PR #5961, opened 2023-05-05
HIP-584: Add MirrorOperationTracer that logs message frame data

### LINKED ISSUE #5951: HIP-584: Add `MessageFrame` interceptor mechanism for debug purposes
### Problem

Currently we have HIP-584 functionality deployed on different environments and we are supporting eth_call and eth_estimateGas. Unfortunately, in case of bugs or unexpected behaviour we don’t have a clear way to determine the problem or to have stack trace with enough depth.

This increases the need of having a way to log as helpful and convenient log messages as possible. A proper way to do this seems to replace the built-in DefaultHederaTracer from hedera-evm with a custom MirrorOperationTracer defined in web3 module. This MirrorOperationTracer can contain debug logging capabilities and enables us to track with much detail which messageFrame failed and what meta data this frame contained. This would help us identify which exact contract has failed, what was the sender, recipient, the remaining gas etc.

### Solution

Implement custom `MirrorOperationTracer` with logging capabilities that uses a new property called TracingProperties, that would allows to fine tune what exactly to log.

### Alternatives

_No response_

IvanKavaldzhiev:
**Description**:

**Related issue(s)**:

Fixes #5951 

**Notes for reviewer**:

**Checklist**


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/MirrorOperationTracer.java
```diff
@@ -0,0 +1,74 @@
...
+
+    private final TracingProperties tracingProperties;
+
+    @Override
+    public void tracePostExecution(final MessageFrame currentFrame, final Operation.OperationResult operationResult) {
+        if (tracingProperties.isEnabled()) {
+            if (!tracingProperties.getStatus().isEmpty()
+                    && !tracingProperties
+                            .getStatus()
+                            .contains(currentFrame.getState().name())) {
```
david-bakin-sl, line 45: my suggestion: pull this condition out into a predicate with a name and then make the predicate _positive_, i.e., instead of `!this && !that` make it `this || that` and do the `!` in the if-statement when called.  IMO expressing complex conditions like this in a "positive" way is easier to understand.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/MirrorOperationTracer.java
```diff
@@ -0,0 +1,74 @@
...
+                return;
+            }
+
+            final var recipientAddress = currentFrame.getRecipientAddress();
+            final var recipientNum = entityIdFromEvmAddress(recipientAddress);
+
+            if (!tracingProperties.getContract().isEmpty()
+                    && !tracingProperties.getContract().contains(recipientNum)) {
```
david-bakin-sl, line 53: Same comment for this one.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/MirrorOperationTracer.java
```diff
@@ -0,0 +1,74 @@
...
+                return;
+            }
+
+            final var inputData = new String(
+                    currentFrame.getInputData() != null
+                            ? currentFrame.getInputData().toArray()
+                            : new byte[0]);
+            log.debug(
```
david-bakin-sl, line 61: Do you want to make this one of those logging things where you first test to see if level debug is enabled before you start formatting strings?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/MirrorOperationTracer.java
```diff
@@ -0,0 +1,74 @@
...
+                    "MessageFrame {} has input data {}, call depth {}, remaining gas {}, sender address {}, recipient address {}, contract address {} and revert reason {}",
```
david-bakin-sl, line 62: I don't know if this is the practice in the mirror node, but for a logging statement like this where multiple fields are logged I suggest using a more "standardized" format like "topic: k1=v1 k2=v2 k3=v3` that logging services have an easier time extracting data from.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/TracingProperties.java
```diff
@@ -0,0 +1,39 @@
...
+
+    @Getter
+    boolean enabled = false;
+
+    @Getter
+    Set<Long> contract = new HashSet<>();
+
+    @Getter
+    Set<String> status = new HashSet<>();
+}
```
david-bakin-sl, line 39: This new class designed for the purpose would be a good place to put those two predicates I suggested above.

david-bakin-sl (changes requested):
Especially needs a test and please don't forget to test the _logging_ statement actually logs the data you think it logs, in the correct formats.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/TracingProperties.java
```diff
@@ -0,0 +1,47 @@
...
+    boolean enabled = false;
+
+    @Getter
+    Set<Long> contract = new HashSet<>();
+
+    @Getter
+    Set<String> status = new HashSet<>();
+
+    public boolean isStateFilterMatches(String state) {
+        return getStatus().isEmpty() || getStatus().contains(state);
```
david-bakin-sl, line 41: Oh darn. oops! _now_ I get what you were talking about in slack.  The null check.  Yeah, you were right, this doesn't work as now written, sorry!  But it's good that it got pulled out as a one-line predicate - still makes it easier to read.  (And below too.)

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/MirrorOperationTracer.java
```diff
@@ -0,0 +1,77 @@
...
+@RequiredArgsConstructor
+public class MirrorOperationTracer implements HederaEvmOperationTracer {
+
+    private final TracingProperties tracingProperties;
+
+    @Override
+    public void tracePostExecution(final MessageFrame currentFrame, final Operation.OperationResult operationResult) {
+        if (tracingProperties.isEnabled()) {
+            if (!tracingProperties.isStateFilterMatches(currentFrame.getState().name())) {
+                return;
```
david-bakin-sl, line 43: IMO one thing language designers could learn from Perl - possibly the _only_ thing - is the suffix-conditional.  (Don't know what it is really called.)

`return if <some condition>;`

`log "this string" if <some condition>;`

Many times that would read more naturally.

david-bakin-sl (commented):
Well.  I misled (*) you in a previous comment.  Sorry about that!

----
(*) Was flat out wrong.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmTxProcessorFacadeImpl.java
```diff
@@ -53,6 +55,7 @@ public class MirrorEvmTxProcessorFacadeImpl implements MirrorEvmTxProcessorFacad
     public MirrorEvmTxProcessorFacadeImpl(
             final MirrorEntityAccess entityAccess,
             final MirrorNodeEvmProperties evmProperties,
+            final TracingProperties tracingProperties,
```
steven-sheehy, line 58: `MirrorOperationTracer` is a bean, so inject it directly and don't leak its dependencies. It's probably the only `HederaEvmOperationTracer` bean so you can also inject the higher level interface.

```suggestion
            final MirrorOperationTracer operationTracer,
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/MirrorOperationTracer.java
```diff
@@ -0,0 +1,71 @@
...
+@CustomLog
+@Named
+@RequiredArgsConstructor
+public class MirrorOperationTracer implements HederaEvmOperationTracer {
+
+    private final TracingProperties tracingProperties;
+
+    @Override
+    public void tracePostExecution(final MessageFrame currentFrame, final Operation.OperationResult operationResult) {
+        if (tracingProperties.isEnabled()) {
```
steven-sheehy, line 41: Fail fast to reduce cognitive complexity and nested conditionals.

```suggestion
        if (!tracingProperties.isEnabled()) {
            return;
        }
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/TracingProperties.java
```diff
@@ -0,0 +1,47 @@
...
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.validation.annotation.Validated;
+
+@Setter
+@Validated
+@ConfigurationProperties(prefix = "hedera.mirror.web3.evm.tracing")
+public class TracingProperties {
+
+    @Getter
+    boolean enabled = false;
```
steven-sheehy, line 32: Remove `@Getter` and `@Setter` everywhere and just use `@Data`.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/MirrorOperationTracer.java
```diff
@@ -0,0 +1,71 @@
...
+            if (!tracingProperties.isStateFilterMatches(currentFrame.getState().name())) {
+                return;
+            }
+
+            final var recipientAddress = currentFrame.getRecipientAddress();
+            final var recipientNum = entityIdNumFromEvmAddress(recipientAddress);
```
steven-sheehy, line 47: Will all recipient addresses be the long zero form and are safe to convert to shard/realm/num? I would've guessed some would be the full random 20 byte evm address. If not, we should make the properties take in a validated collection of 20 byte string with optional 0x prefix.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/TracingProperties.java
```diff
@@ -0,0 +1,47 @@
...
+
+    @Getter
+    Set<Long> contract = new HashSet<>();
```
steven-sheehy, line 35: Unless we need to support EVM address (see other comment), we should change this to `Set<EntityId>` as is our standard in properties elsewhere.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/TracingProperties.java
```diff
@@ -0,0 +1,47 @@
...
+
+    @Getter
+    Set<String> status = new HashSet<>();
```
steven-sheehy, line 38: Both sets need to be marked non-null.
```suggestion
    @NotNull
    private Set<String> status = new HashSet<>();
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/MirrorOperationTracer.java
```diff
@@ -0,0 +1,71 @@
...
+
+            if (!tracingProperties.isContractFilterMatches(recipientNum)) {
+                return;
+            }
+
+            final var inputData = new String(
+                    currentFrame.getInputData() != null
+                            ? currentFrame.getInputData().toArray()
+                            : new byte[0]);
```
steven-sheehy, line 56: Can it always be converted to string? I think we should convert any binary fields to hex instead

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/MirrorOperationTracer.java
```diff
@@ -0,0 +1,71 @@
...
+
+            log.info(
+                    "contractCall: messageFrame={} inputData={}, callDepth={}, remainingGas={}, senderAddress {}, recipientAddress={}, contractAddress={}, revertReason={}",
```
steven-sheehy, line 59: messageFrame is missing `,` and senderAddress is missing `=` like the others

steven-sheehy, line 59: Can omit `contractCall: `

steven-sheehy, line 59: We should have call index as well. Ideally we'd just have a `0.0.1.3` type call depth and index string at the beginning like hashscan.

steven-sheehy, line 59: Omit `Address` everywhere since it's redundant.

steven-sheehy (changes requested):

steven-sheehy:
Please let us know when this ready for another round of reviews by marking it as ready.

docs/configuration.md
```diff
@@ -553,28 +553,31 @@ to configure the application.
...
+`hedera.mirror.web3.evm.maxGasEstimateRetriesCount`         | 20                                                 | Estimate gas' contract call retry treshold 
+`hedera.mirror.web3.evm.maxGasRefundPercentage`             | 20%                                                | Maximal procent of gas refunding
+`hedera.mirror.web3.evm.network`                            | TESTNET                                            | Which Hedera network to use. Can be either `MAINNET`, `PREVIEWNET`, `TESTNET` or `OTHER`
+`hedera.mirror.web3.evm.rateLimit`                          | 100s                                               | Maximum RPS limit
+`hedera.mirror.web3.evm.tracing.enabled`                    | false                                              | Flag enabling tracer
+`hedera.mirror.web3.evm.tracing.contract`                   |                                                    | A set with contract addresses to filter
+`hedera.mirror.web3.evm.tracing.status`                     |                                                    | A set with frame statuses to filter
```
steven-sheehy, line 583:
```suggestion
`hedera.mirror.web3.evm.tracing.contract`                   | []                                                 | A set with contract addresses to filter. By default it is empty to indicate it will trace all contract addresses.
`hedera.mirror.web3.evm.tracing.status`                     | []                                                 | A set with frame statuses to filter. By default it is empty to indicate it will trace all frames regardless of status.
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/TracingProperties.java
```diff
@@ -0,0 +1,47 @@
...
+
+@Data
+@Validated
+@ConfigurationProperties(prefix = "hedera.mirror.web3.evm.tracing")
+public class TracingProperties {
+
+    private boolean enabled = false;
+
+    @NonNull
+    private Set<Address> contract = new HashSet<>();
```
steven-sheehy, line 35: Have you tested this at runtime by populating the address in an `application.yml`? I don't think this will work without some custom converter. Instead, model as a String that supports and validates hex. Please test the below works as you would expect.

```suggestion
    private Set<@Hex(minLength = ADDRESS_LENGTH, maxLength = ADDRESS_LENGTH) String> contract = new HashSet<>();
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/TracingProperties.java
```diff
@@ -0,0 +1,47 @@
...
+import lombok.Data;
+import lombok.NonNull;
+import org.hyperledger.besu.datatypes.Address;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.validation.annotation.Validated;
+
+@Data
+@Validated
+@ConfigurationProperties(prefix = "hedera.mirror.web3.evm.tracing")
+public class TracingProperties {
```
steven-sheehy, line 30: Think we should not use a gerund form of the verb. Adjust other places as well.

```suggestion
@ConfigurationProperties(prefix = "hedera.mirror.web3.evm.trace")
public class TraceProperties {
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/TracingProperties.java
```diff
@@ -0,0 +1,47 @@
...
+
+    private boolean enabled = false;
+
+    @NonNull
+    private Set<Address> contract = new HashSet<>();
+
+    @NonNull
+    private Set<String> status = new HashSet<>();
```
steven-sheehy, line 38: Is there any reason to not use the Besu enum?

```suggestion
    private Set<State> status = new HashSet<>();
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/MirrorOperationTracer.java
```diff
@@ -0,0 +1,89 @@
...
+        }
+    }
+
+    @Override
+    public void tracePostExecution(final MessageFrame currentFrame, final Operation.OperationResult operationResult) {
+        if (!tracingProperties.isEnabled()) {
+            return;
+        }
+        if (tracingProperties.stateFilterCheck(currentFrame.getState().name())) {
+            return;
```
steven-sheehy, line 54: Need separate tests for all the return statements in this method.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/MirrorOperationTracer.java
```diff
@@ -0,0 +1,89 @@
...
+        if (frameState == State.CODE_SUSPENDED) {
+            final String childIndex = "0.0.1." + currentFrame.getMessageStackDepth() + " " + currentFrame.getType();
+            trace(currentFrame, childIndex);
+        }
+    }
+
+    public void trace(final MessageFrame currentFrame, String index) {
+        final var inputData = currentFrame.getInputData() != null
+                ? currentFrame.getInputData().toHexString()
+                : "0x";
```
steven-sheehy, line 74: I don't think this is necessary. Can just directly log `currentFrame.getInputData()` and it will either log null or implicitly call toString which calls toHexString.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/MirrorOperationTracer.java
```diff
@@ -0,0 +1,89 @@
...
+                index
+                        + " messageFrame={}, callDepth={}, remainingGas={}, sender={}, recipient={}, contract={}, revertReason={}, inputData={}",
+                currentFrame.toString(),
+                currentFrame.getMessageStackDepth(),
+                currentFrame.getRemainingGas(),
+                currentFrame.getSenderAddress(),
+                currentFrame.getRecipientAddress(),
+                currentFrame.getContractAddress(),
+                StringUtils.toEncodedString(
+                        currentFrame.getRevertReason().orElse(Bytes.EMPTY).toArray(), StandardCharsets.UTF_16),
```
steven-sheehy, line 86: Is it more likely to be human readable ascii or unreadable binary? Should we just always use hex since it can sometimes be binary?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/MirrorOperationTracer.java
```diff
@@ -0,0 +1,89 @@
...
+        final var recipientAddress = currentFrame.getRecipientAddress();
+        final var recipientNum = mirrorEvmContractAliases.resolveForEvm(recipientAddress);
+
+        if (tracingProperties.contractFilterCheck(recipientNum)) {
+            return;
+        }
+
+        final var frameState = currentFrame.getState();
+        if (frameState == State.CODE_SUSPENDED) {
+            final String childIndex = "0.0.1." + currentFrame.getMessageStackDepth() + " " + currentFrame.getType();
```
steven-sheehy, line 66: The index seems to always start with `0.0.1`, which doesn't really make sense. I was wanting a [hashscan](https://hashscan.io/testnet/transaction/1684266514.414639003) style 1_2_1 call depth plus index prefix that varies for every log written. Not sure if it's possible without tracking prior frames.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/MirrorOperationTracer.java
```diff
@@ -0,0 +1,89 @@
...
+    public void trace(final MessageFrame currentFrame, String index) {
+        final var inputData = currentFrame.getInputData() != null
+                ? currentFrame.getInputData().toHexString()
+                : "0x";
+
+        log.info(
+                index
+                        + " messageFrame={}, callDepth={}, remainingGas={}, sender={}, recipient={}, contract={}, revertReason={}, inputData={}",
+                currentFrame.toString(),
+                currentFrame.getMessageStackDepth(),
```
steven-sheehy, line 80: Call depth is duplicate in the prefix.

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/TracingProperties.java
steven-sheehy: It is [valid](https://docs.jboss.org/hibernate/stable/validator/reference/en-US/html_single/#container-element-constraints) for use there. You need to add `ElementType.TYPE_USE` to the annotation.

docs/configuration.md
```diff
@@ -553,28 +553,31 @@ to configure the application.
...
+`hedera.mirror.web3.evm.expirationCacheTime`                | 10m                                                | Maximum time for contract bytecode's caching
+`hedera.mirror.web3.evm.fundingAccount`                     | 0x0000000000000000000000000000000000000062         | Default Hedera funding account
+`hedera.mirror.web3.evm.maxGasEstimateRetriesCount`         | 20                                                 | Estimate gas' contract call retry treshold 
+`hedera.mirror.web3.evm.maxGasRefundPercentage`             | 20%                                                | Maximal procent of gas refunding
+`hedera.mirror.web3.evm.network`                            | TESTNET                                            | Which Hedera network to use. Can be either `MAINNET`, `PREVIEWNET`, `TESTNET` or `OTHER`
+`hedera.mirror.web3.evm.rateLimit`                          | 100s                                               | Maximum RPS limit
+`hedera.mirror.web3.evm.tracing.enabled`                    | false                                              | Flag enabling tracer
```
steven-sheehy, line 581: These should be `trace`

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/MirrorOperationTracerTest.java
```diff
@@ -0,0 +1,133 @@
...
+
+        assertThat(output).isEmpty();
+    }
+
+    @Test
+    void tracePostExecution(CapturedOutput output) {
+        mirrorOperationTracer = new MirrorOperationTracer(traceProperties, mirrorEvmContractAliases);
+        given(traceProperties.isEnabled()).willReturn(true);
+        given(traceProperties.stateFilterCheck(any())).willReturn(false);
+        given(traceProperties.contractFilterCheck(any())).willReturn(false);
```
steven-sheehy, line 91: Change TraceProperties to not be a mock to provide test coverage

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/MirrorOperationTracerTest.java
```diff
@@ -0,0 +1,133 @@
...
+    private static final Address contract = Address.fromHexString("0x2");
+    private static final Address recipient = Address.fromHexString("0x3");
+    private static final Address sender = Address.fromHexString("0x4");
+    private static final long initialGas = 1000L;
+    private static final Bytes input = Bytes.of("inputData".getBytes());
+
+    @Test
+    void stateFilterTest(CapturedOutput output) {
+        final var topLevelMessageFrame = mock(MessageFrame.class);
+        mirrorOperationTracer = new MirrorOperationTracer(traceProperties, mirrorEvmContractAliases);
```
steven-sheehy, line 62: Move to a `@BeforeEach` or `@InjectMocks`.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/MirrorOperationTracerTest.java
```diff
@@ -0,0 +1,133 @@
...
+        given(topLevelMessageFrame.getState()).willReturn(State.CODE_SUSPENDED);
+
+        mirrorOperationTracer.tracePostExecution(topLevelMessageFrame, operationResult);
+
+        assertThat(output).isEmpty();
+    }
+
+    @Test
+    void contractFilterTest(CapturedOutput output) {
+        final var topLevelMessageFrame = mock(MessageFrame.class);
```
steven-sheehy, line 74: Move to a `@Mock` or a `@BeforeEach`

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/MirrorOperationTracerTest.java
```diff
@@ -0,0 +1,133 @@
...
+    private MirrorEvmContractAliases mirrorEvmContractAliases;
+
+    private MirrorOperationTracer mirrorOperationTracer;
+
+    private static final Address contract = Address.fromHexString("0x2");
+    private static final Address recipient = Address.fromHexString("0x3");
+    private static final Address sender = Address.fromHexString("0x4");
+    private static final long initialGas = 1000L;
+    private static final Bytes input = Bytes.of("inputData".getBytes());
+
```
steven-sheehy, line 58: Need tests for:

* Tracing not enabled.
* state property not empty and passes
* state property not empty and not passes
* contract property not empty and passes
* contract property not empty and not passes
* All checks pass but not frame suspended

Consider using parameterized tests for some of these.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/MirrorOperationTracer.java
```diff
@@ -0,0 +1,81 @@
...
+    public void tracePostExecution(final MessageFrame currentFrame, final Operation.OperationResult operationResult) {
+        if (!traceProperties.isEnabled()) {
+            return;
+        }
+        if (traceProperties.stateFilterCheck(currentFrame.getState())) {
+            return;
+        }
+
+        final var recipientAddress = currentFrame.getRecipientAddress();
+        final var recipientNum = mirrorEvmContractAliases.resolveForEvm(recipientAddress);
```
steven-sheehy, line 56: Why do we need to resolve this address and can't use it directly?

steven-sheehy (changes requested):

steven-sheehy (dismissed):
LGTM

xin-hedera (approved):
LGTM

steven-sheehy (approved):

xin-hedera (approved):

PR #5962, opened 2023-05-05
Add Runbook Restore Document

### LINKED ISSUE #4805: Create a runbook to document disaster recovery steps
### Problem

As a mirror node operator, I would like to know the general steps to take in case I need to revert the mirror node for disaster recovery.

### Solution

* Add a `docs/runbook/restore.md` markdown document that documents the steps necessary to revert to a prior snapshot

### Alternatives

_No response_

edwin-greene:
**Description**:
Add the runbook restore document.

**Related issue(s)**:

Fixes #4805 

**Notes for reviewer**:

**Checklist**


docs/runbook/restore.md
```diff
@@ -0,0 +1,74 @@
...
+   ```
+
+4. Verify that the database has stopped receiving data.
+
+   ```shell
+   mirror_node=> select max(consensus_end) from record_file;
+    1683300839955833003      --> Friday, May 5, 2023 15:33:59.955
+   ```
+
+5. In the Google Console, restore the database from the backup.
```
edwin-greene, line 45: Not sure if there should be more detail for this step. There are command line commands to restore the database, but clicking through the Google Console worked as well.

docs/runbook/restore.md
```diff
@@ -0,0 +1,74 @@
...
+
+6. Verify that the database has been restored to a prior point in time.
+
+   ```shell
+   mirror_node=> select max(consensus_end) from record_file;
+    1683300509619621003      --> Friday, May 5, 2023 15:28:29.619
+   ```
+
+7. Restart the Importer Pod(s) using the number of replicas from step 1.
```
steven-sheehy, line 54: We're missing a step to restart all pods to clear their cache. This should be done before restart importer and monitor since it would be inefficient to restart them after just starting them.

`kubectl delete pods $(kubectl get pod -o name)`

Once we run statefulsets like citus in the namespace, we should revisit this command and maybe add specific labels to pods we want to restart.

docs/runbook/restore.md
steven-sheehy, line 45: We should provide some guidance on which backup to choose as there can be multiple.

steven-sheehy (changes requested):

docs/runbook/restore.md
```diff
@@ -0,0 +1,74 @@
...
+   [pod/mirror-importer-54c477748d-mmxhm/importer] 2023-05-05T09:45:13.177-0600 INFO scheduling-4 c.h.m.i.c.MirrorDateRangePropertiesProcessor RECORD: downloader will download files in time range (2023-05-05T15:28:28.044896003Z, 2262-04-11T23:47:16.854775807Z]
+   [pod/mirror-importer-54c477748d-mmxhm/importer] 2023-05-05T09:45:13.177-0600 INFO scheduling-3 c.h.m.i.c.MirrorDateRangePropertiesProcessor BALANCE: downloader will download files in time range (2023-05-05T15:15:00.168045Z, 2262-04-11T23:47:16.854775807Z]
+   ```
+   
+  9. Start the Monitor. 
+
+      ```shell
+      kubectl scale --replicas=1 deployment/mirror-monitor
+      ```
+
```
xin-hedera, line 73: Add a step before 10 which verifies that traffic is being routed to the cluster, can run rest queries a couple of times and check the query appears in the rest pod's log.

docs/runbook/restore.md
edwin-greene: Thanks I've added the images.

docs/runbook/restore.md
```diff
@@ -0,0 +1,74 @@
...
+Consensus nodes have produced invalid data and the Importer has persisted the data. Restore the database to a previous point in time where the data was valid. The Importer will then read the corrected record stream files and persist the corrected data. 
+
+## Setup
+* Set the kubectl context to the correct cluster and namespace. 
+* In a multicluster environment, perform these steps on one cluster (testnet-eu) in full before performing them on the other cluster (testnet-na).
+
+## Procedure
+
+1. Determine the number of Importer replicas. This is number of replicas that will be restored in step 7.
```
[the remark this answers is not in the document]
edwin-greene: I've removed it.

docs/runbook/restore.md
```diff
@@ -0,0 +1,74 @@
...
+
+   ```shell
+   kubectl describe deployment mirror-importer | grep Replicas:
+   Replicas:           1 desired | 1 updated | 1 total | 1 available | 0 unavailable
+   ```
+
+2. Shut down the Monitor.
```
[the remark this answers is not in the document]
edwin-greene: Good idea, I have added that.

docs/runbook/restore.md
```diff
@@ -0,0 +1,74 @@
...
+   ```shell
+   [pod/mirror-importer-54c477748d-nrsjm/importer] 2023-05-05T09:34:01.470-0600 INFO scheduling-4 c.h.m.i.p.r.RecordFileParser Successfully processed 169 items from 2023-05-05T15_33_58.979940142Z.rcd.gz in 26.32 ms
+   [pod/mirror-importer-54c477748d-nrsjm/importer] 2023-05-05T09:34:01.895-0600 INFO Thread-6 c.h.m.i.u.ShutdownHelper Shutting down.......waiting 10s for internal processes to stop.
+   ```
+
+4. Verify that the database has stopped receiving data.
+
+   ```shell
+   mirror_node=> select max(consensus_end) from record_file;
+    1683300839955833003      --> Friday, May 5, 2023 15:33:59.955
```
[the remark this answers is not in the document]
edwin-greene: Sounds good, I have substituted the rest api query in place of the db query.

docs/runbook/restore.md
edwin-greene: I tried that command but got an error: 

```
kubectl delete pods $(kubectl get pod -o name)
error: there is no need to specify a resource type as a separate argument when passing arguments in resource/name form (e.g. 'kubectl get resource/<resource_name>' instead of 'kubectl get resource resource/<resource_name>'
```

The command I ended up putting the document was able to delete all pods in the namespace: 
`kubectl delete --all pods`

docs/runbook/restore.md
edwin-greene: With the added images there is some more information about which backup to choose. I'm not sure if we need more details or if it will be something to query Ops about.

docs/runbook/restore.md
edwin-greene: I added this step, but I found that the REST query returned the new information from the DB. Are the results from the query itself sufficient, or should we also check the rest api log statements?

docs/runbook/restore.md
```diff
@@ -0,0 +1,100 @@
...
+
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[<img src="backup1.png" width="250"/>](backup1.png "Google Cloud Backups")
+[<img src="backup2.png" width="250"/>](backup2.png "Restore from Backup")
+
+5. After the Restore process has completed verify that the database has been restored to a prior point in time.
+
+   ```shell
+   mirror_node=> select max(consensus_end) from record_file;
+    1683300509619621003      --> Friday, May 5, 2023 15:28:29.619
+   ```
```
edwin-greene, line 51: With testing I found that the blocks rest api was able to confirm that the database had been reverted, so it may be that that command could be used here as well.

docs/runbook/restore.md
xin-hedera: in a multi-cluster setup, we need to check the rest pod's log to confirm traffic is being routed correctly. you also need to run a the query a couple of times since it can be routed to the other cluster

docs/runbook/restore.md
```diff
@@ -0,0 +1,100 @@
...
+   The Importer logs show the shutdown and the last processed record file: 
+   ```shell
+   [pod/mirror-importer-54c477748d-mmxhm/importer] 2023-05-10T10:38:31.723-0600 INFO scheduling-3 c.h.m.i.p.r.RecordFileParser Successfully processed 174 items from 2023-05-10T16_38_28.618260003Z.rcd.gz in 23.11 ms
+   [pod/mirror-importer-54c477748d-mmxhm/importer] 2023-05-10T10:38:31.896-0600 INFO Thread-6 c.h.m.i.u.ShutdownHelper Shutting down.......waiting 10s for internal processes to stop.
+   ```
+
+3. Verify that the Importer has stopped persisting data by querying the REST API multiple times. The block `number` and record file `name` should stop on the last processed record file:
+
+   ```shell
+    curl "https://<mirror-rest-ingress>/api/v1/blocks?limit=1" | json_pp | grep -E "name|number"
```
steven-sheehy, line 35: json_pp is not one of our standard tools. We should prefer jq which can also do it one command. I think just name is sufficient to check.

```suggestion
    curl -sL "https://<hostname>/api/v1/blocks?limit=1&order=desc" | jq -r '.blocks[0].name'
```

docs/runbook/restore.md
steven-sheehy: Yes, use the REST API for both to simplify. Can even say re-run step 3 to verify recent blocks are showing up instead of repeating it.

docs/runbook/restore.md
```diff
@@ -0,0 +1,100 @@
...
+   
+    "name" : "2023-05-10T16_38_28.618260003Z.rcd.gz",
+    "number" : 820492,
+   ``` 
+
+4. In the Google Console, restore the database from the backup. Restore to a backup that was created before the invalid data was persisted.
```
steven-sheehy, line 41: You need to test this scenario for a db with a replica. GCP does not allow restore to instances with a replica and you first have to delete the replica. Please add these steps since any scenario where we execute this runbook would be a production env with replicas. Show images for these steps.

steven-sheehy (changes requested):

docs/runbook/restore.md
steven-sheehy: Don't forget re-adding the replica with the same config as previous and updating any app db config with the new IP

docs/runbook/restore.md
edwin-greene: Thanks I have changed it to use `jq`

docs/runbook/restore.md
edwin-greene: Thanks I have changed the last steps and added details about checking the rest pod logs.

docs/runbook/restore.md
edwin-greene: I've updated the steps with info about the replicas, but I still need to research the part about updating any app db configs with a new IP.

docs/runbook/restore.md
```diff
@@ -0,0 +1,121 @@
...
+
+3. Verify that the Importer has stopped persisting data by querying the REST API multiple times. The record file name should stop changing and show the name of the last processed record file:
+
+   ```shell
+    curl -sL "https://<hostname>/api/v1/blocks?limit=1&order=desc" | jq -r '.blocks[0].name'
+    2023-05-10T16_38_28.618260003Z.rcd.gz
+   ``` 
+4. In the Google Console, click on the database replica and note all the values in the Configuration section. It may be helpful to take a screenshot of the values and to click Edit and view every section in detail as the values will be used to recreate the Replica in step 8. 
+
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[<img src="replicaConfig.png" width="100" height="150"/>](replicaConfig.png "Google Cloud Replica Config")
```
steven-sheehy, line 40: All the images are really small when rendered. Perhaps let them each fill up the space horizontally by just omitting width and height. If it is too big, then maybe just set a width that's close to the 120 char limit.

docs/runbook/restore.md
```diff
@@ -0,0 +1,121 @@
...
+- Data Protection: (Enable deletion protection checked or unchecked)
+- Flags: Each flags and its value
+- Query Insights (Values for each of the checkboxes)
+- Labels
+
+5. In the Google Console Delete the Replica.
+
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[<img src="deleteReplica1.png" width="250" height="150"/>](deleteReplica1.png "Google Cloud Replica Delete part 1")
+[<img src="deleteReplica2.png" width="250" height="150"/>](deleteReplica2.png "Google Cloud Replica Delete part 2")
+6. In the Google Console, restore the database from the backup. Restore to a backup that was created before the invalid data was persisted.
```
steven-sheehy, line 55: The 6 shows as on the same line as the previous image.

docs/runbook/restore.md
```diff
@@ -0,0 +1,121 @@
...
+   [pod/mirror-importer-54c477748d-9kdxr/importer] 2023-05-10T11:04:40.945-0600 INFO scheduling-4 c.h.m.i.c.MirrorDateRangePropertiesProcessor RECORD: downloader will download files in time range (2023-05-05T15:28:28.044896003Z, 2262-04-11T23:47:16.854775807Z]
+   [pod/mirror-importer-54c477748d-9kdxr/importer] 2023-05-10T11:04:40.945-0600 INFO scheduling-3 c.h.m.i.c.MirrorDateRangePropertiesProcessor BALANCE: downloader will download files in time range (2023-05-05T15:15:00.168045Z, 2262-04-11T23:47:16.854775807Z]
+   ```
+   
+12. Start the Monitor. 
+
+    ```shell
+    kubectl scale --replicas=1 deployment/mirror-monitor
+    ```
+13. Get the names of the REST pods for use in step 14.
```
steven-sheehy, line 101: Omit this step and just use `kubectl logs -f --prefix -l app.kubernetes.io/name=rest` in the next step.

docs/runbook/restore.md
```diff
@@ -0,0 +1,121 @@
...
+
+- Verify that traffic is being routed to the cluster by looking for the pod name in the logs.
+- Verify that the record file is at a point in time prior to the date of the invalid data.
+
+
+&nbsp;The REST pods logs should show that the query is being routed to the cluster: 
+   ```shell
+   [pod/mirror-rest-bb68f656-j6bwd/rest] 2023-05-10T11:10:10.134Z INFO 06534719 GET /api/v1/blocks?limit=1 in 5 ms: 200
+   [pod/mirror-rest-bb68f656-vqb8g/rest] 2023-05-10T11:10:17.183Z INFO 194ee756 GET /api/v1/blocks?limit=1 in 4 ms: 200
+   ```
```
steven-sheehy, line 119: This doesn't quite always work. The load balancer routes based upon region so if you're in the NA region and try to test this against the EU cluster you'll always be routed to NA. You can always just verify against a port-forwarded rest service.

docs/runbook/restore.md
steven-sheehy: It might be a bit tricky to show since it's within a sealed secret encrypted file and flux is used to roll it out. Perhaps just state that if the IP changed to update any config using the old IP and leave the details out of scope.

steven-sheehy (changes requested):

docs/runbook/restore.md
edwin-greene: Sounds good I have removed the size limitations. The images seem to automatically shift to a new line depending on the size of the window.

docs/runbook/restore.md
edwin-greene: I've added a space.

docs/runbook/restore.md
edwin-greene: Thanks I have updated it with that kubectl command.

docs/runbook/restore.md
edwin-greene: Sounds good I have added some details to note the IP address value and check if it has changed and then to update other configs.

docs/runbook/restore.md
edwin-greene: Thanks I have updated the documentation with instructions to use the port forward.

jnels124 (approved):
LGTM

steven-sheehy (approved):
LGTM

xin-hedera (approved):
LGTM

PR #5979, opened 2023-05-09
Add support for EVM address to balances endpoint

### LINKED ISSUE #5863: Add support for evmAddress to Balances endpoint
### Problem

Current the balances endpoint `account.id` param only support accounts identifiers of the HederaAccountId.
In some cases calls to Hedera APIs will use the evm address (not a long zero).
In this case developers are foced to first call teh `api/v1/accounts` endpoint to get the AccountID before calling for balances (since the accounts endpoint doesn't support historical timestamps)

### Solution

Update the logic for `account.id` to manage evmAddresses also by utilizing the internal mapping - same logic used on `api/v1/accounts`.
This will the reduce for an extra call in some situations.

### Alternatives

_No response_

edwin-greene:
**Description**:
The balances endpoint now supports a single account.id value in the form of an evm address or an alias.

**Related issue(s)**:

Fixes #5863

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-rest/__tests__/specs/balances/invalid-param-values.json
+++ b/hedera-mirror-rest/__tests__/specs/balances/invalid-param-values.json
@@ -58,7 +58,7 @@
         "_status": {
           "messages": [
             {
-              "message": "Invalid operator. Ethereum accounts only support equals operator."
+              "message": "Invalid operator. Evm address only supports equals operator."
             }
           ]
         }
@@ -75,7 +75,7 @@
         "_status": {
           "messages": [
             {
-              "message": "Only one ethereum address is allowed."
+              "message": "Only one evm address is allowed."
             }
           ]
         }
--- a/hedera-mirror-rest/__tests__/specs/balances/invalid-param-values.json
+++ b/hedera-mirror-rest/__tests__/specs/balances/invalid-param-values.json
@@ -27,9 +27,6 @@
         "_status": {
           "messages": [
             {
-              "message": "Invalid parameter: account.id"
-            },
-            {
               "message": "Invalid parameter: account.balance"
             },
             {
@@ -43,6 +40,9 @@
             },
             {
               "message": "Unknown query parameter: token.id"
+            },
+            {
+              "message": "Invalid parameter: idOrAliasOrEvmAddress"
             }
           ]
         }
@@ -60,7 +60,7 @@
         "_status": {
           "messages": [
             {
-              "message": "Invalid operator. Evm address or alias only supports equals operator."
+              "message": "Invalid parameter: idOrAliasOrEvmAddress"
             }
           ]
         }
--- a/hedera-mirror-rest/__tests__/specs/balances/invalid-param-values.json
+++ b/hedera-mirror-rest/__tests__/specs/balances/invalid-param-values.json
@@ -27,6 +27,9 @@
         "_status": {
           "messages": [
             {
+              "message": "Invalid parameter: account.id"
+            },
+            {
               "message": "Invalid parameter: account.balance"
             },
             {
@@ -40,9 +43,6 @@
             },
             {
               "message": "Unknown query parameter: token.id"
-            },
-            {
-              "message": "Invalid parameter: idOrAliasOrEvmAddress"
             }
           ]
         }
@@ -51,8 +51,9 @@
     {
       "urls": [
         "/api/v1/balances?account.id=gte:ac384c53f03855fa1b3616052f8ba32c6c2a2fec",
-        "/api/v1/balances?account.id=lt:ac384c53f03855fa1b3616052f8ba32c6c2a2fff",
+        "/api/v1/balances?account.id=lt:0xac384c53f03855fa1b3616052f8ba32c6c2a2fff",
         "/api/v1/balances?account.id=lte:0.0.ac384c53f03855fa1b3616052f8ba32c6c2a2fff",
+        "/api/v1/balances?account.id=gte:0x0000000000000000000000000000000000000320",
         "/api/v1/balances?account.id=lte:KGNABD5L3ZGSRVUCSPDR7TONZSRY3D5OMEBKQMVTD2AC6JL72HMQ"
       ],
       "responseStatus": 400,
@@ -60,7 +61,7 @@
         "_status": {
           "messages": [
             {
-              "message": "Invalid parameter: idOrAliasOrEvmAddress"
+              "message": "Invalid operator. Evm address or alias only supports equals operator."
             }
           ]
         }
@@ -71,6 +72,7 @@
         "/api/v1/balances?account.id=ac384c53f03855fa1b3616052f8ba32c6c2a2fec&account.id=ac384c53f03855fa1b3616052f8ba32c6c2a2fff",
         "/api/v1/balances?account.id=0xac384c53f03855fa1b3616052f8ba32c6c2a2fec&account.id=ac384c53f03855fa1b3616052f8ba32c6c2a2fff",
         "/api/v1/balances?account.id=ac384c53f03855fa1b3616052f8ba32c6c2a2fec&account.id=0.0.ac384c53f03855fa1b3616052f8ba32c6c2a2fff",
+        "/api/v1/balances?account.id=0xac384c53f03855fa1b3616052f8ba32c6c2a2fec&account.id=0x0000000000000000000000000000000000000009",
         "/api/v1/balances?account.id=KGNABD5L3ZGSRVUCSPDR7TONZSRY3D5OMEBKQMVTD2AC6JL72HMQ&account.id=0.0.ac384c53f03855fa1b3616052f8ba32c6c2a2fff"
       ],
       "responseStatus": 400,
--- a/hedera-mirror-rest/__tests__/specs/balances/invalid-param-values.json
+++ b/hedera-mirror-rest/__tests__/specs/balances/invalid-param-values.json
@@ -1,33 +1,85 @@
 {
   "description": "Balance api invalid account id format",
-  "setup": {},
-  "urls": [
-    "/api/v1/balances?account.id=lte:0.1.2.3&account.balance=gt:-2&account.publickey=-2&timestamp=-1.2&order=any&token.id=0.0.45678",
-    "/api/v1/balances?account.id=lte:0.1.2.3&account.balance=gt:9223372036854775808&account.publickey=-2&timestamp=-1.2&order=any,&token.id=0.0.45678"
-  ],
-  "responseStatus": 400,
-  "responseJson": {
-    "_status": {
-      "messages": [
-        {
-          "message": "Invalid parameter: account.id"
-        },
-        {
-          "message": "Invalid parameter: account.balance"
-        },
-        {
-          "message": "Invalid parameter: account.publickey"
-        },
-        {
-          "message": "Invalid parameter: timestamp"
-        },
-        {
-          "message": "Invalid parameter: order"
-        },
-        {
-          "message": "Unknown query parameter: token.id"
+  "setup": {
+    "accounts": [
+      {
+        "evm_address": "ac384c53f03855fa1b3616052f8ba32c6c2a2fec",
+        "num": 7
+      },
+      {
+        "num": 8
+      },
+      {
+        "evm_address": "ac384c53f03855fa1b3616052f8ba32c6c2a2fff",
+        "num": 9
+      }
+    ]
+  },
+  "tests": [
+    {
+      "urls": [
+        "/api/v1/balances?account.id=lte:0.1.2.3&account.balance=gt:-2&account.publickey=-2&timestamp=-1.2&order=any&token.id=0.0.45678",
+        "/api/v1/balances?account.id=lte:0.1.2.3&account.balance=gt:9223372036854775808&account.publickey=-2&timestamp=-1.2&order=any,&token.id=0.0.45678"
+      ],
+      "responseStatus": 400,
+      "responseJson": {
+        "_status": {
+          "messages": [
+            {
+              "message": "Invalid parameter: account.id"
+            },
+            {
+              "message": "Invalid parameter: account.balance"
+            },
+            {
+              "message": "Invalid parameter: account.publickey"
+            },
+            {
+              "message": "Invalid parameter: timestamp"
+            },
+            {
+              "message": "Invalid parameter: order"
+            },
+            {
+              "message": "Unknown query parameter: token.id"
+            }
+          ]
         }
-      ]
+      }
+    },
+    {
+      "urls": [
+        "/api/v1/balances?account.id=gte:ac384c53f03855fa1b3616052f8ba32c6c2a2fec",
+        "/api/v1/balances?account.id=lt:ac384c53f03855fa1b3616052f8ba32c6c2a2fff",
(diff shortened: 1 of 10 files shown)


hedera-mirror-rest/balances.js
```diff
@@ -72,10 +74,16 @@ const entityJoin = `join (select id, public_key from entity where type in ('ACCO
  * @return {Promise} Promise for PostgreSQL query
  */
 const getBalances = async (req, res) => {
+  if (req.query[constants.filterKeys.ACCOUNT_ID] !== undefined) {
+    // Rename account-id to id-or-alias-or-evm-address
+    req.query[constants.filterKeys.ID_OR_ALIAS_OR_EVM_ADDRESS] = req.query[constants.filterKeys.ACCOUNT_ID];
+    delete req.query[constants.filterKeys.ACCOUNT_ID];
+  }
   utils.validateReq(req, acceptedBalancesParameters);
```
edwin-greene, line 82: utils.validateReq will not allow an alias value for `account-id`, so rather than open up the validation on all endpoints to allow an alias, this switches the query parameter property name and then does validation for `ID_OR_ALIAS_OR_EVM_ADDRESS` and the `eq` operator. 

An alternative would be to change the endpoint to no longer support a query parameter named `account-id` and introduce a new query parameter, but this would be a breaking change to consumers of the balances api.

hedera-mirror-rest/balances.js
```diff
@@ -72,10 +74,16 @@ const entityJoin = `join (select id, public_key from entity where type in ('ACCO
  * @return {Promise} Promise for PostgreSQL query
  */
 const getBalances = async (req, res) => {
+  if (req.query[constants.filterKeys.ACCOUNT_ID] !== undefined) {
```
Nana-EC, line 77: Q: I forget the logic but why do a remaping of the query object here vs consider the value as either format and decide based on validation?

hedera-mirror-rest/api/v1/openapi.yml
```diff
@@ -3625,6 +3625,33 @@ components:
...
+          value: 0.8
+        shardRealmAccountNum:
+          value: 0.0.8
+        evmAddress:
+          value: ac384c53f03855fa1b3616052f8ba32c6c2a2fec
+        evmAddressWithPrefix:
+          value: 0xac384c53f03855fa1b3616052f8ba32c6c2a2fec
+        evmAddressWithShardAndRealm:
+          value: 0.0.ac384c53f03855fa1b3616052f8ba32c6c2a2fec
+      schema:
```
Nana-EC, line 3652: You could capture at least type and schema as a `AccountIdOrAliasOrEvmAddress` API object and reuse it in `accountIdOrAliasOrEvmAddressQueryParam` and `accountIdOrAliasOrEvmAddressPathParam`

You can refer to `EntityIdQuery` as an example

hedera-mirror-rest/__tests__/specs/balances/invalid-param-values.json
```diff
@@ -1,33 +1,88 @@
...
-        {
-          "message": "Invalid parameter: order"
-        },
-        {
-          "message": "Unknown query parameter: token.id"
+  "setup": {
+    "accounts": [
+      {
+        "alias": "KGNABD5L3ZGSRVUCSPDR7TONZSRY3D5OMEBKQMVTD2AC6JL72HMQ",
+        "evm_address": "ac384c53f03855fa1b3616052f8ba32c6c2a2fec",
```
Nana-EC, line 7: Q: should these have `0x` prefixes. Not sure if that is removed for db performance

hedera-mirror-rest/__tests__/specs/balances/invalid-param-values.json
```diff
@@ -1,33 +1,88 @@
...
+              "message": "Invalid parameter: idOrAliasOrEvmAddress"
+            }
+          ]
         }
-      ]
+      }
+    },
+    {
+      "urls": [
+        "/api/v1/balances?account.id=gte:ac384c53f03855fa1b3616052f8ba32c6c2a2fec",
```
Nana-EC, line 53: For inputs you should expect `0x` prefixes on evm addresses

hedera-mirror-rest/__tests__/specs/balances/invalid-param-values.json
```diff
@@ -1,33 +1,88 @@
...
+        "/api/v1/balances?account.id=lt:ac384c53f03855fa1b3616052f8ba32c6c2a2fff",
```
Nana-EC, line 54: Please add long zero examples for consideration here and below as users may input either a long zero or evm address alias format. Both are valid 20byte hex addresses in this scenario

Nana-EC (commented):
Nice change.
Left some suggestions and questions for consideratuion

hedera-mirror-rest/__tests__/specs/balances/invalid-param-values.json
```diff
@@ -1,33 +1,88 @@
...
+        "/api/v1/balances?account.id=lte:0.0.ac384c53f03855fa1b3616052f8ba32c6c2a2fff",
+        "/api/v1/balances?account.id=lte:KGNABD5L3ZGSRVUCSPDR7TONZSRY3D5OMEBKQMVTD2AC6JL72HMQ"
+      ],
+      "responseStatus": 400,
+      "responseJson": {
+        "_status": {
+          "messages": [
+            {
+              "message": "Invalid parameter: idOrAliasOrEvmAddress"
```
steven-sheehy, line 63: The parameter name must be the query parameter name the customer supplied so they can properly react to the bad input. It would be even better if it told you why it was invalid like unsupported lte operator for EVM address.

```suggestion
              "message": "Invalid parameter: account.id"
```

hedera-mirror-rest/api/v1/openapi.yml
```diff
@@ -3625,6 +3625,33 @@ components:
...
+        pattern: ^(\d{1,10}\.){0,2}(\d{1,10}|(0x)?[A-Fa-f0-9]{40}|(?:[A-Z2-7]{8})*(?:[A-Z2-7]{2}|[A-Z2-7]{4,5}|[A-Z2-7]{7,8}))$
```
steven-sheehy, line 3653: I don't think we should support the weird and complicated `shard.realm.(alias|evmAddress)` format. Let's just support `alias` or `(0x)?evmAddress`.

steven-sheehy (changes requested):

hedera-mirror-rest/balances.js
edwin-greene: Good idea. I ended up using `utils.validateReq` but passing in a custom validator.

hedera-mirror-rest/api/v1/openapi.yml
edwin-greene: Thanks I have updated the openapi spec, it now references a schema and reuses it for both of the account/alias/evm address objects.

hedera-mirror-rest/__tests__/specs/balances/invalid-param-values.json
edwin-greene: Thanks I have added `0x` to the tests.

hedera-mirror-rest/__tests__/specs/balances/invalid-param-values.json
edwin-greene: Good idea, I have added the long zero evm addresses to the test cases.

hedera-mirror-rest/__tests__/specs/balances/invalid-param-values.json
edwin-greene: Thanks I have fixed this.

hedera-mirror-rest/api/v1/openapi.yml
edwin-greene: Thanks I have updated the openapi spec to simplify the schema.

xin-hedera (changes requested):

hedera-mirror-rest/api/v1/openapi.yml
```diff
@@ -3623,7 +3628,34 @@ components:
...
           value: 0.0.ac384c53f03855fa1b3616052f8ba32c6c2a2fec
       schema:
-        pattern: ^(\d{1,10}\.){0,2}(\d{1,10}|(0x)?[A-Fa-f0-9]{40}|(?:[A-Z2-7]{8})*(?:[A-Z2-7]{2}|[A-Z2-7]{4,5}|[A-Z2-7]{7,8}))$
+        $ref: "#/components/schemas/AccountIdOrAliasOrEvmAddress"
+        type: string
+    accountIdOrAliasOrEvmAddressQueryParam:
+      name: account.id
+      in: query
+      description: Account alias or account id or evm address
+      required: true
```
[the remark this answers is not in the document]
edwin-greene: Good catch. I have removed it.

hedera-mirror-rest/balances.js
```diff
@@ -186,6 +190,69 @@ const getTokenAccountBalanceSubQuery = (order) => {
...
+      break;
+    case constants.filterKeys.TIMESTAMP:
+      ret = utils.isValidTimestampParam(val);
+      break;
+    default:
+      ret = false;
+  }
+
+  return ret;
+};
```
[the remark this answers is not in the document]
edwin-greene: Excellent thanks. I have made these simplifications.

hedera-mirror-rest/balances.js
```diff
@@ -186,6 +190,35 @@ const getTokenAccountBalanceSubQuery = (order) => {
...
 };
 
+const parseAccountIdQueryParam = (query, columnName) => {
+  let evmAliasAddressCount = 0;
+  return utils.parseParams(
+    query[constants.filterKeys.ACCOUNT_ID],
+    (value) => {
+      if (!EntityId.isValidEntityId(value, false) && ++evmAliasAddressCount > 1) {
+        throw new InvalidArgumentError(`Only one evm address or alias is allowed.`);
+      }
```
xin-hedera, line 200: we should do this check in `balanceFilterValidator` without the penalty of querying the db in `EntityService.getEncodedId`, can do it a follow-up ticket

xin-hedera (approved):
LGTM

hedera-mirror-rest/api/v1/openapi.yml
```diff
@@ -1751,6 +1751,11 @@ components:
               $ref: "#/components/schemas/Transactions"
             links:
               $ref: "#/components/schemas/Links"
+    AccountIdOrAliasOrEvmAddress:
+      oneOf:
+        - $ref: "#/components/schemas/Alias"
+        - $ref: "#/components/schemas/EntityId"
+        - $ref: "#/components/schemas/EvmAddress"
```
steven-sheehy, line 1758: Revert this type and places the reference it. We can't use oneof this way for reuse as it causes problems with openapi code generation that I had to previously fix to duplicate patterns in places. Just duplicate the patterns so we don't cause regressions.

hedera-mirror-rest/api/v1/openapi.yml
```diff
@@ -3623,7 +3628,33 @@ components:
...
+        realmAccountNum:
+          value: 0.8
+        shardRealmAccountNum:
+          value: 0.0.8
+        evmAddress:
+          value: ac384c53f03855fa1b3616052f8ba32c6c2a2fec
+        evmAddressWithPrefix:
+          value: 0xac384c53f03855fa1b3616052f8ba32c6c2a2fec
+        evmAddressWithShardAndRealm:
+          value: 0.0.ac384c53f03855fa1b3616052f8ba32c6c2a2fec
```
steven-sheehy, line 3655: Still examples with shard and realm for address and alias

hedera-mirror-rest/balances.js
```diff
@@ -186,6 +190,35 @@ const getTokenAccountBalanceSubQuery = (order) => {
...
+      return EntityService.getEncodedId(value);
+    },
+    (op, value) => {
+      if (evmAliasAddressCount > 0 && op !== utils.opsMap.eq) {
+        throw new InvalidArgumentError(`Invalid operator. Evm address or alias only supports equals operator.`);
+      }
+
+      return Array.isArray(value)
+        ? [`${columnName} IN (?`.concat(', ?'.repeat(value.length - 1)).concat(')'), value]
```
steven-sheehy, line 209: In the ticket I state we should only support exactly one EVM address or alias.

steven-sheehy (changes requested):

hedera-mirror-rest/api/v1/openapi.yml
edwin-greene: Sounds good I have removed it.

hedera-mirror-rest/balances.js
edwin-greene: I have updated the validation, it now has reduced the number of db queries.

hedera-mirror-rest/__tests__/specs/balances/invalid-param-values.json
```diff
@@ -1,33 +1,90 @@
...
+        "/api/v1/balances?account.id=lte:0.0.ac384c53f03855fa1b3616052f8ba32c6c2a2fff",
+        "/api/v1/balances?account.id=gte:0x0000000000000000000000000000000000000320",
+        "/api/v1/balances?account.id=lte:KGNABD5L3ZGSRVUCSPDR7TONZSRY3D5OMEBKQMVTD2AC6JL72HMQ"
+      ],
+      "responseStatus": 400,
+      "responseJson": {
+        "_status": {
+          "messages": [
+            {
+              "message": "Invalid operator. Evm address or alias only supports equals operator."
```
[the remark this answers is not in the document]
edwin-greene: Sounds good I have added the details to the error messages that do not arise from the `utils.validateReq`, for those error messages it may be that a new issue could tackle refactoring to add details for other errors.

hedera-mirror-rest/balances.js
edwin-greene: There is an error that will account for more than one evm address or alias; this line allows for multiple uses of the shard realm num account id.

hedera-mirror-rest/api/v1/openapi.yml
edwin-greene: I understand now, support for them has been removed.

hedera-mirror-rest/__tests__/specs/balances/invalid-param-values.json
```diff
@@ -1,33 +1,104 @@
...
+    {
+      "urls": [
+        "/api/v1/balances?account.balance=gt:-2&account.publickey=-2&timestamp=-1.2&order=any&token.id=0.0.45678",
+        "/api/v1/balances?account.balance=gt:9223372036854775808&account.publickey=-2&timestamp=-1.2&order=any,&token.id=0.0.45678"
+      ],
+      "responseStatus": 400,
+      "responseJson": {
+        "_status": {
+          "messages": [
+            {
```
edwin-greene, line 29: For performance, the account.id parameter is now validated upon parsing the account while these other parameters are validated prior to account parsing. Because of this the test for account id's validity has been removed from this test and a new test has been added for it on lines 86-101.

hedera-mirror-rest/balances.js
steven-sheehy: Got it. Thanks for clarifying.

steven-sheehy (approved):
LGTM

xin-hedera (approved):
LGTM

PR #6006, opened 2023-05-15
HIP-584: Create Token Database Accessor

### LINKED ISSUE #5977: HIP-584: Create Token Database Accessor
### Problem

We need to get a token from the database

### Solution

Create a new accessor, that given an address, aggregates data from relevant repositories and returns a Token.

### Alternatives

_No response_

stoyan-lime:
**Description**:
Create a DB accessor, that given an address gets all relevant data from relevant repositories and returns a Token.

**Related issue(s)**:

Fixes #5977


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/TokenDatabaseAccessor.java
```diff
@@ -16,22 +16,89 @@
...
+                parseJkey(databaseToken.getKycKey()),
+                parseJkey(databaseToken.getFreezeKey()),
+                parseJkey(databaseToken.getSupplyKey()),
+                parseJkey(databaseToken.getWipeKey()),
+                parseJkey(entity.getKey()),
+                parseJkey(databaseToken.getFeeScheduleKey()),
+                parseJkey(databaseToken.getPauseKey()),
+                Boolean.TRUE.equals(databaseToken.getFreezeDefault()),
+                null,
+                null,
```
IvanKavaldzhiev, line 83: We need to add `treasury` and `autoRenewAccount` as well.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/TokenDatabaseAccessor.java
```diff
@@ -16,22 +16,115 @@
...
+
+    private JKey parseJkey(byte[] keyBytes) {
+        try {
+            return keyBytes == null ? null : asFcKeyUnchecked(Key.parseFrom(keyBytes));
+        } catch (InvalidProtocolBufferException e) {
+            return null;
+        }
+    }
+
+    private Account getTreasury(EntityId treasuryId) {
```
IvanKavaldzhiev, line 107: You can add a javadoc noting that we don't load all fields for the treasury for performance reasons, since we don't need all of the fields.

stoyan-lime: Created a treasury class to make it clearer.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/model/Treasury.java
```diff
@@ -0,0 +1,32 @@
...
+
+public class Treasury extends Account {
+
+    /**
+     * Create a treasury account with only ID and balance values, as those are the only currently needed ones
+     * @param id account id
+     * @param balance treasury balance
+     */
+    public Treasury(Id id, long balance) {
+        super(id, 0, balance, false, 0, 0, null, 0, null, null, null, 0, 0, 0);
```
david-bakin-sl, line 30: Jeez it's hard to read this.  You're _forced_ to use an IDE to give you argument prompts.  Not the fault of _this_ class, but the `Account` constructor has 7 ints (integral primitive types), 3 different maps, etc. etc.  13 attributes that _must_ be specified (even if `0`/`null`) given the `id` which you've got to have.

If it is the case for the purpose _here_ we only need 2 of the attributes I suggest _adding_ a constructor or better a factory method to `Account` that just takes the fields we need for a `Treasury` thing.  Factory better because it can be given a decent name.  And it would belong over there in `Accounts` so that it is _close_ to the other constructors (and the class defn too) so you could read it over there in context.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/AccessorUtils.java
```diff
@@ -0,0 +1,40 @@
...
+package com.hedera.mirror.web3.evm.store.accessor;
+
+import com.hedera.mirror.common.domain.entity.Entity;
+import java.sql.Date;
+import java.util.concurrent.TimeUnit;
+import lombok.experimental.UtilityClass;
+
+@UtilityClass
+public class AccessorUtils {
+    public static final long DEFAULT_EXPIRY_TIMESTAMP =
```
david-bakin-sl, line 26: The Lombok documentation says that once you mark a class with `@UtilityClass` you don't have to mark each method (or field) `static`, that's done for you.  Yet I see the mirror node code always marks methods `static` when using `@UtilityClass`.  Why?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/AccessorUtils.java
```diff
@@ -0,0 +1,40 @@
...
+            TimeUnit.MILLISECONDS.toNanos(Date.valueOf("2100-1-1").getTime());
+
+    public static Long getEntityExpiration(Entity entity) {
```
david-bakin-sl, line 29: Why return `Long` not `long`?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/AccessorUtils.java
david-bakin-sl, line 26: I'm surprised this isn't somewhere else.  Or maybe it just belongs somewhere else?  Someplace closer to the definition of tokens? (or entities?)  Just curious.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/EntityDatabaseAccessor.java
```diff
@@ -37,9 +37,13 @@ public class EntityDatabaseAccessor extends DatabaseAccessor<Address, Entity> {
...
         if (isMirror(addressBytes)) {
             final var entityId = entityIdNumFromEvmAddress(address);
-            return entityRepository.findByIdAndDeletedIsFalse(entityId);
+            return getById(entityId);
         } else {
             return entityRepository.findByEvmAddressAndDeletedIsFalse(addressBytes);
         }
     }
+
+    public Optional<Entity> getById(Long entityId) {
```
david-bakin-sl, line 46: `@NonNull` like at L36 above?

david-bakin-sl, line 46: Can be named `get` and rely on overloading? (Otherwise you might ask "why isn't the other one `getByAddress`?")

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/TokenDatabaseAccessor.java
```diff
@@ -16,22 +16,102 @@
...
+                Optional.ofNullable(databaseToken.getDecimals()).orElse(0),
+                Optional.ofNullable(entity.getAutoRenewPeriod()).orElse(0L),
+                0L);
+    }
+
+    private JKey parseJkey(byte[] keyBytes) {
+        try {
+            return keyBytes == null ? null : asFcKeyUnchecked(Key.parseFrom(keyBytes));
+        } catch (InvalidProtocolBufferException e) {
+            return null;
```
david-bakin-sl, line 103: What kinds of problems throw `InvalidProtocolBufferException` here such that `null` is an acceptable value to provide the new `Token`?  Because I note that `asFcKeyUnchecked` is happy to throw `IllegalArgumentException`, not caught here and replaced with `null`, in case of `DecoderException`.  So there's a difference here between some errors that get silently replace by null and some that throw.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/TokenDatabaseAccessor.java
```diff
@@ -16,22 +16,102 @@
...
+    }
+
+    private Token tokenFromEntity(Entity entity) {
+        final var databaseToken =
+                tokenRepository.findById(new TokenId(entity.toEntityId())).orElse(null);
+
+        if (databaseToken == null) {
+            return null;
+        }
+        return new Token(
```
david-bakin-sl, line 62: OMG the `Token` constructor is even worse than `Account` (see my comment on `Treasury` below).  And in this case the IDE (IntelliJ) is no help to you trying to figure out what order the _seven consecutive keys_ are supposed to be presented in.  (It doesn't give you the little "argument name" prompts, don't know why.  Perhaps that's a setting?)

image

Are builders discouraged here?  This is the kind of thing they're good for: building something that takes a bunch of attributes, _especially_ with similar types, because each attribute added is _named_ right there.

david-bakin-sl (commented):
Nothing to change as far as I'm concerned.  Just comments on style and questions about "how things are done" in this codebase.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/AccessorUtils.java
steven-sheehy: Just because you can, doesn't mean you should. It's clearer to mark them as static to show that they actually are static. Just like I don't recommend not putting `private final` for `@Value`. If you ever need to remove Lombok (say to switch to `@Data`) then it requires more work and might introduce bugs.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/TokenDatabaseAccessor.java
steven-sheehy: While I agree with you, these classes come from services so they can't use Lombok. While services could transition them to use builders, that would have to be done upstream first and I imagine that's a large effort since these classes are probably used everywhere.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/model/Treasury.java
steven-sheehy: Would agree with this. Don't think we need a sub-class for treasury, just a factory method.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/store/accessor/TokenDatabaseAccessorTest.java
```diff
@@ -17,33 +17,217 @@
...
+        databaseToken.setSupplyKey(SUPPLY_KEY.toByteArray());
+        databaseToken.setWipeKey(WIPE_KEY.toByteArray());
+        databaseToken.setFeeScheduleKey(FEE_SCHEDULE_KEY.toByteArray());
+        databaseToken.setPauseKey(PAUSE_KEY.toByteArray());
+        databaseToken.setFreezeDefault(FREEZE_DEFAULT);
+        databaseToken.setPauseStatus(TokenPauseStatusEnum.UNPAUSED);
+        databaseToken.setName(NAME);
+        databaseToken.setSymbol(SYMBOL);
+        databaseToken.setDecimals(DECIMALS);
+        when(tokenRepository.findById(any())).thenReturn(Optional.ofNullable(databaseToken));
```
steven-sheehy, line 231: Use `domainBuilder.token().get()`

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/EntityDatabaseAccessor.java
```diff
@@ -37,9 +37,13 @@ public class EntityDatabaseAccessor extends DatabaseAccessor<Address, Entity> {
...
+        return entityRepository.findByIdAndDeletedIsFalse(entityId);
+    }
```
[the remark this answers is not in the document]
stoyan-lime: Removed.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/store/accessor/TokenDatabaseAccessorTest.java
```diff
@@ -17,33 +17,217 @@
...
+            .build();
+    private final Key FEE_SCHEDULE_KEY = Key.newBuilder()
+            .setECDSASecp256K1(ByteString.fromHex("03af80b90d25145da28c583359beb47b21796b2fe1a23c1511e443e7a64dfdb27e"))
+            .build();
+
+    private final Key PAUSE_KEY = Key.newBuilder()
+            .setECDSASecp256K1(ByteString.fromHex("03af80b90d25145da28c583359beb47b21796b2fe1a23c1511e443e7a64dfdb27f"))
+            .build();
+
+    private final Key ADMIN_KEY = Key.newBuilder()
```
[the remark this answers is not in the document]
stoyan-lime: Updated.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/store/accessor/TokenDatabaseAccessorTest.java
stoyan-lime: Updated.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/EntityDatabaseAccessor.java
stoyan-lime: Method is removed.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/TokenDatabaseAccessor.java
stoyan-lime: IllegalArgumentException added to catch.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/AccessorUtils.java
stoyan-lime: I assume you meant moving the whole method to the entity class, so I have done this.

steven-sheehy (approved):
LGTM

hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/entity/AbstractEntity.java
```diff
@@ -159,6 +164,18 @@ public EntityId toEntityId() {
         return new EntityId(shard, realm, num, type);
     }
 
+    public long getEffectiveExpiration() {
```
steven-sheehy, line 167: Needs `@JsonIgnore`.

xin-hedera (approved):
LGTM

hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/entity/AbstractEntity.java
stoyan-lime: Updated.

steven-sheehy (approved):

PR #6017, opened 2023-05-17
HIP-584: Refactoring estimate gas tests: removing hardcoded values

### LINKED ISSUE #5946: HIP-584: Refactor `ContractCallService` tests for `eth_estimateGas` to not rely on a hardcoded expected value
### Problem

Currently in `ContractCallServiceTest` we rely on hardcoded fixed values, that may change over time and be impacted on different changes in the params used or the code base itself. We need a more universal and easier to maintain approach.

### Solution

Instead of using hardcoded expected values for the estimate first run a regular transaction simulation via `MirrorEvmTxProcessorFacade` and save the gasUsed field for `TransactionProcessingResult`. Then run an `eth_estimateGas` request and check if the estimate has an offset between 5% and 20% bigger than the `gasUsed` value.

### Alternatives

_No response_

nikolovyanko:
Rather than relying on fixed expected values for the estimate, the approach involves initiating a regular transaction simulation using `MirrorEvmTxProcessorFacade`. The resulting `gasUsed` value from this simulation is obtained and serves as a reference for estimation. Subsequently, the `eth_estimateGas` request is then checked whether its offset is between 5% and 20% bigger than the `gasUsed` value.

In order to be more accurate with the estimation the `MirrorNodeEvmProperties.estimateGasIterationThreshold` values is changed from `1200` to `3600 `.

**Related issue(s)**:

Fixes #5946


hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServiceTest.java
```diff
@@ -302,11 +317,14 @@ void estimateGasForStateChangeCall() {
...
                 serviceParameters(stateChangeHash, 0, ETH_ESTIMATE_GAS, false, 0, ETH_CALL_CONTRACT_ADDRESS);
-        final var expectedGasUsed = 29233L;
 
         persistEntities(false);
+        final var expectedGasUsed = gasUsedAfterExecution(serviceParameters);
 
-        assertThat(contractCallService.processCall(serviceParameters)).isEqualTo(hexValueOf.apply(expectedGasUsed));
+        assertThat(longValueOf.applyAsLong(contractCallService.processCall(serviceParameters)))
+                .as("result must be within at most 20% bigger than the gas used from the first call")
+                .isGreaterThan(expectedGasUsed)
```
IvanKavaldzhiev, line 326: We should check that the estimate is at least 5% bigger than the `gasUsed` field, not directly bigger than `gasUsed`.

nikolovyanko: Done, I had to change `MirrorNodeEvmProperties.estimateGasIterationThreshold` from 1200 to 2200 to get estimations in the 5-20% range.

IvanKavaldzhiev (dismissed):
LGTM

xin-hedera (dismissed):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
```diff
@@ -50,7 +50,7 @@ public class MirrorNodeEvmProperties implements EvmProperties {
 
     @Getter
     @Positive
-    private long estimateGasIterationThreshold = 1200L;
+    private long estimateGasIterationThreshold = 3600;
```
steven-sheehy: Need to update `docs/configuration.md` as well.

steven-sheehy (approved):

PR #6021, opened 2023-05-17
HIP-584: Add Custom Fees to Token

### LINKED ISSUE #5977: HIP-584: Create Token Database Accessor

stoyan-lime:
**Description**:
When building tokens in TokenDatabaseAccessor, add custom fees as well.

**Related issue(s)**:

Fixes #5977


hedera-mirror-web3/src/main/java/com/hedera/services/store/models/Token.java
```diff
@@ -188,7 +193,8 @@ private Token createNewTokenWithNewTotalSupply(Token oldToken, long totalSupply)
                 oldToken.symbol,
                 oldToken.decimals,
                 oldToken.autoRenewPeriod,
-                oldToken.lastUsedSerialNumber);
+                oldToken.lastUsedSerialNumber,
+                customFees);
```
david-bakin-sl, line 197: I don't get these methods.  They aren't `static` even though they're factories.  So they have access to both `oldToken` and the _current token_.  So why for example, in this method, `totalSupply` and now `customFees` come from the _current_ token, and everything else comes from `oldToken` (and there's nothing in the javadoc to explain this)?????? 

Note that in `createNewTokenWithNewLastUsedSerialNumber` - also not `static` - only the new `customFees` comes from the _current_ token.  All else including `totalSupply` comes from `oldToken`.

I bet these are all supposed to be `static` factories that take a prototype and modify one or two fields. (And if they were: These errors - if they are errors - would be caught.)

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/CustomFeeDatabaseAccessor.java
```diff
@@ -16,22 +16,76 @@
...
+    private final EntityDatabaseAccessor entityDatabaseAccessor;
+
     @Override
-    public @NonNull Optional<List<CustomFee>> get(@NonNull Long key) {
-        return Optional.ofNullable(customFeeRepository.findByTokenId(key));
+    public @NonNull Optional<List<CustomFee>> get(@NonNull Long tokenId) {
+        final var customFeesList = customFeeRepository.findByTokenId(tokenId).stream()
+                .filter(customFee -> customFee.getCollectorAccountId() != null)
+                .map(this::mapCustomFee)
+                .collect(Collectors.toList());
```
david-bakin-sl, line 47: `.collect(Collectors.toList())` -> `toList()`

david-bakin-sl (changes requested):
I think something's really wrong in `Token` w.r.t. the factories.  Marking `request changes` to ensure that either it gets fixed or the javadoc is updated to explain about the "current token" (`this`) vs `oldToken`.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/CustomFeeDatabaseAccessor.java
```diff
@@ -16,22 +16,76 @@
...
+
+        return Optional.of(customFeesList);
+    }
+
+    private CustomFee mapCustomFee(com.hedera.mirror.common.domain.transaction.CustomFee customFee) {
```
xin-hedera, line 52: same code also in `TokenAccessorImpl.getCustomFees`

xin-hedera (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/Token.java
IvanKavaldzhiev: We won't to keep this code base similar to that in services but also achieve immutability - that's why we use several constructors with different arguments. However, in all of the cases I think we should load the `customFees` from the `oldToken`, indeed.

david-bakin-sl: Similar is fine to since they start as copies and you'd like to be able to refer to the originals over in services as they change - but you can _add_ without trouble.  E.g., on top of the existing constructors or methods add _new_ ones that use the old, and then just consistently use the _new_ ones.  To provide better abstraction, better readability, etc.  It would still be clear when doing diffs what's going on, and new mirror node code built on the new stuff would be easier to understand.  (Note: I'm not suggesting wrapper classes - just putting a new version of a constructor/method next to the old one in the same class. With, of course, a comment explaining what's going on.)

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/CustomFeeDatabaseAccessor.java
stoyan-lime: Removed duplication.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/EntityDatabaseAccessor.java
```diff
@@ -46,4 +50,23 @@ public class EntityDatabaseAccessor extends DatabaseAccessor<Address, Entity> {
     public Optional<Entity> getById(Long entityId) {
         return entityRepository.findByIdAndDeletedIsFalse(entityId);
     }
+
+    public Address evmAddressFromId(EntityId entityId) {
```
steven-sheehy, line 54: Duplicates logic in TokenAccessorImpl and needs a test.

steven-sheehy (dismissed):
LGTM

xin-hedera (dismissed):
LGTM

david-bakin-sl (approved):
LGTM

PR #5941, opened 2023-05-02
HIP-584: Disable precompiles for eth_estimateGas

IvanKavaldzhiev:
**Description**:

Since we don't have speculative writes support during transaction simulations yet, we don't support precompile calls for eth_estimateGas. Thus, we need to throw a meaningful error, so that the json-rpc relay could redirect those calls to existing logic.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

- [ ] Documented (Code comments, README, etc.)
- [ ] Tested (unit, integration, etc.)

PR #6022, opened 2023-05-17
HIP-584: Adapt MirrorHTSPrecompiledContract to accept precompile calls

### LINKED ISSUE #6005: HIP-584: Adapt MirrorHTSPrecompiledContract to accept precompile calls
### Problem

We need to extend `MirrorHTSPrecompiledContract`, so that it accepts precompile calls from non-static frames. We should throw `UnsupportedOperationException` for all precompiles, that are not yet supported.

### Solution

Adapt `MirrorHTSPrecompiledContract` to accept precompile calls.

### Alternatives

_No response_

IvanKavaldzhiev:
**Description**:

This PR does the following:

- Allow `MirrorHTSPrecompiledContract` to accept precompile calls
- Throw `UnsupportedOperationException` when we try to access a precompile via non-static frame. This is a temporary measure, until we implement a given precompile and add a support for it. We will add precompile support one by one.
- Allows `eth_call` to invoke a dynamic function (non-view) which makes a read only precompile call. These calls would be redirected to the `ViewExecutor` and `RedirectViewExecutor` components and behave the same way like calling these precompiles via a view function. Currently this behaviour was blocked by a previous [PR](https://github.com/hashgraph/hedera-mirror-node/pull/5941). Note: if `eth_call` invokes a dynamic function which makes a call to a modification precompile - `UnsupportedOperationException` would be thrown

Most of the logic is copied from services but obsolete logic was removed to keep the class simpler.

**Related issue(s)**: 

Fixes #6005 

**Notes for reviewer**:

**Checklist**


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contracts/precompile/MirrorHTSPrecompiledContract.java
```diff
@@ -0,0 +1,329 @@
...
+            case AbiConstants.ABI_ID_DELETE_TOKEN -> throw new UnsupportedOperationException(UNSUPPORTED_ERROR_MESSAGE);
+            case AbiConstants.ABI_ID_UPDATE_TOKEN_INFO,
+                    AbiConstants.ABI_ID_UPDATE_TOKEN_INFO_V2,
+                    AbiConstants.ABI_ID_UPDATE_TOKEN_INFO_V3 -> throw new UnsupportedOperationException(UNSUPPORTED_ERROR_MESSAGE);
+            case AbiConstants.ABI_ID_UPDATE_TOKEN_KEYS -> throw new UnsupportedOperationException(UNSUPPORTED_ERROR_MESSAGE);
+            case AbiConstants.ABI_ID_GET_TOKEN_KEY -> throw new UnsupportedOperationException(UNSUPPORTED_ERROR_MESSAGE);
+            case AbiConstants.ABI_ID_REDIRECT_FOR_TOKEN -> {
+                final var target = DescriptorUtils.getRedirectTarget(input);
+
+                final var nestedFunctionSelector = target.descriptor();
```
steven-sheehy, line 251: I hope we're not planning on implementing all of these inline in this giant switch. Instead, we should have some precompile interface that auto-register themselves via some factory.

```java
public interface Precompile {
    Foo invoke(byte[] input);
    PrecompileName getName();
}

public enum PrecompileName {
...
}

@Named
public class PrecompileFactory {

    private final Map<PrecompileName, Precompile> precompiles;

    public PrecompileFactory(Collection<Precompile> precompiles) {
        this.precompiles = precompiles.stream().collect(toMap(Precompile::getName, Function.identity());
    }

    public Precompile lookup(PrecompileName name) {
       return precompiles.getOrDefault(name, UNSUPPORTED);
    }
}

**Review by steven-sheehy — CHANGES_REQUESTED** [review]

File: hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/EntityDatabaseAccessor.java [hunk 2]
```diff
@@ -29,14 +29,14 @@
 
 @Named
 @RequiredArgsConstructor
-public class EntityDatabaseAccessor extends DatabaseAccessor<Address, Entity> {
+public class EntityDatabaseAccessor extends DatabaseAccessor<Object, Entity> {
```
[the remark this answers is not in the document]
IvanKavaldzhiev: We plan to use 4 type of models inside the state. Those are:

- Token
- Account
- UniqueToken
- TokenRelationship

For these models we plan to use the following keys in the same order:

- Address
- Address
- Address, Long (for the serial number)
- Address, Address

So, we have 3 different types of keys that we plan to use inside the state, but the state is designed around a single key type. In order to mitigate this issue, we decided to use Object ignoring the type safety. Another way would be to try to remove and decouple the whole state from the generic key type, but this would be more time-consuming.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contracts/precompile/AbiConstants.java
```diff
@@ -0,0 +1,213 @@
+package com.hedera.mirror.web3.evm.store.contracts.precompile;
+
+import org.apache.tuweni.bytes.Bytes;
+
+public class AbiConstants {
```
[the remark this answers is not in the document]
IvanKavaldzhiev: It contains constants, that are not present in the AbiConstants inside `hedera-evm`. There is another AbiConstants, which is in services repo (outside `hedera-evm`) with all of the rest constants. This class here, copies the values from it. I will move the fields in `com.hedera.services`.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contracts/precompile/MirrorHTSPrecompiledContract.java
steven-sheehy: Since this is new code, I don't think we should write it all the wrong way to then just refactor it right after. That's wasted effort that in particular causes a lot of headache in tests. 

The specifics aren't as important, my main point is that precompiles should not be implemented inline in this class and moved to some separate class that can be tested in isolation. Any precompiles that aren't supported don't need to be explicitly listed and can just be a default statement of the switch that throws not supported exception.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/exception/ResourceLimitException.java
```diff
@@ -0,0 +1,16 @@
+package com.hedera.mirror.web3.exception;
+
+import com.hedera.node.app.service.evm.exceptions.InvalidTransactionException;
+import com.hederahashgraph.api.proto.java.ResponseCodeEnum;
+
+import java.io.Serial;
+
+public class ResourceLimitException extends InvalidTransactionException {
```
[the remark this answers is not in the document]
IvanKavaldzhiev: Done

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
```diff
@@ -85,6 +85,10 @@ public class MirrorNodeEvmProperties implements EvmProperties {
     @DurationMin(seconds = 100)
     private Duration rateLimit = Duration.ofSeconds(100L);
 
+    @Getter
+    @NotNull
+    private long htsDefaultGasCost = 10000;
```
steven-sheehy, line 90: Need to add to `docs/configuration.md`. Please also sort field inline.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/precompile/PrecompileFactory.java
```diff
@@ -0,0 +1,41 @@
...
+ * limitations under the License.
+ */
+
+package com.hedera.mirror.web3.evm.store.contract.precompile;
+
+import com.hedera.services.store.contracts.precompile.Precompile;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+import javax.inject.Named;
```
steven-sheehy, line 23: Need to switch javax to jakarta everywhere after Spring Boot 3 PR.
```suggestion
import jakarta.inject.Named;
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/EntityDatabaseAccessor.java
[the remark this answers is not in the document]
steven-sheehy: I'd like @david-bakin-sl to weigh in here as he designed it. Aren't the database accessors almost entirely used internally within the stacked state classes? I don't see where in the diff that they would be used directly and would have difficulty with using different key types. The only place is where you let Spring collection them into a list but using a wildcard for the generic works there.

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/precompile/PrecompileFactory.java
IvanKavaldzhiev: @steven-sheehy Unfortunately, for the moment, the switch can't happen everywhere, since the `hedera-evm` library relies on the `javax.inject` package and on `mirror-node` side we initialize objects from the library using `javax.inject` arguments in their constructor, so if we switch mirror-node's usage to `jakarta` on these files, it's a breaking change.

I'm not sure what the consequences of this mix between javax.inject and jakarta might be.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/EntityDatabaseAccessor.java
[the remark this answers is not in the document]
IvanKavaldzhiev: The difficulty of using different key types comes with the initialization of `StackedStateFrames<K>` which is bound to a single K type. When we pass the list of DbAccessors, they should have the same key type, so that they follow this rule.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/precompile/PrecompileFactory.java
steven-sheehy: It's fine to have both javax and jakarta in web3 at the same time as long as Spring only loads beans via jakarta. You can keep hedera-evm with javax since we don't use component scanning to load any beans within evm library. It's okay to keep `EvmOperationConstructionUtil` returning `javax.inject.Provider`. But every Spring bean we create inside web3 module should be jakarta.

steven-sheehy: It's working fine now on main with mixing javax and jakarta. I'm only asking that within your diff to change any places from `javax` to `jakarta`.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
IvanKavaldzhiev: Done

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/precompile/MirrorHTSPrecompiledContract.java
```diff
@@ -36,27 +83,167 @@ public Pair<Long, Bytes> computeCosted(
...
+        if (frame.isStatic()) {
+            if (!isTokenProxyRedirect(input) && !isViewFunction(input)) {
+                frame.setRevertReason(STATIC_CALL_REVERT_REASON);
+                return Pair.of(defaultGas(), null);
+            }
+
+            return super.computeCosted(input, frame, viewGasCalculator, tokenAccessor);
+        }
+        final var result = computePrecompile(input, frame);
+        return Pair.of(gasRequirement, result.getOutput());
```
david-bakin-sl, line 98: Things would be clearer IMO if `gasRequirement` wasn't a field but was just passed around.  For example `computePrecompile` could just return this `Pair`.  For example, right now `prepareComputation` sets the field to `0`.  Why?  `computePrecompile` sets it to `defaultGas` before doing real work.  But it might fail earlier (L121) so it must be `0` to be valid globally.  But if L121 just returned `Pair.of(0L, ...)` that would be fine.  Etc.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/precompile/MirrorHTSPrecompiledContract.java
```diff
@@ -36,27 +83,167 @@ public Pair<Long, Bytes> computeCosted(
...
+    }
+
+    @NonNull
+    @Override
+    public PrecompileContractResult computePrecompile(final Bytes input, @NonNull final MessageFrame frame) {
+        // Temporary workaround allowing eth_call to execute precompile methods in a dynamic context (non pure/view).
```
david-bakin-sl, line 104: IMO this would be a valid place for "TODO" regardless of Sonar.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/precompile/MirrorHTSPrecompiledContract.java
```diff
@@ -36,27 +83,167 @@ public Pair<Long, Bytes> computeCosted(
...
+        // This is done by calling ViewExecutor/RedirectViewExecutor logic instead of Precompile classes.
+        // After the Precompile classes are implemented, this workaround won't be needed.
+        if (isViewFunction(input)) {
+            return handleReadsFromDynamicContext(input, frame);
+        }
+
+        if (unqualifiedDelegateDetected(frame)) {
+            frame.setExceptionalHaltReason(Optional.of(ExceptionalHaltReason.PRECOMPILE_ERROR));
```
david-bakin-sl, line 112: Besu's `ExceptionalHaltReason` is extensible.  The services repo does it with `HederaExceptionalHaltReason`.  And they really should have added a new exceptional halt reason for this "unqualified delegate detected" error over there.  (I'll file an [issue](https://github.com/hashgraph/hedera-services/issues/6757).)

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/precompile/MirrorHTSPrecompiledContract.java
```diff
@@ -36,27 +83,167 @@ public Pair<Long, Bytes> computeCosted(
...
+            return INVALID_DELEGATE;
+        }
+        prepareFields(frame);
+        try {
+            prepareComputation(input, updater::unaliased);
+        } catch (InvalidTransactionException e) {
+            final var haltReason = HederaExceptionalHaltReason.ERROR_DECODING_PRECOMPILE_INPUT;
```
david-bakin-sl, line 119: Yep, and right here we see a proper use of `HederaExceptionalHaltReason`.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/precompile/MirrorHTSPrecompiledContract.java
```diff
@@ -36,27 +83,167 @@ public Pair<Long, Bytes> computeCosted(
...
+            return PrecompileContractResult.halt(null, Optional.of(haltReason));
         }
 
-        return super.computeCosted(input, frame, viewGasCalculator, tokenAccessor);
+        gasRequirement = defaultGas();
+        if (this.transactionBody == null) {
+            final var haltReason = Optional.of(ERROR_DECODING_PRECOMPILE_INPUT);
+            frame.setExceptionalHaltReason(haltReason);
+            return PrecompileContractResult.halt(null, haltReason);
+        }
```
david-bakin-sl, line 129: Can't this be merged with L117 above?  Maybe if `prepareComputation` or even `decodeInput` just threw a `InvalidTransactionException` (or something similar) in the case of `transactionBody == null`

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/precompile/MirrorHTSPrecompiledContract.java
```diff
@@ -36,27 +83,167 @@ public Pair<Long, Bytes> computeCosted(
...
+
+        final var now = frame.getBlockValues().getTimestamp();
+        gasRequirement = precompile.getGasRequirement(now);
```
david-bakin-sl, line 132: Why is `gasRequirement` set at L124?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/precompile/MirrorHTSPrecompiledContract.java
```diff
@@ -36,27 +83,167 @@ public Pair<Long, Bytes> computeCosted(
...
+        final Bytes result = computeInternal(frame);
+
+        return result == null
+                ? PrecompiledContract.PrecompileContractResult.halt(null, Optional.of(ExceptionalHaltReason.NONE))
```
david-bakin-sl, line 136: Here, what does it mean (to the user) that the result will have `MessageFrame.State.EXCEPTIONAL_HALT` with reason `ExceptionalHaltReason.NONE`?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/precompile/MirrorHTSPrecompiledContract.java
```diff
@@ -36,27 +83,167 @@ public Pair<Long, Bytes> computeCosted(
...
+        final var stack = frame.getMessageFrameStack();
+        final var frames = stack.iterator();
+        frames.next();
+        if (!frames.hasNext()) {
+            // Impossible to get here w/o a catastrophic EVM bug
+            log.error("Possibly CATASTROPHIC failure - delegatecall frame had no parent");
+            return false;
+        }
+        // If the token redirect contract was called via delegate, then it's a delegate
+        return isDelegateCall(frames.next());
```
david-bakin-sl, line 156: This no longer matches `unqualifiedDelegateDetected` in the services repo - [see here](https://github.com/hashgraph/hedera-services/blob/dad10e18e8bb0305ac99a9d9862407da505fa43c/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/store/contracts/precompile/HTSPrecompiledContract.java#LL261C27-L261C27) - is this intentional, and if so, why don't we care about the `isToken` qualification at least?  And maybe the `permittedDelegateCallers` condition, do we have to replicate that for the mirror node?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/precompile/MirrorHTSPrecompiledContract.java
```diff
@@ -36,27 +83,167 @@ public Pair<Long, Bytes> computeCosted(
...
+        Bytes result;
+        try {
+            validateTrue(frame.getRemainingGas() >= gasRequirement, INSUFFICIENT_GAS);
+
+            precompile.handleSentHbars(frame);
+            precompile.run(frame);
+
+            result = precompile.getSuccessResultFor();
+
+            stackedStateFrames.top().commit();
```
david-bakin-sl, line 169: woohoo!  the `commit` I've been waiting to see!

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/precompile/MirrorHTSPrecompiledContract.java
```diff
@@ -36,27 +83,167 @@ public Pair<Long, Bytes> computeCosted(
...
+            throw e;
+        } catch (final InvalidTransactionException e) {
+            final var status = e.getResponseCode();
+            result = precompile.getFailureResultFor(status);
+            if (e.isReverting()) {
+                frame.setState(MessageFrame.State.REVERT);
+                frame.setRevertReason(e.getRevertReason());
             }
+        } catch (final Exception e) {
+            log.warn("Internal precompile failure", e);
```
david-bakin-sl, line 182: Do the state things need to be set in this case, as at L178-179 above?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/precompile/MirrorHTSPrecompiledContract.java
```diff
@@ -36,27 +83,167 @@ public Pair<Long, Bytes> computeCosted(
...
+        try {
+            this.transactionBody = this.precompile.body(input, aliasResolver);
+        } catch (final Exception e) {
+            transactionBody = null;
+        }
+    }
+
+    void prepareFields(final MessageFrame frame) {
+        this.updater = (HederaEvmStackedWorldStateUpdater) frame.getWorldUpdater();
+        stackedStateFrames.push();
```
david-bakin-sl, line 217: And here's the `push`, woohoo again! But I don't see any `pop` around - there's gotta be one somewhere ....!!??

(No `resetToBase` either but maybe that only comes with `eth_estimateGas`)

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/precompile/MirrorHTSPrecompiledContract.java
```diff
@@ -36,27 +83,167 @@ public Pair<Long, Bytes> computeCosted(
...
+        if (isTokenProxyRedirect(input)) {
+            final var executor =
+                    infrastructureFactory.newRedirectExecutor(input, frame, viewGasCalculator, tokenAccessor);
+            resultFromExecutor = executor.computeCosted();
+        } else if (isViewFunction(input)) {
+            final var executor = infrastructureFactory.newViewExecutor(input, frame, viewGasCalculator, tokenAccessor);
+            resultFromExecutor = executor.computeCosted();
         }
+        return resultFromExecutor == null
+                ? PrecompiledContract.PrecompileContractResult.halt(null, Optional.of(ExceptionalHaltReason.NONE))
```
david-bakin-sl, line 236: (same question as above w.r.t. no result case ... probably same answer ...)

david-bakin-sl (changes requested):
Many of these comments are style comments - don't impact acceptance.  But there are some questions that need addressing - thus "request changes".

1. No stacked state `pop` call found.
2. The type `extends` in `AbstractEvmStackedLedgerUpdater`
3. Question about `unqualifiedDelegateDetected` matching the services repo
4. Catching `Exception` in several places where it appears overly broad.
5. No non-mock implementations of `Precompile`, is that ok at this point?

I'd really like to see the documentation updated as I suggest below, though if not part of this PR I would expect it to come quickly.

Also I haven't looked at the tests at all yet.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/precompile/MirrorHTSPrecompiledContract.java
IvanKavaldzhiev: Actually, the main logic regarding push/pop operations with the stack should happen in the Updater components, since their tied to the specific EVM frames. We have a separate [PR](https://github.com/hashgraph/hedera-mirror-node/pull/6066/files) for this logic.

I even think the `push()` in `MirrorHTSPrecompiledContract` is obsolete, since it will cause adding an extra `RWCachingStateFrame` that is not needed and would break the commit logic. The `push()` should happen in the `Updaters` and they should prepare the StackedStateFrames in advance, before entering the precompile execution. We should accept that the `StackedUpdater` and the `StackedStateFrames` inside, that are received in the `MirrorHTSPrecompiledContract` are ready to be used and only commit() should happen. When the `StackedUpdater` commits afterwards, it would pop the stacked frame that was used for the given precompile execution. I assume this is the correct logic 🤔

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/precompile/MirrorHTSPrecompiledContract.java
IvanKavaldzhiev: The `isToken` qualification seems missed, indeed. Good catch!

I was wondering `permittedDelegateCallers` and I'm not sure mirror-node should keep the same white list of callers, but instead permit more loosely the execution of delegate precompiles. In the worst case, a caller which is not whitelisted would receive an estimate gas value, but afterwards the consensus node transaction would fail, since they would be blocked to execute a delegate call. But this might happen in a variety of other cases like intermediate change of state between the two calls (estimateGas to Archive Node and then EthereumTransaction to Consensus Node).

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/precompile/MirrorHTSPrecompiledContract.java
IvanKavaldzhiev: I don't think so. Here we finish the execution of the precompile with a failing result, which is different from reverting the frame. We would keep this failure result in the EVM and use it for consequent nested calls.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmTxProcessorFacadeImpl.java
```diff
@@ -83,15 +95,17 @@ public HederaEvmTransactionProcessingResult execute(
             final long value,
             final Bytes callData,
             final boolean isStatic) {
+        final var stackedStateFrames = new StackedStateFrames<>(databaseAccessors);
```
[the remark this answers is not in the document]
IvanKavaldzhiev: I aligned the signatures to be the same.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmTxProcessorFacadeImpl.java
```diff
@@ -44,35 +47,44 @@ public class MirrorEvmTxProcessorFacadeImpl implements MirrorEvmTxProcessorFacad
 
     private final MirrorNodeEvmProperties evmProperties;
```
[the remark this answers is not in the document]
IvanKavaldzhiev: Done

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/precompile/MirrorHTSPrecompiledContract.java
IvanKavaldzhiev: I removed this push()

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/precompile/MirrorHTSPrecompiledContract.java
IvanKavaldzhiev: Added missing logic for isToken

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/exception/ResourceLimitException.java
```diff
@@ -0,0 +1,34 @@
...
+package com.hedera.mirror.web3.exception;
+
+import com.hedera.node.app.service.evm.exceptions.InvalidTransactionException;
+import com.hederahashgraph.api.proto.java.ResponseCodeEnum;
+import java.io.Serial;
+
+/**
+ * Exception thrown when a resource limit is exceeded (e.g. max nfts mint exceeded).
+ * **/
+public class ResourceLimitException extends InvalidTransactionException {
```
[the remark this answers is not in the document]
IvanKavaldzhiev: It would be thrown in the future when we implement the Logic types. It's not needed currently. I decided to add it now, so that we don't forget it in the future.

hedera-mirror-web3/src/main/java/com/hedera/services/store/contracts/precompile/Precompile.java
```diff
@@ -47,6 +49,8 @@ public interface Precompile {
 
     long getGasRequirement(long blockTimestamp);
 
+    Set<Integer> getFunctionSelectors();
```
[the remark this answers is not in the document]
IvanKavaldzhiev: Next sprint.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
```diff
@@ -67,6 +67,10 @@ public class MirrorNodeEvmProperties implements EvmProperties {
     @NotBlank
     private String fundingAccount = "0x0000000000000000000000000000000000000062";
 
+    @Getter
+    @NotNull
```
steven-sheehy, line 71: nit: Not needed since it's a primitive.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
```diff
@@ -110,6 +111,7 @@ private HederaEvmTransactionProcessingResult doProcessCall(
                     params.isEstimate() ? estimatedGas : params.getGas(),
                     params.getValue(),
                     params.getCallData(),
+                    Instant.now(),
```
steven-sheehy, line 114: Should adjust the field to use `MirrorEvmTxProcessor mirrorEvmTxProcessor` now that the interfaces are aligned.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/precompile/MirrorHTSPrecompiledContract.java
```diff
@@ -36,27 +98,168 @@ public Pair<Long, Bytes> computeCosted(
...
-        if (frameContainsNonStaticFrameInStack(frame)) {
-            throw new UnsupportedOperationException("Precompile not supported for non-static frames");
+        this.viewGasCalculator = viewGasCalculator;
+        this.tokenAccessor = tokenAccessor;
+
+        if (frame.isStatic()) {
+            if (!isTokenProxyRedirect(input) && !isViewFunction(input)) {
+                frame.setRevertReason(STATIC_CALL_REVERT_REASON);
+                return Pair.of(defaultGas(), null);
+            }
```
steven-sheehy, line 108: There's a lot of complicated logic in this class copied from services that is sure to be hard to keep in sync. Is the plan in the future after modularization to push this logic back into the hedera-evm library? If so, let's add a class comment noting the source of this logic and future direction.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/precompile/MirrorHTSPrecompiledContract.java
```diff
@@ -16,18 +16,80 @@
...
+    private static final Bytes STATIC_CALL_REVERT_REASON = Bytes.of("HTS precompiles are not static".getBytes());
+    private static final String UNSUPPORTED_ERROR = "Precompile not supported for non-static frames";
+    private final MirrorNodeEvmProperties evmProperties;
+    private final EvmInfrastructureFactory infrastructureFactory;
+    private final StackedStateFrames<Object> stackedStateFrames;
+    private final PrecompileMapper precompileMapper;
+    private Optional<Precompile> precompile;
+    private TransactionBody.Builder transactionBody;
+    private long gasRequirement = 0;
+    private Address senderAddress;
```
steven-sheehy, line 64: Please check the code smells

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
IvanKavaldzhiev: We can't directly call `MirrorEvmTxProcessor`, since it's constructed in the Facade which hides this logic.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/precompile/MirrorHTSPrecompiledContract.java
IvanKavaldzhiev: Done

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/precompile/MirrorHTSPrecompiledContract.java
IvanKavaldzhiev: I believe this logic might be pushed back in `smart-contract-service`, but there may still be the need of this class, serving as an adapter. I'm not sure how the design would evolve.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/precompile/PrecompileFactory.java
```diff
@@ -0,0 +1,41 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+@Named
+public class PrecompileFactory {
+
+    private final Map<Integer, Precompile> abiConstantToPrecompile = new HashMap<>();
+
+    public PrecompileFactory(final Set<Precompile> precompiles) {
```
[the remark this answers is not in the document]
IvanKavaldzhiev: Renamed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/precompile/PrecompileFactory.java
```diff
@@ -0,0 +1,41 @@
...
+        for (Precompile precompile : precompiles) {
+            for (Integer selector : precompile.getFunctionSelectors()) {
+                abiConstantToPrecompile.put(selector, precompile);
+            }
+        }
+    }
+
+    public Precompile lookup(int functionSelector) {
```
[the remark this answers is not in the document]
IvanKavaldzhiev: Refactored

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/EvmOperationConstructionUtil.java
```diff
@@ -69,20 +70,108 @@ public static Map<String, Provider<ContractCreationProcessor>> ccps(GasCalculato
...
+                AbiConstants.ABI_ID_UPDATE_TOKEN_EXPIRY_INFO,
+                AbiConstants.ABI_ID_UPDATE_TOKEN_EXPIRY_INFO_V2);
+    }
+
+    private static Map<String, PrecompiledContract> precompiles(final StackedStateFrames<Object> stackedStateFrames) {
         final Map<String, PrecompiledContract> hederaPrecompiles = new HashMap<>();
         final var evmFactory = new EvmInfrastructureFactory(new EvmEncodingFacade());
-        hederaPrecompiles.put(EVM_HTS_PRECOMPILED_CONTRACT_ADDRESS, new MirrorHTSPrecompiledContract(evmFactory));
+        final var mirrorNodeEvmProperties = new MirrorNodeEvmProperties();
+        final var precompileFactory = new PrecompileMapper(new HashSet<>(), getPrecompileFunctionSelectors());
```
steven-sheehy, line 170: PrecompileMapper is a singleton Spring bean yet here you're instantiating it. We should keep as a bean.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/EvmOperationConstructionUtil.java
```diff
@@ -69,20 +70,108 @@ public static Map<String, Provider<ContractCreationProcessor>> ccps(GasCalculato
...
                 EVM_VERSION_0_30,
                 () -> new MessageCallProcessor(evm, new PrecompileContractRegistry()),
                 EVM_VERSION_0_34,
-                () -> new HederaEvmMessageCallProcessor(evm, new PrecompileContractRegistry(), precompiles()));
+                () -> new HederaEvmMessageCallProcessor(
+                        evm, new PrecompileContractRegistry(), precompiles(stackedStateFrames)));
     }
 
-    private static Map<String, PrecompiledContract> precompiles() {
+    public static Set<Integer> getPrecompileFunctionSelectors() {
```
steven-sheehy, line 85: Hopefully this set is just temporary until there's implementations for each of these precompiles? Either way, I think this mapping belongs inside `PrecompileMapper` and that will help with its instantiation by Spring.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/store/contract/precompile/PrecompileMapperTest.java
```diff
@@ -0,0 +1,41 @@
...
+
+    private PrecompileMapper subject;
+
+    @BeforeEach
+    void setUp() {
+        subject = new PrecompileMapper(Set.of(new MockPrecompile()), new HashSet<>());
+    }
+
+    @Test
+    void nonExistingAbiReturnsEmpty() {
```
steven-sheehy, line 36: Needs a test for found and for unsupported

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/store/contract/precompile/MirrorHTSPrecompiledContractTest.java
```diff
@@ -49,55 +84,301 @@ class MirrorHTSPrecompiledContractTest {
...
+    private HederaEvmStackedWorldStateUpdater worldUpdater;
+
+    @Mock
+    private HederaEvmWorldStateTokenAccount account;
+
+    @Mock
+    private Deque<MessageFrame> messageFrameStack;
+
+    @Mock
+    private Iterator<MessageFrame> messageFrameIterator;
```
steven-sheehy, line 100: I don't think we should mock simple objects like collections, iterators and MirrorNodeEvmProperties

steven-sheehy (changes requested):

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/store/contract/precompile/MirrorHTSPrecompiledContractTest.java
IvanKavaldzhiev: Removed unnecessary mocks

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/store/contract/precompile/PrecompileMapperTest.java
IvanKavaldzhiev: Done

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/EvmOperationConstructionUtil.java
IvanKavaldzhiev: Yes, this is a temporary solution. I moved the Set to PrecompileMapper.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/EvmOperationConstructionUtil.java
IvanKavaldzhiev: Done

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
```diff
@@ -67,6 +67,9 @@ public class MirrorNodeEvmProperties implements EvmProperties {
     @NotBlank
     private String fundingAccount = "0x0000000000000000000000000000000000000062";
 
+    @Getter
+    private long htsDefaultGasCost = 10000;
```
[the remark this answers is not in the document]
IvanKavaldzhiev: Done, vim helped to prevent showing extra whitespaces as changes.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/EvmOperationConstructionUtil.java
```diff
@@ -69,20 +70,108 @@ public static Map<String, Provider<ContractCreationProcessor>> ccps(GasCalculato
...
+                AbiConstants.ABI_ID_GET_TOKEN_EXPIRY_INFO,
+                AbiConstants.ABI_ID_UPDATE_TOKEN_EXPIRY_INFO,
+                AbiConstants.ABI_ID_UPDATE_TOKEN_EXPIRY_INFO_V2);
+    }
+
+    private static Map<String, PrecompiledContract> precompiles(final StackedStateFrames<Object> stackedStateFrames) {
         final Map<String, PrecompiledContract> hederaPrecompiles = new HashMap<>();
         final var evmFactory = new EvmInfrastructureFactory(new EvmEncodingFacade());
-        hederaPrecompiles.put(EVM_HTS_PRECOMPILED_CONTRACT_ADDRESS, new MirrorHTSPrecompiledContract(evmFactory));
+        final var mirrorNodeEvmProperties = new MirrorNodeEvmProperties();
```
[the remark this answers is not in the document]
IvanKavaldzhiev: Done

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/EvmOperationConstructionUtil.java
```diff
@@ -69,20 +69,27 @@ public static Map<String, Provider<ContractCreationProcessor>> ccps(GasCalculato
                 () -> new ContractCreationProcessor(gasCalculator, evm, true, List.of(), 1));
     }
 
-    public static Map<String, Provider<MessageCallProcessor>> mcps(GasCalculator gasCalculator) {
+    public static Map<String, Provider<MessageCallProcessor>> mcps(
```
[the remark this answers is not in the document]
IvanKavaldzhiev: We could refactor them in a future PR

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/EntityDatabaseAccessor.java
```diff
@@ -29,14 +29,14 @@
...
     private final EntityRepository entityRepository;
 
     @Override
-    public @NotNull Optional<Entity> get(@NotNull Address address) {
-        var addressBytes = address.toArrayUnsafe();
+    public @NotNull Optional<Entity> get(@NotNull Object address) {
+        var addressBytes = ((Address) address).toArrayUnsafe();
```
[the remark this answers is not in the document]
IvanKavaldzhiev: Currently the type is reverted back to Address

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/precompile/MirrorHTSPrecompiledContract.java
```diff
@@ -36,27 +83,167 @@ public Pair<Long, Bytes> computeCosted(
...
+        try {
+            validateTrue(frame.getRemainingGas() >= gasRequirement, INSUFFICIENT_GAS);
+
+            precompile.handleSentHbars(frame);
+            precompile.run(frame);
+
+            result = precompile.getSuccessResultFor();
+
+            stackedStateFrames.top().commit();
+        } catch (final ResourceLimitException e) {
```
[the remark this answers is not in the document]
IvanKavaldzhiev: I removed ResourceLimitException

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/precompile/MirrorHTSPrecompiledContract.java
IvanKavaldzhiev: Added TODO

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/precompile/MirrorHTSPrecompiledContract.java
IvanKavaldzhiev: I merged the statements

steven-sheehy (approved):
LGTM

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/TokenAllowanceDatabaseAccessor.java
```diff
@@ -26,12 +26,12 @@
...
 @RequiredArgsConstructor
-public class TokenAllowanceDatabaseAccessor extends DatabaseAccessor<Id, TokenAllowance> {
+public class TokenAllowanceDatabaseAccessor extends DatabaseAccessor<Object, TokenAllowance> {
 
     private final TokenAllowanceRepository tokenAllowanceRepository;
 
     @Override
-    public @NonNull Optional<TokenAllowance> get(@NonNull Id key) {
-        return tokenAllowanceRepository.findById(key);
+    public @NonNull Optional<TokenAllowance> get(@NonNull Object key) {
```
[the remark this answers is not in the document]
IvanKavaldzhiev: Changed to `AbstractTokenAllowance.Id`.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/TokenAccountDatabaseAccessor.java
```diff
@@ -26,12 +26,12 @@
...
 @RequiredArgsConstructor
-public class TokenAccountDatabaseAccessor extends DatabaseAccessor<Id, TokenAccount> {
+public class TokenAccountDatabaseAccessor extends DatabaseAccessor<Object, TokenAccount> {
 
     private final TokenAccountRepository tokenAccountRepository;
 
     @Override
-    public @NonNull Optional<TokenAccount> get(@NonNull Id key) {
-        return tokenAccountRepository.findById(key);
+    public @NonNull Optional<TokenAccount> get(@NonNull Object key) {
```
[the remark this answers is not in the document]
IvanKavaldzhiev: Changed.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/HederaEvmStackedWorldStateUpdater.java
```diff
@@ -75,6 +82,32 @@ public EvmAccount getAccount(final Address address) {
...
+     * Returns the mirror form of the given EVM address if it exists; or 20 bytes of binary zeros if
+     * the given address is the mirror address of an account with an EIP-1014 address.
+     *
+     * @param evmAddress an EVM address
+     * @return its mirror form, or binary zeros if an EIP-1014 address should have been used for
+     *     this account
+     */
+    public byte[] unaliased(final byte[] evmAddress) {
+        final var addressOrAlias = Address.wrap(Bytes.wrap(evmAddress));
+        if (!addressOrAlias.equals(tokenAccessor.canonicalAddress(addressOrAlias))) {
```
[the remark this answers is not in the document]
IvanKavaldzhiev: I updated the javadoc.

hedera-mirror-web3/src/main/java/com/hedera/services/store/contracts/precompile/AbiConstants.java
```diff
@@ -0,0 +1,229 @@
+/*
```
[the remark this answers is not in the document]
IvanKavaldzhiev: I added a comment, but we usually copy corresponding services resources in the same package names inside web3 under `com.hedera.services`, so that we know where to find their original sources.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/precompile/MirrorHTSPrecompiledContract.java
```diff
@@ -36,27 +83,167 @@ public Pair<Long, Bytes> computeCosted(
...
+        this.precompile = precompileFactory.lookup(functionId);
+
+        if (precompile != null) {
+            decodeInput(input, aliasResolver);
+        } else {
+            throw new UnsupportedOperationException(UNSUPPORTED_ERROR);
+        }
+    }
+
+    void decodeInput(final Bytes input, final UnaryOperator<byte[]> aliasResolver) {
```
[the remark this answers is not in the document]
IvanKavaldzhiev: I simplified the class logic and removed `decodeInput`

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/precompile/MirrorHTSPrecompiledContract.java
```diff
@@ -36,27 +83,167 @@ public Pair<Long, Bytes> computeCosted(
...
+        this.transactionBody = TransactionBody.newBuilder();
+        try {
+            this.transactionBody = this.precompile.body(input, aliasResolver);
+        } catch (final Exception e) {
+            transactionBody = null;
```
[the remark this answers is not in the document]
IvanKavaldzhiev: Removed the Exception handler since there is no use of it currently. We might add a suitable exception handler if we add specific logic.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/precompile/MirrorHTSPrecompiledContract.java
IvanKavaldzhiev: I'm not sure. Probably this is a default value. It very unlikely to get `null result` from a precompile execution. If we have an error, we would set the corresponding failure result. The only way, IMO, to get a `null result` is to have `null` returned from the `getSuccessResultFor` precompile method, which means something suspicious happened and we haven't thrown an error so we'd better halt the frame.

steven-sheehy:
Need to merge main and fix code smells.

IvanKavaldzhiev:
@steven-sheehy I merged the conflicts. The code smells are intentional since some temporary changes were marked with TODO, as suggested by David.

steven-sheehy:
DCO problem

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/precompile/MirrorHTSPrecompiledContract.java
```diff
@@ -16,18 +16,65 @@
...
 import org.hyperledger.besu.evm.frame.MessageFrame;
+import org.hyperledger.besu.evm.precompile.PrecompiledContract;
 
+@CustomLog
 public class MirrorHTSPrecompiledContract extends EvmHTSPrecompiledContract {
 
-    public MirrorHTSPrecompiledContract(EvmInfrastructureFactory infrastructureFactory) {
+    public static final PrecompileContractResult INVALID_DELEGATE = new PrecompileContractResult(
+            null, true, MessageFrame.State.COMPLETED_FAILED, Optional.of(ExceptionalHaltReason.PRECOMPILE_ERROR));
+    private static final Bytes STATIC_CALL_REVERT_REASON = Bytes.of("HTS precompiles are not static".getBytes());
```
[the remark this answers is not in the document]
david-bakin-sl: ok, though I thought we were moving to the new terms

david-bakin-sl (approved):
LMGTM (looks _mostly_ good to me) - for outstanding issues to be addressed (IMO) I'll list them separately and we'll work on them shortly

steven-sheehy (approved):
LGTM. Please address the comments in a follow up

PR #6032, opened 2023-05-19
HIP-584: Refactor accessors to use StackedStateFrames

### LINKED ISSUE #5959: HIP-584: Refactor `TokenAccessorImpl`, `AccountAccessorImpl` and `MirrorEntityAccess` to use `StackedStateFrames`
### Problem

We need to modify `TokenAccessorImpl`, `AccountAccessorImpl` and `MirrorEntityAccess` classes to use `StackedStateFrames` instead of repositories, so that we read most up-to-date information from the state.

### Solution

Replace repository usage in this classes with `StackedStateFrames`. The only exception is contractStateRepository in MirrorEntityAccess, since we currently don't store contract state changes inside the Stacked State. Those should be handled internally by the Besu EVM.

### Alternatives

_No response_

0xivanov:
**Description**:

We need to modify TokenAccessorImpl, AccountAccessorImpl and MirrorEntityAccess classes to use StackedStateFrames instead of repositories, so that we read most up-to-date information from the state.

**Related issue(s)**:

Fixes #5959

**Notes for reviewer**:

**Checklist**


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/account/AccountAccessorImpl.java
```diff
@@ -31,17 +30,21 @@
...
 public class AccountAccessorImpl implements AccountAccessor {
     public static final int EVM_ADDRESS_SIZE = 20;
+    private StackedStateFrames<Object> stackedStateFrames;
+
     private final MirrorEntityAccess mirrorEntityAccess;
-    private final EntityRepository entityRepository;
+
+    public void setStackedStateFrames(StackedStateFrames<Object> stackedStateFrames) {
+        this.stackedStateFrames = stackedStateFrames;
+    }
```
steven-sheehy, line 39: This class is a singleton bean shared by all requests. We cannot set stateful objects on it and need to keep it stateless.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
```diff
@@ -38,9 +38,13 @@
...
 public class MirrorEntityAccess implements HederaEvmEntityAccess {
-    private final EntityRepository entityRepository;
-    private final ContractRepository contractRepository;
     private final ContractStateRepository contractStateRepository;
+    private final ContractRepository contractRepository;
+    private StackedStateFrames<Object> stackedStateFrames;
+
+    public void setStackedStateFrames(StackedStateFrames<Object> stackedStateFrames) {
+        this.stackedStateFrames = stackedStateFrames;
+    }
```
steven-sheehy, line 47: Remove stateful setter on singleton bean.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/token/TokenAccessorImpl.java
```diff
@@ -36,117 +38,113 @@
...
-    private final TokenAllowanceRepository tokenAllowanceRepository;
-    private final NftAllowanceRepository nftAllowanceRepository;
-    private final CustomFeeDatabaseAccessor customFeeDatabaseAccessor;
-
+    private StackedStateFrames<Object> stackedStateFrames;
     private final MirrorNodeEvmProperties properties;
 
+    public void setStackedStateFrames(StackedStateFrames<Object> stackedStateFrames) {
+        this.stackedStateFrames = stackedStateFrames;
+    }
```
steven-sheehy, line 74: Remove

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/EntityDatabaseAccessor.java
```diff
@@ -33,11 +33,15 @@
...
 @RequiredArgsConstructor
-public class EntityDatabaseAccessor extends DatabaseAccessor<Address, Entity> {
+public class EntityDatabaseAccessor extends DatabaseAccessor<Object, Entity> {
     private final EntityRepository entityRepository;
 
     @Override
-    public @NotNull Optional<Entity> get(@NotNull Address address) {
+    public @NotNull Optional<Entity> get(@NotNull Object key) {
+        if (key instanceof EntityId) {
+            return entityRepository.findByIdAndDeletedIsFalse(((EntityId) key).getId());
```
steven-sheehy, line 42:
```suggestion
        if (key instanceof EntityId entityId) {
            return entityRepository.findByIdAndDeletedIsFalse(entityId.getId());
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/HederaEvmWorldState.java
```diff
@@ -118,18 +120,22 @@ public static class Updater extends AbstractLedgerEvmWorldUpdater<HederaEvmMutab
 
         protected Updater(
                 final HederaEvmWorldState world,
-                final AccountAccessor accountAccessor,
-                final HederaEvmEntityAccess hederaEvmEntityAccess,
-                final TokenAccessor tokenAccessor,
+                final AccountAccessorImpl accountAccessor,
+                final MirrorEntityAccess hederaEvmEntityAccess,
+                final TokenAccessorImpl tokenAccessor,
```
steven-sheehy, line 125: Shouldn't leak implementation class. Especially since we should not be setting state on stateless beans.

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/account/AccountAccessorImpl.java
0xivanov: I agree that we should not have a bean and stateful object inside. We need to create new `StackedStateFrames` for every transaction and assign it to the accessors and the updaters. One option is to convert the accessors to not be singleton beans and create them manually in every `execute` method. Another option is to make the `StackedStateFrames` bean and `invalidate` after every transaction, but this probably would break when executing multiple transaction at the same time.

IvanKavaldzhiev: I want to clarify the usage of these accessors. I know it's confusing but we have different types of accessors. We have several types related to the stacked state design which are:

- AccountDatabaseAccessor
- TokenDatabaseAccessor
- UniqueTokenDatabaseAccessor
- TokenRelationshipDatabaseAccessor

They are stateless and serve as a repository layer.

We have a different kind of accessors, that were introduces a while ago inside the `hedera-evm` design. Those are [AccountAccessor](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/store/contracts/AccountAccessorImpl.java) and [TokenAccessor](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/store/contracts/TokenAccessorImpl.java). Their intent was to serve as a facade for accessing in-memory states and hide the services ledger logic. As it can be seen in the codebase they wrap a `WorldLedgers` instance that hold all type of ledgers which are keeping state. On `mirror-node` side we have `StackedStateFrames` that substitute the Ledger design. That's why here we want to wrap a `StackedStateFrames` instance and perform read related operations. We have interfaces defined in `hedera-evm` and we should provide an implementation for them in order to integrate with the library.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
0xivanov: Fixed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/token/TokenAccessorImpl.java
0xivanov: Fixed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/HederaEvmWorldState.java
0xivanov: Fixed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
```diff
@@ -24,23 +24,21 @@
...
 import org.hyperledger.besu.datatypes.Address;
 
-@Named
 @RequiredArgsConstructor
 public class MirrorEntityAccess implements HederaEvmEntityAccess {
-    private final EntityRepository entityRepository;
-    private final ContractRepository contractRepository;
     private final ContractStateRepository contractStateRepository;
+    private final ContractRepository contractRepository;
+    private final StackedStateFrames<Object> stackedStateFrames;
```
steven-sheehy, line 41: This stateful stacked state class is infecting every other stateless class and making them in turn become stateful. This increases overall memory and causes concurrency issues (see incorrect changes to `MirrorEvmTxProcessorFacadeImpl` in this PR, for an example).

Ideally, every service class is stateless and state is stored in some object that is passed around. If that's not possible, but we can guarantee that any particular request executes on the same request for the entirety of its lifetime then we take another approach. Even though I'm not a fan of them usually, how about using `ThreadLocal` in `StackedStateFrames` and making it stateless?

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccessTest.java
```diff
@@ -53,8 +59,7 @@ class MirrorEntityAccessTest {
     private static final Long ENTITY_ID =
             EntityIdEndec.encode(ENTITY.getShardNum(), ENTITY.getRealmNum(), ENTITY.getEntityNum());
 
-    @Mock
-    private EntityRepository entityRepository;
+    private EntityRepository entityRepository = mock(EntityRepository.class);
```
steven-sheehy, line 62: Why?

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/store/contract/HederaEvmWorldStateTest.java
```diff
@@ -61,17 +53,18 @@ class HederaEvmWorldStateTest {
     final long balance = 1_234L;
 
     @Mock
-    AccountAccessor accountAccessor;
+    AccountAccessorImpl accountAccessor;
 
     @Mock
-    TokenAccessor tokenAccessor;
+    TokenAccessorImpl tokenAccessor;
```
steven-sheehy, line 59: Should use the interfaces so we can use proxies instead of cglib.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/account/AccountAccessorImplTest.java
```diff
@@ -46,12 +52,16 @@ class AccountAccessorImplTest {
     @Mock
     private MirrorEntityAccess mirrorEntityAccess;
 
-    @Mock
-    private EntityRepository entityRepository;
+    private EntityRepository entityRepository = mock(EntityRepository.class);
```
steven-sheehy, line 55: Revert

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/token/TokenAccessorImpl.java
```diff
@@ -390,9 +406,28 @@ private Long entityIdFromAccountAddress(final Address address) {
...
-        return entityRepository
-                .findByEvmAddressAndDeletedIsFalse(addressBytes)
-                .map(AbstractEntity::getId)
-                .orElse(0L);
+        final var topFrame = stackedStateFrames.top();
+        final var entityAccessor = topFrame.getAccessor(Entity.class);
+        return entityAccessor.get(address).map(AbstractEntity::getId).orElse(0L);
+    }
+
+    private Address evmAddressFromId(EntityId entityId) {
```
steven-sheehy, line 414: Missing some test coverage in this method

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/token/TokenAccessorImpl.java
0xivanov: Fixed

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/account/AccountAccessorImplTest.java
0xivanov: Fixed

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/store/contract/HederaEvmWorldStateTest.java
0xivanov: Fixed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmTxProcessorFacadeImpl.java
```diff
@@ -48,40 +50,52 @@
...
 
     private final MirrorNodeEvmProperties evmProperties;
-    private final MirrorOperationTracer mirrorOperationTracer;
+    private MirrorOperationTracer mirrorOperationTracer;
     private final StaticBlockMetaSource blockMetaSource;
-    private final MirrorEvmContractAliases mirrorEvmContractAliases;
-    private final PricesAndFeesImpl pricesAndFees;
-    private final AbstractCodeCache codeCache;
-    private final HederaEvmMutableWorldState worldState;
+    private MirrorEvmContractAliases aliasManager;
```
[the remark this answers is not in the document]
0xivanov: Fixed

IvanKavaldzhiev:
Consider removing `MirrrorEntityAccess`, `TokenAccessorImpl` and `AccountAccessorImpl` in favour of the `StateImpl` class introduced in this [PR](https://github.com/hashgraph/hedera-mirror-node/pull/6173) by making it implement all of the needed 3 interfaces.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
IvanKavaldzhiev: @steven-sheehy I think the two types of nodes differ in nature. The consensus node handles transactions one by one and it's not a problem to have stateful components there, while on the Archive Node we would handle multiple transactions at once.

What about creating new `TokenAccessor`, `AccountAccessor`, `MirrorEntityAccess` for each new transaction? It would increase the memory footprint but it's an easy way to avoid the concurrency issues.

IvanKavaldzhiev:
> Consider removing `MirrrorEntityAccess`, `TokenAccessorImpl` and `AccountAccessorImpl` in favour of the `StateImpl` class introduced in this [PR](https://github.com/hashgraph/hedera-mirror-node/pull/6173) by making it implement all of the needed 3 interfaces.

It seems it can't be achieved without changes in `hedera-evm`, since `AbstractLedgerEvmWorldUpdater` requires 3 concrete different implementations for the 3 interfaces. We can postpone this idea for now.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/StoreImpl.java
```diff
@@ -50,6 +59,37 @@ public Account getAccount(final Address address, final OnMissing throwIfMissing)
...
     }
 
+    @Override
+    public Optional<Entity> getEntity(Address address, OnMissing throwIfMissing) {
+        return stackedStateFrames.top().getAccessor(Entity.class).get(address);
+    }
+
+    @Override
+    public Optional<Entity> getEntity(EntityId entityId, OnMissing throwIfMissing) {
+        return stackedStateFrames.top().getAccessor(Entity.class).get(entityId);
```
IvanKavaldzhiev, line 69: `EntityDatabaseAccessor` works with Address as a key. You have to convert the `EntityId` to `Address`.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/StoreImpl.java
```diff
@@ -50,6 +59,37 @@ public Account getAccount(final Address address, final OnMissing throwIfMissing)
...
+
+    @Override
+    public Optional<TokenAccount> getTokenAccount(AbstractTokenAccount.Id id, OnMissing throwIfMissing) {
+        return stackedStateFrames.top().getAccessor(TokenAccount.class).get(id);
+    }
+
+    @Override
+    @SuppressWarnings("rawtypes")
+    public Optional<List> getCustomFee(Long entityIdNum, OnMissing throwIfMissing) {
+        return stackedStateFrames.top().getAccessor(List.class).get(entityIdNum);
```
IvanKavaldzhiev, line 80: We would most probably update given Token's list of CustomFees, not keeping standalone CustomFees in the state, so if we want to fetch updated CustomFees info, we should first fetch a given Token from the state and then return it's list of CustomFees. Use TokenDatabaseAccessor instead.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/StoreImpl.java
```diff
@@ -50,6 +59,37 @@ public Account getAccount(final Address address, final OnMissing throwIfMissing)
...
+    }
+
+    @Override
+    public Optional<TokenAllowance> getTokenAllowance(AbstractTokenAllowance.Id id, OnMissing throwIfMissing) {
+        return stackedStateFrames.top().getAccessor(TokenAllowance.class).get(id);
```
IvanKavaldzhiev, line 85: I think we should fetch an `Account` model and use its `SortedMap<FcTokenAllowanceId, Long>`

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/UniqueToken.java
```diff
@@ -33,6 +35,7 @@
  * Differences from the original:
  *  1. Added address field for convenience
  *  2. Added factory method that returns empty instance
+ *  3. Added equals() and hashCode()
```
IvanKavaldzhiev, line 38: Document added method `isEmptyUniqueToken`

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/StoreImpl.java
```diff
@@ -50,6 +59,37 @@ public Account getAccount(final Address address, final OnMissing throwIfMissing)
...
+    }
+
+    @Override
+    public Optional<NftAllowance> getNftAllowance(AbstractNftAllowance.Id id, OnMissing throwIfMissing) {
+        return stackedStateFrames.top().getAccessor(NftAllowance.class).get(id);
```
IvanKavaldzhiev, line 90: Fetch an `Account` model and read its set of `approveForAllNfts`

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/token/TokenAccessorImpl.java
```diff
@@ -35,119 +37,96 @@
...
                 .map(acc -> acc.getFreezeStatus().equals(FROZEN))
                 .orElse(false);
     }
 
     @Override
-    public boolean defaultFreezeStatus(final Address token) {
-        final var tokenId = new TokenId(entityIdFromEvmAddress(token));
-        return tokenRepository.findById(tokenId).map(Token::getFreezeDefault).orElse(false);
+    public boolean defaultFreezeStatus(final Address address) {
+        return store.getFungibleToken(address, OnMissing.DONT_THROW).hasFreezeKey();
```
IvanKavaldzhiev, line 116: Use `isFrozenByDefault`

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/token/TokenAccessorImpl.java
```diff
@@ -158,13 +137,8 @@ public Optional<List<CustomFee>> infoForTokenCustomFees(final Address token) {
...
     @Override
-    public TokenType typeOf(final Address token) {
-        final var tokenId = new TokenId(entityIdFromEvmAddress(token));
-
-        return tokenRepository
-                .findById(tokenId)
-                .map(t -> TokenType.valueOf(t.getType().name()))
-                .orElse(null);
+    public TokenType typeOf(final Address address) {
+        return store.getFungibleToken(address, OnMissing.DONT_THROW).getType();
```
IvanKavaldzhiev, line 141: This call would always return `FUNGIBLE` type. We should take into account both fungible and non-fungible cases. I suggest using both `getFungibleToken` and `getUniqueToken` and return type based on which of the two methods actually returns a result.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/token/TokenAccessorImpl.java
```diff
@@ -185,37 +159,31 @@ public EvmKey keyOf(final Address address, final TokenKeyType tokenKeyType) {
     }
 
     @Override
-    public String nameOf(final Address token) {
-        final var tokenId = new TokenId(entityIdFromEvmAddress(token));
-        return tokenRepository.findById(tokenId).map(Token::getName).orElse("");
+    public String nameOf(final Address address) {
+        return store.getFungibleToken(address, OnMissing.DONT_THROW).getName();
```
IvanKavaldzhiev, line 163: Again use both `getFungibleToken` and `getUniqueToken` methods

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/token/TokenAccessorImpl.java
```diff
@@ -185,37 +159,31 @@ public EvmKey keyOf(final Address address, final TokenKeyType tokenKeyType) {
...
     }
 
     @Override
-    public String symbolOf(final Address token) {
-        final var tokenId = new TokenId(entityIdFromEvmAddress(token));
-        return tokenRepository.findById(tokenId).map(Token::getSymbol).orElse("");
+    public String symbolOf(final Address address) {
+        return store.getFungibleToken(address, OnMissing.DONT_THROW).getSymbol();
```
IvanKavaldzhiev, line 168: Again use both `getFungibleToken` and `getUniqueToken` methods

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/token/TokenAccessorImpl.java
```diff
@@ -185,37 +159,31 @@ public EvmKey keyOf(final Address address, final TokenKeyType tokenKeyType) {
...
     }
 
     @Override
-    public long totalSupplyOf(final Address token) {
-        final var tokenId = new TokenId(entityIdFromEvmAddress(token));
-        return tokenRepository.findById(tokenId).map(Token::getTotalSupply).orElse(0L);
+    public long totalSupplyOf(final Address address) {
+        return store.getFungibleToken(address, OnMissing.DONT_THROW).getTotalSupply();
```
IvanKavaldzhiev, line 173: Again use both `getFungibleToken` and `getUniqueToken` methods

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/token/TokenAccessorImpl.java
```diff
@@ -227,23 +195,25 @@ public long staticAllowanceOf(final Address owner, final Address spender, final
         tokenAllowanceId.setSpender(entityIdFromAccountAddress(spender));
         tokenAllowanceId.setTokenId(entityIdNumFromEvmAddress(token));
 
-        return tokenAllowanceRepository
-                .findById(tokenAllowanceId)
+        return store.getTokenAllowance(tokenAllowanceId, OnMissing.DONT_THROW)
```
IvanKavaldzhiev, line 198: Make sure `getTokenAllowance` internally calls `getAccount` and uses its allowance map.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/token/TokenAccessorImpl.java
```diff
@@ -278,56 +251,54 @@ public Address canonicalAddress(final Address addressOrAlias) {
...
-                new AbstractNft.Id(serialNo, entityIdFromEvmAddress(nft).getId());
-
-        return nftRepository
-                .findById(nftId)
-                .map(n -> new String(n.getMetadata(), StandardCharsets.UTF_8))
-                .orElse("");
+    public String metadataOf(final Address address, long serialNo) {
+        final var entityId = entityIdFromEvmAddress(address);
+        final var nft = store.getUniqueToken(
+                new NftId(entityId.getShardNum(), entityId.getRealmNum(), entityId.getEntityNum(), serialNo),
```
IvanKavaldzhiev, line 257: Consider extracting a utility method that constructs `NftId` based on `EntityId` and `serial number`.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/token/TokenAccessorImpl.java
IvanKavaldzhiev: Update: Based on the changes from this [PR](https://github.com/hashgraph/hedera-mirror-node/pull/6239/files#diff-0b44d84d46804689d917a7efa810b48531856c99e62e4840c1a98ff4ca319c7b) you should just invoke `getToken()` method.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/token/TokenAccessorImpl.java
IvanKavaldzhiev: Update: Based on the changes from this [PR](https://github.com/hashgraph/hedera-mirror-node/pull/6239/files#diff-0b44d84d46804689d917a7efa810b48531856c99e62e4840c1a98ff4ca319c7b) you should just invoke getToken() method and actually keep the rest of your logic.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/token/TokenAccessorImpl.java
IvanKavaldzhiev: Update: Based on the changes from this [PR](https://github.com/hashgraph/hedera-mirror-node/pull/6239/files#diff-0b44d84d46804689d917a7efa810b48531856c99e62e4840c1a98ff4ca319c7b) you should just invoke getToken() method and actually keep the rest of your logic.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/token/TokenAccessorImpl.java
IvanKavaldzhiev: Update: Based on the changes from this [PR](https://github.com/hashgraph/hedera-mirror-node/pull/6239/files#diff-0b44d84d46804689d917a7efa810b48531856c99e62e4840c1a98ff4ca319c7b) you should just invoke getToken() method and actually keep the rest of your logic.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/token/TokenAccessorImpl.java
```diff
@@ -278,56 +251,54 @@ public Address canonicalAddress(final Address addressOrAlias) {
...
-                entityDatabaseAccessor.evmAddressFromId(tokenEntity.getTreasuryAccountId()),
+                tokenEntity.getTreasury().getAccountAddress(),
                 tokenEntity.getTotalSupply(),
                 tokenEntity.getMaxSupply(),
                 tokenEntity.getDecimals(),
                 expirationTimeInSec);
         evmTokenInfo.setAutoRenewPeriod(entity.getAutoRenewPeriod() != null ? entity.getAutoRenewPeriod() : 0);
-        evmTokenInfo.setDefaultFreezeStatus(tokenEntity.getFreezeDefault());
-        evmTokenInfo.setCustomFees(getCustomFees(token));
+        evmTokenInfo.setDefaultFreezeStatus(tokenEntity.hasFreezeKey());
```
IvanKavaldzhiev, line 294: Use getFreezeDefault

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/token/TokenAccessorImpl.java
```diff
@@ -348,19 +319,26 @@ private Long expirationTimeSeconds(Entity entity) {
         }
     }
 
-    private List<CustomFee> getCustomFees(final Address token) {
-        return customFeeDatabaseAccessor.get(entityIdNumFromEvmAddress(token)).orElse(Collections.emptyList());
+    @SuppressWarnings("unchecked")
+    private List<CustomFee> getCustomFees(final Address address) {
+        return store.getCustomFee(entityIdNumFromEvmAddress(address), OnMissing.DONT_THROW)
```
IvanKavaldzhiev, line 324: Rework `getCustomFees` to internally use `getToken`

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/UniqueToken.java
```diff
@@ -103,4 +110,14 @@ public String toString() {
...
     }
+
+    @Override
+    public boolean equals(final Object obj) {
+        return EqualsBuilder.reflectionEquals(this, obj);
+    }
+
+    @Override
+    public int hashCode() {
+        return HashCodeBuilder.reflectionHashCode(this);
```
IvanKavaldzhiev, line 121: Why we need `equals` and `hashCode`? Is `UniqueToken` part of a collection?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/StoreImpl.java
0xivanov: Fixed

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/UniqueToken.java
0xivanov: Fixed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/StoreImpl.java
0xivanov: Fixed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/StoreImpl.java
0xivanov: Fixed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/StoreImpl.java
0xivanov: Fixed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/token/TokenAccessorImpl.java
0xivanov: Fixed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/token/TokenAccessorImpl.java
0xivanov: Fixed

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/UniqueToken.java
0xivanov: I use them in `isEmptyUniqueToken` because some of the `Store` methods do not support `Optional` and there is a need for check if store returned default value or not.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/token/TokenAccessorImpl.java
0xivanov: Fixed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/token/TokenAccessorImpl.java
0xivanov: Fixed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/token/TokenAccessorImpl.java
0xivanov: Fixed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/Store.java
```diff
@@ -34,7 +41,17 @@ public interface Store {
...
+
+    Optional<TokenAccount> getTokenAccount(AbstractTokenAccount.Id id, OnMissing throwIfMissing);
+
+    Optional<List<CustomFee>> getCustomFee(Address entityIdNum, OnMissing throwIfMissing);
+
+    Long getTokenAllowance(Address address, FcTokenAllowanceId id, OnMissing throwIfMissing);
+
+    boolean hasNftAllowance(Address address, FcTokenAllowanceId id, OnMissing throwIfMissing);
+
+    Token getToken(Address address, OnMissing throwIfMissing);
```
steven-sheehy, line 54: Up until this, the `Store` was returning services models. Now it returns both mirror node models and services models. This is confusing. I think it makes more sense for it to only return services models as it's used mainly within services code. We don't want to start leaking mirror node internals into copied services code if we can help it.

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/UniqueToken.java
```diff
@@ -103,4 +111,14 @@ public String toString() {
                 .add("spender", spender)
                 .toString();
     }
+
+    @Override
+    public boolean equals(final Object obj) {
+        return EqualsBuilder.reflectionEquals(this, obj);
```
steven-sheehy, line 117: We should never use reflection for equals and hashCode

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/UniqueToken.java
0xivanov: Fixed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/Store.java
0xivanov: Removed the methods, returning mirror node models

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmTxProcessorFacadeImpl.java
```diff
@@ -95,15 +101,17 @@ public HederaEvmTransactionProcessingResult execute(
...
-
-        final var codeCache = new AbstractCodeCache(expirationCacheTime, entityAccess);
-        final var store = new StoreImpl(databaseAccessors);
-
-        final var mirrorEvmContractAliases = new MirrorEvmContractAliases(entityAccess);
+        final Store store = new StoreImpl(databaseAccessors);
+        final MirrorEntityAccess mirrorEntityAccess =
+                new MirrorEntityAccess(contractStateRepository, contractRepository, entityRepository, store);
+        final TokenAccessor tokenAccessor = new TokenAccessorImpl(evmProperties, mirrorEntityAccess, store);
+        final AccountAccessor accountAccessor = new AccountAccessorImpl(mirrorEntityAccess);
```
IvanKavaldzhiev, line 108: You can use `var` for all fields

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/token/TokenAccessorImpl.java
```diff
@@ -184,91 +134,83 @@ public EvmKey keyOf(final Address address, final TokenKeyType tokenKeyType) {
...
-                .map(TokenAllowance::getAmount)
-                .orElse(0L);
+        final var tokenNum = EntityNum.fromEvmAddress(token);
+        final var spenderNum = EntityNum.fromEvmAddress(spender);
+
+        final var fcTokenAllowanceId = new FcTokenAllowanceId(tokenNum, spenderNum);
+
+        final var account = store.getAccount(owner, OnMissing.DONT_THROW);
+        if (account.isEmptyAccount()) return 0L;
+        return account.getFungibleTokenAllowances().get(fcTokenAllowanceId);
```
IvanKavaldzhiev, line 171: You can remove line 170 and use directly:

`account.getFungibleTokenAllowances().getOrDefault(0L)`

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/token/TokenAccessorImpl.java
```diff
@@ -277,58 +219,52 @@ public Address canonicalAddress(final Address addressOrAlias) {
...
         return properties.getNetwork().getLedgerId();
     }
 
-    private Optional<EvmTokenInfo> getTokenInfo(final Address token) {
-        final var tokenEntityOptional =
-                tokenRepository.findById(fromEvmAddress(token.toArray()).getId());
-        final var entityOptional = entityRepository.findByIdAndDeletedIsFalse(entityIdNumFromEvmAddress(token));
+    private Optional<EvmTokenInfo> getTokenInfo(final Address address) {
+        final var tokenEntity = store.getToken(address, OnMissing.DONT_THROW);
+        final var entityOptional = mirrorEntityAccess.findEntity(address);
```
IvanKavaldzhiev, line 238: I think that we should entirely use the `token` model from the store. The first time it's fetched we use `tokenRepository` + `entityRepository`. From then on, we read or update the existing token inside the state. If we want to create a new token, a token model is created and saved in the store and again we should read its fields.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/token/TokenAccessorImpl.java
0xivanov: Fixed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/token/TokenAccessorImpl.java
```diff
@@ -16,139 +16,94 @@
...
-        final var entityAddress = entityDatabaseAccessor.evmAddressFromId(nftEntity.getAccountId());
-        final var creationTime = nftEntity.getCreatedTimestamp();
-        final var metadata = nftEntity.getMetadata();
-        final var spender = nftEntity.getSpender() != null
-                ? entityDatabaseAccessor.evmAddressFromId(nftEntity.getSpender())
-                : Address.ZERO;
-        final var nftInfo = new EvmNftInfo(serialNo, entityAddress, creationTime, metadata, spender, ledgerId);
 
+        final var ledgerId = properties.getNetwork().getLedgerId();
+        final var entityAddress = nft.getOwner() != null ? nft.getOwner().asEvmAddress() : Address.ZERO;
```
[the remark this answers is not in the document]
0xivanov: Fixed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmTxProcessorFacadeImpl.java
0xivanov: Fixed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/token/TokenAccessorImpl.java
0xivanov: Fixed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
```diff
@@ -114,6 +123,15 @@ public Bytes fetchCodeIfPresent(final Address address) {
...
 
+    /**
+     * This method does not use {@link Store} and reads directly from the database.
+     * The returned entity does not contain the most recent changes from the transaction.
+     * We should use this data only for fields that are not contained in {@link com.hedera.services.store.models.Token}
+     * and {@link com.hedera.services.store.models.Account} as they cannot be modified during the transaction.
+     * We can also check if entity exists as we cannot remove entity during a transaction.
+     * @param address of the entity
+     * @return entity from the database
+     */
```
IvanKavaldzhiev, line 134: We can define an `Entity` DTO model that `MirrorEntityAccess` could use. It can be defined in `com.hedera.services` package, even though it won't be copied from services repo.

Then, in `findEntity` we can first search for an account via `store.getAccount`:
- if we find it, map it to the `Entity` model and return it
- if not, try to find a token via `store.getToken`
- if we find a token, again map it to the `Entity` model
- if nothing is found - return `Optional.empty()`

@steven-sheehy I'll be happy to see your opinion as well.

steven-sheehy: I think `MirrorEntityAccess.findEntity()` should never have really been exposed since it's implementing a services interface but also exposing mirror node domain classes for `MirrorEvmContractAliases`. Shouldn't we transition all of `MirrorEntityAccess` to use the Store and remove its reliance on mirror node domain as much as possible? Thus `findEntity` should go away.

0xivanov: Removed `MirrorEntityAccess.findEntity()` . Using the store to fetch `Token` or `Account` in place of `Entity`.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/account/MirrorEvmContractAliases.java
```diff
@@ -91,15 +92,18 @@ private Optional<Address> resolveFromAliases(Address alias) {
...
+            id = store.getAccount(addressOrAlias, OnMissing.THROW).getId(); // if token and account are missing - throw
+            entityId =
+                    new EntityId(id.shard(), id.realm(), id.num(), EntityType.TOKEN); // if account is present - get id
+        } else { // if token is present - get id
+            id = token.getId();
+            entityId = new EntityId(id.shard(), id.realm(), id.num(), EntityType.ACCOUNT);
         }
-
-        final var resolvedAddress = Address.wrap(Bytes.wrap(toEvmAddress(entity.toEntityId())));
+        final var resolvedAddress = Address.wrap(Bytes.wrap(toEvmAddress(entityId)));
```
IvanKavaldzhiev, line 106: Instead of constructing `entityId`, you can directly fetch the address from an `account` or if it's a `token`, you can get it's `id` and use i`d.toEvmAddress`

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/account/MirrorEvmContractAliases.java
```diff
@@ -91,15 +92,18 @@ private Optional<Address> resolveFromAliases(Address alias) {
...
-        if (entity.getType() != EntityType.TOKEN && entity.getType() != EntityType.CONTRACT) {
-            throw new InvalidParametersException("Not a contract or token: " + addressOrAlias);
+        final EntityId entityId;
+        Id id;
+        final var token = store.getToken(addressOrAlias, OnMissing.DONT_THROW);
+        if (token.isEmptyToken()) { // if token is missing - check for account
+            id = store.getAccount(addressOrAlias, OnMissing.THROW).getId(); // if token and account are missing - throw
+            entityId =
+                    new EntityId(id.shard(), id.realm(), id.num(), EntityType.TOKEN); // if account is present - get id
+        } else { // if token is present - get id
```
IvanKavaldzhiev, line 102: You can place all comments above the lines, not sideways, to increase readibility.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/token/TokenAccessorImpl.java
```diff
@@ -184,91 +130,82 @@ public EvmKey keyOf(final Address address, final TokenKeyType tokenKeyType) {
     }
 
     @Override
-    public String nameOf(final Address token) {
-        final var tokenId = entityIdFromEvmAddress(token).getId();
-        return tokenRepository.findById(tokenId).map(Token::getName).orElse("");
+    public String nameOf(final Address address) {
+        return store.getToken(address, OnMissing.DONT_THROW).getName();
```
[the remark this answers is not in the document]
0xivanov: Nice catch! Changed `Token` to return empty strings and added tests for this case.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/account/MirrorEvmContractAliases.java
0xivanov: Fixed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/account/MirrorEvmContractAliases.java
0xivanov: Fixed

steven-sheehy (approved):
LGTM

PR #6034, opened 2023-05-19
Add support for historical account details on accounts endpoint

### LINKED ISSUE #5862: Add support for historical account details on accounts API
### Problem

Currently, the `api/v1/accounts/{account}` supports a `timestamp` param.
This is a legacy param that relates to the transactions listed for the account and not the account details.
This means calling `api/v1/accounts/{account}?timestamp=x` in the past currently returns information that is not correct for that timestamp and therefore the endpoint doesn't support historical information as some users may understand it.

### Solution

Since the mirror node does have historical tables for entities and balances it is possible to surface account details that were valid at the timestamp provided.

### Alternatives

_No response_

jnels124:
**Description**:

Modify the data returned by the `/api/v1/accounts/{ID}` endpoint when passing the timestamp query parameter.

* Ensure provided timestamp params provide effective range
* Union entity query with `entity_history` for specified timestamp
* Return balance by querying the `account_balance` table for matching timestamp query
* Filter token balances from `token_account` by matching `created_timestamp` against timestamp query

**Related issue(s)**:

Fixes #5862 

**Notes for reviewer**:

**Checklist**


hedera-mirror-rest/accounts.js
```diff
@@ -133,45 +139,79 @@ const getEntityBalanceQuery = (
...
+          UNION ALL
+          ${getEntitySql(Entity.historyTableName, balanceSelect, consensusTimestampSelect, orderBy, additionalJoin)} 
+          order by ${Entity.TIMESTAMP_RANGE} desc limit 1 `
+    );
+
+    utils.mergeParams(
+      params,
+      tokenBalanceQuery.params,
+      accountBalanceQuery.params,
+      entityBalanceQuery.params,
```
jnels124, line 195: will fix order

jnels124:
Generated Query

```sql
(with latest_token_balance as (select account_id, balance, token_id, created_timestamp
                                    from token_account
                                    where associated is true)
       select e.alias,
e.auto_renew_period,
e.created_timestamp,
e.decline_reward,
e.deleted,
e.ethereum_nonce,
e.evm_address,
e.expiration_timestamp,
e.id,
e.key,
e.max_automatic_token_associations,
e.memo,
e.receiver_sig_required,
e.staked_account_id,
e.staked_node_id,
e.stake_period_start,
e.type,
e.timestamp_range,
(case when es.pending_reward is null then 0
         when e.decline_reward is true or coalesce(e.staked_node_id, -1) = -1 then 0
         when e.stake_period_start >= es.end_stake_period then 0
         else es.pending_reward
    end) as pending_reward,
                       (select max(consensus_end) from record_file)  as consensus_timestamp,
                       balance,
                       (select json_agg(jsonb_build_object('token_id', token_id, 'balance', balance)) ::jsonb
                        from (select token_id, balance
                              from latest_token_balance
                              where account_id = ? and created_timestamp = ?
                              order by token_id asc
        limit 1000) as account_token_balance) as token_balances
                from entity e
                         left join entity_stake es on es.id = e.id

                where e.type in ('ACCOUNT', 'CONTRACT') and e.id = ? and (es.id = ? OR es.id IS NULL) and e.timestamp_range && ? order by id asc)

          UNION ALL

      (with latest_token_balance as (select account_id, balance, token_id, created_timestamp
                                    from token_account
                                    where associated is true)
       select e.alias,
e.auto_renew_period,
e.created_timestamp,
e.decline_reward,
e.deleted,
e.ethereum_nonce,
e.evm_address,
e.expiration_timestamp,
e.id,
e.key,
e.max_automatic_token_associations,
e.memo,
e.receiver_sig_required,
e.staked_account_id,
e.staked_node_id,
e.stake_period_start,
e.type,
e.timestamp_range,
(case when es.pending_reward is null then 0
         when e.decline_reward is true or coalesce(e.staked_node_id, -1) = -1 then 0
         when e.stake_period_start >= es.end_stake_period then 0
         else es.pending_reward
    end) as pending_reward,
                       ab.consensus_timestamp as consensus_timestamp,
                       ab.balance as balance,
                       (select json_agg(jsonb_build_object('token_id', token_id, 'balance', balance)) ::jsonb
                        from (select token_id, balance
                              from latest_token_balance
                              where account_id = ? and created_timestamp = ?
                              order by token_id asc
        limit 1000) as account_token_balance) as token_balances
                from entity_history e
                         left join entity_stake es on es.id = e.id
                         left join (select account_id, balance, consensus_timestamp
                    from account_balance
                    where account_id = ? and consensus_timestamp = ?
                    limit 1) ab on ab.account_id = e.id
                where e.type in ('ACCOUNT', 'CONTRACT') and e.id = ? and (es.id = ? OR es.id IS NULL) and e.timestamp_range && ? order by timestamp_range desc limit 1)
          order by timestamp_range desc limit 1;
```

hedera-mirror-rest/accounts.js
```diff
@@ -297,23 +346,50 @@ const getAccounts = async (req, res) => {
...
+  const entityAccountQuery = {query: 'e.id = ? and (es.id = ? OR es.id IS NULL)', params: accountIdParams};
+  const tokenBalanceQuery = {
+    query: 'account_id = ?',
+    params: tokenBalanceParams,
+    limit: tokenBalanceResponseLimit.singleAccount,
+  };
+  const accountBalanceQuery = {query: '', params: []};
+
+  if (transactionTsQuery) {
+    tokenBalanceQuery.query += ` and ${transactionTsQuery.replace('t.consensus_timestamp', 'created_timestamp')} `;
```
jnels124, line 376: do we want to use the timestamp_range instead?

hedera-mirror-rest/accounts.js
```diff
@@ -297,23 +346,50 @@ const getAccounts = async (req, res) => {
...
+    tokenBalanceQuery.params = tokenBalanceParams.concat(transactionTsParams);
+
+    entityAccountQuery.query += ` and ${entityTsQuery.join(' and ')}`;
+    entityAccountQuery.params = entityAccountQuery.params.concat(entityTsParams);
+
+    const accountBalanceTs = await balances.getAccountBalanceTimestamp(balanceFileTsQuery, balanceFileTsParams);
+    accountBalanceQuery.query = 'account_id = ? and consensus_timestamp = ?';
+    accountBalanceQuery.params = [encodedId, accountBalanceTs ?? null];
```
jnels124, line 384: in some cases `accountBalanceTs ` can be undefined. In this case, we can't find balance information. Should the response indicate null/zero balance in this case, or should we grab the  balance, timestamp_range end  from the entity`? The only ways I can see this happening are, on a new account that doesn't have a balance file (possibly an account create creates an entry there but didn't verify), or if multiple timestamp parameters are supplied that don't include the range of a balance file.

hedera-mirror-test/k6/src/rest/test/accountsIdTimestampLte.js
```diff
@@ -0,0 +1,34 @@
...
+
+const {options, run, setup} = new RestTestScenarioBuilder()
+  .name('accountsIdTimestampLte') // use unique scenario name among all tests
+  .tags({url: urlTag})
+  .request((testParameters) => {
+    const url = `${testParameters['BASE_URL_PREFIX']}/accounts/${testParameters['DEFAULT_ACCOUNT_ID']}?timestamp=lte:${testParameters['DEFAULT_ACCOUNT_ID_TIMESTAMP']}`;
+    return http.get(url);
+  })
+  .requiredParameters('DEFAULT_ACCOUNT_ID', 'DEFAULT_ACCOUNT_ID_TIMESTAMP')
+  .check('Accounts Id OK', isSuccess)
```
jnels124, line 31: will fix label

hedera-mirror-rest/accounts.js
```diff
@@ -133,45 +139,79 @@ const getEntityBalanceQuery = (
...
+                from ${tableName} e
+                         left join entity_stake es on es.id = e.id
+                         ${additionalJoin}
+                where ${whereCondition} ${orderBy})`;
+  };
+  const queries = [getEntitySql(Entity.tableName)];
+
+  if (accountBalanceQuery.query) {
+    const consensusTimestampSelect = 'ab.consensus_timestamp as consensus_timestamp';
+    const balanceSelect = 'ab.balance as balance';
```
jnels124, line 176: should we coalesce with the entity_history.balance ?

hedera-mirror-rest/api/v1/openapi.yml
```diff
@@ -28,6 +28,8 @@ paths:
       summary: Get account by alias, id, or evm address
       description: |
         Return the account transactions and balance information given an account alias, an account id, or an evm address. The information will be limited to at most 1000 token balances for the account as outlined in HIP-367.
+        When the timestamp parameter is supplied, we will return transactions and account state for the relevant timestamp query. Balance information will be accurate to within 15 minutes of the provided timestamp query.
+        Historical stake and reward information is not available so these fields contain current data.
```
Nana-EC, line 32: Thanks for this line as I was going to ask for it.
According to @steven-sheehy the `ethereum_nonce` value may also not be valid when it's historical. 
Can we note its limited details also and open a ticket for that similar to what you did for staking? 
If it is indeed accurate then great as that was one of the main driving factors behind the ticket.

hedera-mirror-rest/controllers/contractController.js
```diff
@@ -426,6 +375,8 @@ const getAndValidateContractIdAndConsensusTimestampPathParams = async (req) => {
 };
 
 const extractContractIdAndFiltersFromValidatedRequest = (req, acceptedParameters) => {
+  logger.debug('The req query is ');
```
Nana-EC, line 378: SHould this be similar to the below example to be in line with other logging of queries?
```
if (logger.isTraceEnabled()) {
      logger.trace(`? query: ${req.query}, params: ${acceptedParameters}`);
    }
```

jnels124: intended to remove that. thanks

hedera-mirror-rest/api/v1/openapi.yml
jnels124: Yes will do.

hedera-mirror-rest/accounts.js
```diff
@@ -133,45 +139,91 @@ const getEntityBalanceQuery = (
...
+                    where ${accountBalanceQuery.query}
+                    order by consensus_timestamp desc
+                    limit 1) ab on ab.account_id = e.id`;
+    const orderBy = `order by ${Entity.TIMESTAMP_RANGE} desc limit 1`;
+
+    queries.push(
+      `
+          ${getEntitySql(Entity.tableName, balanceSelect, consensusTimestampSelect, orderBy, additionalJoin)}
+          UNION ALL
+          ${getEntitySql(
```
steven-sheehy, line 192: I think this is the wrong way to go about things. Here you're doing two separate queries (one history and one main) where each of those has separate joins on tables whose information would not vary. We should do one sub query for entity union entity_history then join the results of that query on other tables.

hedera-mirror-rest/accounts.js
```diff
@@ -133,45 +139,91 @@ const getEntityBalanceQuery = (
...
+            Entity.historyTableName,
+            balanceSelect,
+            consensusTimestampSelectUnion,
+            orderBy,
+            additionalJoin
+          )} 
+          order by ${Entity.TIMESTAMP_RANGE} desc limit 1 `
+    );
+
+    utils.mergeParams(
```
steven-sheehy, line 202: Per your comment in the ticket, there are a lot of duplicate parameters:

```
[111146,"1684921798808550755",111146,1570810500024187000,111146,111146,"(,1684921798808550755]",111146,"1684921798808550755",111146,1570810500024187000,111146,111146,"(,1684921798808550755]"]
```
We should clean that up so that there's only one `$N` parameter for each distinct value.

hedera-mirror-rest/accounts.js
```diff
@@ -133,45 +139,91 @@ const getEntityBalanceQuery = (
...
+                                    from token_account
+                                    where associated is true)
+       select ${entityFields},
+                       ${consensusTimestampSelect},
+                       ${balanceSelect},
+                       (select json_agg(jsonb_build_object('token_id', token_id, 'balance', balance)) ::jsonb
+                        from (select token_id, balance
+                              from latest_token_balance
+                              where ${tokenBalanceQuery.query}
+                              order by token_id ${order}
```
steven-sheehy, line 166: It's strange that this uses a CTE when it's not necessary here. Can we push the CTE down into the sub-query and remove the CTE?

hedera-mirror-rest/accounts.js
steven-sheehy, line 192: For the union, we can use wildcards for the fields within the sub-queries and only explicitly select the fields outside it. Something like `select e.id, e.balance from (select * from entity union all select * from entity_history) as e`.

hedera-mirror-rest/accounts.js
```diff
@@ -133,45 +139,91 @@ const getEntityBalanceQuery = (
...
+
+  if (accountBalanceQuery.query) {
+    const consensusTimestampSelect =
+      'COALESCE(ab.consensus_timestamp, (select max(consensus_end) from record_file)) as consensus_timestamp';
+    const consensusTimestampSelectUnion =
+      'COALESCE(ab.consensus_timestamp, upper(e.timestamp_range)) as consensus_timestamp';
+    const balanceSelect = 'COALESCE(ab.balance, e.balance) as balance';
+    const additionalJoin = `left join (select account_id, balance, consensus_timestamp
+                    from account_balance
+                    where ${accountBalanceQuery.query}
```
steven-sheehy, line 183: We should first search account_balance_file to get timestamp then use that for account_balance since the former is much smaller.

steven-sheehy (changes requested):

hedera-mirror-rest/accounts.js
xin-hedera: I don't think we have a good solution. What's missing and very important in the `timestamp` query param handling is we don't have application side logic to calculate a single effective range from all the `timestamp` values in the request url.

Without that , when `balances.getAccountBalanceTimestamp` returns `undefined`, we don't know if it's because there is no account balance file's timestamp satisfying the `timestamp` range or the effective `timestamp` range is empty.

The main obstacle is we support the weird `ne` not equal operator, e.g., with `timestamp=ne:100`, the range becomes `timestamp in (,100) or timestamp in (100,)`.

Assume we drop `ne` operator support, we can then

- calculate the effective timestamp range
- if the range is empty, return 400
- if the range has upper bound, disregard the lower bound, and find the last account balance file satisfying the upper bound
- if the range has lower bound and no upper bound, find the first account balance file satisfying the lower bound
- if account balance file is found but the account balance is not in it, use balance from entity / entity_history
- if no account balance file found, use balance from entity / entity_history
- if no row found from entity / entity_history satisfying the timestamp range, return 404

hedera-mirror-rest/accounts.js
jnels124: Not sure the best way to go about this without using named parameters which isn't available. I will look for additional examples in the code base but so far what i have seen is duplicated parameters.

hedera-mirror-rest/accounts.js
jnels124: this is done in the calling method when populating the accountBalanceQuery

hedera-mirror-rest/accounts.js
jnels124: That was a big miss on my part. I will update the query to be more similar to 

```sql
with entity_data as (select *
                     from entity e
                     where e.type in ('ACCOUNT', 'CONTRACT')
                       and e.id = 45385
                       and e.timestamp_range && '(,1680288435601059000]'
                     UNION ALL
                     select *
                     from entity_history e
                     where e.type in ('ACCOUNT', 'CONTRACT')
                       and e.id = 45385
                       and e.timestamp_range && '(,1680288435601059000]'
                     ORDER BY timestamp_range desc
                     LIMIT 1),
     latest_token_balance as (select account_id, balance, token_id, created_timestamp
                              from token_account
                              where associated is true)
SELECT e.alias,
       e.auto_renew_period,
       e.created_timestamp,
       e.decline_reward,
       e.deleted,
       e.ethereum_nonce,
       e.evm_address,
       e.expiration_timestamp,
       e.id,
       e.key,
       e.max_automatic_token_associations,
       e.memo,
       e.receiver_sig_required,
       e.staked_account_id,
       e.staked_node_id,
       e.stake_period_start,
       e.type,
       e.timestamp_range,
       (case
            when es.pending_reward is null then 0
            when e.decline_reward is true or coalesce(e.staked_node_id, -1) = -1 then 0
            when e.stake_period_start >= es.end_stake_period then 0
            else es.pending_reward
           end)                                                                       as pending_reward,
       COALESCE(ab.consensus_timestamp, (select max(consensus_end) from record_file)) as consensus_timestamp,
       COALESCE(ab.balance, e.balance)                                                as balance,
       (select json_agg(jsonb_build_object('token_id', token_id, 'balance', balance)) ::jsonb
        from (select token_id, balance
              from latest_token_balance
              where account_id = 45385
                and created_timestamp <= 1680288435601059000
              order by token_id asc
              limit 1000) as account_token_balance)                                   as token_balances
from entity_data e
         left join entity_stake es on es.id = e.id
         left join account_balance ab on ab.account_id = e.id and consensus_timestamp = 1680288301107978238;
```

hedera-mirror-rest/accounts.js
steven-sheehy: Ok, didn't see it in your example query above.

hedera-mirror-rest/__tests__/specs/accounts/{id}/timestamp-notfound.json
```diff
@@ -0,0 +1,166 @@
...
+            "amount": 200,
+            "is_approval": true
+          }
+        ]
+      }
+    ]
+  },
+  "urls": [
+    "/api/v1/accounts/0.0.8?timestamp=lte:1234567880.000000006",
+    "/api/v1/accounts/0.0.8?timestamp=lte:1234567890.000000007&timestamp=ne:1234567890.000000004",
```
steven-sheehy, line 153: not equals should result in an invalid param error

steven-sheehy (changes requested):

hedera-mirror-rest/accounts.js
```diff
@@ -297,23 +358,91 @@ const getAccounts = async (req, res) => {
...
+
+    entityAccountQuery.query += ` and ${entityTsQuery.join(' and ')}`;
+    entityAccountQuery.params = entityAccountQuery.params.concat(entityTsParams);
+
+    const accountBalanceTs = await balances.getAccountBalanceTimestamp(balanceFileTsQuery, balanceFileTsParams, order);
+    const neqTsFilters = filters.filter(
+      (filter) => filter.key === filterKeys.TIMESTAMP && filter.operator === opsMap.ne
+    );
+    const timestampExcluded = neqTsFilters.reduce(
+      (prev, curr) => prev || parseInteger(curr.value) === accountBalanceTs,
```
xin-hedera, line 427: it should be cheaper to just do `==` with auto type coercion, for timestamp, filter.value is a string, `accountBalanceTs` is either a number or a BigInt
 
```suggestion
      (prev, curr) => prev || curr.value == accountBalanceTs,
```

hedera-mirror-rest/__tests__/specs/accounts/{id}/timestamp-notfound.json
xin-hedera: I think the consensus is don't break it by invalid param error, query the database with `ne` ignored, and apply the `ne` filters on the result

xin-hedera (changes requested):

hedera-mirror-rest/utils.js
```diff
@@ -1384,61 +1440,82 @@ const getStakingPeriod = (stakingPeriod) => {
...
-  const earliest =
-    valuesByOp[opsMap.gt].length > 0 ? BigInt(valuesByOp[opsMap.gt][0]) + 1n : BigInt(valuesByOp[opsMap.gte][0]);
-  const latest =
-    valuesByOp[opsMap.lt].length > 0 ? BigInt(valuesByOp[opsMap.lt][0]) - 1n : BigInt(valuesByOp[opsMap.lte][0]);
-  const difference = latest - earliest + 1n;
+  // there should be exactly one gt/gte and/or one lt/lte at this point
+  const earliest = valuesByOp[opsMap.gt].length ? valuesByOp[opsMap.gt][0] + 1n : valuesByOp[opsMap.gte][0];
+  const latest = valuesByOp[opsMap.lt].length ? valuesByOp[opsMap.lt][0] - 1n : valuesByOp[opsMap.lte][0];
 
+  const difference = latest && earliest ? latest - earliest + 1n : undefined;
```
xin-hedera, line 1509: if latest or earliest is 0n, `latest && earliest` is 0n, same as boolean false. so need to check strictly `latest !== undefined && earliest !== undefined`

hedera-mirror-rest/utils.js
```diff
@@ -1494,6 +1571,77 @@ const convertGasPriceToTinyBars = (gasPrice, hbarsPerTinyCent, centsPerHbar) =>
...
+};
+const buildTimestampQuery = (range, column, neValues = [], eqValues = [], buildAsIn = true) => {
+  const conditions = [];
+  const params = [];
+
+  if (range) {
+    const bounds = [...range.bounds];
+    if (bounds.length !== 2) {
+      return ['', []];
+    }
```
xin-hedera, line 1588: isn't this dead code? range always has a left bound (open or closed) and right bound

hedera-mirror-rest/utils.js
```diff
@@ -1494,6 +1571,77 @@ const convertGasPriceToTinyBars = (gasPrice, hbarsPerTinyCent, centsPerHbar) =>
...
+    if (range.begin) {
+      conditions.push(`${column}${boundToOp[bounds[0]]} ?`);
```
xin-hedera, line 1590: nit
```suggestion
      conditions.push(`${column} ${boundToOp[bounds[0]]} ?`);
```

hedera-mirror-rest/utils.js
```diff
@@ -1494,6 +1571,77 @@ const convertGasPriceToTinyBars = (gasPrice, hbarsPerTinyCent, centsPerHbar) =>
...
+  const params = [];
+
+  if (range) {
+    conditions.push(`${column} && ?`);
+    params.push(range);
+  }
+  if (neValues.length) {
+    neValues.forEach((value) => {
+      conditions.push(`not ${column} @> ?`);
+      params.push(Range(value, value, '[]'));
```
xin-hedera, line 1631: can just use the value
```suggestion
      params.push(value);
```

hedera-mirror-rest/balances.js
```diff
@@ -154,13 +154,16 @@ const getBalances = async (req, res) => {
...
     select consensus_timestamp
     from account_balance_file
-    where ${tsQuery}
-    order by consensus_timestamp desc
+     ${tsQuery ? ' where ' : ''} ${tsQuery}
+    order by consensus_timestamp ${order}
     limit 1`;
+
+  logger.info(`The query is ${query}`);
+  logger.info(tsParams);
```
xin-hedera, line 166:
```suggestion
```

xin-hedera (changes requested):

hedera-mirror-rest/utils.js
jnels124: That is true but in the case of 0n 0n, the difference becomes 1n due to the + 1n. In that case, it will neither be > maxTimestampRangeNs and will not be <= 0n so this will not trigger the invalidArgumentError. I will update it to explicitly check for undefined though because that is more clear.

hedera-mirror-rest/utils.js
jnels124: yes this could probably be removed.

hedera-mirror-rest/utils.js
jnels124: This would have to be a range because it is using the range operator `@>`

hedera-mirror-rest/accounts.js
```diff
@@ -297,23 +358,78 @@ const getAccounts = async (req, res) => {
...
+      [],
+      eqValues,
+      false
+    );
+    const accountBalanceTs = await balances.getAccountBalanceTimestamp(
+      balanceFileTsQuery.replaceAll(opsMap.eq, opsMap.lte),
+      balanceFileTsParams,
+      order
+    );
+    const timestampExcluded = neValues.reduce((prev, curr) => prev || curr === accountBalanceTs, false);
```
xin-hedera, line 416: I see the neValues is an array of `BigInt` now, however, `accountBalanceTs` can be either a `Number` or a `BigInt`, so we'll end up with `10n === 10` evaluated to false, we can do either of the two

- use `utils.parseInteger` instead of `BigInt` in `checkTimestampRange`
- use `curr == accountBalanceTs` here and let js do the coercion

in addition, we should do `Array.prototype.some` instead of reduce
```suggestion
    const timestampExcluded = neValues.some((value) => value == accountBalanceTs);
```

xin-hedera (approved):
LGTM

steven-sheehy (approved):
LGTM

PR #6060, opened 2023-05-22
 Add support for node ID based bucket paths in MetricsExecutionInterce…

### LINKED ISSUE #6023: Update MetricsExecutionInterceptor to recognize node ID based format in S3 URI
It was [pointed out](https://github.com/hashgraph/hedera-mirror-node/pull/5753/#discussion_r1196631440) in the PR review for 4388 that an S3 URI used to download a stream file using the new HIP-679 node ID based bucket structure will not be parsed properly by `MetricsExecutionInterceptor.getNodeAccountId()`.

This does not cause importer ingestion to fail, but generates errors in the logs, and impacts the accuracy of the metrics. In production, this will not become an issue until consensus nodes actually start to write stream files to cloud storage using the new bucket structure, which is as of yet unknown. But, this issue should be resolved soon.

- In `getNodeAccountId()` add check for node ID based download after the existing (account ID) regex fails to match.
- Examine existing code to see how the node should be represented in the metrics.  One approach may be to preserve the current implementation and to convert the HIP-679 node ID path (e.g. `{network}/{shard}/{nodeID}/record/`) into an `EntityId` by utilizing `{shard}`, assuming a realm of `0` and `{NodeID}+3` as the node account ID.

jascks:
**Description**:

Update `MetricsExecutionInterceptor` to recognize and process node ID based format in S3 URI in addition to current node account ID format.

* `MetricsExecutionInterceptor` refactoring and natural incorporation of pattern matching and parsing of both the legacy and new bucket structures. Also, allowed for introduction of mock builders for testing.
* Create `MetricsExecutionInterceptorTest` to test the new functionality and verify the old.

**Related issue(s)**:

Fixes #6023

**Notes for reviewer**:

The new test class is specific to the problem this PR resolves and is not a general test class for all `MetricsExecutionInterceptor` functionality. Perhaps it will grow into one in the future.

**Checklist**


hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/config/MetricsExecutionInterceptor.java
```diff
@@ -87,19 +107,18 @@ public Optional<Publisher<ByteBuffer>> modifyAsyncHttpResponseContent(
...
 
             String[] tags = {
-                "action", getAction(uri),
+                "action", uriAttributes.action(),
                 "method", context.httpRequest().method().name(),
-                "nodeAccount", nodeAccountId.getEntityNum().toString(),
-                "realm", nodeAccountId.getRealmNum().toString(),
-                "shard", nodeAccountId.getShardNum().toString(),
+                "node", uriAttributes.nodeId().toString(),
+                "realm", uriAttributes.realm().toString(),
```
steven-sheehy, line 118: I think we should drop realm now that we can't capture it. We will probably never have a non-zero realm anyway.

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/config/MetricsExecutionInterceptorTest.java
```diff
@@ -0,0 +1,227 @@
...
+ * This test class was introduced with issue 6023, and is focused around the MetricsExecutionInterceptor
+ * afterExecution() method, specifically the micrometer metrics tags and values that are produced depending
+ * on the S3 activity being intercepted.
+ *
+ * This required a bit of refactoring to make it possible to introduce the mock Timer.Builder and
+ * DistributionSummary.Builder instances required for testing. The metric tags are captured via the
+ * mock Timer.Builder only, and the DistributionSummary.Builder mock is ignored. The tags are then examined for the
+ * expected results for the scenario. No other verification of MetricsExecutionInterceptor is performed at this time.
+ */
+class MetricsExecutionInterceptorTest {
```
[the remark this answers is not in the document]
jascks: When I saw there was no unit test class for `MetricsExecutionInterceptor`, my first inclination was to maybe settle for manual testing by looking at the metrics produced in actuator. But I wanted to have a unit test for both for the automation benefits as well as easily throwing a lot of different scenarios at it. And, it helps with test coverage. :)

In my opinion, there is nothing wrong with modifying a class in order to make it more testable, particularly since the class under test and the test class are in the same package, and that can be taken advantage of while leaving existing public access un-impacted.

We're basically only doing string pattern matching on S3 URIs to determine the values to assign to the tags, and then verify the proper values. I can see why there was not a preexisting unit test class, given it was kind of painful to get access to the generated tags. But, I've got it, and it works great!

For this level of testing, is there a need for S3 proxy, s3 client, and hosting actual files etc? I admit, I was unaware of `SimpleMeterRegistry`, and perhaps it can be helpful within the unit test and allow me to eliminate some mocks and revert some of the changes to `MetricsExecutionInterceptor`.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/config/MetricsExecutionInterceptor.java
```diff
@@ -117,6 +136,29 @@ public void afterExecution(Context.AfterExecution context, ExecutionAttributes e
         }
     }
 
+    private UriAttributes getUriAttributes(URI uri) {
+        var decodedUri = URLDecoder.decode(uri.toString(), StandardCharsets.UTF_8);
+        var action = getAction(decodedUri);
+
+        Matcher accountIdMatcher = ENTITY_ID_PATTERN.matcher(decodedUri);
+        if (accountIdMatcher.find() && accountIdMatcher.groupCount() == 1) {
+            var streamType = getType(t -> decodedUri.contains(t.getPath()));
```
steven-sheehy, line 145: If you change the `ENTITY_ID_PATTERN` to `\/(balance|event|record)(s_)?(\d{1,10})\.\d{1,10}\.(\d{1,10})` then you can extract the type, shard and num directly  from the single regex and don't need `EntityId` or `getType()`.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/config/MetricsExecutionInterceptor.java
```diff
@@ -117,6 +136,29 @@ public void afterExecution(Context.AfterExecution context, ExecutionAttributes e
         }
     }
 
+    private UriAttributes getUriAttributes(URI uri) {
+        var decodedUri = URLDecoder.decode(uri.toString(), StandardCharsets.UTF_8);
```
steven-sheehy, line 140: Perhaps instead of decoding the URI or treating as a string we just pass parts of the URI to various matchers/contains. For example, for `ENTITY_ID_PATTERN` and `NODE_ID_PATTERN ` we should only need to pass `uri.getPath()`.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/config/MetricsExecutionInterceptor.java
```diff
@@ -117,6 +136,29 @@ public void afterExecution(Context.AfterExecution context, ExecutionAttributes e
...
+        if (accountIdMatcher.find() && accountIdMatcher.groupCount() == 1) {
+            var streamType = getType(t -> decodedUri.contains(t.getPath()));
+            var entityId = EntityId.of(accountIdMatcher.group(1), EntityType.ACCOUNT);
+            return new UriAttributes(
+                    entityId.getShardNum(), entityId.getRealmNum(), entityId.getEntityNum() - 3L, streamType, action);
+        }
+
+        Matcher nodeIdMatcher = NODE_ID_PATTERN.matcher(decodedUri);
+        if (nodeIdMatcher.find() && nodeIdMatcher.groupCount() == 3) {
+            var streamType = getType(t -> decodedUri.contains(t.getNodeIdBasedSuffix()));
```
steven-sheehy, line 153: We can extract the type from the URI path instead of a separate contains.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/config/MetricsExecutionInterceptor.java
```diff
@@ -130,26 +172,17 @@ private String getAction(String uri) {
...
-            if (uri.contains(streamType.getPath())) {
+            if (matcher.test(streamType)) {
                 return streamType.name();
             }
         }
-
         return "UNKNOWN";
     }
 
+    private record UriAttributes(Long shard, Long realm, Long nodeId, String streamType, String action) {}
```
steven-sheehy, line 184:
```suggestion
    private record UriAttributes(String action, long node, long, shard, String type) {}
```

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/config/MetricsExecutionInterceptorTest.java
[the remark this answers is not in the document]
steven-sheehy: I don't think we have to go the full s3proxy route, but we definitely should not mock the `MeterRegistry` or any Micrometer related classes. So let's change to use `SimpleMeterRegistry` so that we can properly verify the metrics are populated. You can see `SubscribeMetricsTest` for an example.

steven-sheehy (changes requested):

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/config/MetricsExecutionInterceptorTest.java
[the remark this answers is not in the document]
jascks: Ok, thanks!

steven-sheehy:
Need to update the Grafana dashboards as well to remove references to realm or nodeAccount

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/config/MetricsExecutionInterceptor.java
jascks: I wonder if we should drop `method` as well:
```
hedera_mirror_download_response_bytes_sum{action="list",application="hedera-mirror-importer",method="GET",node="3",realm="0",shard="0",status="200",type="BALANCE",} 6284.0
hedera_mirror_download_response_bytes_count{action="signature",application="hedera-mirror-importer",method="GET",node="1",realm="0",shard="0",status="200",type="RECORD",} 1.0
```
It appears list and object get are both http GETs, and those are the only S3 operations we make.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/config/MetricsExecutionInterceptor.java
jascks: Cool, I like it. `NODE_ID_PATTERN` already provides the HIP-679 suffix of same values as well.

xin-hedera (changes requested):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/config/MetricsExecutionInterceptor.java
steven-sheehy: It's fine to drop. Just double check Grafana visualizations to see if it's used.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/config/MetricsExecutionInterceptor.java
jascks: Done

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/config/MetricsExecutionInterceptor.java
jascks: Done.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/config/MetricsExecutionInterceptor.java
jascks: Done

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/config/MetricsExecutionInterceptor.java
jascks: Dropped `realm`, still looking into `method`...

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/config/MetricsExecutionInterceptorTest.java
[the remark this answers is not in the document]
jascks: Okay, this is done. Using `SimpleMeterRegistry` I can access the tags for verification, and I was able to return the `MetricsExecutionInterceptor` constructor and metrics builders to their original state.

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/config/MetricsExecutionInterceptorTest.java
```diff
@@ -0,0 +1,225 @@
...
+                new MetricsExecutionInterceptor(meterRegistry, timerBuilder, distributionSummaryBuilder);
+    }
+
+    @AfterEach
+    void tearDown() throws Exception {
+        openMocksCloseable.close();
+    }
+
+    @ParameterizedTest
+    @ValueSource(strings = {"ACCOUNT_ID", "NODE_ID"})
```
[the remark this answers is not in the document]
jascks: Done. Note that `PathType` as a string must still be used for `s3GetObjectExecution()` given the `@CsvSource`.

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/config/MetricsExecutionInterceptorTest.java
```diff
@@ -0,0 +1,225 @@
...
+
+    @Captor
+    private ArgumentCaptor<String[]> timerTagsCaptor;
+
+    private AutoCloseable openMocksCloseable;
+    private MetricsExecutionInterceptor metricsExecutionInterceptor;
+
+    @BeforeEach
+    void setup() {
+        openMocksCloseable = MockitoAnnotations.openMocks(this);
```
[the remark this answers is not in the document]
jascks: Done

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/config/MetricsExecutionInterceptorTest.java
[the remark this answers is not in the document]
jascks: Well,  never mind, per your later suggestion, those strings can be converted to enum instances automatically. Nice.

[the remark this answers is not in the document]
xin-hedera: that's not true. I just reconfirmed with the following change

```java
void s3GetObjectExecution(CommonDownloaderProperties.PathType pathType, String streamTypeName, String fileName, String expectedAction) {
```

and 
```java
        var prefix = pathType == PathType.ACCOUNT_ID
                ? accountIdPrefix(streamType, SHARD, REALM, ACCOUNT_NUM)
                : nodeIdPrefix(streamType, NETWORK_NAME, SHARD, NODE_ID);
```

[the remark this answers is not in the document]
jascks: Yes, you are correct. :). This has been done.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/config/MetricsExecutionInterceptor.java
```diff
@@ -117,6 +117,30 @@ public void afterExecution(Context.AfterExecution context, ExecutionAttributes e
...
     }
 
+    private UriAttributes getUriAttributes(URI uri) {
+        var query = uri.getQuery();
+        var uriComponent = query != null && query.contains(START_AFTER) ? query : uri.getPath();
+        var action = getAction(uriComponent);
+
+        Matcher accountIdMatcher = ENTITY_ID_PATTERN.matcher(uriComponent);
+        if (accountIdMatcher.find() && accountIdMatcher.groupCount() == 4) {
+            var shard = Long.parseLong(accountIdMatcher.group(3));
```
xin-hedera, line 127: nit, can change the type of `shard` to String in the record, then don't have to parse the string to long here

xin-hedera (approved):
LGTM

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/config/MetricsExecutionInterceptorTest.java
```diff
@@ -0,0 +1,216 @@
...
+        assertNotNull(providedTags);
+        assertEquals(6, providedTags.size());
+
+        // The tags are in a defined order, with the tag name followed by its value. All are Strings.
+        assertEquals(expectedAction, providedTags.get(0).getValue());
+        assertEquals("GET", providedTags.get(1).getValue());
+        assertEquals(expectedNodeId, Long.valueOf(providedTags.get(2).getValue()));
+        assertEquals("0", providedTags.get(3).getValue());
+        assertEquals("200", providedTags.get(4).getValue());
+        assertEquals(expectedStreamType.name(), providedTags.get(5).getValue());
```
steven-sheehy, line 185: We prefer using assertThat() over JUnit asserts and chaining multiple together.

```suggestion
        assertThat(meterRegistry.find("hedera.mirror.download.request").timers())
            .hasSize(1)
            .first()
            .returns(1L, t -> t.count())
            .returns(?, t -> t.mean(TimeUnit.SECONDS))
            .returns(?, t -> t.max(TimeUnit.SECONDS))
            .returns(expectedAction, t -> t.getId().getTag(TAG_ACTION))
            ...
```

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/config/MetricsExecutionInterceptorTest.java
```diff
@@ -0,0 +1,216 @@
...
+        return "%s/%d/%d/%s/".formatted(network, shard, nodeId, streamType.getNodeIdBasedSuffix());
+    }
+
+    private String accountIdPrefix(StreamType streamType, long shard, long realm, long accountNum) {
+        return "%s/%s%d.%d.%d/".formatted(streamType.getPath(), streamType.getNodePrefix(), shard, realm, accountNum);
+    }
+
+    private void verifyTimerTags(String expectedAction, long expectedNodeId, StreamType expectedStreamType) {
+        Collection<Timer> timers =
+                meterRegistry.find("hedera.mirror.download.request").timers();
```
steven-sheehy, line 172: Move the string to a protected class constant for sharing. Same with other uses of metric names.

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/config/MetricsExecutionInterceptorTest.java
```diff
@@ -0,0 +1,216 @@
...
+        assertEquals(1, timers.size());
+
+        List<Tag> providedTags = timers.iterator().next().getId().getTags();
+        assertNotNull(providedTags);
+        assertEquals(6, providedTags.size());
+
+        // The tags are in a defined order, with the tag name followed by its value. All are Strings.
+        assertEquals(expectedAction, providedTags.get(0).getValue());
```
steven-sheehy, line 180: We should lookup via the string so it's clear. Move the tag names to static constants in interceptor.

steven-sheehy (changes requested):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/config/MetricsExecutionInterceptor.java
jascks: Done

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/config/MetricsExecutionInterceptorTest.java
jascks: Done

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/config/MetricsExecutionInterceptorTest.java
jascks: Done

steven-sheehy (approved):
LGTM. Thanks for leaving it cleaner than you found it 👍🏻

xin-hedera (approved):

jascks:
Thank you Steven and Xin for all your helpful comments and suggestions!

Issue #5713, opened 2023-03-27
Add support for network prefix for resettable environments


mgoelswirlds:
### Problem

Currently an example of resettable environment is testnet. 
In order to support that with the following new structure in[ HIP-679](https://github.com/mgoelswirlds/hedera-improvement-proposal/blob/main/HIP/hip-679.md#specification) these environments need to be handled differently.

### Solution

- For networks that can be reset the network key in the path will start with the network name prefix followed by a timestamp, as stated in the above testnet example. On a reset, to get the latest network one would need to filter by the network prefix (e.g. `testnet`) then sort the output to get the latest network to read from.

steven-sheehy:
Depends on #4388

steven-sheehy:
Closing since HIP-679 won't be implemented in favor of block nodes.

Issue #4388, opened 2022-08-30
Support downloading stream files by node ID


steven-sheehy:
### Problem

Consensus nodes currently upload files to the bucket with paths that contain the node account ID (e.g. `/recordstreams/record0.0.3/2022-07-13T08_46_08.041986003Z.rcd`). The account ID for a node can potentially change while its node ID is immutable. The mirror node should support detecting and downloading from paths with node IDs or account IDs.
Reference to the [HIP](https://github.com/hashgraph/hedera-improvement-proposal/blob/06d6754e62bf804e9f32ab35000b2690cf496b39/HIP/hip-679.md)
### Solution

* Add a property `hedera.mirror.importer.downloader.pathType` with values `AUTO`, `ACCOUNT_ID`, and `NODE_ID` with a default of `ACCOUNT_ID`.
* Later once consensus nodes get closer to writing to the new path adjust default to `AUTO`.
* Add a property `hedera.mirror.importer.downloader.pathRefreshInterval=1m`.
* If `AUTO`:
  * Attempt to list signatures from a path with node account IDs.
  * If successful, update a per node expiration timestamp to ensure account ID is used until `pathRefreshInterval` elapses.
  * If no results, attempt to list signatures from a path with node IDs.
  * If successful, set `pathType=NODE_ID` so the node ID path is used going forward.

steven-sheehy:
Also relates to #4388 and we should probably do that one first.

mgoelswirlds:
Adding the following to this ticket:
- Except for mainnet , use the network name as the prefix .
-  Handle OTHER network separately, to have the prefix from the networkPrefix field.

jascks:
Taking over from Mugdha. Got a good download from her yesterday.

PR #6072, opened 2023-05-24
Remove networkPrefix property and rework network property usage.

### LINKED ISSUE #6024: Remove newly introduced networkPrefix property from MirrorProperties
Avoid introducing the new `networkPrefix` property and instead utilize the existing `network` property to achieve the same objective. See [PR comment](https://github.com/hashgraph/hedera-mirror-node/pull/5753#discussion_r1191694395).

- Remove `networkPrefix` from `MirrorProperties`.
- Also remove from `docs/configuration.md`.
- Update usage of `network` to achieve the same goal.

jascks:
**Description**:

Avoid publicly introducing the new `networkPrefix` property and instead utilize the existing `network` property to achieve the same objective. See #4388 [PR comment](https://github.com/hashgraph/hedera-mirror-node/pull/5753#discussion_r1191694395).

- Remove `networkPrefix` from `MirrorProperties`.
- Also remove from `docs/configuration.md`.
- Update usage of `network` within the codebase to achieve the same functionality.

**Related issue(s)**:

Fixes #6024

**Notes for reviewer**:
- While the network prefix is managed by this code, its actual usage and documentation is part of #5713 .

**Checklist**


hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/MirrorProperties.java
```diff
@@ -53,9 +55,10 @@ public class MirrorProperties {
     @NotNull
     private Map<String, MigrationProperties> migration = new CaseInsensitiveMap<>();
 
-    @NotNull
-    private HederaNetwork network = HederaNetwork.DEMO;
+    @NotBlank
+    private String network = HederaNetwork.DEMO;
 
+    @Setter(AccessLevel.NONE) // Not user set-able
     private String networkPrefix;
```
steven-sheehy, line 62: This is currently only used by tests. When we add resettable support, even then it would be called rarely when we detect a potential stream reset and may not be worth caching here and might be fine re-calculating in a `getNetworkPrefix()`. If we really think it's worth caching, can simplify:
```suggestion
    @Getter(lazy = true)
    private String networkPrefix = StringUtils.substringBefore(network, '-');
```

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/MirrorProperties.java
```diff
@@ -89,19 +89,39 @@ public enum ConsensusMode {
...
+        public static final String TESTNET = "testnet";
+
+        private static final Map<String, String> NETWORK_DEFAULT_BUCKETS = Map.of(
+                DEMO, "hedera-demo-streams",
+                MAINNET, "hedera-mainnet-streams",
+                // OTHER has no default bucket
+                PREVIEWNET, "hedera-preview-testnet-streams",
+                TESTNET, "hedera-testnet-streams-2023-01");
+
+        private static final Map<String, String> CANONICAL_REFERENCES = Map.of(
```
steven-sheehy, line 106: Can just use `NETWORK_DEFAULT_BUCKETS.keySet()` instead.

steven-sheehy (changes requested):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/MirrorProperties.java
```diff
@@ -70,17 +73,14 @@ public class MirrorProperties {
...
-                throw new InvalidConfigurationException(
-                        "Unable to retrieve the network prefix for network type " + network);
-            }
-            // Here (5713) we need to add logic to get the complete network prefix for resettable environments.
-            return network.toString().toLowerCase();
-        }
+    public void setNetwork(@NonNull String networkAndPrefix) {
+        var delimiterIndex = networkAndPrefix.indexOf(NETWORK_PREFIX_DELIMITER);
+        var networkName = delimiterIndex < 0 ? networkAndPrefix : networkAndPrefix.substring(0, delimiterIndex);
+        this.network = HederaNetwork.getCanonicalizedNetwork(networkName.toLowerCase());
```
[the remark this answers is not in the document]
jascks: A quick look in IDEA for result of `getNetwork()` being compared (`==` or `!=`) it looks like there are seven places in non-test code.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/MirrorProperties.java
jascks: I'm not sure how the resultant `Set<String>` of canonical network names helps me. I could call `contains()` perhaps, but is a `true` because the reference is equal or that the two Strings just have the same hash code? Java does intern String literals by default, so perhaps my worries are unfounded?

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/MirrorProperties.java
jascks: Taking a closer look at lombok lazy getter, it fires off only the first time it's called and then caches that result. Thus if `setNetwork()` is called again this getter will not reflect any changed network prefix. But in the normal course of reading properties from `application.yml` or elsewhere, this won't be an issue.

You're right though, #5713 is where any network prefix would be utilized. This PR just provided some basic String manipulation to accommodate and makes available any prefix specification; presently only utilized in the tests.

steven-sheehy: It's probably easiest to just compute on read via a `getNetworkPrefix()` and no field.

jascks: I concur. It's shaping up to be something like the following, using a single `network` field with a lombok generated setter.
```
    public String getNetwork() {
        return StringUtils.substringBefore(this.network, NETWORK_PREFIX_DELIMITER).toLowerCase();
    }

    public String getNetworkPrefix() {
        var networkPrefix = StringUtils.substringAfter(this.network, NETWORK_PREFIX_DELIMITER);
        return StringUtils.isEmpty(networkPrefix) ? null : networkPrefix;
    }
```
`StringUtils.substringAfter()` returns an empty string if the delimiter is not found. I prefer a `null` to that, or perhaps `getNetworkPrefix()` could return `Optional<String>`?

jascks: Done

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/MirrorProperties.java
jascks: Done. Removed canonicalization stuff, so this is no longer a thing.

steven-sheehy (approved):
LGTM

xin-hedera (approved):
LGTM

PR #6078, opened 2023-05-25
HIP-584: Add FeeDistribution

dikel:
**Description**:

Copied FeeDistribution from services for TransferLogic

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


hedera-mirror-web3/src/test/java/com/hedera/services/fees/charging/FeeDistributionTest.java
```diff
@@ -0,0 +1,127 @@
+package com.hedera.services.fees.charging;
+
```
steven-sheehy, line 2: Missing license header. Please ensure spotlessApply is ran automatically via git hook or run it manually if it's not. Ivan found that git-gui does not respect `core.hookspath`.

hedera-mirror-web3/src/main/java/com/hedera/services/context/properties/PropertySource.java
```diff
@@ -0,0 +1,45 @@
...
+ * limitations under the License.
+ */
+
+package com.hedera.services.context.properties;
+
+/**
+ * Defines a source of arbitrary properties keyed by strings. Provides strongly typed accessors for commonly used
+ * property types.
+ */
+public interface PropertySource {
```
steven-sheehy, line 23: I'm not sure how you're planning on using this in the future, but we cannot add a new way to load properties from disk even temporarily. This PR is fine but in any future PR that implements PropertySource might have an issue. We can use services interfaces but we would need a custom implementation that can load via Spring Boot techniques like environment variables, application.yaml, application.properties, etc.

hedera-mirror-web3/src/main/java/com/hedera/services/context/properties/GlobalDynamicProperties.java
```diff
@@ -0,0 +1,140 @@
...
+import static com.hedera.services.context.properties.PropertyNames.CONTRACTS_MAX_REFUND_PERCENT_OF_GAS_LIMIT;
+import static com.hedera.services.context.properties.PropertyNames.CONTRACTS_REDIRECT_TOKEN_CALLS;
+import static com.hedera.services.context.properties.PropertyNames.LAZY_CREATION_ENABLED;
+import static com.hedera.services.context.properties.PropertyNames.LEDGER_FUNDING_ACCOUNT;
+import static com.hedera.services.context.properties.PropertyNames.STAKING_FEES_NODE_REWARD_PERCENT;
+import static com.hedera.services.context.properties.PropertyNames.STAKING_FEES_STAKING_REWARD_PERCENT;
+import static com.hedera.services.context.properties.PropertyNames.STAKING_IS_ENABLED;
+
+@Singleton
+public class GlobalDynamicProperties implements EvmProperties {
```
IvanKavaldzhiev, line 45: We should use `MirrorNodeEvmProperties` instead. If needed properties are missing - add them there.

hedera-mirror-web3/src/main/java/com/hedera/services/context/properties/PropertySource.java
IvanKavaldzhiev: Once we delete GlobalDynamicProperties - we can delete also PropertySource. It's not needed in web3.

IvanKavaldzhiev (changes requested):
Generally, `FeeDistribution` is used for paying fees for staking. I don't think we need such a logic in the Archive Node. I suggest that we close the PR.

PR #6081, opened 2023-05-25
Add CORS origin * to POST /contracts/call

leric7:
**Description**:

Add CORS origin * support to `POST /contracts/call` endpoint.
Motivation: Making POST API call from browser errors out with CORS preflight check failed.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/ContractController.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/ContractController.java
@@ -44,6 +44,7 @@ import lombok.RequiredArgsConstructor;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.tuweni.bytes.Bytes;
 import org.hyperledger.besu.datatypes.Address;
+import org.springframework.web.bind.annotation.CrossOrigin;
 import org.springframework.web.bind.annotation.ExceptionHandler;
 import org.springframework.web.bind.annotation.PostMapping;
 import org.springframework.web.bind.annotation.RequestBody;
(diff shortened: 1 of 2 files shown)


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/ContractController.java
```diff
@@ -63,6 +64,7 @@ class ContractController {
     private final ContractCallService contractCallService;
     private final Bucket bucket;
 
+    @CrossOrigin(origins = "*")
```
steven-sheehy, line 67: This needs a test in `ContractControllerTest` that sets an `Origin: example.com` header and verifies the `Access-Control-Allow-Origin: *` in the response.

steven-sheehy (commented):
Thanks for your contribution. It seems our load balancer and/or our proxy in public environment automatically adds the `Access-Control-Allow-Origin: *` necessary so that calling `/api/v1/contracts/call` via our [Swagger](https://mainnet-public.mirrornode.hedera.com/api/v1/docs/#/contracts/contractsCall) page works fine. But if you run it locally it is missing this header so it's a good addition.

jascks (dismissed):
LGTM, the proposed test withstanding...

steven-sheehy (dismissed):
LGTM thanks!

steven-sheehy:
Looks like the test case fails. Can you check?

leric7:
@steven-sheehy , fixed the test case failing.

xin-hedera (approved):
LGTM

PR #6082, opened 2023-05-25
Add migration for synthetic transfer approvals

### LINKED ISSUE #5939: Fix incorrect approval flag for precompile token transfers
### Problem

The consensus nodes were [not](https://github.com/hashgraph/hedera-services/issues/6262) externalizing the `is_approval` flag when it automatically used an approval inside a `transferToken()` precompile.

### Solution

For a contract `0.0.X` with `X > 2119900` (grandfathered number), there are three ways that `0.0.X` can be authorized to debit assets from some account `0.0.A`:

1. `0.0.X` can be `0.0.A` --- you can always spend your own assets
2. `0.0.A` can have a `ThresholdKey{... Key{contractID=0.0.X}}` where the contract key suffices to meet the threshold
3. `0.0.A` can have granted `0.0.X` an approval via either HAPI or contract call

So if you're analyzing a `tokenTransferList` in a `CryptoTransfer` child record where the called contract id is `0.0.X > 2119900`, you can look at each fungible debit `(payer=0.0.A, -1)` or NFT ownership change `(sender=0.0.A, ...)` and use process of elimination with the 3 options above.

### Alternatives

_No response_

edwin-greene:
**Description**:
Add migration for synthetic transfers that are missing the correct approval flag. 

**Related issue(s)**:

Fixes #5939 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/SyntheticCryptoTransferApprovalMigration.java
+++ b/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/SyntheticCryptoTransferApprovalMigration.java
@@ -20,9 +20,11 @@ import com.google.common.base.Stopwatch;
 import com.hedera.mirror.common.domain.entity.EntityId;
 import com.hedera.mirror.importer.MirrorProperties;
 import com.hederahashgraph.api.proto.java.Key;
+import jakarta.inject.Named;
 import java.sql.ResultSet;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.concurrent.atomic.AtomicLong;
-import javax.inject.Named;
 import lombok.SneakyThrows;
 import org.flywaydb.core.api.MigrationVersion;
 import org.springframework.context.annotation.Lazy;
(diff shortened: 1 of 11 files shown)


hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/SyntheticCryptoTransferApprovalMigration.java
```diff
@@ -0,0 +1,215 @@
...
+
+@Named
+public class SyntheticCryptoTransferApprovalMigration extends RepeatableMigration {
+
+    private static final int BATCH_SIZE = 100;
+
+    private static final String TRANSFER_SQL =
+            """
+            with cryptotransfers as (
+              select ct.*, cr.sender_id, e.key
```
xin-hedera, line 40: the key has to be the one valid at the timestamp of the transfer

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/SyntheticCryptoTransferApprovalMigration.java
```diff
@@ -0,0 +1,215 @@
...
+        CRYPTO_TRANSFER,
+        NFT_TRANSFER,
+        TOKEN_TRANSFER
+    }
+
+    private final JdbcTemplate jdbcTemplate;
+
+    @Lazy
+    public SyntheticCryptoTransferApprovalMigration(
+            @Owner JdbcTemplate jdbcTemplate, MirrorProperties mirrorProperties) {
```
xin-hedera, line 133: don't need `@Owner` if we don't delete rows / change schema / etc

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/SyntheticCryptoTransferApprovalMigration.java
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    @Override
+    protected void doMigrate() {
+        AtomicLong count = new AtomicLong(0L);
+        var stopwatch = Stopwatch.createStarted();
+        jdbcTemplate.setFetchSize(BATCH_SIZE);
+        try {
+            jdbcTemplate.query(TRANSFER_SQL, rs -> {
+                if (processTransfer(rs)) {
```
xin-hedera, line 156: so if we fail to parse a single key, the whole db transaction is aborted?

we probably want to process as many rows as we can

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/SyntheticCryptoTransferApprovalMigration.java
```diff
@@ -0,0 +1,215 @@
...
+
+    @SneakyThrows
+    private boolean processTransfer(ResultSet rs) {
+        var senderId = rs.getLong("sender_id");
+        var keyBytes = rs.getBytes("key");
+        var parsedKey = getKey(keyBytes);
+        return !parsedKey.hasThresholdKey() || processThresholdKey(parsedKey, senderId);
+    }
+
+    private boolean processThresholdKey(Key parsedKey, long senderId) {
```
xin-hedera, line 204: I believe `processThresholdKey` should return `true` when none of the contract keys in the key list match `senderId`

and it's better to rename the function to `isContractIdInThresholdKey`

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/SyntheticCryptoTransferApprovalMigration.java
```diff
@@ -0,0 +1,215 @@
...
+                cast(null as bigint) as token_id,
+                'CRYPTO_TRANSFER' as transfer_type
+                from crypto_transfer
+              ) ct
+              join contract_result cr on cr.consensus_timestamp = ct.consensus_timestamp
+              join entity e on e.id = ct.entity_id
+              where
+                cr.sender_id > 2119900 and --grandfathered number
+                cr.sender_id <> ct.entity_id and --sender does not equal receiver
+                ct.is_approval = false
```
xin-hedera, line 59: and ct.amount < 0

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/SyntheticCryptoTransferApprovalMigration.java
```diff
@@ -0,0 +1,215 @@
...
+                  serial_number,
+                  token_id,
+                  'NFT_TRANSFER' as transfer_type
+                  from nft_transfer
+                ) nft
+                join contract_result cr on cr.consensus_timestamp = nft.consensus_timestamp
+                join entity e on e.id = nft.receiver_account_id
+                where
+                  cr.sender_id > 2119900 and
+                  cr.sender_id <> nft.receiver_account_id and
```
xin-hedera, line 79:
```suggestion
                  cr.sender_id <> nft.sender_account_id and
```

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/SyntheticCryptoTransferApprovalMigration.java
```diff
@@ -0,0 +1,215 @@
...
+                token_id,
+                'TOKEN_TRANSFER' as transfer_type
+                from token_transfer
+              ) t
+              join contract_result cr on cr.consensus_timestamp = t.consensus_timestamp
+              join entity e on e.id = t.account_id
+              where
+                cr.sender_id > 2119900 and
+                cr.sender_id <> t.account_id and
+                t.is_approval = false
```
xin-hedera, line 101: `and t.amount < 0`

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/SyntheticCryptoTransferApprovalMigration.java
xin-hedera, line 156: accidentally deleted my previous comment.

the db transaction will rollback if we fail to parse a single entity key, we should instead process as many rows as we can, and at the end have a single log to also indicate the number of rows with issues

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/SyntheticCryptoTransferApprovalMigration.java
```diff
@@ -0,0 +1,215 @@
...
+                'CRYPTO_TRANSFER' as transfer_type
+                from crypto_transfer
+              ) ct
+              join contract_result cr on cr.consensus_timestamp = ct.consensus_timestamp
+              join entity e on e.id = ct.entity_id
+              where
+                cr.sender_id > 2119900 and --grandfathered number
+                cr.sender_id <> ct.entity_id and --sender does not equal receiver
+                ct.is_approval = false
+              ), nfttransfers as (
```
xin-hedera, line 60: nft transfer handling is going to be more fun after it's nested into `transaction` table...

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/SyntheticCryptoTransferApprovalMigration.java
```diff
@@ -0,0 +1,215 @@
...
+
+    @Override
+    protected MigrationVersion getMinimumVersion() {
+        // The version where contract_result sender_id was added
+        return MigrationVersion.fromVersion("1.58.4");
+    }
+
+    @Override
+    protected void doMigrate() {
+        AtomicLong count = new AtomicLong(0L);
```
xin-hedera, line 151: another TBA is we need the timestamps of the first 0.38.x record file in both testnet & mainnet and only run migration for data before that

PR #6087, opened 2023-05-26
HIP-584: Integrate aliases map

### LINKED ISSUE #6052: HIP-584: Integrate aliases map into `StackedStateFrames`
### Problem

We need to provide an in-memory storage for newly created aliases, so that consequent nested transactions are aware of newly created accounts/aliases.

### Solution

Integrate a `Map<Address, Entity>` into `StackedStateFrames`, so that we reuse the stacked approach for storing aliases. We can use the existing `EntityDatabaseAccessor` and use if for upstream loading of data.

### Alternatives

_No response_

stoyan-lime:
**Description**:
Add an aliases map to handle aliases created during transactions.

**Related issue(s)**:

Fixes #6052


IvanKavaldzhiev (commented):
LG, since now `MirrorEvmContractAliases` would be initialized on each new request and it's not a singleton bean, the `aliases` map would be safe to use. Needs unit and integration test scenarios.

david-bakin-sl (commented):
OK for me (but agree with need for more tests as commented by Ivan)

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/account/MirrorEvmContractAliases.java
```diff
@@ -23,30 +23,32 @@
...
         }
 
-        final var entityOptional = mirrorEntityAccess.findEntity(addressOrAlias);
-
-        if (entityOptional.isEmpty()) {
-            throw new EntityNotFoundException("No such contract or token: " + addressOrAlias);
-        }
+        final var resolvedAddress =
+                Optional.ofNullable(aliases.get(addressOrAlias)).orElse(addressOrAlias);
+        final var entity = mirrorEntityAccess
```
steven-sheehy, line 48: Shouldn't it return the resolved alias from the map if present and skip db lookup in that scenario?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/account/MirrorEvmContractAliases.java
```diff
@@ -60,4 +62,12 @@ public Address resolveForEvm(Address addressOrAlias) {
             throw new InvalidParametersException("Not a contract or token: " + addressOrAlias);
         }
     }
+
+    public void link(final Address alias, final Address address) {
```
steven-sheehy, line 66: Who will end up calling link/unlink? Is there a ticket to track that?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/account/MirrorEvmContractAliases.java
```diff
@@ -23,30 +23,32 @@
...
             return addressOrAlias;
         }
 
-        final var entityOptional = mirrorEntityAccess.findEntity(addressOrAlias);
-
-        if (entityOptional.isEmpty()) {
-            throw new EntityNotFoundException("No such contract or token: " + addressOrAlias);
-        }
+        final var resolvedAddress =
+                Optional.ofNullable(aliases.get(addressOrAlias)).orElse(addressOrAlias);
```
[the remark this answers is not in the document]
stoyan-lime: Updated.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/account/MirrorEvmContractAliases.java
IvanKavaldzhiev: Yes, that would be the correct behaviour

stoyan-lime: Fixed.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/account/MirrorEvmContractAliasesTest.java
```diff
@@ -16,7 +16,7 @@
 
 package com.hedera.mirror.web3.evm.account;
 
-import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
+import static org.assertj.core.api.Assertions.assertThat;
```
david-bakin-sl, line 19: (Is there a way to train IntelliJ to _not_ offer certain suggestions for resolving symbols?  It's annoying to always have to pick `org.assertj.core.api.Assertions.assertThat` out of the list when you never ever really want the other ones (even though they are visible).  And sometimes mistakes are made... in fact, there's one on the very next line ....  
image
)

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/account/MirrorEvmContractAliases.java
```diff
@@ -23,30 +23,33 @@
...
 import lombok.RequiredArgsConstructor;
 import org.apache.tuweni.bytes.Bytes;
 import org.hyperledger.besu.datatypes.Address;
 
-@Named
 @RequiredArgsConstructor
 public class MirrorEvmContractAliases extends HederaEvmContractAliases {
     private final MirrorEntityAccess mirrorEntityAccess;
 
+    final Map<Address, Address> aliases = new HashMap<>();
```
david-bakin-sl, line 36: (IMO good use of "package private" access, often neglected.  No need to make this `private` - which people do mostly out of habit I think) and thus make testing (as in  `MirrorEvmContractAliasesTest.resolveForEvmFromAliases`) difficult.)

david-bakin-sl (approved):
I believe some test renaming is desirable but not required.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/account/MirrorEvmContractAliases.java
```diff
@@ -60,4 +63,12 @@ public Address resolveForEvm(Address addressOrAlias) {
             throw new InvalidParametersException("Not a contract or token: " + addressOrAlias);
         }
     }
+
+    public void link(final Address alias, final Address address) {
+        aliases.put(alias, address);
```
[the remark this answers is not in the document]
stoyan-lime: The behavior in services is to replace the value, so I keep it the same here.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/account/MirrorEvmContractAliases.java
stoyan-lime: I did make it private at first, but had difficulty testing it.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/account/MirrorEvmContractAliasesTest.java
```diff
@@ -69,6 +69,15 @@ void resolveForEvmSuccess() {
         assertThat(result).isEqualTo(ADDRESS2);
     }
 
+    @Test
+    void resolveForEvmFromAliases() {
```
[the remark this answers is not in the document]
stoyan-lime: Updated test names.

steven-sheehy (approved):
LGTM

xin-hedera (approved):
LGTM

PR #6115, opened 2023-05-30
Handle null hapiVersion in blocks API

### LINKED ISSUE #6069: hapi_version is null.null.null on blocks API
### Description

Blocks API for older record files that did not have `hapi_version` populated in the protobuf show `"hapi_version":"null.null.null"` instead of `"hapi_version": null`.

### Steps to reproduce

https://mainnet-public.mirrornode.hedera.com/api/v1/blocks?limit=1&order=asc

### Additional context

_No response_

### Hedera network

mainnet

### Version

0.80.0

### Operating system

None

MarcKriguerAtHedera:
**Description**:

* Fix `recordFile.getFullVersion` to return `null` instead of `null.null.null`
* New spec file to test that it works

**Related issue(s)**:

Fixes #6069 

**Checklist**

- [x] Tested (unit)

### DIFF
--- a/hedera-mirror-rest/model/recordFile.js
+++ b/hedera-mirror-rest/model/recordFile.js
@@ -62,7 +62,8 @@ class RecordFile {
   }
 
   getFullHapiVersion() {
-    return `${this.hapiVersionMajor}.${this.hapiVersionMinor}.${this.hapiVersionPatch}`;
+    return `${this.hapiVersionMajor}` == `null` ? null
+        : `${this.hapiVersionMajor}.${this.hapiVersionMinor}.${this.hapiVersionPatch}`;
   }
 }
 
--- a/hedera-mirror-rest/model/recordFile.js
+++ b/hedera-mirror-rest/model/recordFile.js
@@ -62,7 +62,7 @@ class RecordFile {
   }
 
   getFullHapiVersion() {
-    return `${this.hapiVersionMajor}` == `null` ? null
+    return hapiVersionMajor === null || hapiVersionMinor === null || hapiVersionPatch === null ? null
         : `${this.hapiVersionMajor}.${this.hapiVersionMinor}.${this.hapiVersionPatch}`;
   }
 }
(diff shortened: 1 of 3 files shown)


hedera-mirror-rest/model/recordFile.js
```diff
@@ -62,7 +62,8 @@ class RecordFile {
   }
 
   getFullHapiVersion() {
-    return `${this.hapiVersionMajor}.${this.hapiVersionMinor}.${this.hapiVersionPatch}`;
+    return `${this.hapiVersionMajor}` == `null` ? null
```
steven-sheehy, line 65:
```suggestion
    return hapiVersionMajor === null || hapiVersionMinor === null || hapiVersionPatch === null ? null
```

steven-sheehy (changes requested):

hedera-mirror-rest/model/recordFile.js
MarcKriguerAtHedera: Done/

MarcKriguerAtHedera: Redone (the backticks and $ and braces are still gone, but needed to restore the "`this`." for the code to work).

hedera-mirror-rest/model/recordFile.js
```diff
@@ -62,7 +62,8 @@ class RecordFile {
   }
 
   getFullHapiVersion() {
-    return `${this.hapiVersionMajor}.${this.hapiVersionMinor}.${this.hapiVersionPatch}`;
+    return (this.hapiVersionMajor === null || this.hapiVersionMinor === null || this.hapiVersionPatch === null) ? null
```
xin-hedera, line 65: add `nullable: true` to open api spec for `hapi_version`

xin-hedera (changes requested):

hedera-mirror-rest/model/recordFile.js
MarcKriguerAtHedera: Done.

steven-sheehy (approved):

PR #6123, opened 2023-06-01
HIP-584: Implement account deletion in Updater

### LINKED ISSUE #6119: HIP-584: Implement account deletion in Updater
### Problem

We need to update the deleted accounts in the parent `HederaEvmStackedWorldStateUpdater` in the `commit` function.

### Solution

1. Override the `deleteAccount` method and provide implementation.
2. Add the deleted accounts to the parent Updater in the `commit`.

### Alternatives

_No response_

0xivanov:
**Description**:

We need to update the deleted accounts in the parent `HederaEvmStackedWorldStateUpdater` in the `commit` function.

**Related issue(s)**:

Fixes #6119

**Notes for reviewer**:

**Checklist**


david-bakin-sl:
I was 5 files into the review here and so far had only seen whitespace/formatting changes.  Can this be broken into TWO PRs - one to fix formatting and one with the actual interesting changes?

0xivanov:
> I was 5 files into the review here and so far had only seen whitespace/formatting changes. Can this be broken into TWO PRs - one to fix formatting and one with the actual interesting changes?

I think that the `spotlessApply` gradle action, that runs automatically when committing is making these changes, which is a bummer.

0xivanov:
> > I think that the spotlessApply gradle action, that runs automatically when committing is making these changes, which is a bummer.
> 
> It does, but the only reason this is a problem is because some people were committing with a git tool that didn't respect our commit hooks so it bypassed that spotless apply in some PRs. Now that you're touching the same files that were previously missed it's updating those files. If we're sure the everyone on the limechain team now is using the git hook, we can consider doing another mass format PR to fix these up again.

Actually when I am updating from `main/origin` the `spotlessApply` is formatting the changed/updated files from that commit. Maybe the gui I am using is the problem.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/AbstractLedgerWorldUpdater.java
```diff
@@ -0,0 +1,156 @@
...
+import org.hyperledger.besu.datatypes.Wei;
+import org.hyperledger.besu.evm.account.Account;
+import org.hyperledger.besu.evm.account.EvmAccount;
+import org.hyperledger.besu.evm.worldstate.WorldView;
+import org.hyperledger.besu.evm.worldstate.WrappedEvmAccount;
+
+public abstract class AbstractLedgerWorldUpdater<W extends WorldView, A extends Account>
+        extends AbstractLedgerEvmWorldUpdater<W, A> {
+
+    private final StackedStateFrames<Object> stackedStateFrames;
```
IvanKavaldzhiev, line 41: Use `Store`

steven-sheehy (approved):

PR #6153, opened 2023-06-06
Cache and utilize StreamFilename path and incorporate node ID based path access in LocalStreamFileProvider

### LINKED ISSUE #5759: Store/Cache pathName derived for NodeId based path
### Problem

Currently we are creating the bucket download path based on the `pathType` enum for download and then recreating it to write back to. We should be able to cache this path to avoid having to recreate it for every node. 

### Solution

- In `StreamFilename` add a string path to store the path while downloading and then writing it back.
- Modify `LocalStreamFileProvider` to read from the new path where consensus node writes it.
- Change signature and stream file to write back to disk using the same structure, converting paths appropriately.

### Alternatives

_No response_

### LINKED ISSUE #6079: Enable LocalStreamFileProvider to access stream files from new node ID based bucket structure
### Problem

`LocalStreamFileProvider` requires some of the capabilities added to `S3StreamFileProvider` by #4388 in support of HIP-679. Namely it requires the ability to list and get stream files from the local file system using either the legacy node account ID and now the new node ID based bucket structures.

Also, implement a simplified auto-mode, since all stream files are in the local file systems and costly S3 access is not involved. 

### Solution

- Refer to `commonDownloaderProperties.getPathType()` to determine bucket structure mode configured.
- Copy test stream files to file system directory hierarchy to reflect bucket structure.
- For list and get operations, access file system hierarchy according to `PathType` (`getDirectory()`). With #5759 in place, get should utilize cached path.
- Simplified `AUTO` mode is just checking the `ACCOUNT_ID` directory, and if no files a present, go to the `NODE_ID` directory. No need for a refresh interval or map to store state etc. as in `S3StreamFileProvider`.
- Implement test cases to test `NODE_ID` mode.

### Alternatives

_No response_

jascks:
**Description**:

Currently we are creating the bucket download path based on the `PathType` enum for download and then recreating it to write back to. We should be able to cache this path to avoid having to recreate it for every node. Also, `LocalStreamFileProvider` requires the ability to list and get stream files from the local file system using either the legacy node account ID and now the new node ID based bucket structures.

* In `StreamFilename` add a string path to store the path while downloading and then writing it back.
* Modify `LocalStreamFileProvider` to read from the new path where consensus node writes it.
* Change signature and stream file to write back to disk using the same structure, converting paths appropriately.
* For `LocalStreamFileProvder` `list()` operation, access file system hierarchy according to configured `PathType`.
* In addition, implement a simplified `AUTO` mode by just checking the `ACCOUNT_ID` directory, and if no files a present, go to the `NODE_ID` directory. No need for a refresh interval or map to store state etc. as in `S3StreamFileProvider`.

**Related issue(s)**:

Fixes #5759 
Fixes #6079

**Notes for reviewer**:
Note that I am still looking for getting more coverage using account ID, node ID and AUTO testing. So, I will be working on this a bit further during review.

**Checklist**

### DIFF
--- a/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/StreamFileData.java
+++ b/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/StreamFileData.java
@@ -56,8 +56,7 @@ public class StreamFileData {
         try {
             byte[] bytes = FileUtils.readFileToByteArray(file);
             var lastModified = Instant.ofEpochMilli(file.lastModified());
-            // TODO invoke new constructor with name and path
-            return new StreamFileData(new StreamFilename(file.getName()), bytes, lastModified);
+            return new StreamFileData(new StreamFilename(file.getPath()), bytes, lastModified);
         } catch (InvalidStreamFileException e) {
             throw e;
         } catch (Exception e) {
--- a/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/StreamFileData.java
+++ b/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/StreamFileData.java
@@ -23,8 +23,8 @@ import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.Path;
 import java.time.Instant;
-import java.util.function.Supplier;
 import lombok.CustomLog;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
@@ -53,11 +53,11 @@ public class StreamFileData {
 
     private final Instant lastModified;
 
-    private static StreamFileData readStreamFileData(File file, Supplier<StreamFilename> streamFilenameSupplier) {
+    private static StreamFileData readStreamFileData(File file, StreamFilename streamFilename) {
         try {
             byte[] bytes = FileUtils.readFileToByteArray(file);
             var lastModified = Instant.ofEpochMilli(file.lastModified());
-            return new StreamFileData(streamFilenameSupplier.get(), bytes, lastModified);
+            return new StreamFileData(streamFilename, bytes, lastModified);
         } catch (InvalidStreamFileException e) {
             throw e;
         } catch (Exception e) {
@@ -66,11 +66,12 @@ public class StreamFileData {
     }
 
     public static StreamFileData from(@NonNull File file) {
-        return readStreamFileData(file, () -> StreamFilename.from(file.getPath()));
+        return readStreamFileData(file, StreamFilename.from(file.getPath()));
     }
 
-    public static StreamFileData from(@NonNull StreamFilename streamFilename) {
-        return readStreamFileData(new File(streamFilename.getFilePath()), () -> streamFilename);
+    public static StreamFileData from(@NonNull Path basePath, @NonNull StreamFilename streamFilename) {
+        var streamFile = new File(basePath.toFile(), streamFilename.getFilePath());
+        return readStreamFileData(streamFile, streamFilename);
     }
 
     // Used for testing String based files like CSVs
--- a/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/StreamFileData.java
+++ b/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/StreamFileData.java
@@ -66,7 +66,7 @@ public class StreamFileData {
     }
 
     public static StreamFileData from(@NonNull File file) {
-        return readStreamFileData(file, () -> StreamFilename.of(file.getPath()));
+        return readStreamFileData(file, () -> StreamFilename.from(file.getPath()));
     }
 
     public static StreamFileData from(@NonNull StreamFilename streamFilename) {
@@ -76,12 +76,12 @@ public class StreamFileData {
     // Used for testing String based files like CSVs
     public static StreamFileData from(@NonNull String filename, @NonNull String contents) {
         return new StreamFileData(
-                StreamFilename.of(filename), contents.getBytes(StandardCharsets.UTF_8), Instant.now());
+                StreamFilename.from(filename), contents.getBytes(StandardCharsets.UTF_8), Instant.now());
     }
 
     // Used for testing with raw bytes
     public static StreamFileData from(@NonNull String filename, byte[] bytes) {
-        return new StreamFileData(StreamFilename.of(filename), bytes, Instant.now());
+        return new StreamFileData(StreamFilename.from(filename), bytes, Instant.now());
     }
 
     public InputStream getInputStream() {
--- a/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/StreamFileData.java
+++ b/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/StreamFileData.java
@@ -24,6 +24,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
 import java.time.Instant;
+import java.util.function.Supplier;
 import lombok.CustomLog;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
@@ -52,11 +53,11 @@ public class StreamFileData {
 
     private final Instant lastModified;
 
-    public static StreamFileData from(@NonNull File file) {
+    private static StreamFileData readStreamFileData(File file, Supplier<StreamFilename> streamFilenameSupplier) {
         try {
             byte[] bytes = FileUtils.readFileToByteArray(file);
             var lastModified = Instant.ofEpochMilli(file.lastModified());
-            return new StreamFileData(new StreamFilename(file.getPath()), bytes, lastModified);
+            return new StreamFileData(streamFilenameSupplier.get(), bytes, lastModified);
         } catch (InvalidStreamFileException e) {
             throw e;
         } catch (Exception e) {
@@ -64,15 +65,23 @@ public class StreamFileData {
         }
     }
 
+    public static StreamFileData from(@NonNull File file) {
+        return readStreamFileData(file, () -> StreamFilename.of(file.getPath()));
+    }
+
+    public static StreamFileData from(@NonNull StreamFilename streamFilename) {
+        return readStreamFileData(new File(streamFilename.getFilePath()), () -> streamFilename);
+    }
+
     // Used for testing String based files like CSVs
     public static StreamFileData from(@NonNull String filename, @NonNull String contents) {
         return new StreamFileData(
-                new StreamFilename(filename), contents.getBytes(StandardCharsets.UTF_8), Instant.now());
+                StreamFilename.of(filename), contents.getBytes(StandardCharsets.UTF_8), Instant.now());
     }
 
     // Used for testing with raw bytes
     public static StreamFileData from(@NonNull String filename, byte[] bytes) {
-        return new StreamFileData(new StreamFilename(filename), bytes, Instant.now());
+        return new StreamFileData(StreamFilename.of(filename), bytes, Instant.now());
     }
 
     public InputStream getInputStream() {
--- a/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/StreamFileData.java
+++ b/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/StreamFileData.java
@@ -56,6 +56,7 @@ public class StreamFileData {
         try {
             byte[] bytes = FileUtils.readFileToByteArray(file);
             var lastModified = Instant.ofEpochMilli(file.lastModified());
+            // TODO invoke new constructor with name and path
             return new StreamFileData(new StreamFilename(file.getName()), bytes, lastModified);
         } catch (InvalidStreamFileException e) {
             throw e;
--- a/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/StreamFileData.java
+++ b/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/StreamFileData.java
@@ -93,6 +93,10 @@ public class StreamFileData {
         return streamFilename.getFilename();
     }
 
+    public String getFilePath() {
+        return streamFilename.getFilePath();
+    }
+
     @Override
     public String toString() {
         return streamFilename.toString();
--- a/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/StreamFileData.java
+++ b/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/StreamFileData.java
@@ -66,7 +66,7 @@ public class StreamFileData {
     }
 
     public static StreamFileData from(@NonNull File file) {
-        return readStreamFileData(file, StreamFilename.from(file.getPath()));
+        return readStreamFileData(file, StreamFilename.from(file.getPath(), File.separator));
     }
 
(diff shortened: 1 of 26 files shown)


xin-hedera (changes requested):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/StreamFilename.java
```diff
@@ -98,17 +108,42 @@ public StreamFilename(String filename) {
...
+        if (this.fileType == SIDECAR) {
+            builder.append(SIDECAR_FOLDER);
+            builder.append(this.pathSeparator);
+        }
+        builder.append(this.filename);
+        this.filePath = builder.toString();
+    }
+
+    public static StreamFilename from(String filePath) {
+        return from(filePath, File.separator);
```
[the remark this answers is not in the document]
jascks: This particular method is used in numerous unit tests as well, which are referring to the local file system, either stuff in the tests `data` directory or those copied via `FileCopier`. There is a lot outside of `S3StreamFileProvider` where the platform separator is appropriate.

`S3StreamFileProvider.SEPARATOR` is used explicitly when building the `ListObjectsV2Request` as well as within the `StreamFilename` instance within each `StreamFileData` returned by `list()`. In that case as well, the file prefix is derived using only string manipulation and not via the likes of `Path.of()`, `Path.resolve()` etc.

The goal is for the S3 based `StreamFilename` instance to be used as the basis of the subsequent retrieval of the data files, including sidecars, and then utilizing that bucket relative path to archive the files in the appropriate structure in the local file system. It seems I need to look into that more closely along with your `Utility.java` comment.

I don't know if any of this code has ever been run on Windows. I'm trying to keep it in consideration. :)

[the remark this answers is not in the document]
xin-hedera: I don't think we should have changes with potential bugs because that's the preferred default in test cases.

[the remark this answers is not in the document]
jascks: Okay. Just to be clear you want to see:
```
    public static StreamFilename from(String filePath) {
        return from(filePath, S3StreamFileProvider.SEPARATOR);
    }
```
And then where a different path separator character is required, to provide it via one of the other `from()` factory methods.

jnels124 (approved):
LGTM

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/StreamFilename.java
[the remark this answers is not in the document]
xin-hedera: yup that is it

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/util/Utility.java
```diff
@@ -117,8 +118,8 @@ public static String printProtoMessage(GeneratedMessageV3 message) {
         return TextFormat.shortDebugString(message);
     }
 
-    public static void archiveFile(String filename, byte[] contents, Path destinationRoot) {
-        Path destination = destinationRoot.resolve(filename);
+    public static void archiveFile(StreamFilename streamFilename, byte[] contents, Path destinationRoot) {
+        Path destination = destinationRoot.resolve(streamFilename.getFilePath());
```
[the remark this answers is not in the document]
jascks: Done

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/StreamFilename.java
```diff
@@ -98,17 +108,42 @@ public StreamFilename(String filename) {
...
-    /**
-     * Gets the filename with the specified streamType, fileType, and instant.
-     *
-     * @param streamType
-     * @param fileType
-     * @param instant
-     * @return the filename
-     */
     public static String getFilename(StreamType streamType, FileType fileType, Instant instant) {
+
```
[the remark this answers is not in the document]
jascks: Done

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/util/Utility.java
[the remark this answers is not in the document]
jascks: Had to do the same for `StreamFileSignature`.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/StreamFilename.java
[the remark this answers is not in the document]
jascks: Okay, made the above change, and tests run since not on Windows. Looking at test code as to where to add `File.separator` as parameter.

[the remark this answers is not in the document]
xin-hedera: perhaps leave it as is, I don't think we need to support windows.

xin-hedera (approved):
LGTM

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/FileCopier.java
```diff
@@ -102,4 +108,47 @@ public void copy() {
...
+            var destination = destinationAdjuster.apply(to);
+            var networkDir = destination.resolve(network);
+            var sourceNodeDirs = FileUtils.listFilesAndDirs(from.toFile(), TrueFileFilter.INSTANCE, null);
+
+            for (var sourceNodeDir : sourceNodeDirs) {
+
+                var sourceNodeDirName = sourceNodeDir.getName();
+                var streamTypeOpt = Arrays.stream(StreamType.values())
+                        .filter(st -> sourceNodeDirName.startsWith(st.getNodePrefix()))
+                        .findFirst();
```
edwin-greene, line 133: Not sure how many times this will be invoked, but wondering if a regex would be faster than a stream/filter and if a slight performance gain would be meaningful.

edwin-greene (approved):
Looks good

mgoelswirlds (dismissed):
The bucket restructure looks so clean and good now. I am glad that the path modularization is done here. 
LGTM

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/FileCopier.java
xin-hedera: only used in test cases, so unless it's awfully slow, not worth the effort to optimize

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/StreamFileSignature.java
```diff
@@ -75,7 +75,11 @@ public StreamFilename getDataFilename() {
             dataFilename += COMPRESSED_EXTENSION;
         }
 
-        return new StreamFilename(dataFilename);
+        return StreamFilename.from(filename, dataFilename);
+    }
+
+    public String getFilePath() {
```
steven-sheehy, line 81: Method seems as an unnecessary wrapper around the StreamFilename.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/provider/LocalStreamFileProvider.java
```diff
@@ -42,37 +47,63 @@ public class LocalStreamFileProvider implements StreamFileProvider {
 
     @Override
     public Mono<StreamFileData> get(ConsensusNode node, StreamFilename streamFilename) {
-        return Mono.fromSupplier(() -> getDirectory(node, streamFilename))
-                .map(dir -> dir.toPath().resolve(streamFilename.getFilename()).toFile())
-                .map(StreamFileData::from)
+        var basePath =
+                commonDownloaderProperties.getMirrorProperties().getDataPath().resolve(STREAMS);
```
steven-sheehy, line 51: We should not re-resolve the base path against the filename. The full path should already be populated by the list() and directly used in the get()

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/provider/LocalStreamFileProvider.java
```diff
@@ -42,37 +47,63 @@ public class LocalStreamFileProvider implements StreamFileProvider {
...
+                        return Flux.fromArray(dir.listFiles(f -> matches(startAfter, f)));
+                    }
+                    return Flux.empty();
+                }))
                 .sort()
                 .take(batchSize)
-                .map(StreamFileData::from)
-                .doOnSubscribe(s -> log.debug("Searching for the next {} files after {}", batchSize, lastFilename));
+                .map(file -> StreamFilename.from(prefixPathRef.get().toString(), file.getName(), File.separator))
+                .flatMapSequential(streamFilename -> get(node, streamFilename))
```
steven-sheehy, line 88: Why do we need to call get() instead of just doing the previous simpler `.map(StreamFileData::from)`?

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/provider/LocalStreamFileProvider.java
```diff
@@ -42,37 +47,63 @@ public class LocalStreamFileProvider implements StreamFileProvider {
...
+                        var dir = getDirectory(basePath, prefixPathRef.get(), lastFilename);
+                        return Flux.fromArray(dir.listFiles(f -> matches(startAfter, f)));
+                    }
+                    return Flux.empty();
+                }))
                 .sort()
                 .take(batchSize)
-                .map(StreamFileData::from)
-                .doOnSubscribe(s -> log.debug("Searching for the next {} files after {}", batchSize, lastFilename));
+                .map(file -> StreamFilename.from(prefixPathRef.get().toString(), file.getName(), File.separator))
```
steven-sheehy, line 87: This should not need to use prefixPathRef. The `file` should be the full absolute path of the file to return. Thus we can remove `prefixPathRef` entirely.

steven-sheehy (changes requested):
Still reviewing.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/provider/LocalStreamFileProvider.java
jascks: This path in `StreamFilename` is bucket relative in the case of `S3StreamFileProvider` and relative to a data directory in `LocalStreamFileProvider` so that it can be 're-rooted' when the file is written for archive.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/StreamFileSignature.java
jascks: No real reason I guess. Perhaps I pushed things a too little far [here](https://github.com/hashgraph/hedera-mirror-node/pull/6153#discussion_r1229986563).

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/DownloaderProperties.java
```diff
@@ -29,12 +29,10 @@ public interface DownloaderProperties {
 
     MirrorProperties getMirrorProperties();
 
-    default Path getNodeStreamPath(long nodeId) {
-        return getStreamPath().resolve(getStreamType().getNodePrefix() + nodeId);
+    default Path getStreamPath() {
+        return getMirrorProperties().getDataPath();
```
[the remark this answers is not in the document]
jascks: Done.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/StreamFileSignature.java
jascks: Removed.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/StreamFileData.java
```diff
@@ -52,27 +53,36 @@ public class StreamFileData {
...
+            return new StreamFileData(streamFilename, bytes, lastModified);
         } catch (InvalidStreamFileException e) {
             throw e;
         } catch (Exception e) {
             throw new FileOperationException("Unable to read file to byte array", e);
         }
     }
 
+    public static StreamFileData from(@NonNull File file) {
+        return readStreamFileData(file, StreamFilename.from(file.getPath(), File.separator));
```
xin-hedera, line 69: why do we use `File.separator` here?

jascks: This was in response to one of Steven's comments. I felt it was good to use here because we are working with a `File` here. I guess we need to draw a line here if this is a problem. @steven-sheehy, will we never support Windows? I was trying to remain agnostic where it made sense to me.

xin-hedera: I see. didn't notice it's a static method only used in test cases.

we used to have windows support and dropped it later because not widely used afaict.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/provider/LocalStreamFileProvider.java
jascks: I guess it felt natural in my mind, similar to `S3StreamFileProvider` implementation. But, it is not necessary to do that. So, I put it back. Well, at least as far as working relative paths.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/provider/LocalStreamFileProvider.java
jascks: It seems the remaining issue centers around this. My intention is for the path added to `StreamFilename` to be relative to something, not an absolute path in itself. The full bucket structure, relative to the bucket name, is provided, in either legacy account ID or the new HIP-679 node ID format.

The importer cloud storage configuration can be such that the downloader is sourcing stream files from either GCS or S3 via `S3StreamFileProvider` or the local file system using `LocalStreamFileProvider`. Regardless of their source, the downloader may also archive signature and data files to the local file system.

For a given `StreamFilename`, if it was sourced from an S3 bucket (via `S3Proxy` or real life), or if was read from the local file system (a source data path, test directory etc.) the relative path reflecting the 'bucket' structure is the same. And then when that file is archived to a different location in the local file system, that same relative path is used, though relative to a different directory in the file system.

If `S3StreamFileProvider` provides a relative path and `LocalStreamFileProvider` supplies an absolute path, then at archive time this will need to be detected and processed in some manner to yield the same destination for writing the files.

steven-sheehy (approved):
LGTM

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/provider/LocalStreamFileProvider.java
steven-sheehy: I think we could've stored the absolute path and for archiving removed the well known prefix. The archive operation is disabled by default and no one probably uses it while the local filesystem provider will soon become important for historical synchronization, so makes sense to optimize for the latter. But we can keep as is for now.

xin-hedera (approved):
LGTM

PR #6154, opened 2023-06-06
HIP-584: Enable CREATE2 operation by saving new alias to alias map

### LINKED ISSUE #6077: HIP-584: Return correct address for CREATE2 in `HederaEvmStackedWorldStateUpdater`
### Problem

We need to implement `HederaEvmStackedWorldStateUpdater.newAliasedContractAddress()` so that we return correct aliased address during CREATE2 deploy.

### Solution

Implement `HederaEvmStackedWorldStateUpdater.newAliasedContractAddress()`

### Alternatives

_No response_

IvanKavaldzhiev:
**Description**:

This PR:

- Enables `CREATE2` operation by saving then new alias to the in-memory map cache with aliases
- Introduce `pendingChanges` map which will keep each new stacked `Updater` changes. In case of a problem, the `Updater` would revert and clear the `pendingChanges` map, while keeping a safe copy of map with aliases, accumulated from previous child transactions

**Related issue(s)**:

Fixes #6077

**Notes for reviewer**:

**Checklist**


steven-sheehy (approved):

0xivanov (approved):
LGTM

PR #6155, opened 2023-06-06
HIP-584: Provide implementation for transfer precompiles - Part1

mustafauzunn:
**Description**:

We need default implementations for precompiles related to transfer, so we should copy needed logic from hedera-services into hedera-mirror-web3.

*`run(MessageFrame frame)` will be implemented in the next PR

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


hedera-mirror-web3/src/main/java/com/hedera/services/store/contracts/precompile/AbiConstants.java
```diff
@@ -0,0 +1,41 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hedera.services.store.contracts.precompile;
+
+/** All ABI constants used by {@link Precompile} implementations, in one place for easy review. */
+public class AbiConstants {
```
IvanKavaldzhiev, line 20: All `AbiConstants` would be present after you pull from `main`.

steven-sheehy (approved):

hedera-mirror-web3/src/main/java/com/hedera/services/store/contracts/precompile/impl/TransferPrecompile.java
```diff
@@ -0,0 +1,348 @@
...
+    private int numLazyCreates;
+
+    public TransferPrecompile(PrecompilePricingUtils pricingUtils, int functionId, boolean isLazyCreationEnabled) {
+        super(pricingUtils);
+        this.functionId = functionId;
+        this.isLazyCreationEnabled = isLazyCreationEnabled;
+    }
+
+    @Override
+    public void body(Bytes input, UnaryOperator<byte[]> aliasResolver) {
```
IvanKavaldzhiev, line 102: Consider passing the `functionId` as an argument here, wrapped in a specific record for transfer arguments. Soon I will open PR for a precompile with a bean approach that can serve as an example.

xin-hedera (approved):
LGTM

PR #6161, opened 2023-06-07
HIP-584: Alias improvements

### LINKED ISSUE #6052: HIP-584: Integrate aliases map into `StackedStateFrames`

stoyan-lime:
**Description**:
Clean up alias resolution and keep already resolved values in-memory.

**Related issue(s)**:

Fixes #6052


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/account/MirrorEvmContractAliases.java
```diff
@@ -49,18 +48,14 @@ public Address resolveForEvm(Address addressOrAlias) {
...
                 .orElseThrow(() -> new EntityNotFoundException("No such contract or token: " + addressOrAlias));
 
-        final var entityId = entity.toEntityId();
-
-        if (entity.getType() == EntityType.TOKEN) {
-            final var bytes = Bytes.wrap(toEvmAddress(entityId));
-            return Address.wrap(bytes);
-        } else if (entity.getType() == EntityType.CONTRACT) {
-            final var bytes =
-                    Bytes.wrap(entity.getEvmAddress() != null ? entity.getEvmAddress() : toEvmAddress(entityId));
```
steven-sheehy, line 59: Why are we no longer using the evmAddress if available and always encoding the account ID to an address?

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/account/MirrorEvmContractAliases.java
IvanKavaldzhiev: This method should always return the long-zero/mirror address and not the alias one. Currently for contracts we may have an alias address if the create is made via a top-level ethereum transaction and then I think the evm_address would hold the alias format, while we want the regular long-zero address.

steven-sheehy (approved):
LGTM, but need to merge first

steven-sheehy (approved):

PR #6168, opened 2023-06-08
HIP-584: Copy Associate logic

### LINKED ISSUE #5907: HIP-584: Copy `AssociateLogic` into `hedera-mirror-web3`
### Problem

We need to provide `AssociateLogic` implementation in mirror-node, so that we can implement `AssociatePrecompile`.

### Solution

Copy `AssociateLogic` from mono-services by substituting `Store` classes with `StackedStateFrames`. Enhance `Account` model with `associateWith` method and all of the needed validations, except checking `TransactionBody`.

### Alternatives

_No response_

stoyan-lime:
**Description**:
Copy associate logic from services.
Unlike in services associateWith method is implemented in the AssociateLogic class instead of Account model.

Added modification builder to keep account immutability, if it is considered a good approach we can implement it in other places in following PR.

Add `Store` concept that abstracts the `StackedStateFrames` internals and introduces convenient methods for state interaction.

**Related issue(s)**:

Fixes #5907


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
```diff
@@ -88,6 +88,12 @@ public class MirrorNodeEvmProperties implements EvmProperties {
     @DurationMin(seconds = 100)
     private Duration rateLimit = Duration.ofSeconds(100L);
 
+    @Getter
+    private boolean tokenAssociationsLimited = true;
+
+    @Getter
+    private int maxTokensPerAccount = 10_000_000;
```
IvanKavaldzhiev, line 95: Sort alphabetically and update `docs/configuration.md`

IvanKavaldzhiev (dismissed):
Very clean approach! LGTM 🥂

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/Account.java
```diff
@@ -52,6 +53,7 @@ public class Account extends HederaEvmAccount {
     private final int numTreasuryTitles;
     private final long ethereumNonce;
 
+    @Builder(toBuilder = true)
```
steven-sheehy, line 56: We should not use mirror node only libraries in copied services code. Let's keep it exactly as services had it since it will soon be replaced.

hedera-mirror-web3/src/main/java/com/hedera/services/txn/token/AssociateLogic.java
```diff
@@ -0,0 +1,123 @@
...
+import static com.hedera.node.app.service.evm.utils.ValidationUtils.validateFalse;
+import static com.hederahashgraph.api.proto.java.ResponseCodeEnum.FAIL_INVALID;
+import static com.hederahashgraph.api.proto.java.ResponseCodeEnum.TOKENS_PER_ACCOUNT_LIMIT_EXCEEDED;
+import static com.hederahashgraph.api.proto.java.ResponseCodeEnum.TOKEN_ALREADY_ASSOCIATED_TO_ACCOUNT;
+
+import com.hedera.mirror.web3.evm.properties.MirrorNodeEvmProperties;
+import com.hedera.mirror.web3.evm.store.CachingStateFrame.Accessor;
+import com.hedera.mirror.web3.evm.store.StackedStateFrames;
+import com.hedera.mirror.web3.evm.store.accessor.model.TokenRelationshipKey;
+import com.hedera.mirror.web3.exception.InvalidTransactionException;
```
steven-sheehy, line 28: We should not be importing web3 classes into services code. If the stacked state is intended to be used in services code then it should do so upstream and a new version of the library produced.

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/Account.java
stoyan-lime: In services Account is mutable and has setter methods. 
Instead of setters we have methods that copy the whole model with 1 property change.
This is just a cleaner way to do that, as we add a new method for each new property we need to change.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/TokenRelationshipDatabaseAccessor.java
```diff
@@ -34,27 +34,29 @@
...
-                                false,
-                                false,
-                                Boolean.TRUE == tokenAccount.getAutomaticAssociation(),
-                                0))));
+    public @NonNull Optional<TokenRelationship> get(@NonNull Object key) {
+        final var tokenRelationshipKey = (TokenRelationshipKey) key;
+        return findAccount(tokenRelationshipKey.accountAddress())
+                .flatMap(account -> findToken(tokenRelationshipKey.tokenAddress())
+                        .flatMap(token -> findTokenAccount(token, account)
+                                .filter(t -> t.getAssociated() == Boolean.TRUE)
```
david-bakin-sl, line 49: Can't you just use a method reference here?  (like `Token::getAssociated()`?)

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/TokenRelationshipDatabaseAccessor.java
```diff
@@ -34,27 +34,29 @@
...
+                                .map(tokenAccount -> new TokenRelationship(
+                                        token,
+                                        account,
+                                        tokenAccount.getBalance(),
+                                        TokenFreezeStatusEnum.FROZEN == tokenAccount.getFreezeStatus(),
```
david-bakin-sl, line 54: Constructor has 5 booleans in a row ... sad.  Probably too late to change now since these are "copied" from services, right?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/HederaEvmWorldState.java
```diff
@@ -61,16 +61,16 @@ public HederaEvmWorldState(
...
         this.evmProperties = evmProperties;
         this.abstractCodeCache = abstractCodeCache;
         this.accountAccessor = accountAccessor;
         this.tokenAccessor = tokenAccessor;
         this.mirrorEvmContractAliases = mirrorEvmContractAliases;
         this.entityAddressSequencer = entityAddressSequencer;
-        this.stackedStateFrames = stackedStateFrames;
-        stackedStateFrames.push();
+        this.store = store;
+        this.store.wrap();
```
david-bakin-sl, line 73: What's the difference (in purpose) between the `wrap()` here and the wrap [over here, above](https://github.com/hashgraph/hedera-mirror-node/pull/6168/files#diff-6f5c154402b6f237d5560c6746ae8377ad75b9d69f34ffbef75fb95634158fedR69)?  IOW, why are there two different places where you need to `wrap()`?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/HederaEvmWorldState.java
```diff
@@ -166,11 +166,7 @@ public Account getForMutation(final Address address) {
 
         @Override
         public void commit() {
-            final var topFrame = stackedStateFrames.top();
-            if (stackedStateFrames.height() > 1) { // commit only to upstream RWCachingStateFrame
-                topFrame.commit();
-                stackedStateFrames.pop();
-            }
+            store.commit();
```
david-bakin-sl, line 169: And same question here for `commit`: What's the difference between the one here and the [one here, above](https://github.com/hashgraph/hedera-mirror-node/pull/6168/files#diff-fc0f01c4aa5719cefe1ebdcd828ffbf69c62de089ff9700984ee2affb89a75f9R59)?  Why does it need to be called in two separate places?

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/Account.java
```diff
@@ -24,24 +24,32 @@
 import com.hedera.services.utils.EntityNum;
 import java.util.SortedMap;
 import java.util.SortedSet;
+import lombok.Builder;
 import org.apache.commons.lang3.builder.EqualsBuilder;
 import org.apache.commons.lang3.builder.HashCodeBuilder;
 import org.hyperledger.besu.datatypes.Address;
 
 /**
+ * Copied Account model from hedera-services.
```
david-bakin-sl, line 33: Definitely good to have this command and the `differences with original` comment with it, for each of these things copied from services!

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/Account.java
```diff
@@ -24,24 +24,32 @@
...
  *
  * This model is used as a value in a special state (CachingStateFrame), used for speculative write operations. Object
  * immutability is required for this model in order to be used seamlessly in the state.
+ *
+ * Differences with the original:
+ *     1. Missing fields like memo, key, isReceiverSigRequired, isSmartContract
```
david-bakin-sl, line 39: Maybe "removed" instead of "missing"?  Describes the action taken, e.g., like "added" in point 2 below.

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/FcTokenAllowanceId.java
```diff
@@ -24,17 +24,19 @@
...
  * It consists of the information about the token for which allowance is granted to and the spender who is granted the
  * allowance.
  *
  * <p>Having allowance on a token will grant the spender to transfer fungible or non-fungible token
  * units from the owner's account.
+ *
+ * Differences with the original:
+ *  1. Deleted unnecessary fields
  */
 public class FcTokenAllowanceId implements Comparable<FcTokenAllowanceId> {
```
david-bakin-sl, line 39: So ... what does the prefix `Fc` stand for anyway?

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/NftId.java
```diff
@@ -20,6 +20,11 @@
 import edu.umd.cs.findbugs.annotations.NonNull;
 import java.util.Comparator;
 
+/**
+ * Copied Id type from hedera-services.
```
david-bakin-sl, line 24: `Id` -> `NftId`

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/OwnershipTracker.java
```diff
@@ -16,13 +16,27 @@
...
+ * Copied type from hedera-services.
+ *
  * Encapsulates the changes of {@link UniqueToken} ownership within the context of one Transaction
  */
 public class OwnershipTracker {
-    private Map<Id, List<Change>> changes = new HashMap<>();
+    private final Map<Id, List<Change>> changes = new HashMap<>();
+
+    public static Change forMinting(final Id treasury, final long serialNumber) {
+        return new Change(Id.DEFAULT, treasury, serialNumber);
```
david-bakin-sl, line 34: `Change` has an error-prone API (IMO) but it's too late now.

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/Token.java
```diff
@@ -88,6 +90,41 @@ public class Token {
     private final long lastUsedSerialNumber;
     private final List<CustomFee> customFees;
 
+    public Token(Id id) {
+        this(
+                id,
+                Collections.emptyList(),
+                Collections.emptyList(),
```
david-bakin-sl, line 97: is the code coverage bot actually warning me that `Collections.emptyList()` is not covered by our tests?  For this it interrupts my reading flow?  Can't it be a bit smarter about it?

image

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/Token.java
```diff
@@ -88,6 +90,41 @@ public class Token {
...
+                null,
+                0,
+                0,
+                0,
+                Collections.emptyList());
+    }
+
     @SuppressWarnings("java:S107")
     public Token(
             Id id,
```
david-bakin-sl, line 130: This constructor takes
* 2 semantically different lists of `UniqueTokens`, consecutively
* 2 semantically different count values, consecutively
* 7, count'em: **7!!** semantically different _keys_ consecutively - their different uses are rather important, no?
* 2 semantically different `Account`s, consecutively
* now 3 semantically different boolean values, consecutively,
* 3 semantically different strings, consecutively,
* 3 semantically different integral values, consecutively
* and some other longs and booleans stuck in there for fun.

That vibration you're feeling in the seat of your pants is the vibration caused in your location by my teeth grinding over here.

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/Token.java
```diff
@@ -154,6 +191,75 @@ public Token(
...
+                0,
+                Collections.emptyList());
+    }
+
+    // copied from TokenTypesManager in services
+    private static TokenType mapToDomain(final com.hederahashgraph.api.proto.java.TokenType grpcType) {
+        if (grpcType == com.hederahashgraph.api.proto.java.TokenType.NON_FUNGIBLE_UNIQUE) {
+            return TokenType.NON_FUNGIBLE_UNIQUE;
+        }
+        return TokenType.FUNGIBLE_COMMON;
```
david-bakin-sl, line 260: would an explicit `else` help readability?

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/TokenRelationship.java
```diff
@@ -77,6 +79,19 @@ public TokenRelationship(
         this.balanceChange = balanceChange;
     }
 
+    public TokenRelationship(Token token, Account account) {
+        this(
+                token,
+                account,
+                0,
+                token.isFrozenByDefault() && token.hasFreezeKey(),
+                !token.hasKycKey(),
```
david-bakin-sl, line 88: Not clear to me why `!token.hasKycKey()` means "kycGranted" - is that just the known semantics of KYC keys, that if you've got one then it isn't the case that `kycGranted` and if you _don't_ have one then it _is_ the case that `kycGranted`?

hedera-mirror-web3/src/main/java/com/hedera/services/txn/token/AssociateLogic.java
```diff
@@ -0,0 +1,104 @@
...
+    private final MirrorNodeEvmProperties mirrorNodeEvmProperties;
+
+    public AssociateLogic(final StoreImpl store, final MirrorNodeEvmProperties mirrorNodeEvmProperties) {
+        this.store = store;
+        this.mirrorNodeEvmProperties = mirrorNodeEvmProperties;
+    }
+
+    public void associate(final Address accountAddress, final List<Address> tokensAddresses) {
+        /* Load the models */
+        final var account = store.getAccount(accountAddress, true);
```
david-bakin-sl, line 54: See, wouldn't this line and line 56 below be easier to understand if they read `... store.getAccount(accountAddress, OnMissing.THROW_EXCEPTION)` ??

david-bakin-sl (approved):
(I have to come back and look at the unit tests better.)

Lots of suggestions, but as is my practice: approved.  If you make these changes, great!  If you make different changes inspired by my comments, also great! If you choose not to accept a comment, great too, but in that case please leave a counter-comment explaining why so that I can learn.

david-bakin-sl (dismissed):
[**DUPLICATE comment because Github was having issues**]

(I have to come back and look at the unit tests better.)

Lots of suggestions, but as is my practice: approved.  If you make these changes, great!  If you make different changes inspired by my comments, also great! If you choose not to accept a comment, great too, but in that case please leave a counter-comment explaining why so that I can learn.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/TokenRelationshipDatabaseAccessor.java
IvanKavaldzhiev: If we agree on that, I can use the `toBuilder` Lombok generator in all models, to get rid of almost all constructors and just update the properties we need to modify. In this way the chance of mistyping an argument in the long constructor is minimised. These changes can be documented in the javadoc.

david-bakin-sl: I think it is worth doing and would make things clearer - but @steven-sheehy may have other higher-priority ideas about it.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/StoreImpl.java
```diff
@@ -0,0 +1,134 @@
...
+
+    @Override
+    public void addPendingChanges() {
+        stackedStateFrames.top().commit();
+    }
+
+    @Override
+    public void commit() {
+        final var topFrame = stackedStateFrames.top();
+        if (stackedStateFrames.height() > 1) { // commit only to upstream RWCachingStateFrame
```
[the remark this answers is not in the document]
IvanKavaldzhiev: You're right. Internally we check whether we already have an upstream frame.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/HederaEvmWorldState.java
IvanKavaldzhiev: Since the `HederaEvmWorldState` should introduce the first layer of `RWCachingStateFrame`, and then the `StackedUpdater` which is linked to a specific transaction execution (and consequently a message frame) needs also to introduce its own `RWCachingStateFrame`. Having a StackedUpdater basically means we are inside an internal child transaction and we need separate state layer for it, on top of the base main `RWCachingStateFrame` of the top level transaction.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/HederaEvmWorldState.java
IvanKavaldzhiev: This `commit` handles changes from the top level main transaction.
The `commit` from the `StackedStateUpdater` handles changes from an internal child transaction.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/StoreImpl.java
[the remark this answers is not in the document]
0xivanov: We check if there is a `upstreamFrame` , but if this frame happens to be `DatabaseBackedStateFrame` will throw `UnsupportedOperationException` . Maybe it is a good idea to keep this check (or do it internally in the stacked frames).

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/TokenRelationship.java
IvanKavaldzhiev: Yes, it seems this is the logic we have in services.

docs/configuration.md
```diff
@@ -576,8 +576,10 @@ value, it is recommended to only populate overridden properties in the custom `a
...
  `hedera.mirror.web3.evm.trace.enabled`                 | false                                              | Flag enabling tracer                                                                                                                                                                          
  `hedera.mirror.web3.evm.trace.contract`                | []                                                 | A set with contract addresses to filter. By default it is empty to indicate it will trace all contract addresses.                                                                             
  `hedera.mirror.web3.evm.trace.status`                  | []                                                 | A set with frame statuses to filter. By default it is empty to indicate it will trace all frames regardless of status.                                                                        
+ `hedera.mirror.web3.evm.tokenAssociationsLimited`      | true                                               | Limits the number of token associations per account when true
```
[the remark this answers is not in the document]
IvanKavaldzhiev: Fixed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/Store.java
```diff
@@ -0,0 +1,46 @@
...
+
+import com.hedera.mirror.web3.evm.store.accessor.model.TokenRelationshipKey;
+import com.hedera.services.store.models.Account;
+import com.hedera.services.store.models.NftId;
+import com.hedera.services.store.models.Token;
+import com.hedera.services.store.models.TokenRelationship;
+import com.hedera.services.store.models.UniqueToken;
+import org.hyperledger.besu.datatypes.Address;
+
+public interface Store {
```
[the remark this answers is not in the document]
IvanKavaldzhiev: 1. Added javadoc. I decided to remove the `pop()` operation and thus `addPendingChanges()`, so currently we have only `commit()`. I think poping made the whole logic more complicated, while currently I don't see the benefits. The main idea was each new nested call to start with a new `RWCachingStateFrame`, but currently having one `RWCachingStateFrame` for multiple nested transactions seems to also do the trick. There might be some corner cases with some intermediate reverts but I propose to start simple and add additional logic if needed and to test it with more complicated scenarios.
2. I added it. The reason was this functionality is not used in this PR. The interface would grow and capture more methods in the future.
3. Renamed
4. Added the Enum. It made the API clearer, indeed.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/StoreImpl.java
```diff
@@ -0,0 +1,134 @@
...
+        this.stackedStateFrames = new StackedStateFrames<>(databaseAccessors);
+    }
+
+    @Override
+    public Account getAccount(final Address address, final boolean throwIfMissing) {
+        final var accountAccessor = stackedStateFrames.top().getAccessor(Account.class);
+        final var account = accountAccessor.get(address);
+
+        if (throwIfMissing) {
+            return account.orElseThrow(() -> missingEntityException(Account.class.getName(), address));
```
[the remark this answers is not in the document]
IvanKavaldzhiev: 1. Added factory methods.
2. I think using this approach might make things more complicated.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/StoreImpl.java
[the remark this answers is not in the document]
IvanKavaldzhiev: That's correct. I left the check and it's good to be outside of the `commit` operation, since the `commit` is linked to the `CachingStateFrame` instance, which does not know the exact stack height.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/StoreImpl.java
```diff
@@ -0,0 +1,134 @@
...
+        stackedStateFrames.push();
+    }
+
+    private TokenRelationshipKey keyFromRelationship(TokenRelationship tokenRelationship) {
+        final var tokenAddress = tokenRelationship.getToken().getId().asEvmAddress();
+        final var accountAddress = tokenRelationship.getAccount().getAccountAddress();
+        return new TokenRelationshipKey(tokenAddress, accountAddress);
+    }
+
+    private InvalidTransactionException missingEntityException(final String type, Object id) {
```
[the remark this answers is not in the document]
IvanKavaldzhiev: Done

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/TokenRelationshipDatabaseAccessor.java
IvanKavaldzhiev: Done

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/HederaEvmWorldState.java
IvanKavaldzhiev: I decided to leave just only this `wrap()` and have a common `RWCachingState` frame for all internal child transactions. Currently this approach works and simplifies the overall logic.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/HederaEvmWorldState.java
IvanKavaldzhiev: Having only one `wrap()` method, I decided to reduce also the number of commits to the one in the `StackedUpdater`. I removed the commit from the `HederaEvmWorldState`, since there we are at top level and this means we are ending the transaction execution and we don't want to commit anything anymore. Whereas, in services, we definitely want to commit the result of the top-level transaction to the state.

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/NftId.java
IvanKavaldzhiev: Fixed

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/Token.java
IvanKavaldzhiev: Added

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
```diff
@@ -74,6 +74,9 @@ public class MirrorNodeEvmProperties implements EvmProperties {
     @Max(100)
     private int maxGasRefundPercentage = 100;
 
+    @Getter
+    private int maxTokensPerAccount = 1000;
```
steven-sheehy, line 78:
```suggestion
    @Min(1)
    private int maxTokensPerAccount = 1000;
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/CachingStateFrame.java
```diff
@@ -25,6 +25,7 @@
 import java.util.Optional;
 import java.util.stream.Collectors;
 import lombok.NonNull;
+import org.checkerframework.checker.units.qual.K;
```
steven-sheehy, line 28: I don't think you meant to import this and actually wanted a generic K.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/TokenRelationshipDatabaseAccessor.java
steven-sheehy: Again, the guiding principle we use is that we should minimize changes to these copied files. That way if upstream makes a change we can just grab the latest changes. So definitely we should not use Lombok in services code. If you'd like to improve copied code, improve it upstream first.

docs/configuration.md
```diff
@@ -576,8 +576,10 @@ value, it is recommended to only populate overridden properties in the custom `a
...
+ `hedera.mirror.web3.evm.maxTokensPerAccount`           | 1000                                               | Maximum number token associations per account                                                                                                                                                 
  `hedera.mirror.web3.evm.network`                       | TESTNET                                            | Which Hedera network to use. Can be either `MAINNET`, `PREVIEWNET`, `TESTNET` or `OTHER`                                                                                                      
  `hedera.mirror.web3.evm.rateLimit`                     | 100s                                               | Maximum RPS limit                                                                                                                                                                             
+ `hedera.mirror.web3.evm.tokenAssociationsLimited`      | true                                               | Limits the number of token associations per account when true
```
steven-sheehy, line 582: Is this property necessary? Since we allow `Long.MAX_VALUE` for `maxTokensPerAccount ` won't that one property be sufficient?

hedera-mirror-web3/src/main/java/com/hedera/services/txn/token/AssociateLogic.java
```diff
@@ -0,0 +1,101 @@
...
+ *  2. Use abstraction for the state by introducing {@link Store} interface
+ *  3. Use Mirror Node specific properties - {@link MirrorNodeEvmProperties}
+ *  4. Use copied models from hedera-services which are enhanced with additional constructors and/or lombok generated builder for easier setup,
+ *  those are {@link Account}, {@link Token}, {@link TokenRelationship}
+ * */
+public class AssociateLogic {
+    private final MirrorNodeEvmProperties mirrorNodeEvmProperties;
+
+    public AssociateLogic(final MirrorNodeEvmProperties mirrorNodeEvmProperties) {
+        this.mirrorNodeEvmProperties = mirrorNodeEvmProperties;
```
steven-sheehy, line 47: This dependency could be fixed by adding a `EvmProperties.maxTokensPerAccount()` upstream.

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/services/txn/token/AssociateLogic.java
IvanKavaldzhiev: Do you mean adding it in `hedera-smart-contract-service` in services repo?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/CachingStateFrame.java
IvanKavaldzhiev: Correct, that import is a mistake

docs/configuration.md
IvanKavaldzhiev: I removed the property, as well as the Lombok usage in the Account model.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
IvanKavaldzhiev: Fixed

hedera-mirror-web3/src/main/java/com/hedera/services/txn/token/AssociateLogic.java
steven-sheehy: No, EvmProperties class which is in hedera-evm. Seems like a tiny change, but can consider later.

steven-sheehy (approved):

PR #6187, opened 2023-06-12
Add Synthetic Contract Result for HTS operations via HAPI

### LINKED ISSUE #5909: Extend synthetic contract log logic to populate contract result table
### Problem

We’ve recently added support for creating synthetic events for single crypto transfer, token mint, token burn, token wipe and approvals. With we are now logging events for those operations similar to Ethereum. 
Problem is that The Graph still can’t fetch those events, because of it’s flow and the way we log those events.

### Solution

We need to populate contract result table, similar to the way we create and populate synthetic event (contract) logs. 

Extend the current synthetic contract log implementation to add contract results.

### Alternatives

_No response_

georgi-l95:
**Description**:
This PR extends the synthetic event logs when operations on HTS tokens are performed via HAPI.
Add new class SyntheticContractResult which implementation takes care of the logging. 

Some informations is not logged in the contract result, as it's not possible to get it, but it's enough for the needed use cases.

**Related issue(s)**:

Fixes #5909 

**Notes for reviewer**:

**Checklist**


hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/entity/EntityRecordItemListener.java
```diff
@@ -469,6 +472,11 @@ private void insertTokenTransfers(RecordItem recordItem) {
                 insertNonFungibleTokenTransfers(
                         recordItem, tokenId, entityTokenId, payerAccountId, tokenTransferList.getNftTransfersList());
             }
+
+            if (i == 0) {
+                syntheticContractResultService.create(
+                        new TransferContractResult(recordItem, entityTokenId, payerAccountId));
```
steven-sheehy, line 478: The function parameters are not really accurate in regards to ERC20 `transfer(address recipient, uint256 amount)`, right? They only include the signature but not the address or amount in the parameters. Is it not necessary to be 100% accurate since otherwise we'd have to have a 1:1 result to log?

steven-sheehy (changes requested):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/entity/EntityRecordItemListener.java
georgi-l95: Yes, you're correct, only the signature is correct. However, if we want to populate the other two fields, we'd need to get the necessary info from the `tokenTransferList` after they are done processing, because there can be only one contract result with the same consensus timestamp. So I've left it like this in favor to not burden the methods with additional logic for fetching this info or passing it along until they're completed. Would love to hear you're opinion. Just for more context, for the main use case - The Graph, this field is not so important. The Graph only cares for this contract result to exist and this transaction to have correct transaction hash, which later can be queried and logs can be fetched from the contract logs table.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/contractresult/SyntheticContractResultServiceImpl.java
```diff
@@ -0,0 +1,61 @@
...
+    public void create(SyntheticContractResult result) {
+        if (isContract(result.getRecordItem()) || !entityProperties.getPersist().isSyntheticContractResults()) {
+            return;
+        }
+
+        long consensusTimestamp = result.getRecordItem().getConsensusTimestamp();
+        long gasLimit = result.getRecordItem().getTransactionBody().getTransactionFee();
+
+        ContractResult contractResult = new ContractResult();
+
```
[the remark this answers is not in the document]
georgi-l95: I've possible yes, i think we should populated it, but I didn't find the correct info for it, so I've left it blank, for now. For context this field is not needed for our main use case - the graph, but i 100% think that we should strive to populate as many fields as we can with the correct info, regardless of are they used now by some tools or not.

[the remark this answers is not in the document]
steven-sheehy: Don't both transfer and approve just return a boolean? So you can just return `recordItem.isSuccessful()` after encoding it to hex. Which I think would always be true in these scenarios anyway.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/contractresult/SyntheticContractResultServiceImpl.java
```diff
@@ -0,0 +1,61 @@
...
+        contractResult.setConsensusTimestamp(consensusTimestamp);
+        contractResult.setContractId(result.getEntityId().getId());
```
steven-sheehy, line 45: It's confusing since this isn't a contract ID, but a token ID. This will cause the REST API to have `/api/v1/contracts/{tokenId}/results/*` populated with this data. Hashscan or others could attempt to then lookup `{tokenId}` via `/api/v1/contracts/{tokenId}` and it won't return anything as it's not a contract and doesn't have bytecode.

Should we instead put the precompile address here since that is the "contract" being invoked? Of course, grouping them on precompile addresses is not ideal either and will cause an index imbalance. Will The Graph/relay will specifically lookup via `/api/v1/contracts/{tokenId}/results/*`?

georgi-l95: Do you mean 0x167 address? We don't invoke precompiles here, those are pure HAPI operations via the SDK. 
No, the graph or relay will only lookup `/api/v1/contracts/results` and filter by timestamp. Then from the response it gets the all `hash`es of those contract results and search one by one from another endpoint. Basiclly the graph only cares that they exist and have correct hash, nothing else.

steven-sheehy: Xin pointed out we already have contract results for token precompiles, so no concern.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/contractresult/AbstractSyntheticContractResult.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+package com.hedera.mirror.importer.parser.contractresult;
+
+import com.hedera.mirror.common.domain.entity.EntityId;
+import com.hedera.mirror.common.domain.transaction.RecordItem;
+import lombok.Data;
+import org.apache.tuweni.bytes.Bytes;
+
+@Data
```
xin-hedera, line 24: add `@RequiredArgsConstructor` and remove the constructor

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/contractresult/AbstractSyntheticContractResult.java
```diff
@@ -0,0 +1,46 @@
...
+            RecordItem recordItem, EntityId entityId, EntityId senderId, byte[] functionParameters) {
+        this.recordItem = recordItem;
+        this.entityId = entityId;
+        this.senderId = senderId;
+        this.functionParameters = functionParameters;
+    }
+
+    static final String TRANSFER_SIGNATURE = "a9059cbb";
+
+    static final String APPROVE_SIGNATURE = "095ea7b3";
```
xin-hedera, line 41: change these to `byte[]` and use `hexToBytes` to convert from the hex string. use the static byte array in concrete classes to avoid every time creates a new byte array

xin-hedera (changes requested):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/contractresult/AbstractSyntheticContractResult.java
georgi-l95: Added.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/contractresult/AbstractSyntheticContractResult.java
georgi-l95: Changed.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/contractresult/SyntheticContractResultServiceImpl.java
```diff
@@ -0,0 +1,61 @@
...
+        contractResult.setSenderId(result.getSenderId());
+        contractResult.setFunctionParameters(result.getFunctionParameters());
+        contractResult.setGasLimit(gasLimit);
+        contractResult.setTransactionResult(0);
```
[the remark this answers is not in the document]
georgi-l95: Done.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/contractresult/SyntheticContractResultServiceImpl.java
```diff
@@ -0,0 +1,61 @@
...
+    private final EntityListener entityListener;
+    private final EntityProperties entityProperties;
+
+    @Override
+    public void create(SyntheticContractResult result) {
+        if (isContract(result.getRecordItem()) || !entityProperties.getPersist().isSyntheticContractResults()) {
+            return;
+        }
+
+        long consensusTimestamp = result.getRecordItem().getConsensusTimestamp();
```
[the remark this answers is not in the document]
georgi-l95: Done.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/contractresult/SyntheticContractResultServiceImpl.java
```diff
@@ -0,0 +1,61 @@
...
+        contractResult.setConsensusTimestamp(consensusTimestamp);
+        contractResult.setContractId(result.getEntityId().getId());
+        contractResult.setSenderId(result.getSenderId());
+        contractResult.setFunctionParameters(result.getFunctionParameters());
+        contractResult.setGasLimit(gasLimit);
+        contractResult.setTransactionResult(0);
+        contractResult.setPayerAccountId(result.getRecordItem().getPayerAccountId());
+        contractResult.setTransactionHash(result.getRecordItem().getTransactionHash());
+        contractResult.setTransactionIndex(result.getRecordItem().getTransactionIndex());
+
```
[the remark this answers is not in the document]
georgi-l95: Done.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/contractresult/SyntheticContractResultServiceImpl.java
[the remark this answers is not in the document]
georgi-l95: Added.

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/record/transactionhandler/AbstractTransactionHandlerTest.java
```diff
@@ -148,6 +149,9 @@ abstract class AbstractTransactionHandlerTest {
     @Mock
     protected SyntheticContractLogService syntheticContractLogService;
 
+    @Mock
+    protected SyntheticContractResultService syntheticContractResultService;
```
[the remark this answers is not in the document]
georgi-l95: Done.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/contractresult/SyntheticContractResultServiceImpl.java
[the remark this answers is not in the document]
georgi-l95: Done.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/entity/EntityRecordItemListener.java
```diff
@@ -463,9 +466,27 @@ private void insertTokenTransfers(RecordItem recordItem, Transaction transaction
...
+        List<TokenTransferList> tokenTransferListsList =
+                recordItem.getTransactionRecord().getTokenTransferListsList();
+        for (int i = 0; i < tokenTransferListsList.size(); i++) {
+            TokenTransferList tokenTransferList = tokenTransferListsList.get(i);
+
+            if (tokenTransferList.getTransfersCount() > 0) {
+                insertFungibleTokenTransfers(recordItem, tokenTransferList);
+            }
+
+            if (tokenTransferList.getNftTransfersCount() > 0) {
```
steven-sheehy, line 478: Remove these two new `if` checks since the guarded methods already have the same check internally.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/contractresult/SyntheticContractResultServiceImpl.java
```diff
@@ -0,0 +1,70 @@
...
+        contractResult.setConsensusTimestamp(consensusTimestamp);
+        contractResult.setContractId(result.getEntityId().getId());
+        contractResult.setSenderId(result.getSenderId());
+        contractResult.setFunctionParameters(result.getFunctionParameters());
+        contractResult.setGasLimit(gasLimit);
+        contractResult.setTransactionResult(transactionRecord.getReceipt().getStatusValue());
+        contractResult.setPayerAccountId(recordItem.getPayerAccountId());
+        contractResult.setTransactionHash(recordItem.getTransactionHash());
+        contractResult.setTransactionIndex(recordItem.getTransactionIndex());
+        contractResult.setTransactionNonce(transactionBody.getTransactionID().getNonce());
```
steven-sheehy, line 61: nit: sort setters alphabetically and remove `gasLimit` variable

steven-sheehy (changes requested):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/contractresult/SyntheticContractResultServiceImpl.java
georgi-l95: Sorted and removed.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/entity/EntityRecordItemListener.java
georgi-l95: Removed.

steven-sheehy (approved):
LGTM

xin-hedera (dismissed):

steven-sheehy:
merge conflict

xin-hedera (approved):

steven-sheehy (approved):

PR #6201, opened 2023-06-14
Adding tests to check nft_transfers in transaction list API.

### LINKED ISSUE #6011: Add NFT transfers to transactions list API
### Problem

### Solution

After #6008 is implemented:

* Change `/api/v1/transactions` to add info from `transaction.nft_transfer` in a new `nft_transfers` field.
* Fix the API to always show `token_transfers` even if empty

### Alternatives

_No response_

mgoelswirlds:
**Description**:

Adding NFT transfers to transactions list API

This PR modifies
* Transaction list API to read nft_transfers from transactions table 
* return token_transfers as empty array even if there are none

**Related issue(s)**:

Fixes #6011 

**Checklist**


hedera-mirror-rest/__tests__/specs/accounts/{id}/no-balances.json
```diff
@@ -120,12 +120,14 @@
         "max_fee": "33",
         "memo_base64": null,
         "name": "CRYPTOUPDATEACCOUNT",
+        "nft_transfers": [],
```
mgoelswirlds, line 123: Had to add these , since i have mandated returning empty arrays when token_transfers and nft_transfers are not present.

hedera-mirror-rest/api/v1/openapi.yml
```diff
@@ -3337,6 +3337,23 @@ components:
           type: string
         name:
           $ref: "#/components/schemas/TransactionTypes"
+        nft_transfers:
```
mgoelswirlds, line 3340: OpenApi did not have nft transfers, so addign it now.

PR #6210, opened 2023-06-15
Change /transactions/{id} to use nested NFT transfers

### LINKED ISSUE #6009: Change transaction by ID API to use nested NFT transfers
### Problem

### Solution

After #6008 is implemented:

* Change `/api/v1/transactions/:id` to use `transaction.nft_transfer`
* Fix the API to always show `nft_transfers` and `token_transfers` even if empty

### Alternatives

_No response_

MarcKriguerAtHedera:
**Description**:
This PR modifies the `getTransactionQuery` method of **transactions.js** to get the `nft_transger_list` directly from the nested `nft_transfer` column of the transaction table, rather than querying the now-obsolete `nft_transfer` table.

* Change `createNftTransferList()` to a return an empty array (and not `undefined` on empty input.
* Change `createTokenTransferList()` the same way.

**Related issue(s)**:

Fixes #6009 

**Notes for reviewer**:

**Checklist**


xin-hedera:
you need to branch off https://github.com/hashgraph/hedera-mirror-node/tree/4099-nested-nft-transfers since the PR will merge to that branch

PR #6212, opened 2023-06-15
Change /transactions/{id} to use nested NFT transfers

### LINKED ISSUE #6009: Change transaction by ID API to use nested NFT transfers

MarcKriguerAtHedera:
**Description**:
Description:
This PR modifies the `getTransactionQuery` method of `transactions.js` to get the **nft_transfer** list directly from the nested `nft_transfer` column of the `transaction` table, rather than querying the now-obsolete `nft_transfer` table.

* Change `createNftTransferList()` to a return an empty array (instead of **undefined**) on empty input.
* Change `createTokenTransferList()` the same way.

**Related issue(s)**:

Fixes #6009 

**Notes for reviewer**:

**Checklist**


MarcKriguerAtHedera:
There is currently a rest test failing in the parent branch (@mgoelswirlds is running into it as well) that is not impacted by the files changed in this PR.

hedera-mirror-rest/transactions.js
```diff
@@ -715,9 +715,8 @@ const getTransactionQuery = (mainCondition, subQueryCondition) => {
       where ${TokenTransfer.CONSENSUS_TIMESTAMP} = t.consensus_timestamp and ${subQueryCondition}
     ) as token_transfer_list,
     (
-      select ${nftTransferJsonAgg}
-      from ${NftTransfer.tableName} ${NftTransfer.tableAlias}
-      where ${NftTransfer.CONSENSUS_TIMESTAMP} = t.consensus_timestamp and ${subQueryCondition}
+      select ${Transaction.NFT_TRANSFER} from ${Transaction.tableName}
+      where ${subQueryCondition}
     ) as nft_transfer_list,
```
xin-hedera, line 720: the subquery should be removed completely, and replace by just the new jsonb column `${Transaction.NFT_TRANSFER}`

xin-hedera (changes requested):

hedera-mirror-rest/transactions.js
MarcKriguerAtHedera: I originally did this, but now did need to restore a `jsonb_agg` call (and so I don't know if we can remove the "from transaction" from that).

xin-hedera: I don't think that's the type of fix we want

MarcKriguerAtHedera: All the "nonce" tests were failing (and I had to set the number of jest workers to 1 just to see the error message) about "with" needing to return a single value.

steven-sheehy: Please make the change Xin suggested. jsonb_agg is not needed nor is a subquery. And remove `nftTransferJsonAgg` and your MYK logs.

MarcKriguerAtHedera: I removed the two logging lines; if I make the change that Xin requested, all of the various "nonce" transaction tests will fail with
```
ERROR:  more than one row returned by a subquery used as an expression
```
when we limit jest to a single worker, or
```
  ● Test suite failed to run

    TypeError: Do not know how to serialize a BigInt
        at stringify (<anonymous>)

      at messageParent (node_modules/jest-worker/build/workers/messageParent.js:29:19)
```
with multiple jest workers.

I certainly can eliminate `nftTransferJsonAgg` and just copy the two lines into the query, but I want to resolve the above error first (I think if we resolve it with something like `limit 1` we will not be returning the proper results).

steven-sheehy: I don't think you understand Xin's suggestion. Remove the entire `(.*) as nft_transfer_list`.

xin-hedera: to further clarify

- remove the whole subquery `(.*) as nft_transfer_list`
- add `${Transaction.NFT_TRANSFER}` to the `transactionFields` array
- rename any reference like `row.nft_transfer_list` to `row.nft_transfer`
- fix any errors you see in the spec tests result

hedera-mirror-rest/transactions.js
```diff
@@ -273,7 +266,7 @@ const createTransferLists = async (rows) => {
       max_fee: utils.getNullableNumber(row.max_fee),
       memo_base64: utils.encodeBase64(row.memo),
       name: TransactionType.getName(row.type),
-      nft_transfers: createNftTransferList(row.nft_transfer_list),
+      nft_transfers: createNftTransferList(row.nft_transfer),
```
xin-hedera, line 269: `createNftTransferList` should return empty array `[]` if `!row.nft_transfer` is true

same for `createTokenTransferList`

xin-hedera (changes requested):

hedera-mirror-rest/transactions.js
MarcKriguerAtHedera: I had replaced
```
return undefined;
```
with `return [];` in both those functions; somehow, when I created a new workspace, those changes did not get
copied over.  Those two lines are restored now (and the necessary spec tests tweaked to reflect this).

steven-sheehy (approved):
LGTM

hedera-mirror-rest/transactions.js
mgoelswirlds: I seem to have exactly all these changes.

mgoelswirlds:
I see that we have all common code and we made changes to the exact same files. LGTM, I will pull your code and then send out my PR.

mgoelswirlds (approved):

xin-hedera (approved):
LGTM

PR #6249, opened 2023-06-22
Use node_stake for getNodes gRPC API

### LINKED ISSUE #6171: gRPC nodes API returns inaccurate stake
### Description

The nodes gRPC API still returns the address book stake field. It should return the node_stake.stake like the REST API.

### Steps to reproduce

`grpcurl -plaintext -d '{"file_id": {"fileNum": 102}, "limit": 0}' localhost:5600 com.hedera.mirror.api.proto.NetworkService/getNodes`

### Additional context

_No response_

### Hedera network

mainnet

### Version

0.81.0

### Operating system

None

jascks:
**Description**:
The nodes gRPC API still returns the address book stake field. It should return the `node_stake.stake` like the REST API.
* In service layer, replace `AddressBookEntry` stake with that from latest `NodeStake` loaded from the `node_stake` table.
* Load most recent node stake information from `node_stake` based on the consensus timestamp at some reasonable interval so it is ready to go to satisfy grpc service requests.

**Related issue(s)**:

Fixes #6171 

**Notes for reviewer**:

**Checklist**


hedera-mirror-grpc/src/main/java/com/hedera/mirror/grpc/service/NetworkServiceImpl.java
```diff
@@ -86,8 +97,12 @@ private Flux<AddressBookEntry> page(AddressBookContext context) {
...
             var endpoints = new AtomicInteger(0);
 
-            // This hack ensures that the nested serviceEndpoints is loaded eagerly and voids lazy init exceptions
-            nodes.forEach(e -> endpoints.addAndGet(e.getServiceEndpoints().size()));
+            nodes.forEach(node -> {
+                // Replace node stake with latest present in node_stake table
+                node.setStake(getStakeForNode(node.getNodeId()));
+                // This hack ensures that the nested serviceEndpoints is loaded eagerly and voids lazy init exceptions
+                endpoints.addAndGet(node.getServiceEndpoints().size());
+            });
```
xin-hedera, line 105: the NodeStakeRepository should follow the same pattern in `AddressBookRepository` and `AddressBookEntryRepository` to fetch and cache the node stake.

That is, there should a `Optional<Long> findLatestTimestamp()` to return the latest timestamp in `node_stake` table.

There should be a cached function `List<NodeStake> findByConsensusTimestampAndNodeId(long consensusTimestamp, long firstNodeId, long lastNodeId);`

the firstNodeId and the lastNodeId are from what's returned by `addressBookEntryRepository.findByConsensusTimestampAndNodeId`

Then you wouldn't need all the scheduled and manual cache refreshing logic.

Make sure to handle the edge case when a node's stake is not in `node_stake` table

jascks: Thinking more in terms of a lookup table, not finding a way to get `CacheLoader` to just bulk load and replace with an arbitrary set of keys, and [Ben Manes himself suggesting not using a caching library](https://github.com/ben-manes/caffeine/discussions/898) for what I saw as a similar scenario, I took this approach. I think it would work in the end.

But, looking at your suggestions, and the original code in `NetworkServiceImpl`, it looks like I could perhaps save that latest node stake timestamp within the existing `AddressBookContext`, and in `page()`, refer to the first and last node IDs as you suggested and 'lining up' the two lists copying the stake from one to the other, dealing with the edge case you pointed out. I like it! Thank you @xin-hedera !

xin-hedera: there is a very important difference between our use case and the use case in the link. We don't want to refresh either the address book entry cache or the node stake cache in every X minutes. They should be always from the latest state, so we have a separate cheap `findLatestTimestamp` to determine the latest timestamp, then use it and the node id and other parameter deemed necessary to fetch and cache the data.

jascks: Oh, I just noticed the stuff crossed out above. It looks like you prefer an individual node lookup type function in this case? This will result in _n_ queries to the DB, but then the cache will kick in.

xin-hedera: It’s simpler since then you don’t have to match the node id and set the stake value to the correct node. I’m fine either way

jascks: Yes, that is simpler. :) Really I just need the `stake` column, so I could have that method return `Optional<Long>`. Less read from the database, and less in the cache.

jascks: This has been updated.

hedera-mirror-grpc/src/main/java/com/hedera/mirror/grpc/service/AddressBookProperties.java
```diff
@@ -36,6 +36,13 @@ public class AddressBookProperties {
     @Min(0)
     private long cacheSize = 50L;
 
+    @DurationMin(minutes = 1L)
+    @NotNull
+    private Duration nodeStakeCacheExpiry = Duration.ofMinutes(10);
```
xin-hedera, line 41: node stake changes daily, so a 24-hour expiry can provide best cache hit

hedera-mirror-grpc/src/main/java/com/hedera/mirror/grpc/service/AddressBookProperties.java
```diff
@@ -36,6 +36,13 @@ public class AddressBookProperties {
...
+
+    @Min(0)
+    private long nodeStakeCacheSize = 50L;
```
xin-hedera, line 44: the repository method caches stake per node, should use a larger default cache size

xin-hedera (changes requested):

edwin-greene (approved):
Other than value changes Xin noted, looks good to me.

hedera-mirror-grpc/src/main/java/com/hedera/mirror/grpc/service/AddressBookProperties.java
jascks: Sounds good. I was originally thinking such a long expiry would be a disadvantage if we just loaded the cache 5 minutes before the daily node stake update.

But, `NodeStakeRepository.findLatestTimeStamp()` is _not_ cached, it always goes to the DB and returns the latest timestamp. If that has been advanced, then the node stake cache (entries keyed by timestamp and node ID) will then get loaded by the node stake queries made subsequently.

So, a question, should `NodeStakeRepository.findLatestTimeStamp()` be cached for some duration? It has no method arguments so I'm not sure what the cache key would be, but the associated query is:
```
select max(consensus_timestamp) from node_stake
```
Which is going to read `consensus_timestamp` column from all rows. Does the `node_stake` table get trimmed back at all, or does it eventually end up with days, weeks, months, years of node stake updates, making the query slower over time?

xin-hedera: nope, we don't want to cache `NodeStakeRepository.findLatestTimeStamp()`. it's a cheap query since we have primary key `"node_stake_pkey" PRIMARY KEY, btree (consensus_timestamp, node_id)`, it's a btree index and the first column is the consensus_timestamp, getting the max is very efficient

xin-hedera: btw, if it gets really big, we have a ticket to add a `node_stake_history` table, then the `node_stake` table will only have the current stake of all nodes

jascks: Good to know. Thanks Xin!

jascks: Okay, done.

hedera-mirror-grpc/src/main/java/com/hedera/mirror/grpc/service/AddressBookProperties.java
jascks: Made the executive decision to set the max size to 100. We can revisit when we go unpermissioned.

hedera-mirror-grpc/src/main/java/com/hedera/mirror/grpc/repository/NodeStakeRepository.java
```diff
@@ -0,0 +1,35 @@
...
+
+public interface NodeStakeRepository extends CrudRepository<NodeStake, NodeStake.Id> {
+
+    @Query(value = "select max(consensus_timestamp) from node_stake", nativeQuery = true)
+    Optional<Long> findLatestTimeStamp();
+
+    // All results, including nulls, are cached as not all nodes may be present in the table
+    @Cacheable(cacheManager = CacheConfiguration.NODE_STAKE_CACHE, cacheNames = "node_stake")
+    @Query(value = "select stake from node_stake where consensus_timestamp = ? and node_id = ? ", nativeQuery = true)
+    Optional<Long> findStakeByConsensusTimestampAndNodeId(long consensusTimestamp, long nodeId);
```
steven-sheehy, line 34: I don't think we should do N+1 queries, even if cached. It's inefficient and relies upon reflection. Instead, we should do a similar query as the address book entries and get the same N at once. If we move the map construction into the repository it can also be cached.

```suggestion
    Map<Long, Long> findMapByConsensusTimestampAndNodeId(long consensusTimestamp, long nodeId, long limit) {
        return findByConsensusTimestampAndNodeId(consensusTimestamp, nodeId, limit)
            .stream()
            .collect(toMap(NodeStake::getNodeId, NodeStake::getStake);
    }

    @Query(value = "select stake from node_stake where consensus_timestamp = ? and node_id >= ? order by node_id asc limit ?", nativeQuery = true)
    List<NodeStake> findByConsensusTimestampAndNodeId(long consensusTimestamp, long nodeId, long limit);
```

jascks: Unless every address book entry is guaranteed to have a row in in `node_stake`, the paging won't be in sync. What do you think of [Xin's suggestion](https://github.com/hashgraph/hedera-mirror-node/pull/6249#discussion_r1238782749) on doing this with first and last node ID per page?

steven-sheehy: They will be in sync. If not, there's other serious problems with the network that's out of our control.

jascks: Okay, I will make it so. The node stake cache still requires its own expiry I think, but perhaps the cache config should reference the existing `cacheSize` property since we're talking the same number of nodes and pages.

xin-hedera: I believe when a node just gets added to address book, its stake in `node_stake` won't exist until the first UTC midnight afterwards and the behavior also conforms to the stake HIP. The impact is, if we get multiple node stakes, we have to match the node info from address book and the stake from `node_stake` by node id.

jascks: Interesting. When the node stake update does come, it results in a new `consensus_timestamp` and reflects all of the known nodes in the address book at that time? So, if one was added, or removed, since the previous stake update, then the node stake is now present for all?

Then if a node is added or removed right after, we have to wait for the next stake update to get us back in sync?

xin-hedera: yes. New rows are added to `node_stake` table when mirrornode processes the `NodeStakeUpdateTransaction`. The transaction is a system transaction generated by consensus nodes daily as the first transaction after each UTC midnight. The values for nodes in the transaction body reflects the staking state at the beginning of that UTC day, or the beginning of the staking period.

jascks: Is HIP-21 the "staking HIP"?

I think based on the delay between address book updates and node stake updates, it would be possible for the paging to get out of sync. And, once permissionless, community nodes will become a thing. Will they come and go? 

Also, running on other networks, who knows? For example, running locally (`DEMO`), there are four address book entries, but zero rows in `node_stake`, so that would be out of sync. An empty table or missing node ID in `node_stake` is currently handled by assuming node stake of zero.

steven-sheehy: So we can just load all the node stakes into a map instead of paging it. Even if we somehow have 1 million nodes (which we never will since it would be bad for latency and a waste of computing power) that's only 1000000*(8+8)=15MB of memory to store.

jascks: Do you mean something like [this](https://github.com/hashgraph/hedera-mirror-node/pull/6249/commits/849090318ba2634b719b5ae7646182ac7b52bc4f)?  😄 😄 

Well, maybe I can take another look at `CacheLoader` to find a smoother implementation based on Caffeine...

jascks: Okay,  my vote is following Xin's suggestion on caching and querying the node stake based on the latest node stake consensus timestamp and the first and last node ID found in the address book entry page. I think this will be efficient and I can code around missing rows in `node_stake`. A missing node ID results in zero stake. Likewise, if the `node_stake` table is empty, then all stake is zero.

steven-sheehy: Not sure why you ever used a scheduled task or a CacheLoader to cache all. You'd simply used a Cacheable on the repo method like you're currently doing. Just need to change the query to find all. I'm also fine with the find using first and last node IDs.

jascks: I had lookup (by node ID) table on the brain for some reason. Okay, so your suggestion will cache a single `List<NodeStake>`, which is the latest node stake table contents. Then I could iterate through it matching node IDs, rather than do something highly similar a page at a time.

jascks: Okay, for `NodeStakeRepository`, I have the following:
```
    // An empty result may be cached, indicating the node_stake table is empty
    @Cacheable(cacheManager = CacheConfiguration.NODE_STAKE_CACHE, cacheNames = "node_stake")
    @Query(value = "select * from node_stake where consensus_timestamp = (select max(consensus_timestamp) from "
                    + "node_stake) order by node_id asc",
            nativeQuery = true)
    List<NodeStake> findLatest();
```
This means a couple of things:

1. There is only one cache entry, so I will remove the max cache size property.
2. Basically, the whole table is cached for the expiry period. The find latest timestamp query is part of this query and it won't be issued again until the cache entry expires.
3. Given (2), an expiry of 24 hours could now result in the situation where stale data is kept around for too long after the daily node stake update has occurred.
4. There could be a hybrid implementation, where `findLatestTimestamp()` is invoked on demand (`getNodes()` in the service) and the above query will take a `consensus_timestamp` parameter, which can serve as the cache key. Then there will be no expiry issues.

xin-hedera: Not sure if I missed part of the discussion here, I think the focus is on how to reduce `N+1` `findMapByConsensusTimestampAndNodeId` calls to just one.

Since we want to get a map instead of a list, this should work

```java
    List<NodeStake> findAllByConsensusTimestamp(long consensusTimestamp);

    @Cacheable(...)
    default Map<Long, Long> findAllStakeByConsensusTimestamp(long consensusTimestamp) {
      return findAllByConsensusTimestamp(consensusTimestamp).stream().collect(Collectors.toMap(NodeStake::getNodeId, NodeStake::getStake));
    }
```

Note you don't need a custom query for `findAllByConsensusTimestamp`.

we still need a max cache size property, perhaps 5, but can't be 1 since concurrent requests to the grpc service around when the node stake is being updated may cause some responses with the stake in the previous staking period and others with the stake in the current period, thus we need at least 2 to avoid evicting the cache back and forth.

jascks: Oh, I like that! It implies that `findLatestTimestamp()` is still present in the repository and is invoked per `getNodes()` request of the grpc server, making a 24h expiry for the Map appropriate.

jascks: Done

hedera-mirror-grpc/src/main/java/com/hedera/mirror/grpc/repository/NodeStakeRepository.java
```diff
@@ -0,0 +1,35 @@
...
+import com.hedera.mirror.grpc.config.CacheConfiguration;
+import java.util.Optional;
+import org.springframework.cache.annotation.Cacheable;
+import org.springframework.data.jpa.repository.Query;
+import org.springframework.data.repository.CrudRepository;
+
+public interface NodeStakeRepository extends CrudRepository<NodeStake, NodeStake.Id> {
+
+    @Query(value = "select max(consensus_timestamp) from node_stake", nativeQuery = true)
+    Optional<Long> findLatestTimeStamp();
```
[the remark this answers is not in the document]
jascks: Done

steven-sheehy (approved):
LGTM

xin-hedera (approved):
LGTM

PR #6260, opened 2023-06-25
Persist contract nonce and expose it on the contracts REST API

### LINKED ISSUE #6120: HIP-729 Contract nonce
### Problem

[HIP-729](https://hips.hedera.com/hip/hip-729) is [externalizing](https://github.com/hashgraph/hedera-protobufs/pull/257) contract nonce information in the record stream. Mirror nodes should persist this information and make it available via its APIs.

### Solution

```[tasklist]
### Tasks
- [x] Persist nonce information from new proto field into `Entity.ethereum_nonce` inside `ContractResultServiceImpl`
- [x] Add field to `/api/v1/contracts` and `/api/v1/contracts/{id}`
```

### Alternatives

_No response_

mgoelswirlds:
**Description**:

Persisting contract nonces as per [HIP-729](https://hips.hedera.com/hip/hip-729) . Making it available via the contract APIs
`/api/v1/contracts` and `/api/v1/contracts/{id}`

This PR changes
* Persists entity.ethereum_nonce value. This is done behing a feature flag.
* Changed `/api/v1/contracts` and `/api/v1/contracts/{id}` to return the nonce value

**Related issue(s)**:

Fixes #6120

**Notes for reviewer**:

**Checklist**


hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/ContractResultServiceImpl.java
```diff
@@ -203,6 +203,9 @@ private void processContractResult(
             ByteString sidecarFailedInitcode) {
         // create child contracts regardless of contractResults support
         List<Long> contractIds = getCreatedContractIds(functionResult, recordItem, contractEntityId);
+        if (entityProperties.getPersist().isTrackNonce()) {
```
mgoelswirlds, line 206: Created a feature flag, since we are creating entity object for all contracts.

xin-hedera: we already create entity objects for all contracts, updating the nonce won't count as *new* objects so no need of a feature flag

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/domain/ContractResultServiceImplIntegrationTest.java
```diff
@@ -156,14 +156,71 @@ void processContractCreateNoChildren() {
...
-        assertContractResult(recordItem);
+        assertContractResult(recordItem, 1);
         assertContractLogs(recordItem);
         assertContractActions(recordItem);
         assertContractStateChanges(recordItem);
         assertThat(contractRepository.count()).isZero();
         assertThat(entityRepository.count()).isZero();
     }
 
+    @SuppressWarnings("deprecation")
```
mgoelswirlds, line 167: This test case is currently failing. The idea here is to create a parent and child contract and have the nonces set correctly and verify that respective entities with correct nonces are created. This is not currently happening as the processing of the transaction is not creating correct ContractNonceInfo.
I need some guidance on how to proceed with this test.

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/domain/ContractResultServiceImplIntegrationTest.java
```diff
@@ -681,37 +738,48 @@ private void assertContractStateChanges(RecordItem recordItem) {
...
-                        .result(recordItem.getTransactionRecord().getReceipt().getStatusValue())
-                        .transactionHash(DomainUtils.toBytes(
-                                recordItem.getTransactionRecord().getTransactionHash()))
-                        .type(recordItem.getTransactionType())
-                        .validStartNs(DomainUtils.timeStampInNanos((recordItem
-                                .getTransactionRecord()
-                                .getTransactionID()
-                                .getTransactionValidStart()))))
-                .get();
+    protected void process(RecordItem... recordItems) {
```
mgoelswirlds, line 741: This is done so that both records are processed together.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/ContractResultServiceImpl.java
```diff
@@ -249,6 +252,21 @@ private void processContractResult(
         entityListener.onContractResult(contractResult);
     }
 
+    private void processContractNonce(ContractFunctionResult functionResult) {
+        var nonceInfos = functionResult.getContractNoncesList();
+        if (!nonceInfos.isEmpty()) {
+            nonceInfos.forEach(nonceInfo -> {
```
steven-sheehy, line 258: nit: Could be more succinct.
```suggestion
        functionResult.getContractNoncesList().forEach(nonceInfo -> {
```

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/ContractResultServiceImpl.java
```diff
@@ -249,6 +252,21 @@ private void processContractResult(
...
+                var contractId =
+                        entityIdService.lookup(nonceInfo.getContractId()).orElse(EntityId.EMPTY);
```
steven-sheehy, line 260: We don't need to use the entity service here as entity IDs in the record are guaranteed to always be in shard.realm.num format.

```suggestion
                var contractId = EntityId.of(nonceInfo.getContractId());
```

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/ContractResultServiceImpl.java
```diff
@@ -249,6 +252,21 @@ private void processContractResult(
...
+                Entity entity = contractId.toEntity();
+                entity.setBalance(0L);
```
steven-sheehy, line 262: We should not set a balance since it's not changing.
```suggestion
```

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/ContractResultServiceImpl.java
steven-sheehy, line 206: Move this if statement into `processContractNonce` so the method is self contained and more reusable.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/ContractResultServiceImpl.java
```diff
@@ -309,15 +327,16 @@ private List<Long> getCreatedContractIds(
                 createdContractIds.add(contractId.getId());
                 // The parent contract ID can also sometimes appear in the created contract IDs list, so exclude it
                 if (persist && !contractId.equals(parentEntityContractId)) {
-                    processCreatedContractEntity(recordItem, contractId);
+                    processCreatedContractEntity(recordItem, contractId, functionResult);
```
steven-sheehy, line 330: Revert this and below changes.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/entity/EntityProperties.java
```diff
@@ -67,6 +67,8 @@ public static class PersistProperties {
 
         private boolean trackBalance = true;
 
+        private boolean trackNonce = true;
```
steven-sheehy, line 70: Add to docs/configuration.md

hedera-mirror-rest/__tests__/specs/contracts/all-params.json
```diff
@@ -27,6 +29,7 @@
       {
         "auto_renew_period": 1500,
         "created_timestamp": "987654222123456",
+        "ethereum_nonce": 1,
```
steven-sheehy, line 32: nit: Would be good to set each nonce in this file to a distinct value so it's clear the mapping is correct.

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/domain/RecordItemBuilder.java
```diff
@@ -281,12 +282,13 @@ public ContractFunctionResult.Builder contractFunctionResult() {
...
     @SuppressWarnings("deprecation")
     public ContractFunctionResult.Builder contractFunctionResult(ContractID contractId) {
+        var createdContract = contractId();
         return ContractFunctionResult.newBuilder()
                 .setAmount(5_000L)
                 .setBloom(bytes(256))
                 .setContractCallResult(bytes(16))
                 .setContractID(contractId)
-                .addCreatedContractIDs(contractId())
+                .addCreatedContractIDs(createdContract)
```
steven-sheehy, line 291: Seems like this change is no longer really needed.

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/domain/ContractResultServiceImplIntegrationTest.java
```diff
@@ -681,37 +738,48 @@ private void assertContractStateChanges(RecordItem recordItem) {
...
+                    : consensusStart + 1;
             RecordFile recordFile = domainBuilder
                     .recordFile()
                     .customize(x -> x.consensusStart(consensusStart)
-                            .consensusEnd(consensusStart + 1)
+                            .consensusEnd(consensusEnd)
                             .name(filename))
                     .get();
-
+            transaction = transactions.get(0);
```
steven-sheehy, line 778: Storing only a single variable for a multi-transaction test case seems like it would break test cases

steven-sheehy (changes requested):

build.gradle.kts
```diff
@@ -57,8 +57,8 @@ dependencies {
         api("com.graphql-java-generator:graphql-java-client-runtime:2.1")
         api("com.graphql-java:graphql-java-extended-scalars:20.2")
         api("com.graphql-java:graphql-java-extended-validation:20.0")
-        api("com.hedera.evm:hedera-evm:0.39.0")
-        api("com.hedera.hashgraph:hedera-protobuf-java-api:0.39.1")
+        api("com.hedera.evm:hedera-evm:0.38.0")
```
xin-hedera, line 60: revert since this downgrades the dependency

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/domain/ContractResultServiceImplIntegrationTest.java
```diff
@@ -156,14 +156,71 @@ void processContractCreateNoChildren() {
...
 
+    @SuppressWarnings("deprecation")
+    @Test
+    void processChildContractNonce() {
+
+        RecordItem recordItemParent = recordItemBuilder
+                .contractCreate()
+                .record(r -> r.setContractCreateResult(
+                        r.getContractCreateResult().toBuilder().clearCreatedContractIDs()))
+                .build();
```
xin-hedera, line 175: The problem is how the contract nonce is set in RecordItemBuilder.contractFunctionResult

`recordItemBuilder.contractCreate()` allocates a contract id `0.0.x` as the new contract created, then in `recordItemBuilder.contractFunctionResult()`, the nonce info is set for a different contract `0.0.y`

xin-hedera (changes requested):

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/domain/RecordItemBuilder.java
```diff
@@ -309,7 +310,11 @@ public ContractFunctionResult.Builder contractFunctionResult(ContractID contract
                         .addTopic(bytes(32))
                         .addTopic(bytes(32))
                         .build())
-                .setSenderId(accountId());
+                .setSenderId(accountId())
+                .addContractNonces(ContractNonceInfo.newBuilder()
+                        .setContractId(contractId)
```
mgoelswirlds, line 315: @xin-hedera This is the contractId with which the contract is created. I am not sure how they are different?

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/domain/ContractResultServiceImplIntegrationTest.java
```diff
@@ -164,6 +164,34 @@ void processContractCreateNoChildren() {
         assertThat(entityRepository.count()).isZero();
     }
 
+    @Test
+    void processContractNonce() {
```
mgoelswirlds, line 168: This test simply creates a contract and verifies the ethereum nonce is set.

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/record/entity/AbstractEntityRecordItemListenerTest.java
```diff
@@ -204,15 +205,18 @@ protected final void assertAccount(AccountID accountId, Entity dbEntity) {
         assertThat(dbEntity.getType()).isEqualTo(EntityType.ACCOUNT);
     }
 
-    protected void parseRecordItemAndCommit(RecordItem recordItem) {
+    protected void parseRecordItemAndCommit(RecordItem... recordItems) {
```
mgoelswirlds, line 208: This method is modified to support multiple record items.

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/record/entity/EntityRecordItemListenerContractTest.java
```diff
@@ -761,6 +807,16 @@ var record = parentRecordItem.getTransactionRecord();
                         childRecordItem.getTransactionBody(), childRecordItem.getTransactionRecord()),
                 () -> assertThat(entityRepository.findAll()).contains(setupResult.entity),
                 () -> assertCreatedContract(childRecordItem),
+                () -> assertThat(entityRepository
```
mgoelswirlds, line 810: Testing contractCall in an existing test.

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/record/entity/AbstractEntityRecordItemListenerTest.java
steven-sheehy: There's a `parseRecordItemsAndCommit(List)` right below it...

mgoelswirlds: Yup saw that later. Making the changes.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/ContractResultServiceImpl.java
```diff
@@ -249,6 +251,18 @@ private void processContractResult(
         entityListener.onContractResult(contractResult);
     }
 
+    private void processContractNonce(ContractFunctionResult functionResult) {
+        if (entityProperties.getPersist().isTrackNonce()) {
+            functionResult.getContractNoncesList().forEach(nonceInfo -> {
```
edwin-greene, line 256: Not sure of the performance ramifications of using `.forEach`. In many places outside of tests we use a basic `for` loop or a `for(item : list)` loop for better performance.

steven-sheehy: `Iterable.forEach` can vary for different collection implementations. For `ArrayList`, which I believe this is using, it uses `for (int i = 0; i < col.size(); ++i)` which is more efficient than `for(item : list)`.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/ContractResultServiceImpl.java
```diff
@@ -249,6 +251,18 @@ private void processContractResult(
...
+                var contractId = EntityId.of(nonceInfo.getContractId());
+                Entity entity = contractId.toEntity();
```
xin-hedera, line 258: nit
```suggestion
                var entity = contractId.toEntity();
```

docs/configuration.md
```diff
@@ -131,6 +131,7 @@ value, it is recommended to only populate overridden properties in the custom `a
...
 | `hedera.mirror.importer.parser.record.entity.persist.topicMessageLookups`   | true                           | Persist topic message lookups to the database                                                                                                                                                                                                                      |
 | `hedera.mirror.importer.parser.record.entity.persist.trackBalance`          | true                           | Track entity balance changes and persist to the database                                                                                                                                                                                                           |
+| `hedera.mirror.importer.parser.record.entity.persist.trackNonce`            | true                           | Track nonce changes and persist to the entity database                                                                                                                                                                                                             |
```
xin-hedera, line 134: need to update `EthereumTransactionHandler` to also honor the trackNonce property

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/ContractResultServiceImpl.java
```diff
@@ -203,6 +203,8 @@ private void processContractResult(
             ByteString sidecarFailedInitcode) {
         // create child contracts regardless of contractResults support
         List<Long> contractIds = getCreatedContractIds(functionResult, recordItem, contractEntityId);
+        processContractNonce(functionResult);
+
```
xin-hedera, line 207: should move after the `isContractResults` check

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/ContractResultServiceImpl.java
xin-hedera, line 258: nit
```suggestion
                var entity = contractId.toEntity();
```

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/record/entity/EntityRecordItemListenerContractTest.java
```diff
@@ -139,11 +140,110 @@ var record = recordItem.getTransactionRecord();
                 () -> assertEquals(1, contractResultRepository.count()),
                 () -> assertEquals(3, cryptoTransferRepository.count()),
                 () -> assertContractEntity(recordItem),
+                () -> assertThat(entityRepository.findById(entityId.getEntityNum()))
```
xin-hedera, line 143: nit
```suggestion
                () -> assertThat(entityRepository.findById(entityId.getId()))
```

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/record/entity/EntityRecordItemListenerContractTest.java
```diff
@@ -139,11 +140,110 @@ var record = recordItem.getTransactionRecord();
...
+                () -> assertThat(entityRepository.findById(createdId.getEntityNum()))
+                        .get()
+                        .returns(1L, Entity::getEthereumNonce),
+                () -> assertThat(entityRepository.findById(parentId.getEntityNum()))
+                        .get()
+                        .returns(2L, Entity::getEthereumNonce));
+    }
+
+    @Test
+    void contractCallChildNonce() {
```
xin-hedera, line 199: fine with it, but... we already have test cases like `contractCallToExistingWithChildContractCreate`, can just add nonce info and assert it there

xin-hedera (changes requested):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/ContractResultServiceImpl.java
steven-sheehy: Disagree. Nonces should be tracked regardless of if we store results.

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/domain/ContractResultServiceImplIntegrationTest.java
steven-sheehy, line 168: Need a test case for when trackNonce=false. Can make this one a parameterized test.

steven-sheehy (changes requested):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/ContractResultServiceImpl.java
xin-hedera: this looks controversial since nonce info in proto is part of contract result. it'll look weird to reason about it if the whole contract result is skipped while we increase the nonce, users know the called contract has created a new contract however, there is no record of which function it has called to create the new contract.

steven-sheehy: It's also strange that we couple the two properties together. I set `contractResults=false` since I don't want that table populated since I'm not using it but I set `trackNonces=true` because I do want that. Then users will be confused that `trackNonces` requires `contractResults=true`.

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/record/entity/EntityRecordItemListenerContractTest.java
mgoelswirlds: The other test case you mentioned is using CreatedChildContract, which is now deprecated. I did not want to add to use a deprecated API.

xin-hedera: did you mean `CreatedContractIDs`?

no. `contractCallToExistingWithChildContractCreate` does not use that array, because in the test case it sets up the correct HAIP version `HAPI_VERSION_0_23_0`

In fact,  `CreatedContractIDs` is marked as deprecated but services still fills in the child contract ids in the parent's record afaik.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/ContractResultServiceImpl.java
xin-hedera: yeah that's fair

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/record/entity/EntityRecordItemListenerContractTest.java
mgoelswirlds: Yes I meant CreatedContractIDs. I see [this](https://github.com/hashgraph/hedera-mirror-node/blob/2ea722eb3a0d1d7c3f40922025aa026c44b46b5f/hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/record/entity/EntityRecordItemListenerContractTest.java#L832) in the test , which is a CreatedContractID.

xin-hedera: it's being deprecated so consumers of the contract result in record is advised not to use that field. however, it's not a issue that we get the child contract id from it in test since:

- in the setup it's populated
- as I mentioned, services still populates it

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/transactionhandler/EthereumTransactionHandler.java
```diff
@@ -111,7 +111,9 @@ private void updateAccountNonce(RecordItem recordItem, EthereumTransaction ether
 
         if (!EntityId.isEmpty(senderId)) {
             Entity entity = senderId.toEntity();
-            entity.setEthereumNonce(ethereumTransaction.getNonce() + 1);
+            if (entityProperties.getPersist().isTrackNonce()) {
```
xin-hedera, line 114: move the condition into the outer if so it becomes

`if (!EntityId.isEmpty(senderId) && entityProperties.getPersist().isTrackNonce())`

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/domain/ContractResultServiceImplIntegrationTest.java
```diff
@@ -164,6 +167,35 @@ void processContractCreateNoChildren() {
         assertThat(entityRepository.count()).isZero();
     }
 
+    @ParameterizedTest
+    @ValueSource(booleans = {true, false})
+    void processContractNonce(boolean isTrackNonce) {
```
xin-hedera, line 172: add a `@AfterEach` hook to restore persist.trackNone to `true`, the default.

this is causing other test cases failing since they all share the same test application context so when trackNonce is set to `false`, tests ran after it all see the `false` value

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/domain/ContractResultServiceImplIntegrationTest.java
```diff
@@ -164,6 +167,35 @@ void processContractCreateNoChildren() {
...
+        assertContractLogs(recordItem);
+        assertContractActions(recordItem);
+        assertContractStateChanges(recordItem);
+
+        assertThat(contractRepository.count()).isEqualTo(1);
+        assertThat(entityRepository.findAll())
+                .hasSize(1)
+                .first()
+                .returns(entity.getId(), Entity::getId)
+                .returns(1L, Entity::getEthereumNonce);
```
xin-hedera, line 196: if i'm not mistaken, `domainBuilder.entity().persist()` create the entity with nonce=1, and in the contractResult created by `recordItemBuilder.contractCall` the nonce info is also 1. there is no way to tell if we do track nonce or not regardless of the trackNonce property.

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/domain/ContractResultServiceImplIntegrationTest.java
```diff
@@ -164,6 +167,35 @@ void processContractCreateNoChildren() {
         assertThat(entityRepository.count()).isZero();
     }
 
+    @ParameterizedTest
+    @ValueSource(booleans = {true, false})
+    void processContractNonce(boolean isTrackNonce) {
+        entityProperties.getPersist().setTrackNonce(isTrackNonce);
+        // given
+        var entity = domainBuilder.entity().persist();
```
xin-hedera, line 175: change the entity type to `CONTRACT`

xin-hedera (changes requested):

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/record/entity/EntityRecordItemListenerContractTest.java
mgoelswirlds: Ahh, tried changing it , to then realise , I can't do it , since the CreatedContractIDs are obtained later and I need to set the contractID in the builder with the nonce. I tried setting the values with the builder, but doesn't seem to get set.
So I will leave my test case as is and not change the existing test.

xin-hedera: no problem

steven-sheehy (approved):
LGTM except for some merge issues

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.84.0__set_nonce_for_all_contracts.sql
```diff
@@ -0,0 +1,4 @@
+-- set nonce to null for all contracts in the entity table
+update entity
+set ethereum_nonce = null
+where type='CONTRACT';
```
xin-hedera, line 4: version should be `V1.82.0`, and do the same for `entity_history` table

xin-hedera (changes requested):

xin-hedera (dismissed):
LGTM

docs/configuration.md
```diff
@@ -571,6 +572,7 @@ value, it is recommended to only populate overridden properties in the custom `a
  `hedera.mirror.web3.db.name`                           | mirror_node                                        | The name of the database                                                                                                                                                                      
  `hedera.mirror.web3.db.password`                       | mirror_web3_pass                                   | The database password used to connect to the database                                                                                                                                         
  `hedera.mirror.web3.db.port`                           | 5432                                               | The port used to connect to the database                                                                                                                                                      
+ `hedera.mirror.web3.db.port`                           | 5432                                               | The port used to connect to the database
```
steven-sheehy, line 575: Duplicate

steven-sheehy (approved):

xin-hedera (approved):

PR #6276, opened 2023-06-26
HIP-584: Copy TransferLogic into hedera-mirror-web3

mustafauzunn:
**Description**:

We need to provide TransferLogic implementation in mirror-node,  so we should copy the needed logic from hedera-services into hedera-mirror-web3.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


hedera-mirror-web3/src/main/java/com/hedera/services/ledger/TransferLogic.java
```diff
@@ -0,0 +1,194 @@
...
+            final HederaTokenStore hederaTokenStore,
+            final @Nullable AutoCreationLogic autoCreationLogic,
+            final MirrorEvmContractAliases mirrorEvmContractAliases) {
+        this.hederaTokenStore = hederaTokenStore;
+        this.autoCreationLogic = autoCreationLogic;
+        this.mirrorEvmContractAliases = mirrorEvmContractAliases;
+    }
+
+    public void doZeroSum(
+            final List<BalanceChange> changes, Store store, EntityAddressSequencer ids, Address topLevelPayer) {
```
IvanKavaldzhiev, line 62: In the context of transaction simulations we won't have an active `topLevelPayer`, so I think we can delete any logic related to updating the `topLevelPayer` balance and remove it as an argument. We also should document this change in the javadoc.

hedera-mirror-web3/src/main/java/com/hedera/services/ledger/TransferLogic.java
```diff
@@ -0,0 +1,194 @@
...
+import com.hederahashgraph.api.proto.java.Timestamp;
+import edu.umd.cs.findbugs.annotations.Nullable;
+import java.time.Instant;
+import java.util.List;
+import java.util.Map;
+import java.util.TreeMap;
+import org.apache.tuweni.bytes.Bytes;
+import org.hyperledger.besu.datatypes.Address;
+
+public class TransferLogic {
```
IvanKavaldzhiev, line 47: We should make it a bean.

steven-sheehy (approved):
LGTM

PR #6282, opened 2023-06-27
Add support for Inner Transaction From Query Parameter to Contract Results API

### LINKED ISSUE #5874: Contract Results API 'from' query param does not support sender address
### Description

Currently `api/v1/contracts/results` only supports `from` as an address param filter via `api/v1/contracts/results?from={address}` when the address is the Hedera transaction payer.
This limits search capabilities to Hedera payer accounts only.

In HAPI cases of `ContractCreate` and `ContractCall` this is fine.
However, for all `ContrcactFunctionResults` created via `EthereumTransaction` this is not the case by definition.
In this case `ContrcactFunctionResults.sender` is the real from from an EVM perspective i.e. `tx.from` in the inner transaction.
As such it's important to be able to search via `tx.from` to see all the contract results for a given `tx.from` address.

Explore EthereumTransaction transactions where  account 0.0.4081119 (0x5d5702FABeCeE60f775C51973C0b7f241362486B) is the inner transaction `from` but the HAPI transaction is submitted by account 0.0.1417 (0x0000000000000000000000000000000000000589) 
https://hashscan.io/testnet/account/0.0.4081119?app=false&ph=1&pt=1&p2=1
Some transaction are submitted for 

When searching via https://testnet.mirrornode.hedera.com/api/v1/contracts/results?from=0x5d5702FABeCeE60f775C51973C0b7f241362486B no results are found.
However when searching via the Hedera payer https://testnet.mirrornode.hedera.com/api/v1/contracts/results?from=0x0000000000000000000000000000000000000589 results are returned.

### Steps to reproduce

1. Submit an EthereumTransaction by account A with account B as the inner transaction from to the network
2. Query `/api/v1/contracts/results?from={addressA}` to find results
3. Query `/api/v1/contracts/results?from={addressA}` to not find results

### Additional context

_No response_

### Hedera network

mainnet

### Version

v0.78.0

### Operating system

None

edwin-greene:
**Description**:
Add support for sender addresses in the `from` query parameter for the contract result REST API. Add support for the `from` query parameter to be an evm address.

**Related issue(s)**:

Fixes #5874 

**Notes for reviewer**:

**Checklist**


hedera-mirror-rest/controllers/contractController.js
```diff
@@ -393,7 +392,7 @@ class ContractController extends BaseController {
    * @param {string} contractId encoded contract ID
    * @return {Promise<{conditions: [], params: [], order: 'asc'|'desc', limit: number}>}
    */
-  extractContractResultsByIdQuery = async (filters, contractId) => {
+  extractContractResultsByIdQuery = async (filters, contractId, fromEntity) => {
```
jnels124, line 395: missing js doc for fromEntity

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.82.0__add_contract_result_sender_id_index.sql
```diff
@@ -0,0 +1,6 @@
+-------------------
+-- Add index to contract_result to improve rest api query performance
+-------------------
+
+create index if not exists contract_result__sender_timestamp
```
xin-hedera, line 5: should research if from should be contract_result.sender when sender is set and otherwise default to the transaction payer account id.

if so, instead of adding a new index, we should replace the existing two

```
"contract_result__id_payer_timestamp" btree (contract_id, payer_account_id, consensus_timestamp)
"contract_result__payer_timestamp" btree (payer_account_id, consensus_timestamp)
```

with

```
"contract_result__id_from_timestamp" btree (contract_id, coalesce(sender_id, payer_account_id), consensus_timestamp)
"contract_result__from_timestamp" btree (coalesce(sender_id, payer_account_id), consensus_timestamp)
```

and change the queries to use coalesce to match the index

xin-hedera (changes requested):

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.82.0__add_contract_result_sender_id_index.sql
edwin-greene: I have changed the index to be on the sender_id, with no inclusion of the payer_account_id, after adding the migration for the `coalesce(sender_id, payer_accound_id)`.

hedera-mirror-rest/controllers/contractController.js
edwin-greene: Thanks I have added the doc.

hedera-mirror-rest/controllers/contractController.js
```diff
@@ -435,13 +435,16 @@ class ContractController extends BaseController {
 
       switch (filter.key) {
         case filterKeys.FROM:
-          // handle repeated values
+          if (_.isNil(filter.value)) {
+            // utils.buildAndValidateFilters returns null for evm addresses
+            filter.value = await EntityService.getEncodedId(fromEntity);
```
xin-hedera, line 440: we allow multiple `from=`, so we can't use fromEntity as the original value if the parsed value is null. in addition, when there are multiple `from=`, `req.query['from']` is an array

xin-hedera (changes requested):

hedera-mirror-rest/controllers/contractController.js
edwin-greene: Thanks I have refactored that section to support multiple `from` parameters.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/ContractResultServiceImpl.java
```diff
@@ -211,11 +213,13 @@ private void processContractResult(
...
         // Normalize the two distinct hashes into one 32 byte hash
         var transactionHash = recordItem.getTransactionHash();
+        var payerAccountId = recordItem.getPayerAccountId();
 
         ContractResult contractResult = new ContractResult();
         contractResult.setConsensusTimestamp(recordItem.getConsensusTimestamp());
         contractResult.setContractId(contractEntityId.getId());
-        contractResult.setPayerAccountId(recordItem.getPayerAccountId());
+        contractResult.setPayerAccountId(payerAccountId);
+        contractResult.setSenderId(payerAccountId);
```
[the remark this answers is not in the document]
edwin-greene: Sounds good I have added it.

hedera-mirror-rest/controllers/contractController.js
edwin-greene: This change has been refactored out, so there is no longer an additional parameter to that function.

hedera-mirror-rest/controllers/contractController.js
```diff
@@ -1217,6 +1218,28 @@ class ContractController extends BaseController {
   };
 }
 
+/**
+ * Utils.buildAndValidateFilters returns null for evm addresses.
+ * Update null filter values for the 'from' query parameter with the correct encoded id
+ * @param {} fromQuery
+ * @param {[]} filters
+ */
+const getEncodedEvmAddresses = async (fromQuery, filters) => {
```
xin-hedera, line 1227: should change `utils.formatComparator` to assign the `EntityId` value to comparater.value for `FROM` filter key. Then in `extractContractResultsByIdQuery` resolve the evm address if needed

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.82.0__add_contract_result_sender_id_index.sql
```diff
@@ -0,0 +1,6 @@
...
+  on contract_result (sender_id, consensus_timestamp);
```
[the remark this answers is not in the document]
edwin-greene: Thanks for catching that. I have updated the readme.

hedera-mirror-rest/controllers/contractController.js
edwin-greene: Sounds good, I have made that change.

steven-sheehy (approved):
LGTM

xin-hedera (commented):
LGTM

xin-hedera (approved):

PR #6298, opened 2023-06-29
HIP-584: Copy ApproveAllowanceLogic from services

### LINKED ISSUE #5957: HIP-584: Copy allowance-related logic from services
### Problem

We need to provide `ApproveAllowanceLogic` and `DeleteAllowanceLogic` implementations in mirror-node, so that we can implement all `Approve` precompiles.

### Solution

Copy `ApproveAllowanceLogic` `DeleteAllowanceLogic` and  from mono-services by substituting `Store` and `Ledger` classes with `StackedStateFrames`.

### Alternatives

_No response_

0xivanov:
**Description**:  

We need to provide ApproveAllowanceLogic implementation in mirror-node, so that we can implement all Approve precompiles.

**Related issue(s)**:

Fixes #5957

**Notes for reviewer**:

**Checklist**


0xivanov:
Did some refactoring like moving util methods from `EntityIdUtils` to `IdUtils` as it is in services.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/Downloader.java
```diff
@@ -557,16 +556,22 @@ private Collection<ConsensusNode> partialCollection(Collection<ConsensusNode> al
 
         long totalStake = nodes.get(0).getTotalStake();
         // only keep "just enough" nodes to reach/exceed downloadRatio
-        long neededStake = BigDecimal.valueOf(totalStake).multiply(downloadRatio).setScale(0, RoundingMode.CEILING)
+        long neededStake = BigDecimal.valueOf(totalStake)
+                .multiply(downloadRatio)
+                .setScale(0, RoundingMode.CEILING)
```
steven-sheehy, line 561: Might be some merge issue since this code is from a merged PR

0xivanov: I guess it is again spotlessApply formatting from the rebase from main. Since the merged files are considered "changed" in the context of this branch.

steven-sheehy (approved):

hedera-mirror-web3/src/main/java/com/hedera/services/txns/crypto/ApproveAllowanceLogic.java
```diff
@@ -0,0 +1,223 @@
...
+import java.util.TreeMap;
+
+/**
+ *  Copied Logic type from hedera-services. Differences with the original:
+ *  1. Use abstraction for the state by introducing {@link Store} interface
+ *  2. Remove AccountStore, TypedTokenStore and GlobalDynamicProperties
+ *  3. Remove validateAllowanceLimitsOn since mirror-node properties do not have allowance limit
+ *  4. Using exception from {@link com.hedera.mirror.web3.evm.store.StoreImpl} when entity is missing from the state
+ */
+public class ApproveAllowanceLogic {
```
IvanKavaldzhiev, line 46: We should make the Logic type bean and pass the Store as an argument, where needed.

IvanKavaldzhiev: We should also pass the Maps with `touched` accounts.

0xivanov: There might be concurrency problems with `accountsChanged` and `nftsTouched`. Maybe we should use the thread-safe versions of these classes?

IvanKavaldzhiev: If we pass them as arguments, where needed, we would avoid these problems. We can think of a massive refactoring by making the `StoreImpl` thread-safe internally after we finish with all of the precompiles related work. Then, we could make even more components stateless and reduce memory consumption.

0xivanov: Fixed

PR #6300, opened 2023-06-29
HIP-584: Reduce cache expiration period for contract state to minimize stale data reads

IvanKavaldzhiev:
**Description**:

Currently we use a long period of expiration time (5s) for the cache used for contract states. This PR reduces it to 1s, so that when we have a new block to read as much correct data (for the given block) as possible. There might be still a chance of getting stale values from the previous block, but these cases are minimized.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


steven-sheehy (approved):
LGTM

PR #6306, opened 2023-06-30
update zfs init

### LINKED ISSUE #5990: Investigate/Implement solution for GKE when using ZFS
### Problem

With the addition of openebs zfs, we have introduced a problem when attempting to upgrade GKE versions. Or more generally, the issue is getting the persistent volumes from the zfs provisioner moved to a new node. 

- [ ] update zfs on gke doc to describe new process for GKE upgrade
- [x] implement necessary chart changes
- [x] While updating the chart, add support for creating a service account and configuring for use by the init containers. This service account will need permissions needed to perform the actions in the init script.
- [x] Solution ensures that pods from the stateful set are scheduled on the node containing the data correct for their pod 
- [x] The init pod takes longer than it should to run. This is because it has to install the google sdk to do its job. Maybe we should bake this sdk into an image that we use for the init container.
- [x] verify we are able to get disk metrics for the attached storage

### Solution

Solutions need to be investigated. As a start, you can review this [pr](https://github.com/openebs/zfs-localpv/pull/336/files) and the [faqs](https://github.com/openebs/zfs-localpv/blob/master/docs/faq.md#8-how-to-migrate-pvs-to-the-new-node-in-case-old-node-is-not-accessible).

The solution may involve adding various labels to disks and/or nodes. 

 From inside the init container you can access labels from the node using the following command ` gcloud compute instances describe <instanceName> --zone=<zone> --flatten="metadata.kube-labels[]" | grep -o '<labelName>[^, }]*' | sed -n 's/.*<labelName>=\(.*\).*/\1/p'`

If we can get nodeid label from the links above to work, then we can update the init script to attach labels to the disks it creates and then use those labels in the the init script to identify disks to attach to the new node that came up.

### Alternatives

_No response_

jnels124:
**Description**:
* wait for nodes to be labeled before attempting to create zfs disks
* wait for disk to be available to be attached
* add setup for moving daemonset to init container inside zfs pods

**Related issue(s)**:

Fixes #5990

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/charts/hedera-mirror-common/values.yaml
+++ b/charts/hedera-mirror-common/values.yaml
@@ -416,7 +416,8 @@ zfs:
   # move to zfs.serviceAccount.zfsNode.serviceAccount instead and set create to false once init containers pr resolved in zfs local pv repo
   init:
     serviceAccount:
-      name: default
+      iAmName: zfs-driver-preprod-sa@mirrornode-non-prod-314918.iam.gserviceaccount.com
+      name: zfs-service-account
   coordinator:
     diskSize: 128GB
   enabled: false
--- a/charts/hedera-mirror-common/values.yaml
+++ b/charts/hedera-mirror-common/values.yaml
@@ -413,6 +413,10 @@ traefik:
     type: ClusterIP
 
 zfs:
+  # move to zfs.serviceAccount.zfsNode.serviceAccount instead and set create to false once init containers pr resolved in zfs local pv repo
+  init:
+    serviceAccount:
+      name: default
   coordinator:
     diskSize: 128GB
   enabled: false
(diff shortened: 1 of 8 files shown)


charts/hedera-mirror-common/values.yaml
```diff
@@ -431,11 +435,46 @@ zfs:
...
+            value: 256GB
+        securityContext:
+          privileged: true
+        volumeMounts:
+          - name: root-mount
+            mountPath: /root
+          - name: node-init-entrypoint
+            mountPath: /scripts
   zfsNode:
+    allowedTopologyKeys: "citus-role"
```
jnels124, line 467: fix

PR #6329, opened 2023-07-04
Align AbstractAutoCreationLogic with services

0xivanov:
**Description**:
Align AbstractAutoCreationLogic with services.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


hedera-mirror-web3/src/main/java/com/hedera/services/txns/crypto/AutoCreationLogic.java
```diff
@@ -33,13 +39,20 @@
...
+            final SyntheticTxnFactory syntheticTxnFactory) {
+        super(feeCalculator, evmProperties, syntheticTxnFactory);
     }
 
     @Override
     protected void trackAlias(
-            final JKey jKey, final Address alias, final MirrorEvmContractAliases mirrorEvmContractAliases) {
-        mirrorEvmContractAliases.maybeLinkEvmAddress(jKey, alias);
+            final ByteString alias, final Address address, final MirrorEvmContractAliases mirrorEvmContractAliases) {
+        if (isAliasSizeGreaterThanEvmAddress(alias)) {
```
IvanKavaldzhiev, line 52: It seems there is a missing step before the if clause to link the alias as can be seen [here](https://github.com/hashgraph/hedera-services/blob/96ac267a596d611b650ecec98bc29d190e3c85bb/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/txns/crypto/AutoCreationLogic.java#L81)

0xivanov: Added that line of code, but due to changes in `MirrorEvmContractAliases`, additional check should be done. Because we convert the alias to besu `Address` and if it is not 20 bytes, the conversion fails.

steven-sheehy (dismissed):
LGTM

hedera-mirror-web3/src/main/java/com/hedera/services/txns/crypto/AutoCreationLogic.java
```diff
@@ -33,13 +40,25 @@
...
-        super(feeCalculator, evmProperties);
+    public AutoCreationLogic(
+            final FeeCalculator feeCalculator,
+            final EvmProperties evmProperties,
+            final SyntheticTxnFactory syntheticTxnFactory) {
+        super(feeCalculator, evmProperties, syntheticTxnFactory);
     }
 
     @Override
     protected void trackAlias(
```
IvanKavaldzhiev, line 51: I like the approach. You can document the difference in the logic compared to `hedera-services` and point that we keep `Address` as a key, instead of `ByteString`.

0xivanov: Fixed

steven-sheehy (approved):

PR #6330, opened 2023-07-04
Add transactions flag to /api/v1/accounts/idOrAddress

### LINKED ISSUE #6297: Update accounts get by id API to optionally return transactions
### Problem

Currently the `/api/v1/accounts/{accountIdOrEvmAddress}` endpoint always returns the transactions of an account in addition to the account details. In many cases these are unnecessary and users are better served by calling `/api/v1/transactions?account.id={accountId}`. Still in some cases account transactions are valuable but when not applicable they increase the response size and the mirror node REST API work unnecessarily.

### Solution

Make the retrieval and returning of the transactions field on `/api/v1/accounts/{accountIdOrEvmAddress}` optional.

* Add a `transactions=(true|false)` query param with a default of true
* Adjust the OpenAPI spec
* Change `accounts.js` to not query the `transaction` table when `transactions=false`, but still return `"transactions": []`

### Alternatives

* Add an `/api/v1/accounts?evm_address=` for the relay
* GraphQL API

Ivo-Yankov:
**Description**:
Adds a `transactions` flag to the `/accounts/{idOrHash}` route. If set to `false` no transactions will be included to the response.

**Related issue(s)**:

Fixes #6297 

**Notes for reviewer**:

**Checklist**


hedera-mirror-rest/accounts.js
```diff
@@ -361,6 +361,8 @@ const getOneAccount = async (req, res) => {
   // Parse the filter parameters for account-numbers, balance, and pagination
   const filters = utils.buildAndValidateFilters(req.query, acceptedSingleAccountParameters);
   const encodedId = await EntityService.getEncodedId(req.params[constants.filterKeys.ID_OR_ALIAS_OR_EVM_ADDRESS]);
+
+  const transactionsFilter = filters.find((filter) => filter.key === filterKeys.TRANSACTIONS);
```
xin-hedera, line 365: - we honor the last such boolean params in other places
- use `utils.parseBooleanValue` since the validation logic of a boolean param value in `buildAndValidateFilters` is case-insensitive

hedera-mirror-rest/__tests__/specs/accounts/{id}/transactions-false.json
```diff
@@ -0,0 +1,175 @@
+{
```
xin-hedera, line 1: can combine the two spec files into one with the `tests` section:

```json
{
  "tests": [
    {
      "urls": [...], // no transactions param / with 'transactions=true' param
      "responseStatus": 200,
      "responseJson": {...}
    },
    {
      "urls": [...], // with 'transactions=false' param
      "responseStatus": 200,
      "responseJson": {...}
    }
   ]
}
```

xin-hedera (changes requested):

hedera-mirror-rest/accounts.js
xin-hedera, line 365: findLast?

steven-sheehy (commented):
Please update `docs/rest/README.md` to add a new row for when timestamp=true to show that the transaction tables are not used.

docs/rest/README.md
```diff
@@ -9,49 +9,51 @@ our [Swagger UI](https://mainnet-public.mirrornode.hedera.com/api/v1/docs/#/).
...
+| `/api/v1/accounts/:idOrAlias`                                   | `crypto_transfer`, `entity`, `record_file`, `token_account`, `token_transfer`, `transaction`                                               | Transfers & transactions are present only for legacy reasons. |
+| `/api/v1/accounts/:idOrAlias?timestamp=`                        | `account_balance`, `account_balance_file`, `crypto_transfer`, `entity`, `entity_history`, `token_balance`, `token_transfer`, `transaction` | Transfers & transactions are present only for legacy reasons. |
+| `/api/v1/accounts/:idOrAlias?transactions=false`                | `crypto_transfer`, `entity`, `record_file`, `token_account`, `token_transfer`                                                              | Transfers are present only for legacy reasons.                |
+| `/api/v1/accounts/:idOrAlias?timestamp=true&transactions=false` | `account_balance`, `account_balance_file`, `crypto_transfer`, `entity`, `entity_history`, `token_balance`, `token_transfer`                | Transfers are present only for legacy reasons.                |
```
xin-hedera, line 19:
```suggestion
| `/api/v1/accounts/:idOrAlias?transactions=false` |  `entity`, `record_file`, `token_account`                | |
```

steven-sheehy, line 19: We can probably omit this line since it might be tricky to maintain rows for all combinations of params. This is just a union of the other rows.

steven-sheehy (approved):
LGTM

PR #6347, opened 2023-07-07
Use token balance for accounts by id API with timestamp parameter

### LINKED ISSUE #6304: Accounts by id endpoint should use token balance table when timestamp provided
### Description

 /api/v1/accounts/{id}?timestamp={timestamp} should use the token_balance table

### Steps to reproduce

Execute /api/v1/accounts/{id}?timestamp={timestamp}
Notice the balance returned is for the current balance and not the token_balance that adheres to the timestamp param

### Additional context

_No response_

### Hedera network

other

### Version

v0.0.83

### Operating system

None

mgoelswirlds:
**Description**:

Accounts by id endpoint will now use balance values from the token balance table when timestamp provided 

This PR modifies 
* token balances query in the /accounts/{id} api with timestamp parameter
* In case of no parameter use still uses token_account table

**Related issue(s)**:

Fixes #6304

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-rest/__tests__/specs/accounts/{id}/no-params.json
+++ b/hedera-mirror-rest/__tests__/specs/accounts/{id}/no-params.json
@@ -32,6 +32,38 @@
         "num": 98
       }
     ],
+    "balances": [
+      {
+        "timestamp": "1234567890000000004",
+        "id": 8,
+        "balance": 555,
+        "tokens": [
+          {
+            "token_num": 99998,
+            "balance": 71
+          },
+          {
+            "token_num": 99999,
+            "balance": 72
+          }
+        ]
+      },
+      {
+        "timestamp": "1234567880000000004",
+        "id": 7,
+        "balance": 444,
+        "tokens": [
+          {
+            "token_num": 99998,
+            "balance": 61
+          },
+          {
+            "token_num": 99999,
+            "balance": 62
+          }
+        ]
+      }
+    ],
     "entityStakes": [
       {
         "decline_reward_start": false,
--- a/hedera-mirror-rest/__tests__/specs/accounts/{id}/timestamp-lte.json
+++ b/hedera-mirror-rest/__tests__/specs/accounts/{id}/timestamp-lte.json
@@ -37,13 +37,32 @@
       {
         "timestamp": "1234567890000000004",
         "id": 8,
-        "balance": 555
+        "balance": 555,
+        "tokens": [
+          {
+            "token_num": 99998,
+            "balance": 71
+          },
+          {
+            "token_num": 99999,
+            "balance": 72
+          }
+        ]
       },
-
       {
         "timestamp": "1234567880000000004",
         "id": 8,
-        "balance": 444
+        "balance": 444,
+        "tokens": [
+          {
+            "token_num": 99998,
+            "balance": 61
+          },
+          {
+            "token_num": 99999,
+            "balance": 62
+          }
+        ]
       }
     ],
     "recordFiles": [
@@ -263,8 +282,12 @@
       "balance": 555,
       "tokens": [
         {
+          "token_id": "0.0.99998",
+          "balance": 71
+        },
+        {
           "token_id": "0.0.99999",
-          "balance": 88
+          "balance": 72
         }
       ]
     },
--- a/hedera-mirror-rest/__tests__/specs/accounts/{id}/timestamp-noaccountbalance.json
+++ b/hedera-mirror-rest/__tests__/specs/accounts/{id}/timestamp-noaccountbalance.json
@@ -37,12 +37,21 @@
       {
         "timestamp": "1234567890000000010",
         "id": 8,
-        "balance": 555
+        "balance": 555,
+        "tokens": [
+          {
+            "token_num": 99998,
+            "balance": 71
+          },
+          {
+            "token_num": 99999,
+            "balance": 72
+          }
+        ]
       },
-
       {
         "timestamp": "1234567890000000009",
-        "id": 8,
+        "id": 7,
         "balance": 444
       }
     ],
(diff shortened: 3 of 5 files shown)


hedera-mirror-rest/__tests__/specs/accounts/{id}/no-params.json
```diff
@@ -32,6 +32,38 @@
         "num": 98
       }
     ],
+    "balances": [
```
mgoelswirlds, line 35: This test has been modified for the scenario when a timestamp parameter is not provided, then balances are gottne from the token_accounts table.

hedera-mirror-rest/__tests__/specs/accounts/{id}/timestamp-lte.json
```diff
@@ -37,13 +37,32 @@
       {
         "timestamp": "1234567890000000004",
         "id": 8,
-        "balance": 555
+        "balance": 555,
```
mgoelswirlds, line 40: Note in these test cases balances are fetched from token_balances table and not token_accounts.

hedera-mirror-rest/accounts.js
```diff
@@ -151,11 +151,20 @@ const getEntityBalanceQuery = (
...
   );
 
-  const queries = [
-    `with latest_token_balance as (select account_id, balance, token_id, created_timestamp
-                                       from token_account
-                                       where associated is true)`,
-  ];
+  // If timestamp parameter is present then get values from token_balance
+  const queries = [];
+  if (tokenBalanceQuery.params.length === 2) {
```
jnels124, line 156: This can be changed to check if accountBalanceQuery.query is non empty. This check  is already being done on line 179

hedera-mirror-rest/accounts.js
```diff
@@ -413,6 +418,9 @@ const getOneAccount = async (req, res) => {
       balanceFileTsParams,
       order
     );
+    //Setting the timestamp to be the account balance timestamp
+    tokenBalanceQuery.params = tokenBalanceQuery.params.concat(accountBalanceTs);
```
jnels124, line 422: nit: `balanceFileTs` is a little more clear to me than `accountBalanceTs` rename maybe?

hedera-mirror-rest/__tests__/specs/accounts/{id}/timestamp-noaccountbalance.json
```diff
@@ -261,12 +270,7 @@
     "balance": {
       "timestamp": "1234567890.000000007",
       "balance": 30,
-      "tokens": [
-        {
-          "token_id": "0.0.99999",
-          "balance": 88
-        }
-      ]
+      "tokens": []
```
jnels124, line 273: This may be misleading to indicate the user has no tokens. There could be cases where the token was active at the time satisfying the timestamp param but didn't yet have an entry for the token in a balance file. Not sure we need to make changes to handle this scenario.

jnels124 (commented):
Are there k6 tests for v1?

hedera-mirror-rest/accounts.js
```diff
@@ -385,12 +394,8 @@ const getOneAccount = async (req, res) => {
 
   const accountBalanceQuery = {query: '', params: []};
   if (transactionTsQuery) {
-    let tokenBalanceTsQuery = transactionTsQuery
-      .replaceAll('t.consensus_timestamp', 'created_timestamp')
```
steven-sheehy, line 389: I don't see how the `token_account.created_timestamp` is used anymore. Is it now retrieving an arbitrary token_account and not the one at the correct timestamp?

hedera-mirror-rest/accounts.js
```diff
@@ -151,11 +151,20 @@ const getEntityBalanceQuery = (
...
+  // If timestamp parameter is present then get values from token_balance
+  const queries = [];
+  if (tokenBalanceQuery.params.length === 2) {
+    queries.push(
+      `with latest_token_balance as (select account_id, balance, token_id, consensus_timestamp
+                                       from token_balance)`
+    );
+  } else {
+    queries.push(
+      `with latest_token_balance as (select account_id, balance, token_id, created_timestamp
```
steven-sheehy, line 163: Should normalize the field names.

```suggestion
      `with latest_token_balance as (select account_id, balance, token_id, created_timestamp as consensus_timestamp
```

steven-sheehy (changes requested):

hedera-mirror-rest/accounts.js
jnels124: Don't think that logic is needed anymore. That query was added in the last piece of work because I wasn't aware of the token_balance table and I was limiting the token_account based on it being created within the timestamp bounds. 

When we use token_account now, it will only be in cases where we don't provide the timestamp query.

hedera-mirror-rest/__tests__/specs/accounts/{id}/timestamp-noaccountbalance.json
mgoelswirlds: I believe the idea is to indicate the token balances here. So if they are do not have an entry yet in the balance file , then we should not be showing them here right? And showing a 0 balance will also be misleading I think.

hedera-mirror-rest/accounts.js
mgoelswirlds: Didn't refresh so didn't see your comment there Jesse. But agreed.

steven-sheehy: Think we should remove the created_timestamp from the CTE to reduce confusion since it's no longer necessary.

steven-sheehy (approved):
LGTM

xin-hedera (approved):
LGTM

jnels124 (approved):
LGTM

PR #6373, opened 2023-07-11
HIP-584: Add hasAssociation property to TokenRelationship model

IvanKavaldzhiev:
**Description**:

Currently associate and dissociate operations don't update `hasAssociation` property, that is kept in Postgre DB. This creates issues when updating TokenRelationship models into the stacked state frames. This PR adds this field and updates it accordingly based on the associate/dissociate logic.
**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/TokenRelationshipDatabaseAccessor.java
```diff
@@ -56,6 +56,7 @@ public class TokenRelationshipDatabaseAccessor extends DatabaseAccessor<Object,
                                         false,
                                         false,
                                         Boolean.TRUE == tokenAccount.getAutomaticAssociation(),
+                                        true,
```
steven-sheehy, line 59: Shouldn't this be `Objects.requireNonNullElse(tokenAccount.getAssociated(), true)` instead of assuming it's always true?

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/TokenRelationship.java
```diff
@@ -80,10 +84,11 @@ public TokenRelationship(
         this.destroyed = destroyed;
         this.notYetPersisted = notYetPersisted;
         this.automaticAssociation = automaticAssociation;
+        this.hasAssociation = hasAssociation;
         this.balanceChange = balanceChange;
     }
 
-    public TokenRelationship(Token token, Account account) {
+    public TokenRelationship(Token token, Account account, Boolean hasAssociation) {
```
steven-sheehy, line 91: Mismatch between field and parameter.
```suggestion
    public TokenRelationship(Token token, Account account, boolean hasAssociation) {
```

steven-sheehy (changes requested):

steven-sheehy:
Also, there's test failures

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/TokenRelationshipDatabaseAccessor.java
IvanKavaldzhiev: We filter the results with  `.filter(AbstractTokenAccount::getAssociated)` so getting at line 59, we would always have `true` value.

steven-sheehy (approved):
LGTM

PR #6374, opened 2023-07-11
Add entity stake history

### LINKED ISSUE #6113: Entity stake history
### Problem

As a user, I'd like to know the stake my account had at any day in the past.

### Solution

* Add a `entity_stake.timestamp_range` that is populated from end stake period
* Add an `entity_stake_history` table with the same fields
* Add the same primary key but include timestamp_range
* Add a gist timestamp_range on history
* Move `EntityStake` domain to `AbstractEntityStake`
* Add a `EntityStake` and `EntityStakeHistory` classes that extends `AbstractEntityStake`
* Add `DomainBuilder.entityStakeHistory()`
* Add `EntityStakeHistoryRepository`
* Update repeatable entity stake migration to insert into history table on recreation. Consider using a trigger

### Alternatives

_No response_

xin-hedera:
**Description**:

This PR adds `entity_stake_history` table

- Add `timestamp_range` to `entity_stake` and add a `entity_stake_history` table
- Remove the `entity_state_start` materialized view and change it to a temp table
- Change to trigger pending reward calculation when node stake is updated
- Trigger pending reward calculation on `ApplicationReadyEvent`
- Cap pending reward to the last 365 staking periods
- Upsert `entity_stake` and `entity_stake_history` in `EntityStakeRepository.updateEntityStake`
- Create entity history row on staking reward transfer

**Related issue(s)**:

Fixes #6113 

**Notes for reviewer**:

**Checklist**


hedera-mirror-importer/src/main/resources/db/migration/v1/V1.83.2__add_unknown_entity_type.sql
```diff
@@ -0,0 +1 @@
+alter type entity_type add value if not exists 'UNKNOWN' before 'ACCOUNT';
```
xin-hedera, line 1: the changes from `alter type` has to be committed before the new enum value can be set as default for `entity.type`, thus splitting them into two versioned migrations.

the other way is wrap the `alter type` into a child db transaction, i.e.,

```sql
begin;
alter type entity_type ...;
commit;

alter table if exists entity ...;
alter table if exists entity_history ...;
```

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/migration/FixStakedBeforeEnabledMigrationTest.java
```diff
@@ -43,7 +42,7 @@
 @RequiredArgsConstructor(onConstructor = @__(@Autowired))
 @Tag("migration")
 @TestPropertySource(properties = "spring.flyway.target=1.68.2.1")
-class FixStakedBeforeEnabledMigrationTest extends IntegrationTest {
+class FixStakedBeforeEnabledMigrationTest extends AbstractStakingMigrationTest {
```
xin-hedera, line 45: updated due to `entity_stake` table schema change

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/migration/RecalculatePendingRewardMigrationTest.java
```diff
@@ -41,9 +39,8 @@
 @RequiredArgsConstructor(onConstructor = @__(@Autowired))
 @Tag("migration")
 @TestPropertySource(properties = "spring.flyway.target=1.68.3")
-class RecalculatePendingRewardMigrationTest extends IntegrationTest {
+class RecalculatePendingRewardMigrationTest extends AbstractStakingMigrationTest {
```
xin-hedera, line 42: updated due to `entity_stake` table schema change

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/record/entity/sql/SqlEntityListenerTest.java
```diff
@@ -1456,6 +1456,15 @@ void onStakingRewardTransfer() {
         var transfer1 = domainBuilder.stakingRewardTransfer().get();
         var transfer2 = domainBuilder.stakingRewardTransfer().get();
         var transfer3 = domainBuilder.stakingRewardTransfer().get();
+        var expectedEntities = Stream.of(transfer1, transfer2, transfer3)
```
xin-hedera, line 1459: since staking reward transfer now creates a `Entity` update object with non-null `timestampRange`, the upsert SQL will insert a row to `entity` table.

it would violates the `type` column `not null` constraint, so I ended up set `entity.type` default to `'UNKNOWN'`

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/repository/EntityStakeRepository.java
```diff
@@ -91,42 +188,92 @@ select staked_account_id, sum(balance) as staked_to_me
...
-                  staked_to_me         = excluded.staked_to_me,
-                  stake_total_start    = excluded.stake_total_start;
+            set
+              decline_reward_start = excluded.decline_reward_start,
+              end_stake_period = excluded.end_stake_period,
+              pending_reward = excluded.pending_reward,
+              staked_node_id_start = excluded.staked_node_id_start,
+              staked_to_me = excluded.staked_to_me,
+              stake_total_start = excluded.stake_total_start,
+              timestamp_range = excluded.timestamp_range;
```
xin-hedera, line 274: custom SQL to insert rows to current and history tables.

the history row in the first revision is done by database table trigger on update. It turns out it's tricky to support distributed table trigger with citus and manual maintenance may be needed when the table shards change.

there's also performance concern using trigger when every time we update potentially millions rows.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/entity/staking/EntityStakeCalculator.java
```diff
@@ -16,13 +16,20 @@
...
+import org.springframework.context.event.EventListener;
 import org.springframework.scheduling.annotation.Async;
 import org.springframework.transaction.event.TransactionalEventListener;
 
+@Async
 public interface EntityStakeCalculator {
 
-    @Async
-    @TransactionalEventListener(classes = AccountBalanceFileParsedEvent.class)
+    @TransactionalEventListener(classes = NodeStakeUpdatedEvent.class)
```
xin-hedera, line 28: can't have multiple spring `*EventListener` annotations on a single method

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.83.1__entity_stake_history.sql
```diff
@@ -0,0 +1,45 @@
...
+select change_access_privilege(true);
+
+drop materialized view if exists entity_state_start;
+
+select change_access_privilege(false);
+
+drop function change_access_privilege(grant_or_revoke boolean);
+
+-- add index to node_stake
+create index if not exists node_stake__epoch_day on node_stake (epoch_day);
```
steven-sheehy, line 26: Missed in v2

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.83.2__add_unknown_entity_type.sql
steven-sheehy: Would prefer all 3 migrations condensed into one with child transaction

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/entity/staking/EntityStakeCalculatorImpl.java
```diff
@@ -47,16 +48,24 @@ public void calculate() {
         }
 
         try {
-            if (entityStakeRepository.updated()) {
-                log.info("Skipping since the entity stake is up-to-date");
-                return;
-            }
+            while (true) {
```
steven-sheehy, line 51: When would this need to loop? Shouldn't we have some sleep in between so in case there's some query failure it doesn't consume 100% CPU?

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/entity/staking/EntityStakeCalculatorImpl.java
```diff
@@ -47,16 +48,24 @@ public void calculate() {
...
-            entityRepository.refreshEntityStateStart();
-            log.info("Refreshed entity_state_start in {}", stopwatch);
-            int count = entityStakeRepository.updateEntityStake();
-            log.info("Completed pending reward calculation for {} entities in {}", count, stopwatch);
+                transactionOperations.executeWithoutResult(s -> {
+                    var stopwatch = Stopwatch.createStarted();
+                    entityStakeRepository.createEntityStateStart();
+                    log.info("Created entity_state_start in {}", stopwatch);
+                    entityStakeRepository.updateEntityStake();
+                    Optional<Long> endStakePeriod = entityStakeRepository.getEndStakePeriod();
```
steven-sheehy, line 62: Can we return this value in the previous update statement?

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/entity/staking/EntityStakeCalculatorImpl.java
```diff
@@ -47,16 +48,24 @@ public void calculate() {
...
+                    log.info(
+                            "Completed pending reward calculation of end stake period {} in {}",
+                            endStakePeriod.orElse(null),
+                            stopwatch);
+                });
+            }
         } finally {
             running.set(false);
```
steven-sheehy, line 70: This boolean only works if two threads try to run concurrently. If two processes try to run concurrently or an old query continues when a process dies this won't work. Can we instead take a lock on a table?

steven-sheehy (changes requested):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/entity/staking/EntityStakeCalculator.java
xin-hedera: can try it, no sure what happens since `ApplicationReadyEvent` is not published from a transactional context

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/entity/staking/EntityStakeCalculatorImpl.java
xin-hedera: every iteration only updates entity stake for a staking period. With a loop, we can do catch up if the previous run was interrupted (for example, importer crashed).

if there's a query error, the exception would break the while loop.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/entity/staking/EntityStakeCalculatorImpl.java
xin-hedera: don't think we can, it's a limit from either hibernate or spring data, with multiple-statement query statement, we can't get a return value.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/entity/staking/EntityStakeCalculator.java
xin-hedera: nah, won't trigger on spring boot `ApplicationReadyEvent` if combined into `TransactionalEventListener`

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.83.2__add_unknown_entity_type.sql
xin-hedera: made them into one migration

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/entity/staking/EntityStakeCalculatorImpl.java
xin-hedera: added a `EntityStakeRepository.lockFromConcurrentUpdates` method to lock the `entity_stake` table in `share row exclusive` with option `nowait`

jascks (approved):
LGTM

steven-sheehy (approved):
LGTM

edwin-greene (approved):
Looks good

PR #6378, opened 2023-07-12
HIP-584: Copied `AbstractGrantRevokeKycPrecompile`

### LINKED ISSUE #6335: HIP-584: Copy AbstractGrantRevokeKycPrecompile
### Problem

In order to support `GrantKycPrecompile` and `RevokeKycPrecompile`  we need to copy `AbstractGrantRevokeKycPrecompile`.

### Solution

Copy `AbstractGrantRevokeKycPrecompile` in `com.hedera.services`.

### Alternatives

_No response_

agadzhalov:
**Description**:
Copied `AbstractGrantRevokeKycPrecompile` from `hedera-services` into `mirror-node`

In this PR:
- Added comment what is the different
- Removed class fields and adapted constructors in order to achieve stateless behaviour
    - Removed `sigsVerifier`
    - Removed `accountStore`
    - Removed `tokenStore`

**Related issue(s)**:

Fixes #6335

### DIFF
--- a/hedera-mirror-web3/src/main/java/com/hedera/services/store/contracts/precompile/impl/AbstractGrantRevokeKycPrecompile.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/services/store/contracts/precompile/impl/AbstractGrantRevokeKycPrecompile.java
@@ -16,19 +16,12 @@
 
 package com.hedera.services.store.contracts.precompile.impl;
 
-import static java.util.Objects.requireNonNull;
-
 import com.hedera.mirror.web3.evm.account.MirrorEvmContractAliases;
 import com.hedera.mirror.web3.evm.store.Store;
-import com.hedera.node.app.service.evm.store.contracts.precompile.codec.GrantRevokeKycWrapper;
 import com.hedera.services.store.contracts.precompile.Precompile;
 import com.hedera.services.store.contracts.precompile.SyntheticTxnFactory;
 import com.hedera.services.store.contracts.precompile.codec.RunResult;
 import com.hedera.services.store.contracts.precompile.utils.PrecompilePricingUtils;
-import com.hedera.services.store.models.Id;
-import com.hederahashgraph.api.proto.java.AccountID;
-import com.hederahashgraph.api.proto.java.HederaFunctionality;
-import com.hederahashgraph.api.proto.java.TokenID;
 
 /**
  * This class is a modified copy of AbstractGrantRevokeKycPrecompile from hedera-services repo.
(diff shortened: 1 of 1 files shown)


hedera-mirror-web3/src/main/java/com/hedera/services/store/contracts/precompile/impl/AbstractGrantRevokeKycPrecompile.java
```diff
@@ -0,0 +1,64 @@
...
+ *  3. Run method is modified to return {@link RunResult}, so that getSuccessResultFor is based on this record
+ *  4. Run method is modified to accept {@link Store} as a parameter, so that we abstract from the internal state that is used for the execution
+ */
+public abstract class AbstractGrantRevokeKycPrecompile extends AbstractWritePrecompile {
+    protected final MirrorEvmContractAliases aliases;
+    protected GrantRevokeKycWrapper<TokenID, AccountID> grantRevokeOp;
+    protected Id tokenId;
+    protected Id accountId;
+    protected HederaFunctionality function = null;
+    protected SyntheticTxnFactory syntheticTxnFactory;
```
IvanKavaldzhiev, line 48: We shouldn't keep instances of `grantRevokeOp`, `tokenId`, `accountId` and `function` since this breaks the stateless behaviour. They should be passed as arguments when needed in the child classes.

hedera-mirror-web3/src/main/java/com/hedera/services/store/contracts/precompile/impl/AbstractGrantRevokeKycPrecompile.java
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * This class is a modified copy of AbstractGrantRevokeKycPrecompile from hedera-services repo.
+ *
+ * Differences with the original:
+ *  1. Implements a modified {@link Precompile} interface
+ *  2. Removed class fields and adapted constructors in order to achieve stateless behaviour
+ *  3. Run method is modified to return {@link RunResult}, so that getSuccessResultFor is based on this record
+ *  4. Run method is modified to accept {@link Store} as a parameter, so that we abstract from the internal state that is used for the execution
+ */
+public abstract class AbstractGrantRevokeKycPrecompile extends AbstractWritePrecompile {
```
steven-sheehy, line 34: Is there a test we can copy for this? Is there any concrete class we can copy that we can load as beans?

agadzhalov: Similarly to other abstract classes threre's no corresponding test for `AbstractGrantRevokeKycPrecompile`. The code in `AbstractGrantRevokeKycPrecompile` will be covered from `GrantKycPrecompile` and `RevokeKycPrecompile` tests

PR #6379, opened 2023-07-12
disable synthetic contract results

### LINKED ISSUE #5909: Extend synthetic contract log logic to populate contract result table

jnels124:
**Description**:
disable persisting synthetic contract results

**Related issue(s)**:

Fixes #5909 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/contractresult/SyntheticContractResultServiceImplTest.java
+++ b/hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/contractresult/SyntheticContractResultServiceImplTest.java
@@ -27,6 +27,7 @@ import com.hedera.mirror.importer.parser.record.entity.EntityListener;
 import com.hedera.mirror.importer.parser.record.entity.EntityProperties;
 import com.hederahashgraph.api.proto.java.TokenID;
 import com.hederahashgraph.api.proto.java.TokenType;
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Test;
@@ -63,6 +64,11 @@ class SyntheticContractResultServiceImplTest {
         amount = recordItem.getTransactionBody().getTokenMint().getAmount();
 
         entityProperties.getPersist().setSyntheticContractResults(true);
+    }
+
+    @AfterEach
+    void afterEach() {
+        entityProperties.getPersist().setSyntheticContractResults(false);
     }
 
     @Test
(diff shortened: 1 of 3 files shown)


edwin-greene (dismissed):
Looks good

steven-sheehy (dismissed):

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/contractresult/SyntheticContractResultServiceImplTest.java
```diff
@@ -61,6 +61,8 @@ void beforeEach() {
         receiverId =
                 EntityId.of(recordItem.getTransactionBody().getTransactionID().getAccountID());
         amount = recordItem.getTransactionBody().getTokenMint().getAmount();
+
+        entityProperties.getPersist().setSyntheticContractResults(true);
```
steven-sheehy, line 65: Need to set it back to false in an `@After` so as to not impact other tests.

hedera-mirror-importer/src/test/resources/config/application.yml
```diff
@@ -24,6 +24,8 @@ hedera:
           pubsub:
             topicName: testTopic
           entity:
+            persist:
+              syntheticContractResults: true
```
steven-sheehy, line 28: Would it be better to set entityProperties.getPersist().setSyntheticContractResults(true); in the few tests that need it instead of globally?

jnels124: we could do that but I put it here so going forward any tests written don't get slipped in without accounting for this that then break if/when we enable it in the future

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/contractresult/SyntheticContractResultServiceImplTest.java
jnels124: this is a unit test and the other tests that need this property to be false already set it. However, I added the after each anyways

mgoelswirlds (approved):

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/contractresult/SyntheticContractResultServiceImplTest.java
steven-sheehy: Ok, missed that it was a unit test since it's not yet merged into the branch I'm working on to check.

steven-sheehy (approved):

PR #6409, opened 2023-07-17
HIP-584: Fix mint and burn precompile behaviour

IvanKavaldzhiev:
**Description**:

This PR fixes issues related to mint and burn precompiles:

1. Now set expiration timestamp in seconds inside Token model, instead of nanoseconds. The uncorrect and bigger nanoseconds value made an overflow exception in the pricing logic and could break multiple precompile pricing logic calculations.
2. Determine correctly if a token's owner is the token's treasury. Copied `hedera-services` logic had different workflow and behaviour compared to mirror-node mechanism of reading data from DB. In services, owner is set to Id.DEFAULT(0,0,0) if it's the treasury. In mirror-node it's set to what the value is from DB.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


steven-sheehy (approved):
LGTM

steven-sheehy:
Seems like a test issue. Also might be good to add some tests for these two scenarios.

steven-sheehy (approved):

PR #6429, opened 2023-07-18
Remove long to Instant conversions in gRPC API

### LINKED ISSUE #6427: Reduce unnecessary long to Instant conversion in gRPC API
### Problem

The old duplicate `TopicMessage` removed in #6294 used an `Instant` for start and end time but the common `TopicMessage` uses long. As a result, there is now a lot of wasteful long to/from Instant conversion taking place.

### Solution

* Change `TopicMessageFilter` to use long instead of Instant
* Remove grpc converters
* Remove grpc validators in favor of `@AssertTrue`

### Alternatives

_No response_

steven-sheehy:
**Description**:

* Change `TopicMessageFilter` to use long instead of Instant
* Remove converters
* Remove validators in favor of `@AssertTrue`

**Related issue(s)**:

Fixes #6427

**Notes for reviewer**:

**Checklist**


hedera-mirror-grpc/src/test/java/com/hedera/mirror/grpc/service/TopicMessageServiceTest.java
```diff
@@ -127,7 +126,7 @@ void endTimeEqualsStartTime() {
     @Test
     void startTimeAfterNow() {
         TopicMessageFilter filter = TopicMessageFilter.builder()
-                .startTime(Instant.now().plusSeconds(10L))
+                .startTime(now + 60 * 60 * NANOS_PER_SECOND)
```
jascks, line 129: I assume moving from now + 10 seconds to now + 1 hour is intentional?

steven-sheehy: No, but for this test case either value works 😅

hedera-mirror-grpc/src/test/java/com/hedera/mirror/grpc/service/TopicMessageServiceTest.java
```diff
@@ -363,11 +356,11 @@ void incomingMessagesWithLimit() {
 
     @Test
     void incomingMessagesWithEndTimeBefore() {
-        Instant endTime = Instant.now().plusMillis(500);
-        Flux<TopicMessage> generator = domainBuilder.topicMessages(2, endTime.minusNanos(2));
+        long endTime = now + 500_000_000L;
```
jascks, line 359: Like how I've seen `DomanUtils.NANOS_PER_SECOND` being used, what if we had `DomainUtils.NANOS_PER_MILLISECOND` with a value of `1_000_000L`. Then:

```long endTime = now + 500 * NANOS_PER_MILLISECOND;```

jascks (approved):
LGTM

edwin-greene (approved):
Looks good

PR #6446, opened 2023-07-19
HIP-584: Copy ApprovePrecompile

### LINKED ISSUE #5652: HIP-584: Copy implementation for approve precompiles
### Problem

We need default implementations for precompiles related to approve logic, so we should copy needed logic from hedera-services into hedera-mirror-web3.

### Solution

Copy precompile logic for the following precompiles:

- ApprovePrecompile

### Alternatives

_No response_

0xivanov:
**Description**:
We need default implementations for precompiles related to approve logic, so we should copy needed logic from hedera-services into hedera-mirror-web3.

**Related issue(s)**:

Fixes #5652

**Notes for reviewer**:

**Checklist**


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/StoreImpl.java
```diff
@@ -144,6 +148,13 @@ public boolean hasAssociation(TokenRelationshipKey tokenRelationshipKey) {
         return getTokenRelationship(tokenRelationshipKey, OnMissing.DONT_THROW).hasAssociation();
     }
 
+    @Override
+    public boolean hasApprovedForAll(Address ownerAddress, AccountID operatorId, TokenID tokenId) {
+        final Set<FcTokenAllowanceId> approvedForAll =
+                getAccount(ownerAddress, OnMissing.THROW).getApproveForAllNfts();
+        return approvedForAll.contains(FcTokenAllowanceId.from(tokenId, operatorId));
```
steven-sheehy, line 155: Loading all of an accounts NFTs just to check one of them is very inefficient and could crash the service by running out of memory. Can we not just do a query?

hedera-mirror-web3/src/main/java/com/hedera/services/store/contracts/precompile/SyntheticTxnFactory.java
```diff
@@ -103,6 +114,67 @@ public TransactionBody.Builder createDissociate(final Dissociation dissociation)
...
+        return createNonfungibleApproval(approveWrapper, ownerId, null);
+    }
+
+    /**
+     * Copied Logic type from hedera-services.
+     *
+     * Differences with the original:
+     *  1. Using {@link Id} instead of EntityId as types for the owner and operator
+     * */
+    public TransactionBody.Builder createNonfungibleApproval(
```
steven-sheehy, line 134: It's weird that it's `createNonfungibleApproval` but has an `if (approveWrapper.isFungible())`. Is that intentional?

hedera-mirror-web3/src/main/java/com/hedera/services/store/contracts/precompile/codec/ApproveParams.java
```diff
@@ -0,0 +1,26 @@
...
+
+package com.hedera.services.store.contracts.precompile.codec;
+
+import com.hedera.mirror.web3.evm.store.Store;
+import org.hyperledger.besu.datatypes.Address;
+
+/**
+ * Record containing specific body arguments for Approve precompiles.
+ * */
+public record ApproveParams(Address tokenAddress, Address senderAddress, Store store, boolean isFungible)
```
steven-sheehy, line 25: It seems like an abuse to pass around the Store this way. Shouldn't it be retrievable through some other mechanism?

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/services/store/contracts/precompile/SyntheticTxnFactory.java
0xivanov: That is how it's written in hedera-services. I could refactor it but maybe it is better to keep the code closer to the original.

hedera-mirror-web3/src/main/java/com/hedera/services/store/contracts/precompile/codec/ApproveParams.java
0xivanov: In the ApprovePrecompile body method we need to fetch the `ownerId` if we approve NFT. I could pass the `ownerId` instead of `Store` from `HtsPrecompiledContract`, but in order to get the `ownerId` from the store we need to first decode the transaction input (which is internal logic for the precompile). The only optimisation i can think of right now is to pass `null` when the operation is for fungible token and pass the `Store` only when the operation is for NFT.

IvanKavaldzhiev: I see 2 options:

1. Enhance the body() signature with one more argument which is the Store. This would change the signature of all precompiles but would be a clean approach.
2. Add a new utility method inside `ApprovePrecompile`, which would be used only in `web3`, which based on the `Bytes input` retrieves the `tokenId`. Then in `HTSPrecompiledContract` fetch the `ownerId`, using the `tokenId` and pass the ownerId in `ApproveParams`.

Unless we have a second or more precompiles that need the `Store` into the `body()` method I lean towards the second approach.

0xivanov: Implemented the second option

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/StoreImpl.java
0xivanov: Currently, we need to read the whole `Account` from the `Store` in order to get the most up to date information. We can create separate PR with optimisations.

steven-sheehy (dismissed):
LGTM. Just need to merge main and can merge once Ivan K approves

IvanKavaldzhiev (approved):
LGTM

Needs thorough testing since there are a lot of cases here, but that would be covered by acceptance tests.

PR #6454, opened 2023-07-20
HIP-584: Adapt Mirror Archive Node to respect alias addresses

### LINKED ISSUE #6451: HIP-584: Calling eth_call with account alias address returns empty result
### Description

Invoking `eth_call` requests through the Mirror Archive Node when passing as an argument alias form of an address returns empty result instead of the expected value.

### Steps to reproduce

1. Approve given spender with allowance of 200 for a fungible token using the JSON-RPC Relay
2. Perform eth_call request to see what allowance the given spender has
3. Expect the return result to be 200
4. The output result is 0 -> 0x0000000000000000000000000000000000000000000000000000000000000000

### Additional context

_No response_

### Hedera network

testnet

### Version

v0.84.0

### Operating system

None

IvanKavaldzhiev:
**Description**:

Currently when we try to invoke accounts with alias addresses from the in-memory store we don't find them, since we persist them using their long-zero form. This PR fixes the issue and now we first resolve the alias to a long-zero address and then we try to find it from the in-memory store. This issue is present since 0.84.0 release of mirror-node.
**Related issue(s)**:

Fixes #6451 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/account/AccountAccessorImpl.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/account/AccountAccessorImpl.java
@@ -16,6 +16,11 @@
 
 package com.hedera.mirror.web3.evm.account;
 
+import static com.hedera.node.app.service.evm.store.models.HederaEvmAccount.ECDSA_KEY_ALIAS_PREFIX;
+import static com.hedera.node.app.service.evm.utils.EthSigsUtils.recoverAddressFromPubKey;
+import static com.hedera.services.utils.EntityIdUtils.isOfEcdsaAddressSize;
+import static com.hedera.services.utils.EntityIdUtils.isOfEvmAddressSize;
+
 import com.google.protobuf.ByteString;
 import com.hedera.mirror.web3.evm.store.Store;
 import com.hedera.mirror.web3.evm.store.Store.OnMissing;
@@ -29,13 +34,13 @@ import org.hyperledger.besu.datatypes.Address;
 public class AccountAccessorImpl implements AccountAccessor {
     public static final int EVM_ADDRESS_SIZE = 20;
 
-    private final HederaEvmEntityAccess mirrorEntityAccess;
     private final Store store;
+    private final HederaEvmEntityAccess mirrorEntityAccess;
+    private final MirrorEvmContractAliases aliases;
 
     @Override
     public Address canonicalAddress(Address addressOrAlias) {
-        final var account = store.getAccount(addressOrAlias, OnMissing.DONT_THROW);
-        if (!account.isEmptyAccount()) {
+        if (aliases.isInUse(addressOrAlias)) {
             return addressOrAlias;
         }
 
@@ -48,14 +53,29 @@ public class AccountAccessorImpl implements AccountAccessor {
     }
 
     public Address getAddressOrAlias(final Address address) {
-        final ByteString alias;
-        if (!mirrorEntityAccess.isExtant(address)) {
+        // An EIP-1014 address is always canonical
+        if (!aliases.isMirror(address)) {
             return address;
         }
-        alias = mirrorEntityAccess.alias(address);
 
-        if (!alias.isEmpty() && alias.size() == EVM_ADDRESS_SIZE) {
-            return Address.wrap(Bytes.wrap(alias.toByteArray()));
+        final var account = store.getAccount(address, OnMissing.DONT_THROW);
+        final ByteString alias;
+
+        if (account == null) {
+            return address;
+        } else {
+            alias = account.getAlias();
+        }
+
+        if (!alias.isEmpty()) {
+            if (isOfEvmAddressSize(alias)) {
+                return Address.wrap(Bytes.wrap(alias.toByteArray()));
+            } else if (isOfEcdsaAddressSize(alias) && alias.startsWith(ECDSA_KEY_ALIAS_PREFIX)) {
+                final byte[] value = recoverAddressFromPubKey(alias.substring(2).toByteArray());
+                if (value.length > 0) {
+                    return Address.wrap(Bytes.wrap(value));
+                }
+            }
         }
         return address;
     }
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmTxProcessorFacadeImpl.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmTxProcessorFacadeImpl.java
@@ -102,10 +102,10 @@ public class MirrorEvmTxProcessorFacadeImpl implements MirrorEvmTxProcessorFacad
                 (int) evmProperties.getExpirationCacheTime().toSeconds();
         final var store = new StoreImpl(databaseAccessors);
         final var mirrorEntityAccess = new MirrorEntityAccess(contractStateRepository, contractRepository, store);
-        final var tokenAccessor = new TokenAccessorImpl(evmProperties, store);
-        final var accountAccessor = new AccountAccessorImpl(mirrorEntityAccess, store);
-        final var codeCache = new AbstractCodeCache(expirationCacheTime, mirrorEntityAccess);
         final var mirrorEvmContractAliases = new MirrorEvmContractAliases(store);
+        final var tokenAccessor = new TokenAccessorImpl(evmProperties, store, mirrorEvmContractAliases);
+        final var accountAccessor = new AccountAccessorImpl(store, mirrorEntityAccess, mirrorEvmContractAliases);
+        final var codeCache = new AbstractCodeCache(expirationCacheTime, mirrorEntityAccess);
         final var mirrorOperationTracer = new MirrorOperationTracer(traceProperties, mirrorEvmContractAliases);
 
         final var worldState = new HederaEvmWorldState(
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/token/TokenAccessorImpl.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/token/TokenAccessorImpl.java
@@ -22,6 +22,7 @@ import static com.hedera.mirror.web3.evm.utils.EvmTokenUtils.evmKey;
 import static com.hedera.services.utils.MiscUtils.asKeyUnchecked;
 
 import com.hedera.mirror.common.domain.entity.EntityId;
+import com.hedera.mirror.web3.evm.account.MirrorEvmContractAliases;
 import com.hedera.mirror.web3.evm.exception.ParsingException;
 import com.hedera.mirror.web3.evm.properties.MirrorNodeEvmProperties;
 import com.hedera.mirror.web3.evm.store.Store;
@@ -50,6 +51,7 @@ public class TokenAccessorImpl implements TokenAccessor {
 
     private final MirrorNodeEvmProperties properties;
     private final Store store;
+    private final MirrorEvmContractAliases aliases;
 
     @Override
     public Optional<EvmTokenInfo> evmInfoForToken(Address address) {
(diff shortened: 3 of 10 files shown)


hedera-mirror-web3/src/main/java/com/hedera/services/utils/EntityIdUtils.java
```diff
@@ -130,6 +131,14 @@ public static TokenID tokenIdFromEvmAddress(final Address address) {
         return tokenIdFromEvmAddress(address.toArrayUnsafe());
     }
 
+    public static boolean isOfEvmAddressSize(final ByteString alias) {
+        return alias.size() == EVM_ADDRESS_SIZE;
+    }
+
+    public static boolean isOfEcdsaAddressSize(final ByteString alias) {
```
Nana-EC, line 138: This is for a publicKey right? If so
```suggestion
    public static boolean isOfEcdsaPublicKeySize(final ByteString alias) {
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmTxProcessorFacadeImpl.java
```diff
@@ -102,10 +102,10 @@ public HederaEvmTransactionProcessingResult execute(
                 (int) evmProperties.getExpirationCacheTime().toSeconds();
         final var store = new StoreImpl(databaseAccessors);
         final var mirrorEntityAccess = new MirrorEntityAccess(contractStateRepository, contractRepository, store);
-        final var tokenAccessor = new TokenAccessorImpl(evmProperties, store);
-        final var accountAccessor = new AccountAccessorImpl(mirrorEntityAccess, store);
-        final var codeCache = new AbstractCodeCache(expirationCacheTime, mirrorEntityAccess);
         final var mirrorEvmContractAliases = new MirrorEvmContractAliases(store);
+        final var tokenAccessor = new TokenAccessorImpl(evmProperties, store, mirrorEvmContractAliases);
```
Nana-EC, line 106: Q: is there a reason for the positional move?
if so add a comment if it's important to instantiate in a certain order, although that seem odd.
If not then no-op

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/utils/EthSigsUtils.java
```diff
@@ -0,0 +1,58 @@
...
+
+import com.sun.jna.ptr.LongByReference;
+import java.nio.ByteBuffer;
+import org.bouncycastle.jcajce.provider.digest.Keccak;
+import org.hyperledger.besu.nativelib.secp256k1.LibSecp256k1;
+
+/**
+ * This utility class is a copy from hedera-services. It could be used to validate test scenarios using address with alias format.
+ * */
+public final class EthSigsUtils {
```
Nana-EC, line 30: Add an `EthSigsUtilsTest.java` with coverage for these methods

hedera-mirror-web3/src/main/java/com/hedera/services/utils/EntityIdUtils.java
```diff
@@ -130,6 +131,14 @@ public static TokenID tokenIdFromEvmAddress(final Address address) {
         return tokenIdFromEvmAddress(address.toArrayUnsafe());
     }
 
+    public static boolean isOfEvmAddressSize(final ByteString alias) {
```
Nana-EC, line 134: Add UT test coverage

hedera-mirror-web3/src/main/java/com/hedera/services/utils/EntityIdUtils.java
Nana-EC, line 138: Add UT test coverage

Nana-EC (changes requested):
Q's, suggestions and test requests

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmTxProcessorFacadeImpl.java
IvanKavaldzhiev: The reposition is due to we need to first initialize the `mirrorEvmContractAliases`, so that we can pass it in the constructors below.

hedera-mirror-web3/src/main/java/com/hedera/services/utils/EntityIdUtils.java
IvanKavaldzhiev: Renamed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/utils/EthSigsUtils.java
IvanKavaldzhiev: Added tests

hedera-mirror-web3/src/main/java/com/hedera/services/utils/EntityIdUtils.java
IvanKavaldzhiev: Added

hedera-mirror-web3/src/main/java/com/hedera/services/utils/EntityIdUtils.java
IvanKavaldzhiev: Added

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmTxProcessorFacadeImpl.java
Nana-EC: Was on a small screen at 1st pass, I see it now. Thanks

Nana-EC (approved):
LG barring any Mirror Node team review found issues

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/token/TokenAccessorImpl.java
```diff
@@ -80,25 +82,27 @@ public boolean isTokenAddress(final Address address) {
 
     @Override
     public boolean isFrozen(final Address address, final Address token) {
-        final var tokenRelationship =
-                store.getTokenRelationship(new TokenRelationshipKey(token, address), OnMissing.DONT_THROW);
+        final var tokenRelationship = store.getTokenRelationship(
+                new TokenRelationshipKey(token, aliases.resolveForEvm(address)), OnMissing.DONT_THROW);
```
steven-sheehy, line 86: There's a lot of `aliases.resolveForEvm()` calls. Would it make more sense to push this logic into the store itself to stay DRY?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/account/AccountAccessorImpl.java
```diff
@@ -48,14 +53,29 @@ public boolean isTokenAddress(Address address) {
...
+            alias = account.getAlias();
         }
-        alias = mirrorEntityAccess.alias(address);
 
-        if (!alias.isEmpty() && alias.size() == EVM_ADDRESS_SIZE) {
-            return Address.wrap(Bytes.wrap(alias.toByteArray()));
+        if (!alias.isEmpty()) {
+            if (isOfEvmAddressSize(alias)) {
+                return Address.wrap(Bytes.wrap(alias.toByteArray()));
+            } else if (isOfEcdsaPublicAddressSize(alias) && alias.startsWith(ECDSA_KEY_ALIAS_PREFIX)) {
```
steven-sheehy, line 73: codecov indicates lines 72, 74, and 76 are missing test coverage.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/account/AccountAccessorImpl.java
```diff
@@ -48,14 +53,29 @@ public boolean isTokenAddress(Address address) {
...
+        // An EIP-1014 address is always canonical
+        if (!aliases.isMirror(address)) {
+            return address;
+        }
+
+        final var account = store.getAccount(address, OnMissing.DONT_THROW);
         final ByteString alias;
-        if (!mirrorEntityAccess.isExtant(address)) {
+
+        if (account == null) {
```
steven-sheehy, line 64: I don't think the store can return null. Also think you need to check empty account.
```suggestion
        if (account.isEmptyAccount()) {
```

hedera-mirror-web3/src/test/java/com/hedera/services/utils/EthSigsUtilsTest.java
```diff
@@ -0,0 +1,40 @@
...
+
+package com.hedera.services.utils;
+
+import static com.swirlds.common.utility.CommonUtils.unhex;
+import static org.junit.jupiter.api.Assertions.assertArrayEquals;
+
+import com.hedera.node.app.service.evm.utils.EthSigsUtils;
+import org.junit.jupiter.api.Test;
+
+class EthSigsUtilsTest {
```
steven-sheehy, line 25: Strange that this is testing the EthSigsUtils in the evm library and we have a copied one in web3. Think this test and the web3 util should both be removed.

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/utils/EthSigsUtils.java
```diff
@@ -0,0 +1,58 @@
...
+
+    private EthSigsUtils() {}
+
+    public static byte[] recoverAddressFromPubKey(byte[] pubKeyBytes) {
```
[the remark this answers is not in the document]
IvanKavaldzhiev: I missed that we already have `EthSigsUtils` in `hedera-evm`. I removed the class and the associated test.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/token/TokenAccessorImpl.java
IvanKavaldzhiev: I integrated this logic into the Store, where applicable.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/account/AccountAccessorImpl.java
IvanKavaldzhiev: Added coverage

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/account/AccountAccessorImpl.java
IvanKavaldzhiev: Fixed

hedera-mirror-web3/src/test/java/com/hedera/services/utils/EthSigsUtilsTest.java
IvanKavaldzhiev: Removed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/account/MirrorEvmContractAliases.java
```diff
@@ -87,21 +87,21 @@ private Optional<Address> resolveFromAliases(Address alias) {
...
-    private Address resolveFromEntityAccess(Address addressOrAlias) {
-        final Address resolvedAddress;
-        final var token = store.getToken(addressOrAlias, OnMissing.DONT_THROW);
-        // if token is missing - check for account
-        if (token.isEmptyToken()) {
-            // if token and account are missing - throw
-            final var account = store.getAccount(addressOrAlias, OnMissing.THROW);
-            resolvedAddress = account.getAccountAddress();
+    private Address resolveFromStore(Address addressOrAlias) {
+        final var entity = entityRepository.findByEvmAddressAndDeletedIsFalse(addressOrAlias.toArray());
```
steven-sheehy, line 91: This is a non-cached method whereas before it was at least cached by the store. This could probably be called many times for the same address, right? So we should probably add a cache on the repository.

Alternatively, why don't we just keep using the store and whenever we update the account store it with both its long zero and alias address? Then regardless of which address you lookup you'll get the same account back.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/account/MirrorEvmContractAliases.java
```diff
@@ -87,21 +87,21 @@ private Optional<Address> resolveFromAliases(Address alias) {
...
 
-    private Address resolveFromEntityAccess(Address addressOrAlias) {
-        final Address resolvedAddress;
-        final var token = store.getToken(addressOrAlias, OnMissing.DONT_THROW);
-        // if token is missing - check for account
-        if (token.isEmptyToken()) {
-            // if token and account are missing - throw
-            final var account = store.getAccount(addressOrAlias, OnMissing.THROW);
-            resolvedAddress = account.getAccountAddress();
+    private Address resolveFromStore(Address addressOrAlias) {
```
steven-sheehy, line 90: Since we're no longer using the store this method seems inappropriately named.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/account/MirrorEvmContractAliases.java
```diff
@@ -87,21 +87,21 @@ private Optional<Address> resolveFromAliases(Address alias) {
...
+        if (entity.isEmpty()) {
+            return Address.ZERO;
         } else {
-            // if token is present - get id
-            resolvedAddress = token.getId().asEvmAddress();
+            final var resolvedAddress = new Id(
+                            entity.get().getShard(),
+                            entity.get().getRealm(),
+                            entity.get().getNum())
+                    .asEvmAddress();
```
steven-sheehy, line 100:
```suggestion
            final var resolvedAddress = entity.map(e -> new Id(e.getShard(), e.getRealm(), e.getNum()).map(Id::asEvMAddress).get();
```

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/account/MirrorEvmContractAliases.java
IvanKavaldzhiev: I implemented the second approach. It would really provide us with performance benefits and it simplified the logic even more.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/account/MirrorEvmContractAliases.java
IvanKavaldzhiev: I restored the reading from Store

steven-sheehy (approved):
LGTM

Issue #6449, opened 2023-07-19
Upgrade to Citus 12


steven-sheehy:
### Problem

We're not using the latest Citus.

### Solution

* Upgrade Citus to 12 in test image and stackgres
* Upgrade PostgreSQL to 15.3
* Remove `db/scripts/v2/docker/debian`
* Rename `db/scripts/v2/docker/alpine` to `db/scripts/v2/docker`
* Drop `-alpine` suffix from custom image
* Remove pg_partman from custom Docker image
* Update docs
* Remove `R__maintain_partitions.sql`
* Remove `pg_partman` in favor of [create_time_partitions](https://docs.citusdata.com/en/v11.2/develop/api_udf.html#create-time-partitions)
* Change `V2.0.2__static_partitioning.sql` to use `create_time_partitions` with `cron.schedule`

PR #6512, opened 2023-07-27
Handle Kubernetes upgrades when using ZFS volumes

### LINKED ISSUE #5990: Investigate/Implement solution for GKE when using ZFS

jnels124:
**Description**:
* wait for node to be labeled with needed labels
* wait for disk to be available for attachment
* automatically label nodes with openebs.io/nodeid 
* detect nodes no longer in the cluster and reuse nodeid for new node
* custom csi driver image
* add custom service account for zfs driver
* update docs to create cluster with identity pool

**Related issue(s)**:

Fixes #5990 

**Notes for reviewer**:

**Checklist**


charts/hedera-mirror-common/values.yaml
```diff
@@ -430,6 +437,42 @@ zfs:
...
+            value: /root
+          - name: DISK_SIZE_COORDINATOR
+            value: 128GB
+          - name: DISK_SIZE_WORKER
+            value: 256GB
+        securityContext:
+          privileged: true
+        volumeMounts:
+          - name: root-mount
+            mountPath: /root
```
jnels124, line 473: should be /node

charts/hedera-mirror-common/values.yaml
```diff
@@ -438,3 +481,9 @@ zfs:
         key: zfs
         operator: Equal
         value: "true"
+  zfsPlugin:
+    image:
+      registry: docker.io/
+      repository: jesseswirldslabs/zfs-driver
```
jnels124, line 487: will push to gcr.io/mirrornode

charts/hedera-mirror-common/templates/zfs/configmap-init.yaml
```diff
@@ -7,23 +7,146 @@ metadata:
...
+      kubectl patch configmap $CONFIG_MAP_NAME \
+      -n $K8_NAMESPACE \
+      --type merge \
+      -p "${MODIFIED_CONFIG_MAP}"
+      COMMAND_STATUS=$?
+    done
+    
+    DISK_PREFIX="$(sed 's/^.*_\([^_]*\)$/\1/' <<< "$(kubectl config view --minify -o jsonpath='{.clusters[].name}')")"
+    DISK_NAME="${DISK_PREFIX}-${NODE_ID}-zfs"
+    echo "before print disk check"
```
jnels124, line 148: remove

charts/hedera-mirror-common/templates/zfs/rbac.yaml
```diff
@@ -0,0 +1,56 @@
...
+  labels: {{ include "hedera-mirror-common.labels" . | nindent 4 }}
+  name: configmap-updater
+  namespace: {{ include "hedera-mirror-common.namespace" . }}
+rules:
+  - apiGroups: [""]
+    resources: ["configmaps"]
+    verbs: ["get", "list", "put", "patch", "update", "watch"]
+  - apiGroups: [""]
+    resources: ["pods"]
+    verbs: ["get", "list", "delete", "watch"]
```
jnels124, line 40: can remove pods perms. This account doesn't need this anymore

charts/hedera-mirror-common/values.yaml
```diff
@@ -418,18 +418,31 @@ zfs:
...
+      name: zfs-service-account
+      iAmName: sa-name@project.iam.gserviceaccount.com
   parameters:
     compression: zstd-6
     fstype: zfs
     poolname: zfspv-pool
     recordsize: 32k
   worker:
     diskSize: 3200GB
+  zfsController:
```
jnels124, line 438: will remove

charts/hedera-mirror-common/templates/zfs/serviceaccount.yaml
```diff
@@ -0,0 +1,9 @@
+{{- if .Values.zfs.enabled -}}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  annotations:
```
jnels124, line 5: will add common labels

docs/database/zfs.md
```diff
@@ -99,24 +94,3 @@ find the corresponding zvolume `kubectl get zv --all-namespaces` then you may de
 using `kubectl delete zv -n <namespace> <zVolumeName>`. This will do the cleanup on the node itself and make the space
 available in the zpool again.
 
-### GKE Cluster Upgrade
```
jnels124, line 102: There will be a follow up ticket to address this section. With the stackgres deployment, upgrade path is currently unknown.

charts/hedera-mirror-common/templates/zfs/configmap-init.yaml
```diff
@@ -43,9 +165,21 @@ data:
...
-    if (! gcloud compute instances describe "${NODE_NAME}" --zone "${ZONE}" --format '(disks[].source)' | grep "${DISK_NAME}" > /dev/null); then
-        echo "Attaching ${DISK_NAME} to ${NODE_NAME}"
-        gcloud compute instances attach-disk "${NODE_NAME}" --device-name=sdb --disk "${DISK_NAME}" --zone "${ZONE}" --quiet
+    if (! gcloud compute instances describe "${NODE_NAME}" --zone "${ZONE}" --format '(disks[].source)' | grep "${DISK_NAME}" > /dev/null); then        
+        ATTACH_ATTEMPTS=0
+        COMMAND_STATUS=1
+        until [[ ${COMMAND_STATUS} -eq 0 || ${ATTACH_ATTEMPTS} -ge 13 ]]; do
+          echo "Attaching ${DISK_NAME} to ${NODE_NAME} attempts ${ATTACH_ATTEMPTS}" 
+          gcloud compute instances attach-disk "${NODE_NAME}" --device-name=sdb --disk "${DISK_NAME}" --zone "${ZONE}"
+          COMMAND_STATUS=$?
```
jnels124, line 174: will restructure slightly so not waiting 30 seconds in successful cases

charts/hedera-mirror-common/templates/zfs/configmap-init.yaml
```diff
@@ -7,23 +7,150 @@ metadata:
...
+      sleep 10
+    done
+    # Determine correct config for zone
+    CONFIG_MAP_NAME="zfs-node-status"
+    COMMAND_STATUS=1
+    until [[ ${COMMAND_STATUS} -eq 0 ]]; do
+      CONFIG_MAP="$(kubectl get configmap -o json -n ${K8_NAMESPACE} ${CONFIG_MAP_NAME})"
+      MODIFIED_CONFIG_MAP="$(echo "${CONFIG_MAP}" |jq '{metadata: {resourceVersion: (.metadata.resourceVersion)}, data: {"zfs-node-status.json": (.data["zfs-node-status.json"]|fromjson)}}')"
+      ZONE_CONFIG="$(echo "${MODIFIED_CONFIG_MAP}" |jq  --arg ZONE "${ZONE}" --arg CITUS_ROLE "${CITUS_ROLE}" '.data["zfs-node-status.json"]["\($CITUS_ROLE)"]["\($ZONE)"] // []')"
+      NODE_LIST="$(kubectl get nodes -o json)"
```
steven-sheehy, line 76: We should only care about zfs related nodes being drained, right? Can add `-lcsi-type=zfs`.

charts/hedera-mirror-common/templates/zfs/configmap-init.yaml
```diff
@@ -7,23 +7,150 @@ metadata:
...
+      NODE_NAMES="$(echo "${NODE_LIST}" | jq '.items | map(. | {(.metadata.name): {isUpgrade: (.metadata.labels["operation.gke.io/type"] == "drain")}}) | add')"
+      
+      ## This will find all nodes that have been initialized by this script in the past
+      ## and generate its own version of the config map configuration to 
+      ZFS_ZONE_NODES=$(echo "${NODE_LIST}" | jq \
+                      --arg ZONE "${ZONE}" \
+                      --arg CITUS_ROLE "${CITUS_ROLE}" \
+                      --arg NODE_ID_LABEL "${NODE_ID_LABEL}" \
+                      --arg ROLE_LABEL "${ROLE_LABEL}" \
+                      '.items | map ( . | select(.metadata.labels["topology.kubernetes.io/zone"] == "\($ZONE)" and
```
steven-sheehy, line 86: Not sure I understand jq enough and what's going on here to review this. The cognitive complexity of this script is quite high. We definitely need to move this logic to the zfs-controller soon. Let's initiate that conversation soon to see if they're willing.

steven-sheehy (changes requested):

charts/hedera-mirror-common/templates/zfs/configmap-init.yaml
```diff
@@ -7,23 +7,150 @@ metadata:
...
+      if [[ ${ATTEMPT_COUNT} -ge 13 ]]
+      then
+        echo "Timed out waiting for labels please set ${ROLE_LABEL}"
+        exit 1
+      fi
+      ATTEMPT_COUNT=$((ATTEMPT_COUNT+1))
+      echo "Retrying label GET attempt: ${ATTEMPT_COUNT}"
+      NODE_LABELS="$(kubectl get node ${NODE_NAME} -o json | jq '.metadata.labels')"
+      CITUS_ROLE="$(echo ${NODE_LABELS} | jq --arg ROLE_LABEL "${ROLE_LABEL}" -r '.metadata.labels["\($ROLE_LABEL)"]')"
+      sleep 10
```
xin-hedera, line 67: unnecessary 10s sleep when attempt count has reached the max, same for the other loop with attempt count

xin-hedera (changes requested):

charts/hedera-mirror-common/templates/zfs/configmap-init.yaml
jnels124: I agree the script is complex and very much less than ideal. Once this PR is merged, I will link to the issue i created upstream and see if they would be willing to accept a PR in the plugin code that followed similar logic.

charts/hedera-mirror-common/templates/zfs/configmap-init.yaml
```diff
@@ -7,23 +7,150 @@ metadata:
...
+    done
+    # Determine correct config for zone
+    CONFIG_MAP_NAME="zfs-node-status"
+    COMMAND_STATUS=1
+    until [[ ${COMMAND_STATUS} -eq 0 ]]; do
+      CONFIG_MAP="$(kubectl get configmap -o json -n ${K8_NAMESPACE} ${CONFIG_MAP_NAME})"
+      MODIFIED_CONFIG_MAP="$(echo "${CONFIG_MAP}" |jq '{metadata: {resourceVersion: (.metadata.resourceVersion)}, data: {"zfs-node-status.json": (.data["zfs-node-status.json"]|fromjson)}}')"
+      ZONE_CONFIG="$(echo "${MODIFIED_CONFIG_MAP}" |jq  --arg ZONE "${ZONE}" --arg CITUS_ROLE "${CITUS_ROLE}" '.data["zfs-node-status.json"]["\($CITUS_ROLE)"]["\($ZONE)"] // []')"
```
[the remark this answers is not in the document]
jnels124: jq was giving compilation errors without the quotes added. Let me do a quick test and verify if they are needed.

charts/hedera-mirror-common/templates/zfs/configmap-init.yaml
jnels124: I can apply that filter earlier

charts/hedera-mirror-common/templates/zfs/configmap-init.yaml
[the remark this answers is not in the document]
jnels124: Don't need to escape this any longer. It was a learning issue with jq when I started this. Thanks for pointing this out.

charts/hedera-mirror-common/templates/zfs/configmap-init.yaml
jnels124: fixed this by moving some things

charts/hedera-mirror-common/templates/zfs/configmap-init.yaml
steven-sheehy: I don't see that this was added?

jnels124: Instead, I just moved the citus-role and zone filters up. Technically this would still catch citus nodes that may exist and are not using zfs(i.e csi-type != zfs) however they will not effect the resulting calculations for calculating zone configurations because these nodes will never be included in the config map nor will they have the appropriate labels on the node (generated-node-id)

charts/hedera-mirror-common/templates/zfs/rbac.yaml
```diff
@@ -0,0 +1,53 @@
+{{- if .Values.zfs.enabled -}}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels: {{ include "hedera-mirror-common.labels" . | nindent 4 }}
+  name: label-reader
```
steven-sheehy, line 6:
```suggestion
  name: zfs-label-reader
```

charts/hedera-mirror-common/templates/zfs/rbac.yaml
```diff
@@ -0,0 +1,53 @@
...
+rules:
+  - apiGroups: [""]
+    resources: ["nodes"]
+    verbs: ["get", "list", "update", "patch"]
+---
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  labels: {{ include "hedera-mirror-common.labels" . | nindent 4 }}
+  name: label-reader
```
steven-sheehy, line 17:
```suggestion
  name: zfs-label-reader
```

charts/hedera-mirror-common/templates/zfs/rbac.yaml
```diff
@@ -0,0 +1,53 @@
...
+roleRef:
+  kind: ClusterRole
+  name: label-reader
+  apiGroup: rbac.authorization.k8s.io
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  labels: {{ include "hedera-mirror-common.labels" . | nindent 4 }}
+  name: configmap-updater
```
steven-sheehy, line 32: The namespace level RBAC should include both the release name and `zfs` in the name

```suggestion
  name: {{ include "hedera-mirror-common.fullname" . }}-zfs-configmap-updater
```

charts/hedera-mirror-common/templates/zfs/configmap-zfsnodestatus.yaml
```diff
@@ -0,0 +1,11 @@
+{{- if .Values.zfs.enabled -}}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  labels: {{ include "hedera-mirror-common.labels" . | nindent 4 }}
+  name: zfs-node-status
```
steven-sheehy, line 6:
```suggestion
  name: {{ include "hedera-mirror-common.fullname" . }}-zfs-node-status
```

steven-sheehy (changes requested):

charts/hedera-mirror-common/templates/zfs/rbac.yaml
jnels124: makes sense. Thanks for pointing this out.

charts/hedera-mirror-common/templates/zfs/configmap-init.yaml
steven-sheehy: Right, I understand it might logically be the same since you filter in memory, but isn't it more performant and safer to filter out app nodes upfront via the API? It reduces that amount of JSON returned and reduces the likelihood we may accidentally patch an app node or include it in the configmap.

charts/hedera-mirror-common/templates/zfs/configmap-init.yaml
```diff
@@ -7,29 +7,155 @@ metadata:
...
+    until [[ ${COMMAND_STATUS} -eq 0 ]]; do
+      CONFIG_MAP="$(kubectl get configmap -o json -n ${K8S_NAMESPACE} ${CONFIG_MAP_NAME})"
+      MODIFIED_CONFIG_MAP="$(echo "${CONFIG_MAP}" |jq '{metadata: {resourceVersion: (.metadata.resourceVersion)}, data: {"zfs-node-status.json": (.data["zfs-node-status.json"]|fromjson)}}')"
+      ZONE_CONFIG="$(echo "${MODIFIED_CONFIG_MAP}" |jq  --arg ZONE "${ZONE}" --arg CITUS_ROLE "${CITUS_ROLE}" '.data["zfs-node-status.json"][$CITUS_ROLE][$ZONE] // []')"
+      NODE_LIST="$(kubectl get nodes -o json | jq \
+                   --arg ZONE "${ZONE}" \
+                   --arg CITUS_ROLE "${CITUS_ROLE}" \
+                   --arg ROLE_LABEL "${ROLE_LABEL}" \
+                   '.items | map(select(.metadata.labels["topology.kubernetes.io/zone"] == $ZONE and
+                                        .metadata.labels[$ROLE_LABEL] == $CITUS_ROLE))')"
```
steven-sheehy, line 81: We should use kubectl selectors to filter on the server where we can. Also seems like we only need the metadata so can return a subset.

```suggestion
      NODE_LIST="$(kubectl get nodes -lcsi-type=zfs,topology.kubernetes.io/zone=${ZONE},${ROLE_LABEL}=${CITUS_ROLE} -o jsonsonpath='{.items[*].metadata}')"
```

steven-sheehy (changes requested):

charts/hedera-mirror-common/templates/zfs/configmap-init.yaml
```diff
@@ -7,29 +7,155 @@ metadata:
...
+    APISERVER="https://kubernetes.default.svc"
+    SERVICEACCOUNT="/var/run/secrets/kubernetes.io/serviceaccount"
+    TOKEN="$(cat ${SERVICEACCOUNT}/token)"
+    CACERT="${SERVICEACCOUNT}/ca.crt"
+    
+    kubectl config set-cluster ctc --server="${APISERVER}" --certificate-authority="${CACERT}"
+    kubectl config set-context ctc --cluster=ctc
+    kubectl config set-credentials user --token="${TOKEN}"
+    kubectl config set-context ctc --user=user
+    kubectl config use-context ctc
```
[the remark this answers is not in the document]
jnels124: I will try again but initially when I tried I was unable to connect.

charts/hedera-mirror-common/templates/zfs/configmap-init.yaml
jnels124: It would result in less data being returned in cases where we have nodes with a citus-role but not csi-type=zfs which isn't really planned to be the case. I can add the csi-type filter as well. Was mainly just left off to not have more jq logic than necessary but since you have provided other input on moving these filters out of the jq script and into kubectl args I will add this one as well.

+1 on jsonpath. Haven't worked with that before and have always just used jq.

charts/hedera-mirror-common/values.yaml
```diff
@@ -438,3 +443,33 @@ zfs:
         key: zfs
         operator: Equal
         value: "true"
+    additionalVolumes:
+      node:
+        hostPath:
+          path: /
```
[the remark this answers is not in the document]
jnels124: I noticed the error in the output but everything installed / worked correctly so assumed it was just due to how it was defined in the default values upstream

charts/hedera-mirror-common/templates/zfs/configmap-init.yaml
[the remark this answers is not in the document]
jnels124: You are correct it does work without this.

charts/hedera-mirror-common/templates/zfs/configmap-init.yaml
```diff
@@ -7,29 +7,155 @@ metadata:
...
+      then
+        echo "Timed out waiting for labels please set ${ROLE_LABEL}"
+        exit 1
+      fi
+      sleep 10
+      ATTEMPT_COUNT=$((ATTEMPT_COUNT+1))
+      echo "Retrying label GET attempt: ${ATTEMPT_COUNT}"
+      NODE_LABELS="$(kubectl get node ${NODE_NAME} -o json | jq '.metadata.labels')"
+      CITUS_ROLE="$(echo ${NODE_LABELS} | jq --arg ROLE_LABEL "${ROLE_LABEL}" -r '.metadata.labels[$ROLE_LABEL]')"
+    done
```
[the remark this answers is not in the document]
steven-sheehy: Unaddressed?

charts/hedera-mirror-common/values.yaml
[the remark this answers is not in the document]
steven-sheehy: You changed it to the syntax in this comment that I was just highlighting as the ideal syntax. This syntax will not work though. You need to change to the syntax in the two comments directly above this.

charts/hedera-mirror-common/templates/zfs/configmap-init.yaml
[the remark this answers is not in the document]
steven-sheehy: It seems like we don't have `CITUS_ROLE` here and the syntax supported in the wait command is very limited. It doesn't support `!=""` or using `=` in the jsonpath. So we can leave as is.

charts/hedera-mirror-common/templates/zfs/configmap-init.yaml
jnels124: This logic has been updated to use -l csi-type=zfs on the kubectl command

charts/hedera-mirror-common/values.yaml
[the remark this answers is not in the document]
jnels124: The preferred syntax you mentioned does work correctly. However, the syntax you suggested to change to produces errors. Here is the description of the mounts when using the preferred syntax.
Screenshot 2023-08-06 at 12 16 03 PM

docs/database/zfs.md
jnels124: This issue will be resolved by #6449 with the removal of pg partman. A small note will be added to that pr to indicate resources to check pre/post upgrade

charts/hedera-mirror-common/values.yaml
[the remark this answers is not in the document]
steven-sheehy: You're right, it's a map not an array. I expected it to be an array since that's what the `values.yaml` had upstream but as you noted those defaults are wrong. The default upstream should be changed to `{}` and a commented out example added how to supply a value be added.

steven-sheehy (approved):
LGTM

xin-hedera (approved):
LGTM

PR #6557, opened 2023-08-01
HIP-584: Handle read only precompile calls via ViewExecutor/RedirectViewExecutor

IvanKavaldzhiev:
**Description**:

This PR allows all type of calls (view/pure/dynamic) to read only precompiles to be handled via ViewExecutor/RedirectViewExecutor, thus removing the need of having a copy of separate precompile class implementations.

That is possible since the executors read the data from the same place (the Store interface) and they share the same gas calculation logic with the precompile class implementations in `hedera-services` codebase. Consequently, we could just reuse the executor logic to handle `eth_call` and `eth_estimateGas` requests.
**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


0xivanov (approved):
LGTM

bilyana-gospodinova (approved):

agadzhalov (approved):
LGTM

stoyanov-st (approved):
LGTM

kselveliev (approved):

steven-sheehy (approved):

PR #6563, opened 2023-08-01
Suppress protobufjs false positive

steven-sheehy:
**Description**:

Suppress `protobuf.js` false positive since we upgraded to the version in the CVE that has the fix but dependency check plugin still marks it as vulnerable.

**Related issue(s)**:

**Notes for reviewer**:

**Checklist**


edwin-greene (approved):
Looks good

PR #6565, opened 2023-08-01
Rerun conditional migrations

### LINKED ISSUE #6397: Ensure time sensitive migrations run again during parser
### Problem

Database migrations run on startup, but new mirror nodes can startup with older start dates. If the migration corrects data that it has yet to download then the user must somehow know that they should re-run these migrations later once they've processed that data.

### Solution

For each of the following migrations:
* `SyntheticCryptoTransferApprovalMigration`
* `SyntheticNftAllowanceOwnerMigration`
* `SyntheticTokenAllowanceOwnerMigration`

Do the following:
* Implement `RecordStreamFileListener.onEnd()`
* Skip for non-mainnet networks
* Skip if record file has not transitioned to a HAPI version that fixed the issue
* Determine if the migration can be changed to run async

For each of the following migrations:
* `InitializeEntityBalanceMigration`
* `TokenAccountBalanceMigration`

Do the following:
* Implement `BalanceStreamFileListener.onEnd()`
* Skip if already ran since last startup
* Skip if balance table not empty

### Alternatives

Re-implement the migration logic directly when processing each record file

mgoelswirlds:
**Description**:
This PR is created to address re-running certain time sensitive migrations, for newer mirror nodes.

This PR modifies:
* SyntheticCryptoTransferApprovalMigration to rerun when the HAPI version changes to 0.38.10 or above.
* SyntheticNftAllowanceOwnerMigration and SyntheticTokenAllowanceOwnerMigration to rerun when the HAPI version changes to 0.37.0 or above.
* InitializeEntityBalanceMigration and TokenAccountBalanceMigration to rerun, when the first account balance file comes in.

**Related issue(s)**:

Fixes #6397

**Notes for reviewer**:

**Checklist**


hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/balance/AccountBalanceFileParser.java
```diff
@@ -121,5 +127,17 @@ protected void doParse(AccountBalanceFile accountBalanceFile) {
...
+
+        last.compareAndSet(lastBalanceFile, accountBalanceFile);
+    }
+
+    @Nullable
+    private AccountBalanceFile publishInitializeAccountBalanceEvent() {
+        var lastBalanceFile = last.get();
+        // If no account balance file has been processed since last startup.
+        if (lastBalanceFile == null && lastFromDb.get() == null) {
+            applicationEventPublisher.publishEvent(new InitializeBalanceEvent(this));
```
mgoelswirlds, line 139: If this is the first account balance file being processed and the DB is empty then publishing InitializeBalanceEvent which will rerun balance migrations.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/balance/AccountBalanceFileParser.java
```diff
@@ -121,5 +127,17 @@ protected void doParse(AccountBalanceFile accountBalanceFile) {
         accountBalanceFile.setLoadEnd(loadEnd.getEpochSecond());
         streamFileListener.onEnd(accountBalanceFile);
         streamFileRepository.save(accountBalanceFile);
+
+        last.compareAndSet(lastBalanceFile, accountBalanceFile);
+    }
+
+    @Nullable
+    private AccountBalanceFile publishInitializeAccountBalanceEvent() {
```
mgoelswirlds, line 135: Moved this into another function to reduce cognitive complexity.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/SyntheticCryptoTransferApprovalMigration.java
```diff
@@ -140,6 +147,29 @@ with contractresults as (
...
+        // Not Applicable
+    }
+
+    @Override
+    public void onEnd(RecordFile streamFile) throws ImporterException {
+        if (streamFile == null) {
+            return;
+        }
+
+        /*if(shouldRerun(streamFile,HAPI_VERSION_0_38_10,recordFileRepository)) {
```
mgoelswirlds, line 163: These comments will be removed once we Merge Xin's changes to have async migration extend repeatable migration.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/SyntheticNftAllowanceOwnerMigration.java
```diff
@@ -107,9 +115,13 @@ insert into nft_allowance (approved_for_all, owner, payer_account_id, spender, t
     private final JdbcTemplate jdbcTemplate;
 
     @Lazy
-    public SyntheticNftAllowanceOwnerMigration(@Owner JdbcTemplate jdbcTemplate, MirrorProperties mirrorProperties) {
```
mgoelswirlds, line 110: This migration took 779 ms on mainnet.So not converting it to Async.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/SyntheticTokenAllowanceOwnerMigration.java
```diff
@@ -17,15 +17,23 @@
...
+import com.hedera.mirror.importer.repository.RecordFileRepository;
 import jakarta.inject.Named;
 import org.flywaydb.core.api.MigrationVersion;
 import org.springframework.context.annotation.Lazy;
+import org.springframework.data.util.Version;
 import org.springframework.jdbc.core.JdbcTemplate;
 
 @Named
-public class SyntheticTokenAllowanceOwnerMigration extends RepeatableMigration {
+public class SyntheticTokenAllowanceOwnerMigration extends RepeatableMigration implements RecordStreamFileListener {
```
mgoelswirlds, line 33: This migration took 10158 ms on mainnet.So not converting it to Async.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/SyntheticCryptoTransferApprovalMigration.java
```diff
@@ -140,6 +147,32 @@ with contractresults as (
...
+
+    @Override
+    public void onEnd(RecordFile streamFile) throws ImporterException {
+        if (streamFile == null) {
+            return;
+        }
+
+        if ((streamFile.getHapiVersion()).isGreaterThanOrEqualTo(HAPI_VERSION_0_38_10)) {
+            var latestFile = recordFileRepository.findLatestWithOffset(1).orElse(null);
+            if (latestFile != null && latestFile.getHapiVersion().isLessThan(HAPI_VERSION_0_38_10)) {
```
xin-hedera, line 165:
```suggestion
            var latestFile = recordFileRepository.findLatestWithOffset(1);
            if (latestFile.filter(f -> f.getHapiVersion().isLessThan(HAPI_VERSION_0_38_10)).isPresent()) {
```

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/SyntheticNftAllowanceOwnerMigration.java
```diff
@@ -17,15 +17,23 @@
...
 import jakarta.inject.Named;
 import org.flywaydb.core.api.MigrationVersion;
 import org.springframework.context.annotation.Lazy;
+import org.springframework.data.util.Version;
 import org.springframework.jdbc.core.JdbcTemplate;
 
 @Named
-public class SyntheticNftAllowanceOwnerMigration extends RepeatableMigration {
+public class SyntheticNftAllowanceOwnerMigration extends RepeatableMigration implements RecordStreamFileListener {
+    // 779
```
xin-hedera, line 34: leftover?

```suggestion
```

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/MigrationRecordStreamFileListener.java
```diff
@@ -0,0 +1,26 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hedera.mirror.importer.parser.record;
+
+import org.springframework.data.util.Version;
+
+public interface MigrationRecordStreamFileListener {
```
xin-hedera, line 21: where is this interface used?

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/InitializeEntityBalanceMigration.java
```diff
@@ -87,4 +89,12 @@ protected void doMigrate() {
         int count = jdbcOperations.update(INITIALIZE_ENTITY_BALANCE_SQL);
         log.info("Initialized {} entities balance in {}", count, stopwatch);
     }
+
+    @EventListener
+    public void reRunMigration(InitializeBalanceEvent event) {
```
xin-hedera, line 94: we don't want to run it when the transaction in which the event is published fails

```suggestion
   @TransactionalEventListener(classes = InitializeBalanceEvent.class)
    public void reRunMigration() {
```

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/InitializeEntityBalanceMigration.java
```diff
@@ -87,4 +89,12 @@ protected void doMigrate() {
...
+        log.info(
+                "Re-running the InitializeEntityBalanceMigration on InitializeBalanceEvent created at {}",
+                event.getTimestamp());
```
xin-hedera, line 97: don't see much value logging the event's created timestamp
```suggestion
        log.info("Re-running the InitializeEntityBalanceMigration on InitializeBalanceEvent");
```

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/balance/AccountBalanceFileParser.java
```diff
@@ -84,6 +89,7 @@ protected void doParse(AccountBalanceFile accountBalanceFile) {
         DateRangeFilter filter = mirrorDateRangePropertiesProcessor.getDateRangeFilter(StreamType.BALANCE);
         int batchSize = ((BalanceParserProperties) parserProperties).getBatchSize();
         long count = 0L;
+        AccountBalanceFile lastBalanceFile = publishInitializeAccountBalanceEvent();
```
xin-hedera, line 92: why does the function needs to return the balance file?

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/balance/AccountBalanceFileParser.java
```diff
@@ -121,5 +127,17 @@ protected void doParse(AccountBalanceFile accountBalanceFile) {
         accountBalanceFile.setLoadEnd(loadEnd.getEpochSecond());
         streamFileListener.onEnd(accountBalanceFile);
         streamFileRepository.save(accountBalanceFile);
+
+        last.compareAndSet(lastBalanceFile, accountBalanceFile);
+    }
+
+    @Nullable
+    private AccountBalanceFile publishInitializeAccountBalanceEvent() {
+        var lastBalanceFile = last.get();
```
xin-hedera, line 136: make it return void, then the logic is simply

```java
  if (last.get() == null) {
    publish event;
  }
```

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/balance/AccountBalanceFileParser.java
```diff
@@ -121,5 +127,17 @@ protected void doParse(AccountBalanceFile accountBalanceFile) {
         accountBalanceFile.setLoadEnd(loadEnd.getEpochSecond());
         streamFileListener.onEnd(accountBalanceFile);
         streamFileRepository.save(accountBalanceFile);
+
+        last.compareAndSet(lastBalanceFile, accountBalanceFile);
```
xin-hedera, line 131: two issues

- why do we need to do `compareAndSet` instead of just `set`?
- this won't work when transaction fails, for instance, when acount_balance_file table is empty
  - the event gets published
  - parse the first account balance file
  - update last
  - transaction fails
  - re-parse the same account balance file
  - no event publish since last.get() <> null

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/balance/AccountBalanceFileParser.java
```diff
@@ -121,5 +127,17 @@ protected void doParse(AccountBalanceFile accountBalanceFile) {
...
+    }
+
+    @Nullable
+    private AccountBalanceFile publishInitializeAccountBalanceEvent() {
+        var lastBalanceFile = last.get();
+        // If no account balance file has been processed since last startup.
+        if (lastBalanceFile == null) {
+            applicationEventPublisher.publishEvent(new InitializeBalanceEvent(this));
+        }
+        return lastBalanceFile;
```
xin-hedera, line 141: with the pitfalls, perhaps we should

- revert all `last` related changes here and in `AbstractStreamFileParser`
- change `InitializeEntityBalanceMigration` to implement `BalanceStreamFileListener`
- add a `CompositeBalanceStreamFileListener` as the primary bean
- In `InitializeEntityBalanceMigration.onEnd`, publish the event if `account_balance_file` table is empty
- add a `on***Event` method to `InitializeEntityBalanceMigration` with the `TransactionalEventListener` annotation to trigger the migration

xin-hedera (changes requested):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/MigrationRecordStreamFileListener.java
mgoelswirlds: Leftover. will remove it.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/SyntheticCryptoTransferApprovalMigration.java
```diff
@@ -140,6 +147,32 @@ with contractresults as (
...
+    public void onStart() throws ImporterException {
+        // Not Applicable
+    }
+
+    @Override
+    public void onEnd(RecordFile streamFile) throws ImporterException {
+        if (streamFile == null) {
+            return;
+        }
+
```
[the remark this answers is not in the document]
mgoelswirlds: For point number 2 , that's the reason I had it initially, but while some cleanup removed it. Will add it back.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/SyntheticTokenAllowanceOwnerMigration.java
xin-hedera: I doubt it can work in environment where we owner and regular user are different, e.g., `mirror_node` and `mirror_importer`.

The migration gets injected the owner `jdbcTemplate` bean because it delete rows. in some environment, `mirror_importer` may not be able to delete rows.

When the migration runs inside `onEnd`, it's in the same database transaction, and the `connection` in the thread local storage uses regular user `mirror_importer`. The migration most likely uses the same connection, and may fail to delete rows.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/InitializeEntityBalanceMigration.java
```diff
@@ -87,4 +103,17 @@ protected void doMigrate() {
...
+    @TransactionalEventListener(classes = InitializeEntityBalanceEvent.class)
+    public void reRunMigration() {
+        log.info("Re-running the InitializeEntityBalanceMigration on InitializeBalanceEvent");
+        doMigrate();
+    }
+
+    @Override
+    public void onEnd(AccountBalanceFile streamFile) throws ImporterException {
+        if (accountBalanceFileRepository.count() == 0) {
+            applicationEventPublisher.publishEvent(new InitializeEntityBalanceEvent(this));
```
steven-sheehy, line 116: I don't see the point of having the logic split into separate `onEnd()` and a `reRunMigration()`. We don't need the decoupling that Spring events provides if we're invoking it within the same class and it's not async. We should just remove the publisher and `reRunMigration` and execute the migration here in `onEnd()`.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/InitializeEntityBalanceMigration.java
```diff
@@ -87,4 +103,17 @@ protected void doMigrate() {
...
+
+    @TransactionalEventListener(classes = InitializeEntityBalanceEvent.class)
+    public void reRunMigration() {
+        log.info("Re-running the InitializeEntityBalanceMigration on InitializeBalanceEvent");
+        doMigrate();
+    }
+
+    @Override
+    public void onEnd(AccountBalanceFile streamFile) throws ImporterException {
+        if (accountBalanceFileRepository.count() == 0) {
```
steven-sheehy, line 115: A count query in postgresql is extremely slow and should never be used. You should be able to use the existing repository query to find the earliest row.
```suggestion
        if (accountBalanceFileRepository.findNextInRange(0, Long.MAX_VALUE).isEmpty()) {
```

steven-sheehy, line 115: We should optimize this so the query is only executed for the first stream file it's seen. This is different than just storing an atomic boolean that the migration has executed since it could've executed but rolled back.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/SyntheticCryptoTransferApprovalMigration.java
```diff
@@ -27,17 +31,22 @@
...
 import org.springframework.jdbc.core.DataClassRowMapper;
 import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
 import org.springframework.transaction.support.TransactionOperations;
 
 @Named
-public class SyntheticCryptoTransferApprovalMigration extends AsyncJavaMigration<Long> {
+public class SyntheticCryptoTransferApprovalMigration extends AsyncJavaMigration<Long>
+        implements RecordStreamFileListener {
 
+    public static final Version HAPI_VERSION_0_38_10 = new Version(0, 38, 10);
```
steven-sheehy, line 48: There's no `0.38.10`. And we should use the lowest patch release the fix was in. It might've been in all `0.38` releases so we should use `0.38.0`, otherwise we pick the exact version.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/SyntheticCryptoTransferApprovalMigration.java
```diff
@@ -27,17 +31,22 @@
...
+    private static final AtomicBoolean reRanMigration = new AtomicBoolean(false);
```
steven-sheehy, line 49: nit: It's redundant to have the word `migration` in the variable name.

```suggestion
    private static final AtomicBoolean executed = new AtomicBoolean(false);
```

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/SyntheticCryptoTransferApprovalMigration.java
```diff
@@ -141,6 +150,24 @@ with contractresults as (
...
 
+    private final RecordFileRepository recordFileRepository;
+
+    @Override
+    public void onEnd(RecordFile streamFile) throws ImporterException {
+        if (streamFile == null) {
+            return;
+        }
+
+        if ((streamFile.getHapiVersion()).isGreaterThanOrEqualTo(HAPI_VERSION_0_38_10) && !reRanMigration.get()) {
```
steven-sheehy, line 161: There's an extra parenthesis and the boolean check is not atomic, allowing multiple async processes to start.

```suggestion
        if (streamFile.getHapiVersion().isGreaterThanOrEqualTo(HAPI_VERSION_0_38_10) && executed.compareAndSet(false, true)) {
```

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/SyntheticCryptoTransferApprovalMigration.java
```diff
@@ -141,6 +150,24 @@ with contractresults as (
...
+            var lastRecordFile = recordFileRepository.findLatestWithOffset(1);
+            var latestFile = lastRecordFile.orElse(null);
+            if (latestFile != null && latestFile.getHapiVersion().isLessThan(HAPI_VERSION_0_38_10)) {
```
steven-sheehy, line 164: Use optional chaining for all migrations like Xin [mentioned](https://github.com/hashgraph/hedera-mirror-node/pull/6565#discussion_r1284658561).

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/SyntheticCryptoTransferApprovalMigration.java
```diff
@@ -141,6 +150,24 @@ with contractresults as (
...
+                migrateAsync();
+                reRanMigration.set(true);
```
steven-sheehy, line 166: Remove after switching to compareAndSet

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/AbstractStreamFileParser.java
steven-sheehy, line 1: Revert changes to this file if we're using `onEnd` approach

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/balance/CompositeBalanceStreamFileListener.java
```diff
@@ -0,0 +1,57 @@
...
+        }
+    }
+
+    @Override
+    public void onError() {
+        onEach(StreamFileListener::onError);
+    }
+
+    private void onEach(Consumer<BalanceStreamFileListener> consumer) {
+        for (var listener : listeners) {
```
steven-sheehy, line 53: Use indexed based for loop for performance

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/balance/CompositeBalanceStreamFileListener.java
```diff
@@ -0,0 +1,57 @@
...
+    @Override
+    public void onStart() throws ImporterException {
+        onEach(StreamFileListener::onStart);
+    }
+
+    @Override
+    public void onEnd(AccountBalanceFile streamFile) throws ImporterException {
+        for (var listener : listeners) {
+            listener.onEnd(streamFile);
+        }
```
steven-sheehy, line 44:
```suggestion
        onEach(StreamFileListener::onEnd);
```

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/balance/InitializeEntityBalanceEvent.java
steven-sheehy, line 1: Remove both events after previous suggestions.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/RecordFileParser.java
steven-sheehy, line 1: Revert

steven-sheehy (changes requested):

steven-sheehy:
Note: Any comments on one migration can apply to all. I didn't duplicate the comments for each.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/SyntheticCryptoTransferApprovalMigration.java
mgoelswirlds: I will change it to 38.0, but this was the comment  /`/ This problem was fixed by services release 0.38.10, this is last timestamp before that release`
in the code while writing the migration. But I see that there are releases till 0.38.6.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/SyntheticCryptoTransferApprovalMigration.java
mgoelswirlds: Yes I missed that comment.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/SyntheticCryptoTransferApprovalMigration.java
steven-sheehy: Looking in the DB, I do see a 10 for `hapi_version_patch`. So your logic is correct. But we should check with services cause that version number seems like a bug.

steven-sheehy: I asked Michael, still awaiting a response.

I checked the upstream [fix](https://github.com/hashgraph/hedera-services/commit/5d9009307de098695c44414cb3ca680fd68b5f78) and it was in all 0.38 tags. Thus I think since we have a range check we can safely use `0.38.0`.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/InitializeEntityBalanceMigration.java
mgoelswirlds: So even with just the onEnd approach we can ensure that the migration will run once and if the transaction is rolled back this will re-run when the file is parsed again right?

steven-sheehy: Right, since it will happen synchronously and inherit the existing `AccountBalanceFileParser` transaction. The two migrations also don't need to run as owner so they're fine to run as mirror_importer user like the current transaction.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/InitializeEntityBalanceMigration.java
mgoelswirlds: Other than confirming the DB is empty I don't know of a way to verify that this is the first file.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/SyntheticCryptoTransferApprovalMigration.java
mgoelswirlds: So services did not have a 1:1 mapping at that time. protobuf 0.38.10 went in service 0.38.6

mgoelswirlds: But since i am checking the HAPI version, I should keep it as is.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/InitializeEntityBalanceMigration.java
steven-sheehy: You're not confirming it's the first file ever, you're confirming if it's the first processed by the application after startup to reduce unnecessary queries. Can store an `AtomicLong firstConsensusTimestamp = new AtomicLong(0L)` that you `compareAndSet` with zero. Then only run query if `firstConsensusTimestamp.get() == streamFile.getConsensusTimestamp()`.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/SyntheticCryptoTransferApprovalMigration.java
steven-sheehy: No, change to `0.38.0`. Again it should map to in which services the fix went in.

mgoelswirlds: Ahh ok. Had checked with Neeha and she had mentioned 0.38.6 had the changes but will do 0.38.0

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/SyntheticTokenAllowanceOwnerMigration.java
xin-hedera: did you check if running it inside `onEnd` would actually pick a connection with different credentials other than owner?

this is a blocker for both `SyntheticNftAllowanceOwnerMigration` and `SyntheticTokenAllowanceOwnerMigration`

xin-hedera (changes requested):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/InitializeEntityBalanceMigration.java
```diff
@@ -17,14 +17,22 @@
...
 import org.flywaydb.core.api.MigrationVersion;
 import org.springframework.context.annotation.Lazy;
 import org.springframework.jdbc.core.JdbcOperations;
 
 @Named
-public class InitializeEntityBalanceMigration extends RepeatableMigration {
+public class InitializeEntityBalanceMigration extends RepeatableMigration implements BalanceStreamFileListener {
+
+    private final AccountBalanceFileRepository accountBalanceFileRepository;
+    AtomicLong firstConsensusTimestamp = new AtomicLong(0L);
```
steven-sheehy, line 35:
```suggestion
    private final AtomicLong firstConsensusTimestamp = new AtomicLong(0L);
```

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/SyntheticNftAllowanceOwnerMigration.java
```diff
@@ -17,15 +17,25 @@
...
 import org.flywaydb.core.api.MigrationVersion;
 import org.springframework.context.annotation.Lazy;
+import org.springframework.data.util.Version;
 import org.springframework.jdbc.core.JdbcTemplate;
 
 @Named
-public class SyntheticNftAllowanceOwnerMigration extends RepeatableMigration {
+public class SyntheticNftAllowanceOwnerMigration extends RepeatableMigration implements RecordStreamFileListener {
+
+    public static final Version HAPI_VERSION_0_37_0 = new Version(0, 37, 0);
```
steven-sheehy, line 36:
```suggestion
    static final Version HAPI_VERSION_0_37_0 = new Version(0, 37, 0);
```

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/SyntheticCryptoTransferApprovalMigration.java
```diff
@@ -27,22 +31,28 @@
...
 import org.springframework.jdbc.core.DataClassRowMapper;
 import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
 import org.springframework.transaction.support.TransactionOperations;
 
 @Named
-public class SyntheticCryptoTransferApprovalMigration extends AsyncJavaMigration<Long> {
+public class SyntheticCryptoTransferApprovalMigration extends AsyncJavaMigration<Long>
+        implements RecordStreamFileListener {
 
+    public static final Version HAPI_VERSION_0_38_0 = new Version(0, 38, 0);
```
steven-sheehy, line 48:
```suggestion
    static final Version HAPI_VERSION_0_38_0 = new Version(0, 38, 0);
```

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/SyntheticTokenAllowanceOwnerMigration.java
```diff
@@ -17,15 +17,25 @@
...
+    // 10158
+    public static final Version HAPI_VERSION_0_37_0 = new Version(0, 37, 0);
```
steven-sheehy, line 36:
```suggestion
    static final Version HAPI_VERSION_0_37_0 = new Version(0, 37, 0);
```

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/TokenAccountBalanceMigration.java
```diff
@@ -17,14 +17,23 @@
...
 import org.flywaydb.core.api.MigrationVersion;
 import org.springframework.context.annotation.Lazy;
 import org.springframework.jdbc.core.JdbcOperations;
 
 @Named
-public class TokenAccountBalanceMigration extends RepeatableMigration {
+public class TokenAccountBalanceMigration extends RepeatableMigration implements BalanceStreamFileListener {
+
+    private final AccountBalanceFileRepository accountBalanceFileRepository;
+    AtomicLong firstConsensusTimestamp = new AtomicLong(0L);
```
steven-sheehy, line 35:
```suggestion
    private final AtomicLong firstConsensusTimestamp = new AtomicLong(0L);
```

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/migration/SyntheticCryptoTransferApprovalsMigrationTest.java
```diff
@@ -57,11 +58,21 @@
 @Tag("migration")
 class SyntheticCryptoTransferApprovalsMigrationTest extends IntegrationTest {
 
+    public static final long START_TIMESTAMP = 1568415600193620000L;
```
steven-sheehy, line 61:
```suggestion
    private static final long START_TIMESTAMP = 1568415600193620000L;
```

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/migration/SyntheticTokenAllowanceOwnerMigrationTest.java
```diff
@@ -198,6 +154,134 @@ void repeatableMigration() {
         assertThat(firstPassHistory).containsExactlyInAnyOrderElementsOf(secondPassHistory);
     }
 
+    @Test
+    void reRunOnEndMigration() {
```
steven-sheehy, line 158:
```suggestion
    void onEnd() {
```

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/migration/TokenAccountBalanceMigrationTest.java
```diff
@@ -261,6 +263,29 @@ void migrateWhenNoTokenAccount() {
         assertThat(tokenAccountHistoryRepository.findAll()).isEmpty();
     }
 
+    @Test
+    void reRunMigrate() {
```
steven-sheehy, line 267:
```suggestion
    void onEnd() {
```

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/migration/InitializeEntityBalanceMigrationTest.java
```diff
@@ -181,6 +181,23 @@ void migrateWhenNoEntities() {
                         .returns(EntityType.UNKNOWN, Entity::getType));
     }
 
+    @Test
+    void reRunMigrate() {
```
steven-sheehy, line 185:
```suggestion
    void onEnd() {
```

steven-sheehy, line 185: We should have test cases for when database is not empty after startup and when db is empty but not first stream file (i.e. transaction rollback scenario). Same comment for other balance test.

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/migration/SyntheticTokenAllowanceOwnerMigrationTest.java
steven-sheehy, line 158: Missing test coverage for when `onEnd()` is invoked but it should not run migration. Same for other record migration tests.

steven-sheehy (changes requested):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/SyntheticTokenAllowanceOwnerMigration.java
mgoelswirlds: No I did not. Will need to check this.

xin-hedera: we can discuss. running with `mirror_importer` may not break ourselves but can break other mirrornode operators in case in their environment `mirror_importer` doesn't have permission to delete rows.

then we may have to turn these two migrations to be async.

mgoelswirlds: Ok. Lets discuss further offline.

mgoelswirlds: I verified that even when migrate is called from with the onEnd it uses the mirror_node user.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/InitializeEntityBalanceMigration.java
```diff
@@ -87,4 +100,20 @@ protected void doMigrate() {
...
+        }
+
+        if (secondConsensusTimestamp.get() == streamFile.getConsensusTimestamp()
+                && accountBalanceFileRepository
+                                .findLatest()
+                                .map(AccountBalanceFile::getConsensusTimestamp)
+                                .orElse(-1L)
+                        == firstConsensusTimestamp.get()) {
+            doMigrate();
+        }
```
xin-hedera, line 117: can you explain why it becomes so complicated? I don't think I can fully understand the logic here

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/SyntheticCryptoTransferApprovalMigration.java
```diff
@@ -141,6 +152,29 @@ with contractresults as (
...
+
+    @Override
+    public void onEnd(RecordFile streamFile) throws ImporterException {
+        if (streamFile == null) {
+            return;
+        }
+        try {
+            if (streamFile.getHapiVersion().isGreaterThanOrEqualTo(HAPI_VERSION_0_38_0)
+                    && executed.compareAndSet(false, true)) {
+                var latestFile = recordFileRepository.findLatestWithOffset(1);
```
xin-hedera, line 165: `findLatestWithOffset(1)` may or may not get the record file you want.

Issue here is, we do `recordFileRepository.save(recordFile)` in `SqlEntityListener.onEnd()`. So if the bean `sqlEntityListener` is before the bean `syntheticCryptoTransferApprovalMigration` in `compositeRecordStreamFileListener`, the logic is fine.

Otherwise, let say the `streamFile` being processed has index `current`, the repository method is going to return `current - 2`.

Instead, we should find the latest record file whose `consensus_end` is before `streamFile.consensusStart`

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/SyntheticTokenAllowanceOwnerMigration.java
```diff
@@ -17,15 +17,25 @@
...
+import java.util.concurrent.atomic.AtomicBoolean;
 import org.flywaydb.core.api.MigrationVersion;
 import org.springframework.context.annotation.Lazy;
+import org.springframework.data.util.Version;
 import org.springframework.jdbc.core.JdbcTemplate;
 
 @Named
-public class SyntheticTokenAllowanceOwnerMigration extends RepeatableMigration {
+public class SyntheticTokenAllowanceOwnerMigration extends RepeatableMigration implements RecordStreamFileListener {
+    // 10158
```
xin-hedera, line 35: leftover

```suggestion
```

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/viewmodel/DummyRequest.java
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hedera.mirror.restjava.viewmodel;
+
+import lombok.Data;
+
+@Data
+public class DummyRequest {
```
xin-hedera, line 22: mixed up with the other ticket

xin-hedera (changes requested):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/SyntheticCryptoTransferApprovalMigration.java
mgoelswirlds: '''So if the bean sqlEntityListener is before the bean syntheticCryptoTransferApprovalMigration in compositeRecordStreamFileListener''' - This is what I had observed while manually testing.

xin-hedera: what if we move that out of sqlEntityListener.onEnd?

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/InitializeEntityBalanceMigration.java
mgoelswirlds: Yes. That is because the migration update records when there is data in the account_balance_file table.
So while writing some tests i realized that the migration was not updating any records when run on an empty account_balance_file table. Hence we need to run the migration when the second file is being processed, which ensures that the first file has been parsed and committed. I hope that clears it.
And to ensure that the migration is run only when the second file is being parsed and to optimize to not use DB calls everytime, I added a second timestamp.

xin-hedera: then, we should move the `accountBalanceFileRepository.save(accountBalanceFile)` call into `CompositeAccountBalanceFileListener.onEnd` before calling each `listener.onEnd`, you can change this back to the more straight forward check with just the `firstConsensusTimestamp`

If you do the same to `compositeRecordStreamFileListener.onEnd`, you can keep `recordFileRepository.findLatestWithOffset(1);` in my other comment

mgoelswirlds: This will not work. I guess I missed mentioning a point earlier. The data needs to be committed for the migration to work. Unless I actually commit and then do this migration in a separate transaction , it might not work.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/SyntheticCryptoTransferApprovalMigration.java
mgoelswirlds: The sqlEntityListener bean is before syntheticCryptoTransferApprovalMigration, but maybe you earlier approach for timestamp might be better.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/InitializeEntityBalanceMigration.java
xin-hedera: yes it'll work.

- inside the same database transaction, changes before a statement are visible just like they were commited
- `InitializeEntityBalanceMigration` is synchronous, so when `doMigrate` is called in `onEnd`, all statements run in the same db transaction started from `BalanceFileParser.parse`

it didn't work because the migration relied on `account_balance_file `table having the correct row and in `AccountBalanceFileParser.doParse`, saving the account balance file is after `streamFileListener.onEnd`.

That's why I ask you to move `accountBalanceFileRepository.save` from inside `AccountBalanceFileParser.doParse` to `CompositeAccountBalanceFileListener.onEnd` and before any other listener.onEnd

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/SyntheticCryptoTransferApprovalMigration.java
xin-hedera: right I understand what you mean. however depending on implicit condition is error prone.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/TokenAccountBalanceMigration.java
```diff
@@ -89,4 +112,28 @@ protected void doMigrate() {
...
+    @Override
+    public void onEnd(AccountBalanceFile accountBalanceFile) throws ImporterException {
+        if (firstConsensusTimestamp.get() == -1 || succeeded.get()) return;
+        if (firstConsensusTimestamp.get() == 0) {
+            if (accountBalanceFileRepository
+                    .findLatestBefore(accountBalanceFile.getConsensusTimestamp())
+                    .isEmpty()) {
+                firstConsensusTimestamp.set(accountBalanceFile.getConsensusTimestamp());
+            } else {
+                return;
```
xin-hedera, line 125: you need to set first to -1, when it's not set and there are account balance files in db before this one. Doing so then it'll shortcut at the beginning.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/TokenAccountBalanceMigration.java
```diff
@@ -89,4 +112,28 @@ protected void doMigrate() {
         int count = jdbcOperations.update(UPDATE_TOKEN_ACCOUNT_SQL);
         log.info("Migrated {} token account balances in {}", count, stopwatch);
     }
+
+    @Override
+    public void onEnd(AccountBalanceFile accountBalanceFile) throws ImporterException {
+        if (firstConsensusTimestamp.get() == -1 || succeeded.get()) return;
```
xin-hedera, line 118:
```suggestion
        if (firstConsensusTimestamp.get() == -1 || succeeded.get()) {
          return;
        }
```

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/TokenAccountBalanceMigration.java
```diff
@@ -89,4 +112,28 @@ protected void doMigrate() {
...
+            if (accountBalanceFileRepository
+                    .findLatestBefore(accountBalanceFile.getConsensusTimestamp())
+                    .isEmpty()) {
+                firstConsensusTimestamp.set(accountBalanceFile.getConsensusTimestamp());
+            } else {
+                return;
+            }
+        }
+        if (recordFileRepository.findLatest().map(RecordFile::getConsensusEnd).orElse(-1L)
+                >= firstConsensusTimestamp.get()) {
```
xin-hedera, line 129: optional chaining: `findLatest().map(getConsensusEnd).filter(timestamp -> timestamp >= first.get()).isPresent()`

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/TokenAccountBalanceMigration.java
```diff
@@ -89,4 +112,28 @@ protected void doMigrate() {
...
+            TransactionSynchronizationManager.registerSynchronization(this);
+            doMigrate();
+        }
+    }
+
+    @Override
+    public void afterCommit() {
+        succeeded.compareAndSet(false, true);
```
xin-hedera, line 137: just set it to true

```suggestion
        succeeded.set(true);
```

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/TokenAccountBalanceMigration.java
```diff
@@ -89,4 +112,28 @@ protected void doMigrate() {
...
+        if (firstConsensusTimestamp.get() == 0) {
+            if (accountBalanceFileRepository
+                    .findLatestBefore(accountBalanceFile.getConsensusTimestamp())
+                    .isEmpty()) {
+                firstConsensusTimestamp.set(accountBalanceFile.getConsensusTimestamp());
+            } else {
+                return;
+            }
+        }
+        if (recordFileRepository.findLatest().map(RecordFile::getConsensusEnd).orElse(-1L)
```
xin-hedera, line 128: there should be an empty line in between these `if` blocks

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/InitializeEntityBalanceMigration.java
```diff
@@ -87,4 +109,36 @@ protected void doMigrate() {
         int count = jdbcOperations.update(INITIALIZE_ENTITY_BALANCE_SQL);
         log.info("Initialized {} entities balance in {}", count, stopwatch);
     }
+
+    @Override
+    public void onEnd(AccountBalanceFile accountBalanceFile) throws ImporterException {
```
xin-hedera, line 114: should try to pull the logic in `onEnd` to an abstract base class, since it's exactly the same in `InitializeEntityBalanceMigration.java` and `TokenAccountBalanceMigration.java`

also add comments to demystify the logic and the values (especially -1).

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/TimeSensitiveBalanceMigration.java
```diff
@@ -0,0 +1,81 @@
...
+import com.hedera.mirror.importer.repository.AccountBalanceFileRepository;
+import com.hedera.mirror.importer.repository.RecordFileRepository;
+import java.io.IOException;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicLong;
+import org.springframework.transaction.support.TransactionSynchronization;
+import org.springframework.transaction.support.TransactionSynchronizationManager;
+
+abstract class TimeSensitiveBalanceMigration extends RepeatableMigration implements TransactionSynchronization {
```
xin-hedera, line 30: the base class should also implement `BalanceStreamFileListener`, so the `onEnd` logic is completely in the base

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/InitializeEntityBalanceMigration.java
```diff
@@ -18,13 +18,17 @@
...
+import com.hedera.mirror.importer.repository.RecordFileRepository;
 import jakarta.inject.Named;
 import org.flywaydb.core.api.MigrationVersion;
 import org.springframework.context.annotation.Lazy;
 import org.springframework.jdbc.core.JdbcOperations;
 
 @Named
-public class InitializeEntityBalanceMigration extends RepeatableMigration {
+public class InitializeEntityBalanceMigration extends TimeSensitiveBalanceMigration
+        implements BalanceStreamFileListener {
```
xin-hedera, line 31:
```suggestion
public class InitializeEntityBalanceMigration extends TimeSensitiveBalanceMigration {
```

same for the other migration

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/TimeSensitiveBalanceMigration.java
```diff
@@ -0,0 +1,95 @@
...
+    private final RecordFileRepository recordFileRepository;
+    private final AtomicLong firstConsensusTimestamp = new AtomicLong(0L);
+    private final AtomicBoolean succeeded = new AtomicBoolean(false);
+
+    protected TimeSensitiveBalanceMigration(
+            Map<String, MigrationProperties> migrationPropertiesMap,
+            AccountBalanceFileRepository accountBalanceFileRepository,
+            RecordFileRepository recordFileRepository) {
+        super(migrationPropertiesMap);
+        this.accountBalanceFileRepository = accountBalanceFileRepository;
```
mgoelswirlds, line 45: Will move tests for this class.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/balance/CompositeBalanceStreamFileListener.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    @Override
+    public void onError() {
+        onEach(StreamFileListener::onError);
+    }
+
+    private void onEach(Consumer<BalanceStreamFileListener> consumer) {
+        for (int i = 0; i < listeners.size(); i++) {
```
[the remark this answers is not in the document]
xin-hedera: you second part of the response is wrong. That condition of the check is between latest record file in db and the current account balance file in memory, so the migration may end up doing nothing.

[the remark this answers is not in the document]
mgoelswirlds: Agreed.There is still uncertainty with this.While testing locally it does have account balance file in the DB.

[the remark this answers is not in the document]
mgoelswirlds: In the case where the migration does not do anything, one might have to manually run it. I have added the link in the ReadMe.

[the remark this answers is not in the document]
xin-hedera: that's not the resolution, right? if we choose to just ask operators to rerun it manually, what's the point of the updated algorithm?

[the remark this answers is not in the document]
mgoelswirlds: This could happen in some cases. But what you say is fair. I need to fix it to be more deterministic.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/repository/RecordFileRepository.java
```diff
@@ -35,6 +35,11 @@ public interface RecordFileRepository extends StreamFileRepository<RecordFile, L
                     + "(select max(consensus_end) from record_file) - ?1 order by consensus_end desc limit 1")
     Optional<RecordFile> findLatestWithOffset(long offset);
 
+    @Query(
+            nativeQuery = true,
+            value = "select * from record_file where consensus_end < ?1 order by consensus_end desc limit 1")
+    Optional<RecordFile> findLatestBefore(long offset);
```
steven-sheehy, line 41: Needs a test

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/repository/AccountBalanceFileRepository.java
```diff
@@ -28,6 +28,12 @@ public interface AccountBalanceFileRepository
     @Query(value = "select * from account_balance_file order by consensus_timestamp desc limit 1", nativeQuery = true)
     Optional<AccountBalanceFile> findLatest();
 
+    @Query(
+            nativeQuery = true,
+            value =
+                    "select * from account_balance_file where consensus_timestamp < ?1 order by consensus_timestamp desc limit 1")
+    Optional<AccountBalanceFile> findLatestBefore(long timestamp);
```
steven-sheehy, line 35: Needs a test

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/balance/CompositeBalanceStreamFileListener.java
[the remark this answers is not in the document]
steven-sheehy: > Ahh unfortunately I can't do this. Because errata Migration updates the TimeOffset for this in onEnd.

Just save it twice.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/TimeSensitiveBalanceMigration.java
```diff
@@ -0,0 +1,92 @@
...
+import org.springframework.transaction.support.TransactionSynchronization;
+import org.springframework.transaction.support.TransactionSynchronizationManager;
+
+abstract class TimeSensitiveBalanceMigration extends RepeatableMigration
+        implements BalanceStreamFileListener, TransactionSynchronization {
+
+    private final AccountBalanceFileRepository accountBalanceFileRepository;
+    private final RecordFileRepository recordFileRepository;
+    private final AtomicLong firstConsensusTimestamp = new AtomicLong(0L);
+    private final AtomicBoolean succeeded = new AtomicBoolean(false);
```
steven-sheehy, line 38: Better to be consistent with the other migrations.

```suggestion
    private final AtomicBoolean executed = new AtomicBoolean(false);
```

steven-sheehy (changes requested):

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/migration/TokenAccountBalanceMigrationTest.java
```diff
@@ -54,8 +56,9 @@ class TokenAccountBalanceMigrationTest extends IntegrationTest {
     private final TokenBalanceRepository tokenBalanceRepository;
     private final TokenAccountBalanceMigration tokenAccountBalanceMigration;
     private final TokenTransferRepository tokenTransferRepository;
+    private final TransactionTemplate transactionTemplate;
 
-    private AccountBalanceFile accountBalanceFile;
+    private AccountBalanceFile accountBalanceFile1, accountBalanceFile2;
```
steven-sheehy, line 61: Our standard is to always declare on a separate line.

```suggestion
    private AccountBalanceFile accountBalanceFile1;
    private AccountBalanceFile accountBalanceFile2;
```

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/migration/SyntheticCryptoTransferApprovalsMigrationTest.java
```diff
@@ -186,6 +211,181 @@ void repeatableMigration() {
...
+                .persist();
+        domainBuilder
+                .recordFile()
+                .customize(r -> r.hapiVersionMajor(0)
+                        .hapiVersionMinor(37)
+                        .hapiVersionPatch(10)
+                        .consensusStart(1568415600173620000L)
+                        .consensusEnd(1568415600173620001L))
+                .persist();
+        // given
```
steven-sheehy, line 233: duplicate given section

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/migration/SyntheticCryptoTransferApprovalsMigrationTest.java
```diff
@@ -186,6 +211,181 @@ void repeatableMigration() {
...
+    }
+
+    @Test
+    void onEndHapiVersionNotMatched() {
+        // given
+        domainBuilder
+                .recordFile()
+                .customize(r -> r.hapiVersionMajor(0).hapiVersionMinor(39).hapiVersionPatch(0))
+                .persist();
+        // given
```
steven-sheehy, line 317: duplicate given

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/migration/SyntheticCryptoTransferApprovalsMigrationTest.java
```diff
@@ -186,6 +211,181 @@ void repeatableMigration() {
...
+        var contractId = EntityId.of("0.0.2119900", CONTRACT);
+        var contractId2 = EntityId.of("0.0.2119901", CONTRACT);
+        var priorContractId = EntityId.of("0.0.2119899", CONTRACT);
+        var currentKeyUnaffectedEntity = entityCurrentKey(contractId.getEntityNum());
+        var currentKeyAffectedEntity = entityCurrentKey(contractId2.getEntityNum());
+        var noKeyEntity = entityWithNoKey();
+        var pastKeyUnaffectedEntity = entityPastKey(contractId.getEntityNum());
+        var pastKeyAffectedEntity = entityPastKey(contractId2.getEntityNum());
+        var thresholdTwoKeyEntity = domainBuilder
+                .entity()
```
steven-sheehy, line 327: There's a lot of duplicate setup in different tests. Can they be pulled to a common setup method to reduce duplication and so the test is clear what it's specifically testing? Balance migration tests do this and is much more readable.

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/migration/SyntheticCryptoTransferApprovalsMigrationTest.java
```diff
@@ -186,6 +211,181 @@ void repeatableMigration() {
...
+                tokenPastKeyAffectedEntity,
+                thresholdTwoKeyEntity);
+
+        // when
+        migration.onEnd(RECORD_FILE);
+
+        // then
+        var expectedCryptoTransfers = new ArrayList<>(cryptoTransfersPair.getLeft());
+        expectedCryptoTransfers.addAll(cryptoTransfersPair.getRight());
+        assertThat(cryptoTransferRepository.findAll()).containsExactlyInAnyOrderElementsOf(expectedCryptoTransfers);
```
steven-sheehy, line 375: Use the more succinct `List.of()`. Same comment for other assertions.

```suggestion
assertThat(cryptoTransferRepository.findAll()).containsExactlyInAnyOrderElementsOf(List.of(cryptoTransfersPair.getLeft(), cryptoTransfersPair.getRight()));
```

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/migration/SyntheticTokenAllowanceOwnerMigrationTest.java
```diff
@@ -198,6 +164,195 @@ void repeatableMigration() {
...
+        // given
+        // Token allowance and token allowance history entries that have the incorrect owner
+        var contractResultSenderId = EntityId.of("0.0.2001", CONTRACT);
+        long incorrectOwnerAccountId = 1001L;
+        var incorrectTokenAllowancePair = generateTokenAllowance(contractResultSenderId, incorrectOwnerAccountId);
+
+        // A token allowance that has no contract result, but has primary key values that will end up matching those of
+        // a migrated token allowance.
+        var ownerAccountId = 38L;
+        var spender = 39L;
```
steven-sheehy, line 261: Can we move this to a common setup method shared across all tests?

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/migration/SyntheticNftAllowanceOwnerMigrationTest.java
```diff
@@ -84,43 +95,233 @@ void migrate() {
...
+    @Test
+    void onEnd() {
+        // given
+        // Nft allowance and nft allowance history entries that have the incorrect owner
+        var contractResultSenderId = EntityId.of("0.0.2001", CONTRACT);
+        long incorrectOwnerAccountId = 1001L;
+        var incorrectNftAllowancePair = generateNftAllowance(contractResultSenderId, incorrectOwnerAccountId);
+
+        // An nft allowance can occur through HAPI and absent a corresponding Contract Result.
+        // This nft allowance and it's history should be unaffected by the migration
```
steven-sheehy, line 158: Can we move this to a common setup method shared across all tests?

steven-sheehy (changes requested):

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/migration/SyntheticNftAllowanceOwnerMigrationTest.java
mgoelswirlds: Not sure what you mean. Setup for nft allowances and token allowances moved to a common place?

steven-sheehy: I'm speaking in broader terms, not these couple lines above. I can't highlight in a review discussion all impacted lines. In general, each of these `onEnd*()` tests and even the original `migrate` tests setup the test scenario data (e.g. everything in a `given` block), run the method under test, then verify the results. 

They only vary slightly by setting properties, delete some data, etc. Just like in the balance migration tests, we can move this common setup to a helper method to stay DRY.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/TimeSensitiveBalanceMigration.java
```diff
@@ -0,0 +1,92 @@
...
+            RecordFileRepository recordFileRepository) {
+        super(migrationPropertiesMap);
+        this.accountBalanceFileRepository = accountBalanceFileRepository;
+        this.recordFileRepository = recordFileRepository;
+    }
+
+    @Override
+    public void onEnd(AccountBalanceFile accountBalanceFile) throws ImporterException {
+        try {
+            if (firstConsensusTimestamp.get() == -1 || succeeded.get()) {
```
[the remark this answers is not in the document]
steven-sheehy: I see. Well in that case I think we should use static values for 0 and 1 so they're self-descriptive regardless.

```java
private static final int NO_BALANCE_FILE = 0;
private static final int EXECUTED = -1;
```

In which case I think it's much clearer and we can drop the executed/succeeded.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/TimeSensitiveBalanceMigration.java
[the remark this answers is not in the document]
mgoelswirlds: The code coverage is not accurate, because the tests do cover the branches in the method, but it still shows that as not covered.

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/migration/SyntheticNftAllowanceOwnerMigrationTest.java
mgoelswirlds: I understand. I just wanted to understand if it was for both these tests overall.

mgoelswirlds: I have refactored within the test to have a setup methods within the tests.

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/migration/InitializeEntityBalanceMigrationTest.java
```diff
@@ -181,6 +183,76 @@ void migrateWhenNoEntities() {
                         .returns(EntityType.UNKNOWN, Entity::getType));
     }
 
+    @Test
+    @Transactional
+    void onEndWhenNotFirstFile() {
```
xin-hedera, line 188: these test cases don't test what they claim.

The central problem is `initializeEntityBalanceMigration` is a singleton bean, so the `firstConsensusTimestamp` stateful variable once set to -1, it generally doesn't test anything in any test cases which run afterwards.

you may wonder why the `onEnd` test case still passes, that's because the expected balance for the entities is set in `setup` but not in `setupForFirstAccountBalanceFile`.

This also explains why there is missing test coverage and I believe the warnings put by codecov is accurate

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/migration/InitializeEntityBalanceMigrationTest.java
```diff
@@ -66,14 +76,22 @@ void beforeEach() {
         timestamp = new AtomicLong(0L);
     }
 
+    @AfterEach
+    void after() throws Exception {
+        Field activeField =
+                InitializeEntityBalanceMigration.class.getSuperclass().getDeclaredField("firstConsensusTimestamp");
+        activeField.setAccessible(true);
+        activeField.set(migration, new AtomicLong(0L));
+    }
```
xin-hedera, line 85: can just create a new `InitializeEntityBalanceMigration` object for each of those test cases which depend on `firstConsensusTimestamp`

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/TimeSensitiveBalanceMigration.java
```diff
@@ -0,0 +1,92 @@
...
+import org.springframework.transaction.support.TransactionSynchronizationManager;
+
+abstract class TimeSensitiveBalanceMigration extends RepeatableMigration
+        implements BalanceStreamFileListener, TransactionSynchronization {
+
+    private static final int NO_BALANCE_FILE = 0;
+    private static final int EXECUTED = -1;
+    private final AccountBalanceFileRepository accountBalanceFileRepository;
+    private final RecordFileRepository recordFileRepository;
+    private final AtomicLong firstConsensusTimestamp = new AtomicLong(0L);
```
steven-sheehy, line 38:
```suggestion
    private final AtomicLong firstConsensusTimestamp = new AtomicLong(NO_BALANCE_FILE);
```

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/TimeSensitiveBalanceMigration.java
```diff
@@ -0,0 +1,92 @@
...
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicLong;
+import org.springframework.transaction.support.TransactionSynchronization;
+import org.springframework.transaction.support.TransactionSynchronizationManager;
+
+abstract class TimeSensitiveBalanceMigration extends RepeatableMigration
+        implements BalanceStreamFileListener, TransactionSynchronization {
+
+    private static final int NO_BALANCE_FILE = 0;
+    private static final int EXECUTED = -1;
```
steven-sheehy, line 35:
```suggestion
    private static final long EXECUTED = -1L;
    private static final long NO_BALANCE_FILE = 0L;

```

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/migration/SyntheticCryptoTransferApprovalsMigrationTest.java
```diff
@@ -32,39 +33,76 @@
...
+    Entity noKeyEntity;
+    Entity thresholdTwoKeyEntity;
+
+    private final @Owner JdbcTemplate jdbcTemplate;
     private final SyntheticCryptoTransferApprovalMigration migration;
     private final CryptoTransferRepository cryptoTransferRepository;
     private final TransactionRepository transactionRepository;
     private final TokenTransferRepository tokenTransferRepository;
     private final MirrorProperties mirrorProperties;
+    private static final RecordFile RECORD_FILE = RecordFile.builder()
```
steven-sheehy, line 89: nit: Move above with other statics

steven-sheehy (changes requested):

xin-hedera (dismissed):
LGTM

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/migration/InitializeEntityBalanceMigrationTest.java
```diff
@@ -181,6 +191,90 @@ void migrateWhenNoEntities() {
                         .returns(EntityType.UNKNOWN, Entity::getType));
     }
 
+    @Test
+    @Transactional
+    void onEndWhenNotFirstFile() {
+        // given
+        var migration = new InitializeEntityBalanceMigration(
+                jdbcOperations, mirrorProperties, accountBalanceFileRepository, recordFileRepository);
```
steven-sheehy, line 199: This variable hides the injected class bean. We should make the class `migration` non-final and move this creation to a `@BeforeEach`. Same comment for all other test classes.

steven-sheehy (approved):
LGTM

xin-hedera (approved):
LGTM

PR #6567, opened 2023-08-02
HIP-584: Add integration tests for `getTokenExpiryInfo` operation

### LINKED ISSUE #6462: HIP-584: Provide integration tests for GetTokenExpiryInfoPrecompile
### Problem

We need to support GetTokenExpiryInfoPrecompile to enhance eth_estimateGas functionality.

### Solution

Provide integration tests for GetTokenExpiryInfoPrecompile to validate it works properly.

### Alternatives

_No response_

0xivanov:
**Description**:
Add integration tests for `getTokenExpiryInfo` operation

**Related issue(s)**:

Fixes #6462

**Notes for reviewer**:

**Checklist**


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/token/TokenAccessorImpl.java
```diff
@@ -266,20 +265,6 @@ private Optional<EvmTokenInfo> getTokenInfo(final Address address) {
         return Optional.of(evmTokenInfo);
     }
 
-    private Long expirationTimeSeconds(Token token) {
-        Long createdTimestamp = token.getCreatedTimestamp();
-        Long autoRenewPeriod = token.getAutoRenewPeriod();
-        Long expirationTimestamp = token.getExpiry();
-
-        if (expirationTimestamp != null) {
```
IvanKavaldzhiev, line 274: I believe we should keep this method, since it seems we may have cases where `expirationTimestamp` is null and we manually calculate it by adding `createdTimestamp` + `autoRenewPeriod`. The obsolete thing here is 

`expirationTimestamp / NANOS_PER_SECOND;`

We can directly return  `expirationTimestamp`

IvanKavaldzhiev: Which makes me think we should also convert the `createdTimestamp` in seconds when we load it into the `Token` model inside `TokenDatabaseAccessor` to be consistent.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/token/TokenAccessorImpl.java
```diff
@@ -272,9 +270,9 @@ private Long expirationTimeSeconds(Token token) {
         Long expirationTimestamp = token.getExpiry();
 
         if (expirationTimestamp != null) {
-            return expirationTimestamp / NANOS_PER_SECOND;
+            return expirationTimestamp;
         } else if (createdTimestamp != null && autoRenewPeriod != null) {
```
steven-sheehy, line 274: Is it necessary to duplicate this logic from `AbstractEntity.getEffectiveExpirationTimestamp()` which is used in `TokenDatabaseAccessor` to populate the expiry?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/token/TokenAccessorImpl.java
0xivanov: As Steven noted, the logic of this method is executed in `AbstractEntity.getEffectiveExpirationTimestamp()` which we use in the `TokenDatabaseAccessor` when building the token. So the null checks and the other logic is already applied there.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/token/TokenAccessorImpl.java
0xivanov: Nice catch!

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServicePrecompileTest.java
```diff
@@ -375,8 +375,10 @@ enum ContractReadFunctions {
                 "getTokenKeyPublic",
                 new Object[] {NFT_ADDRESS_GET_KEY_WITH_DELEGATABLE_CONTRACT_ID, 64L},
                 new Object[] {false, Address.ZERO, new byte[0], new byte[0], CONTRACT_ADDRESS}),
-        GET_CUSTOM_FEES_FOR_TOKEN("getCustomFeesForToken", new Object[] {FUNGIBLE_TOKEN_ADDRESS}, new Object[] {});
-
+        GET_CUSTOM_FEES_FOR_TOKEN("getCustomFeesForToken", new Object[] {FUNGIBLE_TOKEN_ADDRESS}, new Object[] {}),
+        GET_TOKEN_EXPIRY("getExpiryInfoForToken", new Object[] {FUNGIBLE_TOKEN_ADDRESS_WITH_EXPIRY}, new Object[] {
+            1000L, FUNGIBLE_TOKEN_ADDRESS_WITH_EXPIRY, 1800L
```
IvanKavaldzhiev, line 380: Why `getTokenExpiry` returns the token address as an `auto renew account` - the second returned value? It should be the account that was set as an autoRenew account for this token.

0xivanov: This is how we set it in the persist function:

``` java
 domainBuilder
                .entity()
                .customize(e -> e.id(tokenEntityId.getId())
                        .autoRenewAccountId(tokenEntityId.getId())
```

The autorenew account is the same as the token account. Should I change it?

IvanKavaldzhiev: Yes, that's not correct. We should set an account, not a token as an auto renew account.

0xivanov: Fixed, along with a bug in `TokenDatabaseAccessor`

steven-sheehy (approved):
LGTM

PR #6577, opened 2023-08-03
Add async migration to fix crypto allowance amount

### LINKED ISSUE #6552: Fix crypto allowance amount
### Problem

We dropped the migration to fix remaining crypto allowance amount in ticket https://github.com/hashgraph/hedera-mirror-node/issues/3576 at the last minute due to performance concern. We need to implement a performant solution and test it with mainnet database.

### Solution

Add an async java migration to backfill the remaining crypto allowance amount for the crypto allowances in the current table as of the state when the async java migration starts. 

### Alternatives

_No response_

xin-hedera:
**Description**:

This PR adds an async migration to fix crypto allowance amount

- Add migration to deduct amount in crypto transfer using allowance from remaining allowance amount
- Make `AsyncJavaMigration` extends `RepeatableMigration`
- Workaround to support instantiating `AsyncJavaMigration` with matching `jdbcTemplate` and `transactionOperations` based on non-default `dataSource`

**Related issue(s)**:

Fixes #6552 

**Notes for reviewer**:

Tested with mainnet staging database, takes by average 2-3 minutes for one-day worth crypto transfers at current TPS.

**Checklist**


hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/AsyncJavaMigration.java
```diff
@@ -117,22 +119,24 @@ protected void doMigrate() throws IOException {
...
-                        transactionOperations.execute(t -> migratePartial(previous.get())), Optional.empty());
+                        getTransactionOperations().execute(t -> migratePartial(previous.get())), Optional.empty());
                 count++;
 
-                if (stopwatch.elapsed(TimeUnit.MINUTES) >= minutes) {
+                long elapsed = stopwatch.elapsed(TimeUnit.MINUTES);
+                if (elapsed >= minutes) {
                     log.info("Completed iteration {} with last value: {}", count, last.orElse(null));
-                    minutes++;
+                    minutes = elapsed + 1;
```
xin-hedera, line 139: minutes needs to be the the next to log a message, not self incrementing.

the issue with self incrementing is, e.g., if one iteration takes 30 minutes, and the next several takes 5 seconds, there'll be a log for all of them, instead of at least 1 minute apart.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/FixCryptoAllowanceAmountMigration.java
```diff
@@ -0,0 +1,250 @@
...
+            with sentinel as (
+              insert into crypto_allowance_migration (amount, amount_granted, owner, payer_account_id, spender, timestamp_range)
+              select 0, 0, 0, 0, 0, int8range(consensus_end, null)
+              from record_file
+              order by consensus_end desc
+              limit 1
+            )
+            insert into crypto_allowance_migration (amount, amount_granted, owner, payer_account_id, spender, timestamp_range)
+            select
+              amount_granted - amount, -- the spent amount tracked by importer as of consensus_end
```
steven-sheehy, line 75: In `V1.84.2__allowance_amount.sql`, we `update crypto_allowance set amount = amount_granted;`. That means here `crypto_allowance_migration.amount` will be `0` for all rows that haven't had a new allowance grant since then. Later we `set amount = greatest(ca.amount + m.amount, 0)` making the final product always zero. Or am I missing something?

steven-sheehy: Nevermind, think I see it now. You're starting at zero for those rows, accumulating negative amounts in `crypto_allowance_migration.amount`, then adding that negative amount to the `crypto_allowance.amount`.

steven-sheehy (approved):
LGTM

jascks (approved):
LGTM. Pretty cool stuff!

PR #6591, opened 2023-08-04
Only show active fungible allowance in rest api response

### LINKED ISSUE #6380: Crypto / Fungible token allowance REST API show deleted allowances
### Problem

When a crypto / fungible token allowance is deleted by setting the granted amount to 0, the corresponding mirrornode rest api still shows an allowance entry with amount=0, it confuses users since the number of allowances listed under an owner account can then exceed the configured max for the network. To be consistent, the rest apis should not show deleted allowances.

### Solution

* Update existing REST API query to only include allowances with `amount_granted > 0`
* Add an index on `amount_granted` where it's zero

### Alternatives

_No response_

xin-hedera:
**Description**:

This PR changes crypto / fungible token allowance rest api to only show active allowances

**Related issue(s)**:

Fixes #6380 

**Notes for reviewer**:

Removed the migration which adds index since the tables are tiny now. When the tables become big we need to consider either add the partial index with amount_granted > 0, or delete the current row when there is crypto approve allowance tx with amount_granted = 0

**Checklist**


steven-sheehy (approved):
LGTM

jascks:
Is there any distinction between deleted and spent allowance? I realize this ticket was written before my more recent changes, but `amount_granted > 0` means the allowance is _not_ deleted. If it is also true `amount > 0`, then the allowance exists and has not been completely spent. If `amount == 0`, then it has been spent.

So perhaps I'm reading too much into just basing the decision solely on `amount`.  I agree that a spent allowance is not particularly useful, but it is also true that it has not been explicitly deleted by the owner.

steven-sheehy (approved):

jnels124:
LGTM

xin-hedera:
> Is there any distinction between a deleted and a spent allowance? I realize this ticket was written before my more recent changes, but `amount_granted > 0` means the allowance is _not_ deleted. If it is also true `amount > 0`, then the allowance exists and has not been completely spent. If `amount == 0`, then it has been spent.
> 
> So perhaps I'm reading too much into just basing the decision solely on `amount > 0`. I agree that a spent allowance is not particularly useful, but it is also true that it has not been explicitly deleted by the owner.
> 
> I guess that's why "active" appears in the title of the PR vs "deleted" in the original ticket? :)

The goal of the ticket & PR is to make mirrornode allowance response consistent with consensus nodes state. As I explained yesterday, when an allowance is complete spent, it's removed from consensus nodes state and an slot opens up for the owner to grant another allowance. There is no difference from explicitly deleting / revoking an allowance by setting the `amount_grante` to 0 in that aspect.

jascks (approved):
LGTM

Issue #3576, opened 2022-04-15
HIP-336 Track remaining crypto and fungible token allowances


steven-sheehy:
### Problem

The mirror node tracks the allowances granted via `CryptoApproveAllowance` and `CryptoDeleteAllowance`, but does not currently track the current allowance that result from actually spending those allowances.

### Solution

* Rename `crypto_allowance.amount` to `amount_granted`
* Add a `crypto_allowance.amount`
* Add a `crypto_allowance.created_timestamp`
* Rename `token_allowance.amount` to `amount_granted`
* Add a `token_allowance.amount`
* Add a `token_allowance.created_timestamp`
* Add a `@UpsertColumn(coalesce = "case when {0} >= 0 then {0} else coalesce(e_{0}, 0) + coalesce({0}, 0) end")` to `CryptoAllowance.amount` and `TokenAllowance.amount`
* In `CryptoApproveAllowanceTransactionHandler`, in both `parseCryptoAllowances()` and `parseTokenAllowances()`:
  - Set `created_timestamp` to `consensus_timestamp`
  - Also, set `amount_granted` to approval amount
* In `SqlEntityListener`, merge the allowances in the same record file
* In `EntityRecordItemListener`, for every crypto transfer in the body (not `CryptoTransfer` generated from record) with `is_approval=true` create a `CryptoAllowance` with primary keys, amount from transfer, and no timestamp
* In `EntityRecordItemListener`, for every token transfer in the body (not `TokenTransfer` generated from record) with `is_approval=true` create a `TokenAllowance` with primary keys, amount from transfer, and no timestamp
* Write a migration to calculate the current allowance
  * Populate `amount` in main table using the amount granted and the transfers with `is_approval=true`
  * Populate `created_timestamp` in main and history tables with min `timestamp_range` across both tables
* Fix allowance REST APIs to use the renamed `amount_granted`

PR #6592, opened 2023-08-04
Acceptance tests for token allowance amounts

### LINKED ISSUE #6553: Acceptance tests for remaining crypto and token allowance amounts
### Problem

With the implementation of #3576 completed, implement acceptance tests around the new functionality provided, namely crypto and token allowance definition and that approved crypto/fungible token transfers are properly reflected in the allowance balances.

### Solution

- Create allowances defining appropriate owner and spender IDs.
- Verify existence and expected configuration via mirror node rest api
- Perform crypto and fungible token transfers, both approved utilizing the owner and spender as expected, as well as other scenarios that should not affect the allowance balances.
- Refer to code recently added to recover HBARs spent

### Alternatives

_No response_

jascks:
**Description**:
With the implementation of #3576 completed, implement acceptance tests around the new functionality provided, namely crypto and token allowance definition and that approved crypto/fungible token transfers are properly reflected in the allowance balances.

- Create fungible token and allowance defining appropriate owner and spender IDs.
- Verify existence and expected configuration via mirror node rest api
- Perform fungible token transfer, both approved utilizing the owner and spender as expected, as well as other scenarios that should not affect the allowance balances.
- Refer to code recently added to clean up and recover HBARs spent

**Related issue(s)**:

Fixes #6553

**Notes for reviewer**:

**Checklist**


hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/client/TokenClient.java
```diff
@@ -506,5 +565,32 @@ public NetworkTransactionResponse updateTokenFeeSchedule(
         return response;
     }
 
+    @RequiredArgsConstructor
+    public enum TokenNameEnum {
+        FUNGIBLE(
```
jascks, line 570: Only defined what I required for this issue. Other token configs can be defined in the future as required.

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/props/MirrorTransfer.java
```diff
@@ -16,18 +16,21 @@
...
 @AllArgsConstructor(access = AccessLevel.PRIVATE) // for builder
-@Builder
+@SuperBuilder
 @Data
 @NoArgsConstructor
 public class MirrorTransfer {
     private String account;
     private long amount;
+
+    @JsonProperty("is_approval")
```
jascks, line 34: Apparently Jackson could not figure this out. Even though the API JSON indicated:
```
            "token_transfers": [
                {
                    "account": "0.0.7046",
                    "amount": -100,
                    "is_approval": true,
                    "token_id": "0.0.7047"
                },
```
I was always getting `isApproval` to be false.

hedera-mirror-test/src/test/resources/features/token/tokenAllowance.feature
```diff
@@ -0,0 +1,26 @@
...
+    Given Sender 0 transfers <transferAmount> tokens 0 from the approved allowance to recipient 0
+    Then the mirror node REST API should confirm the debit of <transferAmount> from approved token 0 allowance of <approvedAmount> for sender 0
+    Then the mirror node REST API should confirm the approved transfer of <transferAmount> tokens
+    When I approve sender 0 to transfer up to <approvedAmount> tokens 0
+    Then the mirror node REST API should confirm the approved token 0 allowance of <approvedAmount> for sender 0
+    When I delete the allowance on token 0 for sender 0
+    Then the mirror node REST API should confirm the token allowance deletion
+    Examples:
+      | tokenName  | approvedAmount | transferAmount |
+      | "FUNGIBLE" | 10000          | 100            |
```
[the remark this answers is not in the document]
xin-hedera: p.s., we tend to include additional steps in existing scenarios unless perhaps it becomes too confusing. The reason is we need to make acceptance tests cost effective, here an additional token is created for the scenario, and token creation is quite expensive.

hedera-mirror-test/src/test/resources/features/token/tokenAllowance.feature
```diff
@@ -0,0 +1,26 @@
+@tokenallowance @allowance @fullsuite
+Feature: Account Crypto Allowance Coverage Feature
+
+  @critical @release @acceptance
+  Scenario Outline: Validate approval TokenTransfer affect on TokenAllowance amount
+    Given I use named token <tokenName>
+    Given I associate a new sender account with token 0
+    And the mirror node REST API should return the transaction
+    Given I associate a new recipient account with token 0
```
steven-sheehy, line 9: This creates two new accounts. We should reuse BOB and CAROL like in the crypto allowance test.

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/TokenFeature.java
```diff
@@ -87,7 +87,12 @@ public class TokenFeature {
     private final Map<TokenId, List<CustomFee>> tokenCustomFees = new HashMap<>();
     private final Map<TokenId, List<Long>> tokenSerialNumbers = new HashMap<>();
     private final List<TokenId> tokenIds = new ArrayList<>();
-    private NetworkTransactionResponse networkTransactionResponse;
+
+    @Given("I use named token {string}")
+    public void createOrReuseToken(String tokenName) {
+        var tokenId = tokenClient.getToken(TokenClient.TokenNameEnum.valueOf(tokenName));
+        tokenIds.add(tokenId);
```
steven-sheehy, line 94: Putting it in tokenIds means it gets cleaned up in the `@After`, but we want this token to stick around to be reused in other features. Hence, why we're caching it in TokenClient.

hedera-mirror-test/src/test/resources/features/token/tokenAllowance.feature
```diff
@@ -0,0 +1,26 @@
...
+    And the mirror node REST API should return the transaction
+    Given I approve sender 0 to transfer up to <approvedAmount> tokens 0
+    Then the mirror node REST API should confirm the approved token 0 allowance of <approvedAmount> for sender 0
+    Given I transfer <transferAmount> tokens 0 to recipient 0
```
steven-sheehy, line 13: I don't think we can use the index based approach since it will delete the shared accounts and tokens after the feature. Any new step we add should not use the index based approach. We can convert the existing steps to support a string for name and check if it's enum or number and use appropriately.

steven-sheehy (changes requested):

hedera-mirror-test/src/test/resources/features/token/tokenAllowance.feature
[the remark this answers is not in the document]
jascks: This is similar to `cryptoAllowance.feature`, except that there are additional steps to create a token and associate it with spender and recipient. The story goes:

1. I (owner/operator) create an allowance (crypto or fungible token).
2. I (owner, not spender) transfer crypto/tokens to recipient. The allowance is unchanged (no debit to `amount`)
3. The spender transfers crypto/tokens to recipient. An approved transfer occurred and the allowance`amount` is debited according.
4. I create a new allowance for sender.  The allowance has been essentially re-established.
5. I delete the allowance. The allowance has been deleted.

Of course there are numerous (free) mirror node API calls around that. Let me know what does not need to be tested and I will eliminate it.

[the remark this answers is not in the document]
jascks: Regarding the new feature/merge with existing feature question, I think it is cleaner for the feature to be separate, but the concern for cost is obviously a real one. Based on what we agree actually needs to be tested, perhaps a simpler version of `tokenAllowance.feature` can be integrated into the mentioned `token.feature` scenario.

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/client/TokenClient.java
```diff
@@ -506,5 +565,32 @@ public NetworkTransactionResponse updateTokenFeeSchedule(
...
+                TokenType.FUNGIBLE_COMMON,
+                TokenSupplyType.INFINITE,
+                1_000_000,
+                1,
+                TokenFreezeStatus.FreezeNotApplicable_VALUE,
+                TokenKycStatus.KycNotApplicable_VALUE,
+                Collections.emptyList());
+
+        private final String symbol;
+        private final TokenType tokenType;
```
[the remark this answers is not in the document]
jascks: Why not hardcode it here in the daylight rather than down in some private method somewhere? As this token re-use mechanism gets adopted, a future test writer can see the options available and if one is not suitable they can define a new one.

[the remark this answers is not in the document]
steven-sheehy: YAGNI. We should push users to not need all that customization and keep it simple with a couple generic tokens for all use cases.

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/TokenFeature.java
jascks: Doh. Yes!

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/response/MirrorTokenAllowanceResponse.java
```diff
@@ -0,0 +1,27 @@
...
+package com.hedera.mirror.test.e2e.acceptance.response;
+
+import com.hedera.mirror.test.e2e.acceptance.props.MirrorTokenAllowance;
+import com.hedera.mirror.test.e2e.acceptance.props.MirrorTransferAllowance;
+import java.util.List;
+import lombok.Data;
+
+@Data
+public class MirrorTokenAllowanceResponse extends MirrorTransferAllowance {
+    List<MirrorTokenAllowance> allowances;
```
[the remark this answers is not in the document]
jascks: Done! Extending `MirrorTransferAllowance` serves no purpose. The object contains the deserialized JSON result of the REST API and is used only in one place and the first thing the assert logic does is call `getAllowances()` and go from there. 

The tests still pass. Also fixed this in `MirrorCryptoAllowanceResponse`.

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/props/MirrorTokenAccountBalance.java
```diff
@@ -20,6 +20,6 @@
 
 @Data
 public class MirrorTokenAccountBalance {
-    private String tokenId;
```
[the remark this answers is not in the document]
jascks: Done. Not sure how this happened. This has been restored. This class is referenced only in `MirrorCryptoBalance`, which is then referenced only in `MirrorBalancesResponse`, which then is referenced nowhere, at least according to IDEA.

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/TokenFeature.java
jascks: Done!

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/props/MirrorTokenAccountBalance.java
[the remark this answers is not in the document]
jascks: Done

hedera-mirror-test/src/test/resources/features/token/tokenAllowance.feature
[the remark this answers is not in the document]
jascks: Okay, please let me know what you think of the update to this file and the related clients and token feature. I've not cleaned up `TokenFeature` yet, but before going too much further this direction and proceeding to implement any other features in this manner, I want to hear your feedback on what you see before you.

The steps are all now utilizing the re-usable tokens (new in this PR) and accounts (existing).  The index based stuff is not used here. Also, you will notice the state (tokenId, spenderId etc.) is being kept in the Ghirkin text rather than fields in the Java feature code. I kind of prefer that, but you may have a different opinion.

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/client/AccountClient.java
```diff
@@ -93,7 +93,7 @@ public ExpandedAccountId getAccount(AccountNameEnum accountNameEnum) {
         // retrieve account, setting if it doesn't exist
         ExpandedAccountId accountId = accountMap.computeIfAbsent(accountNameEnum, x -> {
             try {
-                return createNewAccount(SMALL_INITIAL_BALANCE, accountNameEnum);
+                return createNewAccount(DEFAULT_INITIAL_BALANCE, accountNameEnum);
```
jascks, line 96: Looks like token transfer fees are about 835,000 tℏ and `SMALL_INITIAL_BALANCE` is only 500,000, resulting in transaction failure with `INSUFFICIENT_PAYER_BALANCE`.

hedera-mirror-test/src/test/resources/features/token/tokenAllowance.feature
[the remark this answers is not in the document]
xin-hedera: It's fair to have it in a separate scenario if we intend to change all token creation to use the new method.

For the steps outlined, a couple of things

- shouldn't have the step which checks a direct owner to recipient transfer does not affect the allowance, this is a negative test case and usually we don't do negative test in acceptance test suite
- I don't think spender and token association is needed, after all, the spender transfers the token on behalf of the owner

hedera-mirror-test/src/test/resources/features/token/tokenAllowance.feature
jascks: Done. Moved on to named tokens and accounts.

hedera-mirror-test/src/test/resources/features/token/tokenAllowance.feature
jascks: Done.

hedera-mirror-test/src/test/resources/features/token/tokenAllowance.feature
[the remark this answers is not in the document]
jascks: Done! Awesome, I have removed those steps. It runs noticeable faster now. :)

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/props/MirrorTokenAllowance.java
```diff
@@ -0,0 +1,26 @@
...
+package com.hedera.mirror.test.e2e.acceptance.props;
+
+import lombok.Data;
+
+@Data
+public class MirrorTokenAllowance extends MirrorTransferAllowance {
+    private String tokenId;
+    private long amount;
+    private long amountGranted;
+}
```
xin-hedera, line 26: `amount` and `amountGranted` are already in the base class.

btw, we should rename the base class `MirrorTransferAllowance` to either `MirrorAllowance` or `MirrorFungibleAllowance` for better clarity.

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/TokenFeature.java
```diff
@@ -421,6 +528,77 @@ public void verifyMirrorTokenRelationshipNftAPIResponses() {
...
+        var tokenId = tokenIds.get(getIndexOrDefault(tokenIndex));
+        var spenderAccountId = senders.get(getIndexOrDefault(senderIndex));
+        networkTransactionResponse = accountClient.approveToken(tokenId, spenderAccountId.getAccountId(), 0L);
+        assertNotNull(networkTransactionResponse.getTransactionId());
+        assertNotNull(networkTransactionResponse.getReceipt());
+    }
+
+    @Then(
+            "the mirror node REST API should confirm the approved token (?:(.*) )?allowance of (.*) for sender(?: (.*))?$")
+    public void verifyMirrorAPIApprovedTokenAllowanceResponse(
```
xin-hedera, line 551: add the annotation to retry all these public methods which assert certain value from mirrornode rest api since it takes time for the changes being finalized in consensus nodes then propagated and persisted by mirrornode importer

also check the crypto allowance changes make sure it's not missed there.

```suggestion
    @RetryAsserts
    public void verifyMirrorAPIApprovedTokenAllowanceResponse(
```

xin-hedera (changes requested):

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/props/MirrorTokenAllowance.java
jascks: The base class `MirrorTransferAllowance` defines `owner`, `spender` and `timestamp` only, and does not extend anything else.  It defines no amount related fields.

However, both `MirrorTokenAllowance` and `MirrorCryptoAllowance` extend `MirrorTransferAllowance` and add `amount` and `amountGranted`. We could instead have `MirrorTokenAllowance` extend `MirrorCryptoAllowance` and just add  `tokenId`.

The base class `MirrorTransferAllowance` does not really define an allowance of any kind with only `owner`, `spender` and `timestamp`, and should perhaps be abstract. Or, since there are no other subclasses than the two above, it should go a way and `MirrorCryptoAllowance` encompass its fields and also be the base class of `MirrorTokenAllowance` as described above.

jascks: Done! After out of band discussion, changes now committed.

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/client/TokenClient.java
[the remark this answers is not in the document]
jascks: This has been done.

hedera-mirror-test/src/test/resources/features/contract/call.feature
```diff
@@ -2,23 +2,26 @@
 Feature: eth_call Contract Base Coverage Feature
 
   Scenario Outline: Validate eth_call
-    Given I successfully create a new fungible token
-    Given I successfully create a new non-fungible token
+    Given I ensure token <nonFungibleTokenName> has been created
+    Given I ensure token <fungibleTokenName> has been created
```
xin-hedera, line 6: don't think you need the examples section since I don't see a need even in the future we set `nonFungibleTokenName` to sth other than `NFT`, same for `fungibleTokenName`

jascks: Okay, good to know. Done!

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/CallFeature.java
```diff
@@ -144,68 +131,57 @@ public void createNewEstimateTestContract() throws IOException {
...
-        createNewToken(
-                nonFungibleTokenName,
-                TokenFreezeStatus.FreezeNotApplicable_VALUE,
-                TokenKycStatus.KycNotApplicable_VALUE,
-                TokenType.NON_FUNGIBLE_UNIQUE,
-                TokenSupplyType.INFINITE,
-                Collections.emptyList());
+    @Given("I ensure token {string} has been created")
+    public void createNamedToken(String tokenName) {
+        // If just now created, then ensure transaction has been seen by mirror node and receipt was available
```
steven-sheehy, line 136: This doesn't actually ensure that the transaction has been seen by mirror node, just consensus nodes, right? Does it need to to avoid race conditions?

steven-sheehy (approved):
LGTM

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/CallFeature.java
jascks: Oooh... Yes, this is only an interaction with the consensus nodes (HAPI). My comment is a lie! While that did appear to solve the issue, it could become problematic again.

When using the named token mechanism (`TokenClient.getToken()`) the token may or may not be created. So there may or may not be a transaction. Instead, in this step, after calling `getToken()` the mirror node API should be invoked to ensure the token exists (`/api/v1/tokens?token.id=<tokenId>`) with the standard retry template retrying 404.

Does that make sense? I can create a new ticket and PR to address this.

steven-sheehy: Can just ensure the transaction is present on mirror node

jascks: Created issue #6648, opened PR #6649.

PR #6630, opened 2023-08-10
Creating a working example for contract calls

Neurone:
**Description**:
In the Swagger UI, the current example payload for the endpoint [/api/v1/contracts/call](https://testnet.mirrornode.hedera.com/api/v1/docs/#/contracts/contractsCall) contains some errors:
- the amount of gas provided is too much (120M)
- the contract does not exist anymore due to recent testnet reset

This PR modifies the example payload, so the user can test the API without customizing the payload.

**Related issue(s)**:

N/A

**Notes for reviewer**:

N/A

**Checklist**

### DIFF
--- a/hedera-mirror-rest/api/v1/openapi.yml
+++ b/hedera-mirror-rest/api/v1/openapi.yml
@@ -2410,7 +2410,7 @@ components:
           type: string
         data:
           description: Hexadecimal method signature and encoded parameters.
-          example: "0x0198489200000000000000000000000000000000000000000000000000000000000003ee"
+          example: "0x47f1aae7"
           format: binary
           maxLength: 12288
           nullable: true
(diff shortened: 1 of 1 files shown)


hedera-mirror-rest/api/v1/openapi.yml
```diff
@@ -2446,7 +2446,7 @@ components:
           type: integer
         to:
           description: The 20-byte hexadecimal EVM address the transaction is directed to.
-          example: "0x00000000000000000000000000000000000003f4"
+          example: "0xD9d0c5C0Ff85758BdF05A7636F8036d4D065F5B6"
```
steven-sheehy, line 2449: nit: prefer lowercase hex like the other examples

xin-hedera (approved):
LGTM

hedera-mirror-rest/api/v1/openapi.yml
Neurone: Fixed.
I agree in general lowercase is good for any other value, but for addresses I usually prefer the checksum version.

steven-sheehy (approved):
LGTM

PR #6634, opened 2023-08-11
HIP-584: Properly update newly created create2 contract in state

IvanKavaldzhiev:
**Description**:

Currently if we create a `CREATE2` contract and perform operations with it in a single transaction, it's not properly stored in the in-memory state and cannot be fetched, which causes reverts.

**Related issue(s)**:

**Notes for reviewer**:

**Checklist**


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/HederaEvmStackedWorldStateUpdater.java
```diff
@@ -104,10 +104,16 @@ public void trackLazilyCreatedAccount(final Address address) {
     }
 
     private void persistAccount(Address address, long nonce, Wei balance) {
+        final var resolvedAddress = Address.wrap(Bytes.wrap(unaliased(address.toArray())));
+
         final var accountModel = new com.hedera.services.store.models.Account(
-                ByteString.EMPTY,
+                (!Address.ZERO.equals(resolvedAddress) && address != resolvedAddress)
```
steven-sheehy, line 110: `address != resolvedAddress` will always be true since `Address.wrap()` always returns a new `Address` object.

steven-sheehy, line 110: Move `!Address.ZERO.equals(resolvedAddress)` to a variable so it's not called twice.

IvanKavaldzhiev: Fixed

IvanKavaldzhiev: Fixed

steven-sheehy (approved):
LGTM

Issue #6585, opened 2023-08-04
HIP-584: Create integration tests for testing dynamic `eth_call` using `create` and `update` token precompiles


IvanKavaldzhiev:
### Problem

We need to create integration tests that validate `eth_call` works correctly in dynamic cases using create and update precompiles to validate that they update the in-memory state properly.

### Solution

We need to create more tests cases in `hedera-mirror-web3` by combining update and create token precompiles and then using token read precompiles. In this way we would test 2 thing simultaneously:

1. Update and create token precompiles update properly the in-memory state
2. All token read related precompiles are reading properly correct values in dynamic cases (using non-static frames)

The following method combinations could be defined in separate smart contract methods:

- Update token key + get token info key
- Update token expiry + get token info expiry
- Update token info that updates symbol + get token info symbol
- Update token info that updates name + get token info name
- Update token info that updates memo + get token info memo
- Update token info that updates decimals + get token info decimals
- Update token info that updates tokenUri + get token info metadata
- Update auto renew period + get token info auto renew period
- Delete fungible/non-fungible token + get token info isDeleted
- CreateToken for fungible token with/without default freeze/kyc status + name + symbol + decimals + getDefaultFreezeStatus + getDefaultKycStatus + isToken
- CreateToken for nft with/without default freeze/kyc status + name + symbol + tokenUri + ownerOf + getDefaultFreezeStatus + getDefaultKycStatus + isToken

bilyana-gospodinova:
There is no way to update the token decimals so this test will be skipped: "Update token info that updates decimals + get token info decimals".

PR #6613, opened 2023-08-08
HIP-584: Add integration tests for dynamic eth_call - part 1

0xivanov:
**Description**:  

We need to create integration tests that validate eth_call works correctly in dynamic cases using precompiles that modify the token supply and status or the account relation to a token or their privileges related to token.

Fixed the following bugs:

1. `MintPrecompile` and `BurnPrecompile` do not return correct `totalSupply` for NFTs
2. `AssociatePrecompile` creates `TokenRelationship` with `frozen = true` and `kycGranted = false`
3. `ErcTransferPrecompile` never reverts
4. `StoreImpl` persists `TokenRelationship` with with `notYetPersisted = true`

This PR covers the following scenarios:

Mint fungible/non-fungible token + get token info total supply+ get balance of the treasury
Burn fungible/non-fungible token + get token info total supply + get balance of the treasury
Wipe + get token info total supply + get balance of the account which balance was wiped
Pause fungible/non-fungible token + get token info pause status + unpause + get token info pause status
Freeze fungible/non-fungible token + get token info freeze status + unfreeze + get token info freeze status
Associate fungible/non-fungible token transfer (should pass) + dissociate + transfer (should fail)
Approve fungible/non-fungible token + allowance
Approve fungible/non-fungible token + transferFrom with spender + allowance + balance

**Related issue(s)**:

Fixes #6586

**Notes for reviewer**:

**Checklist**

- [ ] Documented (Code comments, README, etc.)
- [x] Tested (unit, integration, etc.)

PR #6647, opened 2023-08-14
HIP-584: Create integration tests for testing dynamic eth_call using create and update token precompiles

bilyana-gospodinova:
**Description**:
We need to create integration tests that validate eth_call works correctly in dynamic cases using create and update precompiles to validate that they update the in-memory state properly.

Fixes the following bugs:
- SyntheticTxnFactory#createTokenCreate does not add the keys.
- AssociatePrecompile creates TokenRelationship with frozen = true and kycGranted = false (from this PR: https://github.com/hashgraph/hedera-mirror-node/pull/6613).
- FunctionEncodeDecoder#encodeToken does not map the keys.

**This PR covers the following scenarios**:
- Update token key + get token info key
- Update token expiry + get token info expiry
- Update token info that updates symbol + get token info symbol
- Update token info that updates name + get token info name
- Update token info that updates memo + get token info memo
- Update auto renew period + get token info auto renew period
- Delete fungible/non-fungible token + get token info isDeleted
- CreateToken for fungible token with/without default freeze/kyc status + name + symbol + decimals + getDefaultFreezeStatus + getDefaultKycStatus + isToken
- CreateToken for nft with/without default freeze/kyc status + name + symbol + tokenUri + ownerOf + getDefaultFreezeStatus + getDefaultKycStatus + isToken

Fixes https://github.com/hashgraph/hedera-mirror-node/issues/6585


hedera-mirror-web3/src/main/java/com/hedera/services/store/models/TokenRelationship.java
```diff
@@ -96,12 +96,23 @@ public TokenRelationship(Token token, Account account, boolean hasAssociation) {
                 token.isFrozenByDefault() && token.hasFreezeKey(),
                 !token.hasKycKey(),
                 false,
-                true,
```
IvanKavaldzhiev, line 99: I believe the default value for `notYetPersisted` should be `true`. If this causes a problem, the issue might be elsewhere and fixed at the appropriate place.

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/TokenRelationship.java
```diff
@@ -96,12 +96,23 @@ public TokenRelationship(Token token, Account account, boolean hasAssociation) {
...
 
+    /**
+     * Constructor for creating new TokenRelationship with AssociatePrecompile
+     * @param token
+     * @param account
+     * @param hasAssociation
+     * @param frozen
+     */
+    public TokenRelationship(Token token, Account account, boolean hasAssociation, boolean frozen) {
+        this(token, account, 0, frozen, true, false, false, false, hasAssociation, 0);
```
IvanKavaldzhiev, line 113: Here we assume `kycGranted` is `true`. I'm not sure that would be always correct. Probably we can also pass this value as a parameter.

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/TokenRelationship.java
bilyana-gospodinova: Changed to true. There are no failing tests after the change.

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/TokenRelationship.java
bilyana-gospodinova: kycGranted is now passed as a parameter. The value is synced from services.

hedera-mirror-web3/src/test/resources/contracts/NestedEthCallsTestContract/NestedEthCallsTestContract.sol
```diff
@@ -0,0 +1,175 @@
+pragma solidity ^0.8.0;
+
+import "./HederaTokenService.sol";
+import "./HederaResponseCodes.sol";
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+
+contract NestedEthCalls is HederaTokenService {
```
IvanKavaldzhiev, line 7: We could make this smart contract agnostic of what type of calls are executed over it - `eth_call` or `eth_estimateGas` and this contract naming could be made more universal. So we can rename it to `NestedCalls` only.

bilyana-gospodinova: The contract is now renamed.

IvanKavaldzhiev (dismissed):
LGTM

0xivanov (approved):
LGTM

steven-sheehy (approved):

PR #6655, opened 2023-08-14
Update History Upserts to persist existing non-updatable values

### LINKED ISSUE #6504: The upsert framework may set history row's created_timestamp wrong
### Description

A hollow account can be finalized into a contract if the contract create transaction's new CREATE2 address matches the hollow account's EVM address. This is the first instance in which the entity's created timestamp should be when the hollow account is created even though importer creates an `Entity` object with type = contract and created_timestamp = the consensus_timestamp of the contract create tx.

However, when

- The hollow account is created in record file X
- The contract is created in record file Y
- The contract is updated in the same record file Y following the contract creation

The history row created for the create contract tx will have incorrect created timestamp.

### Steps to reproduce

Add the following test case to `SqlEntityListenerTest`, it'll fail the second assertion.

```java
    @Test
    void onEntityWithHistoryIncorrectCreatedTimestamp() {
        var entity1 = domainBuilder.entity().get();
        sqlEntityListener.onEntity(entity1);
        completeFileAndCommit();

        long timestamp = entity1.getCreatedTimestamp() + 12345L;
        var entity2 = entity1.toBuilder()
                .createdTimestamp(timestamp)
                .timestampRange(Range.atLeast(timestamp))
                .build();
        var entity3 = entity2.toBuilder()
                .createdTimestamp(null)
                .timestampRange(Range.atLeast(timestamp + 5555L))
                .build();
        sqlEntityListener.onEntity(entity2);
        sqlEntityListener.onEntity(entity3);
        completeFileAndCommit();

        assertThat(entityRepository.findAll()).hasSize(1)
                .first().returns(entity1.getCreatedTimestamp(), Entity::getCreatedTimestamp);
        assertThat(findHistory(Entity.class))
                .extracting("createdTimestamp")
                .containsOnly(entity1.getCreatedTimestamp());
    }
```

### Additional context

_No response_

### Hedera network

other

### Version

v0.86.0-SNAPSHOT

### Operating system

None

edwin-greene:
**Description**:
History table upserts for non-updatable values will now coalesce the existing value with the updated value instead of vice versa.

**Related issue(s)**:

Fixes #6504 

**Notes for reviewer**:

**Checklist**


jnels124 (dismissed):
LGTM

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/repository/upsert/EntityMetadata.java
```diff
@@ -51,4 +51,15 @@ public String columns(Predicate<ColumnMetadata> filter, String pattern) {
...
+    public String columns(String defaultPattern, Predicate<ColumnMetadata> predicate, String predicatePattern) {
+        return columns(defaultPattern, predicate, predicatePattern, ",");
+    }
+
+    public String columns(
+            String defaultPattern, Predicate<ColumnMetadata> predicate, String predicatePattern, String separator) {
+        return columns.stream()
+                .map(c -> predicate.test(c) ? c.format(predicatePattern) : c.format(defaultPattern))
+                .collect(Collectors.joining(separator));
+    }
```
xin-hedera, line 64: do we need a different separator? if no, just remove the second `columns` function and use `,`

edwin-greene: Sounds good, I have removed the additional `columns` method.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/repository/upsert/GenericUpsertQueryGenerator.java
```diff
@@ -77,6 +77,9 @@ public String getUpsertQuery() {
 
         // {0} is column name and {1} is column default. t or blank is the temporary table alias and e is the existing.
         velocityContext.put("coalesceColumns", metadata.columns("coalesce({0}, e_{0}, {1})"));
+        velocityContext.put(
+                "coalesceHistoryColumns",
+                metadata.columns("coalesce({0}, e_{0}, {1})", c -> !c.isUpdatable(), "coalesce(e_{0}, {0}, {1})"));
```
steven-sheehy, line 82: This syntax is a bit wonky with two patterns separated by a predicate. From the signature it's not clear how this works in contrast to the existing method and it's not clear which predicate value uses what pattern. I think it would be cleaner if we reuse the existing method and create two separate lists that we combine:

```suggestion
        var coalesceHistoryColumns = metadata.colummns("coalesce({0}, e_{0}, {1})", EntityMetadata::isUpdatable);
        coalesceHistoryColumns += "," + metadata.colummns("coalesce(e_{0}, {0}, {1})", c -> !c.isUpdatable());

        // {0} is column name and {1} is column default. t or blank is the temporary table alias and e is the existing.
        velocityContext.put("coalesceColumns", metadata.columns("coalesce({0}, e_{0}, {1})"));
        velocityContext.put("coalesceHistoryColumns", coalesceHistoryColumns));
```

edwin-greene: I investigated this. The statement that is used by the upsert sql is: 
```
  insert into $historyTable ($insertColumns)
  select distinct $coalesceHistoryColumns
  ```
  
  So the column order of the `insertColumns` and the `coalesceHistoryColumns` have to match, and concatenating the two calls to `columns` for `coalesceHistoryColumns` results in a different column ordering. 
  
  Another solution might be to change the newly added `columns` method and signature to something more descriptive like `columnsOrElse(Predicate<ColumnMetadata> predicate, String predicatePattern, String elsePattern)`

steven-sheehy: Ok, if it doesn't work then we can leave as is.

steven-sheehy (approved):
LGTM

xin-hedera (approved):
LGTM

xin-hedera (approved):
LGTM

PR #6670, opened 2023-08-17
Add persistence for TransactionRecord

### LINKED ISSUE #432: Store raw TransactionRecord in database
**Problem**

Auditors require the ability to access the raw `TransactionRecord` bytes, historically, for all transactions, and that data is currently not stored in the DB. This is needed in order to be able to generate a hash of the `TransactionRecord` raw bytes which will be used in state proof validation (retrieval of a state proof based on the hash of a `TransactionRecord`). Storing just the transaction record hash is not sufficient as someone could have maliciously modified that data in the mirror node database.

**Solution**

* Add a new `transaction.transaction_record_bytes` column
* Add a new `hedera.mirror.importer.parser.record.entity.persist.transactionRecordBytes` property with a default of `false`
* Persist the `recordItem.getRecordBytes()` if the property is true

**Alternatives**

Store just the hash of the raw TransactionRecord to avoid excess data in the database.

**Additional Context**

This is an auditor-related requirement coming in c/o @gregscullard. All fields from the TransactionRecord are already stored as separate columns in the database, but regenerating the exact raw bytes via protobuf serialization, in order to create the hash of the TransactionRecord is not guaranteed because protobuf serialization is non-deterministic.

edwin-greene:
**Description**:
The TransactionRecord can now be persisted to the database if the `transactionRecordBytes` property is set to true. (It is set to false by default.)

**Related issue(s)**:

Fixes #432 

**Notes for reviewer**:

**Checklist**


hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/entity/EntityProperties.java
```diff
@@ -98,6 +98,12 @@ public static class PersistProperties {
          */
         private boolean transactionBytes = false;
 
+        /**
+         * If configured the mirror node will store the raw transaction record bytes on the transaction table.
+         * The transaction record bytes include the signatures used for stateproof validation.
```
steven-sheehy, line 103: Remove this second line. It is not accurate.

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.86.0__add_transaction_record_bytes.sql
steven-sheehy, line 1: 1.85.3

steven-sheehy (changes requested):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/entity/EntityProperties.java
edwin-greene: Got it, I have removed it.

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.86.0__add_transaction_record_bytes.sql
edwin-greene: Thanks I have renamed it.

steven-sheehy (approved):
LGTM

jascks (approved):
LGTM

xin-hedera (approved):
LGTM

PR #6720, opened 2023-08-23
HIP-584: Aliases are not resolved in web3

### LINKED ISSUE #6707: HIP-584 : aliases are not resolved in web3
### Description

During precompile execution when using alias addresses we convert them to `long zero` addresses. All of the precompiles should work with `long zero` addresses, but there are some places where the addresses are not resolved to `long zero`, which breaks functionalities like `approve`, `transfer`, etc .

### Steps to reproduce

none

### Additional context

none

### Hedera network

testnet

### Version

0.86.1

### Operating system

None

0xivanov:
**Description**: 

During precompile execution when using alias addresses we convert them to long zero addresses. All of the precompiles should work with long zero addresses, but there are some places (usage of `frame.getSender`) where the addresses are not resolved to long zero, which breaks functionalities like approve, transfer, etc .

This PR also addresses some bugs regarding `ApprovePrecompile` and `CreateLogic`

- [x] ApprovePrecompile has unnecessary check that reverts the transaction incorrectly
- [x] ApprovePrecompile does not set the `nftOwnerId` correctly
- [x] Added check for zero address in `StoreImpl` in case the owner of NFT is the zero address
- [x] `AssociateLogic` is not used as bean from Spring context

**Related issue(s)**:

Fixes #6707

**Notes for reviewer**:

**Checklist**


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/StoreImpl.java
```diff
@@ -173,6 +173,7 @@ public boolean hasAssociation(TokenRelationshipKey tokenRelationshipKey) {
 
     @Override
     public boolean hasApprovedForAll(Address ownerAddress, AccountID operatorId, TokenID tokenId) {
+        if (Address.ZERO.equals(ownerAddress)) return false;
```
steven-sheehy, line 176: Standard is to always use braces. Please check your auto format settings.
```suggestion
        if (Address.ZERO.equals(ownerAddress)) {
            return false;
        }

```

hedera-mirror-web3/src/main/java/com/hedera/services/store/contracts/precompile/impl/TokenCreatePrecompile.java
```diff
@@ -270,12 +271,15 @@ public long getMinimumFeeInTinybars(Timestamp consensusTime, TransactionBody tra
     @Override
     public RunResult run(MessageFrame frame, TransactionBody transactionBody) {
         final var store = ((HederaEvmStackedWorldStateUpdater) frame.getWorldUpdater()).getStore();
+        final var updater = ((HederaEvmStackedWorldStateUpdater) frame.getWorldUpdater());
         final var tokenCreateOp = transactionBody.getTokenCreation();
+        final var senderAddress = Address.wrap(Bytes.wrap(
+                updater.permissivelyUnaliased(frame.getSenderAddress().toArray())));
```
steven-sheehy, line 277: We should have a test for this here and in other precompiles.

hedera-mirror-web3/src/main/java/com/hedera/services/store/contracts/precompile/impl/SetApprovalForAllPrecompile.java
```diff
@@ -117,9 +117,11 @@ public RunResult run(final MessageFrame frame, TransactionBody transactionBody)
 
         final var updater = ((HederaEvmStackedWorldStateUpdater) frame.getWorldUpdater());
         final var store = updater.getStore();
+        final var senderAddress = Address.wrap(Bytes.wrap(
```
steven-sheehy, line 120: Perhaps we should add a helper method to `AbstractWritePrecompile` since this logic appears in multiple sub classes?

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/services/store/contracts/precompile/impl/SetApprovalForAllPrecompile.java
0xivanov: The problem with this approach is that we need access to the `frame`, which is available in the `run` method. `AbstractWritePrecompile` does not have access to the frame as it is implemented now.

0xivanov: Added

hedera-mirror-web3/src/main/java/com/hedera/services/store/contracts/precompile/impl/TokenCreatePrecompile.java
0xivanov: Added `DYNAMIC_ETH_CALLS_CONTRACT_ALIAS` that should cover these scenarios. We will add more tests in this contract.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/StoreImpl.java
0xivanov: Fixed

hedera-mirror-web3/src/main/java/com/hedera/services/store/contracts/precompile/impl/TokenCreatePrecompile.java
0xivanov: Also removed some duplicate tests, since we pass only aliases in the calldata

steven-sheehy (approved):
LGTM

PR #6722, opened 2023-08-23
Fix HTS allowance when transferring using Ethereum Tools

### LINKED ISSUE #6721: HTS Allowances are not updated after transfer
### Description

HTS Allowance are not updated when transferFrom is executed, if token is used as ERC20 by Ethereum tools.
Issue is [here](https://github.com/hashgraph/hedera-mirror-node/blob/main/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/entity/EntityRecordItemListener.java#L532C1-L532C1), we are setting the payerAccount as sender, which is wrong in this case. (would be correct, if we make this transfer using the SDK)
It's wrong because we are making contract call using the relay in which case payerAccount is relay operator, but in reality it's not the correct sender. 
Correct sender can be found by checking if this recordItem has contractCallResult, if it does, there the real sender can be found.

Introducing this check, fixes the issue and updates the correct allowances: 
```
boolean isContractCall = recordItem.getTransactionRecord().hasContractCallResult();
long senderId = payerAccountId.getId();
 if (isContractCall) {
         senderId = recordItem.getTransactionRecord().getContractCallResult().getSenderId().getAccountNum();
}
```

### Steps to reproduce

1. Create HTS token
2. Using ethers initilize it as contract following the ERC20 interface.
3. Approve 100 tokens and check allowance
4. TransferFrom using the approved account from X address to Y address.
5. Check allowance to find that it's not updated and it's still 100 tokens.

### Additional context

_No response_

### Hedera network

testnet, previewnet, other

### Version

main

### Operating system

None

georgi-l95:
**Description**:
This PR fixes HTS allowance for tokens used by ethereum tools by applying ERC20 interface, this happens specifically when transferring with `transferFrom` method.

**Related issue(s)**:

Fixes #6721 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/entity/EntityRecordItemListener.java
+++ b/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/entity/EntityRecordItemListener.java
@@ -519,22 +519,20 @@ public class EntityRecordItemListener implements RecordItemListener {
         }
 
         var tokenTransfers = recordItem.getTransactionBody().getCryptoTransfer().getTokenTransfersList();
-        tokenTransfers.forEach(tokenTransfer -> {
+        long senderId = payerAccountId.getId();
+        if (!tokenTransfers.isEmpty() && recordItem.getTransactionRecord().hasContractCallResult()) {
+            senderId = recordItem
+                    .getTransactionRecord()
+                    .getContractCallResult()
+                    .getSenderId()
+                    .getAccountNum();
+        }
+
+        for (TokenTransferList tokenTransfer : tokenTransfers) {
             var tokenId = EntityId.of(tokenTransfer.getToken());
-            tokenTransfer.getTransfersList().forEach(accountAmount -> {
-                // Emit allowance amount representing approved transfer debit
+            for (AccountAmount accountAmount :
+                    tokenTransfer.getTransfersList()) { // Emit allowance amount representing approved transfer debit
                 if (accountAmount.getIsApproval() && accountAmount.getAmount() < 0) {
-                    boolean isContractCall = recordItem.getTransactionRecord().hasContractCallResult();
-                    long senderId = payerAccountId.getId();
-
-                    if (isContractCall) {
-                        senderId = recordItem
-                                .getTransactionRecord()
-                                .getContractCallResult()
-                                .getSenderId()
-                                .getAccountNum();
-                    }
-
                     var tokenAllowance = TokenAllowance.builder()
                             .amount(accountAmount.getAmount())
                             .owner(EntityId.of(accountAmount.getAccountID()).getId())
@@ -545,8 +543,8 @@ public class EntityRecordItemListener implements RecordItemListener {
 
                     entityListener.onTokenAllowance(tokenAllowance);
                 }
-            });
-        });
+            }
+        }
     }
 
     private void insertNonFungibleTokenTransfers(
--- a/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/entity/EntityRecordItemListener.java
+++ b/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/entity/EntityRecordItemListener.java
@@ -519,32 +519,32 @@ public class EntityRecordItemListener implements RecordItemListener {
         }
 
         var tokenTransfers = recordItem.getTransactionBody().getCryptoTransfer().getTokenTransfersList();
-        long senderId = payerAccountId.getId();
+        long spenderId = payerAccountId.getId();
         if (!tokenTransfers.isEmpty() && recordItem.getTransactionRecord().hasContractCallResult()) {
-            senderId = EntityId.of(recordItem
-                            .getTransactionRecord()
-                            .getContractCallResult()
-                            .getSenderId())
-                    .getId();
+            spenderId = recordItem
+                    .getTransactionRecord()
+                    .getContractCallResult()
+                    .getSenderId()
+                    .getAccountNum();
         }
-
-        for (TokenTransferList tokenTransfer : tokenTransfers) {
+        long transferSenderId = spenderId;
+        tokenTransfers.forEach(tokenTransfer -> {
             var tokenId = EntityId.of(tokenTransfer.getToken());
-            for (AccountAmount accountAmount :
-                    tokenTransfer.getTransfersList()) { // Emit allowance amount representing approved transfer debit
+            tokenTransfer.getTransfersList().forEach(accountAmount -> {
+                // Emit allowance amount representing approved transfer debit
                 if (accountAmount.getIsApproval() && accountAmount.getAmount() < 0) {
                     var tokenAllowance = TokenAllowance.builder()
                             .amount(accountAmount.getAmount())
                             .owner(EntityId.of(accountAmount.getAccountID()).getId())
                             .payerAccountId(payerAccountId)
-                            .spender(senderId)
+                            .spender(transferSenderId)
                             .tokenId(tokenId.getId())
                             .build();
 
                     entityListener.onTokenAllowance(tokenAllowance);
                 }
-            }
-        }
+            });
+        });
     }
 
     private void insertNonFungibleTokenTransfers(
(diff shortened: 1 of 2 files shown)


hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/entity/EntityRecordItemListener.java
```diff
@@ -521,15 +521,25 @@ private void insertTokenTransfers(RecordItem recordItem, Transaction transaction
...
+                    boolean isContractCall = recordItem.getTransactionRecord().hasContractCallResult();
+                    long senderId = payerAccountId.getId();
+
+                    if (isContractCall) {
+                        senderId = recordItem
+                                .getTransactionRecord()
+                                .getContractCallResult()
+                                .getSenderId()
+                                .getAccountNum();
+                    }
```
xin-hedera, line 536: move the logic of senderId to line 522

1. initialize `senderId` to `payerAccountId.getId()`
2. if tokenTransfers is not empty and there is contratct call result in transaction record, set senderId to `EntityId.of(contractCallResult().getSenderId()).getId()`
3. then in the inner loop, set senderId for transfers whose isApproval is true

xin-hedera (changes requested):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/entity/EntityRecordItemListener.java
georgi-l95: I've updated the logic, can you check to see if that's what you meant ?

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/entity/EntityRecordItemListener.java
```diff
@@ -519,24 +519,32 @@ private void insertTokenTransfers(RecordItem recordItem, Transaction transaction
...
         var tokenTransfers = recordItem.getTransactionBody().getCryptoTransfer().getTokenTransfersList();
-        tokenTransfers.forEach(tokenTransfer -> {
-            var tokenId = EntityId.of(tokenTransfer.getToken());
+        long senderId = payerAccountId.getId();
+        if (!tokenTransfers.isEmpty() && recordItem.getTransactionRecord().hasContractCallResult()) {
+            senderId = recordItem
+                    .getTransactionRecord()
+                    .getContractCallResult()
+                    .getSenderId()
+                    .getAccountNum();
```
xin-hedera, line 528: should be:
```suggestion
            senderId = EntityId.of(recordItem
                    .getTransactionRecord()
                    .getContractCallResult()
                    .getSenderId())
                    .getId();
```

georgi-l95: ah, yes. makes more sense. thank you, fixed.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/entity/EntityRecordItemListener.java
```diff
@@ -519,24 +519,32 @@ private void insertTokenTransfers(RecordItem recordItem, Transaction transaction
         }
 
         var tokenTransfers = recordItem.getTransactionBody().getCryptoTransfer().getTokenTransfersList();
-        tokenTransfers.forEach(tokenTransfer -> {
-            var tokenId = EntityId.of(tokenTransfer.getToken());
+        long senderId = payerAccountId.getId();
+        if (!tokenTransfers.isEmpty() && recordItem.getTransactionRecord().hasContractCallResult()) {
```
steven-sheehy, line 523: Shouldn't we also be checking `recordItem.getTransactionRecord().hasContractCreateResult()` as well?

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/entity/EntityRecordItemListener.java
```diff
@@ -519,24 +519,32 @@ private void insertTokenTransfers(RecordItem recordItem, Transaction transaction
...
+            senderId = EntityId.of(recordItem
+                            .getTransactionRecord()
+                            .getContractCallResult()
+                            .getSenderId())
+                    .getId();
+        }
 
-            tokenTransfer.getTransfersList().forEach(accountAmount -> {
-                // Emit allowance amount representing approved transfer debit
+        for (TokenTransferList tokenTransfer : tokenTransfers) {
```
steven-sheehy, line 531: An iterator based for loop is slower than a forEach. This is a hot path that needs the optimal approach. Please revert both loops to the previous form.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/entity/EntityRecordItemListener.java
```diff
@@ -519,24 +519,32 @@ private void insertTokenTransfers(RecordItem recordItem, Transaction transaction
...
+            var tokenId = EntityId.of(tokenTransfer.getToken());
+            for (AccountAmount accountAmount :
+                    tokenTransfer.getTransfersList()) { // Emit allowance amount representing approved transfer debit
```
steven-sheehy, line 534: Move comment inside the loop so it doesn't cause the line to split across multiple.

steven-sheehy (changes requested):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/entity/EntityRecordItemListener.java
georgi-l95: Not necessary, because to my knowledge only contract call operations can and should alter allowance.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/entity/EntityRecordItemListener.java
georgi-l95: Moved.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/entity/EntityRecordItemListener.java
georgi-l95: Reverted to how it was. But now added one new final variable to hold the reference, because it's lambda and java is complaining.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/entity/EntityRecordItemListener.java
```diff
@@ -519,17 +519,25 @@ private void insertTokenTransfers(RecordItem recordItem, Transaction transaction
         }
 
         var tokenTransfers = recordItem.getTransactionBody().getCryptoTransfer().getTokenTransfersList();
+        long spenderId = payerAccountId.getId();
+        if (!tokenTransfers.isEmpty() && recordItem.getTransactionRecord().hasContractCallResult()) {
+            spenderId = recordItem
+                    .getTransactionRecord()
+                    .getContractCallResult()
+                    .getSenderId()
+                    .getAccountNum();
```
steven-sheehy, line 528: Have to re-fix:

```suggestion
            spenderId = EntityId.of(recordItem
                    .getTransactionRecord()
                    .getContractCallResult()
                    .getSenderId())
                    .getId();
```

georgi-l95: Yes, you're right. Sorry about that.

steven-sheehy (approved):
LGTM

xin-hedera (approved):
LGTM

steven-sheehy:
We'll need a followup ticket for the migration which can wait for a future sprint. Can you please open that and link to a problem transaction in hashscan?

PR #6732, opened 2023-08-24
Change balance migrations to not use account balance file

### LINKED ISSUE #6678: Change TokenAccountBalanceMigration to not use account_balance_file
### Problem

`TokenAccountBalanceMigration` still uses the account_balance_file which will be going away soon.

### Solution

Replace `account_balance_file` CTE in favor of doing the timestamp filter directly on `account_balance`.

### Alternatives

_No response_

jnels124:
**Description**:
* update balance init migrations to use account_balance and token_balance for timestamp range filter

**Related issue(s)**:

Fixes #6678 

**Notes for reviewer**:

**Checklist**


xin-hedera:
main problem is we should not change the balance migrations to not use `account_balance_file` table because it doesn't make sense to use generated balance info to fix / initialize itself.

PR #6743, opened 2023-08-26
Change network supply REST API to not use account balance file

### LINKED ISSUE #6681: Change network supply REST API to not use account_balance_file
### Problem

The network stake REST API still uses the `account_balance_file` which will be going away soon.

### Solution

Replace `account_balance_file` subquery in favor of doing the timestamp filter directly on `account_balance`.

### Alternatives

_No response_

jnels124:
**Description**:
changes timestamp filter to use account_balance instead of account_balance_file

**Related issue(s)**:

Fixes #6681

**Notes for reviewer**:

**Checklist**


steven-sheehy (commented):
Need to update the specs to remove the insertion of the balance file or ensure it's not getting inserted automatically via test code.

jnels124:
I checked the spec files and there were not any specs inserting balance files. However, I did miss that when inserting an account_balance, it will also insert the balance_file. I will update that

xin-hedera (approved):
LGTM

Issue #6744, opened 2023-08-26
Update Kubernetes awareness property to support Spring Boot v3


wdstorer-bg:
### Problem

Ever since this repo was updated to Spring Boot v3, Kubernetes awareness is no longer disabled by default. This is because `bootstrap.yml` is setting a property that was removed in Spring Boot v3.

This property no longer works:
```
spring:
  cloud:
    kubernetes:
      enabled: false
```

Source: https://github.com/spring-cloud/spring-cloud-release/wiki/Spring-Cloud-2022.0-Release-Notes#breaking-changes

### Solution

The existing configuration should be replaced with the new property for disabling Kubernetes awareness:
```
spring:
  main:
    cloud-platform: NONE
```

I'll submit a corresponding PR for consideration.

PR #674, opened 2020-04-09
Fix Spring Cloud Kubernetes running locally

steven-sheehy:
**Detailed description**:
- Disables Spring Cloud Kubernetes by default to fix warnings and failures when running locally via java -jar
- Enabled Spring Cloud Kubernetes explicitly in the Helm Chart
- Will open ticket upstream to request Kubernetes auto configuration be disabled if not running within Kubernetes

**Which issue(s) this PR fixes**:

**Special notes for your reviewer**:

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

PR #6745, opened 2023-08-26
Disable Kubernetes awareness in Spring Boot v3 (#6744)

wdstorer-bg:
**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


jascks (dismissed):
LGTM

steven-sheehy (changes requested):
This will break our production helm deployments since we manually set `SPRING_CLOUD_KUBERNETES_ENABLED: "true"` there. It also misses this property in `MirrorImporterConfiguration`.

We should also update those or see if we can remove all the manual properties and rely upon the auto detection working. I think previously the Kubernetes auto detection was flaky and mis-detected Kubernetes. If testing proves it's reliable now we can simplify our configuration.

wdstorer-bg:
Thank you so much for the review! I’ve replaced the helm values with the proper environment variable. I do not see the reference you are referring to in `MirrorImporterConfiguration`.

To add additional context: We run our mirror node importer in Kubernetes. With cloud awareness enabled, our importer fails to start due to having insufficient permissions to list configmap resources. Since we don’t utilize any Spring Cloud capability we’re just disabling it.

But like you said, it’s been enabled for a couple months now so maybe the original issue has been resolved? According to https://github.com/hashgraph/hedera-mirror-node/pull/674 it was disabled to `fix warnings and failures when running locally via java -jar`. It’s easy enough for us to disable this on our end so if it’s preferable to remove it, I can close this PR and maybe help test whether we can remove it entirely?

steven-sheehy:
I tested this in Kubernetes, via `java -jar`, and via Gradle. Indeed it does look like the auto-detection has been improved so we no longer need to hardcode the property everywhere. So I think we should do the following:

* Remove both properties from the chart
* Remove both properties from the `bootstrap.yml`
* Delete the `bootstrap.yml` if it is empty
* In your local chart, disable spring cloud manually or add the necessary RBAC (allowing runtime property customization is a useful feature)
* Test

jascks (approved):
LGTM

jascks (approved):
Re-approving. And, thank you!

xin-hedera (approved):

PR #6767, opened 2023-08-28
HIP-584: Remove InvalidTransactionException dependency from services code

### LINKED ISSUE #6766: HIP-584: Remove `InvalidTransactionException` dependency from services code
### Problem

As mentioned [here](https://github.com/hashgraph/hedera-mirror-node/pull/6613#discussion_r1302237683) , services code should not have dependencies on web3 code, with the exception of the properties and Store classes. For this reason we need to replace `InvalidTransactionException` from web3 with `InvalidTransactionException` from services in the precompiles. 

This change in the type of exception also makes `HtsPrecompiledContract` to not catch the exception properly and the flow goes into the generic `catch(Exception e)` block. This can cause confusion for the end user because the mirror node sends wrong error message as result.

### Solution

Replace the `InvalidTransactionException` from web3 with `InvalidTransactionException` from services in the precompiles.

### Alternatives

_No response_

0xivanov:
**Description**:  

As mentioned https://github.com/hashgraph/hedera-mirror-node/pull/6613#discussion_r1302237683 , services code should not have dependencies on web3 code, with the exception of the properties and Store classes. For this reason we need to replace InvalidTransactionException from web3 with InvalidTransactionException from services in the precompiles.
This change in the type of exception also makes HtsPrecompiledContract to not catch the exception properly and the flow goes into the generic catch(Exception e) block. This can cause confusion for the end user because the mirror node sends wrong error message as result.
This PR also covers some refactoring discussed [here](https://github.com/hashgraph/hedera-mirror-node/pull/6720#discussion_r1304636813)

**Related issue(s)**:

Fixes #6766

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-web3/src/main/java/com/hedera/services/store/contracts/precompile/Precompile.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/services/store/contracts/precompile/Precompile.java
@@ -21,8 +21,8 @@ import static com.hederahashgraph.api.proto.java.ResponseCodeEnum.INVALID_FEE_SU
 import static org.hyperledger.besu.evm.frame.MessageFrame.State.REVERT;
 
 import com.hedera.mirror.web3.evm.store.Store;
-import com.hedera.mirror.web3.exception.InvalidTransactionException;
 import com.hedera.node.app.service.evm.accounts.HederaEvmContractAliases;
+import com.hedera.node.app.service.evm.exceptions.InvalidTransactionException;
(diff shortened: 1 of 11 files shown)


bilyana-gospodinova:
LGTM

bilyana-gospodinova (approved):

IvanKavaldzhiev (approved):
LGTM

steven-sheehy (approved):
LGTM! Thanks for taking the time to address previous feedback.

PR #6770, opened 2023-08-28
HIP-584: AssociateLogic creates association with kycGranted=false

### LINKED ISSUE #6768: HIP-584: AssociateLogic creates association with kycGranted=false
### Description

When we attempt to do an `eth_estimateGas` operation for transferFrom precompile using association web3 module returns an exception "ACCOUNT_KYC_NOT_GRANTED_FOR_TOKEN". This happens because when creating new association the kyc status is always false.

### Steps to reproduce

1. Create a Fungible Token with kyc_status = 0 (not applicable)
2. Create 2 accounts(sender&receiver).
3. Associate token with the accounts
4. Mint/supply the sender with the tokens from step 1.
5. Attempt to execute `eth_estimateGas` with `transferFrom` precompile using the token from step 1. and the accounts from step 2.

Expected result: The API returns a result with estimation
Actual result: The API returns 400 bad request and the web3 module returns an exception "ACCOUNT_KYC_NOT_GRANTED_FOR_TOKEN"

### Additional context

_No response_

### Hedera network

other

### Version

0.88.0-SNAPSHOT

### Operating system

None

0xivanov:
**Description**:  

When we attempt to do an eth_estimateGas operation for transferFrom precompile using association web3 module returns an exception "ACCOUNT_KYC_NOT_GRANTED_FOR_TOKEN". This happens because when creating new association the kyc status is always false.

This PR utilises logic in the `token` model like it is in services in order to create new token relationship with the correct kyc status

**Related issue(s)**: 

Fixes #6768

**Notes for reviewer**:

**Checklist**


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/TokenRelationshipDatabaseAccessor.java
steven-sheehy, line 1: Why are we not passing `token_account.associated` to the `TokenRelationship.hasAssociation`?

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/Token.java
```diff
@@ -1565,7 +1565,7 @@ public TokenRelationship newRelationshipWith(final Account account, final boolea
      * @return newly created {@link TokenRelationship}
      */
     public TokenRelationship newEnabledRelationship(final Account account) {
-        return new TokenRelationship(this, account, 0, false, true, false, false, false, false, 0);
+        return new TokenRelationship(this, account, 0, false, true, false, false, false, true, 0);
```
steven-sheehy, line 1568: These long constructors seem to be a frequent source of bugs. I know others have complained in PR reviews about them as well. I thought we had these because of services parity, but looking at services I only see a `TokenRelationship(Token token, Account account)` constructor, which seems reasonable. 

I think we should revisit these constructors and make them match services in a future PR. We need immutability just so we can cache these and allow modifications? Even though I'm against it usually maybe we could add `@Builder(toBuilder=true)` on just these few services models.

0xivanov: Yes, we can address this in future package.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/TokenRelationshipDatabaseAccessor.java
```diff
@@ -56,7 +55,7 @@ public class TokenRelationshipDatabaseAccessor extends DatabaseAccessor<Object,
                                         false,
                                         false,
                                         Boolean.TRUE == tokenAccount.getAutomaticAssociation(),
-                                        true,
+                                        tokenAccount.getAssociated(),
```
steven-sheehy, line 58: Need to handle null case. It would be better to just move true into the query since consensus nodes don't keep dissociated relationships in state.

0xivanov: Do you mean to revert the change to how it was originally?

steven-sheehy: I missed that we filter it above in the stream. But my concern is that I believe consensus nodes don't have a concept of a `TokenRelationship` that is in state but not associated. Or they do, but it's already covered by the `TokenRelationship.destroyed` flag, I believe. So why do we even have the `TokenRelationship.hasAssociated` flag? It should always be true except when dissociating in which case destroyed would temporarily be set true.

But besides that, yes we should probably revert the last change since it better matches services to not load dissociated relationships into memory at all.

0xivanov: Reverted the change

IvanKavaldzhiev (approved):
LGTM!

PR #6776, opened 2023-08-29
HIP-584: Successfully deploy contract with missing sender

### LINKED ISSUE #6775: HIP-584: Contract deploys without from field are failing
### Description

Currently, when we try to perform an `eth_estimateGas` for contract deploys without a `from` field they are failing.

### Steps to reproduce

1. Construct request body without `from` field and `data` with bytecode
2. Perform an `eth_estimateGas` requests against api/v1/contracts/call
3. You get an Internal server error

### Additional context

_No response_

### Hedera network

previewnet

### Version

v0.87.0

### Operating system

None

IvanKavaldzhiev:
**Description**:

This PR fixes an issue where we try to perform `eth_estimateGas` for contract deploy with missing sender.

The solution is to add a default account with zero address and positive balance in the beginning of `eth_estimateGas` executions.

**Related issue(s)**:

Fixes #6775 

**Notes for reviewer**:

**Checklist**


hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServiceTest.java
```diff
@@ -288,8 +289,24 @@ void estimateGasForCreate2ContractDeploy() {
...
+
+        assertGasUsedIsPositive(gasUsedBeforeExecution, ETH_ESTIMATE_GAS);
+    }
+
+    @Test
+    void estimateGasForDirectCreateContractDeployWithMissingSender() {
+        final var gasUsedBeforeExecution = getGasUsedBeforeExecution(ETH_ESTIMATE_GAS);
+
+        final var serviceParameters = serviceParametersForTopLevelContractCreate(
+                ETH_CALL_INIT_CONTRACT_BYTES_PATH, ETH_ESTIMATE_GAS, Address.ZERO);
```
0xivanov, line 309: Is it actually possible for the sender address to be the zero address?

IvanKavaldzhiev: If we don't explicitly set a `from` account in the request body, we assign `Address.ZERO` as a default value. Here we make tests on service level and skip the controller business logic. That's why I pass directly the potentially resolved ZERO address.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
```diff
@@ -41,7 +41,7 @@ public boolean isUsable(final Address address) {
         final var account = store.getAccount(address, OnMissing.DONT_THROW);
         final var balance = account.getBalance();
 
-        if (!account.isEmptyAccount()) {
+        if (!account.isEmptyAccount() || Address.ZERO.equals(address)) {
```
0xivanov, line 44: Can the zero address be found as a valid account in the store?

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServiceTest.java
0xivanov: I get it now, still it is strange to me that it works this way :D

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
IvanKavaldzhiev: I don't think so, the Address.ZERO cannot be linked to a valid account and is just a default value for the Address type.

0xivanov (approved):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
```diff
@@ -41,7 +41,7 @@ public boolean isUsable(final Address address) {
...
             return balance >= 0L;
```
steven-sheehy, line 45: Store will return empty account for address zero, which has a zero balance. So with your change to allow zero address `isUsable(Address.ZERO)` will now return true. Perhaps this might fix this issue but maybe cause problems elsewhere?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
```diff
@@ -41,7 +41,7 @@ public boolean isUsable(final Address address) {
         final var account = store.getAccount(address, OnMissing.DONT_THROW);
```
steven-sheehy, line 41: We should consider optimizing the database accessors to not query and return empty object if it's address zero.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
```diff
@@ -36,33 +36,34 @@ public class MirrorEntityAccess implements HederaEvmEntityAccess {
...
+        if (expirationTimestamp != 0L && expirationTimestamp <= currentTime) {
+            return false;
+        }
+
+        if (createdTimestamp == 0L) {
+            return false;
         }
+
+        return autoRenewPeriod == 0L || (createdTimestamp + autoRenewPeriod) > currentTime;
     }
```
IvanKavaldzhiev, line 67: I aligned this logic closer to `hedera-services` implementation. We should check usability of regular accounts or smart contracts. We should consider more fields than just the positive balance.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmContractCreationCallProcessor.java
```diff
@@ -0,0 +1,94 @@
...
+        }
+        try {
+
+            // Permissively allow the sender to not pay for the contract creation, allowing eth_estimateGas executions
+            // with empty or missing senders
+            final var senderAccount = frame.getWorldUpdater().getSenderAccount(frame);
+            if (senderAccount != null) {
+                final MutableAccount sender = senderAccount.getMutable();
+                sender.decrementBalance(frame.getValue());
+            }
```
IvanKavaldzhiev, line 70: I decided to fix the issue locally in the ContractCreationProcessor by overriding this method. In services it wouldn't be allowed to use zero address or missing address, but here we do simulations and we are interested in the final result, e.g. gas consumption for contract deploys. So, it makes sense to be more permissive and include more use cases, which are usually not valid against the consensus node.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
IvanKavaldzhiev: I adapted `Store.getAccount()` this way.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
IvanKavaldzhiev: There is a possibility, indeed. I removed this check from `isUsable` and instead applied a fix locally into a `ContractCreationProcessor` extension.

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/Account.java
```diff
@@ -49,27 +49,66 @@
  * 6. Added factory method that returns empty instance
  * 7. Added isEmptyAccount() method
  */
+@Getter
 public class Account extends HederaEvmAccount {
     private final Long entityId;
+
+    @Getter
```
steven-sheehy, line 56: Since you have Getter on the class it's redundant to put it on the fields.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
```diff
@@ -36,33 +36,34 @@ public class MirrorEntityAccess implements HederaEvmEntityAccess {
...
+        final var currentTime = Instant.now().getEpochSecond();
+        if (expirationTimestamp != 0L && expirationTimestamp <= currentTime) {
+            return false;
+        }
+
+        if (createdTimestamp == 0L) {
+            return false;
         }
+
+        return autoRenewPeriod == 0L || (createdTimestamp + autoRenewPeriod) > currentTime;
```
steven-sheehy, line 66: The logic here now duplicates what's in `AbstractEntity.getEffectiveTimestamp()`. Since you're populating account.expiry with the effective timestamp and not the underlying raw expiration timestamp this means this logic is now incorrect or effectively unreachable. Can't this just change to `return account.getExpiry() <= Instant.now().getEpochSecond();`?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmContractCreationCallProcessor.java
steven-sheehy: I'm not a fan of this solution since it requires maintaining a fork of code from Besu. So if upstream ever changes we'll never know unless there's a compilation failure. Instead, can we just proactively insert the missing sender into the world state just for gas estimations before this code runs?

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
IvanKavaldzhiev: I agree. However I believe it should be the other way around 

`return account.getExpiry() >= Instant.now().getEpochSecond()`

IvanKavaldzhiev:
I updated the fix to reflect adding a default account in case of `eth_estimateGas`.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmTxProcessorFacadeImpl.java
```diff
@@ -142,6 +145,11 @@ public HederaEvmTransactionProcessingResult execute(
 
         processor.setOperationTracer(mirrorOperationTracer);
 
+        // In case of eth_estimateGas we add a default account with zero address to cover cases with missing sender
+        if (isEstimate) {
+            final var defaultAccount = Account.getDefaultAccount();
+            store.updateAccount(defaultAccount);
```
steven-sheehy, line 151: Should this also have some fake balance to cover some costs?

IvanKavaldzhiev: Yes, inside `getDefaultAccount` I load the account with `1_000_000` balance.

steven-sheehy (approved):
LGTM

PR #6783, opened 2023-08-30
Fast Historical Sync PoC

### LINKED ISSUE #6413: Fast historical sync proof of concept
### Problem

As a mirror node operator, I'd like to startup a mirror node and sync from genesis in hours not months

### Solution

Create a `HistoricalDownloader` class that:
  * Uses `s3Client.listObjectsV2Paginator` and `s3Client.getObject` to download 1/3 of stake (plus a few extra) signatures in parallel to disk. [Each node in 1/3 of stake is listed, and the signature files downloaded.]
  * Uses `s3Client.getObject` to download one of those node's corresponding stream file to disk
  * Retry different nodes until successful
  * Limit total amount of data on disk to a configurable amount per stream type
  * Stop listing and downloading when limit is reached. Resume once space is freed up.

### Alternatives

_No response_

jascks:
# Description:

Fast Historical Sync, Proof of Concept.

**This PR will not leave Draft status, nor be merged into main.** It serves only to make the PoC accessible and to report some results produced by running the code. There are a few bullet items in the associated ticket that are not required at this point to just get a rough idea of performance and whether this approach should be pursued more fully.

Unless your local database already contains an address book, only the classpath genesis address book is available to the PoC. There is no code present to refresh the nodes/stake while downloading when the address book changes. More importantly, since the record files are only being downloaded and not parsed etc., the address book will not change anyway.

## Performance Results
When run using `testnet`, and the classpath genesis address book (7 nodes), stake is not associated with any node, so when 1/3 of stake is calculated, all nodes are considered equal with a stake of 1. So, that means 1/3 of stake is achieved with `n=3` of the 7 nodes.

For a given consensus timestamp, this PoC downloads `n` signature files. Then, for only one of the nodes, the data/record file is also downloaded. For the `testnet` scenario, this is 3 roughly 1000 byte signature files, and 1 varying (4-10KB and larger) data file. For `mainnet` the number of nodes has grown from 13 to 29 over the years. While this PoC would only consider the classpath genesis address book of 13 nodes, a full implementation must accommodate the growth of the address book and adjust its calculation of 1/3 stake as the address book changes over time.

Let's say that today, for `mainnet`, 1/3 of stake is `n=10` of 29 nodes. Thus, 10 signature files and one data file must be downloaded. The current upload rate for any given node is a signature file and data file every two seconds, or 43,200 signature files per day and 43,200 data files per day. 

So to keep pace,  the PoC must download per day:
`testnet`:  3 * 43200 + 43200 = 172,800
`mainnet`: 10 * 43200 + 43200 = 475,200

Of course, the goal is to run much faster than that so that a mirror node can bootstrap from genesis to current time as quickly as possible. And good news, the potential looks great! _Do keep in mind, the performance indicated here only entails downloading the required files to the local file system. Signature validation does not occur, nor are the data files parsed and the database updated._

Using `testnet`, I made several runs to download a 1/2 day, 2 days and then a week's worth of files to tune a few parameters. Download speeds varied from 154 - 2050 files/second or extrapolated out, 13,305,600 - 177,120,000 files/day. Even the slowest result far outstrips the normal production rate. After final tuning I downloaded a week's worth of files from `testnet`, which took just over 14 minutes. This is a rate of 1427 files/second or 123,292,800 files/day.

Therefore 123,292,800 / 475,200 = 259.5 days of `mainnet` data can be downloaded per day. `mainnet` has been producing  these files for about four years, or 365 * 4 = 1460 days worth. So, 1460 / 259.5 = 5.6, which is the number of days it would take to download all of `mainnet`. This does not account for the varying address book over the years, where fewer files would need to be downloaded at genesis where 1/3 of stake results = `n=5` of 13 nodes and then increasing over time.

Further improvements can be made to the code to increase performance, as well as the runtime environment being in the cloud with high network bandwidth, co-located with the data (AWS, GCP) will only make things faster.

### My Runtime Environment

- `testnet` was used as the source of files, configured using the GCP bucket access credentials defined in the `mirrornode` 1Password vault.
- My ISP provides, quite consistently, around 100Mbps download and about 18Mbps upload. Not much else was going on in the home while testing. My laptop is connected via USB-C (5Gbps) to a Dell monitor which has 1 Gb ethernet port to my Netgate firewall via a 2.5 GbE port to ISP.
- My work laptop is a 2021 16" MacbookPro (MacBookPro18,2), M1 Max (10 cores: 8 perf, 2 efficiency), 64GB RAM, MacOS Ventura 13.5.1. Other major programs running at the time were IntelliJ IDEA, Docker Desktop, Slack, lots of Chrome windows and tabs. The system was not solely dedicated to this testing.
- Java is OpenJDK 64-Bit Server VM Temurin-17.0.8+7 (build 17.0.8+7, mixed mode). [Apple Silicon native]
- The new historical downloader is enabled and the S3 client/API is configured to download S3 objects directly to files. They are not read into memory any more than the extent of the `ByteBuffer`s streamed from the service to the file system.
- All other downloaders are disabled via properties.  Therefore, no resources are spent verifying signatures, parsing record files or updating the database.

### The Test Scenario
Being a PoC, some parameters are hardcoded, while others can be more easily manipulated via properties. See *Notes for reviewer* below for some of the choices I made.

`testnet`, was reset part way through 2023-07-27. The first file is `2023-07-27T17_41_10.354040297Z.rcd.gz`. Rather than deal with a partial day, `HistoricalDownloader.java:98`  hardcodes the value `2023-07-28T00_00_00Z.rcd` as the starting point for objects listed in the bucket.

`HistoricalDownloaderProperties` (`hedera.mirror.importer.downloader.historical`) defines two new properties of interest, which are configured to my test defaults in `application.yml`:
- `stopAtPrefix`: The filename prefix when encountered downloading stops. This is a simple `String.startsWith(stopAtKeyPrefix)` implementation. To download only one week's data (starting with `2023-07-28T00_00_00Z` above, I defined the stop prefix as `2023-08-04T`.
- `downloadConcurrency`: The number of files (includes both signature and data files), per node, to download concurrently. This is set to 100. This is referenced in `HistoricalDownloader.java:166`. If you choose to increase this further, or run in an environment with more nodes for which files are to be downloaded, then also check `HistoricalDownloader.java:61` to ensure the backing `ExecutorService` thread pool is large enough to accommodate.

### Running the Historical Downloader

Note that the new property `hedera.mirror.importer.downloadPath` defaults to `../../historical-downloads` because it suited my purposes. This can be overridden. The historical downloader will create this path if it does not already exist, as well as the stream type and node specific sub-directories.
```
historical-downloads
└── recordstreams
    ├── record0.0.3
    ├── record0.0.8
    └── record0.0.9
```
The historical downloader will also overwrite any existing files, so to get a true measure of the files downloaded per run, it is advised to remove the download path (`rm -rf`) before running the downloader again.

Within my runtime environment defined earlier, I used the following command to run the Importer:
```
java -Xms2g -Xmx10g -XX:+CrashOnOutOfMemoryError -jar hedera-mirror-importer/build/libs/hedera-mirror-importer-v0.88.0-SNAPSHOT.jar --spring.config.additional-location=file:/Users/user/Development/mirrornode/config/importer/testnet.yml
```
Do note that with signature verification and data file parsing disabled, a 10g max heap should not be necessary. I tried less (1, 2, and 4g), and while they do not OOM, they are significantly slower, such as 248 seconds for 2 days of files vs 208.  It could be internet variability as well. Dedicate what memory you can on the machine you are using.

Of course, before running it, you must first compile (from the base of the repo):

```
./gradlew clean build -x test
```
Verify the name of the jar file found in `hedera-mirror-importer/build/libs`.

*NOTE:* `-x test` skips the unit tests. Not only does this make building a lot faster, unit tests will fail since the PoC is a bit of a hack and the regular Downloaders are disabled.

When using the `java` command above, the referred to additional Spring configuration is a YAML file that contains my GCP credentials as well as allows me to override settings in `application.yml` without having to recompile.  This looks like:
```
hedera:
  mirror:
    importer:
      #downloadPath: /some/path
      network: testnet
      downloader:
        cloudProvider: GCP
        accessKey: xxx
        secretKey: yyy
        gcpProjectId: zzz
        historical:
          #downloadConcurrency: 100
          #stopAtPrefix: 2023-08-04T
```
I commented out the two `historical` properties that are now defined by default with those values in `application.yml` so you can easily customize them if needed.

Finally, some important information is logged while running. If a DB migration occurred on startup, these first couple of lines may be before that. Search if you need to.
```
2023-09-06T13:09:43.316+0000 INFO scheduling-2 c.h.m.i.d.h.HistoricalDownloader Starting download from 2023-07-28T00_00_00Z.rcd until prefix 2023-08-04T for stream type RECORD, per node download concurrency of 100
2023-09-06T13:09:43.563+0000 INFO scheduling-2 c.h.m.i.d.h.HistoricalDownloader 3 of 7 node downloaders required to meet downloadRatio 0.34
2023-09-06T13:09:43.564+0000 INFO pool-10-thread-1 c.h.m.i.d.h.HistoricalDownloader Downloading files for node 5 (0.0.8)
2023-09-06T13:09:43.564+0000 INFO pool-10-thread-2 c.h.m.i.d.h.HistoricalDownloader Downloading files for node 0 (0.0.3)
2023-09-06T13:09:43.563+0000 INFO pool-10-thread-3 c.h.m.i.d.h.HistoricalDownloader Downloading files for node 6 (0.0.9)
```
This confirms your settings of `stopAtPrefix` and `downloadConcurrency`, as well as indicates what constitutes 1/3 stake and the nodes for which files are going to be downloaded. The nodes selected will change from one run to the next.

The S3 async client is used and each node lists its specific bucket contents and paginates through 1000 objects per page. This is reflected by:
```
2023-09-07T17:35:14.301+0000 INFO sdk-async-response-1-2 c.h.m.i.d.p.S3StreamFileProvider Node 0 (0.0.3) loaded page 12 containing 1000 S3 objects, starting with: recordstreams/record0.0.3/2023-07-28T03_03_22.055194003Z.rcd.gz
2023-09-07T17:35:14.310+0000 INFO sdk-async-response-1-9 c.h.m.i.d.p.S3StreamFileProvider Node 6 (0.0.9) loaded page 18 containing 1000 S3 objects, starting with: recordstreams/record0.0.9/2023-07-28T04_43_22.268103768Z.rcd.gz
2023-09-07T17:35:14.331+0000 INFO sdk-async-response-1-4 c.h.m.i.d.p.S3StreamFileProvider Node 3 (0.0.6) loaded page 14 containing 1000 S3 objects, starting with: recordstreams/record0.0.6/2023-07-28T03_36_42.086753003Z.rcd.gz
```
Note the key for the first object on each page is output as an indication of progress. Keep in mind that listing bucket contents is asynchronous from downloading those objects, and you may well see many pages listed that go well beyond your `stopAtPrefix`. Those objects will not be downloaded, however.

For a given node, for every 1000 files downloaded, a progress indicator is given:
```
2023-09-07T17:35:16.894+0000 INFO sdk-async-response-1-7 c.h.m.i.d.h.HistoricalDownloader Node 0 (0.0.3) downloaded cumulative total of 6000 files
2023-09-07T17:35:17.101+0000 INFO sdk-async-response-1-3 c.h.m.i.d.h.HistoricalDownloader Node 6 (0.0.9) downloaded cumulative total of 5000 files
```
These are intermingled with the object list paging output, until the paging is paused by the S3 client.

When all downloads have completed you are given a summary:
```
2023-09-06T13:23:50.049+0000 INFO pool-10-thread-2 c.h.m.i.d.h.HistoricalDownloader Downloaded 397034 files for node: 0 (0.0.3) in 14.11 min
2023-09-06T13:23:50.069+0000 INFO pool-10-thread-3 c.h.m.i.d.h.HistoricalDownloader Downloaded 413278 files for node: 6 (0.0.9) in 14.11 min
2023-09-06T13:23:50.498+0000 INFO pool-10-thread-1 c.h.m.i.d.h.HistoricalDownloader Downloaded 398932 files for node: 5 (0.0.8) in 14.12 min
2023-09-06T13:23:50.534+0000 INFO scheduling-2 c.h.m.i.d.h.HistoricalDownloader Total download time 14.12 min, total number of files 1209244
```
Note the java process will not exit on its own. Hit ctrl-C to kill it at this point.

That last line is the total number of files downloaded and the time it took to do it. The per node results are also reported. You can confirm this via the file system as well:
```
$ find historical-downloads -name \*.rcd_sig | wc -l                                                                                                    
  906933
$ find historical-downloads -name \*.rcd.gz | wc -l
  302311
$ du -sk  historical-downloads
5229436	historical-downloads
```
That is a total of 906,933 signature files, or 302,311 per node. And, 302,311 data files for a total of 1,209,244, resulting in about 5GB of data. Recall only one data file is downloaded from one of the nodes. 

So, you might be wondering why the per node log messages all indicate different file totals. While all signature files are downloaded for a node, only the first one to reach the point of downloading the data file does so, and the others do not. This is not deterministic and those 302,311 data files got scattered around for the 3 nodes.

# Related issue(s)

Fixes #6413 

# Notes for reviewer
Some things to keep in mind regarding this PoC. There are a number of hacks that most likely will not remain in place if we move beyond PoC and work on integrating this functionality into main.

- The PoC is executed by running the Importer. All other downloaders have their `enable` property set to false. `HistoricalDownloader` is enabled.
- Like the other downloaders, `HistoricalDownloader` is triggered via Spring `@Scheduled`, though only to run once and not to repeat.
- At one point I got frustrated trying to get ahold of the `AsyncS3Client` instance used by `S3StreamFileProvider` to access S3. In the end, I chose to update `StreamFileProvider` and define two methods to do the work I required. This also means that `CompositeStreamFileProvider` and `S3StreamFileProvider` were modified to implement things. It is what it is for the PoC.
- Therefore, the main body of work is `HistoricalDownloader` and `S3StreamFileProvider`.
- Some code was cut and pasted into `HistoricalDownloader` to get things done w/o refactoring existing code etc. just for the PoC.
- Everything is reactor based. Objects are downloaded from S3 directly to files using `AsyncResponseTransformer.toFile()`. `MirrorProperties.downloadPath` defines the top-level directory in which files are created. The hierarchy follows the S3 key/prefix structure defined in the bucket. Directories are created if not present.
- No additional tests have been written for the PoC.

**Checklist**

### DIFF
--- a/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/historical/HistoricalDownloader.java
+++ b/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/historical/HistoricalDownloader.java
@@ -16,6 +16,9 @@
 
 package com.hedera.mirror.importer.downloader.historical;
 
+import static com.hedera.mirror.importer.config.CacheConfiguration.CACHE_MANAGER_STREAM_DOWNLOADER;
+import static java.util.stream.Collectors.groupingBy;
+
 import com.google.common.base.Stopwatch;
 import com.hedera.mirror.common.domain.StreamType;
 import com.hedera.mirror.importer.addressbook.ConsensusNode;
@@ -23,7 +26,6 @@ import com.hedera.mirror.importer.addressbook.ConsensusNodeService;
 import com.hedera.mirror.importer.domain.StreamFilename;
 import com.hedera.mirror.importer.downloader.DownloaderProperties;
 import com.hedera.mirror.importer.downloader.provider.StreamFileProvider;
-import com.hedera.mirror.importer.downloader.provider.StreamFileProvider.GetObjectResponseWithKey;
 import com.hedera.mirror.importer.exception.FileOperationException;
 import com.hedera.mirror.importer.leader.Leader;
 import jakarta.inject.Named;
@@ -31,23 +33,18 @@ import java.io.IOException;
 import java.math.BigDecimal;
 import java.math.RoundingMode;
 import java.nio.file.Path;
-import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
-import java.util.Set;
+import java.util.Map;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
-import java.util.concurrent.ConcurrentSkipListMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
-import java.util.concurrent.atomic.AtomicInteger;
-import java.util.concurrent.atomic.AtomicReference;
 import lombok.CustomLog;
-import lombok.Value;
 import org.apache.commons.io.FileUtils;
+import org.springframework.cache.Cache;
+import org.springframework.cache.CacheManager;
 import org.springframework.scheduling.annotation.Scheduled;
 import software.amazon.awssdk.services.s3.model.S3Object;
 
@@ -56,6 +53,7 @@ import software.amazon.awssdk.services.s3.model.S3Object;
 public class HistoricalDownloader {
 
     public static final String SEPARATOR = "/";
+    private static final String TEMPLATE_ACCOUNT_ID_PREFIX = "%s/%s%s/";
 
     private final ConsensusNodeService consensusNodeService;
     private final Path downloadPath;
@@ -63,34 +61,31 @@ public class HistoricalDownloader {
     private final ExecutorService executorService = Executors.newFixedThreadPool(100);
     private final StreamFileProvider streamFileProvider;
     private final StreamType streamType;
-    private final ConcurrentMap<String, FileSpecificInfo> downloadsInfoMap = new ConcurrentSkipListMap<>();
-    private final AtomicReference<ConsensusNodeInfo> nodeInfoRef = new AtomicReference<>();
+    private final Cache dataDownloadsCache;
     private final int downloadConcurrency = 3; // Debug, make a property
 
     public HistoricalDownloader(
             ConsensusNodeService consensusNodeService,
             HistoricalDownloaderProperties downloaderProperties,
-            StreamFileProvider streamFileProvider) {
+            StreamFileProvider streamFileProvider,
+            @Named(CACHE_MANAGER_STREAM_DOWNLOADER) CacheManager cacheManager) {
 
         this.consensusNodeService = consensusNodeService;
+        this.dataDownloadsCache = cacheManager.getCache("dataDownloads");
         this.downloaderProperties = downloaderProperties;
         this.streamFileProvider = streamFileProvider;
         this.streamType = downloaderProperties.getStreamType();
         this.downloadPath = downloaderProperties.getMirrorProperties().getDownloadPath();
     }
 
+    private static String getPrefix(ConsensusNode node, StreamType streamType) {
+        var nodeAccount = node.getNodeAccountId().toString();
+        return TEMPLATE_ACCOUNT_ID_PREFIX.formatted(streamType.getPath(), streamType.getNodePrefix(), nodeAccount);
+    }
+
     private static String s3Basename(String s3Key) {
         var lastSeparatorIndex = s3Key.lastIndexOf(SEPARATOR);
-        return lastSeparatorIndex < 0 ? s3Key : s3Key.substring(lastSeparatorIndex + 1);
-    }
-
-    private static String s3Prefix(String s3Key) {
-        var lastSeparatorIndex = s3Key.lastIndexOf(SEPARATOR);
-        return lastSeparatorIndex < 0 ? null : s3Key.substring(0, lastSeparatorIndex + 1);
-    }
-
-    private static boolean isSignatureFileObject(S3Object s3Object) {
-        return isSignatureFileName(s3Object.key());
+        return lastSeparatorIndex < 0 ? s3Key : s3Key.substring(lastSeparatorIndex + 1); // keep trailing /
     }
 
     private static boolean isSignatureFileName(String filename) {
@@ -101,13 +96,13 @@ public class HistoricalDownloader {
     // Run once
     @Scheduled(initialDelay = 1000 * 5, fixedDelay = Long.MAX_VALUE)
     public void downloadAll() {
-        var stopwatch = Stopwatch.createStarted();
-        log.info("Starting download from epoc for stream type {}", streamType);
         downloadAll(StreamFilename.EPOCH);
-        log.info("Total download time {}", stopwatch);
     }
 
     public void downloadAll(StreamFilename startFilename) {
+
+        var methodStopwatch = Stopwatch.createStarted();
+        log.info("Starting download from {}} for stream type {}", startFilename, streamType);
 
         /* NOTE: For first part (6413), the address book will not change while downloading since the data files
          * are not being imported.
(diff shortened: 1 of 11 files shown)


hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/historical/HistoricalDownloader.java
```diff
@@ -0,0 +1,250 @@
...
+                                    "Unable to create local directory %s".formatted(filesDownloadDir), e);
+                        }
+
+                        var fileCounter = new AtomicLong(0L);
+                        try {
+                            var count = streamFileProvider
+                                    .listAllPaginated(node, startFilename)
+                                    .filter(this::isDownloadProspect)
+                                    .flatMap(streamFileProvider::getAsFile, downloadConcurrency)
+                                    .doOnNext(response -> {
```
jascks, line 158: This is only to provide some feedback/heatbeat during a long running process.

steven-sheehy:
The PoC was a success in proving we could quickly download lots of files at a rate quicker than we can ingest. There will be follow up tickets to integrate a historical downloader into the codebase.

PR #6786, opened 2023-08-31
Generate balance info from record file

### LINKED ISSUE #6700: Generate balance info from record file
### Problem

Since the balance file will be going away soon, we need to change how the balance tables are populated.

### Solution

* Add a `hedera.mirror.importer.parser.record.historicalBalances.enabled=true`
* Add a `hedera.mirror.importer.parser.record.historicalBalances.maxFrequency=15m`
* Disable `hedera.mirror.importer.downloader.balance.enabled`
* Ensure even if downloader balance is disabled, the first balance file is downloaded and processed if the table is empty
* Add a `HistoricalBalanceService`
* Notify `HistoricalBalanceService` asynchronously after every record file is completed
* Ignore record file if its consensus end does not exceed the last execution plus the balance max frequency
* Execute a SQL that will loop through all active accounts and contracts and insert an `account_balance`
* Execute a SQL that will loop through all active token accounts and insert a `token_balance`
* Use Repeatable read [isolation](https://www.postgresql.org/docs/14/transaction-iso.html) to ensure later transactions are not seen
* Handle the scenario if the balance service takes a long time, is down, or misses notification of record files that it loops over missing and catches up

### Alternatives

_No response_

xin-hedera:
**Description**:

This PR adds support of generating balance info from record file

- Add `HistoricalBalancesService` to generate account balance and token balance
- Trigger `HistoricalBalancesService` on each parsed record file
- Add `synthetic` column to `account_balance_file` table
- Add `RecordFileRepository.findFirst()`
- Update `InitializeEntityBalanceMigration` and `TokenAccountBalanceMigration` to skip synthetic account balance file rows
- Limit downloader to process at most 1 file when it's disabled
- Fix `LocalStreamFileProvider.list` returning more files than asked
- Make `MirrorDateRangePropertiesProcessor` independent of whether downloader is enabled
- Disable account balance file downloader by default
- Don't generate token balances in rosetta all-in-one container image

**Related issue(s)**:

Fixes #6700 

**Notes for reviewer**:

Tested against mainnet, the longest time to generate the balances info is 76s, right after importer startup. Once more data is cached database server memory, the average time is around 40s with high at 50s.

```
2023-08-31T03:29:26.150+0000 INFO Historical balances service-1 c.h.m.i.p.r.h.HistoricalBalancesService Generating historical balances after processing record file with consensusEnd 1693378801999335304
2023-08-31T03:30:42.059+0000 INFO Historical balances service-1 c.h.m.i.p.r.h.HistoricalBalancesService Generated historical account balance file 2023-08-30T07_00_01.999335304Z_Balances.csv with 3489930 account balances and 4460167 token balances for timestamp 1693378801999335304 in 1.265 min
...
2023-08-31T04:16:00.520+0000 INFO Historical balances service-1 c.h.m.i.p.r.h.HistoricalBalancesService Generated historical account balance file 2023-08-30T14_15_35.999898590Z_Balances.csv with 3491322 account balances and 4460606 token balances for timestamp 1693404935999898590 in 36.30 s
2023-08-31T04:16:55.160+0000 INFO Historical balances service-1 c.h.m.i.p.r.h.HistoricalBalancesService Generating historical balances after processing record file with consensusEnd 1693405837981907491
2023-08-31T04:17:26.158+0000 INFO Historical balances service-1 c.h.m.i.p.r.h.HistoricalBalancesService Generated historical account balance file 2023-08-30T14_30_37.981907491Z_Balances.csv with 3491376 account balances and 4460614 token balances for timestamp 1693405837981907491 in 31.00 s
2023-08-31T04:18:23.378+0000 INFO Historical balances service-1 c.h.m.i.p.r.h.HistoricalBalancesService Generating historical balances after processing record file with consensusEnd 1693406739991374003
2023-08-31T04:19:04.012+0000 INFO Historical balances service-1 c.h.m.i.p.r.h.HistoricalBalancesService Generated historical account balance file 2023-08-30T14_45_39.991374003Z_Balances.csv with 3491432 account balances and 4460626 token balances for timestamp 1693406739991374003 in 40.63 s
2023-08-31T04:20:02.686+0000 INFO Historical balances service-1 c.h.m.i.p.r.h.HistoricalBalancesService Generating historical balances after processing record file with consensusEnd 1693407641972630003
2023-08-31T04:20:37.687+0000 INFO Historical balances service-1 c.h.m.i.p.r.h.HistoricalBalancesService Generated historical account balance file 2023-08-30T15_00_41.972630003Z_Balances.csv with 3491480 account balances and 4460638 token balances for timestamp 1693407641972630003 in 35.00 s
```
**Checklist**


hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/config/MirrorDateRangePropertiesProcessor.java
```diff
@@ -72,12 +69,6 @@ public DateRangeFilter getDateRangeFilter(StreamType type) {
     }
 
     private DateRangeFilter newDateRangeFilter(StreamType streamType) {
-        DownloaderProperties downloaderProperties = getDownloaderProperties(streamType);
-
-        if (!downloaderProperties.isEnabled()) {
-            return DateRangeFilter.empty();
-        }
```
xin-hedera, line 79: this breaks account balance file parser test since by default the downloader is disabled and the filter is empty. Making it independent of wether download is enabled shouldn't break anything

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/config/MirrorDateRangePropertiesProcessorTest.java
```diff
@@ -208,13 +204,6 @@ void startDateNotBeforeEndDate(Instant startDate, Instant endDate, Instant lastF
         }
     }
 
-    @ParameterizedTest(name = "timestamp {0} does not pass empty filter")
-    @ValueSource(longs = {-10L, 0L, 1L, 10L, 8L, 100L})
-    void emptyFilter(long timestamp) {
-        DateRangeFilter filter = DateRangeFilter.empty();
-        assertThat(filter.filter(timestamp)).isFalse();
-    }
-
```
xin-hedera, line 217: will add the test case back. the `empty()` function was removed at first, however added back later since other test cases use it

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/provider/S3StreamFileProvider.java
```diff
@@ -67,32 +70,30 @@ public Mono<StreamFileData> get(ConsensusNode node, StreamFilename streamFilenam
...
         var startAfter = prefix + lastFilename.getFilenameAfter();
 
         var listRequest = ListObjectsV2Request.builder()
-                .bucket(commonDownloaderProperties.getBucketName())
+                .bucket(properties.getBucketName())
                 .prefix(prefix)
                 .delimiter(SEPARATOR)
                 .startAfter(startAfter)
-                .maxKeys(batchSize)
+                .maxKeys(listBatchSize)
```
jascks, line 91: S3 will return a max of 1000 keys. Not sure what happens if this setting is larger than that. Could result in an error, or perhaps it's simply ignored (javadoc is not clear). Either way, it may not be what is expected.

`AbstractStreamFileProvider.getListBatchSize()` utilizes the existing `batchSize` property (when `USE_DEFAULT_BATCH_SIZE`), and then doubles it. That property in `CommonDownloaderProperties` has no validation annotation currently. Perhaps a `@Max(500)` would make sense, which when doubled will top out at 1000.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/Downloader.java
```diff
@@ -377,7 +390,8 @@ private boolean verifySignatures(Collection<StreamFileSignature> signatures) {
             try {
                 var dataFilename = signature.getDataFilename();
                 var node = signature.getNode();
-                var streamFileData = streamFileProvider.get(node, dataFilename).block();
+                var streamFileData = Objects.requireNonNull(
+                        streamFileProvider.get(node, dataFilename).block());
```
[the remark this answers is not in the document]
jascks: I see, `Mono.block()` returns null when the mono completes empty. Not sure when that might occur, but it _could_, so good call generating an NPE on the spot.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/balance/AccountBalancesDownloader.java
```diff
@@ -65,4 +72,30 @@ public AccountBalancesDownloader(
...
+    protected boolean shouldDownload() {
+        if (downloaderProperties.isEnabled()) {
+            return true;
+        }
+
+        if (accountBalanceFileExists.get()) {
+            return false;
+        }
+
+        if (accountBalanceFileRepository.findLatest().isPresent() || verifiedCount.get() > 0) {
```
jascks, line 93: Would it be a slight optimization to test `verifiedCount.get() > 0` ahead of the DB query?

From what I've seen so far, for the non-balance file world of the future, `AccountBalancesDownloader` will be disabled in its properties. Then, based on the 'should download' criteria here, and using the new stream file provider batch size when listing files, a single balance file will be listed, downloaded, processed, and results written to the database.  And then this downloader shall not download anything else.

xin-hedera: Not sure. The code path when active, runs by default every 30s, so I don’t think it’s worth considering unless it’s a huge performance difference,

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/historicalbalance/HistoricalBalancesService.java
```diff
@@ -0,0 +1,181 @@
...
+     * @param event The record file parsed event published by {@link RecordFileParser}
+     */
+    @Async
+    @TransactionalEventListener
+    public void onRecordFileParsed(RecordFileParsedEvent event) {
+        consensusEndSink.next(event.getConsensusEnd());
+    }
+
+    private void generate(long consensusEnd) {
+        boolean shouldGenerate = accountBalanceFileRepository
```
jnels124, line 123: maybe I am missing from elsewhere but should we record this as metric so we can monitor easier?

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/balance/AccountBalancesDownloader.java
edwin-greene, line 93: Is there a reason that `verifiedCount` is a count? It seems that it is may be used as a boolean -> [it may be either 0 or greater than 0] so could it be an AtomicBoolean instead?

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/provider/S3StreamFileProvider.java
xin-hedera: The SDK documentation is vague by saying `By default, the action returns up to 1,000 key names.` It's unclear if this also applies to when `maxKeys` is set to more than 1000. In addition, it doesn't mention if it'll throw exception when the value is > 1000. I'd just leave it as is.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/balance/AccountBalancesDownloader.java
xin-hedera: can change it to `AtomicBoolean` since `getAndIncrement` should be more expensive and is common code in all downloaders.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/historicalbalance/HistoricalBalancesService.java
xin-hedera: will add metrics

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/balance/AccountBalancesDownloader.java
jascks: Tried `maxKeys(5000)` in my historical downloader PoC in request made via `listObjectsV2Paginator()`, and each page still has 1000 `S3Object`s. Looks like it just quietly trims to 1000.

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/migration/InitializeEntityBalanceMigrationTest.java
```diff
@@ -310,6 +310,14 @@ private void setup() {
...
         persistAccountBalance(450L, contract.toEntityId(), accountBalanceTimestamp2);
 
+        // A synthetic account balance file
+        long syntheticBalanceTimestamp =
+                accountBalanceTimestamp2 + Duration.ofHours(1).toNanos();
+        domainBuilder
+                .accountBalanceFile()
+                .customize(a -> a.consensusTimestamp(syntheticBalanceTimestamp).synthetic(true))
+                .persist();
+
```
edwin-greene, line 320: The tests still pass if `synthetic` is false, so it doesn't seem to be exercising the new part of the CTE (```where synthetic is false
                limit 1```)

xin-hedera: good catch, that's because the synthetic account balance file's timestamp is after max(record_file.consensus_end)

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/Downloader.java
```diff
@@ -234,7 +245,7 @@ private Multimap<StreamFilename, StreamFileSignature> downloadAndParseSigFiles()
 
                 try {
                     var count = streamFileProvider
-                            .list(node, startAfterFilename)
+                            .list(node, startAfterFilename, getBatchSize())
```
steven-sheehy, line 248: It's weird to pass through the batch size except that it's not the batch size by default, it's a sentinel value that indicates the provider should look up the batch size. Prefer a simpler solution that doesn't cascade to the provider classes. It's slightly less efficient since it will retrieve 200 signatures then only use one, but this code will only be called once on the very first start so doesn't need to be optimal. `getMaxSignatures()` returns Long.MAX_VALUE by default.

```suggestion
                            .list(node, startAfterFilename, getBatchSize())
                            .take(getMaxSignatures())
```

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/historicalbalance/HistoricalBalancesService.java
```diff
@@ -0,0 +1,202 @@
...
+@Named
+public class HistoricalBalancesService {
+
+    private static final String GENERATE_ACCOUNT_BALANCES_SQL =
+            """
+            insert into account_balance (account_id, balance, consensus_timestamp)
+            select id, balance, ?
+            from entity
+            where deleted is not true and balance is not null and type in ('ACCOUNT', 'CONTRACT')
+            order by id
```
steven-sheehy, line 63: Queries should live in the repository layer and we should avoid the use of JdbcTemplate, if possible. Move both queries into a `AccountBalanceRepository.balanceSnapshot(long)` and `TokenBalanceRepository.balanceSnapshot(long)`. Add a new `BalanceSnapshotRepository` interface to each.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/historicalbalance/HistoricalBalancesService.java
```diff
@@ -0,0 +1,202 @@
...
+                // Drop the oldest to minimize the delay of generating next historical balances
+                .onBackpressureBuffer(bufferSize, BufferOverflowStrategy.DROP_OLDEST)
+                // Ensure generate runs in the same thread
+                .publishOn(Schedulers.newSingle("historical-balances-service"))
+                .doOnNext(this::generate)
+                .doOnError(t -> log.error("Error processing data triggered by parsed record file", t))
+                .subscribe();
+
+        // metrics
+        var generateDurationTimerBuilder = Timer.builder("hedera.mirror.historicalbalances.duration")
```
steven-sheehy, line 118:
```suggestion
        var generateDurationTimerBuilder = Timer.builder("hedera.mirror.importer.historicalbalance.duration")
```

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/historicalbalance/HistoricalBalancesProperties.java
```diff
@@ -0,0 +1,66 @@
...
+    @DurationUnit(ChronoUnit.MINUTES)
+    @NotNull
+    private Duration minFrequency = Duration.ofMinutes(15);
+
+    private boolean tokenBalances = true;
+
+    @DurationMin(seconds = 30)
+    @DurationUnit(ChronoUnit.SECONDS)
+    @NotNull
+    protected Duration transactionTimeout = Duration.ofSeconds(300);
```
steven-sheehy, line 57:
```suggestion
    private Duration transactionTimeout = Duration.ofMinutes(5);
```

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/historicalbalance/HistoricalBalancesService.java
```diff
@@ -0,0 +1,202 @@
...
+
+    /**
+     * Listens on {@link RecordFileParsedEvent} and emits the consensusEnd to trigger historical balances generation.
+     *
+     * @param event The record file parsed event published by {@link RecordFileParser}
+     */
+    @Async
+    @TransactionalEventListener
+    public void onRecordFileParsed(RecordFileParsedEvent event) {
+        consensusEndSink.next(event.getConsensusEnd());
```
steven-sheehy, line 134: Why do we need to buffer record file events? We should be able to check if we need to start a balance snapshot within 2s and we should be able to complete the snapshot within 15 minutes. If we don't, then we have other problems.

For catch up mode, a buffer doesn't help because we'd quickly fill it up. We're already in a separate thread than RecordFileParser thanks to `@Async` so we don't need the reactor threading model. Think we should remove the buffer.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/historicalbalance/HistoricalBalancesService.java
```diff
@@ -0,0 +1,202 @@
...
+                        .synthetic(true)
+                        .build();
+                accountBalanceFileRepository.save(accountBalanceFile);
+
+                log.info(
+                        "Generated historical account balance file {} with {} account balances and {} token balances for timestamp {} in {}",
+                        filename,
+                        accountBalancesCount,
+                        tokenBalancesCount,
+                        timestamp,
```
steven-sheehy, line 190: Having timestamp and filename is a bit redundant. Can drop timestamp.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/historicalbalance/HistoricalBalancesService.java
```diff
@@ -0,0 +1,202 @@
...
+                .findLatest()
+                .map(AccountBalanceFile::getConsensusTimestamp)
+                .or(() -> recordFileRepository
+                        .findFirst()
+                        .map(RecordFile::getConsensusEnd)
+                        .map(timestamp ->
+                                timestamp + properties.getInitialDelay().toNanos()))
+                .filter(lastTimestamp -> consensusEnd - lastTimestamp
+                        >= properties.getMinFrequency().toNanos())
+                .isPresent();
```
steven-sheehy, line 148: Move to a separate function to reduce cognitive complexity.

steven-sheehy, line 148: Think it would be simpler and more performant to lazily `select max(consensus_timestamp) from account_balance` on startup, then store an `AtomicLong lastTimestamp` that we compare the latest record file consensus end against.

steven-sheehy (changes requested):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/balance/AccountBalancesDownloader.java
```diff
@@ -65,4 +75,36 @@ public AccountBalancesDownloader(
...
+        }
+
+        if (accountBalanceFileRepository.findLatest().isPresent() || verified.get()) {
+            log.info("Disable downloader since there is at least one account balance file parsed or verified");
+            accountBalanceFileExists.set(true);
+            return false;
+        }
+
+        return true;
+    }
```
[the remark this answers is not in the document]
xin-hedera: There's complexity in how we create the first start after filename. For instance, the filename can be created soly from the configured startDate when the set date is after the latest timestamp of the stream file in db.

The correct action here should be disable account balance file downloader, solely base on the fact that `account_balance_file` table is not empty.

While it's still possible to create the start after streamfile object with a flag to indicate there's actually file in db, the complexity is still there and it's harder to maintain.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/historicalbalance/HistoricalBalancesService.java
xin-hedera: I think the main reason was to support generate balance snapshots in exactly every 15 minutes in catch up mode but due to other constraint we can't do it.

Without that, the other benefit with reactor and single-threaded scheduler is we don't have to explicitly handle race conditions in the async event listener. btw, the main purpose of `@Async` is to avoid exceptions in the listener getting propagated to the event publisher.

So if we use flux, we need to explicitly set overflow strategy regardless of the buffer size since the default is buffer all.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/historicalbalance/HistoricalBalancesProperties.java
xin-hedera: will keep `Duration.ofSeconds(300)` since that's consistent with the unit used in DurationMin and DurationUnit

steven-sheehy: DurationUnit only tells you what unit it uses by default if you omit the character unit. We could just as easily switch it to minutes or omit it entirely. In general, we should use the highest unit that fits the time without loss of precision. This makes it more self documenting in the code and in the configuration docs for users.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/historicalbalance/HistoricalBalancesService.java
steven-sheehy: I prefer the approach that is simplest and easiest to understand as that will be easiest to maintain in the long run and avoid bugs. Spring cache is always tricky to get 100% correct. So think we should use the stateful approach.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/historicalbalance/HistoricalBalancesProperties.java
```diff
@@ -0,0 +1,66 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.convert.DurationUnit;
+import org.springframework.validation.annotation.Validated;
+
+@Data
+@ConfigurationProperties("hedera.mirror.importer.parser.record.historical-balances")
+@RequiredArgsConstructor(onConstructor = @__(@Autowired))
+@Validated
+public class HistoricalBalancesProperties {
```
steven-sheehy, line 36:
```suggestion
@ConfigurationProperties("hedera.mirror.importer.parser.record.historicalBalance")
@RequiredArgsConstructor(onConstructor = @__(@Autowired))
@Validated
public class HistoricalBalanceProperties {
```

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/repository/AccountBalanceRepository.java
```diff
@@ -22,9 +22,25 @@
...
 
 public interface AccountBalanceRepository
-        extends CrudRepository<AccountBalance, AccountBalance.Id>, RetentionRepository {
+        extends BalanceSnapshotRepository, CrudRepository<AccountBalance, AccountBalance.Id>, RetentionRepository {
+
+    @Modifying
+    @Override
+    @Query(
+            value =
+                    """
```
steven-sheehy, line 34: nit: Spotless formatted this and the other one weird. Maybe try moving the nativeQuery param first to see if it formats better.
```suggestion
    @Query(nativeQuery = true, value = """
```

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/historicalbalance/HistoricalBalanceService.java
```diff
@@ -0,0 +1,179 @@
...
+            TokenBalanceRepository tokenBalanceRepository) {
+        this.accountBalanceFileRepository = accountBalanceFileRepository;
+        this.accountBalanceRepository = accountBalanceRepository;
+        this.properties = properties;
+        this.recordFileRepository = recordFileRepository;
+        this.tokenBalanceRepository = tokenBalanceRepository;
+
+        // Set repeatable read isolation level and transaction timeout
+        this.transactionTemplate = new TransactionTemplate(platformTransactionManager);
+        this.transactionTemplate.setIsolationLevel(TransactionDefinition.ISOLATION_REPEATABLE_READ);
```
steven-sheehy, line 83: `@TransactionEventListener` states that `Interactions with the underlying transactional resource will not be committed in this phase`. So this transaction might be able to see some of the committed resources. Just to be safe:

```suggestion
        this.transactionTemplate.setIsolationLevel(TransactionDefinition.ISOLATION_REPEATABLE_READ);
        this.transactionTemplate.setPropagationBehavior(PROPAGATION_REQUIRES_NEW);
```

steven-sheehy (changes requested):

jascks (dismissed):
While I'm not as familiar as others with the broad cross section of functionality this PR encompasses, looking at downloader/parser code, and DB schema and extrapolating a bit in other areas, I am comfortable with this PR.

Using the demo bucket, I can see:
```
2023-09-05T22:24:48.872+0000 INFO scheduling-5 c.h.m.i.p.b.AccountBalanceFileParser Successfully processed 62464 items from 2019-10-11T13_45_00.028792Z_Balances.csv in 284.1 ms
```
Followed by several:
```
2023-09-05T22:27:21.696+0000 INFO task-8 c.h.m.i.p.r.h.HistoricalBalanceService Generating historical balances after processing record file with consensusEnd 1570811591052763001
2023-09-05T22:27:21.701+0000 INFO task-8 c.h.m.i.p.r.h.HistoricalBalanceService Generated historical account balance file 2019-10-11T16_33_11.052763001Z_Balances.csv with 555 account balances and 0 token balances in 5.606 ms
```
Resulting in:
```
mirror_node=# select * from account_balance_file;
 consensus_timestamp | count | load_start |  load_end  |                                            file_hash                                             |                    name                     | bytes | time_offset | node_id | synthetic 
---------------------+-------+------------+------------+--------------------------------------------------------------------------------------------------+---------------------------------------------+-------+-------------+---------+-----------
 1570801500028792000 | 62464 | 1693952688 | 1693952688 | 2961b851d18d3917b31210be074012a12c11b098abbb4b86406be4f931be3cfc9ec8defd9cf1597db7483d577ec9ee35 | 2019-10-11T13_45_00.028792Z_Balances.csv    |       |           0 |       0 | f
 1570802423951905001 |     4 | 1693952690 | 1693952690 |                                                                                                  | 2019-10-11T14_00_23.951905001Z_Balances.csv |       |           0 |      -1 | t
 1570803364604288000 |   307 | 1693952702 | 1693952702 |                                                                                                  | 2019-10-11T14_16_04.604288Z_Balances.csv    |       |           0 |      -1 | t
 1570804365577024000 |   522 | 1693952716 | 1693952716 | 
...
```
There are in the end 11 synthetic balance file rows after the first, 'real' balance file processed.

So, at this time, it looks good to me.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/historicalbalance/HistoricalBalancesProperties.java
steven-sheehy: Missed the property path change

xin-hedera: Oh will change. Thought it was just the singular form in class name.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/balance/AccountBalancesDownloader.java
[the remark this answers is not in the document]
xin-hedera: will keep the current implementation.

other than the startDate after existing data in db corner case, special care is also needed to handle the situation that there is just no account balances files in cloud bucket and then inside `getStartAfterFilename` we also need the `accountBalanceFileRepository.findLatest()` so downloader gets disabled once the first synthetic account balance file is generated. Otherwise, downloader will be always enabled and try to list files in cloud bucket every 30s.

jascks (approved):
Reaffirming my original LGTM.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/balance/AccountBalancesDownloader.java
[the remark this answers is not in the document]
xin-hedera: the current implementation checks if there's account balance file row in db, so it'll disable the downloader when the first synthetic balance file is generated.

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/downloader/balance/AccountBalancesDownloaderTest.java
```diff
@@ -110,4 +125,45 @@ void downloadWithMixedStreamFileExtensions() {
...
+        downloader.download();
+
+        // then
+        verifyStreamFiles(List.of(file1));
+
+        // when
+        downloader.download();
+
+        // then no new files are downloaded
+        verifyStreamFiles(List.of(file1));
```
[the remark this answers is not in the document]
xin-hedera: it doesn't verify it's downloaded again, since we use a mock, unless we reset the mock, the value captured stays. perhaps it's clearer to reset the mock and verify with empty list?

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/balance/AccountBalancesDownloader.java
```diff
@@ -65,4 +73,29 @@ public AccountBalancesDownloader(
...
+    protected boolean shouldDownload() {
+        if (downloaderProperties.isEnabled()) {
+            return true;
+        }
+
+        if (accountBalanceFileExists.get()) {
+            return false;
+        }
+
+        if (accountBalanceFileRepository.findLatest().isPresent() || verified.get()) {
```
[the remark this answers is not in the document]
xin-hedera: thanks, I can remove `verified`

steven-sheehy (approved):
LGTM

PR #6808, opened 2023-09-04
Fix acceptance tests race condition

### LINKED ISSUE #6807: Acceptance tests race condition
### Description

After adding the estimateGas+precompile tests there is a race condition happening while running tests with @ acceptance tag in parallel. 

### Steps to reproduce

Run tests using the @ acceptance tag.

### Additional context

_No response_

### Hedera network

other

### Version

0.88.0-SNAPSHOT

### Operating system

None

yiliev0:
**Description**: After the addition of the estimateGas+precompile tests, we observed a race condition when running the tests with the @acceptance tag in parallel.

Changes in this PR:

- Modified getTokenInfo to fetch totalSupply for use in the eth_call tests.
- Updated getAccountDetail to fetch tokens and their respective balances for the eth_call tests.
- Updated secondReceiver account in estimateGas+precompile tests to "Dave"

Benefits:

- When running acceptance tests in parallel (like the NFT TotalSupply test), even if other suites modify the totalSupply, our test will now dynamically pick up the correct amount and assert it.
- This dynamic approach also ensures that if the test is run individually, it will still pass due to the accurate amount fetched.

**Related issue(s)**:

Fixes #6807

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/props/MirrorAccountBalance.java
+++ b/hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/props/MirrorAccountBalance.java
@@ -16,10 +16,21 @@
 
 package com.hedera.mirror.test.e2e.acceptance.props;
 
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.List;
 import lombok.Data;
 
 @Data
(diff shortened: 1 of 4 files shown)


steven-sheehy (approved):
LGTM. Confirmed it passes against integration.

PR #6827, opened 2023-09-07
Nullable `to` address for reverted transactions on `/contracts/results/<txHash>` route

### LINKED ISSUE #6826: Nullable `to` address for reverted transactions on `/contracts/results/<txHash>` route
### Description

**Description**:

The `/contracts/results/<txHash>` route returns null for `to` field if the ethereum transaction is reverted (no contract function result has been created).

**Solutions**:

Get the actual `to` value from the `EthereumTransaction` table.

**Linked issue**:
- https://github.com/hashgraph/hedera-json-rpc-relay/issues/1711

### Steps to reproduce

Call a non-existing contract, the tx should revert, then execute `/contracts/results/<txHash>` and observe `to` field.

### Additional context

https://testnet.mirrornode.hedera.com/api/v1/transactions/0.0.1064-1694498644-146651309
https://testnet.mirrornode.hedera.com/api/v1/contracts/results/0.0.1064-1694498644-146651309

### Hedera network

testnet

### Version

latest

### Operating system

Linux

natanasow:
**Description**:

The `/contracts/results/<txHash>` route returns null for `to` field if the ethereum transaction is reverted (no contract function result has been created).

**Solutions**:

Get the actual `to` value from the `EthereumTransaction` table.

**Steps to reproduce**:

Call a non-existing contract, the tx should revert, then execute `/contracts/results/<txHash>` and observe `to` field.

**Related issue(s)**:

Related to https://github.com/hashgraph/hedera-json-rpc-relay/issues/1711
Fixes #6826 

**Notes for reviewer**:

**Checklist**


xin-hedera (changes requested):
I don't see a new json spec test case for the scenario in the ticket description, did I miss anything?

IvanKavaldzhiev (dismissed):
LGTM

hedera-mirror-rest/__tests__/specs/contracts/results/{id}/eth-tx-by-transactionhash.json
```diff
@@ -0,0 +1,183 @@
...
+        "function_result": [8, 8],
+        "gas_limit": 1000000,
+        "gas_used": 123,
+        "payer_account_id": 8001,
+        "sender_id": 8001,
+        "transaction_result": 22,
+        "transaction_hash": "4a563af33c4871b51a8b108aa2fe1dd5280a30dfb7236170ae5e5e7957eb6392",
+        "transaction_index": 1,
+        "transaction_nonce": 0
+      }
```
xin-hedera, line 29: the ticket says there is no contract call function result since the ethereum transaction is reverted, however we have a matching contract result here. same issue for contract logs and state changes.

xin-hedera (changes requested):

hedera-mirror-rest/__tests__/specs/contracts/results/order.json
```diff
@@ -82,7 +82,8 @@
         "signature_s": "0x1092806a99727a20c31836959133301b65a2bfa980f9795522d21a254e629110",
         "recovery_id": 1,
         "nonce": 3,
-        "value": ["0x14"]
+        "value": ["0x14"],
+        "to_address": "0x0000000000000000000000000000000000001389"
```
steven-sheehy, line 86: For all the files with `to_address` added should sort it by key like the rest of the fields.

hedera-mirror-rest/__tests__/specs/contracts/results/order.json
```diff
@@ -95,7 +96,8 @@
         "signature_s": "0x1092806a99727a20c31836959133301b65a2bfa980f9795522d21a254e629110",
         "recovery_id": 1,
         "nonce": 4,
-        "value": ["0x1e"]
+        "value": ["0x1e"],
+        "to_address": "0x0000000000000000000000000000000000001389"
```
steven-sheehy, line 100: Instead of modifying all these files, you could've just set this as the default in `integrationDomainOps.ethereumTransactionDefaults`.

hedera-mirror-rest/__tests__/specs/contracts/results/{id}/eth-tx-by-transactionhash.json
steven-sheehy: Agree with Xin, except that state changes comes from the sidecar so not sure that will or won't be present. Can you confirm @natanasow ?

steven-sheehy (changes requested):

hedera-mirror-rest/__tests__/specs/contracts/results/{id}/eth-tx-by-transactionhash.json
natanasow: The exact flow for reproduction:
- Execute `erc20.transfer` to a non-existing address, ethers code snippet
```typescript
    const NON_EXISTING_CONTRACT_ADDRESS = '0x01d42839A486DaAa853499De763E4908276A39e5';
    const EXISTING_RECEIVER_ADDRESS = '0x2ed4df6ec66f55a5765def0a24bfa3bac29e795e';

    const erc20Mock = await ethers.getContractAt('ERC20_ABI', NON_EXISTING_CONTRACT_ADDRESS);
    await erc20Mock.transfer(EXISTING_RECEIVER_ADDRESS, 10, {gasLimit: 1_000_000});
```
- Testnet txId - `0.0.1064@1694498644.146651309`
- [Link](https://testnet.mirrornode.hedera.com/api/v1/transactions/0.0.1064-1694498644-146651309) to mirror node `/transactions/<id>`
- [Link](https://testnet.mirrornode.hedera.com/api/v1/contracts/results/0.0.1064-1694498644-146651309) to mirror node `/contracts/results/<tx-id-or-hash>` - returns `null` for `to` address. It has to be the non-existing address to which the tx had been made (in this case 0x01d42839A486DaAa853499De763E4908276A39e5).

steven-sheehy: So the above testnet transaction has a `contract_result`, an `ethereum_transaction` and no logs. So the description in the ticket was incorrect with regards to no result but the test case is correct. Thanks for confirming.

steven-sheehy (approved):
LGTM

hedera-mirror-rest/__tests__/specs/contracts/results/{id}/eth-tx-by-transactionhash.json
xin-hedera: believe that contract result isn't from transaction record, it's generated by importer in case an eth tx fails and there's no contract result in record, it's filled with values importer knows for some fields and best effort for other fields...

xin-hedera (approved):
LGTM

PR #6832, opened 2023-09-08
Switch tests to use OutputCaptureExtension

steven-sheehy:
**Description**:

* Change tests to use Spring's logging framework agnostic `OutputCaptureExtension`
* Fix monitor and web3 including sl4fj implementations on the classpath

**Related issue(s)**:

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/config/LoggingFilterTest.java
+++ b/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/config/LoggingFilterTest.java
@@ -18,60 +18,38 @@ package com.hedera.mirror.web3.config;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
-import com.hedera.mirror.web3.exception.InvalidParametersException;
-import java.io.StringWriter;
 import java.time.Duration;
-import lombok.CustomLog;
 import org.apache.logging.log4j.Level;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.core.Logger;
-import org.apache.logging.log4j.core.appender.WriterAppender;
-import org.apache.logging.log4j.core.layout.PatternLayout;
-import org.junit.jupiter.api.AfterEach;
-import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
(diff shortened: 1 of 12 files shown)


hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/config/LoggingFilterTest.java
```diff
@@ -103,44 +81,38 @@ void filterXForwardedFor() {
...
     }
 
-    private void assertLog(Level level, String pattern) {
-        assertThat(logOutput)
-                .asString()
-                .hasLineCount(1)
-                .contains(level.toString())
-                .containsPattern(pattern);
+    private void assertLog(CapturedOutput logOutput, String level, String pattern) {
+        assertThat(logOutput).asString().hasLineCount(1).contains(level).containsPattern(pattern);
```
jascks, line 116: This is the fourth definition of this identical method I've seen. I don't see any reason it cannot be a static method, and then defined once in some test utilities class. There is `TestUtils` in `importer`, but that won't help here in `web3`. But, we could define such a class in `common`.

In addition, `monitor`'s `PublishMetricsTest` does not define this method, but instead interacts more directly with the captured output:
```
110   assertThat(output).asString().contains("No publishers");
```
Maybe a utility class could look like:
```
@UtilityClass
public final class TestUtils {

    public static void assertLog(CapturedOutput logOutput, String level, String pattern) {
        assertThat(logOutput).asString().hasLineCount(1).contains(level).containsPattern(pattern);
    }

    public static void assertLog(CapturedOutput logOutput, String pattern) {
        assertThat(logOutput).asString().containsPattern(pattern);
    }
}
```
While there are only five test classes that would benefit today, I image the number will grow over time.

steven-sheehy: I think the only reason it's duplicated is because the entire test and LoggingFilter is duplicated. We should think of ways to solve that problem and not worry about duplicating one line from the duplicated test. But that's tricky since both use web dependencies and we don't want the common module to have web dependencies. Don't think it's worth it just for this line and because not every test will have exactly one log line or care about log levels.

jascks: I see, more entanglement than I was aware of.

jascks (approved):
LGTM

edwin-greene (approved):
Looks good

PR #6836, opened 2023-09-09
Set daemonsets to a higher priority class

### LINKED ISSUE #6835: Can’t scale up a node pool because of a failing scheduling predicate
### Description

Seen a couple times where GKE node pools can't scale because of a failing scheduling predicate. Logs indicate it was promtail and node exporter daemonset. These daemonset pods could not get scheduled because other items on the node had higher priority. This would be fine by itself since these pods are not the highest priority, but if it causes the node to not scale then it needs to be addressed.

### Steps to reproduce

Monitor GKE with real life workloads.

### Additional context

_No response_

### Hedera network

other

### Version

0.87.0

### Operating system

None

steven-sheehy:
**Description**:

* Add `priorityClassName` to zfs-init daemonset
* Change all daemon sets to a higher priority class
* Fix use of deprecated `serviceAccount` field
* Remove global `priorityClass.enabled` since it's required elsewhere

**Related issue(s)**:

Fixes #6835

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/charts/hedera-mirror-common/templates/priorityclass.yaml
+++ b/charts/hedera-mirror-common/templates/priorityclass.yaml
@@ -1,4 +1,3 @@
-{{- if .Values.priorityClass.enabled -}}
 apiVersion: scheduling.k8s.io/v1
 kind: PriorityClass
 metadata:
@@ -30,4 +29,3 @@ metadata:
 value: 250
 globalDefault: false
 description: "Used for low priority workloads that no workloads depend upon"
-{{- end -}}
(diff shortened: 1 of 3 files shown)


charts/hedera-mirror-common/templates/priorityclass.yaml
```diff
@@ -1,4 +1,3 @@
-{{- if .Values.priorityClass.enabled -}}
```
steven-sheehy, line 1: Remove since we depend upon `critical` priority class elsewhere and we can't conditionalize it there.

jnels124 (approved):
LGTM

xin-hedera (approved):
LGTM

PR #6862, opened 2023-09-14
HIP-584: Register missing besu internal precompiles

### LINKED ISSUE #6861: HIP-584: Register missing besu internal precompiles
### Description

Currently if we try to make a smart contract that depends its logic on the EVM's internal precompiles we would get errors. The root cause seems to be we don't register those precompiles when we build the EVM object.

### Steps to reproduce

Create a smart contract that invokes an internal precompile and observe its behaviour.

### Additional context

_No response_

### Hedera network

testnet

### Version

0.88.0

### Operating system

None

IvanKavaldzhiev:
**Description**:

**Related issue(s)**:

Fixes #6861 

**Notes for reviewer**:

**Checklist**


0xivanov (commented):
Does this mean that operations like `ecrecover` do not work right now? I believe we have tests about that.

steven-sheehy (changes requested):
Agree, needs some tests.

steven-sheehy (approved):
LGTM

PR #6865, opened 2023-09-15
Update web3 module README with contracts/call API support

### LINKED ISSUE #6726: Update web3 module README with contracts/call API support
### Problem

Currently the openAPI docs under `api/v1/docs` give a summary of what the `contracts/call` POST method supports across call and estimate scenarios.
However, some users are still unclear of what is capable and the openAPI doc isn't the best place for more detailed info.

### Solution

Update the https://github.com/hashgraph/hedera-mirror-node/blob/main/docs/web3/README.md with current support of which scenarios and system contract functions are implemented and supported 

### Alternatives

_No response_

0xivanov:
**Description**:  

Currently the openAPI docs under api/v1/docs give a summary of what the contracts/call POST method supports across call and estimate scenarios. But some of the operations are missing.

**Solution**:
Update the openApi spec with the newest supported precompile calls.

**Related issue(s)**:

Fixes #6726

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-rest/api/v1/openapi.yml
+++ b/hedera-mirror-rest/api/v1/openapi.yml
@@ -321,27 +321,27 @@ paths:
         |                  | non-static contract state reads and modifications for associate precompile functions                         | 0.83+                                   |
         |                  | non-static contract state reads and modifications for non precompile functions with lazy account creation    | 0.84+                                   |
         |                  | static contract state reads for ERC precompile functions (balance, symbol, tokenURI, name, decimals, etc. )  | 0.84+                                   |
-        |                  | HTS system contract MintPrecompile                                                                           | 0.84+                                   |
-        |                  | HTS system contract GrantKycPrecompile                                                                       | 0.85+                                   |
-        |                  | HTS system contract WipePrecompile                                                                           | 0.85+                                   |
-        |                  | HTS system contract BurnPrecompile                                                                           | 0.85+                                   |
-        |                  | HTS system contract DissociatePrecompile                                                                     | 0.85+                                   |
-        |                  | HTS system contract SetApprovalForAllPrecompile                                                              | 0.86+                                   |
-        |                  | HTS system contract ApprovePrecompile                                                                        | 0.86+                                   |
-        |                  | HTS system contract RevokeKycPrecompile                                                                      | 0.86+                                   |
-        |                  | HTS system contract UnpausePrecompile                                                                        | 0.86+                                   |
-        |                  | HTS system contract PausePrecompile                                                                          | 0.86+                                   |
-        |                  | HTS system contract DeleteTokenPrecompile                                                                    | 0.86+                                   |
-        |                  | HTS system contract PausePrecompile                                                                          | 0.86+                                   |
-        |                  | HTS system contract FreezePrecompile                                                                         | 0.86+                                   |
-        |                  | HTS system contract UnfreezePrecompile                                                                       | 0.86+                                   |
+        |                  | non-static contract state reads and modifications for HTS system contract MintPrecompile                     | 0.84+                                   |
+        |                  | non-static contract state reads and modifications for HTS system contract GrantKycPrecompile                 | 0.85+                                   |
+        |                  | non-static contract state reads and modifications for HTS system contract WipePrecompile                     | 0.85+                                   |
+        |                  | non-static contract state reads and modifications for HTS system contract BurnPrecompile                     | 0.85+                                   |
+        |                  | non-static contract state reads and modifications for HTS system contract DissociatePrecompile               | 0.85+                                   |
+        |                  | non-static contract state reads and modifications for HTS system contract SetApprovalForAllPrecompile        | 0.86+                                   |
+        |                  | non-static contract state reads and modifications for HTS system contract ApprovePrecompile                  | 0.86+                                   |
+        |                  | non-static contract state reads and modifications for HTS system contract RevokeKycPrecompile                | 0.86+                                   |
+        |                  | non-static contract state reads and modifications for HTS system contract UnpausePrecompile                  | 0.86+                                   |
+        |                  | non-static contract state reads and modifications for HTS system contract PausePrecompile                    | 0.86+                                   |
+        |                  | non-static contract state reads and modifications for HTS system contract DeleteTokenPrecompile              | 0.86+                                   |
+        |                  | non-static contract state reads and modifications for HTS system contract PausePrecompile                    | 0.86+                                   |
+        |                  | non-static contract state reads and modifications for HTS system contract FreezePrecompile                   | 0.86+                                   |
+        |                  | non-static contract state reads and modifications for HTS system contract UnfreezePrecompile                 | 0.86+                                   |
         |                  | static contract state reads for IsApprovedForAllPrecompile                                                   | 0.87+                                   |
         |                  | static contract state reads for AllowancePrecompile                                                          | 0.87+                                   |
         |                  | static contract state reads for GetApprovedPrecompile                                                        | 0.87+                                   |
         |                  | HTS system contract UpdateTokenKeysPrecompile                                                                | 0.87+                                   |
         |                  | HTS system contract UpdateTokenExpiryInfoPrecompile                                                          | 0.87+                                   |
-        |                  | HTS system contract ExchangeRatePrecompiledContract                                                          | 0.88+                                   |
-        |                  | HTS system contract PrngSystemPrecompiledContract                                                            | 0.89+                                   |
+        |                  | non-static contract state reads and modifications for system contract ExchangeRatePrecompiledContract        | 0.88+                                   |
+        |                  | non-static contract state reads and modifications for system contract PrngSystemPrecompiledContract          | 0.89+                                   |
 
         The operations types which are not currently supported should return 501 error status.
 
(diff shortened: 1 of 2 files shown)


hedera-mirror-rest/api/v1/openapi.yml
```diff
@@ -320,7 +320,28 @@ paths:
...
+        |                  | static contract state reads for IsApprovedForAllPrecompile                                                   | 0.87+                                   |
+        |                  | static contract state reads for AllowancePrecompile                                                          | 0.87+                                   |
+        |                  | static contract state reads for GetApprovedPrecompile                                                        | 0.87+                                   |
+        |                  | HTS system contract UpdateTokenKeysPrecompile                                                                | 0.87+                                   |
+        |                  | HTS system contract UpdateTokenExpiryInfoPrecompile                                                          | 0.87+                                   |
+        |                  | HTS system contract ExchangeRatePrecompiledContract                                                          | 0.88+                                   |
```
IvanKavaldzhiev, line 343: ExchangeRatePrecompiledContract is not a HTS contract

hedera-mirror-rest/api/v1/openapi.yml
```diff
@@ -320,7 +320,28 @@ paths:
...
+        |                  | HTS system contract PrngSystemPrecompiledContract                                                            | 0.89+                                   |
```
IvanKavaldzhiev, line 344: PrngSystemPrecompiledContract is not a HTS contract

hedera-mirror-rest/api/v1/openapi.yml
```diff
@@ -320,7 +320,28 @@ paths:
...
+        |                  | HTS system contract UnpausePrecompile                                                                        | 0.86+                                   |
+        |                  | HTS system contract PausePrecompile                                                                          | 0.86+                                   |
+        |                  | HTS system contract DeleteTokenPrecompile                                                                    | 0.86+                                   |
+        |                  | HTS system contract PausePrecompile                                                                          | 0.86+                                   |
+        |                  | HTS system contract FreezePrecompile                                                                         | 0.86+                                   |
+        |                  | HTS system contract UnfreezePrecompile                                                                       | 0.86+                                   |
```
IvanKavaldzhiev, line 337: I believe for all precompiles it's valid that we support `non-static contract state reads and modifications`.

0xivanov: Is this true for read only precompiles? The modification part might not be right

hedera-mirror-rest/api/v1/openapi.yml
0xivanov: Removed the HTS part

hedera-mirror-rest/api/v1/openapi.yml
0xivanov: Removed the HTS part

hedera-mirror-rest/api/v1/openapi.yml
0xivanov: Added `non-static contract state ...`

hedera-mirror-rest/api/v1/openapi.yml
```diff
@@ -320,7 +320,28 @@ paths:
...
+        |                  | non-static contract state reads and modifications for HTS system contract UnpausePrecompile                  | 0.86+                                   |
+        |                  | non-static contract state reads and modifications for HTS system contract PausePrecompile                    | 0.86+                                   |
+        |                  | non-static contract state reads and modifications for HTS system contract DeleteTokenPrecompile              | 0.86+                                   |
+        |                  | non-static contract state reads and modifications for HTS system contract PausePrecompile                    | 0.86+                                   |
+        |                  | non-static contract state reads and modifications for HTS system contract FreezePrecompile                   | 0.86+                                   |
+        |                  | non-static contract state reads and modifications for HTS system contract UnfreezePrecompile                 | 0.86+                                   |
```
steven-sheehy, line 337: Instead of listing everything we do support, is there enough supported now that we can just list what is not supported? Like `non-static contract state reads and modifications except FooPrecompile`

hedera-mirror-rest/api/v1/openapi.yml
```diff
@@ -320,7 +320,28 @@ paths:
...
+        |                  | non-static contract state reads for AllowancePrecompile                                                      | 0.87+                                   |
+        |                  | non-static contract state reads for GetApprovedPrecompile                                                    | 0.87+                                   |
+        |                  | non-static contract state reads and modifications for HTS system contract UpdateTokenKeysPrecompile          | 0.87+                                   |
+        |                  | non-static contract state reads and modifications for HTS system contract UpdateTokenExpiryInfoPrecompile    | 0.87+                                   |
+        |                  | non-static contract state reads and modifications for system contract ExchangeRatePrecompiledContract        | 0.88+                                   |
+        |                  | non-static contract state reads and modifications for system contract PrngSystemPrecompiledContract          | 0.89+                                   |
```
steven-sheehy, line 344: I think we should drop the `Mirror Node Version` column and just have a `Supported` column. It makes the table too complicated to show the history of support. We should just keep this table up to date to show the current level of support. We should try to collapse the table into as few as rows as possible to simplify it.

hedera-mirror-rest/api/v1/openapi.yml
```diff
@@ -320,7 +320,28 @@ paths:
...
         |                  | non-static contract state reads and modifications for associate precompile functions                         | 0.83+                                   |
         |                  | non-static contract state reads and modifications for non precompile functions with lazy account creation    | 0.84+                                   |
-        |                  | non-static contract state reads and modifications for precompile functions except associate                  | Not supported                           |
+        |                  | static contract state reads for ERC precompile functions (balance, symbol, tokenURI, name, decimals, etc. )  | 0.84+                                   |
+        |                  | non-static contract state reads and modifications for HTS system contract MintPrecompile                     | 0.84+                                   |
+        |                  | non-static contract state reads and modifications for HTS system contract GrantKycPrecompile                 | 0.85+                                   |
```
steven-sheehy, line 325: The ticket mentions splitting out static vs non-static into a separate `Static` column. I think we should do that since every row mentions it.

hedera-mirror-rest/api/v1/openapi.yml
```diff
@@ -320,7 +320,28 @@ paths:
         | true             | non-static contract state reads and modifications for non precompile functions except lazy account creation  | 0.83+                                   |
```
steven-sheehy, line 320: Maybe switch to `Y` or `N` for boolean columns for brevity like the ticket shows as an example.

hedera-mirror-rest/api/v1/openapi.yml
```diff
@@ -320,7 +320,28 @@ paths:
...
         |                  | non-static contract state reads and modifications for associate precompile functions                         | 0.83+                                   |
         |                  | non-static contract state reads and modifications for non precompile functions with lazy account creation    | 0.84+                                   |
-        |                  | non-static contract state reads and modifications for precompile functions except associate                  | Not supported                           |
+        |                  | static contract state reads for ERC precompile functions (balance, symbol, tokenURI, name, decimals, etc. )  | 0.84+                                   |
+        |                  | non-static contract state reads and modifications for HTS system contract MintPrecompile                     | 0.84+                                   |
```
steven-sheehy, line 324: The ticket mentions moving this table from the OpenAPI doc to `docs/web3/README.md` and linking to it. Were we not going to do that at this time?

steven-sheehy (changes requested):

hedera-mirror-rest/api/v1/openapi.yml
0xivanov: I think that every operation is supported now

IvanKavaldzhiev: Hey @Nana-EC, @steven-sheehy if we want to list the support of functionality only for the current release, I think we should delete the whole table now. We support each case for HIP-584 related to "latest" block in 0.89.0, so no point in keeping the table. The only value would be having support documentation for 87 and 88 release.

Nana-EC: I agree with Steven.
We should have enough to say in summary what is supported and in details what is left.
We should be able to say static/non-static HTS/non-HTS  are supported and that historical blocks for these flows are not.

Also better for the open-api to be simple and the README to highlight this detail

hedera-mirror-rest/api/v1/openapi.yml
0xivanov: Added

hedera-mirror-rest/api/v1/openapi.yml
0xivanov: Fixed

hedera-mirror-rest/api/v1/openapi.yml
0xivanov: Added the table in readme.md and linked it

hedera-mirror-rest/api/v1/openapi.yml
0xivanov: Implemented

hedera-mirror-rest/api/v1/openapi.yml
0xivanov: Added the updated table in readme.md and linked it

steven-sheehy (approved):

Nana-EC (approved):

Issue #6619, opened 2023-08-09
HIP-584: Add K6 tests for modification precompiles


IvanKavaldzhiev:
### Problem

We need to add K6 tests related to modification precompiles, so that we can simulate multiple users, which perform changes to the in-memory state asynchronously. In this way we could validate that concurrent processes don't contaminate the state of another process after we rework the state to use ThreadLocals.

### Solution

Implement the following K6 tests:

- [x] Write `k6` test for `MINT_NFT` (actual gas: `732250`)
- [x] Write `k6` test for `APPROVE_ERC` (actual gas: `731632`)
- [x] Write `k6` test for `ASSOCIATE_TOKENS` (actual gas: `730641`)
- [x] Write `k6` test for `DISSOCIATE_TOKENS` (actual gas: `730641`)
- [x] Write `k6` test for `SET_APPROVAL_FOR_ALL` (actual gas: `729608`)
- [x] Write `k6` test for `APPROVE` (actual gas: `729571`)
- [x] Write `k6` test for `APPROVE_NFT` (actual gas: `729569`)

PR #6866, opened 2023-09-15
HIP-584: k6 tests for modification precompiles

bilyana-gospodinova:
**Description**:
We need to add K6 tests related to modification precompiles, so that we can simulate multiple users, which perform changes to the in-memory state asynchronously. In this way we could validate that concurrent processes don't contaminate the state of another process after we rework the state to use ThreadLocals.

Fixes https://github.com/hashgraph/hedera-mirror-node/issues/6619


hedera-mirror-test/k6/src/web3/test/index.js
```diff
@@ -79,6 +92,24 @@ const tests = {
   contractCallTokenType,
   contractCallTokenURI,
   contractCallTotalSupply,
+  contractCallEstimateFungibleTokenCustomFees,
```
kselveliev, line 95: We should remove tests from line 95-112, we add them later when we do the check - > 
if (__ENV.RUN_ESTIMATE_TESTS) {}

bilyana-gospodinova: The tests were moved.

kselveliev:
LGTM

steven-sheehy (approved):
LGTM

PR #6867, opened 2023-09-15
Improve importer memory usage

### LINKED ISSUE #6729: Importer shows high memory consumption at the beginning of 10K TPS perf test
### Description

Importer shows very high memory consumption at the beginning of 10K TPS perf test.

The high memory consumption often times causes high CPU usage since a lot of cycles spent on GC, then the pod becomes unresponsive and liveness/ readiness probes fail multiple times in a row, and eventually triggers kubelet to restart the pod. For instance, the k8s node logs [here](https://console.cloud.google.com/logs/query;query=resource.type%3D%22k8s_node%22%0Aresource.labels.project_id%3D%22mirrornode-non-prod-314918%22%0Aresource.labels.location%3D%22us-central1%22%0Aresource.labels.cluster_name%3D%22mirrornode-preprod%22%0Aresource.labels.node_name%3D%22gke-mirrornode-preprod-preprod-325b8a6c-jhxx%22%0Aseverity%3E%3DDEFAULT%0AjsonPayload.MESSAGE%3D~%22gt2sr%22;cursorTimestamp=2023-08-23T00:07:27.793674Z;startTime=2023-08-23T00:03:07.864Z;endTime=2023-08-23T00:10:00.000Z?project=mirrornode-non-prod-314918).

CPU usage around the time the pod gets restarted (2023-08-23 00:05:00 UTC)

Screenshot 2023-08-23 at 3 03 11 PM

Memory usage:

Screenshot 2023-08-23 at 3 08 25 PM

When it's worse, the pod can also get OOM-killed and recreated. 

### Steps to reproduce

1. Run importer to ingest 10K TPS record stream files from perfnet bucket, set the following in config to disable balance downloading and limit batch size:

```yaml
hedera:
  mirror:
    importer:
      downloader:
        balance:
          enabled: false
        batchsize: 5
```

2. Watch pod memory usage and restart / recreation due to OOM

### Additional context

_No response_

### Hedera network

other

### Version

v0.86.1

### Operating system

None

steven-sheehy:
**Description**:

* Bump PostgreSQL version in chart to 14.9
* Change importer pod resources to guaranteed QoS
* Change `hedera.mirror.importer.downloader.batchSize` from `100` to `25` to reduce memory usage
* Change `hedera.mirror.importer.parser.(event|record).queueCapacity` to `1` to reduce memory usage
* Convert signature downloading from using a thread pool to project reactor
* Disable unused GraphQL and Rosetta sub-charts in Helm chart
* Fix creating fixed sized thread pools for disabled stream types
* Fix record file accumulating gas used on rollback
* Optimize `ProtoRecordFileReader` to lazily convert `Transaction` and `TransactionRecord` to bytes
* Optimize `RecordItem` to remove lazy `AtomicReference` objects and reduce unnecessary objects
* Optimize last `RecordFile` to clear large fields
* Remove `hedera.mirror.importer.downloader.threads` property

**Related issue(s)**:

Fixes #6729

**Notes for reviewer**:

Comparing this PR against 0.88 showed similar numbers for all metrics except memory and CPU, which showed a noticeable reduction in usage. Below is a graph of both runs where the first run is this PR and the later run with 0.88:

Screenshot 2023-09-18 at 14 35 36

**Checklist**


hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/transaction/RecordItem.java
```diff
@@ -65,47 +66,31 @@ public class RecordItem implements StreamItem {
...
+    private final boolean successful;
     private final Transaction transaction;
-    private final byte[] transactionBytes;
+    private final TransactionBody transactionBody;
     private final int transactionIndex;
+    private final TransactionRecord transactionRecord;
+    private final int transactionType;
 
-    // Lazily calculated fields
-    @Getter(lazy = true)
```
steven-sheehy, line 77: Saw a lot of temporary `AtomicReference` objects in heap dumps due to these lazy fields. Refactored them to eagerly calculate these fields in `build()` to reduce memory usage. Lazy fields are fine usually but we create so many `RecordItem` objects that we need to micro-optimize this particular class and most of the fields didn't need to be lazy since they're always used.

hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/transaction/RecordFile.java
```diff
@@ -127,6 +115,13 @@ public class RecordFile implements StreamFile<RecordItem> {
 
     private int version;
 
+    @Override
+    public void clear() {
+        StreamFile.super.clear();
+        setLogsBloom(null);
+        setSidecars(null);
```
edwin-greene, line 122: Sidecars is initialized to `Collections.emptyList()`, should it be reset to an empty list here as well?

hedera-mirror-common/src/main/java/com/hedera/mirror/common/util/DomainUtils.java
```diff
@@ -239,6 +240,10 @@ public static byte[] toBytes(ByteString byteString) {
             return null;
         }
 
+        if (ByteString.EMPTY.equals(byteString)) {
+            return Internal.EMPTY_BYTE_ARRAY;
+        }
+
```
edwin-greene, line 246: DomainUtils has a constant `EMPTY_BYTE_ARRAY` declared, looks like it could be deleted in favor of using this one.

steven-sheehy: I don't think we want to use a constant from a class named `Internal` too widely. I only use it here since internally `ByteString.EMPTY` uses that array and wanted to ensure any `==` comparisons work the same.

hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/transaction/RecordFile.java
steven-sheehy: Good idea. Updated to that.

edwin-greene (approved):
LGTM

xin-hedera (approved):
LGTM

hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/transaction/RecordItem.java
```diff
@@ -116,114 +101,51 @@ public class RecordItem implements StreamItem {
...
+
+        getEntityTransactions().computeIfAbsent(entityId.getId(), id -> entityTransactionBuilder
+                .entityId(id)
+                .build());
     }
 
-    public void addEntityId(EntityId entityId) {
-        if (!entityTransactionPredicate.test(entityId)) {
-            return;
+    public Map<Long, EntityTransaction> getEntityTransactions() {
```
jascks, line 122: I was getting all concerned about thread safety, along with `HashMap.computeItAbsent()` above, but it does look like this phase of the downloader is indeed single threaded. Each data file is downloaded in a serial, blocking manner, verified and then processed.

Is this because of hash chain validation? We could download multiple data files in parallel and then perhaps process them overall more quickly even if serially? But, that would also increase memory usage, which seems to be the key problem being solved here...

steven-sheehy: Entity transaction stuff is only used in the parser, not downloader. Parser code is a scheduled and synchronized so can only have one thread.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/Downloader.java
```diff
@@ -222,62 +212,37 @@ protected boolean shouldDownload() {
...
-        var stopwatch = Stopwatch.createStarted();
-        long timeoutMillis = downloaderProperties.getCommon().getTimeout().toMillis();
-        signatureDownloadThreadPool.invokeAll(tasks, timeoutMillis, TimeUnit.MILLISECONDS);
-        int total = sigFilesMap.size();
-
+        final var signatures = Objects.requireNonNull(Flux.fromIterable(nodes)
+                .flatMap(node -> streamFileProvider
+                        .list(node, startAfterFilename)
+                        .take(listLimit)
+                        .map(s -> {
```
jascks, line 227: nit: move to private method and use a method reference to make this even neater.

jascks (approved):
Looks good!

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/Downloader.java
steven-sheehy: Wouldn't be that much cleaner IMO since I wouldn't be able to use a method reference due to having to pass 3 parameters.

steven-sheehy: Wouldn't be that much cleaner IMO since I wouldn't be able to use a method reference due to having to pass 3 parameters.

jascks: Doh!

PR #6885, opened 2023-09-18
Enable entity balance tracking and migration in rosetta

### LINKED ISSUE #6884: Enable entity balance tracking and migration in rosetta
### Problem

Entity balance tracking and initialization migration were disabled by default in rosetta image in its `entrypoint.sh`, with account balance file sunsetting, these need to be turned on by default.

### Solution

- Remove function `config_entity_balance_tracking` from `entrypoint.sh`
- Add `HEDERA_MIRROR_IMPORTER_PARSER_RECORD_ENTITY_PERSIST_TRACKBALANCE=true` to supervisord.conf

### Alternatives

_No response_

xin-hedera:
**Description**:

This PR changes the rosetta image to enable entity balance tracking and migration by default.

**Related issue(s)**:

Fixes #6884

**Notes for reviewer**:

Note, v0.88.0 needs to be deployed with `ENTITY_PERSIST_TRACKBALANCE=true` and `INITIALIZEENTITYBALANCEMIGRATION_ENABLED=true`

From v0.89.0 with the changes in the PR, custom configuration will no longer  be needed.

**Checklist**


hedera-mirror-rosetta/container/supervisord.conf
```diff
@@ -53,6 +53,7 @@ environment=HEDERA_MIRROR_IMPORTER_PARSER_RECORD_ENTITY_PERSIST_CONTRACTS=false,
   HEDERA_MIRROR_IMPORTER_PARSER_RECORD_ENTITY_PERSIST_SYSTEMFILES=false,
   HEDERA_MIRROR_IMPORTER_PARSER_RECORD_ENTITY_PERSIST_TOKENS=false,
   HEDERA_MIRROR_IMPORTER_PARSER_RECORD_ENTITY_PERSIST_TOPICMESSAGELOOKUPS=false,
+  HEDERA_MIRROR_IMPORTER_PARSER_RECORD_ENTITY_PERSIST_TRACKBALANCE=true,
```
steven-sheehy, line 56: This is the default. Do we need to explicitly set it here?

xin-hedera: adding it here to hardcode it, so that there's no way to override it to `false` through env var set in docker command line or compose file

steven-sheehy (approved):

jascks (approved):
LGTM

PR #6901, opened 2023-09-19
Option to halt importer on recoverable error

### LINKED ISSUE #6860: Add an option to halt importer on recoverable error
### Problem

Mirror node importer prioritizes availability over finding issues in the record stream. It would be useful to prefer the latter in pre-production environments to catch issues earlier in the lifecycle.

### Solution

* Add a `HEDERA_MIRROR_IMPORTER_PARSER_HALTONERROR=false` system property (can't be regular property since we'll use it in a static utility)
* Add a `Utility.handleRecoverableError(String message, String... args)` that either logs the message with the prefix or throws a `ParserException`
* Change everywhere that uses `Utility.RECOVERABLE_ERROR` to use the new utility
* In all environments except mainnet and testnet, set it to true - separate PR on `develop` branch.

### Alternatives

_No response_

jascks:
**Description**:
Mirror node importer prioritizes availability over finding issues in the record stream. It would be useful to prefer the latter in pre-production environments to catch issues earlier in the lifecycle.

This PR modifies:

* Utilize the `HEDERA_MIRROR_IMPORTER_PARSER_HALTONERROR=false` system property (can't be regular property since we'll use it in a static utility)
* Add a `Utility.handleRecoverableError(String message, String... args)` that either logs the message with the prefix or throws a `ParserException`
* Change everywhere that logs uses `Utility.RECOVERABLE_ERROR` to use the new utility

**Related issue(s)**:

Fixes #6860

**Notes for reviewer**:
- I'm still working on that final bullet item regarding environment specific property setting, but otherwise this PR defines the implementation of this feature. By default, the system will behave as previous releases.
- Issue calls for **not** documenting the new property.

**Checklist**

### DIFF
--- a/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/util/Utility.java
+++ b/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/util/Utility.java
@@ -90,7 +90,7 @@ public class Utility {
             }
         } catch (Exception e) {
             var aliasHex = Hex.encodeHexString(alias);
-            log.error(RECOVERABLE_ERROR + "Unable to decode alias to EVM address: {}", aliasHex, e);
+            handleRecoverableError("Unable to decode alias to EVM address: {}", aliasHex, e);
         }
 
         return evmAddress;
--- a/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/util/Utility.java
+++ b/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/util/Utility.java
@@ -25,6 +25,7 @@ import com.google.protobuf.ByteString;
 import com.google.protobuf.GeneratedMessageV3;
 import com.google.protobuf.TextFormat;
 import com.hedera.mirror.common.util.DomainUtils;
+import com.hedera.mirror.importer.exception.ParserException;
 import com.hederahashgraph.api.proto.java.AccountID;
 import com.hederahashgraph.api.proto.java.ContractLoginfo;
 import com.hederahashgraph.api.proto.java.Key;
@@ -44,6 +45,7 @@ import org.apache.commons.codec.binary.Hex;
 import org.apache.commons.lang3.StringUtils;
 import org.bouncycastle.jcajce.provider.digest.Keccak;
 import org.hyperledger.besu.nativelib.secp256k1.LibSecp256k1;
+import org.slf4j.helpers.MessageFormatter;
 
 @CustomLog
 @UtilityClass
@@ -52,7 +54,8 @@ public class Utility {
     public static final Instant MAX_INSTANT_LONG = Instant.ofEpochSecond(0, Long.MAX_VALUE);
 
     public static final String RECOVERABLE_ERROR = "Recoverable error. ";
-
+    static final String HALT_ON_ERROR_PROPERTY = "HEDERA_MIRROR_IMPORTER_PARSER_HALTONERROR";
+    static final String HALT_ON_ERROR_DEFAULT = "false";
     private static final int ECDSA_SECP256K1_COMPRESSED_KEY_LENGTH = 33;
 
     /**
@@ -188,6 +191,22 @@ public class Utility {
             return text;
         }
         return CaseFormat.UPPER_CAMEL.to(CaseFormat.LOWER_UNDERSCORE, text);
+    }
+
+    public static void handleRecoverableError(String message, Object... args) {
+        var haltOnError = Boolean.parseBoolean(System.getProperty(HALT_ON_ERROR_PROPERTY, HALT_ON_ERROR_DEFAULT));
+
+        if (haltOnError) {
+            var formattingTuple = MessageFormatter.arrayFormat(message, args);
+            var throwable = formattingTuple.getThrowable();
+            var formattedMessage = formattingTuple.getMessage();
+
+            throw throwable == null
+                    ? new ParserException(formattedMessage)
+                    : new ParserException(formattedMessage, throwable);
+        } else {
+            log.error(RECOVERABLE_ERROR + message, args);
+        }
     }
 
     // This method is copied from hedera-services EthTxSigs::recoverAddressFromPubKey and should be kept in sync
--- a/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/util/Utility.java
+++ b/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/util/Utility.java
@@ -210,16 +210,13 @@ public class Utility {
      *                the cause of the thrown ParserException.
      */
     public static void handleRecoverableError(String message, Object... args) {
-        var haltOnError = Boolean.parseBoolean(System.getProperty(HALT_ON_ERROR_PROPERTY, HALT_ON_ERROR_DEFAULT));
+        var haltOnError = Boolean.parseBoolean(System.getProperty(HALT_ON_ERROR_PROPERTY));
 
         if (haltOnError) {
             var formattingTuple = MessageFormatter.arrayFormat(message, args);
             var throwable = formattingTuple.getThrowable();
             var formattedMessage = formattingTuple.getMessage();
-
-            throw throwable == null
-                    ? new ParserException(formattedMessage)
-                    : new ParserException(formattedMessage, throwable);
+            throw new ParserException(formattedMessage, throwable);
         } else {
             log.error(RECOVERABLE_ERROR + message, args);
         }
--- a/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/util/Utility.java
+++ b/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/util/Utility.java
@@ -53,7 +53,7 @@ public class Utility {
 
     public static final Instant MAX_INSTANT_LONG = Instant.ofEpochSecond(0, Long.MAX_VALUE);
 
-    public static final String RECOVERABLE_ERROR = "Recoverable error. ";
+    static final String RECOVERABLE_ERROR = "Recoverable error. ";
     static final String HALT_ON_ERROR_PROPERTY = "HEDERA_MIRROR_IMPORTER_PARSER_HALTONERROR";
     static final String HALT_ON_ERROR_DEFAULT = "false";
     private static final int ECDSA_SECP256K1_COMPRESSED_KEY_LENGTH = 33;
--- a/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/util/Utility.java
+++ b/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/util/Utility.java
@@ -193,6 +193,22 @@ public class Utility {
         return CaseFormat.UPPER_CAMEL.to(CaseFormat.LOWER_UNDERSCORE, text);
     }
 
+    /**
+     * Handle a parser recoverable error. Depending on the value of the system property
+     * HEDERA_MIRROR_IMPORTER_PARSER_HALTONERROR, when false (default), the provided message and arguments are logged at
+     * ERROR level, with the message prepended with the String defined by Utility.RECOVERABLE_ERROR, identifying it as a
+     * recoverable error.
+     * <p/>
+     * When the system property is set to true, then ParserException is thrown, and the provided message and arguments
+     * are used for generating the exception message string. Nothing is logged in this mode.
+     *
+     * @param message The message string to be logged. This may contain ordered placeholders in the format of {} just
+     *                like the rest of the logging in Mirror Node.
+     * @param args    the variable arguments list to match each placeholder. Simply omit this if there are no mess
+     *                placeholders. The final, or only, argument may be a reference to a Throwable, in which case it is
+     *                identified as the cause of the error; either logged in the stacktrace following the message, or as
+     *                the cause of the thrown ParserException.
+     */
     public static void handleRecoverableError(String message, Object... args) {
         var haltOnError = Boolean.parseBoolean(System.getProperty(HALT_ON_ERROR_PROPERTY, HALT_ON_ERROR_DEFAULT));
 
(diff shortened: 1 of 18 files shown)


hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/util/Utility.java
```diff
@@ -190,6 +193,38 @@ public static String toSnakeCase(String text) {
...
+     *
+     * @param message The message string to be logged. This may contain ordered placeholders in the format of {} just
+     *                like the rest of the logging in Mirror Node.
+     * @param args    the variable arguments list to match each placeholder. Simply omit this if there are no mess
+     *                placeholders. The final, or only, argument may be a reference to a Throwable, in which case it is
+     *                identified as the cause of the error; either logged in the stacktrace following the message, or as
+     *                the cause of the thrown ParserException.
+     */
+    public static void handleRecoverableError(String message, Object... args) {
+        var haltOnError = Boolean.parseBoolean(System.getProperty(HALT_ON_ERROR_PROPERTY, HALT_ON_ERROR_DEFAULT));
```
jascks, line 213: I'm not certain, but I would not think recoverable errors occur real often. Fetching the property value each invocation makes this more easily testable, but also allows for the property to be updated on a running system should that be helpful.

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/util/UtilityTest.java
```diff
@@ -131,4 +146,82 @@ void instantToTimestamp(long seconds, int nanos) {
...
+    public static class StringArrayConverter extends SimpleArgumentConverter {
+
+        @Override
+        protected Object convert(Object source, Class<?> targetType) throws ArgumentConversionException {
+            if (source == null) {
+                return null;
+            }
+
+            if (source instanceof String && String[].class.isAssignableFrom(targetType)) {
+                return ((String) source).split("\\s*,\\s*");
```
steven-sheehy, line 220:
```suggestion
            if (source instanceof String str && String[].class.isAssignableFrom(targetType)) {
                return str.split("\\s*,\\s*");
```

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/util/Utility.java
steven-sheehy: I don't believe system properties are refreshed after startup, so your latter point is probably not applicable. But I'm not sure how you'd test it if it was static without some nasty reflection.

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/util/UtilityTest.java
```diff
@@ -131,4 +146,82 @@ void instantToTimestamp(long seconds, int nanos) {
...
     }
+
+    @ParameterizedTest(name = "Log {0} with cause {2}")
+    @CsvSource(
+            textBlock =
+                    """
+                                plain message, plain message, false,
+                                {} arg message, one arg message, true, one
+                                {} {} message, a b message, false, 'a, b'
+                                {} {} {} {}, a b c d, true, 'a, b, c, d'
```
steven-sheehy, line 157: It might be cleaner and easier to understand if the exception was moved into the args and the expected message. So that you just have `String format, String[] args, String expected` parameters. The converter could detect the string `exception` and turn it into a RuntimeException.

```suggestion
                                plain message, '', plain message
                                {} arg message, 'one,exception', one arg message: RuntimeException: exception
                                {} {} message, 'a,b', a b message
                                {} {} {} {}, 'a,b,c,d,exception', a b c d: RuntimeException: exception
```

steven-sheehy (changes requested):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/util/Utility.java
```diff
@@ -52,7 +54,8 @@ public class Utility {
     public static final Instant MAX_INSTANT_LONG = Instant.ofEpochSecond(0, Long.MAX_VALUE);
 
     public static final String RECOVERABLE_ERROR = "Recoverable error. ";
```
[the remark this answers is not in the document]
jascks: That makes sense. However, `com.hedera.mirror.importer.domain.ContractResultServiceImplTest` currently references it and resides outside of the package.

I do like shutting off future use though, so I could define a private constant with the same value in `ContractResultServiceImplTest` itself and use that.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/util/Utility.java
jascks: Yep, okay that makes sense. I could be wrong, but I would not think reporting recoverable errors is a high volume thing?

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/util/UtilityTest.java
jascks: Good catch on the old school Java. It's only been seven releases... 😄

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/util/Utility.java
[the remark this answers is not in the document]
steven-sheehy: Could also switch that test to halt on error.

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/util/UtilityTest.java
```diff
@@ -131,4 +146,82 @@ void instantToTimestamp(long seconds, int nanos) {
...
+        var allOutput = capturedOutput.getAll();
+        assertThat(allOutput).contains(RECOVERABLE_ERROR + formatted);
+        if (addCause) {
+            assertThat(allOutput).contains(cause.getClass().getName() + ": " + causeMessage);
+        }
+
+        /*
+         * With halt on error set, ensure ParserException is thrown with expected information.
+         */
+        System.setProperty(HALT_ON_ERROR_PROPERTY, "true");
```
steven-sheehy, line 197: We should probably set this back to false in an AfterEach so it doesn't impact other tests

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/util/UtilityTest.java
jascks: Done

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/util/UtilityTest.java
jascks: Revamped this test case, and the converter significantly. I like it much better. 👍

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/util/UtilityTest.java
jascks: Done

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/util/Utility.java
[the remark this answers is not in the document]
jascks: That test would then have to be cognizant of the new system property setting which I'd like to keep to the `util` package.  `UtilityTest` already does a good job of testing the property and its effects.

`ContractResultServiceImplTest` is all about capturing log output and looking for `RECOVERABLE_ERROR`. I'm thinking defining that constant locally is the way to go for now.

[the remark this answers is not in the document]
jascks: Done

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/util/Utility.java
steven-sheehy: Correct

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/util/UtilityTest.java
```diff
@@ -131,4 +153,81 @@ void instantToTimestamp(long seconds, int nanos) {
...
+            Utility.handleRecoverableError(format);
+        } else {
+            Utility.handleRecoverableError(format, args);
+        }
+
+        var allOutput = capturedOutput.getAll();
+        assertThat(allOutput).contains(RECOVERABLE_ERROR + formatted);
+        if (causeProvided != null) {
+            assertThat(allOutput).contains(causeProvided.getClass().getName() + ": " + causeProvided.getMessage());
+        }
```
steven-sheehy, line 188: Is it not possible to just put the `RuntimeException: provided cause` into the expected message so that the entire output sans prefix is in the `@CsvSource`?

```suggestion
        assertThat(capturedOutput.getAll()).contains(RECOVERABLE_ERROR + formatted);
```

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/util/UtilityTest.java
```diff
@@ -131,4 +153,81 @@ void instantToTimestamp(long seconds, int nanos) {
...
+
+        /*
+         * With halt on error not set, verify expected log output is generated.
+         */
+        System.setProperty(HALT_ON_ERROR_PROPERTY, "false");
+        if (args.length == 0) {
+            Utility.handleRecoverableError(format);
+        } else {
+            Utility.handleRecoverableError(format, args);
+        }
```
steven-sheehy, line 182: Even if you don't pass a `[]` the JVM will generate an empty array regardless. Might as well reduce the conditionals and always pass the args.

```suggestion
        Utility.handleRecoverableError(format, args);
```

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/util/UtilityTest.java
```diff
@@ -131,4 +153,81 @@ void instantToTimestamp(long seconds, int nanos) {
...
+        ParserException parserException;
+        if (args.length == 0) {
+            parserException = assertThrows(ParserException.class, () -> Utility.handleRecoverableError(format));
+        } else {
+            parserException = assertThrows(ParserException.class, () -> Utility.handleRecoverableError(format, args));
+        }
+        assertThat(parserException.getMessage()).isEqualTo(formatted);
+        if (causeProvided != null) {
+            assertThat(parserException.getCause()).isSameAs(causeProvided);
+        }
```
steven-sheehy, line 203: Use assertj and always pass args.

```suggestion
        assertThatThrownBy(() -> Utility.handleRecoverableError(format, args))
            .isInstanceOf(ParserException.class)
            .hasMessage(formatted)
            .hasCause(causeProvided);
```

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/util/UtilityTest.java
jascks: Done

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/util/UtilityTest.java
jascks: Indeed. Nice series of refinements. Thank you. Done.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/util/Utility.java
```diff
@@ -190,6 +193,38 @@ public static String toSnakeCase(String text) {
...
+     *
+     * @param message The message string to be logged. This may contain ordered placeholders in the format of {} just
+     *                like the rest of the logging in Mirror Node.
+     * @param args    the variable arguments list to match each placeholder. Simply omit this if there are no message
+     *                placeholders. The final, or only, argument may be a reference to a Throwable, in which case it is
+     *                identified as the cause of the error; either logged in the stacktrace following the message, or as
+     *                the cause of the thrown ParserException.
+     */
+    public static void handleRecoverableError(String message, Object... args) {
+        var haltOnError = Boolean.parseBoolean(System.getProperty(HALT_ON_ERROR_PROPERTY, HALT_ON_ERROR_DEFAULT));
```
xin-hedera, line 213: nit, can just do the following since `Boolean.parseBoolean(value)` returns false when `value` is null

```suggestion
        var haltOnError = Boolean.parseBoolean(System.getProperty(HALT_ON_ERROR_PROPERTY));
```

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/util/Utility.java
```diff
@@ -190,6 +193,38 @@ public static String toSnakeCase(String text) {
...
+
+        if (haltOnError) {
+            var formattingTuple = MessageFormatter.arrayFormat(message, args);
+            var throwable = formattingTuple.getThrowable();
+            var formattedMessage = formattingTuple.getMessage();
+
+            throw throwable == null
+                    ? new ParserException(formattedMessage)
+                    : new ParserException(formattedMessage, throwable);
```
xin-hedera, line 222: nit, can just pass in the throwable regardless null or not

```suggestion
            throw new ParserException(formattingTuple.getMessage(), formattingTuple.getThrowable());
```

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/util/Utility.java
jascks: Ah, okay, thank you. Done.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/util/Utility.java
jascks: Hmm, all the way up to `Throwable` itself I see. Thanks! Done.

steven-sheehy (approved):

xin-hedera (approved):
LGTM

PR #6676, opened 2023-08-18
HIP-584: Adapt stacked state to use ThreadLocals

IvanKavaldzhiev:
**Description**:

Reducing memory footprint by using ThreadLocals for the state frame caches. K6 tests would be performed to analyze the difference before and after these changes.
**Related issue(s)**:

Fixes #6618 

**Notes for reviewer**:

Do not merge before @david-bakin-sl approval.

It would be good to run existing acceptance tests against this PR to validate its correctness.
**Checklist**

- [ ] Documented (Code comments, README, etc.)
- [ ] Tested (unit, integration, etc.)

PR #6906, opened 2023-09-20
HIP-584: Introduce timestamp field that will be used for filtering hi…

### LINKED ISSUE #6895: HIP-584: Introduce timestamp field into ThreadLocalHolder class
### Problem

Introduce a timestamp field which we will be using to filter data in store, db accessors and db queries.
The new timestamp field needs to added in the current ThreadLocalHolder class as a public static final field.

### Solution

The new timestamp field needs to added in the current ThreadLocalHolder class as a public static final field.

### Alternatives

_No response_

kselveliev:
**Description**:
This pull request introduces a new field in our  **ThreadLocalHolder** class.
The new field will be a holding a timestamp which will be used as a filter in our db queries to
be able to obtain historical data.

Fixes #6895

**Checklist**


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/common/ThreadLocalHolder.java
```diff
@@ -30,6 +30,19 @@
...
+     */
+    public static final long UNSET_TIMESTAMP = -1L;
+
+    /**
+     * Long value which stores the block timestamp used for filtering of historical data.
+     * A value of UNSET_TIMESTAMP indicates that the timestamp is unset or disabled for filtering.
+     * Any value other than UNSET_TIMESTAMP that is a valid timestamp should be considered for filtering operations.
+     */
+    @NonNull
+    public static final ThreadLocal<Long> blockTimestamp = ThreadLocal.withInitial(() -> UNSET_TIMESTAMP);
```
steven-sheehy, line 44: Guess I haven't looked at the other PR since it already does this everywhere, but I don't think we should have `N` ThreadLocals each with a single field. We should instead have one `ThreadLocal` with a single object with all these fields within it. Won't block this PR but something to consider for the other PR. Something like:

```java
class ContractCallContext {
    private long blockTimestamp = UNSET_TIMESTAMP;
    private boolean create = false;
    private final Map<Address, Address> aliases = new HashMap<>();
    ...
}

kselveliev: Agreed. But this modification will happen in this [PR](https://github.com/hashgraph/hedera-mirror-node/pull/6676) like you said. So if you approve this code we can merge it first in 06618-adapt-state-to-use-thread-locals. And then we can address the comment.

steven-sheehy (approved):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/common/ThreadLocalHolder.java
0xivanov: We have some instances where we remove some of the thread locals and keep others. If we encapsulate the whole thing in one object we would lose this flexibility. 
Just something to consider.

steven-sheehy: How so? The context object can be mutable and you can add and remove things from the individual fields via its methods.

PR #6913, opened 2023-09-20
Fix custom fee schedule not updated properly

### LINKED ISSUE #6908: Custom token fees are not updated when cleared
### Description

Currently, a token admin can remove all custom fees by submitting a TokenFeeScheduleUpdateTransaction with an empty list of token fees. The mirror node used to get updated as well. However, in the recent update, it appears that if the custom fee is removed from the token, the mirror node still displays the custom fees, which should be removed from the mirror node database.

### Steps to reproduce

1. Create a token with custom fee
2. Remove the custom fee by submitting `TokenFeeScheduleUpdateTransaction` with an empty list of token fees.
3. Query token info from the consensus node, the fee is gone. On the other hand, the fee is still shown from the mirror node API

The example code to reproduce the problem provided below

```
  const customFee = new CustomFixedFee()
    .setAmount(1000000000)
    .setFeeCollectorAccountId(myAccountId) // The account that will receive the royalty fee
    .setAllCollectorsAreExempt(false);

  const customFee2 = new CustomFixedFee()
    .setAmount(2000000000)
    .setFeeCollectorAccountId(myAccountId) // The account that will receive the royalty fee
    .setAllCollectorsAreExempt(false);

  console.log("CreateToken---------------------");
  let tokenCreateTx = await new TokenCreateTransaction()
    .setTokenName("MyToken")
    .setTokenSymbol("MYT")
    .setTokenType(TokenType.FungibleCommon)
    .setDecimals(0)
    .setInitialSupply(10000)
    .setTreasuryAccountId(myAccountId)
    .setSupplyType(TokenSupplyType.Infinite)
    .setCustomFees([customFee, customFee2])
    .setFeeScheduleKey(myPrivateKey)
    .setMaxTransactionFee(new Hbar(40))
    .freezeWith(client);

  let tokenCreateSubmit = await tokenCreateTx.execute(client);
  console.log("- Executed");
  let tokenCreateRx = await tokenCreateSubmit.getReceipt(client);
  let tokenId = tokenCreateRx.tokenId;
  console.log(`- Created token with ID: ${tokenId}`);
  console.log(`- Token Address ${tokenId.toSolidityAddress()}`);
  console.log("-----------------------------------");

  const query = await new TokenInfoQuery().setTokenId(tokenId).execute(client);
  console.log(query);

  //Create the transaction and freeze for manual signing
  const transaction = await new TokenFeeScheduleUpdateTransaction()
    .setTokenId(tokenId)
    .setCustomFees([])
    .freezeWith(client);

  //Submit the signed transaction to a Hedera network
  const txResponse = await transaction.execute(client);

  //Request the receipt of the transaction
  const receipt = await txResponse.getReceipt(client);

  //Get the transaction consensus status
  const transactionStatus = receipt.status.toString();
  console.log("Remove Token Fee", transactionStatus.toString());

  const query2 = await new TokenInfoQuery().setTokenId(tokenId).execute(client);
  console.log(query2);
  ```

### Additional context

_No response_

### Hedera network

mainnet, testnet, previewnet

### Version

latest

### Operating system

None

### PR DESCRIPTION
**Pull request by xin-hedera**
**Description**:

This PR fixes the bug that custom fee schedule isn't updated properly

- Add `shouldCoalesce()` flag to `UpsertColumn` annotation and skip any coalesce when false
- Annotate `fixedFees`, `fractionalFees`, and `royaltyFees` with `shouldCoalesce = false`
- Add integration tests

**Related issue(s)**:

Fixes #6908 

**Notes for reviewer**:

The migration is left out from the PR, since we want to think about if there's way to fix the existing data without re-downloading and parsing record stream files.

Also want to point out, unable to clear the whole custom fee schedule is a special case of the issue. Generally, if any type of fee is not present in the token fee schedule update tx being processed and the previous such tx has that type of fee, importer will end up copying the old type fee instead of clearing it.

**Checklist**

### DIFF
--- a/hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/UpsertColumn.java
+++ b/hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/UpsertColumn.java
@@ -39,4 +39,12 @@ public @interface UpsertColumn {
      * @return the SQL clause
      */
     String coalesce() default "";
+
+    /**
+     * Specify if the column should coalesce with existing and default value. If false, {@link #coalesce()} is ignored.
+     * The default is true.
+     *
+     * @return Whether to coalesce the column
+     */
+    boolean shouldCoalesce() default true;
 }
--- a/hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/token/AbstractCustomFee.java
+++ b/hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/token/AbstractCustomFee.java
@@ -21,6 +21,7 @@ import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import com.google.common.collect.Range;
 import com.hedera.mirror.common.converter.ObjectToStringSerializer;
 import com.hedera.mirror.common.domain.History;
+import com.hedera.mirror.common.domain.UpsertColumn;
 import com.hedera.mirror.common.domain.Upsertable;
 import io.hypersistence.utils.hibernate.type.json.JsonBinaryType;
 import jakarta.persistence.Id;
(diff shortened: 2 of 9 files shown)

### CODE REVIEW

File: hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/UpsertColumn.java [hunk 1]
```diff
@@ -39,4 +39,12 @@
      * @return the SQL clause
      */
     String coalesce() default "";
```
xin-hedera, line 41: originally wanted to use `coalesce = null` to indicate the generator should not do any coalesce on the column, however java doesn't allow `null` as annotation attribute value.

edwin-greene (approved):
Looks good!

steven-sheehy (approved):
LGTM

PR #6916, opened 2023-09-21
Follow up on regression on contract/results/<transactionHash> address to field as 0x

natanasow:
**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-rest/__tests__/specs/contracts/results/{id}/ethereum-transaction-contract-create-with-empty-to-address.json
+++ b/hedera-mirror-rest/__tests__/specs/contracts/results/{id}/ethereum-transaction-contract-create-with-empty-to-address.json
@@ -1,5 +1,5 @@
 {
-  "description": "Contract results api call using transactionId or an ethereum transaction hash for contract creation",
+  "description": "Contract results api call using transactionId or an ethereum transaction hash for contract creation with empty to address in ethereum transaction table",
   "setup": {
     "entities": [
       {
@@ -55,7 +55,7 @@
         "payer_account_id": "8001",
         "value": "0x0077359400",
         "_value_description": "2000000000 tinybars, in importer java BigInteger.toByteArray always adds extra 0x00",
-        "to_address": "47e8474502a5466f0d47d30ca5c98659cdb92fb1"
+        "to_address": []
       }
     ]
   },
@@ -97,7 +97,7 @@
     "state_changes": [],
     "status": "0x1",
     "timestamp": "167654.000123456",
-    "to": "0x47e8474502a5466f0d47d30ca5c98659cdb92fb1",
+    "to": "0x0000000000000000000000000000000000001389",
     "transaction_index": 1,
     "type": 2,
     "v": 1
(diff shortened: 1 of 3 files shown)


hedera-mirror-rest/__tests__/specs/contracts/results/{id}/ethereum-transaction-contract-create-with-empty-to-address.json
```diff
@@ -0,0 +1,105 @@
...
+    ],
+    "ethereumtransactions": [
+      {
+        "call_data": "0x0707",
+        "consensus_timestamp": 167654000123456,
+        "hash": "4a563af33c4871b51a8b108aa2fe1dd5280a30dfb7236170ae5e5e7957eb6392",
+        "payer_account_id": "8001",
+        "value": "0x0077359400",
+        "_value_description": "2000000000 tinybars, in importer java BigInteger.toByteArray always adds extra 0x00",
+        "to_address": []
```
Nana-EC, line 58: q: should this be null?
```suggestion
        "to_address": null
```

hedera-mirror-rest/viewmodel/contractResultDetailsViewModel.js
```diff
@@ -118,7 +118,9 @@ class ContractResultDetailsViewModel extends ContractResultViewModel {
       this.nonce = ethTransaction.nonce;
       this.r = utils.toHexStringNonQuantity(ethTransaction.signatureR);
       this.s = utils.toHexStringNonQuantity(ethTransaction.signatureS);
-      this.to = utils.toHexStringNonQuantity(ethTransaction.toAddress?.length ? ethTransaction.toAddress : contractResult.evmAddress);
+      if (ethTransaction.toAddress?.length) {
```
natanasow, line 121: @xin-hedera @steven-sheehy @Nana-EC 

The `ethTransaciton.toAddress` field is nullable but could it be empty? 
- If yes - we must add a `?.length` check (as @xin-hedera suggested at the start of the discussion), 
- If no - I will reduce the check to `if (ethTransaction.toAddress) {` as @steven-sheehy suggested and remove [that](https://github.com/hashgraph/hedera-mirror-node/pull/6916/files#diff-d9ef9a0fccadfef7c6fab6195af079c3715df2947839d4c5994f5b2c21d1b53c) test because it will be unnecessary.

hedera-mirror-rest/__tests__/specs/contracts/results/{id}/ethereum-transaction-contract-create-with-empty-to-address.json
natanasow: I asked the same question but you were first :racing_car: .

hedera-mirror-rest/viewmodel/contractResultDetailsViewModel.js
xin-hedera: It could be empty.

by db schema, it's a nullable column so in theory it can also be null. Even though the fact is in importer I believe it never sets it to null, it's still good to be a bit more defensive.

steven-sheehy (approved):
LGTM

Nana-EC (approved):

xin-hedera (approved):
LGTM

PR #6917, opened 2023-09-21
Set dynamic value for create token in acceptance tests

### LINKED ISSUE #6905: Set dynamic value for create token tests
### Problem

Exchange rates in local / testnet / previewnet envs are different and token create operations cost 1 usd and 2 usd for custom fees.
Currently the value passed in the create token tests are hardcoded and work only in local environment.

### Solution

Implement dynamic setting of the hbars sent to the create token call, so the fee is sufficient.

### Alternatives

_No response_

0xivanov:
**Description**:  

Exchange rates in local / testnet / previewnet envs are different and token create operations cost 1 usd and 2 usd for custom fees.
Currently the value passed in the create token tests are hardcoded and work only in local environment.
This PR implements dynamic setting of the hbars sent to the create token call, so the fee is sufficient.

**Related issue(s)**:

Fixes #6905

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/client/MirrorNodeClient.java
+++ b/hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/client/MirrorNodeClient.java
@@ -212,7 +212,7 @@ public class MirrorNodeClient {
     }
 
     public ExchangeRateResponse getExchangeRates() {
-        log.debug("Verify contract result '{}' is returned by Mirror Node");
+        log.debug("Get exchange rates by Mirror Node");
         return callRestEndpoint("/network/exchangerate", ExchangeRateResponse.class);
     }
 
--- a/hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/client/MirrorNodeClient.java
+++ b/hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/client/MirrorNodeClient.java
@@ -30,6 +30,7 @@ import com.hedera.mirror.test.e2e.acceptance.props.MirrorNetworkNode;
 import com.hedera.mirror.test.e2e.acceptance.props.MirrorNetworkNodes;
 import com.hedera.mirror.test.e2e.acceptance.props.MirrorNetworkStake;
 import com.hedera.mirror.test.e2e.acceptance.response.ContractCallResponse;
+import com.hedera.mirror.test.e2e.acceptance.response.ExchangeRateResponse;
 import com.hedera.mirror.test.e2e.acceptance.response.MirrorAccountResponse;
 import com.hedera.mirror.test.e2e.acceptance.response.MirrorContractResponse;
 import com.hedera.mirror.test.e2e.acceptance.response.MirrorContractResultResponse;
@@ -208,6 +209,11 @@ public class MirrorNodeClient {
         log.debug("Verify contract result '{}' is returned by Mirror Node", transactionId);
         return callRestEndpoint(
                 "/contracts/results/{transactionId}", MirrorContractResultResponse.class, transactionId);
+    }
+
+    public ExchangeRateResponse getExchangeRates() {
+        log.debug("Verify contract result '{}' is returned by Mirror Node");
+        return callRestEndpoint("/network/exchangerate", ExchangeRateResponse.class);
     }
 
     public ContractCallResponse contractsCall(ContractCallRequest request) {
--- a/hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/AbstractFeature.java
+++ b/hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/AbstractFeature.java
@@ -28,6 +28,7 @@ import com.hedera.mirror.test.e2e.acceptance.client.FileClient;
 import com.hedera.mirror.test.e2e.acceptance.client.MirrorNodeClient;
 import com.hedera.mirror.test.e2e.acceptance.props.CompiledSolidityArtifact;
 import com.hedera.mirror.test.e2e.acceptance.props.MirrorTransaction;
+import com.hedera.mirror.test.e2e.acceptance.response.ExchangeRateResponse;
 import com.hedera.mirror.test.e2e.acceptance.response.MirrorTransactionsResponse;
 import com.hedera.mirror.test.e2e.acceptance.response.NetworkTransactionResponse;
 import java.io.IOException;
@@ -52,6 +53,22 @@ abstract class AbstractFeature {
 
     @Autowired
     protected ObjectMapper mapper;
+
+    protected ExchangeRateResponse exchangeRates;
+
+    protected long calculateCreateTokenFee(double usdFee, boolean useCurrentFee) {
+        if (exchangeRates == null) {
+            throw new RuntimeException("Exchange rates are not initialized.");
+        }
+        final var fee = useCurrentFee ? exchangeRates.getCurrentRate() : exchangeRates.getNextRate();
+        final long hbarPriceInCents = fee.getCentEquivalent() / fee.getHbarEquivalent();
+        final int usdInCents = 100;
+        // create token requires 1 usd in fees
+        // create token with custom fees requires 2 usd in fees
+        // usdInCents / hbarPriceInCents = amount of hbars equal to 1 usd. Increment that number with 1 for safety and
+        // multiply that number with 10 ^ 8 to convert hbar to tinybar
+        return (long) ((usdInCents * usdFee / hbarPriceInCents + 1) * 100000000);
+    }
 
     protected MirrorTransaction verifyMirrorTransactionsResponse(MirrorNodeClient mirrorClient, int status) {
         String transactionId = networkTransactionResponse.getTransactionIdStringNoCheckSum();
@@ -120,7 +137,7 @@ abstract class AbstractFeature {
     protected record DeployedContract(
             FileId fileId, ContractId contractId, CompiledSolidityArtifact compiledSolidityArtifact) {}
 
-    protected record ExchangeRate(int centEquivalent, int hbarEquivalent) {}
+    //    protected record ExchangeRate(int centEquivalent, int hbarEquivalent) {}
 
     protected CompiledSolidityArtifact readCompiledArtifact(InputStream in) throws IOException {
         return mapper.readValue(in, CompiledSolidityArtifact.class);
--- a/hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/AbstractFeature.java
+++ b/hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/AbstractFeature.java
@@ -61,7 +61,7 @@ abstract class AbstractFeature {
             throw new RuntimeException("Exchange rates are not initialized.");
         }
         final var fee = useCurrentFee ? exchangeRates.getCurrentRate() : exchangeRates.getNextRate();
-        final long hbarPriceInCents = fee.getCentEquivalent() / fee.getHbarEquivalent();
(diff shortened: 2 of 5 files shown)


hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/client/MirrorNodeClient.java
```diff
@@ -210,6 +211,11 @@ public MirrorContractResultResponse getContractResultByTransactionId(String tran
                 "/contracts/results/{transactionId}", MirrorContractResultResponse.class, transactionId);
     }
 
+    public ExchangeRateResponse getExchangeRates() {
+        log.debug("Verify contract result '{}' is returned by Mirror Node");
```
IvanKavaldzhiev, line 215: Do we need this log?

0xivanov: Changed the log to be appropriate

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/EstimatePrecompileFeature.java
```diff
@@ -104,6 +107,22 @@ public void createNewERCContract(int supply) throws IOException {
         ercTestContractSolidityAddress = deployedErcTestContract.contractId().toSolidityAddress();
     }
 
+    @Given("I get exchange rates")
+    public void getExchangeRate() throws IOException {
+        final var exchangeRates = mirrorClient.getExchangeRates();
+        currentExchangeRate = new ExchangeRate(
```
IvanKavaldzhiev, line 113: Probably we should determine whether to use `currentExchangeRate` or `nextExchangeRate` based on the current timestamp and check exchange rates `expiration` timestamp?

0xivanov: Yeah, i guess it could happen if we get the currentExchangeRate when there are few seconds/milliseconds left before the next

0xivanov: The question is what is the timeframe between the call that gets the exchange rates and the create token estimate gas call.

0xivanov: The simplest/safest option i can think of is to make these tests to retry with `nextExchangeRate` calculations in case the first call fails. And if both fail the whole test is considered failed.

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/EstimatePrecompileFeature.java
```diff
@@ -104,6 +107,22 @@ public void createNewERCContract(int supply) throws IOException {
...
+                exchangeRates.getCurrentRate().getCentEquivalent(),
+                exchangeRates.getCurrentRate().getHbarEquivalent());
+        final long hbarPriceInCents = currentExchangeRate.centEquivalent() / currentExchangeRate.hbarEquivalent();
+        final int usdInCents = 100;
+        // create token requires 1 usd in fees
+        // create token with custom fees requires 2 usd in fees
+        // usdInCents / hbarPriceInCents = amount of hbars equal to 1 usd. Increment that number with 1 for safety and
+        // multiply that number with 10 ^ 8 to convert hbar to tinybar
+        createTokenValue = (usdInCents / hbarPriceInCents + 1) * 100000000;
+        createTokenValueWithCustomFees = (usdInCents * 2 / hbarPriceInCents + 1) * 100000000;
```
steven-sheehy, line 123: We should encapsulate this in a shared method in `AbstractFeature`. Then we should just call this everywhere needed instead of storing `createTokenValue*` as class variables.

```java
protected long calculateFee(double usdFee) {
    final double hbarPriceInCents = (double) currentExchangeRate.centEquivalent() / currentExchangeRate.hbarEquivalent();
    final double rate = hbarPriceInCents / 100.0;
    return (usdFee * rate + 1) * 100_000_000L;
}

**Review by steven-sheehy — CHANGES_REQUESTED** [review]

File: hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/EstimatePrecompileFeature.java (hunk 3 above)
Comment (0xivanov) [reply hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/EstimatePrecompileFeature.java:123]: Fixed

File: hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/EstimatePrecompileFeature.java [hunk 4]
```diff
@@ -78,6 +78,9 @@ public class EstimatePrecompileFeature extends AbstractEstimateFeature {
     private TokenId nonFungibleTokenId;
     private DeployedContract deployedEstimatePrecompileContract;
     private DeployedContract deployedErcTestContract;
+    private ExchangeRate currentExchangeRate;
```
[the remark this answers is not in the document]
0xivanov: Fixed

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/AbstractFeature.java
```diff
@@ -120,6 +120,8 @@ protected ContractId verifyCreateContractNetworkResponse() {
     protected record DeployedContract(
             FileId fileId, ContractId contractId, CompiledSolidityArtifact compiledSolidityArtifact) {}
 
+    protected record ExchangeRate(int centEquivalent, int hbarEquivalent) {}
```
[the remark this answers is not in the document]
0xivanov: Fixed

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/EstimatePrecompileFeature.java
0xivanov: Implemented the retry mechanism

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/AbstractFeature.java
```diff
@@ -53,6 +54,22 @@ abstract class AbstractFeature {
...
     protected ObjectMapper mapper;
 
+    protected ExchangeRateResponse exchangeRates;
+
+    protected long calculateCreateTokenFee(double usdFee, boolean useCurrentFee) {
+        if (exchangeRates == null) {
+            throw new RuntimeException("Exchange rates are not initialized.");
+        }
+        final var fee = useCurrentFee ? exchangeRates.getCurrentRate() : exchangeRates.getNextRate();
+        final long hbarPriceInCents = fee.getCentEquivalent() / fee.getHbarEquivalent();
```
steven-sheehy, line 64: This should be double arithmetic so we don't lose precision, right? Services uses BigInteger

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/AbstractFeature.java
```diff
@@ -53,6 +54,22 @@ abstract class AbstractFeature {
     @Autowired
     protected ObjectMapper mapper;
 
+    protected ExchangeRateResponse exchangeRates;
+
+    protected long calculateCreateTokenFee(double usdFee, boolean useCurrentFee) {
+        if (exchangeRates == null) {
+            throw new RuntimeException("Exchange rates are not initialized.");
+        }
+        final var fee = useCurrentFee ? exchangeRates.getCurrentRate() : exchangeRates.getNextRate();
```
steven-sheehy, line 63: Can we not just detect which is closest from now?

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/EstimatePrecompileFeature.java
```diff
@@ -754,17 +759,45 @@ public void createFungibleTokenEstimateGas() {
...
+                    calculateCreateTokenFee(1, true));
+            return;
+
+        } catch (AssertionError e) {
+            log.warn("Assertion failed for estimateGas with current exchange rates. Trying with next exchange rates.");
+        } catch (Exception e) {
+            log.warn(
+                    "Exception occurred for estimateGas with current exchange rates. Trying with next exchange rates.");
+        }
         validateGasEstimationForCreateToken(
```
steven-sheehy, line 775: Make this a lambda `Consumer<Boolean>` that takes in current flag then move try/catch retry into a helper method

```suggestion
        var estimateFunction = current -> validateGasEstimationForCreateToken(
                    Strings.encode(encodedFunctionCall),
                    ContractMethods.CREATE_FUNGIBLE_TOKEN.getActualGas(),
                    calculateCreateTokenFee(1, current));
        validateGasEstimation(estimateFunction);
```

steven-sheehy (changes requested):

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/AbstractFeature.java
0xivanov: Does this mean getting rid of the boolean `useCurrentFee` . Your suggestions include the usage if this boolean.

steven-sheehy: Asking if it's possible to get rid of and just use a timestamp. Not sure how that works eith your retry logic

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/AbstractFeature.java
0xivanov: changed to using double

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/EstimatePrecompileFeature.java
0xivanov: Implemented

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/AbstractFeature.java
0xivanov: I added documentation to `executeAndRetryWithNextExchangeRates` explaining why I am doing retry and the problem i see with comparing the currentExchangeRate expiration timestamp with the system timestamp

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/AbstractFeature.java
```diff
@@ -53,6 +54,22 @@ abstract class AbstractFeature {
...
+        final double hbarPriceInCents = fee.getCentEquivalent() / fee.getHbarEquivalent();
```
steven-sheehy, line 64: This is still integer arithmetic. You need to cast one of the numerator or denominator to double as well.

0xivanov: Addressed

steven-sheehy (approved):

PR #6919, opened 2023-09-21
Modify `Token` model to make it immutable

### LINKED ISSUE #6918: Store models collections are not immutable
### Description

We need to modify the store models to not modify the state of the collections they hold. This breaks the immutability and causes issues for estimateGas, where we call doProcessCall multiple times in one transaction.
This is currently not a problem but is seen in https://github.com/hashgraph/hedera-mirror-node/pull/6676 as potential bug

### Steps to reproduce

Create new collections instead of modifying the old ones.

### Additional context

_No response_

### Hedera network

previewnet

### Version

0.89

### Operating system

None

0xivanov:
**Description**:
We need to modify the store models to not modify the state of the collections they hold. This breaks the immutability and causes issues for estimateGas, where we call doProcessCall multiple times in one transaction.
This is currently not a problem but is seen in https://github.com/hashgraph/hedera-mirror-node/pull/6676 as potential bug

**Related issue(s)**:

Fixes #6918

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-web3/src/main/java/com/hedera/services/store/models/Token.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/services/store/models/Token.java
@@ -932,7 +932,51 @@ public class Token {
     }
 
     /**
-     * Creates new instance of {@link Token} with updated decimals in order to keep the object's immutability and avoid
+     * Creates new instance of {@link Token} with updated removed unique tokens in order to keep the object's immutability and avoid
+     * entry points for changing the state.
+     *
+     * @param oldToken
+     * @param removedUniqueTokens
+     * @return new instance of {@link Token} with updated {@link #removedUniqueTokens} property
+     */
+    private Token createNewTokenWithRemoveddUniqueTokens(Token oldToken, List<UniqueToken> removedUniqueTokens) {
+        return new Token(
+                oldToken.entityId,
+                oldToken.id,
+                oldToken.mintedUniqueTokens,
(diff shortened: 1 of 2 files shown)


IvanKavaldzhiev (approved):
LGTM

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/Token.java
```diff
@@ -932,7 +932,51 @@ private Token createNewTokenWithDecimals(Token oldToken, int decimals) {
...
     /**
-     * Creates new instance of {@link Token} with updated decimals in order to keep the object's immutability and avoid
+     * Creates new instance of {@link Token} with updated removed unique tokens in order to keep the object's immutability and avoid
+     * entry points for changing the state.
+     *
+     * @param oldToken
+     * @param removedUniqueTokens
+     * @return new instance of {@link Token} with updated {@link #removedUniqueTokens} property
+     */
+    private Token createNewTokenWithRemoveddUniqueTokens(Token oldToken, List<UniqueToken> removedUniqueTokens) {
```
bilyana-gospodinova, line 942: Misspelled method name - there is an extra "d" at the end of "Removed".

0xivanov: Fixed

steven-sheehy (approved):

Issue #6856, opened 2023-09-12
HIP-584: Adapt in-memory stack state to utilize pop() functionality


IvanKavaldzhiev:
### Problem

Currently, the use of the in-memory stacked state frames requires that we add new RWCachingState frame by pushing it on top of the stack for each nested transaction. The normal execution flow would consist of popping this added RWCachingState at the end of each nested transaction during commit(). The current code base does not pop the added frames and the stack keeps increasing in height.

### Solution

Pop the RWCachingStateFrame at the end of each nested transaction's lifecycle.

PR #6920, opened 2023-09-21
HIP-584: Adapt stack state to utilize pop functionality

bilyana-gospodinova:
**Description**:
Currently, the use of the in-memory stacked state frames requires that we add new RWCachingState frame by pushing it on top of the stack for each nested transaction. The normal execution flow would consist of popping this added RWCachingState at the end of each nested transaction during commit(). The current code base does not pop the added frames and the stack keeps increasing in height.
This PR implements the missing pop of the RWCachingStateFrame from the stack at the end of each nested transaction's lifecycle.
More integration tests are added and in all gas estimation tests we are going to check now the height of the stack frame.

Fixes https://github.com/hashgraph/hedera-mirror-node/issues/6856

**Notes for reviewer**:
This PR needs to be merged after:
- https://github.com/hashgraph/hedera-mirror-node/pull/6919


david-bakin-sl:
This PR is difficult to review because it is against main.  It should be stacked - that is, against the previous PR #6676 by specifying [06618-adapt-state-to-use-thread-locals](https://github.com/hashgraph/hedera-mirror-node/tree/06618-adapt-state-to-use-thread-locals) at the branch to merge into.  Then this PR would only list the diffs for _this_ PR.  After merging #6676 then you merge/rebase this branch from main and reset the branch to merge against to main.

I'm not even sure why this is separate and _later_ than #6676.  A problem with reviewing #6676 is that it has both thread local behavior _and_ new stack behavior and that stack behavior is _wrong_ (because of the missing pops) and that, with the thread local behavior, makes it hard to know that _that_ PR is correct.  I suggest just putting these changes into _that_ PR directly so it can be reviewed as one correct "state machine" on the stacked state w.r.t. eth_call and eth_estimategas calls.

bilyana-gospodinova:
The target branch was changed to 06618-adapt-state-to-use-thread-locals so that this PR can be reviewed easier. This is a separate PR as agreed in the comments of https://github.com/hashgraph/hedera-mirror-node/pull/6676.

IvanKavaldzhiev (approved):
LGTM

0xivanov (approved):
LGTM

0xivanov:
There is one code smell.

PR #6925, opened 2023-09-21
Add migration for missing nft history timestamp ranges

### LINKED ISSUE #6712: Mirror Node missing nft transaction data
### Description

I'm currently trying to pull all transactions associated with a NFT Collection via Mirror Node.
I see some transactions on hashscan that I don't see in the mirror node pull.

Example:

https://hashscan.io/mainnet/transaction/1685285944.321751033 

https://mainnet-public.mirrornode.hedera.com/api/v1/tokens/0.0.2235264/nfts/247/transactions

The above hashscan transaction was for token id: 0.0.2235264 serial 247

### Steps to reproduce

Example:

https://hashscan.io/mainnet/transaction/1685285944.321751033 

https://mainnet-public.mirrornode.hedera.com/api/v1/tokens/0.0.2235264/nfts/247/transactions

The above hashscan transaction was for token id: 0.0.2235264 serial number 247 and does not show up in the mirror node transaction list for the same token id / serial number

### Additional context

_No response_

### Hedera network

mainnet

### Version

v0.1.0

### Operating system

None

edwin-greene:
**Description**:
Add nft history entries to account for missing timestamp ranges.

**Related issue(s)**:

Fixes #6712 

**Notes for reviewer**:

**Checklist**


hedera-mirror-importer/src/main/resources/db/migration/v1/V1.88.0__add_nft_history_timestamp_ranges.sql
```diff
@@ -0,0 +1,41 @@
...
+for current in
+      select * from nft
+      union all
+      select * from nft_history
+      order by token_id, serial_number, timestamp_range
+    loop
+        if (prior is null) then
+            prior = current;
+            continue;
+        elsif (current.token_id = prior.token_id and current.serial_number = prior.serial_number
```
xin-hedera, line 15: can simplify this to
```suggestion
        if prior is not null and current.token_id = prior.token_id and current.serial_number = prior.serial_number
```

steven-sheehy, line 15: FYI for the future, this type of query would be better written as a window function with `lag()` to get access to previous rows.

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.88.0__add_nft_history_timestamp_ranges.sql
```diff
@@ -0,0 +1,41 @@
...
+                    deleted,
+                    metadata,
+                    serial_number,
+                    spender,
+                    timestamp_range,
+                    token_id)
+                values (
+                    current.account_id,
+                    current.created_timestamp,
+                    current.delegating_spender,
```
xin-hedera, line 30: let's set both `delegating_spender` and `spender` to null instead of copy what's in the row immediately after the missing row.

reason is the missing row at its lower timestamp can only correspond to a nft transfer and whenever a transfer happens, its spender info should be erased.

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/migration/AddNftHistoryRangesMigrationTest.java
```diff
@@ -0,0 +1,199 @@
...
+@DisableRepeatableSqlMigration
+@EnabledIfV1
+@RequiredArgsConstructor(onConstructor = @__(@Autowired))
+@Tag("migration")
+@TestPropertySource(properties = "spring.flyway.target=1.87.2")
+class AddNftHistoryRangesMigrationTest extends IntegrationTest {
+
+    private static final RowMapper<NftHistory> NFT_HISTORY_ROW_MAPPER = rowMapper(NftHistory.class);
+
+    private final @Owner JdbcOperations jdbcOperations;
```
xin-hedera, line 49: don't need `@Owner` since the sql migration doesn't delete rows
```suggestion
    private final JdbcOperations jdbcOperations;
```

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.88.0__add_nft_history_timestamp_ranges.sql
```diff
@@ -0,0 +1,41 @@
+do $$
+declare
+  current record;
+  prior   record;
+begin
+for current in
+      select * from nft
```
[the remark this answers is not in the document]
xin-hedera: for union, the order by at the end applies to the output of the union operator

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.88.0__add_nft_history_timestamp_ranges.sql
edwin-greene: Thanks, after you referenced it here I looked into `lag` and researched it. I rewrote the migration and found that the insertions that only involved the nft history table worked, but the insertion based on the nft table didn't occur, so I didn't go much further into the investigation.

xin-hedera (approved):
LGTM

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.88.0__add_nft_history_timestamp_ranges.sql
xin-hedera: don't think for nft_history table we can use `lag()` because the row physically before nft `(T1, S1)` may be nft `(T2, S2)`

steven-sheehy (approved):

PR #6930, opened 2023-09-22
Add MirrorEntityAccess contract expiry management logic

### LINKED ISSUE #6928: web3 module MirrorEntityAccess.isUsable is missing logic for autorenew consideration
### Description

### Steps to reproduce

### Additional context

_No response_

### Hedera network

mainnet

### Version

0.88.0

### Operating system

None

Nana-EC:
**Description**:
MirrorEntityAccess is applying exclusionary logic for expiration of contracts.
This is leading to valid contracts being set as not usable even though we don't yet have expiration implemented

- Update `balance` check to return false on < 0 to improve considerations (may not be possible)
- Add a check for expiry, if valid check for expiry validity
- Break out contract vs account consideration of expiry. Contract check should consider flag. Leave account flag consideration to minimize changes

**Related issue(s)**:

Fixes #6928 

**Notes for reviewer**:
- [ ] Is the balance check necessary? The `AccountDatabaseAccessor.accountFromEntity()` defaults balance to 0 when null 

**Checklist**

### DIFF
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
@@ -57,22 +57,20 @@ public class MirrorEntityAccess implements HederaEvmEntityAccess {
             return false;
         }
 
-        if (balance <= 0) {
+        if (balance < 0) {
             return false;
         }
 
         // expiry validation
-        if (account.getExpiry() != null && account.getExpiry() > 0L) {
-            if (account.isSmartContract()) {
-                // only consider contract accounts expiry if renewal is enabled
-                if (mirrorNodeEvmProperties.shouldAutoRenewContracts()
-                        && account.getExpiry() < Instant.now().getEpochSecond()) {
-                    return false;
-                }
-            } else {
-                // true for accounts if expiry is defined and within range
-                return account.getExpiry() >= Instant.now().getEpochSecond();
+        if (account.isSmartContract()) {
+            // only consider contract accounts expiry if renewal is enabled
+            if (mirrorNodeEvmProperties.shouldAutoRenewContracts()
+                    && account.getExpiry() < Instant.now().getEpochSecond()) {
+                return false;
             }
+        } else {
+            // true for accounts if expiry is defined and within range
+            return account.getExpiry() >= Instant.now().getEpochSecond();
         }
 
         return true;
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
@@ -42,11 +42,6 @@ public class MirrorEntityAccess implements HederaEvmEntityAccess {
     // We should check only accounts usability
     @Override
     public boolean isUsable(final Address address) {
-        // check address validity first before querying store
-        if (Address.ZERO.equals(address)) {
-            return false;
-        }
-
         final var account = store.getAccount(address, OnMissing.DONT_THROW);
 
         final var balance = account.getBalance();
@@ -58,6 +53,10 @@ public class MirrorEntityAccess implements HederaEvmEntityAccess {
         }
 
         if (balance < 0) {
+            return false;
+        }
+
+        if (Address.ZERO.equals(address)) {
             return false;
         }
 
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
@@ -25,7 +25,9 @@ import com.hedera.mirror.web3.evm.store.Store;
 import com.hedera.mirror.web3.evm.store.Store.OnMissing;
 import com.hedera.mirror.web3.repository.ContractRepository;
 import com.hedera.mirror.web3.repository.ContractStateRepository;
+import com.hedera.node.app.service.evm.exceptions.InvalidTransactionException;
 import com.hedera.node.app.service.evm.store.contracts.HederaEvmEntityAccess;
+import com.hedera.services.store.models.Account;
 import java.time.Instant;
 import lombok.RequiredArgsConstructor;
 import org.apache.tuweni.bytes.Bytes;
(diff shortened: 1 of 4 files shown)


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
```diff
@@ -36,27 +37,43 @@ public class MirrorEntityAccess implements HederaEvmEntityAccess {
...
+
     // We should check only accounts usability
     @Override
     public boolean isUsable(final Address address) {
+        // check address validity first before querying store
+        if (Address.ZERO.equals(address)) {
+            return false;
+        }
+
         final var account = store.getAccount(address, OnMissing.DONT_THROW);
```
david-bakin-sl, line 45: I'm not sure you want `OnMissing.DONT_THROW` here.  That'll return an "empty account" for which the defaults might not be (i.e., won't be) appropriate for anything.

Suggest using the alternative `OnMissing.THROW` and then either handling the resulting `InvalidTransactionException` for the missing account or maybe the right thing to do is just let it fly by so something else can handle it.

(See `StoreImpl.getAccount()`)

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
```diff
@@ -36,27 +37,43 @@ public class MirrorEntityAccess implements HederaEvmEntityAccess {
...
+        // expiry validation
+        if (account.isSmartContract()) {
+            // only consider contract accounts expiry if renewal is enabled
+            if (mirrorNodeEvmProperties.shouldAutoRenewContracts()
+                    && account.getExpiry() < Instant.now().getEpochSecond()) {
+                return false;
+            }
+        } else {
+            // true for accounts if expiry is defined and within range
+            return account.getExpiry() >= Instant.now().getEpochSecond();
```
david-bakin-sl, line 72: Shouldn't this also be conditional, in this case on `shouldAutoRenewAccounts()` ?

david-bakin-sl: I addressed this in `[a882f8f](https://github.com/hashgraph/hedera-mirror-node/pull/6930/commits/a882f8fd9f0ee3cc261b7713ae6335b1c4fed3d6)`

david-bakin-sl:
Unfortunately a bunch of tests failed with my last push.  Looking at that ... may have to revert ...

david-bakin-sl:
With de324b3 (just pushed) two web3 unit tests fail:
* `ContractControllerTest.exceedingDataCreateSizeOnEstimate` - which appears to have nothing whatsoever to do with this PR
* `ContractCallServiceTest.estimateGasForDirectCreateContractDeployWithMissingSender` - which **may have to do with this PR**

Investigating.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
```diff
@@ -36,27 +39,40 @@ public class MirrorEntityAccess implements HederaEvmEntityAccess {
...
     // We should check only accounts usability
     @Override
     public boolean isUsable(final Address address) {
-        final var account = store.getAccount(address, OnMissing.DONT_THROW);
 
-        final var balance = account.getBalance();
-        final var isDeleted = account.isDeleted();
+        if (Address.ZERO.equals(address)) {
+            return false;
+        }
```
steven-sheehy, line 50: Should revert zero address change. Believe it was added for https://github.com/hashgraph/hedera-mirror-node/pull/6776 which pre-adds a zero address user to the store.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
```diff
@@ -36,27 +39,40 @@ public class MirrorEntityAccess implements HederaEvmEntityAccess {
...
         }
 
-        if (balance > 0) {
-            return true;
+        if (account.isDeleted()) {
+            return false;
         }
 
-        if (Address.ZERO.equals(address)) {
+        if (account.getBalance() < 0) {
```
steven-sheehy, line 63: Balance check is related to expiry. We should match services `ContextOptionValidator.expiryStatusGiven()` and return true if balance > 0 and bypass expiry check. So revert logic.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
steven-sheehy: Not confident in this change. We should revert and fix the exact scenario in the ticket. You can follow up with LimeChain about throw vs dont throw separately.

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
```diff
@@ -36,27 +39,40 @@ public class MirrorEntityAccess implements HederaEvmEntityAccess {
...
+            return false;
+        }
+
+        // expiry validation
+        final var accountsCanExpire = account.isSmartContract()
+                ? mirrorNodeEvmProperties.shouldAutoRenewContracts()
+                : mirrorNodeEvmProperties.shouldAutoRenewAccounts();
+        if (accountsCanExpire && account.getExpiry() >= Instant.now().getEpochSecond()) {
```
steven-sheehy, line 71: One code smell here

david-bakin-sl: IMO the suggested sonar change makes the code _less understandable_.  This code is a ladder of tests - each one that fails returns false.  If you make it to the end it returns true.  To conflate the last test with the final return obscures that.  I'll suppress the sonar warning.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
Nana-EC: Ahh good to know.
On it's own it doesn't make sense, together it does.
Basically if there's a valid non zero balance there the account can't be expired as there's something for the system to charge. An expired account would first have paid it's rent and not topped up for renewal, then it make sense to check for expiry.

david-bakin-sl: So the test is `==0` + expiry -> false ( because the balance can never be negative) right?

Nana-EC:
Replaced by simplified alternate version in #6940

PR #6940, opened 2023-09-22
Make `MirrorEntityAccess.isUsable` **not** consider expiry

### LINKED ISSUE #6928: web3 module MirrorEntityAccess.isUsable is missing logic for autorenew consideration

david-bakin-sl:
**Description**:

`MirrorEntityAccess` is using expiry as a gate for `isUsable` for accounts, incl contracts

Should be gated on feature flags - but isn't.  Expiry isn't enabled now but this code thought it was.

Solution for now is to simply _not_ check for expiry as that is the network behavior right now.

Also, make sure balance behaves properly: >0 balance means isUsable, <0 balance means _not_ isUsable, and ==0 balance is a special case.

**Related issue(s)**:

Fixes #6928

Will need #6941  to reimplement expiry here, properly

**Checklist**

- [x] Tested (unit)

### DIFF
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
@@ -25,7 +25,6 @@ import com.hedera.mirror.web3.evm.store.Store.OnMissing;
 import com.hedera.mirror.web3.repository.ContractRepository;
 import com.hedera.mirror.web3.repository.ContractStateRepository;
 import com.hedera.node.app.service.evm.store.contracts.HederaEvmEntityAccess;
-import java.time.Instant;
 import lombok.RequiredArgsConstructor;
 import org.apache.tuweni.bytes.Bytes;
 import org.hyperledger.besu.datatypes.Address;
@@ -36,9 +35,15 @@ public class MirrorEntityAccess implements HederaEvmEntityAccess {
     private final ContractRepository contractRepository;
     private final Store store;
 
-    // We should check only accounts usability
+    // An account is usable if it isn't deleted, if it has a balance >0, or if it has balance==0
+    // but is not the 0-address or the empty account.  (This allows the special case where
+    // a synthetic 0-address account is used in eth_estimateGas.)
     @Override
     public boolean isUsable(final Address address) {
+        // Do not consider expiry/renewal at this time.  It is not enabled in the network.
+        // When it is handled it must be gated on (already existing) mirror node feature flags
+        // (properties).
+
         final var account = store.getAccount(address, OnMissing.DONT_THROW);
 
         final var balance = account.getBalance();
@@ -52,11 +57,19 @@ public class MirrorEntityAccess implements HederaEvmEntityAccess {
             return true;
         }
 
+        if (balance < 0) {
+            return false;
+        }
+
         if (Address.ZERO.equals(address)) {
             return false;
         }
 
-        return account.getExpiry() >= Instant.now().getEpochSecond();
+        if (account.isEmptyAccount()) {
+            return false;
+        }
+
+        return true;
     }
 
     @Override
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
@@ -38,6 +38,10 @@ public class MirrorEntityAccess implements HederaEvmEntityAccess {
     // An account is usable if it isn't deleted, if it has a balance >0, or if it has balance==0
     // but is not the 0-address or the empty account.  (This allows the special case where
     // a synthetic 0-address account is used in eth_estimateGas.)
+    @SuppressWarnings("java:S1126") // "replace this if-then-else statement by a single return"
+    // ^ Complaining about the final if- before the `return true`. But it is clear that this
+    // method is a ladder of tests and merging the final `return true` into the last if- would
+    // impair readability, in this case.
     @Override
     public boolean isUsable(final Address address) {
         // Do not consider expiry/renewal at this time.  It is not enabled in the network.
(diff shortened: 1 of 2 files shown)


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
```diff
@@ -52,11 +57,19 @@ public boolean isUsable(final Address address) {
...
+        if (balance < 0) {
+            return false;
+        }
+
         if (Address.ZERO.equals(address)) {
             return false;
         }
 
-        return account.getExpiry() >= Instant.now().getEpochSecond();
+        if (account.isEmptyAccount()) {
```
Nana-EC, line 68: Q: Is this already covered by an existing test or does it need a new one in `MirrorEntityAccessTest`?

Nana-EC (commented):
LG, one test clarification

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
```diff
@@ -52,11 +57,19 @@ public boolean isUsable(final Address address) {
             return true;
         }
 
+        if (balance < 0) {
```
Nana-EC, line 60: Q: not possible right. Do we need this given the accessor?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
steven-sheehy, line 68: I don't see how this scenario is possible against a real db. The empty account is the zero address and there should never be a `0.0.0` in the db.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
steven-sheehy, line 60: I don't see how this scenario is possible against a real db. There's no negative balance accounts or we'd have a serious problem with the mirror node.

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
david-bakin-sl: It's not not possible because of the accessor, it's "not possible" because we think that there's no way for an account to be in arrears, given the network behaving properly.  A unit test covers this.  Sort of belt-and-suspenders I guess.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
david-bakin-sl: Test `MirrorEntityAccessTest.isNotUsableWithWrongAlias` says this is what happens with a "wrong alias". It's a given to mockito in that test.  Whether it actually happens in the codebase I don't know.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
david-bakin-sl: Well I agree with you.  But someone thought it was worth testing at `MirrorEntityAccessTest.isNotUsableWithNegativeBalance'.   I don't have enough information to know what triggered the writing of that test.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
david-bakin-sl: Are you referring to new L68?  This condition is tested in `MirrorEntityAccessTest.isNotUsableWithWrongAlias`.  As I mention in a separate comment below, I don't know if this actually happens in real life.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
david-bakin-sl: removed the test and the code (per offline discussion)

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
david-bakin-sl: removed the test and the code (per offline discussion)

david-bakin-sl: removed the test and the code (per offline discussion)

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccessTest.java
```diff
@@ -85,37 +86,23 @@ void isUsableWithPositiveBalance() {
...
-        final long balance = -1L;
-        when(store.getAccount(ADDRESS, OnMissing.DONT_THROW)).thenReturn(account);
-        when(account.getBalance()).thenReturn(balance);
-        final var result = mirrorEntityAccess.isUsable(ADDRESS);
-        assertThat(result).isFalse();
-    }
-
-    @Test
-    void isNotUsableWithWrongAlias() {
-        final var address = Address.fromHexString("0x3232134567785444e");
```
Nana-EC, line 99: This was wrong due to to bad entry. I had changed it to 
```suggestion
   final var address = Address.fromHexString("0x0");
```

Nana-EC (approved):
LG, one suggestion on the address.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccessTest.java
david-bakin-sl: Accepted (but I did my own push of a new commit to _reinclude_ the test)

steven-sheehy (approved):

PR #6942, opened 2023-09-23
Make `MirrorEntityAccess.isUsable` not consider expiry (0.88)

### LINKED ISSUE #6928: web3 module MirrorEntityAccess.isUsable is missing logic for autorenew consideration

Nana-EC:
**Description**:
Cherry pick #6940 to 0.88

MirrorEntityAccess is using expiry as a gate for isUsable for accounts, incl contracts

Should be gated on feature flags - but isn't. Expiry isn't enabled now but this code thought it was.

Solution for now is to simply not check for expiry as that is the network behavior right now.

**Related issue(s)**:

Fixes #6928 

**Notes for reviewer**:

**Checklist**


steven-sheehy (approved):

david-bakin-sl (approved):
Cherry pick LGTM

PR #6962, opened 2023-09-26
06812: Extend token db query to filter by timestamp, add token balanc…

### LINKED ISSUE #6812: HIP-584: Extend needed DB queries to filter by timestamp

kselveliev:
**Description**:
Changes in this PR address task 1 from the following [issue](https://github.com/hashgraph/hedera-mirror-node/issues/6812).

NOTES: The timestamp filter used in the db queries will be set from **consensus_end** time of a record_file result based on block number.

Changes applied:
Added TokenBalanceRepository in web3 module with a method for getting **token_balance** based on timestamp filter and added unit tests for it.
Added TokenTransferRepository in web3 module with a method for getting **token_transfer** based on timestamp filter and 
added unit tests for it.

Added new DB query in **TokenRepository** that allows for filtering historical token based on timestamp filter.
Added unit tests for the new method in **TokenRepository**

Fixed a problem in **DomainBuilder.java** where some of the entity objects had the same lower and upper values in their 
timestampRange field. This caused problems when we were quering historical data since we based filtering on the 
timestamp_range column.

**Related issue(s)**:

Fixes #6812

**Notes for reviewer**:

**Checklist**


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenTransferRepository.java
```diff
@@ -0,0 +1,22 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hedera.mirror.web3.repository;
+
+import com.hedera.mirror.common.domain.token.TokenTransfer;
+import org.springframework.data.repository.CrudRepository;
+
+public interface TokenTransferRepository extends CrudRepository<TokenTransfer, TokenTransfer.Id> {}
```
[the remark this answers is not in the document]
IvanKavaldzhiev: I believe it's a convention in mirror-node to define new repository instances in the modules where they are needed. That's why we defined several repositories in web3.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/repository/TokenTransferRepositoryTest.java
```diff
@@ -0,0 +1,38 @@
...
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+
+@RequiredArgsConstructor(onConstructor = @__(@Autowired))
+class TokenTransferRepositoryTest extends Web3IntegrationTest {
+
+    private final TokenTransferRepository tokenTransferRepository;
+
+    @Test
+    void save() {
```
IvanKavaldzhiev, line 31: We shouldn't allow anything in the web3 module to persist entries in DB, so no point in testing save operations.

kselveliev: Comment addressed

bilyana-gospodinova:
LGTM

bilyana-gospodinova (approved):

hedera-mirror-common/src/test/java/com/hedera/mirror/common/domain/DomainBuilder.java
```diff
@@ -343,13 +343,14 @@ public DomainWrapper<ContractStateChange, ContractStateChange.ContractStateChang
...
         long amount = id() + 1000;
         var spender = entityId();
+        long timestamp = timestamp();
         var builder = CryptoAllowanceHistory.builder()
                 .amount(amount)
                 .amountGranted(amount)
                 .owner(id())
                 .payerAccountId(spender)
                 .spender(spender.getId())
-                .timestampRange(Range.closedOpen(timestamp(), timestamp()));
```
steven-sheehy, line 352: I don't see how this change or any change within the file has any impact. Either way you're calling `timestamp()` separately for lower and upper bound. `timestamp()` is guaranteed to produced a monotonically increasing timestamp. Just putting the lower bound into a variable doesn't change its behavior.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenTransferRepository.java
```diff
@@ -0,0 +1,36 @@
...
+
+import com.hedera.mirror.common.domain.token.TokenTransfer;
+import java.util.Optional;
+import org.springframework.data.jpa.repository.Query;
+import org.springframework.data.repository.CrudRepository;
+
+public interface TokenTransferRepository extends CrudRepository<TokenTransfer, TokenTransfer.Id> {
+
+    @Query(
+            value = "SELECT * FROM token_transfer "
```
steven-sheehy, line 27: In mirror node, we prefer all SQL to be written as lower case. Please adjust any file in the PR.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenTransferRepository.java
```diff
@@ -0,0 +1,36 @@
...
+                    + "WHERE "
+                    + "token_id = ?1 AND "
+                    + "account_id = ?2 AND "
+                    + "consensus_timestamp < ?3 "
+                    + "ORDER BY consensus_timestamp DESC "
+                    + "LIMIT 1 ",
```
steven-sheehy, line 33: It's hard to know if this is an appropriate query without seeing how it's used. But I'm struggling to find how it's useful to select one token transfer before a block's timestamp. What's the use case here? Wouldn't you want to sum all token transfers between two timestamps to calculate the historical balance?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenRepository.java
```diff
@@ -21,11 +21,31 @@
...
+            key = "#tokenId + '-' + #blockTimestamp",
+            cacheManager = CACHE_MANAGER_TOKEN,
+            unless = "#result == null")
+    @Query(
+            value = " SELECT * FROM ( "
+                    + " (SELECT * FROM token WHERE token_id = ?1 "
+                    + " AND lower(timestamp_range) < ?2 "
+                    + " UNION ALL "
+                    + " SELECT * FROM token_history WHERE token_id = ?1 "
+                    + " AND lower(timestamp_range) < ?2 AND upper(timestamp_range) > ?2 "
```
steven-sheehy, line 44: We should use range [operators](https://www.postgresql.org/docs/14/functions-range.html) to efficiently utilize the GIST index. Searching the lower and upper bound with AND and ORs is inefficient. Please explain analyze your queries to profile them.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenRepository.java
```diff
@@ -21,11 +21,31 @@
...
+
+    @Cacheable(
+            cacheNames = "tokenUnionCache",
+            key = "#tokenId + '-' + #blockTimestamp",
+            cacheManager = CACHE_MANAGER_TOKEN,
+            unless = "#result == null")
+    @Query(
+            value = " SELECT * FROM ( "
+                    + " (SELECT * FROM token WHERE token_id = ?1 "
+                    + " AND lower(timestamp_range) < ?2 "
```
steven-sheehy, line 41: There's only one token in token table for that token_id, so no need to filter on timestamp range in token subquery. The filter on timestamp range should be moved outside the union subquery as I indicated via slack

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenRepository.java
```diff
@@ -21,11 +21,31 @@
...
 
     @Override
     @Cacheable(cacheNames = "token", cacheManager = CACHE_MANAGER_TOKEN, unless = "#result == null")
     Optional<Token> findById(Long tokenId);
+
+    @Cacheable(
+            cacheNames = "tokenUnionCache",
+            key = "#tokenId + '-' + #blockTimestamp",
+            cacheManager = CACHE_MANAGER_TOKEN,
+            unless = "#result == null")
```
steven-sheehy, line 37: Not sure we should cache this until testing proves it's necessary. Also not sure if the 2s expiry for CACHE_MANAGER_TOKEN is appropriate or could be higher for historical queries.

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenTransferRepository.java
kselveliev: Comment addressed. All of the SQL in this pr is lowercase.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenTransferRepository.java
kselveliev: My mistake. This query cannot be used. I've deleted this repository and created a query in TokenBalanceRepository where the sum of all token transfers between two timestamps is added to the historical token balance.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenRepository.java
kselveliev: I've refactored the query where we would use only "where lower(timestamp_range) < ?" outside of the UNION subquery.
I ran this query with explain and analyze in pgAdmin and the timestamp_range index seems to used with this approach.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenRepository.java
kselveliev: I've refactored the query. The timestamp_range filtering is simplified and moved outside of the UNION subquery.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenRepository.java
kselveliev: Comment addressed. I've removed the caching logic until it is proven that is necessary.

hedera-mirror-common/src/test/java/com/hedera/mirror/common/domain/DomainBuilder.java
```diff
@@ -922,6 +928,17 @@ public DomainWrapper<TokenTransfer, TokenTransfer.TokenTransferBuilder> tokenTra
         return new DomainWrapperImpl<>(builder, builder::build);
     }
 
+    public DomainWrapper<TokenTransfer, TokenTransfer.TokenTransferBuilder> tokenTransferCustom(
```
xin-hedera, line 931: prefer a common custom function in the test class or a utility class, not here.

you can do

```java
  domainBuilder.tokenTransfer()
    .customize(t -> t.amount(...).id(...))
    ...

```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenBalanceRepository.java
```diff
@@ -0,0 +1,98 @@
...
+    @Query(
+            value =
+                    """
+                select * from token_balance
+                where
+                token_id = ?1 and
+                account_id = ?2 and
+                consensus_timestamp < ?3
+                order by consensus_timestamp desc
+                limit 1
```
xin-hedera, line 46: nit, this is slightly better indentation
```suggestion
                select * from token_balance
                where
                  token_id = ?1 and
                  account_id = ?2 and
                  consensus_timestamp < ?3
                order by consensus_timestamp desc
                limit 1
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenBalanceRepository.java
```diff
@@ -0,0 +1,98 @@
...
+     *
+     * @param tokenId         the ID of the token.
+     * @param accountId       the ID of the account.
+     * @param blockTimestamp  the block timestamp used to filter the results.
+     * @return an Optional containing the historical balance at the specified timestamp.
+     *         If there are no token transfers, an empty Optional is returned.
+     *         In the case of an empty Optional, consider using {@link #findByIdAndTimestampLessThan(long, long, long)}
+     *         to get the balance at the given timestamp.
+     */
+    @Query(
```
xin-hedera, line 64: Two issues in the query

- correctness. It's possible the subquery on table `token_balance` returns nothing because there's no token balance row for the `(account, token, blockTimestamp)`, then the result from the query is incorrect
- performance. We should go back only one token balance snapshot before `blockTimestamp` (from `account_balance_file` table), assume the timestamp is `T1`, then calculate the balance with `coalesce(token balance at T1, 0) + sum(transfer amount in (T1, blockimestamp])`

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenRepository.java
```diff
@@ -21,11 +21,39 @@
...
+               select * from (
+                   select * from token where token_id = ?1
+                   union all
+                   select * from token_history where token_id = ?1
+               ) as t
+               where lower(timestamp_range) < ?2
+               order by timestamp_range desc
+               limit 1""",
+            nativeQuery = true)
+    Optional<Token> findByTokenIdAndTimestampRange(long tokenId, long blockTimestamp);
```
xin-hedera, line 58: why is the function named `findByTokenIdAndTimestampRange` while `blockTimestamp` is a point in time not a range?

The query is inefficient since it does not trim unnecessary rows in `token_history` table in the `union all`, should rewrite it to

```sql
(
  select *
  from token
  where token_id = ?1 and lower(timestamp_range) <= ?2
)
union all
(
  select *
  from token_history
  where token_id = ?1 and lower(timestamp_range) <= ?2
  order by lower(timestamp_range) desc
  limit 1
)
```

I'm not sure from the design whether the query should return the latest token info effective at `blockTimestamp` or before `blockTimestamp`. The above query with `<= ?2` is for effective at `blockTimestamp`.

xin-hedera (changes requested):

hedera-mirror-common/src/test/java/com/hedera/mirror/common/domain/DomainBuilder.java
steven-sheehy: Ok, thanks for clarifying.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenBalanceRepository.java
xin-hedera: follow up, the query should work for both fungible and non fungible tokens, right?

so it will be more complicated, for non fungible token, the transfers are in `transaction.nft_transfer`, it's a json array, e.g.,

```json
[
  {"token_id": 1682826, "is_approval": false, "serial_number": 1161, "sender_account_id": 1518550, "receiver_account_id": 1385889},
  ...
]
```

hedera-mirror-common/src/test/java/com/hedera/mirror/common/domain/DomainBuilder.java
kselveliev: Comment addressed. I removed the tokenTransferCustom method and now using the common custom function as proposed.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenBalanceRepository.java
kselveliev: Comment addressed.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenRepository.java
```diff
@@ -21,11 +21,53 @@
...
+                        select *
+                        from token_history
+                        where token_id = ?1 and lower(timestamp_range) <= ?2
+                        AND NOT EXISTS (
+                            SELECT 1
+                            FROM token_history as inner_th
+                            WHERE inner_th.token_id = token_history.token_id
+                            AND lower(inner_th.timestamp_range) > lower(token_history.timestamp_range)
+                            AND lower(inner_th.timestamp_range) <= ?2
+                        )
```
xin-hedera, line 65: why use the complicated logic instead of just `order by lower(timestamp_range) desc limit 1`?

there is an index on `token_history (token_id, lower(timestamp_range))` so it should be fast enough.

xin-hedera (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenRepository.java
kselveliev: Comment addressed. I've fixed the naming of the method since we are not using the range.

I've refactored the query as proposed.

I've tested and improved it to trim the unnecessary rows in token_history.

Here is some analysis done on the current query and the improved one.

Testing the following queries:

image

In local DB I’ve setup to have **1 million** records that satisfy the condition lower(timestamp_range) < 1795888431263421169 and token_id=1024

Here is the query result.

image

Execution plan:

image

Usually it takes around 500.000 ms to complete.

I've enhanced the query to trim the history token only to the one closest to the block timestamp.
This way we save time from ordering by millions of records to only one.

Enhanced query:

image

For second part that queries token_history, I’ve created a clause which will trim all of the records that satisfy token_id = 1024 and lower(timestamp_range) <= 1795888431263421169. and will leave only the one record that is closest or equal to 1795888431263421169 which is our filter in this case.

Query result:
image

Execution plan:
image

It usually takes less under 1.000 ms to complete 

So this query performs a few hundred times faster than the first one.
Let me know if you agree with this change

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenRepository.java
kselveliev: Could you check https://github.com/hashgraph/hedera-mirror-node/pull/6962#discussion_r1347868412.
I can revert the complicated logic if you believe it is okay. My concern was about the order by since it may do the order by millions of records.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenBalanceRepository.java
kselveliev: I've fixed the correctness part as proposed.

In terms of performance I tried to trim to check only the closest token balance snapshot before blockTimestamp.
Something similar to what I did [here](https://github.com/hashgraph/hedera-mirror-node/pull/6962#discussion_r1347868412). But it somehow slowed down the query when I did the analysis. 
So I left it as it is. Let me know if this is okay for you.

As for the question if it is for non fungible token. I was not aware of that json array complexity.
So for now I believe in our business logic in the code we should only use this query when dealing with fungible tokens. We will create a separate query for non fungible in another PR.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenRepository.java
kselveliev: I've reverted the complex part

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenBalanceRepository.java
xin-hedera: can you test this query?

```sql
select coalesce((
    select balance
    from token_balance
    where
      consensus_timestamp = s.consensus_timestamp and
      account_id = :accountId and
      token_id = :tokenId
  ), 0) + coalesce((
    select sum(amount)
    from token_transfer
    where
      account_id = :accountId and
      token_id = :tokenId and
      consensus_timestamp > s.consensus_timestamp and
      consensus_timestamp <= :blockTimestamp
  ), 0)
from (
  select consensus_timestamp
  from account_balance_file
  where consensus_timestamp <= :blockTimestamp
  limit 1
) as s;

```

kselveliev: One question about this proposal. Is the modified query intended for increased query performace? 

Im am asking this bacause I was not aware of this account balance file table and how is it used.

And also what is the relation with token balance table . Because to test it I need to persist also the data in the account balance file table for the unit test setup. 

So if I understand correctly we should first find a record in account_balance_file.

Then we take its consensus_timestamp based on the filter, and then we should always have a record in token balance table where consensus_timestamp = token_balance_file.consensus_timestamp and also the  token_id and account_id provided.

Is this the correct logic?

xin-hedera: for both performance and some corner cases I mentioned about the original query.

`account_balance_file` has one row for each account balance file, it includes the account balance file name, its consensus timestamp, and etc.

For each row in account balance file, there should be corresponding account balance rows in `account_balance` table with the same consensus timestamp. same applies to `token_balance` table.

However for a timestamp `T1`, a token balance for `(accountId, tokenId)` may not exist since the account wasn't associated with the token at  `T1`, thus we should do a `coalesce`

kselveliev: Understood, I’ve replaced the the previous query with this one.

I’ve setup the tests so that we have account_balance_file record and 

a token balance record with the same consensus_timestamp and tests 

are passing. Let me know if this is sufficient.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenBalanceRepository.java
```diff
@@ -0,0 +1,93 @@
...
+                          consensus_timestamp <= ?3
+                      ), 0)
+                    from (
+                      select consensus_timestamp
+                      from account_balance_file
+                      where consensus_timestamp <= ?3
+                      order by consensus_timestamp desc
+                      limit 1
+                    ) as s;
+                    """,
```
xin-hedera, line 90: failed with v2 database schema due to citus limitations, so have to rewrite the query to make citus happy

```sql
with balance_snapshot as (
  select consensus_timestamp
  from account_balance_file
  where consensus_timestamp <= ?3
  limit 1
), base as (
    select balance
    from token_balance as tb, balance_snapshot as s
    where
      tb.consensus_timestamp = s.consensus_timestamp and
      token_id = ?1 and
      account_id = ?2
), change as (
    select sum(amount) as amount
    from token_transfer as tt, balance_snapshot as s
    where
      token_id = ?1 and
      account_id = ?2 and
      tt.consensus_timestamp > s.consensus_timestamp and
      tt.consensus_timestamp <= ?3
)
select coalesce((select balance from base), 0) + coalesce((select amount from change), 0)
```

kselveliev: comment addressed

hedera-mirror-common/src/test/java/com/hedera/mirror/common/domain/DomainBuilder.java
```diff
@@ -349,7 +349,7 @@ public DomainWrapper<ContractStateChange, ContractStateChange.ContractStateChang
                 .owner(id())
                 .payerAccountId(spender)
                 .spender(spender.getId())
-                .timestampRange(Range.closedOpen(timestamp(), timestamp()));
+                .timestampRange(Range.closedOpen(timestamp(), timestamp(10)));
```
xin-hedera, line 352: I made the change to add the delta to `id` so `timestamp()` will always increase, otherwise it breaks the assumption made in `FixCryptoAllowanceAmountMigrationTest`

xin-hedera (dismissed):
LGTM

0xivanov (dismissed):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenBalanceRepository.java
```diff
@@ -0,0 +1,93 @@
...
+     * @return an Optional containing the historical balance at the specified timestamp.
+     *         If there are no token transfers between the consensus_timestamp of token_balance and the block timestamp,
+     *         the method will return the balance present at consensus_timestamp.
+     */
+    @Query(
+            value =
+                    """
+                    with balance_snapshot as (
+                      select consensus_timestamp
+                      from account_balance_file
```
steven-sheehy, line 69: Not sure why Xin recommended using account_balance_file when we're trying to purge its use everywhere. But we can go with it as is and open a ticket to follow up.

steven-sheehy (dismissed):
LGTM

kselveliev:
I will close this and reopen it from a fresh branch since i forgot to verify one of the initial commits and due to that merging is blocked.

Issue #6963, opened 2023-09-26
Unneeded store.wrap() call in ApproveAllowanceLogic


bilyana-gospodinova:
### Description

In ApproveAllowanceLogic#applyNftAllowances() there is an extra store.wrap() call and the stack frames grow unnecessary.

### Steps to reproduce

none

### Hedera network

testnet

### Version

v0.88.0

### Operating system

None

PR #6964, opened 2023-09-26
HIP-584: Remove unneeded store.wrap() call in ApproveAllowanceLogic

bilyana-gospodinova:
**Description**:
In ApproveAllowanceLogic#applyNftAllowances() there is an extra store.wrap() call and the stack frames grow unnecessary. This call is now removed.

Fixes https://github.com/hashgraph/hedera-mirror-node/issues/6963

### DIFF
--- a/hedera-mirror-web3/src/main/java/com/hedera/services/txns/crypto/ApproveAllowanceLogic.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/services/txns/crypto/ApproveAllowanceLogic.java
@@ -180,7 +180,6 @@ public class ApproveAllowanceLogic {
         if (nftAllowances.isEmpty()) {
             return;
         }
-        Account modifiedAccount = null;
         for (final var allowance : nftAllowances) {
             final var owner = allowance.getOwner();
             var approvingAccount = fetchOwnerAccount(owner, payerAccount, store, accountsChanged);
@@ -201,7 +200,7 @@ public class ApproveAllowanceLogic {
                 final var accountWithUpdatedApproveForAllNfts =
                         approvingAccount.setApproveForAllNfts(approveForAllNfts);
                 if (!accountWithUpdatedApproveForAllNfts.equals(approvingAccount)) {
-                    modifiedAccount = accountWithUpdatedApproveForAllNfts;
+                    accountsChanged.put(approvingAccount.getId().num(), approvingAccount);
                 }
             }
 
@@ -214,9 +213,6 @@ public class ApproveAllowanceLogic {
                     store, approvingAccount.getId(), spenderId, tokenId, allowance.getSerialNumbersList());
             for (final var nft : nfts) {
                 nftsTouched.put(nft.getNftId(), nft);
-            }
-            if (modifiedAccount != null) {
-                accountsChanged.put(approvingAccount.getId().num(), approvingAccount);
             }
         }
     }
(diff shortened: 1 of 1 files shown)


IvanKavaldzhiev (approved):
LGTM

kselveliev (dismissed):
LGTM

hedera-mirror-web3/src/main/java/com/hedera/services/txns/crypto/ApproveAllowanceLogic.java
```diff
@@ -207,11 +212,12 @@ protected void applyNftAllowances(
 
             final var nfts = updateSpender(
                     store, approvingAccount.getId(), spenderId, tokenId, allowance.getSerialNumbersList());
-            store.wrap();
             for (final var nft : nfts) {
                 nftsTouched.put(nft.getNftId(), nft);
             }
-            accountsChanged.put(approvingAccount.getId().num(), approvingAccount);
+            if (modifiedAccount != null) {
+                accountsChanged.put(approvingAccount.getId().num(), approvingAccount);
```
0xivanov, line 219: I believe you can move this line in line 203 and revert the other changes. This will decrease complexity and might fix the code smell.

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/services/txns/crypto/ApproveAllowanceLogic.java
```diff
@@ -197,7 +197,11 @@ protected void applyNftAllowances(
                     // Need not validate anything here to revoke the approval
                     approveForAllNfts.remove(key);
                 }
-                approvingAccount = approvingAccount.setApproveForAllNfts(approveForAllNfts);
+                final var accountWithUpdatedApproveForAllNfts =
+                        approvingAccount.setApproveForAllNfts(approveForAllNfts);
+                if (!accountWithUpdatedApproveForAllNfts.equals(approvingAccount)) {
```
0xivanov, line 202: Isn't this always true. I mean we either add or remove from the set and then create new account with the updated set.
Isn't this enough? Like without the if statements and stuff:
``` java
approvingAccount = approvingAccount.setApproveForAllNfts(approveForAllNfts);
accountsChanged.put(approvingAccount.getId().num(), approvingAccount);
```

bilyana-gospodinova: I am not sure. What if the set is empty and we are in the else case and try to remove an element that does not exist? In this case the set is the same and there will be no need to update the account.

hedera-mirror-web3/src/main/java/com/hedera/services/txns/crypto/ApproveAllowanceLogic.java
bilyana-gospodinova: Updated as requested.

hedera-mirror-web3/src/main/java/com/hedera/services/txns/crypto/ApproveAllowanceLogic.java
```diff
@@ -180,6 +180,7 @@ protected void applyNftAllowances(
         if (nftAllowances.isEmpty()) {
             return;
         }
+        Account modifiedAccount = null;
```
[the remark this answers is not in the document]
bilyana-gospodinova: I made the changes as in Ivan's suggestion.

hedera-mirror-web3/src/main/java/com/hedera/services/txns/crypto/ApproveAllowanceLogic.java
0xivanov: Yeah I agree on this point.

0xivanov (approved):

david-bakin-sl:
This PR does more than simply remove the `wrap()` call - it's doing some logic.  PR title & comment (and subsequent commit message) should reflect that.

steven-sheehy (approved):
LGTM

PR #6966, opened 2023-09-27
HIP-584: Extend entity, account balance and crypto transfer db queries 

### LINKED ISSUE #6812: HIP-584: Extend needed DB queries to filter by timestamp

kselveliev:
**Description**:

Fixes part two of #6812. The timestamp filter used in the db queries will be set from `consensus_end` time of a record_file result based on block number.

* Add two queries to `EntityRepository`
* Add new unit tests in `EntityRepositoryTest` to test the new methods
* Add `AccountBalanceRepository` interface with a query to find account balance based on timestamp filter and added unit tests for it. `AccountBalanceRepository` has a method `findHistoricalAccountBalanceUpToTimestamp` which calculates the account balance by account id up to the specified block timestamp and sums up the crypto transfers and adds them to the balance. Those two will be needed in the future to calculate entity balance.
* Add documentation in code for all repository methods.

**Related issue(s)**:

Fixes #6812

**Checklist**


hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/repository/CryptoTransferRepositoryTest.java
```diff
@@ -0,0 +1,42 @@
...
+    @Test
+    void findById() {
+        var cryptoTransfer1 = domainBuilder.cryptoTransfer().persist();
+        var cryptoTransfer2 = domainBuilder.cryptoTransfer().persist();
+        Assertions.assertThat(cryptoTransferRepository.findByIdAndTimestamp(
+                        cryptoTransfer1.getEntityId(), cryptoTransfer1.getConsensusTimestamp()))
+                .get()
+                .isEqualTo(cryptoTransfer1);
+
+        Assertions.assertThat(cryptoTransferRepository.findAll()).isEqualTo(List.of(cryptoTransfer1, cryptoTransfer2));
```
bilyana-gospodinova, line 40: This is not relevant for a findById test case or at least to me the placement should not be at the end of the test. It makes more sense to persist the two crypto transfers -> assert that they are persisted correctly (this line here) -> assert that findByIdAndTimestamp returns the correct crypto transfer.

kselveliev: Comment addressed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/EntityRepository.java
```diff
@@ -36,39 +36,43 @@ public interface EntityRepository extends CrudRepository<Entity, Long> {
...
+    @Query(value =
+            "SELECT * FROM ( "
+          + " (SELECT * FROM entity "
+          + "  WHERE evm_address = ?1 "
+          + "  AND deleted = false "
+          + "  UNION ALL "
+          + "  SELECT * FROM entity_history "
+          + "  WHERE evm_address = ?1 "
+          + "  AND deleted = false) "
+          + ") AS t "
```
IvanKavaldzhiev, line 51: It could be changed to `AS e`

kselveliev: Comment addressed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/AccountBalanceRepository.java
```diff
@@ -0,0 +1,35 @@
...
+package com.hedera.mirror.web3.repository;
+
+import com.hedera.mirror.common.domain.balance.AccountBalance;
+import java.util.Optional;
+import org.springframework.data.jpa.repository.Query;
+import org.springframework.data.repository.CrudRepository;
+
+public interface AccountBalanceRepository extends CrudRepository<AccountBalance, AccountBalance.Id> {
+    @Query(
+            value = "SELECT * FROM account_balance "
```
steven-sheehy, line 26: Use lower case for SQL everywhere.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/AccountBalanceRepository.java
```diff
@@ -0,0 +1,35 @@
...
+                    + "WHERE "
```
steven-sheehy, line 27: These queries span too many lines. I think it's better to combine the where clause into a single line and we should use `"""` for multiline strings. Same comment elsewhere.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/CryptoTransferRepository.java
```diff
@@ -0,0 +1,37 @@
...
+    @Query(
+            value = "SELECT * FROM crypto_transfer "
+                    + "WHERE "
+                    + "entity_id = ?1 AND "
+                    + "amount = ?2 AND "
+                    + "consensus_timestamp < ?3 "
+                    + "ORDER BY consensus_timestamp DESC "
+                    + "LIMIT 1 ",
+            nativeQuery = true)
+    Optional<CryptoTransfer> findByIdAndTimestampLessThan(Long entityId, long amount, Long consensusTimestamp);
```
steven-sheehy, line 36: Again, don't see the value in this query. The notion doc only mentions querying transfer tables to sum their amount between two timestamps. Is that what you intended here? If so, should sum the amounts and return a Long

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/EntityRepository.java
```diff
@@ -32,4 +33,50 @@ public interface EntityRepository extends CrudRepository<Entity, Long> {
...
+            unless = "#result == null")
+    @Query(value =
+            "SELECT * FROM ( "
+          + " (SELECT * FROM entity "
+          + "  WHERE evm_address = ?1 "
+          + "  AND deleted = false "
+          + "  AND lower(timestamp_range) < ?2"
+          + "  UNION ALL "
+          + "  SELECT * FROM entity_history "
+          + "  WHERE evm_address = ?1 "
```
steven-sheehy, line 50: We don't have an index on `entity_history.evm_address` so this would be slow. Since evm_address is immutable we could just query for it against `entity.evm_address`. We don't need to query against the history table. Since it's possible an account could be deleted and a different account assign the same EVM address you'll also need to order by timestamp_range and limit to return the appropriate one.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/EntityRepository.java
```diff
@@ -32,4 +33,50 @@ public interface EntityRepository extends CrudRepository<Entity, Long> {
...
+    @Query(value =
+           "SELECT * FROM ( "
+           + " (SELECT * FROM entity "
+           + "  WHERE id = ?1 "
+           + "  AND deleted = false "
+           + "  AND lower(timestamp_range) < ?2"
+           + "  UNION ALL "
+           + "  SELECT * FROM entity_history "
+           + "  WHERE id = ?1 "
+           + "  AND deleted = false "
```
steven-sheehy, line 74: Move the where clauses to the outer query and use range operators

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/AccountBalanceRepository.java
kselveliev: Comment addressed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/AccountBalanceRepository.java
kselveliev: Comment addressed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/EntityRepository.java
kselveliev: Comment addressed. Modified the query to search only in entity table.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/CryptoTransferRepository.java
kselveliev: Understood. This repository is removed and its tests. Added the new logic in AccountBalanceRepository where the amount is calculated summing up the crypto transfer. The query is made following the structure proposed [here](https://github.com/hashgraph/hedera-mirror-node/pull/6962#discussion_r1347939728)

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/EntityRepository.java
```diff
@@ -32,4 +33,62 @@ public interface EntityRepository extends CrudRepository<Entity, Long> {
...
+     */
+    @Query(
+            value =
+                    """
+                    select *
+                    from entity
+                    where evm_address = ?1 and lower(timestamp_range) <= ?2
+                    and deleted = false
+                    order by timestamp_range desc
+                    limit 1
```
xin-hedera, line 53: 1. The query should union all `entity` and `entity_history`
2. use `delete is not true` since delete is a nullable column. `null is not true` evaluates to be `true` while `null = false` evaluates to `null`

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/EntityRepository.java
```diff
@@ -32,4 +33,62 @@ public interface EntityRepository extends CrudRepository<Entity, Long> {
...
+     *         If there is no record found for the given criteria, an empty Optional is returned.
+     */
+    @Query(
+            value =
+                    """
+                    (
+                        select *
+                        from entity
+                        where id = ?1 and lower(timestamp_range) <= ?2
+                        and deleted = false
```
xin-hedera, line 78: deleted is not true

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/EntityRepository.java
```diff
@@ -32,4 +33,62 @@ public interface EntityRepository extends CrudRepository<Entity, Long> {
...
+                    )
+                    union all
+                    (
+                        select *
+                        from entity_history
+                        where id = ?1 and lower(timestamp_range) <= ?2
+                        and deleted = false
```
xin-hedera, line 85: deleted is not true

xin-hedera (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/EntityRepository.java
kselveliev: Commend addressed.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/EntityRepository.java
kselveliev: Comment addressed.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/EntityRepository.java
kselveliev: Addressed part 2 of the comment.
However part 1 clashes with Steven's comment [above](https://github.com/hashgraph/hedera-mirror-node/pull/6966#discussion_r1345145298). 
Can you check it and let me know if we should use the entity_history table when making query by evm_address.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/repository/EntityRepositoryTest.java
```diff
@@ -71,4 +73,159 @@ void findByEvmAddressAndDeletedTrueCall() {
...
+    @Test
+    void findHistoricalEntityByIdAndTimestampRangeAndDeletedTrueCall() {
+        EntityHistory entityHistory =
+                domainBuilder.entityHistory().customize(e -> e.deleted(true)).persist();
+
+        assertThat(entityRepository.findByIdAndTimestampAndDeletedIsFalse(
+                        entityHistory.getId(), entityHistory.getCreatedTimestamp()))
+                .isEmpty();
+    }
+
```
bilyana-gospodinova, line 197: Suggestion for one more junit test:
- Persist two entries with the same fields in EntityHistory. They will only differ in the timestamp range.
- Assert that the new repository methods return the entry with the latest lower timestamp.

kselveliev: Comment addressed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/EntityRepository.java
steven-sheehy: My comment was to separate the evm address lookup from the find entity at block lookup. For evm address lookup we can take advantage of the fact that for any given evm_address it will only be active for one account during a particular time range. So we can do this query:

```sql
select * from entity where evm_address = ?1 and created_timestamp <= ?2 order by created_timestamp desc limit 1;
```

For entity lookup we will still need do union with history but just looking at id from previous evm address lookup.

steven-sheehy (changes requested):

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/repository/EntityRepositoryTest.java
```diff
@@ -71,4 +74,167 @@ void findByEvmAddressAndDeletedTrueCall() {
...
+    private void assertEntityFields(EntityHistory entityHistory, Optional<Entity> queryResult) {
```
bilyana-gospodinova, line 207: You can use usingRecursiveComparison() is the assert in order to avoid this large method that does the same.
Example:
```
assertThat(entityRepository.findByIdAndTimestampAndDeletedIsFalse(
                entityHistory.getId(), entityHistory.getTimestampLower()))
.get()
.usingRecursiveComparison()
.isEqualTo(entityHistory)
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/EntityRepository.java
IvanKavaldzhiev: Even though the `evm_address` could be active during a particular range, couldn't it be that it was active for a given account in a time range from the past? If that's true, won't this mean we also need the `entity_history` table?

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/repository/EntityRepositoryTest.java
kselveliev: Commend addressed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/EntityRepository.java
steven-sheehy: It will be active for a given account at some time in the past, hence why I sort and filter by the immutable created_timestamp field.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/AccountBalanceRepository.java
```diff
@@ -0,0 +1,88 @@
...
+     * @return an Optional containing the historical balance at the specified timestamp.
+     *         If there are no crypto transfers between the consensus_timestamp of account_balance and the block timestamp,
+     *         the method will return the balance present at consensus_timestamp.
+     */
+    @Query(
+            value =
+                    """
+                       with balance_snapshot as (
+                          select consensus_timestamp
+                          from account_balance_file
```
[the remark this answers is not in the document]
kselveliev: Comment addressed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/EntityRepository.java
kselveliev: I've refactored the find by evm address db query as I understood it from above conversation.
Let me know if it looks good now.

xin-hedera: It should be sth like

```sql
with contract as (
  select id
  from entity
  where evm_address = ?1 and created_timestamp <= ?2
  order by created_timestamp desc
  limit 1
)
(
  select *
  from entity e
  join contract c on c.id = e.id
  where ...
) union all
(
  select *
  from entity_history e
  join contract c on c.id = e.id
  where ...
  order by lower(timestamp_range) desc
  limit 1
)
order by timestamp_range desc
limit 1
```

xin-hedera (approved):
LGTM

steven-sheehy (approved):
LGTM

PR #6968, opened 2023-09-28
Add contract caching mechanism for acceptance tests

### LINKED ISSUE #6789: Implement contract caching mechanism
### Problem

Contracts are pretty expensive to create on main net and we are now creating the same contract(i.e. precompile contract) in multiple test suites(estimatePrecomple, precompile, call).

### Solution

Implement a caching mechanism that would get already created contract(singleton) similar to the `AccoutClient` and `TokenClient` and it would be used across all test suites.

### Alternatives

_No response_

yiliev0:
**Description**: This PR introduces a caching mechanism to retrieve already created contracts, reducing the cost and redundancy of deploying the same contract in multiple test suites. 

**Related issue(s)**:

Fixes #6789

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/AbstractFeature.java
+++ b/hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/AbstractFeature.java
@@ -35,15 +35,21 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
 import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import lombok.CustomLog;
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.core.io.Resource;
+import org.springframework.core.io.ResourceLoader;
 import org.springframework.http.HttpStatus;
 
 @CustomLog
 abstract class AbstractFeature {
     protected NetworkTransactionResponse networkTransactionResponse;
     protected ContractId contractId;
+    private static final Map<ContractResource, DeployedContract> contractIdMap = new ConcurrentHashMap<>();
 
     @Autowired
     protected ContractClient contractClient;
@@ -55,6 +61,9 @@ abstract class AbstractFeature {
     protected ObjectMapper mapper;
 
     protected ExchangeRateResponse exchangeRates;
+
+    @Autowired
+    private ResourceLoader resourceLoader;
 
     protected long calculateCreateTokenFee(double usdFee, boolean useCurrentFee) {
         if (exchangeRates == null) {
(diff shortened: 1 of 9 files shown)


hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/AbstractFeature.java
```diff
@@ -95,23 +104,30 @@ protected MirrorTransaction verifyMirrorTransactionsResponse(MirrorNodeClient mi
...
-                            .getSdkClient()
-                            .getAcceptanceTestProperties()
-                            .getFeatureProperties()
-                            .getMaxContractFunctionGas(),
-                    initialBalance == 0 ? null : Hbar.fromTinybars(initialBalance),
-                    null);
-            contractId = verifyCreateContractNetworkResponse();
-            return new DeployedContract(fileId, contractId, compiledSolidityArtifact);
-        }
+    protected synchronized DeployedContract getContract(ContractResource contractResource)
```
0xivanov, line 107: In Java, when you have a synchronized method in an abstract class and you create two different subclasses that inherit from this abstract class, the synchronization is based on the current object instance. Each subclass instance will have its own lock associated with it.
In this case we can modify the code like this, in order to synchronize on a shared object:
``` java
protected DeployedContract getContract(ContractResource contractResource) {
    synchronized (contractIdMap) {
        // the rest of the method
    }
}
```

0xivanov: As mentioned [here](https://stackoverflow.com/questions/14947723/is-concurrenthashmap-totally-safe), `ConcurrentHashMap` is not 100% thread safe when we do sequence of get and put operations. That is why we need to make the whole method atomic IMO.

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/AbstractFeature.java
```diff
@@ -95,23 +104,30 @@ protected MirrorTransaction verifyMirrorTransactionsResponse(MirrorNodeClient mi
...
+                                .getAcceptanceTestProperties()
+                                .getFeatureProperties()
+                                .getMaxContractFunctionGas(),
+                        contractResource.initialBalance == 0 ? null : Hbar.fromTinybars(contractResource.initialBalance),
+                        null);
+                ContractId contractId = verifyCreateContractNetworkResponse();
+
+                return new DeployedContract(fileId, contractId, compiledSolidityArtifact);
+            } catch (IOException e) {
+                throw new RuntimeException(e);
```
0xivanov, line 128: Might be better to catch all exceptions and log some additional info. For example                 
`log.warn("Issue creating contract: {}, ex: {}", contractResource, e);`

yiliev0: Done

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/AbstractFeature.java
yiliev0: Implemented, thanks for the heads up.

attest08:
LGTM

0xivanov (approved):
LGTM

IvanKavaldzhiev (approved):
LGTM

steven-sheehy (approved):
LGTM

PR #6974, opened 2023-09-28
Update accounts and balances to use entity.balance_timestamp

### LINKED ISSUE #6701: Use balance timestamp in REST API
### Problem

The new balance timestamp added in https://github.com/hashgraph/hedera-mirror-node/issues/6683 will provide a more accurate timestamp for when the balance was last updated.

### Solution

* Change `/api/v1/accounts` to use `entity.balance_timestamp`
* Change `/api/v1/accounts/{id}` to use `entity.balance_timestamp`
* Change `/api/v1/balances` to use `entity.balance_timestamp`

### Alternatives

_No response_

edwin-greene:
The accounts and balances endpoints will now utilize the entity.balance_timestamp.

**Related issue(s)**:

Fixes #6701 

**Notes for reviewer**:

**Checklist**


hedera-mirror-rest/__tests__/specs/balances/no-params.json
steven-sheehy, line 1: It's strange that this spec is called no-params yet has a test with the timestamp param.

hedera-mirror-rest/accounts.js
```diff
@@ -408,31 +408,7 @@ const getOneAccount = async (req, res) => {
...
-    //Allow type coercion as the neValues will always be bigint and balanceFileTs may be a number
-    const timestampExcluded = neValues.some((value) => value == balanceFileTs);
-
-    if (timestampExcluded) {
-      throw new NotFoundError('Not found');
-    }
-
-    accountBalanceQuery.query = `ab.account_id = e.id and ab.consensus_timestamp = $${++paramCount}`;
-    accountBalanceQuery.params = [balanceFileTs ?? null];
+    accountBalanceQuery.query = `ab.account_id = e.id and ab.consensus_timestamp = e.balance_timestamp`;
```
steven-sheehy, line 411: We can't join account_balance timestamp which is generated by 15 minute snapshot process with the balance_timestamp. We should only join on account_balance for when timestamp parameter is passed, in which case we should not change that flow.

steven-sheehy (changes requested):

hedera-mirror-rest/__tests__/specs/accounts/{id}/timestamp.json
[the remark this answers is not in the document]
edwin-greene: Thanks I have removed them.

hedera-mirror-rest/__tests__/specs/balances/no-params.json
edwin-greene: I removed the timestamp param test and the setup related to it except for a few items that demonstrated that the endpoint discerns between entries with older timestamps and entries with more recent ones.

hedera-mirror-rest/accounts.js
edwin-greene: Thanks I have removed this change.

hedera-mirror-rest/accounts.js
```diff
@@ -42,7 +42,7 @@ const processRow = (row) => {
       ? null
       : {
           balance: row.balance,
-          timestamp: utils.nsToSecNs(row.consensus_timestamp),
+          timestamp: utils.nsToSecNs(row.balance_timestamp),
```
steven-sheehy, line 45: I don't see where we're removing the `select max(consensus_end) from record_file` sub-query. There might be multiple.

steven-sheehy (changes requested):

hedera-mirror-rest/accounts.js
edwin-greene: Thanks for spotting those, I have removed them.

xin-hedera (changes requested):

xin-hedera (approved):
LGTM

steven-sheehy (approved):

Issue #6969, opened 2023-09-28
HIP-584: Enhance ContractController to return NOT_SUPPORTED error code for block values “pending”, “safe”, “finalized”


kselveliev:
### Problem

Currently only "latest" block value is supported by api calls.
With the introduction of the historical support "earliest" and hex value of block number will also be supported.

Values "pending", "safe", "finalized" will not be supported and this change needs to be added.

### Solution

Modify ContractController so that NOT_SUPPORTED error codes are returned if block with value "pending", "safe", "finalized" is passed.
Add tests for the new changes.

PR #6975, opened 2023-09-29
Enhance ContractController to handle all remaining block types (pending, safe, finalized)

bilyana-gospodinova:
**Description**:
ContractController is modified, so that block with value "pending", "safe", "finalized" is treated as a "latest" block and could be passed by users.

Fixes https://github.com/hashgraph/hedera-mirror-node/issues/6969


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/ContractController.java
```diff
@@ -144,6 +151,13 @@ private Mono<GenericErrorResponse> invalidJson(final ServerWebInputException e)
         return errorResponse(e.getReason(), "Unable to parse JSON", StringUtils.EMPTY);
     }
 
+    @ExceptionHandler
+    @ResponseStatus(BAD_REQUEST)
```
IvanKavaldzhiev, line 155: I think it would be better to return `NOT_IMPLEMENTED` response status, so that the user knows we don't support these values and it's not a user mistake passing such block values.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/viewmodel/BlockType.java
```diff
@@ -16,15 +16,20 @@
...
     public static final BlockType PENDING = new BlockType("pending", Long.MAX_VALUE);
+    public static final BlockType SAFE = new BlockType("safe", Long.MAX_VALUE);
+    public static final BlockType FINALIZED = new BlockType("finalized", Long.MAX_VALUE);
     private static final String HEX_PREFIX = "0x";
 
+    private static final BlockType[] UNSUPPORTED_TYPES = {PENDING, SAFE, FINALIZED};
+
     public static BlockType of(final String value) {
 
         if (StringUtils.isEmpty(value) || BlockType.LATEST.name().equalsIgnoreCase(value)) {
```
IvanKavaldzhiev, line 35: We can put standalone check in the beginning that returns BlockType.LATEST if we don't have block values passed.

Then we can use switch statement for all other types.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/viewmodel/BlockType.java
```diff
@@ -50,4 +76,9 @@ public static BlockType of(final String value) {
             throw new IllegalArgumentException("Invalid block value: " + value, e);
         }
     }
+
+    public static boolean isSupported(String blockType) {
+        return Arrays.stream(UNSUPPORTED_TYPES)
```
steven-sheehy, line 81: This is incorrect since block can also be a hex number. We should revert to the previous logic. BlockType.of() will be called by BlockTypeDeserializer which is called by spring validation/conversion. Also, streams are slow and we shouldn't use them here.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/viewmodel/BlockType.java
```diff
@@ -16,23 +16,49 @@
...
-            return BlockType.LATEST;
-        } else if (BlockType.EARLIEST.name().equalsIgnoreCase(value)) {
-            return BlockType.EARLIEST;
-        } else if (BlockType.PENDING.name().equalsIgnoreCase(value)) {
-            return BlockType.PENDING;
+        try {
+            final BlockTypeName blockTypeName = BlockTypeName.valueOf(value.toUpperCase());
+            switch (blockTypeName) {
+                case EARLIEST -> {
+                    return new BlockType(BlockTypeName.EARLIEST.name, 0L);
```
steven-sheehy, line 54: EARLIEST and LATEST should be static values like before since they're the same for all requests.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/viewmodel/BlockType.java
```diff
@@ -16,23 +16,49 @@
...
+                }
+                case LATEST, PENDING, SAFE, FINALIZED -> {
```
steven-sheehy, line 56: We should just throw an exception if pending, safe or finalized.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/viewmodel/BlockType.java
```diff
@@ -16,23 +16,49 @@
...
+        }
 
-        if (StringUtils.isEmpty(value) || BlockType.LATEST.name().equalsIgnoreCase(value)) {
-            return BlockType.LATEST;
-        } else if (BlockType.EARLIEST.name().equalsIgnoreCase(value)) {
-            return BlockType.EARLIEST;
-        } else if (BlockType.PENDING.name().equalsIgnoreCase(value)) {
-            return BlockType.PENDING;
+        try {
+            final BlockTypeName blockTypeName = BlockTypeName.valueOf(value.toUpperCase());
```
steven-sheehy, line 51: valueOf is inefficient since it throws an exception for hex number and exceptions are relatively slow to create to be used for conditional logic. Instead, we can skip the enum and just convert value to lowercase and use switch statement with string.

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/ContractController.java
bilyana-gospodinova: As explained in Steven's comment the status will be BAD_REQUEST.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/viewmodel/BlockType.java
bilyana-gospodinova: Done

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/ContractController.java
```diff
@@ -82,6 +84,11 @@ Mono<ContractCallResponse> call(@RequestBody @Valid ContractCallRequest request)
     }
 
     private CallServiceParameters constructServiceParameters(ContractCallRequest request) {
+        final var blockType = request.getBlock();
+        if (!BlockType.isSupported(blockType.name())) {
+            throw new InvalidBlockTypeException(String.format("Unsupported block type passed: %s", blockType.name()));
+        }
```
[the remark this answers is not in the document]
bilyana-gospodinova: If we throw an exception inside BlockType.of() then the returned error will be something like "Malformed JSON" because the deserialization will throw an exception and the message will be misleading. I introduced a placeholder static UNSUPPORTED block type value for "pending", "safe", "finalized" so that we can check for this value in the ContractController and throw the proper exception that will then be caught by the exception handler and return a proper message that the block type value is not supported. What do you think about this approach?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/ContractController.java
```diff
@@ -144,6 +151,13 @@ private Mono<GenericErrorResponse> invalidJson(final ServerWebInputException e)
         return errorResponse(e.getReason(), "Unable to parse JSON", StringUtils.EMPTY);
     }
 
+    @ExceptionHandler
+    @ResponseStatus(NOT_IMPLEMENTED)
+    private Mono<GenericErrorResponse> invalidBlockType(final InvalidBlockTypeException e) {
```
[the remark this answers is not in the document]
bilyana-gospodinova: Done

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/viewmodel/BlockType.java
bilyana-gospodinova: Done

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/viewmodel/BlockType.java
bilyana-gospodinova: Done

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/viewmodel/BlockType.java
bilyana-gospodinova: Please check my other comment.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/viewmodel/BlockType.java
bilyana-gospodinova: Done

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/viewmodel/BlockType.java
```diff
@@ -16,23 +16,33 @@
...
 
 public record BlockType(String name, long number) {
 
+    private static final String HEX_PREFIX = "0x";
+
     public static final BlockType EARLIEST = new BlockType("earliest", 0L);
     public static final BlockType LATEST = new BlockType("latest", Long.MAX_VALUE);
-    public static final BlockType PENDING = new BlockType("pending", Long.MAX_VALUE);
-    private static final String HEX_PREFIX = "0x";
+    public static final BlockType UNSUPPORTED = new BlockType("", 0L);
```
Nana-EC, line 28: Q: should the number differ from EARLIEST?
This will avoid misconception should someone ever compare the block number or use it for inference.
```suggestion
    public static final BlockType UNSUPPORTED = new BlockType("", -1L);
```
or if you make number an Integer (not making an argument for it)
```suggestion
    public static final BlockType UNSUPPORTED = new BlockType("", null);
```

Nana-EC (changes requested):
I think we should incorporate finality into the  block considerations

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/viewmodel/BlockType.java
```diff
@@ -16,23 +16,33 @@
...
-        } else if (BlockType.EARLIEST.name().equalsIgnoreCase(value)) {
-            return BlockType.EARLIEST;
-        } else if (BlockType.PENDING.name().equalsIgnoreCase(value)) {
-            return BlockType.PENDING;
+        final String blockTypeName = value.toLowerCase();
+        switch (blockTypeName) {
+            case "earliest" -> {
+                return EARLIEST;
+            }
+            case "latest" -> {
```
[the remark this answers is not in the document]
IvanKavaldzhiev: That's an interesting perspective.

In the first place, it should be uncommon and wrong a user to send such a block value to the Hedera network, so it would be a bit of a corner case.

The Hedera architecture, indeed embeds finality for each new block/record file and the `safe`/`pending`/`finalized` block behaviour should be already incorporated into the `latest` block we have.

The user shouldn't expect to pass for examples 2 requests with `pending` and `latest` block and to expect different results, so we should be safe of applying the same logic for all these values.

It would be also great to await for @steven-sheehy opinion before finalizing this PR.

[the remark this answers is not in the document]
steven-sheehy: Sounds reasonable to me. It's better to be more permissive of user input if possible.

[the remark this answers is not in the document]
bilyana-gospodinova: Changed as requested.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/viewmodel/BlockType.java
bilyana-gospodinova: UNSUPPORTED is now removed as it is no longer needed after the changes from https://github.com/hashgraph/hedera-mirror-node/pull/6975#discussion_r1351090690.

IvanKavaldzhiev (dismissed):
LGTM

Nana-EC (approved):
LG

IvanKavaldzhiev:
We can add default statement to  the switch, which invokes an extracted method for parsing numeric block values, so that we resolve code smells. Then we can merge the PR.

kselveliev (approved):

Nana-EC (approved):

PR #7000, opened 2023-10-02
Improve the efficiency of loading contract initcode

### LINKED ISSUE #6980: Improve efficiency of loading contract initcode
### Problem

Loading contract initcode can be less than ideal in some cases.

### Solution

Improve its efficiency

### Alternatives

_No response_

mgoelswirlds:
**Description**:

Improving the efficiency of loading contract initcode.

This PR modifies
* The contracts by id call to filter by some file_ids.

**Related issue(s)**:

Fixes #6980

**Notes for reviewer**:

**Checklist**


hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/repository/FileDataRepositoryTest.java
```diff
@@ -93,4 +95,17 @@ void getFileForFeeSchedules() throws InvalidProtocolBufferException {
...
+    void getNullBytesForLargeFile() {
+        domainBuilder
+                .fileData()
+                .customize(f -> f.fileData(feeSchedules.toByteArray())
+                        .entityId(LARGE_FILE_ENTITY_ID)
+                        .consensusTimestamp(expiry))
+                .persist();
+
+        final var actualBytes = fileDataRepository.getFileAtTimestamp(LARGE_FILE_ENTITY_ID.getId(), expiry);
+        assertNull(actualBytes);
```
steven-sheehy, line 109: Prefer not to hardcode fields unless necessary or increase their scope. Also always use assertj.
```suggestion
        var fileId = EntityId.of(0L, 0L, 151L);
        var file = domainBuilder.fileData().customize(f -> f.entityId(fileId).persist();

        assertThat(fileDataRepository.getFileAtTimestamp(fileId.getId(), file.getConsensusTimestamp()).isNull();
```

hedera-mirror-rest/service/fileDataService.js
```diff
@@ -63,7 +63,8 @@ class FileDataService extends BaseService {
            ) data
         from ${FileData.tableName}
         where
-           ${FileData.ENTITY_ID} = $1
+           ${FileData.ENTITY_ID} = $1 
+        and ${FileData.ENTITY_ID} > 1000 -- < 1000 would be invalid user files
```
steven-sheehy, line 67: This `getFileData` is currently only used to retrieve initcode, but in theory it could be used elsewhere in the future. We should probably move this check into a util and check it in code.

hedera-mirror-rest/service/fileDataService.js
```diff
@@ -63,7 +63,8 @@ class FileDataService extends BaseService {
...
         and ${FileData.CONSENSUS_TIMESTAMP} >= (
         select ${FileData.CONSENSUS_TIMESTAMP}
```
steven-sheehy, line 69: There's a couple other places that call `FileDataService.getLatestFileDataContents()` with an callDataId that we also need to cover. But we can't modify this shared query so need to check in code via the util.

steven-sheehy (changes requested):

hedera-mirror-rest/service/fileDataService.js
mgoelswirlds: If we do this check in the code, what should we return for file data when the ids don't match? 0x?

steven-sheehy: 0x sounds good. Basically just skip query if file id is in bad range

mgoelswirlds:
Still working on adding test cases for the callDataId case.

hedera-mirror-rest/controllers/contractController.js
```diff
@@ -45,6 +45,7 @@ import {
   ContractStateViewModel,
   ContractViewModel,
 } from '../viewmodel';
+import {isValidUserFileId} from '../utils';
```
xin-hedera, line 48: duplicate import since the whole `utils` module is imported on line 38

hedera-mirror-rest/controllers/contractController.js
```diff
@@ -699,7 +700,7 @@ class ContractController extends BaseController {
       throw new NotFoundError();
     }
     const contract = rows[0];
-    if (contract.file_id !== null) {
+    if (contract.file_id !== null && isValidUserFileId(contract.file_id)) {
```
xin-hedera, line 703: change `isValidUserFileId` to first do `_.isNil`, then this line can be simplified to
```suggestion
    if (utils.isValidUserFileId(contract.file_id)) {
```

hedera-mirror-rest/controllers/contractController.js
```diff
@@ -966,7 +967,7 @@ class ContractController extends BaseController {
     const ethTransaction = ethTransactions[0];
 
     let fileData = null;
-    if (ethTransaction && !_.isNil(ethTransaction.callDataId)) {
+    if (ethTransaction && !_.isNil(ethTransaction.callDataId) && isValidUserFileId(ethTransaction.callDataId)) {
```
xin-hedera, line 970: simplify this line the other same line to
```suggestion
    if (utils.isValidUserFileId(ethTransaction?.callDataId)) {
```

hedera-mirror-rest/service/fileDataService.js
```diff
@@ -63,7 +63,7 @@ class FileDataService extends BaseService {
            ) data
         from ${FileData.tableName}
         where
-           ${FileData.ENTITY_ID} = $1
+           ${FileData.ENTITY_ID} = $1
```
xin-hedera, line 66: revert the extra space at the end?

hedera-mirror-rest/utils.js
```diff
@@ -410,6 +410,10 @@ const isValidContractIdQueryParam = (op, val) => {
   return EntityId.isValidEntityId(val, false);
 };
 
+const isValidUserFileId = (val) => {
+  return val > 1000;
```
xin-hedera, line 414: do `_.isNil` first, and add unit test

xin-hedera (changes requested):

hedera-mirror-rest/utils.js
```diff
@@ -410,6 +410,10 @@ const isValidContractIdQueryParam = (op, val) => {
   return EntityId.isValidEntityId(val, false);
 };
 
+const isValidUserFileId = (val) => {
```
xin-hedera, line 413: missing unit test

mgoelswirlds: Ahh yes.

hedera-mirror-rest/__tests__/utils.test.js
```diff
@@ -2046,3 +2046,13 @@ describe('Utils extractTimestampRangeConditionFilters', () => {
...
 });
+
+describe('Utils isValidUserFileId', () => {
+  test('Null', () => expect(utils.isValidUserFileId(null)).toBeFalse());
+  test('Empty', () => expect(utils.isValidUserFileId('')).toBeFalse());
+  test('Undefined', () => expect(utils.isValidUserFileId(undefined)).toBeFalse());
+  test('1001', () => expect(utils.isValidUserFileId('1001')).toBeTrue());
+  test('5001', () => expect(utils.isValidUserFileId('1001')).toBeTrue());
+  test('1000', () => expect(utils.isValidUserFileId('1000')).toBeFalse());
+  test('150', () => expect(utils.isValidUserFileId('150')).toBeFalse());
```
xin-hedera, line 2057: better test with `Number` or `BigInt` since that's the type passed to the function in the contract controller

```suggestion
  test('1001', () => expect(utils.isValidUserFileId(1001)).toBeTrue());
  test('5001', () => expect(utils.isValidUserFileId(5001)).toBeTrue());
  test('1000', () => expect(utils.isValidUserFileId(1000)).toBeFalse());
  test('150', () => expect(utils.isValidUserFileId(150)).toBeFalse());
```

steven-sheehy (approved):
LGTM

xin-hedera (approved):
LGTM

PR #7003, opened 2023-10-03
Disable shard replication by default

### LINKED ISSUE #6972: Update Stackgres Deployment To Not Use Replicas
### Problem

We no longer want to deploy our citus cluster with read replicas. 

### Solution

Update the configuration to have only one instance per shard.

### Alternatives

_No response_

jnels124:
**Description**:
disable replicas for workers

**Related issue(s)**:

Fixes #6972 

**Notes for reviewer**:
Only disabled replicas for workers.

**Checklist**

### DIFF
--- a/charts/hedera-mirror/values-prod.yaml
+++ b/charts/hedera-mirror/values-prod.yaml
@@ -121,6 +121,7 @@ stackgres:
       storageClass: zfs
   worker:
     instances: 3
+    replicasPerInstance: 1
     resources:
       cpu: 3700m
       memory: 24Gi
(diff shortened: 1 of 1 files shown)


charts/hedera-mirror/values-prod.yaml
```diff
@@ -121,7 +121,6 @@ stackgres:
       storageClass: zfs
   worker:
     instances: 3
-    replicasPerInstance: 2
```
steven-sheehy, line 124: Prefer we keep the property but explicitly set it to 1.

jnels124: got it. Removed it from here as we specify the default in values.yaml. Not against having it here to though.

xin-hedera (approved):
LGTM

steven-sheehy (approved):
LGTM

steven-sheehy:
Has this change been applied to the cluster yet?

Issue #7016, opened 2023-10-05
Table size on disk metrics


steven-sheehy:
### Problem

As a mirror node operator, I'd like to have a way to track the size the database tables take up on disk.

### Solution

* Add a `database.index.bytes` metric
* Add a `database.table.bytes` metric
* Add graphs for the above

PR #7017, opened 2023-10-06
Add table disk size metrics & graphs

steven-sheehy:
**Description**:

* Add a `database.index.bytes` metric
* Add a `database.table.bytes` metric
* Add `imagePullSecrets`, `nodeSelector`, `priorityClassName`, and `tolerations` to monitor deployment
* Add `Table Size`, `Index Size`, and `Total Size` graphs to the importer dashboard
* Add `Cache Hit`, `Cache Size`, `Cache Get Rate` and `Cache Put Rate` graphs to grpc, importer, and web3 dashboards
* Add `recordStats` option to all Caffeine caches
* Fix cache metrics not being registered by using static cache names
* Rename `database.table.size` to `database.table.rows` to disambiguate with new metrics

**Related issue(s)**:

Fixes: #7016

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-grpc/src/main/java/com/hedera/mirror/grpc/config/CacheConfiguration.java
+++ b/hedera-mirror-grpc/src/main/java/com/hedera/mirror/grpc/config/CacheConfiguration.java
@@ -19,6 +19,7 @@ package com.hedera.mirror.grpc.config;
 import com.github.benmanes.caffeine.cache.Caffeine;
 import com.hedera.mirror.grpc.GrpcProperties;
 import com.hedera.mirror.grpc.service.AddressBookProperties;
+import java.util.Set;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.cache.CacheManager;
 import org.springframework.cache.annotation.EnableCaching;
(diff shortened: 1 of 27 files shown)


edwin-greene (approved):
Looks good

hedera-mirror-grpc/src/main/java/com/hedera/mirror/grpc/config/CacheConfiguration.java
```diff
@@ -35,22 +36,27 @@ public class CacheConfiguration {
     public static final String ADDRESS_BOOK_ENTRY_CACHE = "addressBookEntryCache";
     public static final String NODE_STAKE_CACHE = "nodeStakeCache";
     public static final String ENTITY_CACHE = "entityCache";
+    public static final String CACHE_NAME = "default";
 
     @Bean(ADDRESS_BOOK_ENTRY_CACHE)
     CacheManager addressBookEntryCache(AddressBookProperties addressBookProperties) {
         CaffeineCacheManager caffeineCacheManager = new CaffeineCacheManager();
+        caffeineCacheManager.setCacheNames(Set.of(CACHE_NAME)); // We have to eagerly set cache name to register metrics
```
steven-sheehy, line 44: The `CacheManager` abstraction doesn't provide much value for us, especially if we now have to use static cache names. Going forward, we should probably just create one `CacheManager` for every cache instead of trying to reuse them if they have the same configuration. That way, we can always hardcode a static `default` cache name and we won't break one cache because we adjusted it for a different cache using the same `CacheManager`.

xin-hedera (approved):
LGTM

PR #7035, opened 2023-10-09
Add deduplication of entries for Account Balance and Token Balance 

### LINKED ISSUE #3507: Dedupe balance information
### Problem

The size of the token balance table on testnet is growing rapidly with every 15 minutes bringing another 230K+ rows. Likewise for account balances. We need a way to reduce the size of the tables without information loss.

### Solution

* Change the indexes for balance tables in v2 to have the account or token ID first
* Add a `hedera.mirror.importer.parser.record.historicalbalance.dedupe=false` property
* Change `HistoricalBalanceService` to use a different query if `dedupe=true` to only insert into balance table if `balance_timestamp` changed between the consensus timestamp of the last run.
* Ensure `0.0.2` always gets inserted into the balance snapshot regardless of its balance timestamp
* Ensure each partition contains each entity ID at least once
* Adjust balance queries to include `0.0.2` or spinoff to separate ticket if too many changes

### Alternatives

_No response_

edwin-greene:
Duplicate entries will not be placed in the account_balance or token_balance tables when the dedup property is set and when the current partition already has the balance for the account/token.

**Related issue(s)**:

Fixes #3507 

**Notes for reviewer**:

**Checklist**


hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/historicalbalance/HistoricalBalanceService.java
```diff
@@ -175,4 +184,31 @@ private boolean shouldGenerate(long consensusEnd) {
...
+            // There is no previous account balance snapshot, include all balances
+            return true;
+        }
+
+        var partitions = timePartitionService.getOverlappingTimePartitions(
+                ACCOUNT_BALANCE_TABLE_NAME, lastBalanceTimestamp, timestamp);
+        if (partitions.isEmpty()) {
+            // Within the current partition, do not include all balances
+            return false;
+        }
```
xin-hedera, line 203: this is wrong. if `[lastBalanceTimestamp, timestamp]` is within the current partition, `partitions.size()` will be 1.

if it's empty, it's because either the table is not partitioned or sth is seriously wrong.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/historicalbalance/HistoricalBalanceService.java
```diff
@@ -175,4 +184,31 @@ private boolean shouldGenerate(long consensusEnd) {
...
+
+        var lowerEndpoint =
+                partitions.get(partitions.size() - 1).getTimestampRange().lowerEndpoint();
+        if (lastBalanceTimestamp < lowerEndpoint && timestamp > lowerEndpoint) {
+            // Include all balances when creating a snapshot for a new partition
+            return true;
+        }
```
xin-hedera, line 210: should just return true when `partitions.size() > 1`

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/repository/AccountBalanceRepository.java
```diff
@@ -42,13 +43,39 @@ insert into account_balance (account_id, balance, consensus_timestamp)
...
+        select e.id, e.balance, :consensusTimestamp
+        from entity e left join account_balance_snapshot abs on e.id = abs.account_id
+        where e.id = 2 or
+            (e.balance is not null and
+            e.balance_timestamp is not null and
+            e.deleted is not true and
+            (abs.consensus_timestamp is null or e.balance_timestamp > abs.consensus_timestamp))
+        order by e.id
+        """)
+    @Transactional
```
xin-hedera, line 66: there is performance issue because the CTE part will scan all the time partitions to determine max(consensus_timestamp) for each account.

Instead, to create a new snapshot with deduped info, we should

1. Find the current time partition
2. Find the latest account balance snapshot in the current partition, however, return (account_id, balance) instead of (account_id, consensus_timestamp)
3. insert new row for an account when it's account 2, or its balance has changed

Note, when comparing balance to dedupe, the condition should be `(abs.balance is null or e.balance <> abs.balance)`

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/repository/AccountBalanceRepository.java
```diff
@@ -42,13 +43,39 @@ insert into account_balance (account_id, balance, consensus_timestamp)
...
+
     @Override
     @EntityGraph("AccountBalance.tokenBalances")
     List<AccountBalance> findAll();
 
     @EntityGraph("AccountBalance.tokenBalances")
     List<AccountBalance> findByIdConsensusTimestamp(long consensusTimestamp);
 
+    @Query(value = "select * from account_balance order by consensus_timestamp desc limit 1", nativeQuery = true)
+    Optional<AccountBalance> findLatest();
```
xin-hedera, line 77: Same performance issue because

1. the query doesn't use the sentinel account id 2
2. the query will go to all time partitions

so similarly, we need to add a partition range, however, we may need to search in two partitions, i.e., the current and the next, instead of just the current.

hedera-mirror-importer/src/main/resources/db/migration/v2/V2.0.3__index_init.sql
```diff
@@ -235,7 +235,7 @@ create index if not exists token_allowance_history__owner_spender_token_lower_ti
 
 -- token_balance
 alter table if exists token_balance
-    add constraint token_balance__pk primary key (consensus_timestamp, account_id, token_id);
+    add constraint token_balance__pk primary key (token_id, account_id, consensus_timestamp);
```
xin-hedera, line 238: shouldn't the index be `(account_id, token_id, consensus_timestamp)`?

xin-hedera (changes requested):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/historicalbalance/HistoricalBalanceService.java
```diff
@@ -121,7 +127,10 @@ public void onRecordFileParsed(RecordFileParsedEvent event) {
                         .map(RecordFile::getConsensusEnd)
                         // This should never happen since the function is triggered after a record file is parsed
                         .orElseThrow(() -> new ParserException("Record file table is empty"));
-                int accountBalancesCount = accountBalanceRepository.balanceSnapshot(timestamp);
+
+                int accountBalancesCount = (properties.isDedupe() && !includeAllBalances(timestamp))
```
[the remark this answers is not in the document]
edwin-greene: Thanks I have reworked this.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/historicalbalance/HistoricalBalanceService.java
edwin-greene: I see, please review again as I have reworked this area.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/historicalbalance/HistoricalBalanceService.java
edwin-greene: I have updated this, it should now work with no partitions and partitions of variable sizes.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/repository/AccountBalanceRepository.java
edwin-greene: Thanks I have refactored this query to include the partition key, it should now be more performant on v2.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/repository/AccountBalanceRepository.java
edwin-greene: Thanks I have refactored things and removed this.

hedera-mirror-importer/src/main/resources/db/migration/v2/V2.0.3__index_init.sql
edwin-greene: Got it, I have fixed this.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/historicalbalance/HistoricalBalanceService.java
```diff
@@ -175,4 +185,35 @@ private boolean shouldGenerate(long consensusEnd) {
...
+        return tokenBalanceRepository.balanceSnapshot(timestamp);
+    }
+
+    private Range<Long> getPartitionRange(long timestamp, String tableName) {
+        var partitions = timePartitionService.getTimePartitions(tableName);
+        return partitions.stream()
+                .filter(p -> p.getTimestampRange().contains(timestamp))
+                .findFirst()
+                .map(TimePartition::getTimestampRange)
+                .orElse(Range.closedOpen(LOWER_RANGE, UPPER_RANGE));
```
steven-sheehy, line 217: Can cache the entire `Range` object instead of the longs.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/repository/AccountBalanceRepository.java
```diff
@@ -42,6 +42,32 @@ insert into account_balance (account_id, balance, consensus_timestamp)
...
+        from entity e left join account_balance_snapshot abs on e.id = abs.account_id
+        where
+            e.id = 2 or
+            (e.balance is not null and
+             e.deleted is not true and
+             (abs is null or abs.balance is null or e.balance <> abs.balance))
+        order by e.id
+        """)
+    @Transactional
+    int updateBalanceSnapshot(long lowerRangeTimestamp, long upperRangeTimestamp, long consensusTimestamp);
```
steven-sheehy, line 69: Add this method to `BalanceSnapshotRepository`

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/repository/AccountBalanceRepository.java
```diff
@@ -42,6 +42,32 @@ insert into account_balance (account_id, balance, consensus_timestamp)
...
+          select account_id, balance from account_balance
+          where (account_id, consensus_timestamp)
+            in (select account_id, max(consensus_timestamp)
+                from account_balance
+                where consensus_timestamp >= :lowerRangeTimestamp and consensus_timestamp < :upperRangeTimestamp
+                group by account_id)
+        )
+        insert into account_balance (account_id, balance, consensus_timestamp)
+        select e.id, e.balance, :consensusTimestamp
+        from entity e left join account_balance_snapshot abs on e.id = abs.account_id
```
steven-sheehy, line 60: We're not even using the `entity.balance_timestamp` which is the main purpose of that column. The ticket states `only insert into balance table if balance_timestamp changed between the consensus timestamp of the last run`.

```sql
insert into account_balance (account_id, balance, consensus_timestamp)
select e.id, e.balance, :consensusTimestamp
from entity e
where e.id = 2 or (e.balance is not null and e.deleted is not true and 
e.balance_timestamp > (select max(consensus_timestamp) from account_balance where account_id = 2))
```

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/historicalbalance/HistoricalBalanceService.java
```diff
@@ -175,4 +185,35 @@ private boolean shouldGenerate(long consensusEnd) {
...
+
+        if (properties.isDedupe()) {
+            var range = getPartitionRange(timestamp, TOKEN_BALANCE_TABLE_NAME);
+            return tokenBalanceRepository.updateBalanceSnapshot(
+                    range.lowerEndpoint(), range.upperEndpoint(), timestamp);
+        }
+        return tokenBalanceRepository.balanceSnapshot(timestamp);
+    }
+
+    private Range<Long> getPartitionRange(long timestamp, String tableName) {
```
steven-sheehy, line 211: We should not need to do any partition logic here. Instead, modify the partition cron to copy the last balance snapshot from the previous snapshot to the newly created partition.

steven-sheehy (changes requested):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/repository/AccountBalanceRepository.java
xin-hedera: Ideally we should compare balance since balance timestamp was changed <> balance was changed. However, comparing to account 2's max consensus timestamp should be much faster than first finding the all the account's last balance and then comparing by balance, so agree with the approach.

xin-hedera: We still want to limit the query `select max(consensus_timestamp) from account_balance where account_id = 2` to the current partition with timestamp range `[lowerRangeTimestamp, upperRangeTimestamp)`

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/historicalbalance/HistoricalBalanceService.java
```diff
@@ -53,30 +56,39 @@
 @Named
 public class HistoricalBalanceService {
 
+    private static final String ACCOUNT_BALANCE_TABLE_NAME = "account_balance";
+    private static final String TOKEN_BALANCE_TABLE_NAME = "token_balance";
+    private static final long LOWER_RANGE = 0L;
+    private static final long UPPER_RANGE = 9223372036854775807L;
```
xin-hedera, line 62: nit

```suggestion
    private static final long UPPER_RANGE = Long.MAX_VALUE;
```

xin-hedera (changes requested):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/repository/TokenBalanceRepository.java
```diff
@@ -40,6 +40,30 @@ insert into token_balance (account_id, balance, consensus_timestamp, token_id)
...
+                where consensus_timestamp >= :lowerRangeTimestamp and consensus_timestamp < :upperRangeTimestamp
+                group by account_id, token_id)
+        )
+        insert into token_balance (account_id, balance, consensus_timestamp, token_id)
+        select ta.account_id, ta.balance, :consensusTimestamp, ta.token_id
+        from token_account ta left join token_balance_snapshot tbs on ta.account_id = tbs.account_id and
+          ta.token_id = tbs.token_id
+        where ta.associated is true and
+          (tbs is null or tbs.balance is null or ta.balance <> tbs.balance)
+        order by ta.account_id, ta.token_id\s
```
xin-hedera, line 62: interesting this didn't cause any issue

```suggestion
        order by ta.account_id, ta.token_id
```

docs/configuration.md
```diff
@@ -146,6 +146,7 @@ value, it is recommended to only populate overridden properties in the custom `a
...
 | `hedera.mirror.importer.parser.record.entity.sql.enabled`                        | true                                             | Whether to use PostgreSQL Copy mechanism to insert into the database                                                                                                                                                                                               |
 | `hedera.mirror.importer.parser.record.frequency`                                 | 100ms                                            | How often to poll for new messages. Can accept duration units like `10s`, `2m` etc.                                                                                                                                                                                |
+| `hedera.mirror.importer.parser.record.historicalBalance.dedupe`                  | false                                            | Whether to deduplicate balance information                                                                                                                                                                                                                         |
```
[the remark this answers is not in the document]
edwin-greene: Thanks I have updated it to `deduplicate`

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/historicalbalance/HistoricalBalanceService.java
edwin-greene: Good idea, I have added that.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/repository/AccountBalanceRepository.java
edwin-greene: Got it, I have updated `BalanceSnapshotRepository`.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/repository/AccountBalanceRepository.java
edwin-greene: Sounds good, the queries have now been updated to use `balance_timestamp` and they have been simplified to this version.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/historicalbalance/HistoricalBalanceService.java
```diff
@@ -175,4 +185,35 @@ private boolean shouldGenerate(long consensusEnd) {
                         >= properties.getMinFrequency().toNanos())
                 .isPresent();
     }
+
+    private int balanceSnapshot(long timestamp) {
+        if (properties.isDedupe()) {
+            var range = getPartitionRange(timestamp, ACCOUNT_BALANCE_TABLE_NAME);
+            return accountBalanceRepository.updateBalanceSnapshot(
+                    range.lowerEndpoint(), range.upperEndpoint(), timestamp);
```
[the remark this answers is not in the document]
edwin-greene: The full snapshot is now implemented in the java code.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/historicalbalance/HistoricalBalanceService.java
```diff
@@ -175,4 +185,35 @@ private boolean shouldGenerate(long consensusEnd) {
...
+
+    private int tokenBalanceSnapshot(long timestamp) {
+        if (!properties.isTokenBalances()) {
+            return 0;
+        }
+
+        if (properties.isDedupe()) {
+            var range = getPartitionRange(timestamp, TOKEN_BALANCE_TABLE_NAME);
+            return tokenBalanceRepository.updateBalanceSnapshot(
+                    range.lowerEndpoint(), range.upperEndpoint(), timestamp);
```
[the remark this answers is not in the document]
edwin-greene: This should now be seen in the java code.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/historicalbalance/HistoricalBalanceService.java
edwin-greene: Thanks I have updated it to the constant.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/repository/TokenBalanceRepository.java
edwin-greene: Thanks I have removed that typo.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/historicalbalance/HistoricalBalanceService.java
```diff
@@ -121,10 +133,22 @@ public void onRecordFileParsed(RecordFileParsedEvent event) {
...
+                            partitionRange.lowerEndpoint(), partitionRange.upperEndpoint());
+                    accountBalancesCount = updateBalanceSnapshot(
+                            lastConsensusTimestamp, maxConsensusTimestamp, partitionRange, timestamp);
+                    tokenBalancesCount = properties.isTokenBalances()
+                            ? updateTokenBalanceSnapshot(lastConsensusTimestamp, maxConsensusTimestamp, timestamp)
+                            : 0;
+                } else {
+                    accountBalancesCount = accountBalanceRepository.balanceSnapshot(timestamp);
+                    tokenBalancesCount =
+                            properties.isTokenBalances() ? tokenBalanceRepository.balanceSnapshot(timestamp) : 0;
```
xin-hedera, line 151: The logic can be simplified if we first determine if the snapshot should be full or deduped, we'd reduce duplicate code and no longer need  the function `updateBalanceSnapshot()` / `updateTokenBalanceSnapshot()`.

Overall, it should be:

```
   bool fullSnapshot = true;
   if (properties.isDeduplicate()) {
      check if this is not the first balance snapshot in a new partition, if so, set fullSnapshot to false, as well as finding other variables needed by `*BalanceRepository.updateBalanceSnapshot`
   }

  if (fullSnapshot) {
    *BalanceRepository.balanceSnapshot(...);
  } else {
    *BalanceRepository.updateBalanceSnapshot(...);
  }
```

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/historicalbalance/HistoricalBalanceService.java
```diff
@@ -121,10 +133,22 @@ public void onRecordFileParsed(RecordFileParsedEvent event) {
...
-                int accountBalancesCount = accountBalanceRepository.balanceSnapshot(timestamp);
-                int tokenBalancesCount = 0;
-                if (properties.isTokenBalances()) {
-                    tokenBalancesCount = tokenBalanceRepository.balanceSnapshot(timestamp);
+
+                int tokenBalancesCount;
+                int accountBalancesCount;
+                if (properties.isDeduplicate()) {
+                    var partitionRange = getPartitionRange(timestamp, ACCOUNT_BALANCE_TABLE_NAME);
+                    long maxConsensusTimestamp = accountBalanceRepository.getMaxConsensusTimestampInRange(
```
xin-hedera, line 141: `accountBalanceRepository.getMaxConsensusTimestampInRange` should return `Optional<Long>` and return an empty optional instead of coalescing to 0 if no value found. That way, we immediately know we need to create a full snapshot, i.e., we no longer need to compare `lastConsensusTimestamp` and `partitionRange.lowerEndPoint`

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/historicalbalance/HistoricalBalanceService.java
```diff
@@ -171,8 +195,42 @@ private boolean shouldGenerate(long consensusEnd) {
...
+                ?
+                // Between the last snapshot and this one a partition boundary has been crossed.
+                // Add all token balances to the snapshot
+                tokenBalanceRepository.balanceSnapshot(timestamp)
+                : tokenBalanceRepository.updateBalanceSnapshot(
+                        maxConsensusTimestamp, partitionRange.upperEndpoint(), timestamp);
+    }
+
+    private Range<Long> getPartitionRange(long timestamp, String tableName) {
+        var partitions = timePartitionService.getTimePartitions(tableName);
```
xin-hedera, line 229: we really should use `timePartitionService.getOverlappingTimePartitions(tableName, timestamp, timestamp)` since it's a binary search v.s. a linear search of all the partitions returned by `timePartitionService.getTimePartitions`. I'm fine to keep it as is to reduce the churn.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/historicalbalance/HistoricalBalanceService.java
```diff
@@ -171,8 +195,42 @@ private boolean shouldGenerate(long consensusEnd) {
...
+        return partitions.stream()
+                .filter(p -> p.getTimestampRange().contains(timestamp))
+                .findFirst()
+                .map(TimePartition::getTimestampRange)
+                .orElse(DEFAULT_RANGE);
```
xin-hedera, line 234: If the table is partitioned, and no partition containing `timestamp` found, shouldn't we throw an exception to fail early?

The fact is if this ever happens, when we insert data into `account_balance` table, it'll still fail because the corresponding partition doesn't exist.

xin-hedera (changes requested):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/repository/TokenBalanceRepository.java
```diff
@@ -40,6 +40,23 @@ insert into token_balance (account_id, balance, consensus_timestamp, token_id)
...
     int balanceSnapshot(long consensusTimestamp);
 
+    @Override
+    @Modifying
+    @Query(
+            nativeQuery = true,
+            value =
+                    """
+        insert into token_balance (account_id, balance, consensus_timestamp, token_id)
+        select account_id, balance, :consensusTimestamp, token_id
```
steven-sheehy, line 50: Shouldn't the `consensus_timestamp` of the balance tables be populated with the `balance_timestamp` and not the snapshot's timestamp? This would be more accurate but probably we have to wait until we stop inserting account balance file entries.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/repository/BalanceSnapshotRepository.java
```diff
@@ -25,4 +25,16 @@ interface BalanceSnapshotRepository {
...
+     * Updates the balance snapshot from state in database.
+     * Only adds entries for items with a balance_timestamp that is greater or equal to the maxConsensusTimestamp
+     * and less than the upperRangeTimestamp.
+     *
+     * @param maxConsensusTimestamp
+     * @param upperRangeTimestamp The upper range timestamp of the partition.
+     * @param consensusTimestamp  The consensus timestamp of the balance snapshot.
+     * @return The number of balance rows inserted
+     */
+    int updateBalanceSnapshot(long maxConsensusTimestamp, long upperRangeTimestamp, long consensusTimestamp);
```
steven-sheehy, line 39: You're not really updating a previous snapshot since it's only inserting rows. Maybe `balanceSnapshotDeduplicate`, `balanceSnapshotPartition`, etc.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/repository/AccountBalanceRepository.java
```diff
@@ -42,6 +42,37 @@ insert into account_balance (account_id, balance, consensus_timestamp)
...
+                    """
+        insert into account_balance (account_id, balance, consensus_timestamp)
+        select id, balance, :consensusTimestamp
+        from entity
+        where
+          id = 2 or
+          (balance is not null and
+           deleted is not true and
+           balance_timestamp > :maxConsensusTimestamp and
+           balance_timestamp < :upperRangeTimestamp)
```
xin-hedera, line 59:
```suggestion
           balance_timestamp > :maxConsensusTimestamp)
```

Check the upper bound is unnecessary. There is an inherent on the caller to guarantee that when the method is called, the entity table state should be at the time of `consensusTimestamp`.

Since the consensusTimestamp is the new account balance snapshot's timestamp, and it's inside the current partition, it should never happen that any entity's balanceTimestamp >= upperRangeTimestamp.

Instead, we can add a javadoc  to the interface method.

Note, in the historical balance service, we set transaction isolation level to repeatable reads to guarantee it.

xin-hedera (changes requested):

steven-sheehy (changes requested):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/repository/TokenBalanceRepository.java
edwin-greene: From what I can tell that should be an easy change when we are ready for it.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/historicalbalance/HistoricalBalanceService.java
```diff
@@ -171,8 +195,42 @@ private boolean shouldGenerate(long consensusEnd) {
                         .map(RecordFile::getConsensusEnd)
                         .map(timestamp ->
                                 timestamp + properties.getInitialDelay().toNanos()))
-                .filter(lastTimestamp -> consensusEnd - lastTimestamp
-                        >= properties.getMinFrequency().toNanos())
-                .isPresent();
+                .orElse(0L);
```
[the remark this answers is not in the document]
edwin-greene: Sounds good I have updated it to `Long.MAX`.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/historicalbalance/HistoricalBalanceService.java
edwin-greene: Thanks I have reworked this section, it is now similar to what you have described.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/historicalbalance/HistoricalBalanceService.java
edwin-greene: Great thanks. I have updated it to return the Optional.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/historicalbalance/HistoricalBalanceService.java
edwin-greene: I found that switching to `getOverlappingTimePartitions` could also simplify some other logic so I have updated it to use that method instead.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/historicalbalance/HistoricalBalanceService.java
edwin-greene: I updated it with a comment to throw an exception once v1 is partitioned.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/repository/AccountBalanceRepository.java
edwin-greene: Thanks I have made this change and removed the `upperRangeTimestamp`.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/repository/BalanceSnapshotRepository.java
edwin-greene: Sounds good I have refactored its name.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/historicalbalance/HistoricalBalanceService.java
```diff
@@ -171,8 +194,35 @@ private boolean shouldGenerate(long consensusEnd) {
...
+            // After v1 account_balance and token_balance are partitioned a size of 0 should throw an
+            // exception.
+            // In the meantime use default range for non-partitioned tables
+            var partitionRange =
+                    partitions.isEmpty() ? DEFAULT_RANGE : partitions.get(0).getTimestampRange();
+            maxConsensusTimestamp = accountBalanceRepository.getMaxConsensusTimestampInRange(
+                    partitionRange.lowerEndpoint(), partitionRange.upperEndpoint());
+        }
+
+        return maxConsensusTimestamp;
```
xin-hedera, line 221: let completely remove `lastConsensusTimestamp`.

the logic here should be:

``` java
var partitions = timePartitionService.getOverlappingTimePartitions(ACCOUNT_BALANCE_TABLE_NAME, timestamp, timestamp);
if (partitions.isEmpty()) {
  return Optional.empty(); // once v1 is partitioned, throw exception instead
}

var partitionRange = partitions.get(0).getTimestampRange();
return accountBalanceRepository.getMaxConsensusTimestampInRange(partitionRange.lowerEndpoint(), partitionRange.upperEndpoint);
```

So now we no longer need `lastConsensusTimestamp`, we can revert the changes to `shouldGenerate` and remove the function `getLastConsensusTimestamp()`.

Let me know if you have questions.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/repository/AccountBalanceRepository.java
```diff
@@ -42,6 +43,36 @@ insert into account_balance (account_id, balance, consensus_timestamp)
...
+            nativeQuery = true,
+            value =
+                    """
+        insert into account_balance (account_id, balance, consensus_timestamp)
+        select id, balance, :consensusTimestamp
+        from entity
+        where
+          id = 2 or
+          (balance is not null and
+           deleted is not true and
```
xin-hedera, line 58:
```suggestion
```

when dedup, we can't skip deleted entities. we need to add the deleted entity's balance into the snapshot (for timestamp at or after the entity is deleted) if balance_timestamp > max. Otherwise, the query to find the deleted entity's balance at or after the entity is deleted will return its balance before deletion.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/repository/TokenBalanceRepository.java
```diff
@@ -40,6 +40,22 @@ insert into token_balance (account_id, balance, consensus_timestamp, token_id)
...
+        from token_account
+        where associated is true and
```
xin-hedera, line 52: for the same reason in the comment for AccountBalanceRepository.balanceSnapshotDeduplicate, remove the associated is true check

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/historicalbalance/HistoricalBalanceService.java
edwin-greene: Great idea, I have updated it.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/repository/AccountBalanceRepository.java
edwin-greene: Thanks I have updated the query.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/repository/TokenBalanceRepository.java
edwin-greene: This is now fixed.

steven-sheehy (approved):
LGTM

xin-hedera (approved):
LGTM

Issue #7040, opened 2023-10-10
Estimate gas acceptance tests fail sometimes - estimated gas under expected range


bilyana-gospodinova:
### Description

When running the acceptance tests in `estimate.feature` sometimes some of them (such as `I call estimateGas with function that changes contract slot information by updating global contract field with the passed argument`) fail because the estimated gas is below the lower expected range. After debugging it seems that in some cases the contract's bytecode could not be fetched from the repository and the initial gas estimation is 21000 which is the minimum gas cost and this affects the further gas calculations.

### Steps to reproduce

Run `estimate.feature`.

### Hedera network

testnet

### Version

v0.89.0

### Operating system

None

PR #7041, opened 2023-10-10
HIP-584: Estimated gas under expected range

bilyana-gospodinova:
**Description**:
When running the acceptance tests in `estimate.feature` sometimes some of them (such as `I call estimateGas with function that changes contract slot information by updating global contract field with the passed argument`) fail because the estimated gas is below the lower expected range. After debugging it seems that in some cases the contract's bytecode could not be fetched from the repository and the initial gas estimation is 21000 which is the minimum gas cost and this affects the further gas calculations.

This PR adds check in estimate gas acceptance tests to ensure that the contract is deployed and that the bytecode will be fetched successfully in web3. As a result the gas estimation will be within the expected range.

Fixes https://github.com/hashgraph/hedera-mirror-node/issues/7040

### DIFF
--- a/hedera-mirror-test/src/test/resources/features/contract/estimate.feature
+++ b/hedera-mirror-test/src/test/resources/features/contract/estimate.feature
@@ -3,6 +3,7 @@ Feature: EstimateGas Contract Base Coverage Feature
 
   Scenario Outline: Validate EstimateGas
     Given I successfully create EstimateGas contract from contract bytes
+    Then the mirror node REST API should return status 200 for the estimate contract creation
     Given I successfully create fungible token
     And lower deviation is 5% and upper deviation is 20%
     Then I call estimateGas without arguments that multiplies two numbers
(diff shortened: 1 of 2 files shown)


hedera-mirror-test/src/test/resources/features/contract/estimate.feature
```diff
@@ -3,6 +3,7 @@ Feature: EstimateGas Contract Base Coverage Feature
 
   Scenario Outline: Validate EstimateGas
     Given I successfully create EstimateGas contract from contract bytes
+    Then the mirror node REST API should return status 200 for the estimate contract creation
```
IvanKavaldzhiev, line 6: It would be good to have this check after

`Given I successfully create fungible token`

In this way, we would validate that all HAPI transactions sent to the consensus node are ingested in the mirror-node. Then we can start performing `eth_estimateGas` requests.

yiliev0 (approved):
LGTM

kselveliev (dismissed):
LGTM

hedera-mirror-test/src/test/resources/features/contract/estimate.feature
bilyana-gospodinova: The checks are rearranged.

steven-sheehy (approved):

PR #7087, opened 2023-10-17
Fix balances API with timestamp performance in v2

### LINKED ISSUE #6936: /balances?timestamp=X Shows Performance Degradation in v2
### Problem

When running this query on a one off basis, it performs well. However, under k6 tests it consistently performs poorly.

This may be related to #5229 

### Solution

determine bottleneck and optimize

### Alternatives

_No response_

edwin-greene:
Update the balances consensus timestamp query to include an account_id. This increases performance as the account id is the first value of the composite primary key from https://github.com/hashgraph/hedera-mirror-node/issues/3507.

**Related issue(s)**:

Fixes #6936 

**Notes for reviewer**:
K6 test results on Citus: 
```
| balances_with_timestamp | /balances?timestamp=X | 1500 | 87.29 | 393.50/s | 343.49/s | 2608.49ms | No | |
```

**Checklist**


hedera-mirror-rest/balances.js
```diff
@@ -108,6 +107,9 @@ const getBalances = async (req, res) => {
   };
 
   if (tsQuery) {
+    // Add the treasury account to the query as it will always be in the balance snapshot and account_id is the primary key of the table thus it will speed up queries on v2
+    tsQuery = tsQuery.concat(' and account_id = ?');
+    tsParams.push('2');
```
jnels124, line 112: Need to make sure the account id isn't already provided on the request. Maybe best to move logic into the get balance method?

jnels124 (changes requested):

hedera-mirror-rest/balances.js
jnels124: Actually this may be more of a problem. We may not even be able to add this hint here. Reason is we need to be able to return account balances for all accounts.

hedera-mirror-rest/balances.js
```diff
@@ -158,10 +157,16 @@ const getBalances = async (req, res) => {
 };
 
 const getAccountBalanceTimestamp = async (tsQuery, tsParams, order = 'desc') => {
+  if (!tsQuery || !tsQuery.includes('account_id')) {
+    // Add the treasury account to the query as it will always be in the balance snapshot and account_id is the primary key of the table thus it will speed up queries on v2
+    tsQuery = tsQuery ? tsQuery.concat(' and account_id = ?') : ' account_id = ?';
+    tsParams.push('2');
+  }
```
xin-hedera, line 164: `tsQuery` built by the caller shouldn't have any `account_id = ` query.
  
```suggestion
  // Add the treasury account to the query as it will always be in the balance snapshot and account_id is the primary key of the table thus it will speed up queries on v2
  tsQuery = tsQuery ? tsQuery.concat(' and account_id = ?') : ' account_id = ?';
  tsParams.push('2');
```

xin-hedera (changes requested):

hedera-mirror-rest/balances.js
xin-hedera: `tsQuery` passed in from the caller shouldn't have any `account_id = ` query

edwin-greene: Thanks I have updated this and removed the check.

hedera-mirror-rest/balances.js
edwin-greene: Got it, I have removed the if statement

hedera-mirror-rest/balances.js
```diff
@@ -108,24 +111,34 @@ const getBalances = async (req, res) => {
...
+      tsParams = [balanceTimestamp, balanceTimestamp];
+    } else {
+      // For queries with account(s) specified, a timestamp range including the lower bound of the partition increases performance
+      consensusTsQuery = `ab.consensus_timestamp >= ? and ab.consensus_timestamp <= ?`;
+      // Get partition lower bound by subtracting 31 days from the timestamp
+      // Queries with this lower bound will hit two partitions at most
+      const upperBound = tsParams[0];
+      let oneMonthTimestamp = math.subtract(upperBound, NS_IN_MONTH);
+      const lowerPartitionBound = oneMonthTimestamp >= 0 ? oneMonthTimestamp : 0;
+      tsParams = [lowerPartitionBound, upperBound, lowerPartitionBound, upperBound];
```
xin-hedera, line 127: let's not implement queries assuming account balance is deduped before the ingestion side is complete and data is populated.

There is an issue for deleted accounts with full account balance snapshot v.s. with deduped account balance snapshot:

- for full account balance snapshot, once an account is deleted, its balance info disappears from any account balance file after that, so the balances query would simply not return balance info for the account when the timestamp is after the deletion
- for deduped account balance snapshot, the historical balance service should have one last row for the account with 0 balance after its deletion, the modified query would then return 0 balance for the account, that's good

However, the problem arises when we use the modified query with full account balance snapshot, since then the balances API will show the last balance of those deleted accounts.

I don't think there is a way to use the same query for both types of account balance info, besides the `distinct on ... order by ..., ... desc limit 1` also has performance impact. We may end up with two sets of different queries and add a property to rest api configuration to switch between.

xin-hedera (changes requested):

hedera-mirror-rest/balances.js
edwin-greene: Thanks. I have added a configuration for deduplicate and added back the original queries for non-deduplicated data.

hedera-mirror-rest/balances.js
```diff
@@ -170,6 +178,61 @@ const getAccountBalanceTimestamp = async (tsQuery, tsParams, order = 'desc') =>
...
+  tsQuery
+) => {
+  let consensusTsQuery = `ab.consensus_timestamp = ?`;
+  if (!accountQuery) {
+    // Use a single balanceTimestamp to improve query performance
+    const balanceTimestamp = await getAccountBalanceTimestamp(tsQuery, tsParams, order);
+    tsParams = [balanceTimestamp, balanceTimestamp];
+  } else {
+    // For queries with account(s) specified, a timestamp range including the lower bound of the partition increases performance
+    consensusTsQuery = `ab.consensus_timestamp >= ? and ab.consensus_timestamp <= ?`;
```
jnels124, line 212: Doesn't this miss if the user was to query for a specific timestamp and not provide an account query? Need to be able to handle gt,lt, etc

jnels124 (changes requested):
Looks like the schema in the db doesn't match what is in our migration files. Please update the index init migration to update the order of the index.

hedera-mirror-rest/balances.js
```diff
@@ -170,6 +178,61 @@ const getAccountBalanceTimestamp = async (tsQuery, tsParams, order = 'desc') =>
...
+    // Get partition lower bound by subtracting 31 days from the timestamp
+    // Queries with this lower bound will hit two partitions at most
+    const upperBound = tsParams[0];
+    let oneMonthTimestamp = math.subtract(upperBound, NS_IN_MONTH);
```
jnels124, line 216: This still potentially scans multiple partitions. Couldn't we just roll/roll-truncate the input to stay in the correct partition?

jascks: Are you saying that since `NS_IN_MONTH` represents 31 days, depending on where `upperBound` falls in the month/partition, subtracting 31 days could involve up to three partitions, if the previous partition represents a month of less than 31 days?

What is the "correct partition"? I realize the fewer partitions accessed the better, but is there some heuristic to develop like if day of month is > 25, then truncate to start of month and involve only a single partition, else generate a query that would access two partitions?

Then for, say, `upperBound` defined somewhere in March 1, subtracting 31 days would hit January, or a three partition query. Truncation here would drop January and just utilize two partitions?

If this is the case, then this code would be useful outside of this file. I know I would use it. 😄

xin-hedera: then we'll need a library to handle natural month, since a month can have 28/29/30/31 days, with that, the boundary should exactly match the partitions in db.

hedera-mirror-rest/balances.js
```diff
@@ -170,6 +178,61 @@ const getAccountBalanceTimestamp = async (tsQuery, tsParams, order = 'desc') =>
...
+  const joinEntityClause = pubKeyQuery ? entityJoin : '';
+  const tokenBalanceSubQuery = getTokenBalanceDeduplicateSubQuery(order, accountQuery);
+  const whereClause = `
+      where ${[consensusTsQuery, accountQuery, pubKeyQuery, balanceQuery].filter(Boolean).join(' and ')}`;
+  const sqlQuery = `
+      select distinct on (account_id) ab.*, (${tokenBalanceSubQuery}) as token_balances
+      from account_balance ab
+      ${joinEntityClause}
+      ${whereClause}
+      order by ab.account_id ${order}, ab.consensus_timestamp desc
```
jnels124, line 231: this should use the order passed as a param

hedera-mirror-rest/balances.js
jascks: Hopefully one without 15 transitive dependencies of its own. 😄 It looks like the built in `Date` can be given a Unix epoch time, and be used to return the month and full year. It's only February that changes between 28 and 29 days, the rest are fixed at either 30 or 31. So, we could just define an array to look things up based on the month number, unless it's 1 (zero based February), in which case if the full year is divisible by 4 then use 29 else 28.

That would tied us over until 2100, which is not a leap year because it's divisible by 100. 😄

xin-hedera: yeah. there's more work if we want to query the partition metadata. fun fact is the partition boundary is stored in `pg_class.relpartbound` and in a (perhaps binary) format internal to postgres. `pg_get_expr(relpartbound, oid)` will resolve it into a human-readable string like `for values from ('xxx') to ('xxx')`

So to retrieve and cache partition metadata in the js rest module, we'll have to re-implement the `TimePartitionService` in importer...

jascks: Oh, that's too bad, but good to know. I was trying to find a utility function that would help with this. Something akin to:
```
mirror_node=> SELECT get_shard_id_for_distribution_column('transaction', 1833905);
 get_shard_id_for_distribution_column
--------------------------------------
                               102599
(1 row)
```

Given literal values for the partition key, Citus is able to limit the partitions accessed:

```
                                               ->  Merge Append  (cost=0.85..6778.96 rows=278 width=8) (actual time=1.295..14.744 rows=25 loops=1)
                                                     Sort Key: t_2.consensus_timestamp DESC
                                                     ->  Index Scan Backward using transaction_p2023_03_pkey_112792 on transaction_p2023_03_112792 t_2_1  (cost=0.42..5604.33 rows=275 width=8) (actual time=1.257..11.985 rows=22 loops=1)
                                                           Index Cond: ((consensus_timestamp >= '1678818548505133688'::bigint) AND (consensus_timestamp <= '1680908778141046768'::bigint) AND (payer_account_id = 1833905))
                                                           Filter: ((nonce = 0) OR (parent_consensus_timestamp IS NOT NULL))
                                                     ->  Index Scan Backward using transaction_p2023_04_pkey_112808 on transaction_p2023_04_112808 t_2_2  (cost=0.42..1171.83 rows=3 width=8) (actual time=0.038..2.751 rows=3 loops=1)
                                                           Index Cond: ((consensus_timestamp >= '1678818548505133688'::bigint) AND (consensus_timestamp <= '1680908778141046768'::bigint) AND (payer_account_id = 1833905))
                                                           Filter: ((nonce = 0) OR (parent_consensus_timestamp IS NOT NULL))
```
I'm not sure how those indexes got named (some other shard ID?), but that partition portion comports with the Citus `time_partitions` table:
```
       parent_table       |  partition_column   |             partition             |     from_value      |      to_value       | access_method
--------------------------+---------------------+-----------------------------------+---------------------+---------------------+---------------
 transaction              | consensus_timestamp | transaction_p2023_03              | 1677628800000000000 | 1680307200000000000 | heap
 transaction              | consensus_timestamp | transaction_p2023_04              | 1680307200000000000 | 1682899200000000000 | heap
```
Maybe we can leverage that in some manner?

jascks: One last thing before I switch to another PR; In `time_partitions` above, `from_value` and `to_value` are the actual boundary values for the partition, not the current lowest and highest values present.

xin-hedera: `time_partitions` is a view provided by citus so in v1 (vanilla postgresl) we have to craft the query and parse the partition boundary by ourselves.

edwin-greene: Thanks, I have updated this with a utility method to get the first millisecond of a month.

hedera-mirror-rest/balances.js
edwin-greene: Thanks, I have added a spec test for this in `deduplicate-balances.json`.

hedera-mirror-rest/balances.js
```diff
@@ -170,6 +173,42 @@ const getAccountBalanceTimestamp = async (tsQuery, tsParams, order = 'desc') =>
...
+      ${whereClause}
+      order by ab.account_id ${order}
+      ${limitQuery}`;
+};
+
+const getDeduplicateBalancesQuery = async (accountQuery, balanceQuery, limitQuery, order, pubKeyQuery, tsParams) => {
+  const upperBound = tsParams[0];
+  const lowerPartitionBound = utils.getFirstDayOfMonth(upperBound);
+  tsParams = [lowerPartitionBound, upperBound, lowerPartitionBound, upperBound];
+
```
edwin-greene, line 195: This now gets the partition bounds no matter if an accountQuery was received; tests for this have been added to `deduplcate-balances.json` which now has timestamps at the beginning and end of a month and also tests the balances with a timestamp and balances with accounts and timestamp.

hedera-mirror-rest/__tests__/specs/balances/deduplicated-balances.json
```diff
@@ -0,0 +1,212 @@
...
+  "description": "Balances api call with deduplicated account balances",
+  "setup": {
+    "config": {
+      "query": {
+        "deduplicateBalances": true
+      }
+    },
+    "balances": [
+      {
+        "timestamp": 1564617600000000000,
```
edwin-greene, line 11: This is the beginning of a month, August 1, 2019 12:00:00 AM

hedera-mirror-rest/__tests__/specs/balances/deduplicated-balances.json
```diff
@@ -0,0 +1,212 @@
...
+            "balance": 7
+          },
+          {
+            "token_num": 70007,
+            "balance": 700
+          }
+        ]
+      },
+      {
+        "timestamp": 1564618500000000000,
```
edwin-greene, line 32: Fifteen minutes later, August 1, 2019 12:15:00 AM

hedera-mirror-rest/__tests__/specs/balances/deduplicated-balances.json
```diff
@@ -0,0 +1,212 @@
...
+        "balance": 2
+      },
+      {
+        "timestamp": 1566618500000000000,
+        "id": 19,
+        "realm_num": 1,
+        "balance": 90
+      },
+      {
+        "timestamp": 1567295999000000000,
```
edwin-greene, line 66: The end of the month, August 31, 2019 11:59:59 PM

hedera-mirror-rest/__tests__/specs/balances/deduplicated-balances.json
```diff
@@ -0,0 +1,212 @@
...
+        "links": {
+          "next": null
+        }
+      }
+    },
+    {
+      "urls": [
+        "/api/v1/balances?account.id=gte:0.1.16&account.id=lt:0.1.21&timestamp=1567296000.000000000",
+        "/api/v1/balances?timestamp=1567296000.000000000"
+      ],
```
edwin-greene, line 201: The beginning of the next month, September 1, 2019 12:00:00 AM

hedera-mirror-rest/utils.js
```diff
@@ -912,6 +912,49 @@ const nsToSecNsWithHyphen = (ns) => {
...
+const getFirstDayOfMonth = (secNs) => {
+  if (_.isNil(secNs)) {
+    return null;
+  }
+
+  // Convert nanoseconds to milliseconds for Date object
+  const timestampMs = math.floor(Number(secNsToMs(secNs)));
+  const timestampDate = new Date(timestampMs);
+  const firstDayMs = Date.UTC(timestampDate.getUTCFullYear(), timestampDate.getUTCMonth());
+  return secMsToNs(firstDayMs);
```
edwin-greene, line 929: There is some loss of precision here with the conversion of nanoseconds to milliseconds, as to get a Date object in Javascript a timestamp in milliseconds is required.

hedera-mirror-rest/__tests__/specs/balances/deduplicated-balances.json
```diff
@@ -0,0 +1,212 @@
+{
+  "description": "Balances api call with deduplicated account balances",
```
jnels124, line 2: maybe should have a test case where user is providing both bounds?
`/api/v1/balances?account.id=gte:0.1.16&account.id=lt:0.1.21&timestamp=lt:1567295999.000000000&timestamp=gt:1564617600.000000000`

jnels124 (changes requested):

hedera-mirror-rest/balances.js
edwin-greene: Further investigation has found that this is still a problem concerning the gt/lt operators. Working on a fix.

hedera-mirror-rest/balances.js
```diff
@@ -170,6 +173,42 @@ const getAccountBalanceTimestamp = async (tsQuery, tsParams, order = 'desc') =>
...
+      select ab.*, (${tokenBalanceSubQuery}) as token_balances
+      from account_balance ab
+      ${joinEntityClause}
+      ${whereClause}
+      order by ab.account_id ${order}
+      ${limitQuery}`;
+};
+
+const getDeduplicateBalancesQuery = async (accountQuery, balanceQuery, limitQuery, order, pubKeyQuery, tsParams) => {
+  const upperBound = tsParams[0];
```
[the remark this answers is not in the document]
edwin-greene: Thanks I have added a method to get the partition bounds from the tsQuery.

hedera-mirror-rest/__tests__/specs/balances/deduplicated-balances.json
edwin-greene: Thanks! I have added this test and a fix so that it should now provide the correct results.

hedera-mirror-rest/balances.js
```diff
@@ -170,6 +176,84 @@ const getAccountBalanceTimestamp = async (tsQuery, tsParams, order = 'desc') =>
...
+
+const getDeduplicateTsQuery = (tsParams, tsQuery) => {
+  let upperBound = 0;
+  let gteParam = Number.MAX_VALUE;
+
+  // Combine the tsParams into a single upperBound and gteParam if present
+  tsQuery.split('?').forEach((value, index) => {
+    // eq operator has already been converted to the lte operator
+    if (value.includes(opsMap.lte)) {
+      upperBound = upperBound < tsParams[index] ? tsParams[index] : upperBound;
```
jnels124, line 227: shouldn't this logic be inversed? If we have two lte, shouldn't the upper bound be the minimum of the two?

edwin-greene: Sounds good, I have implemented the fix. The upperBound will now be the minimum of the lt/lte parameters and the lowerBound will be based off of the greatest of the gt/gte parameters.

hedera-mirror-rest/balances.js
```diff
@@ -170,6 +176,94 @@ const getAccountBalanceTimestamp = async (tsQuery, tsParams, order = 'desc') =>
...
+      ${joinEntityClause}
+      ${whereClause}
+      order by ab.account_id ${order}, ab.consensus_timestamp desc
+      ${limitQuery}`;
+  return [sqlQuery, deduplicateTsParams];
+};
+
+const getDeduplicateTsQuery = (tsParams, tsQuery) => {
+  let upperBound = Number.MAX_VALUE;
+  let gteParam = 0;
```
xin-hedera, line 221: both need to be `bigint`, timestamp in nanos precision isn't going to be a safe number

hedera-mirror-rest/balances.js
```diff
@@ -170,6 +176,94 @@ const getAccountBalanceTimestamp = async (tsQuery, tsParams, order = 'desc') =>
...
+  const neParams = [];
+
+  // Combine the tsParams into a single upperBound and gteParam if present
+  tsQuery.split('?').forEach((value, index) => {
+    // eq operator has already been converted to the lte operator
```
xin-hedera, line 226: should have a `const paramValue = BigInt(tsParams[index])` to avoid repeating `tsParams[index]`

can also rename the arrow function param name from `(value , index)` to `(query, index)` then can assign `tsParams[index]` to a variable `value` which is more straightfoward

hedera-mirror-rest/balances.js
```diff
@@ -170,6 +176,94 @@ const getAccountBalanceTimestamp = async (tsQuery, tsParams, order = 'desc') =>
...
+      upperBound = upperBound < ltValue ? upperBound : ltValue;
+    } else if (value.includes(opsMap.gte)) {
+      gteParam = gteParam > tsParams[index] ? gteParam : tsParams[index];
+    } else if (value.includes(opsMap.gt)) {
+      // Convert gt to gte to simplify query
+      const gtValue = math.add(math.bignumber(tsParams[index]), 1).toString();
+      gteParam = gteParam > gtValue ? gteParam : gtValue;
+    } else if (value.includes(opsMap.ne)) {
+      neParams.push(tsParams[index]);
+    }
```
xin-hedera, line 241: please convert all the math operations into `BigInt` (not `math.bignumber`)

hedera-mirror-rest/balances.js
```diff
@@ -170,6 +176,94 @@ const getAccountBalanceTimestamp = async (tsQuery, tsParams, order = 'desc') =>
...
+  });
+
+  const minTimestamp = gteParam !== 0 ? gteParam : upperBound;
```
xin-hedera, line 244: can't use `upperBound` as `minTimestamp` if not provided by user. I can send a query `/balances?timestamp=lte:year 3000`, the api will then return an empty list, because, there is simply no partition for year 3000 at year 2023.

hedera-mirror-rest/balances.js
```diff
@@ -170,6 +176,94 @@ const getAccountBalanceTimestamp = async (tsQuery, tsParams, order = 'desc') =>
...
+  return getDeduplicatePartitionedTsQuery(upperBound, lowerBound, neParams);
+};
+
+const getDeduplicatePartitionedTsQuery = (upperBound, lowerBound, neParams) => {
+  const neQuery = Array.from(neParams, (v) => '?').join(', ');
+  let consensusTsQuery = neQuery
+    ? `ab.consensus_timestamp >= ? and ab.consensus_timestamp not in (${neQuery})`
+    : `ab.consensus_timestamp >= ?`;
+
+  if (upperBound === Number.MAX_VALUE) {
```
xin-hedera, line 257: unsafe use of `Number`

hedera-mirror-rest/utils.js
```diff
@@ -912,6 +912,49 @@ const nsToSecNsWithHyphen = (ns) => {
...
+ * Converts milliseconds to nanoseconds
+ * @param {String|Number} secNs milliseconds since epoch
+ * @return {String} nanoseconds since epoch
+ */
+const secNsToMs = (secNs) => {
+  if (_.isNil(secNs)) {
+    return null;
+  }
+
+  return math.divide(math.bignumber(secNs), math.bignumber(1e6)).toString();
```
xin-hedera, line 942: prefer `BigInt` over math.bignumber functions

hedera-mirror-rest/utils.js
```diff
@@ -912,6 +912,49 @@ const nsToSecNsWithHyphen = (ns) => {
...
+};
+
+/**
+ * Converts nanoseconds to milliseconds
+ * @param {String} nanoseconds since epoch
+ * @return {String} milliseconds since epoch
+ */
+const secMsToNs = (secMs) => {
```
xin-hedera, line 950: `seMsToNs` and `secNsToMs` do not seem to really usable as their own functions

They are only used in `getFirstDayOfMonth` and directly multiple / divide in `getFirstDayOfMonth` will be much simpler

xin-hedera (changes requested):

hedera-mirror-rest/balances.js
xin-hedera: We should use current month as a reference timestamp for the range if no lower / upper bound is provided by the user, with consideration of the `order`.

- if order is `desc`
  - if upper bound's year and month is before the current month, use the begging of the month prior to the upper bound as the lower bound, i.e., suppose upper bound is in month X, the lower bound should be 00:00 of day one in month X - 1
  - otherwise, use the beginning of the month before current month as the lower bound. Reason is, the account balance partition for current month might be empty since the snapshots is generated by default every 15 minutes

- if order is `asc`, since we don't have running information of the first balance snapshot, I can't think of a way to optimize other than use the lower / upper bound provided by user as is

xin-hedera: using current month as reference is going to make it impossible to test certain query parameters in spec based integration test. so have to create a separate integration test without using spec files

docs/configuration.md
```diff
@@ -521,6 +521,7 @@ value, it is recommended to only populate overridden properties in the custom `a
 | `hedera.mirror.rest.openapi.specFileName`                          | 'openapi'               | The file name of the OpenAPI spec file                                                                                                                                                        |
 | `hedera.mirror.rest.openapi.swaggerUIPath`                         | '/docs'                 | Swagger UI path for your REST API                                                                                                                                                             |
 | `hedera.mirror.rest.port`                                          | 5551                    | The REST API port                                                                                                                                                                             |
+| `hedera.mirror.rest.query.deduplicateBalances`                     | false                   | Enables REST API queries for deduplicated balances                                                                                                                                            |
```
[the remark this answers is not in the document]
edwin-greene: Sounds good I have removed the property and the areas in balances.js that were not deduplicate queries.

hedera-mirror-rest/balances.js
edwin-greene: Thanks I have updated it to `BigInt`.

hedera-mirror-rest/balances.js
edwin-greene: Sounds good I have made these refactorings.

hedera-mirror-rest/balances.js
edwin-greene: Got it, I have updated everything to use `BigInt`.

hedera-mirror-rest/balances.js
edwin-greene: Thanks I have updated these sections and added a new test that is a non-spec test. The new test also has testcases for future timestamps and their expected behavior.

hedera-mirror-rest/balances.js
edwin-greene: Thanks I have fixed this.

hedera-mirror-rest/utils.js
edwin-greene: I have updated it to use `BigInt`.

hedera-mirror-rest/utils.js
edwin-greene: Sounds good I have removed those methods.

hedera-mirror-rest/balances.js
```diff
@@ -157,28 +150,111 @@ const getBalances = async (req, res) => {
...
+      const lowerBound = utils.getFirstDayOfMonth(gteParam);
+      return getPartitionedTsQuery(lowerBound, neParams, upperBound);
+    }
+  } else {
+    // There is only a lower bound, set the lower bound to the minimum of (gteParam, the first day of the month before the current month)
+    const firstDayOfCurrentMonth = utils.getFirstDayOfMonth(nowInNs);
+    const firstDayOfPreviousMonth = utils.getFirstDayOfMonth(firstDayOfCurrentMonth - 1n);
+    const lowerBound = gteParam > firstDayOfPreviousMonth ? firstDayOfPreviousMonth : gteParam;
+    return getPartitionedTsQuery(lowerBound, neParams);
+  }
```
xin-hedera, line 224: this is not what we discussed, as follows:

1. If both gt and lt are provided, we do nothing, use it as is
2. If there is only an upper bound, and the upper bound is in a month before the current month, we use `[beginning of the month upper bound is in, upper bound]`;  if upper bound is in the current or later month, we use `[beginning of the month before current month, upper bound]`
3. If there is only a lower bound, we set it to `max(lower bound, the beginning of the month before the current month)`

The major problem is lower bound is adjusted backwards. For instance, in the implementation, when there are both lower and upper bounds, `lowerBound` is set to the beginning of the month of `getParam`. Assume there is a query with `timestamp=gt:tomorrow&timestamp=lt:now`, obviously the effective timestamp range is empty so the api response should be empty. However the implementation will set `lowerBound` to `2023-11-01 00:00`, and return the last balance snapshot if there is one.

Similar issue in adjusting lowerBound when there is only lower bound from the http request. The proposed operation is `max`, however the implementation does `min`. Then the query with `timestamp=gt:tomorrow` will return balances from the latest snapshot, instead of an empty response.

Additional changes:

-  further shrink the timestamp range for case 1
- shortcut and return empty response if user provided lowerBound > upperBound

hedera-mirror-rest/utils.js
```diff
@@ -913,12 +913,21 @@ const nsToSecNsWithHyphen = (ns) => {
...
  */
-const secNsToNs = (secNs) => {
-  return math.multiply(math.bignumber(secNs), math.bignumber(1e9)).toString();
+const getFirstDayOfMonth = (secNs) => {
+  if (_.isNil(secNs)) {
+    return null;
+  }
+
+  // Convert nanoseconds to milliseconds for Date object
+  const timestampMs = secNs / 1000000n;
```
xin-hedera, line 926: make 10n ** 6n a constant

xin-hedera (changes requested):

hedera-mirror-rest/balances.js
```diff
@@ -157,28 +150,111 @@ const getBalances = async (req, res) => {
...
+};
+
+const getPartitionedTsQuery = (lowerBound, neParams, upperBound) => {
+  const neQuery = Array.from(neParams, (v) => '?').join(', ');
+  let consensusTsQuery = neQuery
+    ? `ab.consensus_timestamp >= ? and ab.consensus_timestamp not in (${neQuery})`
+    : `ab.consensus_timestamp >= ?`;
+
+  const lowerBoundString = lowerBound.toString();
```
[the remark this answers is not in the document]
edwin-greene: Thanks I have removed the strings.

hedera-mirror-rest/utils.js
edwin-greene: Got it, I have updated it to a constant.

hedera-mirror-rest/balances.js
```diff
@@ -21,11 +21,13 @@ import EntityId from './entityId';
 import {EntityService} from './service/index.js';
 import {EvmAddressType} from './constants';
 import {InvalidArgumentError} from './errors/index.js';
-import * as math from 'mathjs';
 import * as utils from './utils';
+import {opsMap} from './utils';
```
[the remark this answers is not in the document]
edwin-greene: Thanks! I have fixed that.

hedera-mirror-rest/balances.js
```diff
@@ -21,11 +21,13 @@ import EntityId from './entityId';
...
 
 const {tokenBalance: tokenBalanceLimit} = getResponseLimit();
 
+const MAX_UPPER_BOUND = BigInt(Number.MAX_VALUE);
```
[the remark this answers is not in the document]
edwin-greene: Sounds good I have moved it to constants.

hedera-mirror-rest/balances.js
edwin-greene: Thanks! I have updated this section and it should have those changes with the exception of the `further shrink the timestamp range for case 1`. I will look into that tomorrow.

hedera-mirror-rest/accounts.js
```diff
@@ -522,6 +521,23 @@ const getOneAccount = async (req, res) => {
...
+    select consensus_timestamp
+    from account_balance
+    where ${tsQuery}
+    order by consensus_timestamp ${order}
+    limit 1`;
+
+  const pgSqlQuery = utils.convertMySqlStyleQueryToPostgres(query);
+  const {rows} = await pool.queryQuietly(pgSqlQuery, tsParams);
+  return rows[0]?.consensus_timestamp;
+};
```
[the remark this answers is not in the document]
jnels124: I have reached out to Nana to discuss.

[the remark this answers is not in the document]
jnels124: spoke with nana and he is ok with the balance always being the latest within the specified range regardless of order. His only request was to make sure this was documented in the api docs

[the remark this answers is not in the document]
xin-hedera: thanks. let's switch back to use `balances.getAccountBalanceTimestamp`

hedera-mirror-rest/balances.js
```diff
@@ -157,28 +146,108 @@ const getBalances = async (req, res) => {
...
+  const neParams = [];
+
+  // Combine the tsParams into a single upperBound and gteParam if present
+  tsQuery
+    .split('?')
+    .filter(Boolean)
+    .forEach((query, index) => {
+      const value = BigInt(tsParams[index]);
+      // eq operator has already been converted to the lte operator
+      if (query.includes(utils.opsMap.lte)) {
```
xin-hedera, line 178: add a comment to explain why check `lte` before `lt`. same for `gte` and `gt`

hedera-mirror-rest/balances.js
```diff
@@ -157,28 +146,108 @@ const getBalances = async (req, res) => {
...
+        // Convert gt to gte to simplify query
+        const gtValue = value + 1n;
+        gteParam = gteParam > gtValue ? gteParam : gtValue;
+      } else if (query.includes(utils.opsMap.ne)) {
+        neParams.push(value);
+      }
+    });
+
+  if (gteParam > upperBound) {
+    throw new InvalidArgumentError('Invalid timestamp ranges: lt value is less than gt value');
```
xin-hedera, line 196: wouldn't this break the endpoint's behavior? This is what's returned now

```
$ curl -s 'https://previewnet.mirrornode.hedera.com/api/v1/balances?timestamp=gt:1700183696&timestamp=lt:1700180000' | jq .
{
  "timestamp": null,
  "balances": [],
  "links": {
    "next": null
  }
}
```

hedera-mirror-rest/balances.js
```diff
@@ -157,28 +146,108 @@ const getBalances = async (req, res) => {
...
+
+  const nowInNs = BigInt(Date.now()) * constants.NANOSECONDS_PER_MILLISECOND;
+  if (upperBound !== constants.MAX_LONG) {
+    if (gteParam === 0n) {
+      // There is an upper bound but no lowerBound
+      const firstDayOfUpperBoundMonth = utils.getFirstDayOfMonth(upperBound);
+      const firstDayOfCurrentMonth = utils.getFirstDayOfMonth(nowInNs);
+      const lowerBound =
+        firstDayOfUpperBoundMonth < firstDayOfCurrentMonth
+          ? // If the upper bound is in a month prior to the current month, the lower bound is the beginning of the month the upper bound is in
```
xin-hedera, line 207: Let's change to: if the upper bound is in a month prior to the current month, the lower bound is the beginning of the month prior to the month the upper bound is in.

It's the same corner case than for instance when the upper bound is merely 1 minute into a month, there's most likely no balance snapshot in that 1 minute range, however, there's no to little chance with no balance snapshot in that 1 month + 1 minute range.

hedera-mirror-rest/balances.js
```diff
@@ -157,28 +146,108 @@ const getBalances = async (req, res) => {
...
+            firstDayOfUpperBoundMonth
+          : // If upper bound is in the current or later month the lower bound is the beginning of the month before current month
+            utils.getFirstDayOfMonth(firstDayOfCurrentMonth - 1n);
+      return getPartitionedTsQuery(lowerBound, neParams, upperBound);
```
xin-hedera, line 211: nit, `getPartitionedTsQuery` is called three times. we can leave the `if ... else ...` block to figure out the upper bound and lower bound, and call `getPartitionedTsQuery` once at the end.

for `upperBound`, if it's not set, it's long max, so it's perfectly fine to just pass it to `getPartitionedTsQuery`, which means `getPartitionedTsQuery` no longer to check if `upperBound` set or not.

xin-hedera (changes requested):

hedera-mirror-rest/balances.js
```diff
@@ -157,28 +146,108 @@ const getBalances = async (req, res) => {
...
-    limit 1`;
-
-  const pgSqlQuery = utils.convertMySqlStyleQueryToPostgres(query);
-  const {rows} = await pool.queryQuietly(pgSqlQuery, tsParams);
-  return rows[0]?.consensus_timestamp;
+const getBalancesQuery = (accountQuery, balanceQuery, limitQuery, order, pubKeyQuery, tsParams, tsQuery) => {
+  const [consensusTsQuery, deduplicateTsParams] = getTsQuery(tsParams, tsQuery);
+  // Only need to join entity if we're selecting on publickey
+  const joinEntityClause = pubKeyQuery ? entityJoin : '';
+  const tokenBalanceSubQuery = getTokenBalanceSubQuery(order, consensusTsQuery);
```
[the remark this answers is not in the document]
xin-hedera: since we need to change it back to the old 2-step flow, given `T1` is the balance snapshot timestamp found in the first step, for the second step, the timestamp query should be `consensus_timestamp >= T2 and consensus_timestamp <= T1` where `T2 = beginning of the month T1 is in`

hedera-mirror-rest/accounts.js
```diff
@@ -409,7 +408,7 @@ const getOneAccount = async (req, res) => {
       eqValues,
       false
     );
-    const balanceFileTs = await balances.getAccountBalanceTimestamp(
```
xin-hedera, line 412: both account balance query and token balance query for `/accounts/{id}?timestamp=` are not updated to use the deduped balance table. the queries still get balance info from the table at exactly the balance snapshot timestamp, instead of the latest balance info in`[beginning of partition, balance snapshot timestamp]`

xin-hedera (changes requested):

hedera-mirror-rest/balances.js
```diff
@@ -170,15 +169,117 @@ const getAccountBalanceTimestamp = async (tsQuery, tsParams, order = 'desc') =>
...
+    lowerBound = gteParam > firstDayOfPreviousMonth ? gteParam : firstDayOfPreviousMonth;
+  }
+
+  return getPartitionedTsQuery(lowerBound, neParams, upperBound);
+};
+
+const getPartitionedTsQuery = async (lowerBound, neParams, upperBound) => {
+  const neQuery = Array.from(neParams, (v) => '?').join(', ');
+  let accountBalanceQuery = `consensus_timestamp >= ? and consensus_timestamp <= ?`;
+  const accountBalanceTimestamp = await getAccountBalanceTimestamp(accountBalanceQuery, [lowerBound, upperBound]);
```
xin-hedera, line 252: `neQuery` needs to go into `getAccountBalanceTimestamp` instead of being added to `consensusTsQuery` and then passed to the main query.

Two reasons

- avoid the picked treasury account's balance snapshot timestamp matching any of the `ne` values
- suppose the picked treasury account's balance snapshot timestamp is `T1`, and ne value is `T2` (which is earlier than `T1`), if for an account `X`, the latest balance at or before `T1` is in snapshot at time `T2`, we *should* use it, instead of otherwise when the ne value `T2` is in the main query, account `X`'s balance at an even earlier timestamp `T3` is picked

hedera-mirror-rest/balances.js
```diff
@@ -170,15 +169,117 @@ const getAccountBalanceTimestamp = async (tsQuery, tsParams, order = 'desc') =>
...
+  if (!accountBalanceTimestamp) {
+    return [undefined, []];
+  }
+
+  const beginningOfMonth = utils.getFirstDayOfMonth(accountBalanceTimestamp);
+  const consensusTsQuery = neQuery
+    ? `ab.consensus_timestamp >= ? and ab.consensus_timestamp <= ? and ab.consensus_timestamp not in (${neQuery})`
+    : `ab.consensus_timestamp >= ? and ab.consensus_timestamp <= ?`;
+  // Double the params to account for the query values for account_balance and token_balance together
```
xin-hedera, line 261: remove the logic about `neQuery` here

xin-hedera (changes requested):

hedera-mirror-rest/balances.js
edwin-greene: Thanks I have updated this.

hedera-mirror-rest/balances.js
edwin-greene: Sounds good, I have removed it.

hedera-mirror-rest/balances.js
edwin-greene: Thanks I have updated the comment

xin-hedera (approved):
LGTM

jnels124 (approved):
LGTM

PR #7090, opened 2023-10-18
HIP-584: Historical support - 6970 map block number to timestamp

### LINKED ISSUE #6970: HIP-584: Map block value from request body to timestamp and set it in the ThreadLocalHolder timestamp.
### Problem

In order to query historical data we need a timestamp filter to be used in our DB queries.

We already have a timestamp field added in ThreadLocalHolder class with a default value of -1.

We need to map the block value passed in request body to a timestamp.
Using the block value in request (which can be "earliest" or 
hexadecimal number of the block) we need to query record_file table with that block.

The query result will have a consensus_start and consensus_end timestamp. 
We need to get the consensus_end timestamp and set in our ThreadLocal timestamp.
We will later use this timestamp for filtering in our DB accessors.

### Solution

Map block value to a timestamp and set it the ThreadLocalHolder timestamp.
If value of block "latest", maintain current behavior and do not do the mapping.

### Alternatives

_No response_

kselveliev:
**Description**:
This PR fixes the following two issues: 

#6895 - Since the timestamp field was deleted at some point in [PR](https://github.com/hashgraph/hedera-mirror-node/pull/6676) - the same change is introduced here

#6970 - When a block number is passed in contracts/call request body for eth_call, we need to map the block passed
to a certain timestamp which we will be using to later query historical data in db.

Modifications in this **PR**

Added new timestamp field in **ContractCallContext** which will be used for filtering.
Added new method in **RecordFileRepository** that will query to find consensus_end timestamp for record file based on block number 

Added a logic in MirrorEvmTxProcessorImpl that checks if we are in eth_call context. Then gets the block passed in request and finds the record file with that number and gets its consensus_end timestamp and sets in the **ContractCallContext**( later on this value will be read and used for filtering in db queries).
 If no value is found for that block number the timestamp field will be left with default value and this will be treated as if **BlockType.LATEST** is passed.

Added unit tests for ContractCallContext -> **ContractCallContextTest**

Modified **MirrorEvmTxProcessorTest** where we pass different combinations of is estimate gas or eth call with different block number values 

Modified **RecordFileRepositoryTest** - added test for the new query 

Modified **ContractCallServiceTest** - added a test for eth call where custom block number is passed

Modified shared method used to build service parameters for **processCall** - now we can pass **block** parameter to it.

**Related issue(s)**:

Fixes #6970, #6895 

**Checklist**

### DIFF
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/ContractController.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/ContractController.java
@@ -89,6 +89,7 @@ class ContractController {
         final var data = request.getData() != null ? Bytes.fromHexString(request.getData()) : EMPTY;
         final var isStaticCall = false;
         final var callType = request.isEstimate() ? ETH_ESTIMATE_GAS : ETH_CALL;
+        final var block = request.getBlock();
 
         return CallServiceParameters.builder()
                 .sender(sender)
@@ -99,6 +100,7 @@ class ContractController {
                 .isStatic(isStaticCall)
                 .callType(callType)
                 .isEstimate(request.isEstimate())
+                .block(block)
                 .build();
     }
 
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/config/EvmConfiguration.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/config/EvmConfiguration.java
@@ -29,6 +29,7 @@ import com.hedera.mirror.web3.evm.properties.StaticBlockMetaSource;
 import com.hedera.mirror.web3.evm.store.Store;
 import com.hedera.mirror.web3.evm.store.contract.EntityAddressSequencer;
 import com.hedera.mirror.web3.evm.store.contract.HederaEvmWorldState;
+import com.hedera.mirror.web3.repository.RecordFileRepository;
 import com.hedera.mirror.web3.repository.properties.CacheProperties;
 import com.hedera.node.app.service.evm.store.contracts.AbstractCodeCache;
 import com.hedera.services.contracts.execution.LivePricesSource;
@@ -153,7 +154,8 @@ public class EvmConfiguration {
             final BasicHbarCentExchange basicHbarCentExchange,
             final PrngSystemPrecompiledContract prngSystemPrecompiledContract,
             final HederaPrngSeedOperation prngSeedOperation,
-            final Store store) {
+            final Store store,
+            final RecordFileRepository recordFileRepository) {
         return new MirrorEvmTxProcessorImpl(
                 worldState,
                 pricesAndFees,
@@ -174,6 +176,7 @@ public class EvmConfiguration {
                 mirrorEvmContractAliases,
                 abstractCodeCache,
                 mirrorOperationTracer,
-                store);
+                store,
+                recordFileRepository);
     }
 }
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmTxProcessorImpl.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmTxProcessorImpl.java
@@ -21,6 +21,8 @@ import com.hedera.mirror.web3.evm.account.MirrorEvmContractAliases;
 import com.hedera.mirror.web3.evm.contracts.execution.traceability.MirrorOperationTracer;
 import com.hedera.mirror.web3.evm.store.Store;
 import com.hedera.mirror.web3.exception.MirrorEvmTransactionException;
+import com.hedera.mirror.web3.repository.RecordFileRepository;
+import com.hedera.mirror.web3.viewmodel.BlockType;
 import com.hedera.node.app.service.evm.contracts.execution.BlockMetaSource;
 import com.hedera.node.app.service.evm.contracts.execution.EvmProperties;
 import com.hedera.node.app.service.evm.contracts.execution.HederaEvmTransactionProcessingResult;
@@ -34,6 +36,7 @@ import com.hederahashgraph.api.proto.java.HederaFunctionality;
 import com.hederahashgraph.api.proto.java.ResponseCodeEnum;
 import java.time.Instant;
 import java.util.Map;
+import java.util.Optional;
 import javax.inject.Provider;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.tuweni.bytes.Bytes;
@@ -49,6 +52,7 @@ public class MirrorEvmTxProcessorImpl extends HederaEvmTxProcessor implements Mi
     private final AbstractCodeCache codeCache;
     private final MirrorEvmContractAliases aliasManager;
     private final Store store;
+    private final RecordFileRepository recordFileRepository;
 
     @SuppressWarnings("java:S107")
     public MirrorEvmTxProcessorImpl(
@@ -62,13 +66,15 @@ public class MirrorEvmTxProcessorImpl extends HederaEvmTxProcessor implements Mi
             final MirrorEvmContractAliases aliasManager,
             final AbstractCodeCache codeCache,
             final MirrorOperationTracer operationTracer,
-            final Store store) {
+            final Store store,
+            final RecordFileRepository recordFileRepository) {
         super(worldState, pricesAndFeesProvider, dynamicProperties, gasCalculator, mcps, ccps, blockMetaSource);
 
         super.setOperationTracer(operationTracer);
         this.aliasManager = aliasManager;
         this.codeCache = codeCache;
         this.store = store;
+        this.recordFileRepository = recordFileRepository;
     }
 
     public HederaEvmTransactionProcessingResult execute(
@@ -79,13 +85,19 @@ public class MirrorEvmTxProcessorImpl extends HederaEvmTxProcessor implements Mi
             final Bytes callData,
             final Instant consensusTime,
             final boolean isStatic,
-            final boolean isEstimate) {
+            final boolean isEstimate,
+            final BlockType block) {
         final long gasPrice = gasPriceTinyBarsGiven(consensusTime, true);
         // in cases where the receiver is the zero address, we know it's a contract create scenario
         super.setupFields(receiver.equals(Address.ZERO));
         final var contractCallContext = ContractCallContext.get();
         contractCallContext.setCreate(Address.ZERO.equals(receiver));
         contractCallContext.setEstimate(isEstimate);
+
+        if (!isEstimate) {
+            Optional<Long> blockTimestamp = recordFileRepository.findConsensusEndByBlockNumber(block.number());
+            blockTimestamp.ifPresent(contractCallContext::setBlockTimestamp);
+        }
 
         store.wrap();
         if (isEstimate) {
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/RecordFileRepository.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/RecordFileRepository.java
@@ -47,4 +47,7 @@ public interface RecordFileRepository extends PagingAndSortingRepository<RecordF
             unless = "#result == null")
     @Query(value = "select * from record_file order by consensus_end desc limit 1", nativeQuery = true)
     Optional<RecordFile> findLatest();
+
+    @Query(value = "select consensus_end from record_file where index = ?1", nativeQuery = true)
+    Optional<Long> findConsensusEndByBlockNumber(long blockNumber);
 }
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
@@ -119,7 +119,8 @@ public class ContractCallService {
                     params.getCallData(),
                     Instant.now(),
                     params.isStatic(),
-                    isEstimate);
+                    isEstimate,
+                    params.getBlock());
         } catch (IllegalStateException | IllegalArgumentException e) {
             throw new MirrorEvmTransactionException(e.getMessage(), EMPTY, EMPTY);
         }
--- a/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/repository/RecordFileRepositoryTest.java
+++ b/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/repository/RecordFileRepositoryTest.java
@@ -52,4 +52,22 @@ class RecordFileRepositoryTest extends Web3IntegrationTest {
 
         assertThat(recordFileRepository.findLatest()).get().isEqualTo(latest);
     }
+
+    @Test
+    void findConsensusEndByBlockNumber() {
+        domainBuilder.recordFile().persist();
+        var latest = domainBuilder.recordFile().persist();
+        long blockNumber = latest.getIndex();
+
+        assertThat(recordFileRepository.findConsensusEndByBlockNumber(blockNumber))
+                .get()
+                .isEqualTo(latest.getConsensusEnd());
+    }
+
+    @Test
+    void findConsensusEndByBlockNumberNotExists() {
+        long nonExistentBlockNumber = 1L;
+        assertThat(recordFileRepository.findConsensusEndByBlockNumber(nonExistentBlockNumber))
+                .isEmpty();
+    }
 }
--- a/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServiceTest.java
+++ b/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServiceTest.java
@@ -37,6 +37,7 @@ import org.junit.jupiter.params.provider.EnumSource;
 
 class ContractCallServiceTest extends ContractCallTestSetup {
 
+    // add tests in this class
     private static final String GAS_METRICS = "hedera.mirror.web3.call.gas";
 
     @BeforeEach
--- a/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServiceTest.java
+++ b/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServiceTest.java
@@ -26,6 +26,7 @@ import static org.assertj.core.api.AssertionsForClassTypes.assertThatThrownBy;
 
 import com.hedera.mirror.web3.exception.MirrorEvmTransactionException;
 import com.hedera.mirror.web3.service.model.CallServiceParameters.CallType;
+import com.hedera.mirror.web3.viewmodel.BlockType;
 import lombok.RequiredArgsConstructor;
 import org.apache.tuweni.bytes.Bytes;
 import org.assertj.core.data.Percentage;
@@ -37,7 +38,6 @@ import org.junit.jupiter.params.provider.EnumSource;
 
 class ContractCallServiceTest extends ContractCallTestSetup {
 
-    // add tests in this class
     private static final String GAS_METRICS = "hedera.mirror.web3.call.gas";
 
     @BeforeEach
@@ -54,7 +54,27 @@ class ContractCallServiceTest extends ContractCallTestSetup {
         final var pureFuncHash = "8070450f";
         final var successfulReadResponse = "0x0000000000000000000000000000000000000000000000000000000000000004";
         final var serviceParameters = serviceParametersForExecution(
-                Bytes.fromHexString(pureFuncHash), ETH_CALL_CONTRACT_ADDRESS, ETH_CALL, 0L);
+                Bytes.fromHexString(pureFuncHash), ETH_CALL_CONTRACT_ADDRESS, ETH_CALL, 0L, BlockType.LATEST);
+
+        assertThat(contractCallService.processCall(serviceParameters)).isEqualTo(successfulReadResponse);
+
+        assertGasUsedIsPositive(gasUsedBeforeExecution, ETH_CALL);
+    }
+
+    @Test
+    void pureCallWithCustomBlock() {
+        domainBuilder
+                .recordFile()
+                .customize(recordFileBuilder -> recordFileBuilder.index(1L))
+                .persist();
+
+        final var gasUsedBeforeExecution = getGasUsedBeforeExecution(ETH_CALL);
+
+        final var pureFuncHash = "8070450f";
+        final var successfulReadResponse = "0x0000000000000000000000000000000000000000000000000000000000000004";
+
+        final var serviceParameters = serviceParametersForExecution(
+                Bytes.fromHexString(pureFuncHash), ETH_CALL_CONTRACT_ADDRESS, ETH_CALL, 0L, BlockType.of("0x1"));
 
         assertThat(contractCallService.processCall(serviceParameters)).isEqualTo(successfulReadResponse);
 
@@ -66,7 +86,7 @@ class ContractCallServiceTest extends ContractCallTestSetup {
         final var pureFuncHash = "8070450f";
         final var gasUsedBeforeExecution = getGasUsedBeforeExecution(ETH_ESTIMATE_GAS);
         final var serviceParameters = serviceParametersForExecution(
-                Bytes.fromHexString(pureFuncHash), ETH_CALL_CONTRACT_ADDRESS, ETH_ESTIMATE_GAS, 0L);
+                Bytes.fromHexString(pureFuncHash), ETH_CALL_CONTRACT_ADDRESS, ETH_ESTIMATE_GAS, 0L, BlockType.LATEST);
 
         final var expectedGasUsed = gasUsedAfterExecution(serviceParameters);
 
@@ -79,8 +99,8 @@ class ContractCallServiceTest extends ContractCallTestSetup {
 
     @Test
     void estimateGasWithoutReceiver() {
-        final var serviceParameters =
-                serviceParametersForExecution(Bytes.fromHexString("0x"), Address.ZERO, ETH_ESTIMATE_GAS, 0L);
+        final var serviceParameters = serviceParametersForExecution(
+                Bytes.fromHexString("0x"), Address.ZERO, ETH_ESTIMATE_GAS, 0L, BlockType.LATEST);
 
         final var expectedGasUsed = gasUsedAfterExecution(serviceParameters);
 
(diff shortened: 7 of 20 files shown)


0xivanov (commented):
Suggestion:
See where `RecordFileRepository.findLatest` is used and call `findConsensusEndByBlockNumber` instead of it _if it is required_ .

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/RecordFileRepository.java
```diff
@@ -47,4 +47,7 @@ public interface RecordFileRepository extends PagingAndSortingRepository<RecordF
             unless = "#result == null")
     @Query(value = "select * from record_file order by consensus_end desc limit 1", nativeQuery = true)
     Optional<RecordFile> findLatest();
+
+    @Query(value = "select consensus_end from record_file where index = ?1", nativeQuery = true)
+    Optional<Long> findConsensusEndByBlockNumber(long blockNumber);
```
steven-sheehy, line 52: There's already a `findHashByIndex` that's cached. Instead of adding another finder method every time we need a different field, we should convert the existing one to return a `RecordFile` so we can centralize the query for caching purposes.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
```diff
@@ -119,7 +119,8 @@ private HederaEvmTransactionProcessingResult doProcessCall(
                     params.getCallData(),
                     Instant.now(),
                     params.isStatic(),
-                    isEstimate);
+                    isEstimate,
+                    params.getBlock());
```
steven-sheehy, line 123: This is outside the scope of this change, but I never liked this method signature and this change makes it obvious. I never understood why we take in a `ContractCallRequest` just to convert into an almost identical `CallServiceParameters`, to then deconstruct its fields and pass as explicit parameters here. We should at minimum just pass `CallServiceParameters` to `MirrorEvmTxProcessor` and possibly just pass `ContractCallRequest`. That way when you change the params you don't have to adjust a bunch of interfaces and it's encapsulated in one object.

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/RecordFileRepository.java
kselveliev: Commend addressed. I deleted the new method and changed findHashByIndex to findRecordFileByIndex. I adjusted the places where this method is used accordingly.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmTxProcessorImpl.java
```diff
@@ -79,14 +85,20 @@ public HederaEvmTransactionProcessingResult execute(
...
+            final BlockType block) {
         final long gasPrice = gasPriceTinyBarsGiven(consensusTime, true);
         // in cases where the receiver is the zero address, we know it's a contract create scenario
         super.setupFields(receiver.equals(Address.ZERO));
         final var contractCallContext = ContractCallContext.get();
         contractCallContext.setCreate(Address.ZERO.equals(receiver));
         contractCallContext.setEstimate(isEstimate);
 
+        if (!isEstimate) {
+            Optional<Long> blockTimestamp = recordFileRepository.findConsensusEndByBlockNumber(block.number());
```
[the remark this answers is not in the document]
kselveliev: Comment addressed. I added a check to skip db check when we are in BlockType.LATEST.

For testing coverage, there is one test in MirrorEvmTxProcessorTest which passes different block values. However, the test’s scope is limited and the DB result will not be utilized. 

Proper testing will be possible once we add the future changes where the db result timestamp will be used. Then we can do integration and acceptance test which will cover such cases that you mentioned properly.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
kselveliev: I agree with this. 

I tried to see if it would be easy just to pass CallServiceParameters instead of the current version. 
In the estimate gas we have one corner case: 

image
image

When we do process call we have a check **params.isEstimate() ? estimatedGas : params.getGas(),**
then based on the flag we pass different value in **mirrorEvmTxProcessor.execute(.** **for gas value**
This means that if we pass the whole CallServiceParameters object to the method we will either have to mutate the object's gas field (**currently the object is aimed to be immmutable)** or pass one more parameter for estimatedGas to the execute() method. 

Since this change will require some refactoring and also a lot of test classes will be changed.
I prefer to create a new task that will specifically target this because I do not want to add more complexity to the changes that are aimed for this PR.

Let me know if you agree with this. (I can also try to make the change in this PR)

steven-sheehy: Agree it should be separate task and gas can be a separate method parameter.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmTxProcessorImpl.java
```diff
@@ -139,4 +151,9 @@ protected MessageFrame buildInitialFrame(
                         .code(code)
                         .build();
     }
+
+    public void setBlockTimestampToContractCallContext(BlockType block, ContractCallContext contractCallContext) {
+        Optional<RecordFile> recordFile = recordFileRepository.findRecordFileByIndex(block.number());
```
steven-sheehy, line 156: There's a corner case for partial mirror nodes that don't have data in their database since they set startDate=some date. For them, EARLIEST block can vary from other mirror nodes so we should do a separate lookup to find min(index) so it finds the min block the node has available.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/repository/RecordFileRepositoryTest.java
```diff
@@ -52,4 +55,22 @@ void findLatestFile() {
...
+        var latest = domainBuilder.recordFile().persist();
+        long blockNumber = latest.getIndex();
+
+        assertThat(recordFileRepository.findRecordFileByIndex(blockNumber))
+                .map(RecordFile::getConsensusEnd)
+                .hasValue(latest.getConsensusEnd());
+    }
+
+    @Test
+    void findConsensusEndByBlockNumberNotExists() {
```
steven-sheehy, line 71: Change method names to match corresponding repository method.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServiceTest.java
```diff
@@ -53,7 +54,27 @@ void pureCall() {
...
-                Bytes.fromHexString(pureFuncHash), ETH_CALL_CONTRACT_ADDRESS, ETH_CALL, 0L);
+                Bytes.fromHexString(pureFuncHash), ETH_CALL_CONTRACT_ADDRESS, ETH_CALL, 0L, BlockType.LATEST);
+
+        assertThat(contractCallService.processCall(serviceParameters)).isEqualTo(successfulReadResponse);
+
+        assertGasUsedIsPositive(gasUsedBeforeExecution, ETH_CALL);
+    }
+
+    @Test
+    void pureCallWithCustomBlock() {
```
steven-sheehy, line 65: We should have one test here for EARLIEST to have integration test coverage with db.

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmTxProcessorImpl.java
IvanKavaldzhiev: Shouldn't we enable the web3 module only on full mirror nodes? By definition an archive node in Ethereum stores all data from genesis and that's why it's called an archival node. I believe in HIP-584 we defined that we would support these web3 REST calls only on full mirror nodes with enabled sidecars ingestion.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/repository/RecordFileRepositoryTest.java
kselveliev: Comment addressed

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServiceTest.java
kselveliev: Comment addressed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmTxProcessorImpl.java
steven-sheehy: I think we should support a best effort basis for partial mirror node like we do elsewhere. Nana agrees with my above comment saying that what earliest returns is based on the data the node has available.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServiceTest.java
```diff
@@ -452,4 +492,8 @@ private enum RevertFunctions {
         private final String errorDetail;
         private final String errorData;
     }
+
+    static Stream<BlockType> provideBlockTypes() {
+        return Stream.of(BlockType.EARLIEST, BlockType.of("0x1"), BlockType.of("0x100"));
```
IvanKavaldzhiev, line 497: We can also add tests for `PENDING`, `SAFE`, `FINALIZED` and block value of `Long.MAX` to see how the system behaves with non-existing block.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmTxProcessorImpl.java
kselveliev: Comment addressed. I added a logic to determine the earliest possible block when a user passes a block that is earlier than the earliest possible in the DB.

Let me know if the changes look good.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/config/EvmConfiguration.java
```diff
@@ -137,6 +139,18 @@ CacheManager cacheManagerRecordFileLatest() {
         return caffeineCacheManager;
     }
 
+    @Bean(CACHE_MANAGER_RECORD_FILE_EARLIEST)
+    CacheManager cacheManagerRecordFileEarliest() {
+        final var caffeine = Caffeine.newBuilder()
+                .expireAfterWrite(10, TimeUnit.MINUTES)
+                .maximumSize(1000)
+                .recordStats();
```
IvanKavaldzhiev, line 147: This is a value that can never change. This cached values should never expire. The maximum size could be 1, since we would keep only a single value and reduce the cache size.

kselveliev: Comment addressed

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServiceTest.java
kselveliev: Comment addressed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmTxProcessorImpl.java
```diff
@@ -79,14 +86,20 @@ public HederaEvmTransactionProcessingResult execute(
...
         // in cases where the receiver is the zero address, we know it's a contract create scenario
         super.setupFields(receiver.equals(Address.ZERO));
         final var contractCallContext = ContractCallContext.get();
         contractCallContext.setCreate(Address.ZERO.equals(receiver));
         contractCallContext.setEstimate(isEstimate);
 
+        if (shouldSetBlockTimestamp(isEstimate, block)) {
+            long blockNumber = determineAppropriateBlockNumber(block);
+            setBlockTimestampToContractCallContext(blockNumber, contractCallContext);
+        }
```
steven-sheehy, line 101: It's wasteful to query for min index if block type is not `EARLIEST`. We should just have this somewhere:

```java
        return recordFileRepository.findMinimumIndex().orElseThrow();
    }
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmTxProcessorImpl.java
```diff
@@ -139,4 +152,18 @@ protected MessageFrame buildInitialFrame(
...
+    }
+
+    private long determineAppropriateBlockNumber(BlockType block) {
+        long minIndex = recordFileRepository.findMinimumIndex().orElse(BlockType.EARLIEST.number());
+        return Math.max(block.number(), minIndex);
+    }
+
+    public void setBlockTimestampToContractCallContext(long blockIndex, ContractCallContext contractCallContext) {
+        Optional<RecordFile> recordFile = recordFileRepository.findRecordFileByIndex(blockIndex);
+        recordFile.map(RecordFile::getConsensusEnd).ifPresent(contractCallContext::setBlockTimestamp);
```
steven-sheehy, line 167: If we can't find the given block number, shouldn't we throw or short circuit the logic and return some default value? What do ethereum nodes do?

Here, if we can't find we're just silently assuming they want `BlockType.LATEST`, which could provide inaccurate values. If they provide a block number that's greater than latest, it might make sense we assume latest but probably not if less than latest.

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmTxProcessorImpl.java
IvanKavaldzhiev: I believe Archive Ethereum nodes use 0 as earliest block and Full Nodes use the 128th block before the current one. If an Ethereum node does not find a result for the specified block - it returns and empty output. This is what we can also do - depending on the mirror node (archive or partial) - we should find the earliest block we have in the DB and search for information based on it. If we don't find anything - just return empty result and don't fallback to latest or another block value. 

I believe that if we pass a block number greater than the latest block number - again empty value is returned, but I can double check that.

steven-sheehy: I'm okay with return default value but the main point though is if a block is not found or out of range (negative, greater than latest, earlier than min available), we should not proceed to execute the EVM with potentially inaccurate block info. Currently the code proceeds as normal.

IvanKavaldzhiev: I agree. We should add conditional logic.

If block < min -> throw error with Unknown block number
If earliest -> use min available
If min < block < current -> execute transaction (normal case)
If block > max -> throw error with Unknown block number

What do you think?

steven-sheehy: > If block < min -> throw error with Unknown block number

It's unnecessary since if it's less than min it won't be found. So basically just change this to if block not found throw error. Unless you're saying if min=100 and block=99 then we should not throw error but return default value? However, if block < 0 we should not do any lookup and just throw unknown block number.

> If block > max -> throw error with Unknown block number

Same as above.

> If min < block < current -> execute transaction (normal case)

Need to be inclusive of values `min <= block <= current`

Also probably obvious but need if `latest -> use max available`

IvanKavaldzhiev: I believe if block < 0, then we should return BAD_REQUEST, since the user has sent invalid block value.

For the case of `min = 100` and `block = 99`, we should return default empty value, since this is a valid positive block number less than `max block` and it is the partial mirror node that does not support this block index.

steven-sheehy: Ok, that sounds fine.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/config/EvmConfiguration.java
```diff
@@ -136,6 +139,17 @@ CacheManager cacheManagerRecordFileLatest() {
         return caffeineCacheManager;
     }
 
+    @Bean(CACHE_MANAGER_RECORD_FILE_EARLIEST)
+    CacheManager cacheManagerRecordFileEarliest() {
+        final var caffeine = Caffeine.newBuilder()
+                .maximumSize(1)
+                .recordStats();
+        final CaffeineCacheManager caffeineCacheManager = new CaffeineCacheManager();
+        caffeineCacheManager.setCacheNames(Set.of(CACHE_NAME_RECORD_FILE_EARLIEST_INDEX));
```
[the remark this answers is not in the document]
kselveliev: Comment addressed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmTxProcessorImpl.java
```diff
@@ -139,4 +152,18 @@ protected MessageFrame buildInitialFrame(
...
+    private boolean shouldSetBlockTimestamp(boolean isEstimate, BlockType block) {
+        return !isEstimate && block != BlockType.LATEST;
+    }
+
+    private long determineAppropriateBlockNumber(BlockType block) {
+        long minIndex = recordFileRepository.findMinimumIndex().orElse(BlockType.EARLIEST.number());
+        return Math.max(block.number(), minIndex);
+    }
+
+    public void setBlockTimestampToContractCallContext(long blockIndex, ContractCallContext contractCallContext) {
```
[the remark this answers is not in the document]
kselveliev: This code was removed after my last refactoring

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmTxProcessorImpl.java
```diff
@@ -139,4 +152,18 @@ protected MessageFrame buildInitialFrame(
...
                         .build();
     }
+
+    private boolean shouldSetBlockTimestamp(boolean isEstimate, BlockType block) {
+        return !isEstimate && block != BlockType.LATEST;
+    }
+
+    private long determineAppropriateBlockNumber(BlockType block) {
+        long minIndex = recordFileRepository.findMinimumIndex().orElse(BlockType.EARLIEST.number());
+        return Math.max(block.number(), minIndex);
```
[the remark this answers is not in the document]
kselveliev: This was removed after refactoring

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmTxProcessorImpl.java
kselveliev: Refactored the code to:

If block < 0 -> throw Bad request -> Invalid block value
If block < min (still valid block) -> return default empty value
If earliest -> use min available
If min <= block <= current -> execute transaction (normal case)
If block > max possible -> return bad request with Unknown block number

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmTxProcessorImpl.java
```diff
@@ -79,14 +88,24 @@ public HederaEvmTransactionProcessingResult execute(
...
         contractCallContext.setEstimate(isEstimate);
 
+        if (shouldSetBlockTimestamp(isEstimate, block)) {
+            Optional<RecordFile> recordFileOptional = findRecordFileByBlock(block);
+            if (recordFileOptional.isPresent()) {
+                contractCallContext.setBlockTimestamp(recordFileOptional.get().getConsensusEnd());
+            } else {
+                throw new BlockNumberNotFoundException("Block number not found");
+            }
+        }
```
steven-sheehy, line 107: Think we should do all this logic inside `ContractCallService.processCall()`. Then block checking is encapsulated in one spot, don't need to use slow exception handling for block not found, don't check block for each gas iteration, and don't need to load slow ThreadLocal multiple times. Can do in a follow up if it's too big of change.

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/viewmodel/BlockType.java
```diff
@@ -55,6 +55,9 @@ private static BlockType extractNumericBlock(String value) {
 
         try {
             long blockNumber = Long.parseLong(cleanedValue, radix);
+            if (blockNumber < 0) {
+                throw new IllegalArgumentException("Invalid block value: " + value);
```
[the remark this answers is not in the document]
kselveliev: Comment address. Added a check for "-" before.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmTxProcessorImpl.java
```diff
@@ -139,4 +158,20 @@ protected MessageFrame buildInitialFrame(
...
     }
+
+    private boolean shouldSetBlockTimestamp(boolean isEstimate, BlockType block) {
+        return !isEstimate && block != BlockType.LATEST;
+    }
+
+    private Optional<RecordFile> findRecordFileByBlock(BlockType block) {
+        if (block == BlockType.EARLIEST) {
+            return recordFileRepository.findRecordFileByIndex(
+                    recordFileRepository.findMinimumIndex().get());
```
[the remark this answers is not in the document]
kselveliev: Comment addressed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/exception/BlockNumberNotFoundException.java
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hedera.mirror.web3.exception;
+
+import com.hedera.mirror.common.exception.MirrorNodeException;
+import java.io.Serial;
+
+public class BlockNumberNotFoundException extends MirrorNodeException {
```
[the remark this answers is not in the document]
kselveliev: Comment addressed.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/ContractController.java
```diff
@@ -123,6 +126,13 @@ private Mono<GenericErrorResponse> validationError(final WebExchangeBindExceptio
         return Mono.just(new GenericErrorResponse(errors));
     }
 
+    @ExceptionHandler
+    @ResponseStatus(BAD_REQUEST)
+    private Mono<GenericErrorResponse> blockValidationError(final BlockNumberOutOfRangeException e) {
```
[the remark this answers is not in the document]
kselveliev: Comment addressed. Could not make it for all of the methods that return BAD_REQUEST. Some of them do not extend Web3Exception.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmTxProcessorImpl.java
```diff
@@ -79,14 +88,24 @@ public HederaEvmTransactionProcessingResult execute(
...
         final var contractCallContext = ContractCallContext.get();
         contractCallContext.setCreate(Address.ZERO.equals(receiver));
         contractCallContext.setEstimate(isEstimate);
 
+        if (shouldSetBlockTimestamp(isEstimate, block)) {
+            Optional<RecordFile> recordFileOptional = findRecordFileByBlock(block);
+            if (recordFileOptional.isPresent()) {
+                contractCallContext.setBlockTimestamp(recordFileOptional.get().getConsensusEnd());
+            } else {
+                throw new BlockNumberNotFoundException("Block number not found");
```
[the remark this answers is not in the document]
kselveliev: Comment addressed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/RecordFileRepository.java
```diff
@@ -37,9 +38,13 @@ public interface RecordFileRepository extends PagingAndSortingRepository<RecordF
...
 
+    @Cacheable(cacheNames = CACHE_NAME, cacheManager = CACHE_MANAGER_RECORD_FILE_EARLIEST, unless = "#result == null")
+    @Query("select min(r.index) from RecordFile r")
+    Optional<Long> findMinimumIndex();
+
     @Cacheable(cacheNames = CACHE_NAME, cacheManager = CACHE_MANAGER_RECORD_FILE_INDEX, unless = "#result == null")
-    @Query("select r.hash from RecordFile r where r.index = ?1")
-    Optional<String> findHashByIndex(long index);
+    @Query("select r from RecordFile r where r.index = ?1")
+    Optional<RecordFile> findRecordFileByIndex(long index);
```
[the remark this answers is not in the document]
kselveliev: Comment addressed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
```diff
@@ -63,13 +71,26 @@ public String processCall(final CallServiceParameters params) {
             if (params.isEstimate()) {
                 result = estimateGas(params);
             } else {
+                final var contractCallContext = ContractCallContext.get();
```
steven-sheehy, line 74: We already have `ctx` variable above.

```suggestion
```

steven-sheehy (dismissed):
LGTM

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
kselveliev: Comment addressed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/StaticBlockMetaSource.java
```diff
@@ -34,9 +34,9 @@ public class StaticBlockMetaSource implements BlockMetaSource {
 
     @Override
     public Hash getBlockHash(long blockNo) {
```
0xivanov, line 36: I think we should modify `computeBlockValues` also to work with `findByIndex`

kselveliev: You mean computeBlockValues() method, this will be done in another task

steven-sheehy (approved):

IvanKavaldzhiev (approved):
LGTM

Issue #7045, opened 2023-10-11
Returning a block hash from within a solidity contract causes an exception when running against the mirror node.


ebadiere:
### Description

An error is returned when running the following snippet on the mirror node:
```
    function getBlockHash() public view returns (bytes32) {
        return blockhash(block.number - 1);
    }
```

The ethCall request body to the mirror node is the following:
```
{
  "gas": 9021272,
  "from": "0x05fba803be258049a27b820088bab1cad2058871",
  "to": "0xc659482c3b814d684457d7764b0738d8f0681bff",
  "data": "0x9663f88f",
  "value": null,
  "estimate": false,
}
```

curl:
```
curl --location 'http://127.0.0.1:8545/api/v1/contracts/call' \
--header 'Content-Type: application/json' \
--data '{
  "gas": 9021272,
  "from": "0x05fba803be258049a27b820088bab1cad2058871",
  "to": "0xc659482c3b814d684457d7764b0738d8f0681bff",
  "data": "0x9663f88f",
  "value": null,
  "estimate": false,
}'
```

Running this against the mirror node has also returned 0x0 values as well, which is also incorrect.  Running against the consensus node always returns a block hash value.

### Steps to reproduce

1. There is a reproducer test in the smart contracts repository, branch `430-block-properties`, in BlockInfo called `should be able to get the hash of a given block when the block number is one of the 256 most recent blocks`
2. You can deploy a smart contract with the solidity function listed in the description and call it through a hard hat application, which is what the test does.

### Hedera network

other

### Version

v0.89.0

### Operating system

None

PR #7091, opened 2023-10-18
HIP-584: Fix block hash exception due to hash value not being sliced

bilyana-gospodinova:
**Description**:
This PR fixes the IllegalArgumentException ("Hex value is too large: expected at most 32 bytes but got 48") that is thrown when we call the blockhash function in a solidity contract. The exception is thrown in StaticBlockMetaSource#getBlockHash -> org.hyperledger.besu.datatypes.Hash::fromHexString. The cause for this is that the persisted hash values in the record files are of length 48 bytes but org.hyperledger.besu.datatypes.Hash::fromHexString expects the input to be of length 32 bytes. Following HIP-415 (https://hips.hedera.com/hip/hip-415) that was implemented in services the returned hash value from StaticBlockMetaSource#getBlockHash needs to be constructed from the first 32 prefix bytes of the hash.

In this PR some integration tests are added in order to cover some typical scenarios related to HIP-415 such as:
1. The block hash of the genesis block needs to be all zeroes.
2. The block hash of the current block needs to be all zeroes.
3. The block hash of the last 256 blocks (not including the current one) is the 32 byte prefix from the 48 byte hash.

In the integration tests the following changes were made:
- The persisted record file hash values are changed to be numeric in order to be parsable later as a hex.
- Preserve a record file in ContractCallTestSetup that is not in the latest block in order to get its index (block number) and test scenario 3.

Fixes https://github.com/hashgraph/hedera-mirror-node/issues/7045


hedera-mirror-common/src/test/java/com/hedera/mirror/common/domain/DomainBuilder.java
```diff
@@ -730,19 +730,19 @@ public DomainWrapper<RecordFile, RecordFile.RecordFileBuilder> recordFile() {
                 .consensusEnd(consensusEnd)
                 .count(1L)
                 .digestAlgorithm(DigestAlgorithm.SHA_384)
-                .fileHash(text(96))
+                .fileHash(numeric(96))
```
steven-sheehy, line 733: Please undo changes to the `DomainBuilder`. The intent of the `DomainBuilder` is to generate random data that matches the format of the real data as much as possible. The real hashes are random byte arrays not numeric. It's also a code smell to depend upon their values being numeric in a test so we really shouldn't rely on such an assumption.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/utils/BlockHashUtil.java
```diff
@@ -0,0 +1,32 @@
...
+import org.apache.tuweni.bytes.Bytes32;
+
+@UtilityClass
+public class BlockHashUtil {
+
+    public static org.hyperledger.besu.datatypes.Hash ethHashFrom(final String hash) {
+        final byte[] hashBytesToConvert = Bytes.fromHexString(hash).toArrayUnsafe();
+        final byte[] prefixBytes = new byte[32];
+        System.arraycopy(hashBytesToConvert, 0, prefixBytes, 0, 32);
+        return org.hyperledger.besu.datatypes.Hash.wrap(Bytes32.wrap(prefixBytes));
```
steven-sheehy, line 30: Doesn't this work?

```suggestion
        return Hash.fromHexString(StringUtils.substring(hash, 0, 64))
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/utils/BlockHashUtil.java
```diff
@@ -0,0 +1,32 @@
...
+package com.hedera.mirror.web3.evm.utils;
+
+import lombok.experimental.UtilityClass;
+import org.apache.tuweni.bytes.Bytes;
+import org.apache.tuweni.bytes.Bytes32;
+
+@UtilityClass
+public class BlockHashUtil {
+
+    public static org.hyperledger.besu.datatypes.Hash ethHashFrom(final String hash) {
```
steven-sheehy, line 26: No need to use fully qualified package names.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallEvmCodesTest.java
```diff
@@ -138,9 +146,45 @@ void getBlockPrevrandao() {
...
+    void getGenesisBlockHashReturnsZero() {
+        final var functionHash = functionEncodeDecoder.functionHashFor("getBlockHash", EVM_CODES_ABI_PATH, 0L);
+        final var serviceParameters = serviceParametersForEvmCodes(functionHash);
+
+        assertThat(contractCallService.processCall(serviceParameters))
+                .isEqualTo("0x0000000000000000000000000000000000000000000000000000000000000000");
+    }
+
+    @Test
+    void getLatestBlockHashReturnsZero() {
```
Nana-EC, line 175: Q: why should the latest block hash return 0's?
We have valid hashes for the latest block - https://testnet.mirrornode.hedera.com/api/v1/blocks?order=desc&limit=1

Is there a specific edge case you're going for?

Nana-EC (changes requested):
Looks mostly good.
Unsure about the all zeros for latest block hash.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/StaticBlockMetaSource.java
```diff
@@ -34,8 +35,9 @@ public class StaticBlockMetaSource implements BlockMetaSource {
     @Override
     public Hash getBlockHash(long blockNo) {
         final var fileHash = recordFileRepository.findHashByIndex(blockNo);
-        return fileHash.map(Hash::fromHexString)
-                .orElseThrow(() -> new MissingResultException(String.format("No record file with index: %d", blockNo)));
+
+        return BlockHashUtil.ethHashFrom(fileHash.orElseThrow(
```
steven-sheehy, line 39: Is the change from `map` necessary to not throw a certain exception type? If not, prefer older map approach.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/properties/StaticBlockMetaSourceTest.java
```diff
@@ -47,9 +50,13 @@ void setUp() {
...
     void getBlockHashReturnsCorrectValue() {
-        final var fileHash = "0x00000000000000000000000000000000000000000000000000000000000004e4";
+        final var fileHash =
+                "37313862636664302d616365352d343861632d396430612d36393036316337656236626333336466323864652d346100";
         given(repository.findHashByIndex(1)).willReturn(Optional.of(fileHash));
-        assertThat(subject.getBlockHash(1)).isEqualTo(Hash.fromHexString(fileHash));
+        final var expected = Hash.wrap(
+                Bytes32.wrap(Bytes.fromHexString("0x37313862636664302d616365352d343861632d396430612d3639303631633765")
+                        .toArrayUnsafe()));
+        assertThat(subject.getBlockHash(1)).isEqualTo(expected);
```
steven-sheehy, line 59:
```suggestion
        final var expected = Hash.fromHexString("0x37313862636664302d616365352d343861632d396430612d3639303631633765");
        assertThat(subject.getBlockHash(1)).isEqualTo(expected);
```

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallEvmCodesTest.java
```diff
@@ -138,9 +146,45 @@ void getBlockPrevrandao() {
...
+        persistEntities();
+        areEntitiesPersisted = true;
+
+        final var functionHash = functionEncodeDecoder.functionHashFor(
+                "getBlockHash", EVM_CODES_ABI_PATH, recordFileForBlockHash.getIndex());
+        final var serviceParameters = serviceParametersForEvmCodes(functionHash);
+
+        assertThat(contractCallService.processCall(serviceParameters))
+                .isEqualTo(BlockHashUtil.ethHashFrom(recordFileForBlockHash.getHash())
+                        .toString());
```
steven-sheehy, line 162: No need for the util.
```suggestion
                .isEqualTo(recordFileForBlockHash.getHash().substring(0, 64));
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/utils/BlockHashUtil.java
steven-sheehy, line 1: I don't think one method warrants its own utility class, especially since it's not used outside of `StaticBlockMetaSource` and tests. I have provided suggestions to remove its use in tests while making them simpler.

steven-sheehy (changes requested):

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallEvmCodesTest.java
IvanKavaldzhiev: Does the execution fall into this logic?

 // If the current block is the genesis block or the sought block is
    // not within the last 256 completed blocks, zero is returned.
    if (currentBlockNumber == 0
        || soughtBlock < (mostRecentBlockNumber - MAX_RELATIVE_BLOCK)
        || soughtBlock > mostRecentBlockNumber) {
      frame.pushStackItem(Bytes32.ZERO);

hedera-mirror-common/src/test/java/com/hedera/mirror/common/domain/DomainBuilder.java
bilyana-gospodinova: "The intent of the DomainBuilder is to generate random data that matches the format of the real data as much as possible" - I agree with that. However, when we pass in the DomainBuilder a random string, e.g. "xyz" it gets persisted in the DB as "xyz" so when we read this value later and try to parse it to get the Hash we get an exception as this is not a parseable hex string:
> Illegal character 'v' found at index 1 in hex binary representation
> com.hedera.mirror.web3.exception.MirrorEvmTransactionException: Illegal character 'v' found at index 1 in hex binary representation. 

This is why I modified the persisted file hashes to be numeric so that we can parse them later successfully. 

Instead of the numeric() method and in order to be able to generate values as close as possible to the real data I would suggest this method:
public String hash(int characters) {
    return RandomStringUtils.random(characters, "0123456789abcdef");
}

What do you think about that?

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallEvmCodesTest.java
bilyana-gospodinova: Yes, exactly.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/utils/BlockHashUtil.java
bilyana-gospodinova: It does. The change is addressed.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/utils/BlockHashUtil.java
bilyana-gospodinova: The file was removed as suggested.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/StaticBlockMetaSource.java
bilyana-gospodinova: The change is reverted.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/properties/StaticBlockMetaSourceTest.java
bilyana-gospodinova: Comment addressed.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallEvmCodesTest.java
bilyana-gospodinova: Comment addressed.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/utils/BlockHashUtil.java
bilyana-gospodinova: The file was removed as suggested.

hedera-mirror-common/src/test/java/com/hedera/mirror/common/domain/DomainBuilder.java
steven-sheehy: That sounds reasonable.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallEvmCodesTest.java
steven-sheehy: This seems like a bug when applied to Hedera then. Mirror nodes can return block hashes beyond the last 256 blocks and can return the block hash for the latest block since it's final. Here because of `long mostRecentBlockNumber = currentBlockNumber - 1` we're explicitly ignoring latest block even when the code loads the same values for `soughtBlock` and `currentBlockNumber`. Can we open a ticket to investigate?

steven-sheehy (approved):
LGTM

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallEvmCodesTest.java
Nana-EC: Agreed the 256 blocks is a design choice by EVM specs/Ethereum. Mirror Node is more capable so we may be able to support it or we can just match equivalence and not do the extra work.
Agreed with Steven @IvanKavaldzhiev  can we open a ticket to investigate this and decide

Nana-EC (approved):
LG
Let's follow up on that ticket around the 256 blocks consideration

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallEvmCodesTest.java
IvanKavaldzhiev: Yes, definitely we should investigate this behaviour. Will open a ticket.

PR #7095, opened 2023-10-19
Remove generation of token history row when total supply changes for Mint/Wipe/Burn

### LINKED ISSUE #6998: Don't generate a token history row when total supply changes
### Problem

We currently upsert a history row when the `token.total_supply` column is accumulated for token mints, burns, etc. We should only add history rows for create, update or delete operations on the token entity itself.

### Solution

Change `token.total_supply` to not cause a history row when modified

### Alternatives

_No response_

mgoelswirlds:
**Description**:

Remove the creation of history rows when the `token.total_supply` column is adjusted for token mints, burns, or wipes. History rows are added for create, update, or delete operations on the token entity itself.

* Remove setting of timestamp range on above mentioned token operations.
* Add token history row check to existing tests.

**Related issue(s)**:

Fixes #6998

**Notes for reviewer**:

**Checklist**


jascks (dismissed):
LGTM

xin-hedera (changes requested):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/entity/sql/SqlEntityListener.java
```diff
@@ -852,67 +852,94 @@ private Schedule mergeSchedule(Schedule cachedSchedule, Schedule schedule) {
...
+        // If previous doesn't have history, merge reversely from current to previous
+        var src = previous.isHistory() ? previous : current;
+        var dest = previous.isHistory() ? current : previous;
+
+        boolean isSameTimestampLower = Objects.equals(current.getTimestampLower(), previous.getTimestampLower());
+        if (current.isHistory() && isSameTimestampLower) {
+            // Copy from current to previous if the updates have the same lower timestamp (thus from same transaction)
+            src = current;
+            dest = previous;
+        }
```
xin-hedera, line 876: more than one token updates with the same lower timestamp shouldn't happen

xin-hedera (changes requested):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/entity/sql/SqlEntityListener.java
```diff
@@ -852,67 +852,94 @@ private Schedule mergeSchedule(Schedule cachedSchedule, Schedule schedule) {
...
+        Long currentTotalSupply = current.getTotalSupply();
+        Long previousTotalSupply = previous.getTotalSupply();
+
+        if (!current.isHistory()) {
+            if (currentTotalSupply > 0) {
+                previous.setTotalSupply(previousTotalSupply + currentTotalSupply);
+            } else {
+                previous.setTotalSupply(previousTotalSupply - currentTotalSupply);
+            }
+            return previous;
```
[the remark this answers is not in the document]
mgoelswirlds: Will add the method. But the idea was to add history rows only for entity updates.As per the ticket` add history rows for create, update or delete operations on the token entity itself.` So did remove creation in other places.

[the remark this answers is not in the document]
xin-hedera: paused or not is part of token's state, even though it's not done by token update tx. pause / unpause shouldn't be an often operation on tokens so we should change it back with history rows.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/entity/sql/SqlEntityListener.java
mgoelswirlds: ok.

hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/token/AbstractToken.java
```diff
@@ -109,4 +109,14 @@ public void setName(String name) {
     public void setSymbol(String symbol) {
         this.symbol = DomainUtils.sanitize(symbol);
     }
+
+    public void updateTotalSupply(Long newTotalSupply) {
+        if (newTotalSupply != null && this.totalSupply != null) {
```
xin-hedera, line 114: what if totalSupply is null and newTotalSupply is not null?

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/record/entity/sql/SqlEntityListenerTest.java
```diff
@@ -1946,6 +1946,94 @@ void onTokenMerge() {
...
+
+        var tokenHistory = TestUtils.clone(tokenCreate);
+        tokenHistory.setTimestampUpper(tokenUpdate.getTimestampLower());
+        assertThat(findHistory(Token.class)).containsExactlyInAnyOrder(tokenHistory);
+    }
+
+    @Test
+    void onTokenMergeWithCreateThenMintThenUpdate() {
+        // given
+        // Create does not have a history row
```
xin-hedera, line 1996: create definitely has timestamp range set

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/record/entity/sql/SqlEntityListenerTest.java
```diff
@@ -1946,6 +1946,94 @@ void onTokenMerge() {
...
+        var tokenCreate = domainBuilder.token().get();
+        sqlEntityListener.onToken(tokenCreate);
+
+        // Token mint does not have history
+        var tokenMint = domainBuilder
+                .token()
+                .customize(t ->
+                        t.tokenId(tokenCreate.getTokenId()).timestampRange(null).totalSupply(-5_000L))
```
xin-hedera, line 2004: why mint will end up with a negative total supply?

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/record/entity/sql/SqlEntityListenerTest.java
```diff
@@ -1946,6 +1946,94 @@ void onTokenMerge() {
...
+                .get();
+        sqlEntityListener.onToken(tokenMint);
+
+        // Update has a history row
+        var tokenUpdate = domainBuilder
+                .token()
+                .customize(t -> t.freezeDefault(true)
+                        .pauseStatus(TokenPauseStatusEnum.PAUSED)
+                        .supplyType(TokenSupplyTypeEnum.FINITE)
```
xin-hedera, line 2013: token's supply type is immutable

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/record/entity/sql/SqlEntityListenerTest.java
```diff
@@ -1946,6 +1946,94 @@ void onTokenMerge() {
...
+        tokenMerged.setTotalSupply(tokenMint.getTotalSupply());
+        assertThat(tokenRepository.findAll()).containsExactlyInAnyOrder(tokenMerged);
+
+        var tokenHistory = TestUtils.clone(tokenCreate);
+        tokenHistory.setTimestampUpper(tokenUpdate.getTimestampLower());
+        assertThat(findHistory(Token.class)).containsExactlyInAnyOrder(tokenHistory);
+    }
+
+    @Test
+    void onTokenMergeWithCreateThenMintThenUpdate() {
```
xin-hedera, line 1994: this should be a parameterized test, take a look at `onCryptoAllowanceHistory`

xin-hedera:
There is one more `setTimestampLower` to remove: https://github.com/hashgraph/hedera-mirror-node/blob/5eb770b61265920463508fd39e7aadaff40cee9b/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/entity/EntityRecordItemListener.java#L397C17-L397C30

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/record/entity/sql/SqlEntityListenerTest.java
xin-hedera: misunderstood the comment, thought it meant token create has not timestamp range set

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/entity/sql/SqlEntityListener.java
```diff
@@ -852,72 +854,82 @@ private Schedule mergeSchedule(Schedule cachedSchedule, Schedule schedule) {
...
-        if (currentTotalSupply == null) {
-            current.setTotalSupply(previousTotalSupply);
-        } else if (previousTotalSupply != null && currentTotalSupply < 0) {
+        if (destTotalSupply == null) {
+            dest.setTotalSupply(srcTotalSupply);
+        } else if (srcTotalSupply != null && destTotalSupply < 0) {
             // Negative from a token transfer of a token dissociate of a deleted token, so we aggregate the change.
-            current.setTotalSupply(previousTotalSupply + currentTotalSupply);
+            dest.setTotalSupply(srcTotalSupply + destTotalSupply);
         }
```
xin-hedera, line 925: should replace these with `updateTotalSupply` since now the logic is encapsulated, might be good to also copy the comment here to the new function

mgoelswirlds: The logic is a little different. Because the negative balance check is for the destination here where as the equivalent check is for the source. But I think the one in the method needs to match this right?

xin-hedera: if you replace these with `dest.updateTotalSupply(src.getTotalSupply())` it's the exact same logic, right?

mgoelswirlds: I did replace it , when I realized that the difference is of the totalSupply amount being checked for a negative value.
In the method we are checking for newTotalSupply which is equivalent to srcTotalSupply and in the other case we are checking for destTotalSupply to be negative. So I believe the check should be for src value in both the places and not destination. Does that seem right?

mgoelswirlds: to simplify the question in the above method what happens if we have a negative totalSupply?

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/record/entity/sql/SqlEntityListenerTest.java
```diff
@@ -2004,7 +2116,7 @@ void onTokenMergeNegativeTotalSupply() {
         // then
         var expected = TestUtils.clone(tokenCreate);
         expected.setTimestampLower(tokenDissociateDeleted.getTimestampLower());
-        expected.setTotalSupply(expected.getTotalSupply() - 10);
```
mgoelswirlds, line 2007: @xin-hedera This is the case I was talking about. the negative balance is ignored because we don't check for the totalSupply value.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/entity/sql/SqlEntityListener.java
```diff
@@ -852,72 +854,81 @@ private Schedule mergeSchedule(Schedule cachedSchedule, Schedule schedule) {
...
-        if (current.getFeeScheduleKey() == null) {
-            current.setFeeScheduleKey(previous.getFeeScheduleKey());
+        if (!current.hasHistory()) {
+            previous.updateTotalSupply(current.getTotalSupply());
+            return previous;
         }
 
-        if (current.getFreezeKey() == null) {
-            current.setFreezeKey(previous.getFreezeKey());
+        if (!previous.hasHistory() && current.hasHistory()) {
```
mgoelswirlds, line 863: There actually is a test case for this but it still complains of partial coverage.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/entity/sql/SqlEntityListener.java
```diff
@@ -852,72 +854,81 @@ private Schedule mergeSchedule(Schedule cachedSchedule, Schedule schedule) {
...
+            current.updateTotalSupply(previous.getTotalSupply());
+            return current;
         }
 
-        if (current.getKycKey() == null) {
-            current.setKycKey(previous.getKycKey());
+        // If previous doesn't have history, merge reversely from current to previous
+        var src = previous.hasHistory() ? previous : current;
+        var dest = previous.hasHistory() ? current : previous;
```
steven-sheehy, line 870: This is dead code since you've already ruled out previous not having history above. This logic can only run when both previous and current are regular updates. Can remove dest/src in favor of using more straightforward previous/current. Basically revert the bottom part of the original method to how it was before except using updateTotalSupply.

```suggestion
```

hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/token/AbstractToken.java
```diff
@@ -109,4 +109,17 @@ public void setName(String name) {
     public void setSymbol(String symbol) {
         this.symbol = DomainUtils.sanitize(symbol);
     }
+
+    public void updateTotalSupply(Long newTotalSupply) {
```
steven-sheehy, line 113: Would prefer an overloaded `setTotalSupply` unless that causes issues.

steven-sheehy (changes requested):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/entity/sql/SqlEntityListener.java
xin-hedera: it's partial coverage because not all 4 combinations were tested. not requesting changes in test cases though, just want to clarify.

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/domain/token/AbstractTokenTest.java
```diff
@@ -0,0 +1,40 @@
...
+import org.junit.jupiter.params.provider.CsvSource;
+
+class AbstractTokenTest {
+
+    @ParameterizedTest
+    @CsvSource(
+            textBlock = """
+            -100, , -100
+            , 500, 500
+            1200,-500, 700
```
xin-hedera, line 32: add one more test data `, -500, -500`

xin-hedera (changes requested):

xin-hedera (approved):
LGTM for all the changes I requested.

jascks (approved):
Can't speak to any outstanding requested changes/comments, but otherwise this PR looks good based on the description and my own similar experience in the past.

steven-sheehy (approved):
LGTM

Issue #7052, opened 2023-10-13
HIP-584: Historical support - Extend nft, nft allowance and token account DB queries, to filter by to filter by timestamp


kselveliev:
### Problem

For historical support we need to extend db queries to have a timestamp filter.
Using this filter we can obtain records based on a historical block passed in eth_call.

### Solution

We need the extend the following queries in web3.

Extend query for nft.

Extend query for nft allowance.

Extend query for token account.

Document and test new query methods. 

Use this [issue](https://github.com/hashgraph/hedera-mirror-node/issues/6812) for reference.

PR #7096, opened 2023-10-20
HIP-584 Historical: Extend nft, nft allowance and token account DB queries to filter by timestamp

bilyana-gospodinova:
**Description**:
For historical support we need to extend db queries to have a timestamp filter.
Using this filter we can obtain records based on a historical block passed in eth_call.

Fixes https://github.com/hashgraph/hedera-mirror-node/issues/7052

**Checklist**


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/NftRepository.java
```diff
@@ -40,6 +40,38 @@ public interface NftRepository extends CrudRepository<Nft, AbstractNft.Id> {
...
+     * @param blockTimestamp the block timestamp used to filter the results.
+     * @return an Optional containing the nft's state at the specified timestamp.
+     *         If there is no record found for the given criteria, an empty Optional is returned.
+     */
+    @Query(
+            value = "( "
+                    + "select n.* from nft n "
+                    + "left join entity e on e.id = n.token_id "
+                    + "where n.token_id=:tokenId and n.serial_number=:serialNumber and lower(n.timestamp_range)<=:blockTimestamp "
+                    + "and n.deleted is false and e.deleted is not true "
```
IvanKavaldzhiev, line 59: n.deleted is false -> n.deleted is not true

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/NftRepository.java
```diff
@@ -40,6 +40,38 @@ public interface NftRepository extends CrudRepository<Nft, AbstractNft.Id> {
...
+                    + ") "
+                    + "union all "
+                    + "( "
+                    + "select n.* from nft_history n "
+                    + "left join entity_history e on e.id = n.token_id "
+                    + "where n.token_id=:tokenId and n.serial_number=:serialNumber and lower(n.timestamp_range)<=:blockTimestamp "
+                    + "and n.deleted is false and e.deleted is not true "
```
IvanKavaldzhiev, line 66: n.deleted is false -> n.deleted is not true

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenAccountRepository.java
```diff
@@ -39,4 +39,40 @@ public interface TokenAccountRepository extends CrudRepository<TokenAccount, Abs
...
+     *         If there is no record found for the given criteria, an empty Optional is returned.
+     */
+    @Query(
+            value =
+                    """
+                    (
+                        select *
+                        from token_account
+                        where account_id = :#{#id.accountId}
+                            and token_id = :#{#id.tokenId}
```
steven-sheehy, line 60: Don't use reflection based SPEL.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenAccountRepository.java
```diff
@@ -39,4 +39,40 @@ public interface TokenAccountRepository extends CrudRepository<TokenAccount, Abs
...
+     * @param blockTimestamp  the block timestamp used to filter the results.
+     * @return an Optional containing the token account's state at the specified timestamp.
+     *         If there is no record found for the given criteria, an empty Optional is returned.
+     */
+    @Query(
+            value =
+                    """
+                    (
+                        select *
+                        from token_account
```
steven-sheehy, line 58: Not sure of the use case, but probably you also want to return it if it's associated at the time since consensus nodes don't really keep dissociated token relationships? Or will you check that in memory?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/NftRepository.java
steven-sheehy, line 59: It's trickier since you would also need to check entity_history since it might be deleted now, but not at the block timestamp.

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/NftRepository.java
```diff
@@ -40,6 +40,38 @@ public interface NftRepository extends CrudRepository<Nft, AbstractNft.Id> {
...
+     * and returns the one that was valid just before or equal to the provided block timestamp.
+     *
+     * @param tokenId the token id of the nft to be retrieved.
+     * @param serialNumber the serial number of the nft to be retrieved.
+     * @param blockTimestamp the block timestamp used to filter the results.
+     * @return an Optional containing the nft's state at the specified timestamp.
+     *         If there is no record found for the given criteria, an empty Optional is returned.
+     */
+    @Query(
+            value = "( "
```
xin-hedera, line 55: can you change it to use java text block?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/NftRepository.java
xin-hedera: we can change it to `(e.deleted is not true or (lower(e.timestamp_range) > :blockTimestamp)` since lower(e.timestamp_range) is when the nft collection is deleted.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/NftRepository.java
```diff
@@ -40,6 +40,38 @@ public interface NftRepository extends CrudRepository<Nft, AbstractNft.Id> {
...
+                    + "select n.* from nft n "
+                    + "left join entity e on e.id = n.token_id "
+                    + "where n.token_id=:tokenId and n.serial_number=:serialNumber and lower(n.timestamp_range)<=:blockTimestamp "
+                    + "and n.deleted is false and e.deleted is not true "
+                    + ") "
+                    + "union all "
+                    + "( "
+                    + "select n.* from nft_history n "
+                    + "left join entity_history e on e.id = n.token_id "
```
xin-hedera, line 64: remove the `entity_history` join. a nft_history row can only correlate with its entity row by nft_history.timestamp_range, which means, a nft history row can span over multiple nft collection entity rows.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/NftRepository.java
```diff
@@ -40,6 +40,38 @@ public interface NftRepository extends CrudRepository<Nft, AbstractNft.Id> {
...
+                    + "where n.token_id=:tokenId and n.serial_number=:serialNumber and lower(n.timestamp_range)<=:blockTimestamp "
+                    + "and n.deleted is false and e.deleted is not true "
+                    + "order by lower(n.timestamp_range) desc "
+                    + "limit 1 "
+                    + ") "
+                    + "order by timestamp_range desc "
+                    + "limit 1 ",
```
xin-hedera, line 71: instead of join `entity` / `entity_history` in the subqueries to check if the collection is deleted, we should join in the outer query, sth like

```
select n.*
from (
  (nft table subquery)
  union all
  (nft_history table subquery)
) as n
join entity e on e.id = n.token_id
where e.deleted is not true or lower(e.timestamp_range) > :blockTimestamp
order by n.timestamp_range desc
limit 1
```

xin-hedera (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/NftRepository.java
bilyana-gospodinova: Would you please explain this part: "or lower(e.timestamp_range) > :blockTimestamp"? Why do we need that?

xin-hedera: if an entity is deleted, only in the `entity` table, `deleted = true`, and `lower(entity.timestamp_range` is when it's deleted, so `e.deleted is not true or lower(e.timestamp_range) > :blockTimestamp`

- if the entity is not deleted
- or (implicitly it's deleted) it's deleted after blockTimestamp

bilyana-gospodinova: Oh, I get it now. Thank you, I will update the query as suggested.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/NftRepository.java
kselveliev: Comment addressed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/NftRepository.java
kselveliev: Comment addressed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/NftAllowanceRepository.java
```diff
@@ -33,4 +34,42 @@ public interface NftAllowanceRepository extends CrudRepository<NftAllowance, Id>
...
+     */
+    @Query(
+            value =
+                    """
+                    (
+                        select *
+                        from nft_allowance
+                        where token_id = :#{#id.tokenId}
+                            and owner = :#{#id.owner}
+                            and spender = :#{#id.spender}
```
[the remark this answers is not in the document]
kselveliev: Comment addressed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenAccountRepository.java
kselveliev: Comment addressed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenAccountRepository.java
kselveliev: Will check dissociated in memory, for current data we are filtering out dissociated in memory.

image

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/NftRepository.java
kselveliev: Applied the (e.deleted is not true or (lower(e.timestamp_range) > :blockTimestamp) suggestion.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/NftRepository.java
kselveliev: Comment addressed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/NftRepository.java
kselveliev: Removed the entity history check and applied what you suggested below.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/NftRepository.java
kselveliev: Applied the suggested change above. Please check the refactored query

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenAccountRepository.java
kselveliev: Hm now that I think about this.
Considering the scenario where the closest token account to the timestamp is dissociated, while the previous one remains associated, the current in-memory filtering might not be effective for historical data, as it's designed for latest data. Moving the 'associated is true' condition into the SQL query, instead of applying it in memory as in the snippet, would ensure we fetch the nearest associated token account to the block timestamp. What do you think about this?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/NftAllowanceRepository.java
```diff
@@ -33,4 +34,48 @@ public interface NftAllowanceRepository extends CrudRepository<NftAllowance, Id>
...
     List<NftAllowance> findByOwnerAndApprovedForAllIsTrue(long owner);
+
+    /**
+     * Retrieves the most recent state of a nft allowance by its owner up to a given block timestamp.
+     * The method considers both the current state of the nft allowance and its historical states
+     * and returns the latest valid just before or equal to the provided block timestamp.
+     *
+     * @param owner the ID of the owner
+     * @param blockTimestamp  the block timestamp used to filter the results.
+     * @return an Optional containing the nft allowance's state at the specified timestamp.
```
xin-hedera, line 45: nit, it returns a list

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenAccountRepository.java
kselveliev: I agree with the last comment, I believe we should not check association in sql, and simply apply the filter as in the above screenshot. No new changes were added to the code after this comment.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/NftAllowanceRepository.java
kselveliev: Comment addressed

xin-hedera (approved):
LGTM

steven-sheehy (approved):
LGTM

PR #7098, opened 2023-10-20
Extend tests for estimateGas with contract transaction execution

### LINKED ISSUE #7086: Extend estimateGas tests to include transaction execution
### Problem

Currently, our tests verify the `/api/v1/call` endpoint for estimateGas functionality. To improve our testing further, it would be beneficial to add tests to check if the estimated gas is adequate for processing the same transaction through the consensus node.

### Solution

Implement additional tests that will perform the following actions:

1. Retrieve the estimateGas value from the `/api/v1/call` endpoint for a specified operation.
2. Execute the identical transaction with the gas limit acquired from step 1 through the consensus node.

### Alternatives

_No response_

yiliev0:
**Description**: This PR introduces additional tests to validate the estimateGas functionality provided by the `/api/v1/call` endpoint. The new tests retrieve the estimated gas for a specified operation, then execute the identical transaction through the consensus node with the acquired gas limit, ensuring accurate gas estimations for successful transaction processing. 

**Related issue(s)**:

Fixes #7086

**Notes for reviewer**: I've added an extra test for estimateGas w/o precompile, covering a scenario where estimateGas is called for contract deployment with contract bytecode as data. This test is out of scope of the original issue.

**Checklist**

### DIFF
--- a/hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/EstimatePrecompileFeature.java
+++ b/hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/EstimatePrecompileFeature.java
@@ -120,19 +120,30 @@ import static com.hedera.mirror.test.e2e.acceptance.util.TestUtil.asByteArray;
 import static com.hedera.mirror.test.e2e.acceptance.util.TestUtil.asLongArray;
 import static com.hedera.mirror.test.e2e.acceptance.util.TestUtil.nftAmount;
 import static com.hedera.mirror.test.e2e.acceptance.util.TestUtil.to32BytesString;
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import com.esaulpaugh.headlong.abi.Tuple;
 import com.google.protobuf.InvalidProtocolBufferException;
 import com.hedera.hashgraph.sdk.AccountId;
+import com.hedera.hashgraph.sdk.AccountUpdateTransaction;
+import com.hedera.hashgraph.sdk.ContractFunctionParameters;
+import com.hedera.hashgraph.sdk.KeyList;
 import com.hedera.hashgraph.sdk.NftId;
+import com.hedera.hashgraph.sdk.PrecheckStatusException;
+import com.hedera.hashgraph.sdk.ReceiptStatusException;
 import com.hedera.hashgraph.sdk.TokenId;
+import com.hedera.hashgraph.sdk.TokenUpdateTransaction;
 import com.hedera.mirror.test.e2e.acceptance.client.AccountClient;
 import com.hedera.mirror.test.e2e.acceptance.client.AccountClient.AccountNameEnum;
+import com.hedera.mirror.test.e2e.acceptance.client.ContractClient.ExecuteContractResult;
 import com.hedera.mirror.test.e2e.acceptance.client.TokenClient;
 import com.hedera.mirror.test.e2e.acceptance.props.ContractCallRequest;
 import com.hedera.mirror.test.e2e.acceptance.props.ExpandedAccountId;
+import com.hedera.mirror.test.e2e.acceptance.props.MirrorTransaction;
 import com.hedera.mirror.test.e2e.acceptance.response.ContractCallResponse;
+import com.hedera.mirror.test.e2e.acceptance.response.MirrorTransactionsResponse;
+import com.hedera.mirror.test.e2e.acceptance.response.NetworkTransactionResponse;
 import io.cucumber.java.en.And;
 import io.cucumber.java.en.Given;
 import io.cucumber.java.en.Then;
@@ -142,6 +153,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import java.util.concurrent.TimeoutException;
 import java.util.function.Consumer;
 import lombok.CustomLog;
 import lombok.Getter;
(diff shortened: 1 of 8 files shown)


hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/EstimatePrecompileFeature.java
```diff
@@ -1614,6 +1626,290 @@ public void exchangeRateTinyBarsToTinyCentsEstimateGas() {
...
+
+        ExecuteContractResult executeContractResult = contractClient.executeContract(
+                deployedContract.contractId(), gas, contractMethods.getSelector(), parameters, null);
+
+        networkTransactionResponse = executeContractResult.networkTransactionResponse();
+        verifyMirrorTransactionsResponse(mirrorClient, 200);
+        return executeContractResult;
+    }
+
+    private void verifySuccessfulTransaction(ExecuteContractResult contractResult) {
```
steven-sheehy, line 1643: This method seems to duplicate the functionality of the existing `verifyMirrorTransactionsResponse()` and can be deleted since `executeContractTransaction()` already calls `verifyMirrorTransactionsResponse()`.

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/EstimatePrecompileFeature.java
```diff
@@ -1614,6 +1626,290 @@ public void exchangeRateTinyBarsToTinyCentsEstimateGas() {
...
+                mirrorTransactionsResponse.getTransactions().get(0);
+
+        assertThat(mirrorTransaction.getResult()).isEqualTo("SUCCESS");
+    }
+
+    public long estimateGas(String data, String contractAddress) {
+        var response = estimateContract(data, contractAddress);
+        return response.getResultAsNumber().longValue();
+    }
+
```
steven-sheehy, line 1656: Getting the below error when running this branch against previewnet. Can you check?

```
    2023-10-23T22:02:47.654Z ERROR ForkJoinPool-1-worker-4 c.h.m.t.e.a.s.CallFeature Scenario failure: 
    I ensure token "NFT" has been created
    I ensure token "FUNGIBLE" has been created
    I successfully create ERC contract
    I successfully create Precompile contract
    I successfully create EstimateGas contract
    the mirror node REST API should return status 200 for the estimate contract transaction <- Cannot invoke "com.hedera.mirror.test.e2e.acceptance.response.NetworkTransactionResponse.getTransactionIdStringNoCheckSum()" because "this.networkTransactionResponse" is null
    I call function with IERC721Metadata token "NFT" name
    I call function with IERC721Metadata token "NFT" symbol
    I call function with IERC721Metadata token "NFT" totalSupply
    I call function with IERC721 token "NFT" balanceOf owner
    I call function with HederaTokenService isToken token "FUNGIBLE"
    I call function with HederaTokenService isFrozen token "FUNGIBLE", account
    I call function with HederaTokenService isKyc token "FUNGIBLE", account
    I call function with HederaTokenService getTokenDefaultFreezeStatus token "FUNGIBLE"
    I call function with HederaTokenService getTokenDefaultKycStatus token "FUNGIBLE"
    I call function with update and I expect return of the updated value
    I call function that makes N times state update
    I call function with nested deploy using create function
    I call function with nested deploy using create2 function
    I call function with transfer that returns the balance
     java.lang.NullPointerException: Cannot invoke "com.hedera.mirror.test.e2e.acceptance.response.NetworkTransactionResponse.getTransactionIdStringNoCheckSum()" because "this.networkTransactionResponse" is null
        at com.hedera.mirror.test.e2e.acceptance.steps.AbstractFeature.verifyMirrorTransactionsResponse(AbstractFeature.java:90)
        at com.hedera.mirror.test.e2e.acceptance.steps.CallFeature.verifyMirrorAPIResponses(CallFeature.java:90)
        at ✽.the mirror node REST API should return status 200 for the estimate contract transaction(classpath:features/contract/call.feature:10)
```

hedera-mirror-test/src/test/resources/features/contract/estimate.feature
```diff
@@ -42,3 +42,4 @@ Feature: EstimateGas Contract Base Coverage Feature
     Then I call estimateGas with IERC20 token approve using evm address as receiver
```
steven-sheehy, line 42: This file does not have `@acceptance` so isn't getting ran in any environment. Can we test and add?

hedera-mirror-test/src/test/resources/features/contract/estimatePrecompile.feature
```diff
@@ -168,4 +168,23 @@ Feature: EstimateGas Contract Base Coverage Feature
     Then I call estimateGas with exchange rate tinycents to tinybars
     Then I call estimateGas with exchange rate tinybars to tinycents
     #Then I call estimateGas with pseudo random seed
-    #Then I call estimateGas with pseudo random number
\ No newline at end of file
+    #Then I call estimateGas with pseudo random number
+    #### These tests below are recommended to be kept at the end as they change the state #####
+    Then I call estimateGas with balanceOf function for fungible token and execute it using gas limit
```
steven-sheehy, line 173: There's no indication that these are executing against HAPI or use `ContractCallTransaction`. Same comment for others.

```suggestion
    Then I call estimateGas with balanceOf function for fungible token and verify the estimated gas against HAPI
```

hedera-mirror-test/src/test/resources/features/contract/estimatePrecompile.feature
```diff
@@ -168,4 +168,23 @@ Feature: EstimateGas Contract Base Coverage Feature
...
+    Then I call estimateGas with balanceOf function for NFT and execute it with gas limit
```
steven-sheehy, line 174: It's inefficient to call mirror contract/call with the same inputs twice. We should move the pre-existing calls from above that these mirror down here so they only get executed once and verify both gas range and against HAPI.

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/EstimatePrecompileFeature.java
```diff
@@ -1614,6 +1626,290 @@ public void exchangeRateTinyBarsToTinyCentsEstimateGas() {
...
+                ALLOWANCE,
+                asAddress(fungibleKycUnfrozenTokenId),
+                asAddress(admin),
+                asAddress(receiverAccountAlias));
+        var estimateGasValue = estimateGas(data, estimatePrecompileContractSolidityAddress);
+
+        var parameters = new ContractFunctionParameters()
+                .addAddress(fungibleKycUnfrozenTokenId.toSolidityAddress())
+                .addAddress(admin.getAccountId().toSolidityAddress())
+                .addAddress(receiverAccountAlias);
```
steven-sheehy, line 1778: A form of code duplication in generating contract data.. We should find a way to not have to encode the same data twice just because we're invoking MAPI and HAPI separately. Can we use `ContractFunctionParameters` to generate input to mirror contract/call or vice versa?

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/EstimatePrecompileFeature.java
```diff
@@ -1614,6 +1626,290 @@ public void exchangeRateTinyBarsToTinyCentsEstimateGas() {
...
+
+        var parameters = new ContractFunctionParameters()
+                .addAddress(nonFungibleKycUnfrozenTokenId.toSolidityAddress())
+                .addAddress(admin.getAccountId().toSolidityAddress());
+        var result = executeContractTransaction(deployedErcTestContract, estimateGasValue, BALANCE_OF, parameters);
+        verifySuccessfulTransaction(result);
+    }
+
+    @And("I update the account and token keys")
+    public void updateAccountAndTokenKeys() throws PrecheckStatusException, TimeoutException, ReceiptStatusException {
```
steven-sheehy, line 1682: Why do we need all these updates?

steven-sheehy (changes requested):

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/EstimatePrecompileFeature.java
```diff
@@ -1614,6 +1626,290 @@ public void exchangeRateTinyBarsToTinyCentsEstimateGas() {
...
+
+        var parameters = new ContractFunctionParameters()
+                .addAddress(fungibleKycUnfrozenTokenId.toSolidityAddress())
+                .addAddress(admin.getAccountId().toSolidityAddress());
+        var result = executeContractTransaction(deployedErcTestContract, estimateGasValue, BALANCE_OF, parameters);
+        verifySuccessfulTransaction(result);
+    }
+
+    @Then("I call estimateGas with balanceOf function for NFT and execute it with gas limit")
+    public void executeBalanceOfFunctionNonFungibleWithLimitedGas() {
```
IvanKavaldzhiev, line 1670: We can extract the logic inside

I call estimateGas with balanceOf function for fungible token and execute it using gas limit
and
I call estimateGas with balanceOf function for NFT and execute it with gas limit

in a common method, since the only difference is the token address.

yiliev0: Extracted the logic into a common method as suggested. Thanks.

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/EstimatePrecompileFeature.java
yiliev0: Removed. In fact, we can just use the `verifyMirrorTransactionsResponse()` inside `executeContractTransaction` and it would do the same thing. Thanks!

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/EstimatePrecompileFeature.java
yiliev0: I found the problem. The problem originates when the tests are run parallel. So the `networkTransactionResponse` never gets populated because the singleton method for the contract creation does not invoke the creation itself, but the already stored contract in the mapper, thus leaving us without a transaction and a `networkTransactionResponse`. The solution I found is that we can change the order of the `Given` methods and leave the token creation at last.  In the following method:

```
@Given("I ensure token {token} has been created")
    public void createNamedToken(TokenNameEnum tokenName) {
        var tokenAndResponse = tokenClient.getToken(tokenName);
        if (tokenAndResponse.response() != null) {
            this.networkTransactionResponse = tokenAndResponse.response();
            verifyMirrorTransactionsResponse(mirrorClient, 200);
        }
   }
```
We have the verifyMirrorTransactionsResponse in this method which would be executed after the contract creation. So in both cases -> running it individually or in parallel(@acceptance) we secure that we have networkTransactionResponse and all the transactions are verified.

hedera-mirror-test/src/test/resources/features/contract/estimate.feature
yiliev0: I guess we missed it from the begging. I've added it now and tested running the tests with the @acceptance tag. It seems all are working fine together.

hedera-mirror-test/src/test/resources/features/contract/estimatePrecompile.feature
yiliev0: I've changed the names of the tests so they are more descriptive. Thanks for suggestion!

hedera-mirror-test/src/test/resources/features/contract/estimatePrecompile.feature
yiliev0: I have moved the identical calls/tests that check the gas within the range at the bottom. So now we make only 1 call against the mirror node for the gas estimation, verify that it's correct and use it afterwards for the `executeContractTransaction`

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/EstimatePrecompileFeature.java
yiliev0: These updates are needed for the contract transaction execution when we attempt to do mints&transfers. After the v2 security changes, if we attempt to execute it without the account and token keys update, we get `NVALID_FULL_PREFIX_SIGNATURE_FOR_PRECOMPILE`.

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/EstimatePrecompileFeature.java
yiliev0: I've attempted initially to find a solution to combine these two, however, I couldn't find a way to make it happen. Тhe data that we encode with `headlong` and `ContractFunctionParamets` are the same but these two are working/constructed differently and not interchangeable. I looked for any util methods but I couldn't find any suitable any that would help us for our case.

steven-sheehy: You can use headlong to construct the data, then use `new ContractExecuteTransaction().setFunctionParameters(ByteString.copyFrom(data))`. `ContractClient.executeContract()` could be overloaded to pass `byte[]` instead of `ContractFunctionParameters`. But this isn't critical for this PR.

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/EstimatePrecompileFeature.java
```diff
@@ -1614,6 +1502,250 @@ public void exchangeRateTinyBarsToTinyCentsEstimateGas() {
...
+
+    public int validateAndReturnGas(String data, ContractMethods contractMethods, String contractAddress) {
+        var response = estimateContract(data, contractAddress);
+        var estimateGasValue = response.getResultAsNumber().intValue();
+        assertTrue(isWithinDeviation(contractMethods.getActualGas(), estimateGasValue, lowerDeviation, upperDeviation));
+        return estimateGasValue;
+    }
+
+    @Then("I call estimateGas with balanceOf function for {string} and verify the estimated gas against HAPI")
+    public void executeBalanceOfFunctionWithLimitedGas(String tokenName) {
```
steven-sheehy, line 1526: We should avoid manual conversion to enum.

```suggestion
    @Then("I call estimateGas with balanceOf function for {token} and verify the estimated gas against HAPI")
    public void executeBalanceOfFunctionWithLimitedGas(TokenNameEnum tokenName) {
```

yiliev0: Removed the manual conversion.

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/EstimatePrecompileFeature.java
yiliev0: Following your steps I've managed to do it. Thanks for the suggestions it was pretty helpful!

steven-sheehy (approved):
LGTM

Issue #7092, opened 2023-10-18
Update the Besu EVM library to version 23.10.0


Nana-EC:
### Problem

Besu recently released a new 23.10.0 version

### Solution

Update the Besu EVM library to version 23.10.0

PR #7117, opened 2023-10-25
Update the Besu EVM library version to 23.10.0

bilyana-gospodinova:
**Description**:
This PR updates the Besu EVM library version to 23.10.0 in order to be equivalent with hedera-services.

Fixes https://github.com/hashgraph/hedera-mirror-node/issues/7092

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -95,7 +95,6 @@ dependencies {
         api("org.mockito:mockito-inline:5.2.0")
         api("software.amazon.awssdk:bom:2.21.5")
         api("uk.org.webcompere:system-stubs-jupiter:2.1.3")
-        api("tech.pegasys:jc-kzg-4844:0.7.0")
     }
 }
 
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -83,7 +83,7 @@ dependencies {
         api("org.apache.tuweni:tuweni-bytes:2.3.1")
         api("org.apache.velocity:velocity-engine-core:2.3")
         api("org.gaul:s3proxy:2.1.0")
-        api("org.hyperledger.besu:secp256k1:0.6.1")
+        api("org.hyperledger.besu:secp256k1:0.8.0")
         api("org.hyperledger.besu:evm:23.10.0")
         api("org.jetbrains:annotations:24.0.1")
         api("org.mapstruct:mapstruct:$mapStructVersion")
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -57,7 +57,7 @@ dependencies {
         api("com.graphql-java-generator:graphql-java-client-runtime:2.3.1")
         api("com.graphql-java:graphql-java-extended-scalars:21.0")
         api("com.graphql-java:graphql-java-extended-validation:21.0")
-        api("com.hedera.evm:hedera-evm:0.39.0")
+        api("com.hedera.evm:hedera-evm:0.43.0")
         api("com.hedera.hashgraph:hedera-protobuf-java-api:0.43.0")
         api("com.hedera.hashgraph:sdk:2.29.0")
         api("com.ongres.scram:client:2.1")
@@ -84,7 +84,7 @@ dependencies {
         api("org.apache.velocity:velocity-engine-core:2.3")
         api("org.gaul:s3proxy:2.1.0")
         api("org.hyperledger.besu:secp256k1:0.6.1")
-        api("org.hyperledger.besu:evm:22.7.6")
+        api("org.hyperledger.besu:evm:23.10.0")
         api("org.jetbrains:annotations:24.0.1")
         api("org.mapstruct:mapstruct:$mapStructVersion")
         api("org.mapstruct:mapstruct-processor:$mapStructVersion")
@@ -95,6 +95,7 @@ dependencies {
         api("org.mockito:mockito-inline:5.2.0")
         api("software.amazon.awssdk:bom:2.21.5")
         api("uk.org.webcompere:system-stubs-jupiter:2.1.3")
+        api("tech.pegasys:jc-kzg-4844:0.7.0")
     }
 }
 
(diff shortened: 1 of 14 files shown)


lukelee-sl (approved):
LGTM

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/store/contract/HederaEvmStackedWorldStateUpdaterTest.java
```diff
@@ -167,7 +168,7 @@ void commitsDeletedAccountsAsExpected() {
 
     @Test
     void accountTests() {
-        updatedHederaEvmAccount.setBalance(Wei.of(100));
+        updatedHederaEvmAccount.setBalance(Wei.ONE);
```
Nana-EC, line 171: Nit: value was changed from 100 to 1.
Not sure if there's a possibility it might change behavior but want to make sure this is an intentional reduction in the value

hedera-mirror-common/build.gradle.kts
```diff
@@ -36,6 +40,7 @@ dependencies {
     api("org.apache.tuweni:tuweni-bytes")
     api("org.jetbrains:annotations")
     api("org.springframework.boot:spring-boot-starter-data-jpa")
+    api("tech.pegasys:jc-kzg-4844")
```
Nana-EC, line 43: Q: why is this needed here and in other places?

Nana-EC (changes requested):
Looking good.
2 clarifying points to confirm

build.gradle.kts
```diff
@@ -95,6 +95,7 @@ dependencies {
         api("org.mockito:mockito-inline:5.2.0")
         api("software.amazon.awssdk:bom:2.21.5")
         api("uk.org.webcompere:system-stubs-jupiter:2.1.3")
+        api("tech.pegasys:jc-kzg-4844:0.7.0")
```
lukelee-sl, line 98: Can you explain this dependency - looks like it should come in with the besu jar.

bilyana-gospodinova: There was an error for this dependency at compile time. It seems that only adding this repository is enough (maven { url = uri("https://artifacts.consensys.net/public/maven/maven/")}). I will remove the dependency in the next commit.

hedera-mirror-common/build.gradle.kts
bilyana-gospodinova: Answered in https://github.com/hashgraph/hedera-mirror-node/pull/7117#discussion_r1371819723.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/store/contract/HederaEvmStackedWorldStateUpdaterTest.java
bilyana-gospodinova: Thanks for pointing it out. The change in the value was not needed.

kselveliev (dismissed):
LGTM

hedera-mirror-rest-java/build.gradle.kts
```diff
@@ -23,6 +23,10 @@ plugins {
     id("spring-conventions")
 }
 
+repositories {
+    maven { url = uri("https://artifacts.consensys.net/public/maven/maven/") }
+}
+
```
steven-sheehy, line 29: Remove this repository everywhere except web3 gradle.

hedera-mirror-web3/build.gradle.kts
```diff
@@ -38,6 +40,7 @@ dependencies {
     implementation("org.springframework.cloud:spring-cloud-starter-bootstrap")
     implementation("org.springframework.cloud:spring-cloud-starter-kubernetes-fabric8-config")
     implementation("net.java.dev.jna:jna")
+    implementation("com.swirlds:swirlds-common")
```
steven-sheehy, line 43: Do we really need this separate dependency? Shouldn't hedera-evm depend upon it? There's no version for it in root build.gradle.kts so that implies it really is a transitive dependency.

steven-sheehy (changes requested):

build.gradle.kts
```diff
@@ -84,7 +84,7 @@ dependencies {
         api("org.apache.velocity:velocity-engine-core:2.3")
         api("org.gaul:s3proxy:2.1.0")
         api("org.hyperledger.besu:secp256k1:0.6.1")
```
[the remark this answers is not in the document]
bilyana-gospodinova: org.hyperledger.besu:secp256k1 is now updated to 0.8.0

hedera-mirror-rest-java/build.gradle.kts
bilyana-gospodinova: Done

hedera-mirror-web3/build.gradle.kts
bilyana-gospodinova: com.swirlds:swirlds-common is removed as a dependency from web3.

hedera-mirror-rest-java/build.gradle.kts
bilyana-gospodinova: If I remove it the build fails with:

Execution failed for task ':common:test'.
> Could not resolve all files for configuration ':common:testRuntimeClasspath'.
   > Could not find tech.pegasys:jc-kzg-4844:0.7.0.
     Required by:
         project :common > org.hyperledger.besu:evm:23.10.0

steven-sheehy: Ok, then please remove everywhere but add to https://github.com/hashgraph/hedera-mirror-node/blob/bc8f5a34129f0e6b513b625381606a362611f844/buildSrc/src/main/kotlin/java-conventions.gradle.kts#L38

IvanKavaldzhiev (dismissed):
LGTM

steven-sheehy (approved):
LGTM

Nana-EC (approved):
LGTM

PR #7118, opened 2023-10-25
HIP-584 Historical: Added new query for contract state to return historical contract state storage value

### LINKED ISSUE #6812: HIP-584: Extend needed DB queries to filter by timestamp

kselveliev:
Description:

This PR fixes part of the following [issue](https://github.com/hashgraph/hedera-mirror-node/issues/6812)

**Description**:
For historical support we need to be able to get historical snapshot of the contract state.
To do this we need to add new method to retrieve a historical contract state storage value based on a timestamp filter.

This PR modifies:
Added new method in **ContractStateRepository** to retrieve historical contract state storage value.
Added new unit tests in **ContractStateRepositoryTest** to test all the possible scenarios for the newly added repository method.

**Related issue(s)**:

Fixes #6812

**Checklist**

### DIFF
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/ContractStateRepository.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/ContractStateRepository.java
@@ -30,4 +30,44 @@ public interface ContractStateRepository extends CrudRepository<ContractState, L
     @Query(value = "select value from contract_state where contract_id = ?1 and slot =?2", nativeQuery = true)
     @Cacheable(cacheNames = CACHE_NAME, cacheManager = CACHE_MANAGER_CONTRACT_STATE, unless = "#result == null")
     Optional<byte[]> findStorage(final Long contractId, final byte[] key);
+
+
+    @Query(value =
+            """
+            select value
+            from (
+                (
+                    select
+                        contract_id,
+                        created_timestamp,
+                        modified_timestamp,
+                        slot,
+                        value
+                    from contract_state
+                    where contract_id = ?1
+                    and slot = ?2
+                    and created_timestamp <= ?3
+                    order by created_timestamp desc
+                    limit 1
+                )
+                union all
+                (
+                    select
+                        contract_id,
+                        consensus_timestamp as created_timestamp,
+                        consensus_timestamp as modified_timestamp,
+                        slot,
+                        value_written as value
+                    from contract_state_change
+                    where contract_id = ?1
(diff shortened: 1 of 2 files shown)


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/ContractStateRepository.java
```diff
@@ -30,4 +30,60 @@ public interface ContractStateRepository extends CrudRepository<ContractState, L
...
+                    from contract_state_change
+                    where contract_id = ?1
+                    and slot = ?2
+                    and consensus_timestamp <= ?3
+                    order by consensus_timestamp desc
+                    limit 1
+                )
+                order by created_timestamp desc
+                limit 1
+            ) as combined_result
```
xin-hedera, line 86: should just query `contract_state_change`, the union all will run both subqueries and then end up perhaps either throwing away one duplicate copy or no matching result found from `contract_state`

when querying contract_state_change, it should `coalesce(value_written, value_read)`  since value_written can be null, indicating the slot is read but not written to

xin-hedera (changes requested):

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/repository/ContractStateRepositoryTest.java
```diff
@@ -37,6 +38,30 @@ void findStorageSuccessfulCall() {
                 .isEqualTo(contractState.getValue());
     }
 
+    @Test
+    void findStorageOfContractStateChangeByBlockTimestampSuccessfulCall() {
```
xin-hedera, line 42: Need multiple rows contract_state_change for the same (contract_id, slot) so we know the method returns the correct value w.r.t. to the blockTimestamp.

also need a row where `value_written` is null to verify the method returns the value from `value_read`

xin-hedera (changes requested):

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/repository/ContractStateRepositoryTest.java
kselveliev: Comment addressed.

xin-hedera (approved):
LGTM

IvanKavaldzhiev (approved):
LGTM

steven-sheehy (approved):
LGTM

PR #7125, opened 2023-10-27
Fix acceptance tests race condition for EstimateFeature

### LINKED ISSUE #7124: Acceptance test: Fix race condition for estimateFeature
### Problem

Following the log, we find out that there's race condition when running all of the tests in parallel. In this case networkTransactionResponse is null because the token was taken from the mapper used by our creation singleton method and leaving us without networkTransactionResponse.

```
 java.lang.NullPointerException: Cannot invoke "com.hedera.mirror.test.e2e.acceptance.response.NetworkTransactionResponse.getTransactionIdStringNoCheckSum()" because "this.networkTransactionResponse" is null
	at com.hedera.mirror.test.e2e.acceptance.steps.AbstractFeature.verifyMirrorTransactionsResponse(AbstractFeature.java:91)
	at com.hedera.mirror.test.e2e.acceptance.steps.EstimateFeature.verifyMirrorAPIResponses(EstimateFeature.java:102)
	at ✽.the mirror node REST API should return status 200 for the estimate contract creation(classpath:features/contract/estimate.feature:7)
```

### Solution

Add a check that would verify if networkTransaction is null:
* If networkTransaction is null -> token is taken from the mapper, so we don't need to verify the mirror transaction response because the token already exisists.
* If networkTransaction is not null -> token was created within the test suite, so we need to verify the mirror transaction response.

### Alternatives

_No response_

yiliev0:
**Description**:  This PR addresses a race condition observed when tests are run concurrently, particularly affecting the handling of tokens and contracts. The race condition could lead to a scenario where a contract and/or token is already created (within another test suite) and loaded from the Map, which leads to networkTransactionResponse being null and throwing java.lang.NullPointerException.
To solve this:
* Added a null check for networkTransactionResponse before proceeding to verify the mirror transaction response.

**Related issue(s)**:

Fixes #7124

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/EstimateFeature.java
+++ b/hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/EstimateFeature.java
@@ -97,14 +97,21 @@ public class EstimateFeature extends AbstractEstimateFeature {
         receiverAccountId = accountClient.getAccount(AccountClient.AccountNameEnum.BOB);
     }
 
-    @Then("the mirror node REST API should return status {int} for the estimate contract creation")
-    public void verifyMirrorAPIResponses(int status) {
-        verifyMirrorTransactionsResponse(mirrorClient, status);
-    }
-
     @Given("I successfully create fungible token")
     public void createFungibleToken() {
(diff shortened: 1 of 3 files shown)


steven-sheehy (approved):
LGTM

PR #7146, opened 2023-11-01
HIP-584: Token transfer does not fail after dissociate

### LINKED ISSUE #7114: TransferToken does not fail after dissociate in a nested function
### Description

When calling a nested contract function which does token associate, token dissociate and transferToken we expect the transfer to fail as the token is no longer associated to the account. however, the transfer is successful. For NFT the transfer fails and the error is successfully caught with the correct status code, but for Fungible token it is not.

### Steps to reproduce

1. Create a contract nested function which has token dissociate and transferToken. 
2. Call this function and expect the transfer status to be failing.

The test can be found here https://github.com/hashgraph/hedera-mirror-node/pull/7085/files

### Additional context

This scenario is tested only with "transferToken" and "transferNFT". Will test with the rest of the transfer methods and update the description accordingly. 

### Hedera network

other

### Version

0.92.0-SNAPSHOT

### Operating system

None

kselveliev:
**Description**

With the introduction of `hasAssociation` flag in `TokenRelationship`:

After dissociate operation `hasAssociation` flag is set to `false` in the `TokenRelationship`.

After associate operation `hasAssociation` flag is set to `true` in the `TokenRelationship`.

However after the addition of `HederaTokenStore` in mirror-node during `sanityChecked`.

hedera-services checks → `if (!tokenRelsLedger.exists(tokenRelationShipKey)) { do validation   }`

mirror-node checks → `if (tokenRelationship.getAccount().getId().equals(Id.*DEFAULT*)) { do validation }`

In mirror node currently we do NOT check `hasAssociation` flag of the `TokenRelationship` and due to this the validation is skipped and we are not returning status `TOKEN_NOT_ASSOCIATED_TO_ACCOUNT` every time we expect to.

* `HederaTokenStore` `sanityChecked()`, `autoAssociate()` methods and modifies the if statements in them where we should check if token relationship has association.
* Modified token relationship associations in unit tests in `HederaTokenStoreTest`.

**Related issue(s)**:

Fixes #7114

**Checklist**


IvanKavaldzhiev (approved):
LGTM

hedera-mirror-web3/src/main/java/com/hedera/services/store/tokens/HederaTokenStore.java
```diff
@@ -148,7 +148,7 @@ public ResponseCodeEnum autoAssociate(AccountID aId, TokenID tId) {
             final var tokenRelationship =
                     store.getTokenRelationship(asTokenRelationshipKey(aId, tId), OnMissing.DONT_THROW);
 
-            if (!tokenRelationship.getAccount().getId().equals(Id.DEFAULT)) {
+            if (!tokenRelationship.getAccount().getId().equals(Id.DEFAULT) || tokenRelationship.hasAssociation()) {
```
steven-sheehy, line 151: Think this is the wrong approach. `store.getTokenRelationship()` is used in a number of places and this only fixes a few of its uses. To fix everywhere we should push this fix into the store. Consensus nodes do the following:

```java
    if (tokenRelsLedger.contains(Pair.of(aId, tId))) {
        return TOKEN_ALREADY_ASSOCIATED_TO_ACCOUNT;
    }
```

So inside their store they do not even have dissociated relationships at all. I think we should change the store to be similar so we only query associated relationships from the database and return an empty relationship for dissociated.

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/services/store/tokens/HederaTokenStore.java
0xivanov: In `TokenRelationshipDatabaseAccessor` we already only query associated relationships.
When we want to get dissociated one, the store will either return default/empty token rel object or throw an error (depending on OnMissing argument).

steven-sheehy: I see, so this is for scenarios where token relationship is associated in db, but dissociated in memory via some operation? Regardless, the point still stands that we should only return associated relationships from the store. Otherwise we probably have other areas that we'd have to check associate flag to match services logic.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/StoreImpl.java
```diff
@@ -173,6 +173,23 @@ public void updateTokenRelationship(final TokenRelationship updatedTokenRelation
         }
     }
 
+    @Override
+    public void deleteTokenRelationship(TokenRelationship tokenRelationship) {
+        final var topFrame = stackedStateFrames.top();
+        final var tokenRelationshipAccessor = topFrame.getAccessor(TokenRelationship.class);
+        try {
+            final var tokenRelationshipKey = keyFromRelationship(tokenRelationship);
+            final var tokenRel = tokenRelationshipAccessor.get(tokenRelationshipKey);
```
IvanKavaldzhiev, line 182: We should search also for a `TokenRelationshipKey` where the account address has an alias (if applicable). We can use - keyFromRelationshipWithAlias(persistedTokenRel);

If we find a record where the account has an alias - we should also delete it. Basically we can have 2 token relationships between a token and an account - one where the account has long zero address and one, where they have an alias.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/StoreImpl.java
```diff
@@ -173,6 +173,23 @@ public void updateTokenRelationship(final TokenRelationship updatedTokenRelation
         }
     }
 
+    @Override
+    public void deleteTokenRelationship(TokenRelationship tokenRelationship) {
```
steven-sheehy, line 177: Missing test coverage in StoreImplTest for all code paths.

hedera-mirror-web3/src/main/java/com/hedera/services/store/tokens/HederaTokenStore.java
kselveliev: After the above conversations I've removed the hasAssociation flag in the TokenRelationship model as it does not exist in services. Once we do dissociate, the relationship is marked as destroyed and we now delete the relationship from store. And if when we try to get token relationship from the store if it is deleted we simply return an empty relationship.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/StoreImpl.java
kselveliev: Comment addressed. I added a logic to also check and delete possible alias token relationship.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/StoreImpl.java
kselveliev: Commentd addressed. Added tests for the delete method in StoreImplTest.

steven-sheehy (approved):
LGTM

IvanKavaldzhiev (approved):
LGTM!

Issue #7123, opened 2023-10-27
eth_call returns 400 bad request for nested contract deploy


yiliev0:
### Description

Currently the `/api/v1/call` endpoint returns 400 bad request for `eth_call` with nested contract deploy via create or create2 operations.

### Steps to reproduce

1. Deploy the contract below
2. Attempt to call eth_call on function deployNestedContracts() or deployNestedContracts2()

**Expected result**: eth_call returns two addresses for the contracts
**Actual result:** eth_call returns 400 bad request.
 
   ```
     contract MockContract {
    function getAddress() public view returns (address) {
        return address(this);
    }

    function destroy() public {
        selfdestruct(payable(msg.sender));
    }
}
    contract EstimateGasContract is Caller {
    uint256 public salt = 1;
    uint256 public salt2 = 2;
    MockContract mockContract;
    
    function deployNestedContracts() public returns (address, address) {
        MockContract newContract1 = new MockContract();
        MockContract newContract2 = new MockContract();

        return (address(newContract1), address(newContract2));
    }

    function deployNestedContracts2() public returns (address, address) {
        MockContract newContract1 = new MockContract{salt: bytes32(salt)}();
        MockContract newContract2 = new MockContract{salt: bytes32(salt2)}();

        return (address(newContract1), address(newContract2));
    }
```

### Hedera network

other

### Version

0.92.0

### Operating system

None

PR #7148, opened 2023-11-01
HIP-584: Fix nested contract deploy fail

bilyana-gospodinova:
**Description**:
When deploying nested contracts after the Besu upgrade to 23.10.0 (and hedera-evm upgrade to 0.43.0) the target addresses of the created contracts are now created as address aliases based on the sender address and the account nonce. The account nonce was returned always as 0 from the default getNonce(Address) method from HederaEvmEntityAccess and this caused collisions in the generated aliases for the created contracts. This method is now overriden in MirrorEntityAccess to return the correct nonce value from the DB.

Fixes https://github.com/hashgraph/hedera-mirror-node/issues/7123


steven-sheehy (approved):
LGTM

IvanKavaldzhiev (approved):
LGTM

Issue #7122, opened 2023-10-27
HIP-584: Blockhash() returns empty hash for latest block number


IvanKavaldzhiev:
### Description

Currently when we use the internal solidity `blockhash()` function with `block.number` we get empty hash. That's not correct, since mirror-nodes have the block hashes for all blocks that were imported in the DB including the latest one.

### Steps to reproduce

1. Execute `blockhash(block.number)` on any environment
2. Observe that empty hash is returned

### Additional context

The logic inside besu's BlockHashOperation ignores latest block from mirror node's DB and we default to empty hash. Probably we should adapt the logic for getting block number by incrementing it with 1, so that it matches consensus node behaviour.

### Hedera network

previewnet

### Version

v0.92.0

### Operating system

None

PR #7156, opened 2023-11-02
HIP-584: Update the blockhash operation to return value for all blocks

bilyana-gospodinova:
**Description**:
In this PR a custom implementation of the blockhash operation is added. With this change we can get the block hash value of any block up to the latest one.

Fixes https://github.com/hashgraph/hedera-mirror-node/issues/7122


hedera-mirror-web3/src/main/java/com/hedera/services/evm/contracts/operations/HederaBlockHashOperation.java
```diff
@@ -0,0 +1,62 @@
...
+import org.hyperledger.besu.evm.operation.BlockHashOperation;
+import org.hyperledger.besu.evm.operation.Operation;
+
+/**
+ * Custom version of the Besu's BlockHashOperation class. The difference is
+ * that in the mirror node we have the block hash values of all the blocks
+ * so the restriction for the latest 256 blocks is removed. The latest
+ * block value can be returned as well.
+ */
+public class HederaBlockHashOperation extends BlockHashOperation {
```
IvanKavaldzhiev, line 35: Since this opcode contains mirror node specific logic, which we would want to keep, we should move this class into `com.hedera.mirror.web3.evm.contracts.operations` package.

After we start deleting the logic from `com.hedera.services` and replacing it with the modularized code from hedera-services repo, we should investigate. the option of keeping this logic. If we don't find a way to add this custom opcode to the EVM configuration, which most probably would be moved inside the new `hedera-smart-contracts` service, we may need to find a different fix for the BLOCKHASH issue, but for the moment, this fix would be sufficient.

hedera-mirror-web3/src/main/java/com/hedera/services/evm/contracts/operations/HederaBlockHashOperation.java
```diff
@@ -0,0 +1,62 @@
...
+        // Short-circuit if value is unreasonably large
+        if (blockArg.size() > 8) {
+            frame.pushStackItem(UInt256.ZERO);
+            return successResponse;
+        }
+
+        final long soughtBlock = blockArg.toLong();
+        final Function<Long, Hash> blockHashLookup = frame.getBlockHashLookup();
+        final Hash blockHash = blockHashLookup.apply(soughtBlock);
+        frame.pushStackItem(blockHash);
```
steven-sheehy, line 58: The old logic would return zero if the sought block was greater than latest or negative. This new logic will throw an error. There's also the question does it throw error or return zero if positive & less than latest but not found. I think we need to add back some bounds checking to match the previous behavior.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallEvmCodesTest.java
```diff
@@ -161,21 +161,26 @@ void getBlockHashReturnsCorrectHash() {
...
         final var serviceParameters = serviceParametersForEvmCodes(functionHash);
 
         assertThat(contractCallService.processCall(serviceParameters))
-                .isEqualTo("0x0000000000000000000000000000000000000000000000000000000000000000");
+                .isEqualTo("0x" + genesisRecordFileForBlockHash.getHash().substring(0, 64));
     }
 
     @Test
-    void getLatestBlockHashReturnsZero() {
+    void getLatestBlockHashIsNotEmpty() {
```
steven-sheehy, line 178: We should have some more tests for negative, beyond latest, below min, etc.

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/services/evm/contracts/operations/HederaBlockHashOperation.java
bilyana-gospodinova: The class was moved to the suggested package.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallEvmCodesTest.java
bilyana-gospodinova: Added test for values after latest block.

hedera-mirror-web3/src/main/java/com/hedera/services/evm/contracts/operations/HederaBlockHashOperation.java
bilyana-gospodinova: I added some of the previous if-statements. We can't really have negative block values as in Solidity the blockhash function accepts unsigned values. I have added an if-statement for this case anyway and the result will be zero.

In case the block value is positive and less than latest but not found, an error will be thrown and this was the same in the previous logic. Can there be such case at all (other than in the integration tests)?

steven-sheehy: > In case the block value is positive and less than latest but not found, an error will be thrown and this was the same in the previous logic. Can there be such case at all (other than in the integration tests)?

It's a valid scenario for partial mirror nodes who start after block zero by setting a startDate (or leaving it unset to default to now). Since the previous and current logic throw an error for block not found perhaps its fine as it is now.

steven-sheehy (dismissed):
LGTM. One code smell plus workflows are stuck. Maybe fix the code smell and that should re-trigger the workflows.

IvanKavaldzhiev (dismissed):
LGTM

kselveliev (approved):
LGTM

PR #7177, opened 2023-11-07
HIP-584: Make `MirrorEvmTxProcessor` singleton

### LINKED ISSUE #6612: HIP-584: Reduce memory footprint by making state asynchronous
### **Motivation**

Ongoing support of `eth_call` and `eth_estimateGas` executions would increase the number of users that start using these new functionalities. Eventually this would increase the number of requests against the Archive Node and performance issues might be faced. This epic would aim to reduce the memory footprint and the number of newly created objects by making the in-memory state agnostic of different requests.

### **Implementation details**

- [x] #6619 
- [x] #6618 
- [ ] #6617 

### **Additional bug fixes**
- [x] #6620

### LINKED ISSUE #6617: HIP-584: Analyze thread and heap dump with jmeter
### Problem

We need to make sure that a transaction execution in the Archive Node uses only a single thread in order to perform refactoring of the in-memory state, using ThreadLocal.

Additionally we should inspect what is the heap dump of the application before and after the refactoring to see how much the memory footprint has been reduced.

Expecting the thread dump would help us analyze how the web3 is handling threads and ThreadLocal instances and whether we have memory leaks or threads that continue work in the background.

### Solution

Setup a `jmeter` against the web3 module and using K6 tests load the system with requests, so that we can analyze its memory and thread management performance.

### Alternatives

_No response_

IvanKavaldzhiev:
**Description**:

It seems we can save additional memory footprint by making `MirrorEvmTxProcessor` bean to be Singleton. 

In the current code base we have defined `MirrorEvmTxProcessor` to be with factory scope and have a new instance for each new transaction. Analysis of heap dumps during K6 tests with multiple VUs showed that currently we have 2 dominator objects, which are `ContractCallContext` and `MirrorEvmTxProcessor`. They both retain memory size for the same classes and we have duplication of unnecessary classes in memory.

By making  `MirrorEvmTxProcessor` singleton we have one single instance for all requests and we remove the duplication of these objects. To make this bean singleton, we should also make the `HTSPrecompiledContract` to be singleton bean and extract the stateful fields into `ContractCallContext`. So we have the side effect of increasing the thread local context. In addition `buildInitialFrame` and `execute` methods inside `MirrorEvmTxProcessor` needed to be make `synchronized`, so that each transaction executes on top of its own `initial message frame`. Without `synchronized` we get mixed behaviour and one transaction could use the initial frame of another thread.

In theory this PR should reduce the retained memory size at least in half.

**Related issue(s)**:

Fixes #6612 
Fixes #6617

**Notes for reviewer**:

Metrics with current code base (factory bean definition):

factory_bean_dominator_classes_2
factory_bean_dominator_classes
factory_bean_memory_retained_size

Metrics after new changes (singleton bean definition):

singleton_bean_dominator_classes_2
singleton_bean_dominator_classes
singleton_bean_memory_retained_size

**Checklist**


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/common/ContractCallContext.java
```diff
@@ -68,6 +70,22 @@ public class ContractCallContext implements AutoCloseable {
...
+    @Setter
+    private long gasRequirement = 0L;
+
+    /** HTS Precompile field keeping the transactionBody needed for a given precompile execution */
+    @Setter
+    private TransactionBody.Builder transactionBody;
+
+    /** HTS Precompile field keeping the sender address of the account that initiated a given precompile execution */
+    @Setter
+    private Address senderAddress;
```
bilyana-gospodinova, line 87: ContractCallContext has a reset() method that is used in the BinaryGasEstimator. Don't we need to reset these new fields in this method as well?

IvanKavaldzhiev: I have already added a reset for these fields there.

bilyana-gospodinova: Oh, my bad. I was looking at the wrong method.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmTxProcessorImpl.java
```diff
@@ -71,7 +71,7 @@ public MirrorEvmTxProcessorImpl(
         this.store = store;
     }
 
-    public HederaEvmTransactionProcessingResult execute(
+    public synchronized HederaEvmTransactionProcessingResult execute(
```
bilyana-gospodinova, line 74: Wouldn't this slow down the performance significantly? This way we will basically process only 1 transaction at a time.

IvanKavaldzhiev: Having the `MirrorEvmTxProcessor` to be singleton, we save a lot of memory footprint but sacrifice some of the TPS handling. My assumption, though, is that while some threads may be blocked to execute the transaction via `execute` method, others may read data from the DB and prepare their own `ContractCallContext` to be ready for the execution, so executions of transactions are not totally blocked and threads have some other tasks to do. Most of transactions are executed below `200ms` I believe, but we can think of internal mechanism inside `MirrorEvmTxProcessor`, that saves us from very heavy malicious transactions and throw exception if certain time period is reached 🤔 

We should decide which factor is more blocking for us - increased memory retained size and web3 K8 pods starting to scale a lot or safe memory size and sacrifice some of the TPS. If we don't have user complaints for getting a lot of HTTP 429 requests, probably its safe to go with the second approach, introduced in this PR.

kselveliev (dismissed):
LGTM

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmTxProcessorImpl.java
IvanKavaldzhiev: Based on K6 tests and metrics monitored in VisualVM it seems with this PR we have relatively 4 times quicker requests and a lot less CPU spikes and overall CPU consumption, making it possible to even increase our rate limit value.

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/config/EvmConfiguration.java
```diff
@@ -147,43 +179,134 @@ CacheManager cacheManagerRecordFileEarliest() {
...
+            final EVM evm,
+            final PrecompileContractRegistry precompileContractRegistry,
+            final MirrorHTSPrecompiledContract mirrorHTSPrecompiledContract) {
+        final Map<String, PrecompiledContract> hederaPrecompiles = new HashMap<>();
+        hederaPrecompiles.put(EVM_HTS_PRECOMPILED_CONTRACT_ADDRESS, mirrorHTSPrecompiledContract);
+        hederaPrecompiles.put(PRNG_PRECOMPILE_ADDRESS, prngSystemPrecompiledContract);
+        hederaPrecompiles.put(
+                EXCHANGE_RATE_SYSTEM_CONTRACT_ADDRESS,
+                new ExchangeRatePrecompiledContract(
+                        gasCalculator, basicHbarCentExchange, mirrorNodeEvmProperties, Instant.now()));
```
steven-sheehy, line 297: Move this precompile map to be a separate bean then remove `MirrorEvmMessageCallProcessor` bean and used `@Named`.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/config/EvmConfiguration.java
```diff
@@ -147,43 +179,134 @@ CacheManager cacheManagerRecordFileEarliest() {
...
+                operationRegistry,
+                gasCalculator,
+                mirrorNodeEvmProperties.chainIdBytes32().toBigInteger());
+        Set.of(
+                        new HederaBalanceOperation(gasCalculator, validator),
+                        new HederaDelegateCallOperation(gasCalculator, validator),
+                        new HederaEvmChainIdOperation(gasCalculator, mirrorNodeEvmProperties),
+                        new HederaEvmCreate2Operation(
+                                gasCalculator, mirrorNodeEvmProperties, createOperationExternalizer),
+                        new HederaEvmCreateOperation(gasCalculator, createOperationExternalizer),
```
steven-sheehy, line 257: nit: Can avoid the extra param. Spring is smart enough to only create one.

```suggestion
                                gasCalculator, mirrorNodeEvmProperties, createOperationExternalizer()),
                        new HederaEvmCreateOperation(gasCalculator, createOperationExternalizer()),
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/config/EvmConfiguration.java
```diff
@@ -147,43 +179,134 @@ CacheManager cacheManagerRecordFileEarliest() {
...
+            final MessageCallProcessor messageCallProcessor,
+            final MirrorEvmMessageCallProcessor mirrorEvmMessageCallProcessor,
+            final PrecompileContractRegistry precompileContractRegistry) {
+        final Map<String, Provider<ContractCreationProcessor>> contractCreationProcessors = Map.of(
+                EVM_VERSION_0_30, () -> contractCreationProcessor, EVM_VERSION_0_34, () -> contractCreationProcessor);
+
+        MainnetPrecompiledContracts.populateForIstanbul(precompileContractRegistry, gasCalculator);
+
+        final Map<String, Provider<MessageCallProcessor>> messageCallProcessors = Map.of(
+                EVM_VERSION_0_30, () -> messageCallProcessor, EVM_VERSION_0_34, () -> mirrorEvmMessageCallProcessor);
```
steven-sheehy, line 207: Move these maps to separate beans, then you can make `MirrorEvmTxProcessor` `@Named`.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/config/ServicesConfiguration.java
```diff
@@ -691,4 +698,39 @@ TokenUpdateKeysPrecompile tokenUpdateKeysPrecompile(
...
+    }
+
+    @Bean
+    EvmEncodingFacade evmEncodingFacade() {
+        return new EvmEncodingFacade();
+    }
+
+    @Bean
+    EvmInfrastructureFactory evmInfrastructureFactory(final EvmEncodingFacade evmEncodingFacade) {
+        return new EvmInfrastructureFactory(evmEncodingFacade);
```
steven-sheehy, line 719:
```suggestion
    EvmInfrastructureFactory evmInfrastructureFactory() {
        return new EvmInfrastructureFactory(evmEncodingFacade());
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmTxProcessorImpl.java
steven-sheehy, line 1: We need to make `HederaEvmTxProcessor` stateless for this to be singleton. Luckily, we don't use many of the stateful fields:

* `coinbase`: Is a property so can move to constructor
* Move `tracer` to a final field populated in constructor
* Move `messageCallProcessor` and `contractCreationProcessor` assignment into constructor and make final since we don't seem to support dynamic evm version at runtime
* Push `intrinsicGas` into `execute()` and calculate it there
* Push `gasUsed` into local variable since it's only used within execute()
* Push `initialFrame` into local variable since it's only used within execute()
* Push `sbhRefund` into local variable since it's only used within execute()
* Remove unused setWorldState() and make field final
* Remove unused setBlockMetaSource() and make field final

steven-sheehy (changes requested):

PR #7178, opened 2023-11-07
HIP-584: Account balance cache delays is reflecting to proper eth_call balance update

### LINKED ISSUE #7043: Account balance cache delays is reflecting to proper eth_call balance update
### Description

We've identified an issue where the caching mechanism for an account balance (in HBARS) is causing a delay in reflecting real-time balance updates. When performing an `eth_call` operation to retrieve an account balance, followed by a transaction to change the account balance, and then another `eth_call` operation to retrieve the updated balance, the latter eth_call does not return the updated balance immediately. It retains the old balance from the cache for about 30 seconds before updating to reflect the new balance.

### Steps to reproduce

1. Perform an `eth_call` operation to an account to retrieve the current balance.
2. Send funds to the account to change its balance.
3. Perform another `eth_call` operation to the account to retrieve the updated balance.

### Additional context

Upon investigation, it appears that the caching mechanism triggered during the first `eth_call` operation is holding onto the old balance value for a duration of 30 seconds. This delay in cache update is leading to incorrect balance being returned in the subsequent `eth_call` operation post the account balance change.

### Hedera network

other

### Version

0.89.0-rc3

### Operating system

None

0xivanov:
**Description**:
We've identified an issue where the caching mechanism for an account balance (in HBARS) is causing a delay in reflecting real-time balance updates. When performing an eth_call operation to retrieve an account balance, followed by a transaction to change the account balance, and then another eth_call operation to retrieve the updated balance, the latter eth_call does not return the updated balance immediately. It retains the old balance from the cache for about 30 seconds before updating to reflect the new balance.

Steps to reproduce: 

1. Perform an eth_call operation to an account to retrieve the current balance.
2. Send funds to the account to change its balance and wait for the transaction to be received in the mirror node.
3. Perform another eth_call operation to the account to retrieve the updated balance.

Additional context:
Upon investigation, it appears that the caching mechanism triggered during the first eth_call operation is holding onto the old balance value for a duration of 30 seconds. This delay in cache update is leading to incorrect balance being returned in the subsequent eth_call operation post the account balance change.

**Related issue(s)**:

Fixes #7043

**Notes for reviewer**:

**Checklist**


hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/CallFeature.java
```diff
@@ -100,7 +102,7 @@ public void createNewPrecompileTestContract() throws IOException {
     public void createNewEstimateTestContract() throws IOException {
         deployedContract = getContract(ESTIMATE_GAS);
         estimateContractAddress = deployedContract.contractId().toSolidityAddress();
-        receiverAccountId = accountClient.getAccount(AccountNameEnum.BOB);
+        receiverAccountId = accountClient.getAccount(AccountNameEnum.ALICE);
```
0xivanov, line 105: Accounts that have `ECDSA_SECP256K1` like BOB fail the `accountClient.sendCryptoTransfer` transaction, so i use ALICE here. Not sure why that happens though.

yiliev0: This is because BOB account has receiverSigRequired = true and we need to pass his privateKey similar to the tokenClient `transferFungibleToken` function. Maybe you can try modifying the `sendCryptoTransfer` to accept privateKey as well or we can keep it this way if it's too much effort.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/properties/CacheProperties.java
```diff
@@ -26,15 +26,20 @@
...
 public class CacheProperties {
 
+    /**
+     *  entity, contract and token models are used mostly simultaneously, so we use the same configuration for consistent reads
+     */
+    private static final String ENTITY_CACHE_CONFIG = "expireAfterWrite=1s,maximumSize=10000,recordStats";
+
     @NotBlank
-    private String contractState = "expireAfterWrite=1s,maximumSize=10000,recordStats";
+    private String contractState = ENTITY_CACHE_CONFIG;
```
steven-sheehy, line 35: Need to update default in `docs/configuration.md`.

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/CallFeature.java
steven-sheehy: We can also probably just remove receiverSigRequired=true from BOB if it's easier. Setting it doesn't really test anything related to mirror node functionality.

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/CallFeature.java
```diff
@@ -272,6 +274,20 @@ public void ethCallNestedDeployViaCreate2Function() {
         validateAddresses(addresses);
     }
 
+    @Then("I successfully update the balance of an account and get the updated balance")
+    public void getBalance() {
+        var data = encodeData(
+                ESTIMATE_GAS,
+                ADDRESS_BALANCE,
+                asAddress(receiverAccountId.getAccountId().toSolidityAddress()));
```
steven-sheehy, line 282: nit: Move to separate line so `encodeData` line can be on a single line.

steven-sheehy (changes requested):

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/CallFeature.java
yiliev0: @steven-sheehy In my opinion it would be better if we keep BOB for now with receiverSigRequired as he might come in handy in our next in-equivalence tests that are coming up to the mirror node. Secondly, if we decide to get rid of receiverSigRequired we might need to refactor all the places where we use BOB and we do transfers. 
The approach that @0xivanov decided to take(using ALICE) is fine imo.

0xivanov: Implementing the setting of private key in the transaction is also valid solution...

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/CallFeature.java
0xivanov: extracted `asAddress(receiverAccountId.getAccountId().toSolidityAddress())` to a variable

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/properties/CacheProperties.java
0xivanov: Updated

steven-sheehy (approved):
LGTM

PR #7180, opened 2023-11-07
Support local files grouped by day

### LINKED ISSUE #7075: Support local files grouped by day
### Problem

The `LocalStreamFileProvider` is not performant when the directory it's reading files contains millions of files due to filesystems are inherently slow sorting large numbers of files.

### Solution

* Add a `hedera.mirror.importer.downloader.groupByDay=false` property
* If property is true, search for files in a sub-folder with the last file's date
* If property is true and no files in the directory with the last file's date, search the next day's folder
* If property is false, use the parent directory

### Alternatives

_No response_

steven-sheehy:
**Description**:

* Add a `hedera.mirror.importer.downloader.local.deleteAfterProcessing` property to allow keeping local files
* Add support for writing sidecar files to disk
* Change write files properties to write to `streams` sub-directory
* Fix null handling when no files in directory
* Support local files grouped by day

**Related issue(s)**:

Fixes #7075

**Notes for reviewer**:

**Checklist**


hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/provider/LocalStreamFileProvider.java
```diff
@@ -42,80 +41,94 @@
...
+                                        String.valueOf(node.getNodeId()),
+                                        streamType.getNodeIdBasedSuffix());
+                            };
+
+                    return basePath.resolve(prefix).toFile();
+                })
+                .doOnNext(f -> log.debug("Listing files in {}", f))
+                .filter(File::exists)
+                .flatMapSequential(dir -> Flux.fromArray(
+                        requireNonNullElse(dir.listFiles(f -> matches(streamFilename.getFilenameAfter(), f)), EMPTY)));
```
edwin-greene, line 102: It might be ok to remove the `requireNonNullElse` and the `EMPTY` constant. If they are removed the tests still pass. It looks like the `filter(File::exists)` is removing invalid values from reaching the next step, so that that require and constant are not needed.

steven-sheehy: I ran into this error during manual testing with `filter(File::exists)` in place. `filter(File::exists)` only filters out missing directories. The directory could exist but have an I/O error and return null, per javadocs.

edwin-greene (approved):
Looks good!

jnels124 (approved):
LGTM

jascks (approved):
Looks good!

mgoelswirlds (approved):

PR #7181, opened 2023-11-07
Time partition and deduplicate existing balance data in v1

### LINKED ISSUE #7005: Reduce balance snapshot interval for older data
### Problem

Balance information makes up about half of the database on mainnet and with the growing number of accounts we won't be able to retain every 15 minute snapshot indefinitely.

### Solution

* Rename `account_balance` to `account_balance_old`
* Rename `token_balance` to `token_balance_old`
* Create a new `account_balance` table with the same columns but primary key reversed to `account_id, consensus_timestamp`
* Create a new `token_balance` table with the same columns but primary key reversed to `account_id, token_id, consensus_timestamp`
* Copy last 24 hours starting from last snapshot consensus timestamp worth of balances from old to new synchronously
* Write an async migration that copies from old to new in chunks starting with newer data first
  * For the first week, keep all snapshots
  * Afterwards, keep only one snapshot per day
* Drop `account_balance_old` and `token_balance_old`

### Alternatives

_No response_

xin-hedera:
**Description**:

This PR time partitions `account_balance` and `token_balance`, and dedups the existing balance data in v1

- Add a versioned SQL migration for helper functions to dedup balance data
- Add a versioned SQL migration to time partition and dedup the last 6-hour account_balance and token_balance data
- Add a repeatable SQL migration for partition maintenance procedures in v1
- Add an async java migration to backfill and deduplicate remaining balance data
- Remove `deduplicate` from `HistoricalBalanceProperties`
- Fix `(Account|Token)BalanceRepository.balanceSnapshot` to also include balance info deleted right at or before a full snapshot
- Fix flyway placeholder `partitionStartDate` to use absolute date
- Change web3 `TokenBalanceRepository` to use `account_balance` instead of `account_balance_file`
- Change rosetta to use deduped account_balance data
 
**Related issue(s)**:

Fixes #7005

**Notes for reviewer**:

**Checklist**


hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/db/PartitionMaintenance.java
```diff
@@ -42,10 +40,10 @@ public class PartitionMaintenance {
     @Owner
     private final JdbcTemplate jdbcTemplate;
 
-    @Scheduled(cron = "${hedera.mirror.importer.db.maintenance.cron:0 0 0 * * ?}")
-    @EventListener(ApplicationReadyEvent.class)
-    @Retryable
     @Leader
+    @Retryable
+    @Scheduled(cron = "${hedera.mirror.importer.db.maintenance.cron:0 0 0 * * ?}")
+    @Scheduled(initialDelay = 60, fixedDelay = Long.MAX_VALUE, timeUnit = TimeUnit.SECONDS)
```
xin-hedera, line 46: The problem of `@EventListener(ApplicationReadyEvent.class)` is the integration test for V1.89.0 migration will fail when `partitionMaintenance.runMaintenance` is triggered synchronously by application ready event in the spring test context and the SQL procedure raises exceptions since the tables are yet to be partitioned, the spring application fails to start.

Making it `@Async` while keeping `@EventListener(ApplicationReadyEvent.class)` can avoid the exception bubbling up to the main process, however makes testing more difficult.

Perhaps we can just remove `@EventListener(ApplicationReadyEvent.class)` and the new `@Scheduled` workaround above

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/BackfillAndDeduplicateBalanceMigration.java
```diff
@@ -0,0 +1,266 @@
...
+@Named
+@Profile("!v2")
+public class BackfillAndDeduplicateBalanceMigration extends AsyncJavaMigration<Long> {
+
+    private static final String ACCOUNT_BALANCE_TABLE_NAME = "account_balance";
+    private static final long EPOCH = 0L;
+
+    // Can't use spring data repositories because the migration runs with a different role, so it can drop the tables
+    // at the end
+    // To simplify the test case, the create_(deduped|full)_*_balance_snapshot functions are not dropped at the end
```
xin-hedera, line 53: If we want to drop these no longer needed SQL functions and make the integration test not so complicated, we can split V1.89.0 to two sql migrations

- the first creates the `create*` functions
- the second creates other functions, partitions and deduplicates the balance tables

Then in the integration test, we just need to reapply the first SQL migration in the after each hook

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/BackfillAndDeduplicateBalanceMigration.java
```diff
@@ -0,0 +1,266 @@
...
+
+    private static final String SELECT_INITIAL_CONSENSUS_TIMESTAMP_SQL =
+            """
+            select coalesce(max(consensus_timestamp), 0)
+            from account_balance
+            where account_id = 2 and consensus_timestamp < ?
+            """;
+
+    private static final String SELECT_LAST_CONSENSUS_TIMESTAMP_SQL =
+            "select balance from account_balance_old where consensus_timestamp = -1 and account_id = -1";
```
xin-hedera, line 137: the balance column of the row (account_id, consensus_timestamp) = (-1, -1) keeps the consensus timestamp of the first snapshot processed by migration V1.89.0

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/repository/AccountBalanceFileRepository.java
```diff
@@ -18,11 +18,9 @@
 
 import com.hedera.mirror.common.domain.balance.AccountBalanceFile;
 import java.util.Optional;
-import org.springframework.data.jpa.repository.Modifying;
 import org.springframework.data.jpa.repository.Query;
 
-public interface AccountBalanceFileRepository
-        extends StreamFileRepository<AccountBalanceFile, Long>, RetentionRepository {
+public interface AccountBalanceFileRepository extends StreamFileRepository<AccountBalanceFile, Long> {
```
xin-hedera, line 23: when `account_balance` and `token_balance` are partitioned, we should drop partitions to remove old data. I can revert the changes to these repositories to make the PR simpler.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/repository/AccountBalanceRepository.java
```diff
@@ -37,7 +37,12 @@ public interface AccountBalanceRepository
         insert into account_balance (account_id, balance, consensus_timestamp)
         select id, balance, :consensusTimestamp
         from entity
-        where deleted is not true and balance is not null
+        where balance is not null and
+          (deleted is not true or balance_timestamp > coalesce((
+              select max(consensus_timestamp)
+              from account_balance
+              where account_id = 2 and consensus_timestamp > :consensusTimestamp - 2592000000000000
```
xin-hedera, line 44: 2592000000000000 is 30 days

hedera-mirror-importer/src/main/resources/application.yml
```diff
@@ -119,7 +119,7 @@ spring:
       idPartitionSize: 1000000
       maxEntityId: 5000000
       maxEntityIdRatio: 2.0
-      partitionStartDate: "'4 years'"
+      partitionStartDate: "'2019-09-01'"
```
xin-hedera, line 122: default value for mainnet

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/record/historicalbalance/HistoricalBalanceServiceIntegrationTest.java
```diff
@@ -47,18 +46,19 @@
...
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
-import org.junit.jupiter.params.provider.CsvSource;
+import org.junit.jupiter.params.provider.ValueSource;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.ApplicationEventPublisher;
 import org.springframework.transaction.support.TransactionTemplate;
 import org.testcontainers.shaded.org.apache.commons.lang3.tuple.Pair;
 
+@Disabled("Disabled WIP")
```
xin-hedera, line 61: all test cases fail in this class, it's quite difficult to understand the test cases itself so have it disabled. will fix and re-enable it

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/BackfillAndDeduplicateBalanceMigration.java
```diff
@@ -0,0 +1,266 @@
...
+
+        return Optional.of(timestamp);
+    }
+
+    private Long getBalanceTimestamp(Long last) {
+        if (last == NO_BALANCE) {
+            return null;
+        }
+
+        long lowerBound = last + 4 * ONE_HOUR_IN_NS;
```
xin-hedera, line 235: will replace `4 * ONE_HOUR_IN_NS` with `MIN_INTERVAL`, looks like I made the change then lost it

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/repository/AccountBalanceFileRepository.java
xin-hedera: update, actually we have to make them not implement `RetentionRepository`

problem is if the repo is pruned with a non partition boundary timestamp, account's balance info can get lost since it's deduped

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.89.0__add_balance_deduplicate_functions.sql
```diff
@@ -0,0 +1,126 @@
...
+    from account_balance_old
+    where consensus_timestamp = balance_timestamp
+  )
+  insert into account_balance (account_id, balance, consensus_timestamp)
+  select coalesce(c.account_id, p.account_id), coalesce(c.balance, 0), balance_timestamp
+  from current as c
+  full outer join previous as p using (account_id);
+
+  get diagnostics count = row_count;
+  return count;
```
xin-hedera, line 27: return the number of inserted rows, useful info in the async Java migration

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/BackfillAndDeduplicateBalanceMigration.java
```diff
@@ -0,0 +1,274 @@
...
+            """
+            drop function if exists create_full_account_balance_snapshot(bigint, bigint);
+            drop function if exists create_deduped_account_balance_snapshot(bigint, bigint);
+            drop function if exists create_full_token_balance_snapshot(bigint, bigint);
+            drop function if exists create_deduped_token_balance_snapshot(bigint, bigint);
+
+            truncate account_balance_old;
+            truncate token_balance_old;
+            drop table account_balance_old;
+            drop table token_balance_old;
```
xin-hedera, line 63: truncating the tables took 35 minutes while directly dropping them took 1 hour

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/BackfillAndDeduplicateBalanceMigration.java
```diff
@@ -0,0 +1,274 @@
...
+    @Override
+    protected MigrationVersion getMinimumVersion() {
+        return MigrationVersion.fromVersion("1.89.1"); // The version balance table partitions are created
+    }
+
+    @Nonnull
+    @Override
+    protected Optional<Long> migratePartial(Long last) {
+        var stopwatch = Stopwatch.createStarted();
+        jdbcTemplate.execute("set local work_mem = '512MB'"); // Use higher work_mem to avoid temp files
```
xin-hedera, line 190: since it's `set local`, the 512MB work_mem is only effective in the current db transaction

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/repository/TokenBalanceRepository.java
```diff
@@ -34,7 +34,13 @@ public interface TokenBalanceRepository
         insert into token_balance (account_id, balance, consensus_timestamp, token_id)
         select account_id, balance, :consensusTimestamp, token_id
         from token_account
-        where associated is true
+        where associated is true or balance_timestamp > (
+            select coalesce(max(consensus_timestamp), 0)
+            from account_balance
+            where account_id = 2 and
+              consensus_timestamp > :consensusTimestamp - 2592000000000000 and
+              consensus_timestamp < :consensusTimestamp
```
xin-hedera, line 42: unlike `AccountBalanceRepository.balanceSnapshot`, there is an additional `consensus_timestamp < :consensusTimestamp` condition, this is to handle the corner case that account_balance snapshot at `:consensusTimestamp` is generated before token_balance snapshot

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/record/historicalbalance/HistoricalBalanceServiceIntegrationTest.java
```diff
@@ -76,16 +74,25 @@ class HistoricalBalanceServiceIntegrationTest extends IntegrationTest {
...
-    private List<TokenAccount> tokenAccounts = new ArrayList<>();
+    private List<Entity> entities;
+    private List<TokenAccount> tokenAccounts;
 
     void setup() {
         // common database setup
-        var balanceTimestamp = domainBuilder.timestamp() + Durations.ONE_MINUTE.toNanos();
-        var treasuryAccount =
-                domainBuilder.entity().customize(e -> e.id(2L).num(2L)).persist();
+        long now = DomainUtils.convertToNanosMax(Instant.now());
```
xin-hedera, line 87: main changes in the integration test

- revert the overall test flow to the state before deduplication is implemented
- change `verifyGeneratedBalances` to reduce complexity, let the test case main logic pass in list of updated entities and token accounts which should have new balance info added to the snapshot
- remove conditional testing of dedup, since now it's always deduped and time partitioned

hedera-mirror-importer/src/main/resources/application.yml
xin-hedera, line 122: The old values is misleading and not easy to config for production environments.

The name indicates it's a date while the value used to be an interval.

When deploying, for instance, for mainnet, one has to override the value so the absolute date in the migration `current_timestamp - ${partitionStartDate}` is not after 2019-09

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/db/PartitionMaintenance.java
steven-sheehy: We can listen for Flyway events to determine when baseline is complete, then run this maintenance. See `MigrationHealthIndicator` for an example. Perhaps we can have some other class that adapts flyway events to Spring events to decouple the two.

xin-hedera: I think I can use a mock bean in that single integration test, it won't affect others because that integration test has its own test application context due to a specific flyway migration target set.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/BackfillAndDeduplicateBalanceMigration.java
```diff
@@ -0,0 +1,274 @@
...
+
+    private static final String SELECT_INITIAL_CONSENSUS_TIMESTAMP_SQL =
+            """
+            select coalesce(max(consensus_timestamp), 0)
+            from account_balance
+            where account_id = 2 and consensus_timestamp < ?
+            """;
+
+    private static final String SELECT_LAST_CONSENSUS_TIMESTAMP_SQL =
+            "select balance from account_balance_old where consensus_timestamp = -1 and account_id = -1";
```
steven-sheehy, line 143: I don't see where we remove this sentinel row at the end?

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/BackfillAndDeduplicateBalanceMigration.java
```diff
@@ -0,0 +1,274 @@
...
+
+    private static final String IS_ACCOUNT_BALANCE_PARTITION_EMPTY_SQL =
+            """
+            select not exists(
+              select * from account_balance
+              where consensus_timestamp >= :lowerBound and consensus_timestamp < :upperBound
+            )
+            """;
+
+    private static final long MIN_INTERVAL = Duration.ofHours(4).toNanos();
```
steven-sheehy, line 86: Should probably make this a migration property in case we need to adjust it.

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.89.2__time_partition_balance_tables.sql
```diff
@@ -0,0 +1,129 @@
...
+    from account_balance_file
+    where consensus_timestamp >= start_timestamp
+    order by consensus_timestamp
+  loop
+    -- find the partition
+    with partition_info as (
+      select
+        child.relname as name,
+        -- extract the from_timestamp from the string "FOR VALUES FROM ('xxx') to ('yyy')"
+        substring(pg_get_expr(child.relpartbound, child.oid) from 'FROM \(''(\d+)''\)')::bigint as from_timestamp
```
steven-sheehy, line 74: I see this query in a number of places including in go code. Should we create a permanent procedure for this to centralize this pg-specific partition lookup logic? Or even better a view?

steven-sheehy (changes requested):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/BackfillAndDeduplicateBalanceMigration.java
xin-hedera: the table is dropped at the end of the async migration

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/repository/AccountBalanceFileRepository.java
xin-hedera: correct that's the improvement we need for retention, especially for v2

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/migration/AbstractAsyncJavaMigrationTest.java
```diff
@@ -0,0 +1,59 @@
...
+              order by installed_rank desc
+              limit 1
+            )
+            """;
+
+    protected void resetChecksum(AsyncJavaMigration<?> migration) {
+        jdbcOperations.update(RESET_CHECKSUM_SQL, getScript(migration));
+    }
+
+    protected void waitForCompletion(AsyncJavaMigration<?> migration) {
```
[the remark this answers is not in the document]
xin-hedera: will add an abstract getter `T getMigration()` in `AbstractAsyncJavaMigrationTest`, don't want to make it a member since then have to add constructors to the concrete migration test classes v.s. generated constructor by lombok `RequiredArgsConstructor`

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/migration/AbstractAsyncJavaMigrationTest.java
```diff
@@ -0,0 +1,59 @@
...
+        await().atMost(Duration.ofSeconds(5))
+                .pollDelay(Duration.ofMillis(100))
+                .pollInterval(Duration.ofMillis(100))
+                .untilAsserted(() -> assertThat(isMigrationCompleted(migration)).isTrue());
```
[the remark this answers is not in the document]
xin-hedera: we might want to adopt the methods in `AbstractAsyncJavaMigrationTest`.

I believe it was flaky because we didn't disable  `SyntheticCryptoTransferApprovalsMigration` in test application.yaml so when the test case `onEnd` runs and checks `flyway_schema_history` the checksum might be already 1 before the actual migration completes.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/BackfillAndDeduplicateBalanceMigration.java
xin-hedera: made it a migration property

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.89.2__time_partition_balance_tables.sql
xin-hedera: made it into a view `mirror_node_time_partitions`

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/historicalbalance/HistoricalBalanceProperties.java
```diff
@@ -37,15 +38,14 @@ public class HistoricalBalanceProperties {
...
-    private boolean deduplicate = false;
-
     private boolean enabled = true;
 
     @DurationMin(minutes = 2)
     @DurationUnit(ChronoUnit.MINUTES)
     @NotNull
     private Duration initialDelay = Duration.ofMinutes(2);
 
+    @DurationMax(days = 7)
```
xin-hedera, line 48: limit it to a max of 7 days, since in the query for full snapshot, we put a lower bound of new snapshot timestamp - 30days to limit the number of time partitions to search for the last balance timestamp of account 2.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/historicalbalance/HistoricalBalanceService.java
```diff
@@ -164,7 +164,8 @@ public void onRecordFileParsed(RecordFileParsedEvent event) {
                 accountBalanceFileRepository.save(accountBalanceFile);
 
                 log.info(
-                        "Generated historical account balance file {} with {} account balances and {} token balances in {}",
+                        "Generated {} historical account balance file {} with {} account balances and {} token balances in {}",
```
xin-hedera, line 167: change the log format to show if it's a full or deduped balance snapshot

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/db/TimePartition.java
```diff
@@ -33,7 +33,7 @@ public class TimePartition implements Comparable<TimePartition> {
 
     private String name;
     private String parent;
-    private String range;
+    //    private String range;
```
steven-sheehy, line 36: We shouldn't have commented out code.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/historicalbalance/HistoricalBalanceProperties.java
steven-sheehy: Can we add a javadoc with a similar comment. Also note the max 7 days in configuration.md.

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.89.1__add_balance_deduplicate_functions.sql
```diff
@@ -0,0 +1,139 @@
...
+  -- insert a row if its balance no longer exists (i.e., the token association no longer exists at balance_timestamp),
+  -- or its balance has changed since last snapshot
+  where c.balance is null or (p.balance is null or c.balance <> p.balance);
+
+  get diagnostics count = row_count;
+  return count;
+end;
+$$ language plpgsql;
+
+create or replace view mirror_node_time_partitions as
```
steven-sheehy, line 128: Might be better to have a `R__view.sql` to create this in case we need to recreate it. But I guess repeatable get created after versioned?

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/migration/TransferTransactionPayerMigrationTest.java
```diff
@@ -57,11 +57,13 @@
 import org.junit.jupiter.api.Test;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.Import;
 import org.springframework.jdbc.core.JdbcOperations;
 import org.springframework.test.context.TestPropertySource;
 
 @DisableRepeatableSqlMigration
 @EnabledIfV1
+@Import(PartitionMaintenanceDisabledMigrationTestConfiguration.class)
```
steven-sheehy, line 66: nit: Could simplify name
```suggestion
@Import(DisablePartitionMaintenanceConfiguration.class)
```

hedera-mirror-importer/src/main/resources/db/migration/v2/R__mirror_node_time_partitions.sql
steven-sheehy, line 1: Prefer `R__view.sql`

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/BackfillAndDeduplicateBalanceMigration.java
```diff
@@ -0,0 +1,281 @@
...
+    }
+
+    @Override
+    public String getDescription() {
+        return "Backfill and deduplicate old balance information";
+    }
+
+    @Override
+    protected Long getInitial() {
+        minFrequency = DurationStyle.SIMPLE
```
steven-sheehy, line 173: Does this need to be set here or can we set in constructor?

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/migration/AbstractAsyncJavaMigrationTest.java
```diff
@@ -0,0 +1,63 @@
...
+              select checksum from flyway_schema_history
+              where script = ?
+              order by installed_rank desc
+              limit 1
+            )
+            """;
+
+    protected abstract T getMigration();
+
+    @AfterEach
```
steven-sheehy, line 43: It would be safer for future use to use `@BeforeEach` in case it gets ran automatically elsewhere. Or can maybe add both before and after annotations.

steven-sheehy (changes requested):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/db/TimePartition.java
xin-hedera: leftover when changing TimePartitionService to use the new db view, `range` is no longer needed since now we parse the range string in SQL not in java code

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.89.1__add_balance_deduplicate_functions.sql
xin-hedera: yeah, that's the problem, V1.89.2 depends on the view

hedera-mirror-importer/src/main/resources/db/migration/v2/R__mirror_node_time_partitions.sql
xin-hedera: sure

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/BackfillAndDeduplicateBalanceMigration.java
xin-hedera: set it here instead of inside the constructor so testing is easier with a stateless bean, or a bean when data / config changes, run it again would apply the new values.

Otherwise, I have to change the integration test pattern to create a new `BackfillAndDeduplicationBalanceMigration` object for each test.

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/migration/AbstractAsyncJavaMigrationTest.java
xin-hedera: sure

docs/configuration.md
```diff
@@ -158,7 +158,7 @@ value, it is recommended to only populate overridden properties in the custom `a
...
-| `hedera.mirror.importer.parser.record.historicalBalance.minFrequency`            | 15m                                              | How often at least to generate balances information. Can accept duration units like `10s`, `2m` etc.                                                                                                                                                               |
+| `hedera.mirror.importer.parser.record.historicalBalance.minFrequency`            | 15m                                              | How often at least to generate balances information. Can accept duration units like `10s`, `2m` etc. The minimium allowed value is `15m`, and the maximum is `7d`.                                                                                                                                                                |
```
steven-sheehy, line 161: minimum is misspelled.

steven-sheehy (dismissed):
LGTM 🥂

jnels124 (approved):
LGTM

steven-sheehy (approved):
LGTM

PR #7182, opened 2023-11-07
Add Health check on Rest Network Stake

### LINKED ISSUE #7141: Add a database health check to monitor
### Problem

If one Kubernetes cluster is having database problems, we should failover to another cluster automatically.

### Solution

Update `ClusterHealthIndicator` to call `/api/v1/network/stake` via `WebClient` and return `DOWN` if `5xx` and `UP` otherwise

### Alternatives

_No response_

edwin-greene:
**Description**:
Add a health check on the Rest Network Stake endpoint. If this endpoint is down it may be indicative of a database problem.

**Related issue(s)**:

Fixes #7141 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-monitor/src/main/java/com/hedera/mirror/monitor/health/ClusterHealthIndicator.java
+++ b/hedera-mirror-monitor/src/main/java/com/hedera/mirror/monitor/health/ClusterHealthIndicator.java
@@ -22,6 +22,7 @@ import com.hedera.mirror.monitor.subscribe.Scenario;
 import com.hedera.mirror.monitor.subscribe.rest.RestApiClient;
 import jakarta.inject.Named;
 import java.time.Duration;
+import lombok.CustomLog;
 import lombok.RequiredArgsConstructor;
 import org.apache.commons.lang3.StringUtils;
 import org.springframework.boot.actuate.health.Health;
@@ -29,13 +30,11 @@ import org.springframework.boot.actuate.health.ReactiveHealthIndicator;
 import org.springframework.boot.actuate.health.Status;
 import reactor.core.publisher.Mono;
 
+@CustomLog
 @Named
 @RequiredArgsConstructor
 public class ClusterHealthIndicator implements ReactiveHealthIndicator {
 
-    private static final Mono<Health> NETWORK_STAKE_DOWN = health(Status.DOWN, "Rest Network stake endpoint is down");
-    private static final Mono<Health> NETWORK_STAKE_UNKNOWN =
-            health(Status.UNKNOWN, "Rest Network stake endpoint is unknown");
     private static final Mono<Health> UNKNOWN = health(Status.UNKNOWN, "Publishing is inactive");
     private static final Mono<Health> UP = health(Status.UP, "");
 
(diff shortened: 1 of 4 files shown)


hedera-mirror-monitor/src/main/java/com/hedera/mirror/monitor/health/ClusterHealthIndicator.java
```diff
@@ -71,4 +79,20 @@ private Mono<Health> subscribing() {
...
+                    if (statusCode.is2xxSuccessful()) {
+                        return UP;
+                    } else if (statusCode.is5xxServerError()) {
+                        return NETWORK_STAKE_DOWN;
+                    }
+
+                    return NETWORK_STAKE_UNKNOWN;
+                })
+                .timeout(Duration.ofSeconds(5))
+                .onErrorResume(e -> NETWORK_STAKE_UNKNOWN);
```
steven-sheehy, line 96: For api down or unknown, a non-static health with a descriptive reason would be beneficial. Since errors should be infrequent, just call `health()` here and pass the exception message or exact status code. We should also log the error for those.

hedera-mirror-monitor/src/test/java/com/hedera/mirror/monitor/health/ClusterHealthIndicatorTest.java
```diff
@@ -50,19 +60,31 @@ class ClusterHealthIndicatorTest {
...
-        "1.0, 1.0, UP", // healthy
-        "0.0, 1.0, UNKNOWN", // publishing inactive
-        "1.0, 0.0, UNKNOWN", // subscribing inactive
-        "0.0, 0.0, UNKNOWN", // publishing and subscribing inactive
+        "1.0, 1.0, 200, UP", // healthy
+        "0.0, 1.0, 200, UNKNOWN", // publishing inactive
+        "1.0, 0.0, 200, UNKNOWN", // subscribing inactive
+        "0.0, 0.0, 200, UNKNOWN", // publishing and subscribing inactive
+        "1.0, 1.0, 400, UNKNOWN", // unknown network stake
+        "1.0, 1.0, 500, DOWN", // network stake down
```
steven-sheehy, line 68: Add test case for 0/0/500, 0/1/500, 1/0/500 for completeness even though they should be equivalent to this.

steven-sheehy (changes requested):

hedera-mirror-monitor/src/main/java/com/hedera/mirror/monitor/health/ClusterHealthIndicator.java
edwin-greene: Sounds good I have made those changes.

hedera-mirror-monitor/src/test/java/com/hedera/mirror/monitor/health/ClusterHealthIndicatorTest.java
edwin-greene: Got it, I have updated the test with those additions.

steven-sheehy (approved):
LGTM

xin-hedera (approved):
LGTM

PR #7185, opened 2023-11-08
HIP-584 Historical: Set historical block values

### LINKED ISSUE #7184: HIP-584 Historical: Set historical block values
### Problem

We already have a timestamp field added in `ContractCallContext` class with a default value of -1.
In order to support evm operations we need to set historical block values like: `gasLimit`, `blockNo`, `consTimestamp`

### Solution

Based on the timestamp field extract the needed block values data and set in `StaticBlockMetaSource`.

### Alternatives

_No response_

0xivanov:
**Description**:

We already have a timestamp field added in `ContractCallContext` class with a default value of -1.
In order to support evm operations we need to set historical block values like: `gasLimit`, `blockNo`, `consTimestamp`.

Based on the timestamp field extract the needed block values data and set in `StaticBlockMetaSource`.

**Related issue(s)**:

Fixes #7184

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/common/ContractCallContext.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/common/ContractCallContext.java
@@ -33,11 +33,6 @@ public class ContractCallContext implements AutoCloseable {
 
     private static final ThreadLocal<ContractCallContext> THREAD_LOCAL = ThreadLocal.withInitial(() -> null);
 
-    /**
-     * Constant for representing an unset or disabled timestamp for filtering.
-     */
-    public static final long UNSET_TIMESTAMP = -1L;
-
     /** Map of account aliases that were committed */
     private final Map<Address, Address> aliases = new HashMap<>();
 
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/common/ContractCallContext.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/common/ContractCallContext.java
@@ -16,6 +16,7 @@
 
 package com.hedera.mirror.web3.common;
 
+import com.hedera.mirror.common.domain.transaction.RecordFile;
 import com.hedera.mirror.web3.evm.store.CachingStateFrame;
 import com.hedera.mirror.web3.evm.store.StackedStateFrames;
 import java.util.EmptyStackException;
@@ -52,7 +53,7 @@ public class ContractCallContext implements AutoCloseable {
      * Any value other than UNSET_TIMESTAMP that is a valid timestamp should be considered for filtering operations.
      */
     @Setter
-    private long blockTimestamp = UNSET_TIMESTAMP;
+    private RecordFile recordFile;
 
     /** Boolean flag which determines whether we should make a contract call or contract init transaction simulation */
     @Setter
@@ -103,7 +104,7 @@ public class ContractCallContext implements AutoCloseable {
         estimate = false;
         pendingAliases.clear();
         pendingRemovals.clear();
-        blockTimestamp = UNSET_TIMESTAMP;
+        recordFile = null;
         stack = stackBase;
     }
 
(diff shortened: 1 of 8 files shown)


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/StaticBlockMetaSource.java
```diff
@@ -42,11 +43,11 @@ public Hash getBlockHash(long blockNo) {
 
     @Override
     public BlockValues computeBlockValues(long gasLimit) {
-        final var latestRecordFile = recordFileRepository
-                .findLatest()
+        final var recordFile = recordFileRepository
+                .findByTimestamp(ContractCallContext.get().getBlockTimestamp())
```
steven-sheehy, line 47: This is inefficient. We already query for the block timestamp to set `ContractCallContext.blockTimestamp`. We should just set the entire `RecordFile` inside the context so we don't have to re-query it here.

0xivanov: I now saw that i am missing logic. If we have non-historical call the timestamp is -1 and the query will fail. So i should have check for that and query latest block if timestamp is -1.
Anyhow, should i set the record file only when we have historical call? Because if not, we would still make 2 identical queries when doing non-historical call: findByIndex returns latest record file and findLatest returns the same.

0xivanov: I could set the record file regardless of historical/non-historical type of call, but this would change the logic quite a bit in ContractCallService

0xivanov: Implemented the record file setting, i think we could optimise it even more, so feel fee to give suggestions.
@IvanKavaldzhiev 
@steven-sheehy

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/RecordFileRepository.java
```diff
@@ -46,6 +47,10 @@ public interface RecordFileRepository extends PagingAndSortingRepository<RecordF
     @Query("select r from RecordFile r where r.index = ?1")
     Optional<RecordFile> findByIndex(long index);
 
+    @Cacheable(cacheNames = CACHE_NAME, cacheManager = CACHE_MANAGER_RECORD_FILE_TIMESTAMP, unless = "#result == null")
+    @Query("select r from RecordFile r where r.consensusEnd = ?1")
+    Optional<RecordFile> findByTimestamp(long timestamp);
```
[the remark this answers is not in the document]
0xivanov: Removed

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/common/ContractCallContext.java
```diff
@@ -52,7 +53,7 @@ public class ContractCallContext implements AutoCloseable {
      * Any value other than UNSET_TIMESTAMP that is a valid timestamp should be considered for filtering operations.
      */
     @Setter
-    private long blockTimestamp = UNSET_TIMESTAMP;
```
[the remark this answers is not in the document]
0xivanov: Removed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/StaticBlockMetaSource.java
```diff
@@ -34,19 +36,26 @@ public class StaticBlockMetaSource implements BlockMetaSource {
 
     @Override
     public Hash getBlockHash(long blockNo) {
-        final var recordFile = recordFileRepository.findByIndex(blockNo);
-        return recordFile
-                .map(rf -> ethHashFrom(rf.getHash()))
-                .orElseThrow(() -> new MissingResultException(String.format("No record file with index: %d", blockNo)));
+        var recordFile = ContractCallContext.get().getRecordFile();
+        if (Objects.isNull(recordFile)) {
```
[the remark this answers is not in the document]
0xivanov: Nice catch. We are not guaranteed as far as I know. I ran the acceptance tests to see how this method is handled, but it was not called even once. I reverted the change for this method just to be sure.

steven-sheehy (approved):
LGTM

IvanKavaldzhiev (approved):
LGTM

PR #7194, opened 2023-11-10
Set correct sender account in `TransferPrecompile`

### LINKED ISSUE #7192: CONTRACT_REVERT_EXECUTED on estimateGas for transferFrom operation
### Description

**CONTRACT_REVERT_EXECUTED is received while calling estimateGas for transferFrom operation.**

**The transfer is executed successfully and the amount is present to account:**
     const tx2 = await contract.transferFromExternal.populateTransaction(tokenAddress, account0LongZero, account1LongZero, amount);

**While calling estimate gas:**
   const estimateGasResponse = await relay.call(RelayCalls.ETH_ENDPOINTS.ETH_ESTIMATE_GAS, [tx2]);
**the following response is observed:**

`  Error: execution reverted (no data present; likely require(false) occurred (action="estimateGas", data="0x", reason="require(false)", transaction={ "data": "0x75a85472000000000000000000000000000000000000000000000000000000000000040e000000000000000000000000000000000000000000000000000000000000040b000000000000000000000000000000000000000000000000000000000000040c0000000000000000000000000000000000000000000000000000000000000001", "to": "0x000000000000000000000000000000000000040A" }, invocation=null, revert=null, code=CALL_EXCEPTION, version=6.8.0)`

### Steps to reproduce

TBD

### Additional context

_No response_

### Hedera network

other

### Version

0.90

### Operating system

None

0xivanov:
**Description**:

Sender account is not being set properly in `TransferPrecompile`. The correct account should be `senderAddress` from `HtsPrecompiledContract`.
This causes incorrect behaviour compared to hedera-services.

Affected operations: `transferFrom`

**Related issue(s)**:

Fixes #7192

**Notes for reviewer**:
We have false positive tests in web3 and test modules that will be corrected with this PR.

**Checklist**


hedera-mirror-web3/src/main/java/com/hedera/services/store/contracts/precompile/Precompile.java
```diff
@@ -59,7 +60,8 @@ long getGasRequirement(
             long blockTimestamp,
             TransactionBody.Builder transactionBody,
             Store store,
-            HederaEvmContractAliases mirrorEvmContractAliases);
```
0xivanov, line 62: If someone has better alternative for getting the `senderAddress` from `HtsPrecompiledContract` feel free to share

hedera-mirror-web3/src/test/resources/contracts/DynamicEthCalls/DynamicEthCalls.sol
```diff
@@ -402,3 +407,12 @@ contract DynamicEthCalls is HederaTokenService {
         if(isKyc) revert("Kyc status mismatch");
     }
 }
+
+contract SpenderContract {
+    function spendFungible(address token, uint256 amount, address from) public {
+        IERC20(token).transferFrom(from, address(this), amount);
```
IvanKavaldzhiev, line 413: Here we have a case where the `spender` and `recipient` accounts are the same. It would be good to also have a test where the recipient is a different account, to cover more test cases. This can be achieved by passing the recipient address as an argument.

0xivanov: Added `to` argument and now the function is called 4 times. 2 for nft with 2 different `to` addresses and the same for fungible.

hedera-mirror-web3/src/main/java/com/hedera/services/store/contracts/precompile/Precompile.java
steven-sheehy: In mono service, it's a class field in `TransferPrecompile` since the class is stateful and constructed for each request. Here, `TransferPrecompile` is stateless so we can't use that approach. Whenever there is per-request state that some previously stateful class needs we should use `ContractCallContext`. This minimizes changes to the copied services code. This is also the approach taken in #7177 which makes `HtsPrecompiledContract` stateless, so you wouldn't be able to use this approach if that PR is merged.

I also think `TransactionBody` should be removed from `getMinimumFeeInTinybars(Timestamp consensusTime, final TransactionBody transactionBody)` for the same reason and moved to `ContractCallContext`. But that can be outside the scope of this PR.

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/EstimatePrecompileFeature.java
```diff
@@ -448,9 +448,12 @@ public void ercTransferFromEstimateGasWithoutApproval() {
...
     }
 
-    @And("I approve receiver account to use fungible token")
+    @And("I approve the contract to use fungible token")
     public void approveFungibleWithReceiver() {
-        networkTransactionResponse = accountClient.approveToken(fungibleTokenId, receiverAccount.getAccountId(), 10);
+        networkTransactionResponse = accountClient.approveToken(
+                fungibleTokenId, AccountId.fromSolidityAddress(ercTestContractSolidityAddress), 10);
+        networkTransactionResponse = accountClient.approveToken(
+                fungibleTokenId, AccountId.fromSolidityAddress(precompileTestContractSolidityAddress), 10);
```
steven-sheehy, line 456: nit: Move account ids to separate variables so approveToken fits on one line.

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/services/store/contracts/precompile/Precompile.java
0xivanov: Moved the sender address to `ContractCallContext`.
About #7177 we are not sure yet if it is worth making `HtsPrecompiledContract` bean.
Also i agree with the comment about the `TransactionBody`

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/EstimatePrecompileFeature.java
0xivanov: Fixed

hedera-mirror-web3/src/test/java/com/hedera/services/store/contracts/precompile/AssociatePrecompileTest.java
```diff
@@ -89,6 +93,11 @@ class AssociatePrecompileTest {
     private final TransactionBody.Builder transactionBody =
             TransactionBody.newBuilder().setTokenAssociate(TokenAssociateTransactionBody.newBuilder());
 
+    private MockedStatic<ContractCallContext> staticMock;
+
+    @Mock
+    private ContractCallContext contractCallContext;
```
steven-sheehy, line 99: Not sure why all these changes are necessary when currently only TransferPrecompile needs the `ContractCallContext`?

0xivanov: They call `subject.prepareFields(frame);` so we need to mock the `contractCallContext.get().setSenderAddress(senderAddress);`  method, or it will throw NPE for some tests.

steven-sheehy: Got it, thanks

steven-sheehy (approved):
LGTM

IvanKavaldzhiev (approved):
LGTM

Issue #7147, opened 2023-11-01
Inconsistent eth_call response: random 0x return instead of expected value


yiliev0:
### Description

eth_call for `isToken` should return either true or false, represented as 0x000...001 or 0x000...000, but on random occasions, it's returning 0x.
The same fails for `getTokenInfo` which returns only 0x but should return a wide range of information.

Additional note: After quite some time(5-10m) the same request starts returning valid response. 

### Steps to reproduce

1. Create a fungible token
2. Deploy a contract with a function that consists of  `getTokenInfo` and `isToken`
3. Execute eth_call using the function selector+token(from step1) to the contract(step2)

### Hedera network

other

### Version

0.92.0

### Operating system

None

PR #7203, opened 2023-11-14
HIP-584: 0x returned when contract's runtime bytecode is not available yet

bilyana-gospodinova:
**Description**:
Some of the acceptance tests for isToken and getTokenInfo scenarios were failing randomly. The response was HTTP 200 and a body with content "0x". After further investigation it seems that if a contract is not fully deployed (it has an address but the runtime bytecode is not imported yet) then we get this ambiguos response.

In this PR the following changes were made:
- `MirrorEntityAccess#fetchCodeIfPresent` is changed to return null if the entity is not found. This change was necessary because the AbstractCodeCache depends on this value in order to function correctly. Otherwise, the cache would contain addresses as keys and null values for some time even when the code is available at a later point. 
- `MirrorEvmTxProcessorImpl#buildInitialFrame` was changed to correspond to the implementation in services.
- As a result from the other two bullet points in case the runtime bytecode of a contract is not available yet in buildInitialFrame a `MirrorEvmTransactionException` will be thrown and the user will get HTTP 400 with error code "INVALID_TRANSACTION".

The acceptance tests are now enhanced to verify that a contract's runtime bytecode is available and that the token exists before executing the other tests for isToken and getTokenInfo. 

Fixes https://github.com/hashgraph/hedera-mirror-node/issues/7147

### DIFF
--- a/hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/TokenFeature.java
+++ b/hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/TokenFeature.java
@@ -100,6 +100,8 @@ public class TokenFeature extends AbstractFeature {
             this.networkTransactionResponse = tokenAndResponse.response();
             verifyMirrorTransactionsResponse(mirrorClient, 200);
         }
+        var tokenInfo = mirrorClient.getTokenInfo(tokenAndResponse.tokenId().toString());
+        log.info("Get token info for token {}: {}", tokenId, tokenInfo);
     }
 
     @Given("I associate account {account} with token {token}")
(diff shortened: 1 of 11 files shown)


0xivanov (approved):
LGTM

kselveliev (approved):
LGTM

IvanKavaldzhiev (approved):
LGTM

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/TokenFeature.java
```diff
@@ -100,6 +100,8 @@ public void createNamedToken(TokenNameEnum tokenName) {
             this.networkTransactionResponse = tokenAndResponse.response();
             verifyMirrorTransactionsResponse(mirrorClient, 200);
         }
+        var tokenInfo = mirrorClient.getTokenInfo(tokenAndResponse.tokenId().toString());
+        log.info("Get token info for token {}: {}", tokenId, tokenInfo);
```
steven-sheehy, line 104: In general, we don't print out API responses at INFO level in acceptance tests. Should remove or set to debug.

steven-sheehy (approved):

PR #7206, opened 2023-11-14
Topic message optimization

### LINKED ISSUE #6976: Topic Message Endpoints Show Regression in V2
### Problem

`/topics/{id}/messages`,  `/topics/{id}/messages/{sequenceNumber}`, and `/topics/{id}/messages?sequencenumber={sequenceNumber}` show large performance decrease in V2 when compared to V1

### Solution

When `topic_message` table is partitioned by timestamp range, we need to utilize `topic_message_lookup` table to narrow down the number of partitions to query for better performance.

- Set `hedera.mirror.importer.parser.record.entity.persist.topicMessageLookups=true` to enable `topic_message_lookup` and run the migration to backfill the table from existing data in `topic_message` table
- In Rest API
  -  Add a utility function `isV2Schema` to lazily detect if it's running v2 by checking if migration version `2.0.0` is in `flyway_schema_history` table. The function should cache the result
  - For `/topics/{id}/messages`, parse the `sequencenumber` query filters to build a sequence number range, query `topic_message_lookups` table to find the covering timestamp ranges, e.g.
     ```sql
     select timestamp_range
     from topic_message_lookup
     where topic_id = $1 and sequence_number_range && $2
     order by sequence_number_range
     ```
     `&&` is the PostgreSQL range overlap operator, `$2` is the sequence number range built from query filters. 
     A couple of things to note for building the sequence number range
     -  Consider the limit and order to modify either the lower or the upper bound
     - When there is no `sequencenumber` query filters, depending on the order, use different queries to get the first or the last X topic messages' timestamp range from `topic_message_lookup` table
     - Consider shortcut and return empty response if the range built is empty, e.g., `sequencenumber=gt:10&&sequencenumber=lt:10`
     - Reject the query if there is any `ne` operator, e.g., `sequencenumber=ne:10`
     - Reject the query if there are multiple `eq` operators, eg., `sequencenumber=10&&sequencenumber=eq:11`
  - For `/topics/{id}/messages/{sequenceNumber}`, the query to find the timestamp range is similar, just need to use the range include operation, e.g., `sequence_number_range @> $2`
  - Limit the number of timestamp partitions by adding the timestamp ranges to the where conditions
 
### Alternatives

_No response_

mgoelswirlds:
**Description**:
Optimizing topic message endpoints for V2. This change will be under a feature flag , and can be enabled for v1 if needed.

This PR modifies 
* /topics/{id}/messages endpoint to use topic_message_lookup table based on sequence numbers
* it also handles the case for when sequence numbers are absent.
* Enable hedera.mirror.rest.query.v2.topicMessageLookups for v2.

**Related issue(s)**:

Fixes #6976 

**Notes for reviewer**:

**Checklist**


mgoelswirlds:
Work in progress:
- Tests need to be written
- the scenario for returning an empty reponse needs to be handled.

hedera-mirror-rest/topicmessage.js
```diff
@@ -158,16 +163,89 @@ const getTopicMessages = async (req, res) => {
...
+        console.log(`upperLimit is ${upperLimit}`);
+      } else if (lowerLimit === undefined && order === constants.orderFilterValues.DESC) {
+        lowerLimit = upperLimit - limit;
+      }
+
+      // add sequence number range
+      hasSequenceNumber = true;
+    }
+  }
+  if (!hasSequenceNumber) {
```
xin-hedera, line 220: if there is no sequence number in the request URL, we need to query `topic_message_lookup` table for either the min or max sequence number depending on `order`

hedera-mirror-rest/topicmessage.js
```diff
@@ -202,6 +284,21 @@ const extractSqlFromTopicMessagesRequest = (topicId, filters) => {
...
 
+const getTopicMessageTimestamps = async (topicId, filters) => {
+  const pgObject = await extractSqlForTopicMessagesLookup(topicId, filters);
+  if (pgObject === null) {
+    return null;
+  }
+  const params = pgObject.params;
+  const query = pgObject.query;
+
+  const row = await pool.queryQuietly(query, params);
```
xin-hedera, line 295:
```suggestion
  const {rows} = await pool.queryQuietly(query, params);
```

xin-hedera (changes requested):

docs/configuration.md
```diff
@@ -534,6 +534,7 @@ value, it is recommended to only populate overridden properties in the custom `a
 | `hedera.mirror.rest.query.maxTimestampRange`                       | 7d                      | The maximum amount of time a timestamp range query param can span for some APIs.                                                                                                              |
 | `hedera.mirror.rest.query.maxTransactionConsensusTimestampRange`   | 35m                     | The maximum amount of time of a transaction's consensus timestamp from its valid start timestamp.                                                                                             |
 | `hedera.mirror.rest.query.strictTimestampParam`                    | true                    | Enables strict checking of timestamp query param (currently only effects /api/v1/accounts/{id}?timestamp={timestamp}                                                                          |
+| `hedera.mirror.rest.query.v2.topicMessageLookups`                  | false                   | Enables topic message lookup querying when enable for v2                                                                                                                                      |
```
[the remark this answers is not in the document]
xin-hedera: talked to Steven offline. Let's keep the feature flag, it's less work than detecting v2 during runtime.

hedera-mirror-rest/topicmessage.js
```diff
@@ -158,16 +163,89 @@ const getTopicMessages = async (req, res) => {
...
+        }
+      }
+      console.log(`lowerLimit is ${lowerLimit}`);
+      // we add or subtract the limit depending on the order to obtain the range.
+      if (upperLimit === undefined && order === constants.orderFilterValues.ASC) {
+        upperLimit = lowerLimit + limit;
+        console.log(`upperLimit is ${upperLimit}`);
+      } else if (lowerLimit === undefined && order === constants.orderFilterValues.DESC) {
+        lowerLimit = upperLimit - limit;
+      }
```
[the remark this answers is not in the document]
mgoelswirlds: yes.

hedera-mirror-rest/topicmessage.js
mgoelswirlds: You mean get the the min or max sequence_number_range from topic_message_lookup right?

hedera-mirror-rest/topicmessage.js
```diff
@@ -158,16 +163,89 @@ const getTopicMessages = async (req, res) => {
...
+  let pgSqlQuery = `select numrange(min(lower(timestamp_range)), max(upper(timestamp_range)))
+                    from topic_message_lookup
+                    where ${TopicMessageLookup.TOPIC_ID} = $1 `;
+  const pgSqlParams = [topicId.getEncodedId()];
+
+  // add filters
+  let limit = defaultLimit;
+  let order = constants.orderFilterValues.ASC;
+  let hasSequenceNumber = false;
+  const bound = new Bound(filterKeys.SERIAL_NUMBER, 'serial_number');
```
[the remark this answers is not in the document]
mgoelswirlds: should be SEQUENCE_NUMBER. But not using Bound anymore.

hedera-mirror-rest/topicmessage.js
```diff
@@ -158,16 +163,89 @@ const getTopicMessages = async (req, res) => {
...
+      // add sequence number range
+      hasSequenceNumber = true;
+    }
+  }
+  if (!hasSequenceNumber) {
+    pgSqlQuery += `  group by ${TopicMessageLookup.PARTITION} order by ${TopicMessageLookup.PARTITION} ${order}`;
+    // add limit
+    pgSqlQuery += ` limit ${limit}`;
+  } else {
+    pgSqlQuery += ` and  sequence_number_range && '[${lowerLimit},${upperLimit})'::int8range`;
```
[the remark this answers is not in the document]
mgoelswirlds: Changed it to be inclusive.

hedera-mirror-rest/topicmessage.js
```diff
@@ -158,16 +163,89 @@ const getTopicMessages = async (req, res) => {
   res.locals[constants.responseDataLabel] = topicMessagesResponse;
 };
 
-const extractSqlFromTopicMessagesRequest = (topicId, filters) => {
+const extractSqlForTopicMessagesLookup = async (topicId, filters) => {
+  let pgSqlQuery = `select numrange(min(lower(timestamp_range)), max(upper(timestamp_range)))
```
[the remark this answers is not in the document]
mgoelswirlds: Modified it.

hedera-mirror-rest/topicmessage.js
mgoelswirlds: I have a follow-up question. When we do get the min/max sequence number, we should still get the limit number of topic messages from a given partition which has the min/max sequence number respectively right? Currently I am getting only one but that doesn't seem right.

hedera-mirror-rest/__tests__/specs/topics/{id}/messages/none.json
```diff
@@ -26,6 +33,14 @@
...
+    "topicmessagelookups": [
+      {
+        "partition": "topic_message_p2020_07",
+        "timestamp_range": "[1234567890000000001,1234567890000000003)",
+        "sequence_number_range": "[1,4)",
+        "topic_id": 7
+      }
     ]
   },
   "url": "/api/v1/topics/7/messages?sequencenumber=2&sequencenumber=gte:3&timestamp=lt:1234567890.000000004",
```
mgoelswirlds, line 46: Need to handle the scenario where a missing range is built. Once that is fixed this test should pass.

hedera-mirror-rest/__tests__/specs/topics/{id}/messages/specific-sequence.json
```diff
@@ -26,6 +33,14 @@
...
+    ],
+    "topicmessagelookups": [
+      {
+        "partition": "topic_message_p2020_07",
+        "timestamp_range": "[1234567890000000001,1234567890000000004)",
+        "sequence_number_range": "[1,4)",
+        "topic_id": 7
+      }
     ]
   },
```
mgoelswirlds, line 45: For v2 now the code generating a range and it is returning a limit number of topic Messages which is more topic messages than the one eq to indicates.  Need input on handling this case.

hedera-mirror-rest/topicmessage.js
```diff
@@ -158,7 +161,87 @@ const getTopicMessages = async (req, res) => {
...
+      if (filter.operator === utils.opsMap.lte) {
+        upperLimit = upperLimit === undefined ? Number.MAX_SAFE_INTEGER : upperLimit;
+        upperLimit = Math.min(upperLimit, Number(filter.value));
+      }
+    }
+  }
+  // If the range built is empty shortcut with an empty response
+  if (lowerLimit !== undefined && upperLimit !== undefined) {
+    if (lowerLimit > upperLimit) {
+      // need to return an empty response
```
mgoelswirlds, line 226: This branch of code when an empty range is built needs to be handles to return an empty response.

mgoelswirlds:
There are still unanswered questions which I am waiting for and then the builds would be passing.

hedera-mirror-rest/topicmessage.js
```diff
@@ -158,7 +161,87 @@ const getTopicMessages = async (req, res) => {
...
+      return null;
+    }
+  }
+  // we add or subtract the limit depending on the order to obtain the range.
+  if (upperLimit === undefined) {
+    upperLimit = lowerLimit + Number(limit);
+  } else if (lowerLimit === undefined) {
+    lowerLimit = upperLimit - Number(limit);
+  }
```
xin-hedera, line 235: can't do this. because the endpoint also supports `timestamp=XXX` filters.

so the response of  `/topics/{id}/messages?sequencenumber=gt:10&timestamp=gt:1234567890.100&limit=2` may include messages with sequence number 15 and 16 because the `gt` timestamp takes precedence.

hedera-mirror-rest/topicmessage.js
```diff
@@ -158,7 +161,87 @@ const getTopicMessages = async (req, res) => {
...
+
+  pgSqlQuery += ` and  sequence_number_range && '[${lowerLimit},${upperLimit}]'::int8range`;
+  pgSqlQuery += ` order by ${TopicMessageLookup.SEQUENCE_NUMBER_RANGE} ${order}`;
+
+  pgSqlQuery += ` limit ${limit}`;
```
xin-hedera, line 240: this is still not addressed. limit applies to the number of topic messages, not the number of rows from `topic_message_lookup`.

hedera-mirror-rest/topicmessage.js
```diff
@@ -158,7 +161,87 @@ const getTopicMessages = async (req, res) => {
...
+
+  return utils.buildPgSqlObject(pgSqlQuery, pgSqlParams, order, limit);
+};
+const extractSqlFromTopicMessagesRequest = async (topicId, filters) => {
```
xin-hedera, line 244: format, add a new line in between

hedera-mirror-rest/topicmessage.js
```diff
@@ -189,6 +279,35 @@ const extractSqlFromTopicMessagesRequest = (topicId, filters) => {
...
+    // If there is no sequence number in the request URL,
+    // query the topic_message_lookup table for either the min or max sequence number depending on order.
+    if (!hasSequenceNumberForV2) {
+      if (order === constants.orderFilterValues.ASC) {
+        pgSqlQuery += ` and sequence_number >= (select MIN(lower(sequence_number_range)) 
+      from topic_message_lookup where ${TopicMessageLookup.TOPIC_ID} = $1)`;
+      } else {
+        pgSqlQuery += ` and sequence_number <= (select MAX(upper(sequence_number_range)) 
+      from topic_message_lookup where ${TopicMessageLookup.TOPIC_ID} = $1)`;
+      }
```
xin-hedera, line 293: this doesn't optimize stuff.

what we need to do is a three-step process

- get the min or max sequence number of the topic from `topic_message_lookup` table
- depending on order and limit, get the sequence number range `[lower, upper]` and query `topic_message_lookup` to find out the timestamp_range
- finally build the SQL query to get the messages from `topic_message` table with the timestamp range

hedera-mirror-rest/topicmessage.js
```diff
@@ -202,6 +321,21 @@ const extractSqlFromTopicMessagesRequest = (topicId, filters) => {
   return utils.buildPgSqlObject(pgSqlQuery, pgSqlParams, order, limit);
 };
 
+const getTopicMessageTimestamps = async (topicId, filters) => {
+  const pgObject = extractSqlForTopicMessagesLookup(topicId, filters);
+  if (pgObject === null) {
+    return null;
+  }
+  const params = pgObject.params;
+  const query = pgObject.query;
```
xin-hedera, line 330: think this is a common coding style issue.  please space the code blocks properly

hedera-mirror-rest/topicmessage.js
```diff
@@ -189,6 +279,35 @@ const extractSqlFromTopicMessagesRequest = (topicId, filters) => {
...
+        return {};
+      }
+      pgSqlQuery += ` and ((${TopicMessage.CONSENSUS_TIMESTAMP} >= ${_.first(rows).timestamp_start} and ${
+        TopicMessage.CONSENSUS_TIMESTAMP
+      } < ${_.first(rows).timestamp_end})`;
+      rows.shift();
+      rows.map((row) => {
+        pgSqlQuery += ` or (${TopicMessage.CONSENSUS_TIMESTAMP} >= ${row.timestamp_start} and ${TopicMessage.CONSENSUS_TIMESTAMP} < ${row.timestamp_end})`;
+      });
+      pgSqlQuery += `)`;
```
xin-hedera, line 307: construct the timestamp range conditions using functional programming patteren
```suggestion
  if (rows.length) {
     const condition = rows.map((row) => `(consensus_timestamp >= ${...} and consensus_timestamp < ${...})`).join(' or ');
     pgSqlQuery += ` and (${condition})`;
   }
```

hedera-mirror-rest/topicmessage.js
```diff
@@ -189,6 +279,31 @@ const extractSqlFromTopicMessagesRequest = (topicId, filters) => {
...
+      from topic_message_lookup where ${TopicMessageLookup.TOPIC_ID} = $1)`;
+      }
+    } else {
+      // this needs to be fixed and gotten values
+      const rows = await getTopicMessageTimestamps(topicId, filters);
+      if (rows === null) {
+        return {};
+      }
+  if (rows.length) {
+     const condition = rows.map((row) => `(consensus_timestamp >= ${...} and consensus_timestamp < ${...})`).join(' or ');
```
xin-hedera, line 301: i didn't mean to literally copy the `${...}`, you need to replace them with the correct expression.

hedera-mirror-rest/topicmessage.js
mgoelswirlds: I just merged this with git. will fix it.

hedera-mirror-rest/topicmessage.js
```diff
@@ -158,7 +161,87 @@ const getTopicMessages = async (req, res) => {
...
+                    from topic_message_lookup
+                    where ${TopicMessageLookup.TOPIC_ID} = $1 `;
+  const pgSqlParams = [topicId.getEncodedId()];
+
+  // add filters
+  let limit = defaultLimit;
+  let order = constants.orderFilterValues.ASC;
+  let equal = null;
+  let lowerLimit = undefined;
+  let upperLimit = undefined;
```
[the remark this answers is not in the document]
mgoelswirlds: This is done specifically to be able to shortcut an empty range.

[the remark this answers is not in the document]
xin-hedera: I don't think so. You'll probably agree with me after reviewing how empty range detection done in the current implementation.

hedera-mirror-rest/model/topicMessageLookup.js
```diff
@@ -0,0 +1,49 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import _ from 'lodash';
+
+class TopicMessageLookup {
```
xin-hedera, line 19: nice to have the model.

however my personal opinion is we should avoid using it when constructing SQL queries, it only makes the query less readable.

no changes needed btw. however it's an item we may want to address in general.

hedera-mirror-rest/topicmessage.js
```diff
@@ -132,6 +135,12 @@ const getTopicMessages = async (req, res) => {
     },
   };
 
+  if (!query) {
+    topicMessagesResponse.messages = [];
+    res.locals[constants.responseDataLabel] = topicMessagesResponse;
```
xin-hedera, line 140: move these out of the if block, also remove line 167 since it only needs to set once

hedera-mirror-rest/topicmessage.js
```diff
@@ -158,16 +167,113 @@ const getTopicMessages = async (req, res) => {
   res.locals[constants.responseDataLabel] = topicMessagesResponse;
 };
 
-const extractSqlFromTopicMessagesRequest = (topicId, filters) => {
+async function getSequenceNumberFromTopicMessageLookup(topicId, order, limit) {
```
xin-hedera, line 170: follow the existing coding style
```suggestion
const getSequenceNumberFromTopicMessageLookup = async (topicId, order, limit) {
```

hedera-mirror-rest/topicmessage.js
```diff
@@ -21,6 +21,9 @@ import {InvalidArgumentError, NotFoundError} from './errors';
 import {TopicMessage} from './model';
 import * as utils from './utils';
 import {TopicMessageViewModel} from './viewmodel';
+import _ from 'lodash';
+import config from './config.js';
+import TopicMessageLookup from './model/topicMessageLookup.js';
```
xin-hedera, line 26:
```suggestion
import config from './config';
import {TopicMessageLookup} from './model';
```

You'll need to import and export `TopicMessageLookup` model in `model/index.js`

hedera-mirror-rest/topicmessage.js
```diff
@@ -202,6 +332,49 @@ const extractSqlFromTopicMessagesRequest = (topicId, filters) => {
...
+
+  const params = pgObject.params;
+  const query = pgObject.query;
+
+  const {rows} = await pool.queryQuietly(query, params);
+  if (rows.length === 0) {
+    throw new NotFoundError();
+  }
+
+  return _.isNil(rows) ? null : rows;
```
xin-hedera, line 352: rows won't be null or undefined
```suggestion
  return rows;
```

hedera-mirror-rest/topicmessage.js
```diff
@@ -202,6 +332,49 @@ const extractSqlFromTopicMessagesRequest = (topicId, filters) => {
...
+      query =
+        'select MIN(lower(sequence_number_range)) as sequence_number from topic_message_lookup where topic_id = $1';
+      break;
+    case constants.orderFilterValues.DESC:
+      query =
+        'select MAX(upper(sequence_number_range)) as sequence_number from topic_message_lookup where topic_id = $1';
+      break;
+  }
+
+  const rows = await pool.queryQuietly(query, params);
```
xin-hedera, line 370:
```suggestion
  const {rows} = await pool.queryQuietly(query, params);
```

hedera-mirror-rest/topicmessage.js
```diff
@@ -202,6 +332,49 @@ const extractSqlFromTopicMessagesRequest = (topicId, filters) => {
...
+  if (rows.length === 0) {
+    throw new NotFoundError();
+  }
+
+  return _.isNil(rows) ? null : rows.rows[0].sequence_number;
```
xin-hedera, line 375:
```suggestion
  return rows[0].sequence_number;
```

hedera-mirror-rest/__tests__/application.yml
```diff
@@ -3,3 +3,6 @@ hedera:
     rest:
       metrics:
         enabled: false
+      query:
+        v2:
+          topicMessageLookups: false
```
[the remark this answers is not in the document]
steven-sheehy: Agree, except would also drop the s: `hedera.mirror.rest.query.topicMessageLookup`

hedera-mirror-rest/topicmessage.js
```diff
@@ -158,16 +167,113 @@ const getTopicMessages = async (req, res) => {
...
+  return utils.buildPgSqlObject(pgSqlQuery, pgSqlParams, order, limit);
+};
+
+const extractSqlFromTopicMessagesRequest = async (topicId, filters) => {
   let pgSqlQuery = `select *
                     from ${TopicMessage.tableName}
                     where ${TopicMessage.TOPIC_ID} = $1`;
   let nextParamCount = 2;
   const pgSqlParams = [topicId.getEncodedId()];
+  let hasSequenceNumberForV2 = false;
```
steven-sheehy, line 271: Should not have variables named that are v2 specific

hedera-mirror-rest/__tests__/application.yml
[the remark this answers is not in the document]
mgoelswirlds: I agree. That makes sense.

hedera-mirror-rest/__tests__/application.yml
```diff
@@ -3,3 +3,5 @@ hedera:
     rest:
       metrics:
         enabled: false
+      query:
+        topicMessageLookups: false
```
xin-hedera, line 7: partially addressed. I also requested moving this change to config/application.yml

the test application.yml is not where we put configuration default.

In addition, Steven also asked to drop the `s` so it should be `topicMessageLookup`

docs/configuration.md
```diff
@@ -499,60 +499,61 @@ recursively merged into) the current configuration:
...
-| `hedera.mirror.rest.stateproof.streams.cloudProvider`              | S3                      | The cloud provider to download files from. Either `S3` or `GCP`                                                                                                                               |
-| `hedera.mirror.rest.stateproof.streams.endpointOverride`           |                         | Can be specified to download streams from a source other than S3 and GCP. Should be S3 compatible                                                                                             |
-| `hedera.mirror.rest.stateproof.streams.gcpProjectId`               |                         | GCP project id to bill for requests to GCS bucket which has Requester Pays enabled.                                                                                                           |
+| Name                                                              | Default                 | Description                                                                                                                                                                                   |
```
xin-hedera, line 502: this make review difficult.

should redo it with intellij markdown autoreformat disabled, or just revert and use a different text editor to add the single line.

xin-hedera (changes requested):

docs/configuration.md
mgoelswirlds: ok.

mgoelswirlds: Working on it. But to make it simpler , we can hide whitespaces , which shows only the changed line for this particular file.

xin-hedera (approved):
LGTM

Issue #7053, opened 2023-10-13
HIP-584: Historical support - Create db queries for calculating historical crypto allowance, token allowance


kselveliev:
### Problem

For historical support we need to extend db queries to have a timestamp filter.
Using this filter we can obtain records based on a historical block passed in eth_call.

### Solution

We need the extend the following queries in web3.

Create query for calculation of crypto allowance.

Create query for calculation of token allowance.

Document and test new query methods.

Use this [issue](https://github.com/hashgraph/hedera-mirror-node/issues/6812) for reference.

PR #7209, opened 2023-11-17
HIP-584 Historical: DB queries for historical crypto and token allowances

bilyana-gospodinova:
**Description**:
For historical support we need to extend db queries to have a timestamp filter.
Using this filter we can obtain records based on a historical block passed in eth_call.

Fixes https://github.com/hashgraph/hedera-mirror-node/issues/7053


xin-hedera (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenAllowanceRepository.java
```diff
@@ -34,4 +35,59 @@ public interface TokenAllowanceRepository extends CrudRepository<TokenAllowance,
...
+                                select tt.token_id, account_id, sum(tt.amount) as amount
+                                from token_transfer tt join token_allowances ta on tt.account_id = ta.spender
+                                where account_id in (select spender from token_allowances)
+                                    and tt.token_id in (select token_id from token_allowances)
+                                    and is_approval = true
+                             		and consensus_timestamp <= :blockTimestamp
+                             		and consensus_timestamp > lower((select timestamp_range from token_allowance))
+                                group by tt.token_id, account_id
+                            )
+                            select amount_granted, owner, payer_account_id, spender, timestamp_range, token_id, coalesce(amount - coalesce((select amount from transfers tr where tr.token_id = ta.token_id and tr.account_id = ta.spender), 0), 0) as amount
```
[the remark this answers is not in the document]
bilyana-gospodinova: I checked PR https://github.com/hashgraph/hedera-mirror-node/pull/6722. In `EntityRecordItemListener#insertTokenTransfers` (rows 536-552) it seems that the token allowances are decreased directly by the amount of the token transfers that were made. The `transferSpenderId` could be the payer account id or the contract result sender id but regardless the token allowances should have the correct amount at the end of the record file import. I think that in the query we do not need to subtract the transfers at all as they were already subtracted on import. In summary, the query should return only the latest token allowances for the given owner partitioned by (spender, tokenId) up to the passed block timestamp and having amount_granted <> 0. What do you think about that?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/CryptoAllowanceRepository.java
```diff
@@ -19,8 +19,63 @@
...
+                            and is_approval is true
+                            and consensus_timestamp <= :blockTimestamp
+                            and consensus_timestamp > lower(ca.timestamp_range)
+                        group by entity_id
+                        )
+                     select amount_granted, owner, payer_account_id, spender, timestamp_range, coalesce(amount - coalesce((select amount from transfers tr where tr.entity_id = ca.spender), 0), 0) as amount
+                     from crypto_allowances ca
+                    """,
+            nativeQuery = true)
+    List<CryptoAllowance> findByOwnerAndTimestamp(long owner, long blockTimestamp);
```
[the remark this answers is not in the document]
bilyana-gospodinova: Comment addressed.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/CryptoAllowanceRepository.java
```diff
@@ -19,8 +19,63 @@
...
+                                union all
+                                (
+                                    select *
+                                    from crypto_allowance_history
+                                    where owner = :owner
+                                        and lower(timestamp_range) <= :blockTimestamp
+                                )
+                            ) as all_crypto_allowances
+                        ) as grouped_crypto_allowances
+                        where row_number = 1
```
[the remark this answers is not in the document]
bilyana-gospodinova: Comment addressed.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenAllowanceRepository.java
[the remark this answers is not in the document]
bilyana-gospodinova: If we subtract the tranferred amount from the amountGranted in the query won't that duplicate the logic in `EntityRecordItemListener#insertTokenTransfers` (rows 536-552)?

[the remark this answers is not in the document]
xin-hedera: no, the importer logic changes `amount` not `amountGranted`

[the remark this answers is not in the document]
xin-hedera: Perhaps this is a clearer explanation.

Assume there is an old token allowance,

- amountGranted = 100
- amount = 20
- timestamp range = `[t1, t2)`

and when the repo function is called to provide the token allowance as of a certain block timestamp `t3`, where `t1 < t3 < t2`, we need to compute `amount` as `amountGranted - total amount using the allowance in [t1, t3]`.

Suppose the last such token transfer using the above allowance happened at time `t4`, where `t1 < t4 < t2`, the `amount` in db is only accurate for the timestamp range `[t4, t2)`. For the corner case that the allowance is never used, `amount` will be equal to `amountGranted` and be valid for `[t1, t2)`.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/CryptoAllowanceRepository.java
```diff
@@ -19,8 +19,69 @@
...
+                            on ct.entity_id = ca.owner
+                                and ct.payer_account_id = ca.spender
+                        where is_approval is true
+                            and consensus_timestamp <= :blockTimestamp
+                            and consensus_timestamp > lower(ca.timestamp_range)
+                        group by spender
+                        )
+                    select *
+                    from (
+                        select amount_granted, owner, payer_account_id, spender, timestamp_range, amount_granted - coalesce((select amount from transfers tr where tr.spender = ca.spender), 0) as amount
```
xin-hedera, line 80: missed in the previous review, amount should be

`amount_granted + coalesce((select amount from transfers tr where tr.spender = ca.spender), 0)`

Note, the only difference is change `-` to `+`, because the amount of a transfer using allowance is a debit, the amount is negative.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenAllowanceRepository.java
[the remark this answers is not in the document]
IvanKavaldzhiev: For the scope of the historical block support, when we have a given historical `block X` with timestamp `[t1, t2)`, we need to use the state of the block after all transactions inside have been processed. That means we are interested in the correct allowance value at point `t2`, which is the `block_end` timestamp. We don't need any intermediate values.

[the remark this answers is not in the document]
IvanKavaldzhiev: Probably this means we should get the closest `token_allowance` record to `t2` timestmap (`block_end`) and use it's `amount` value? That would be the allowance left at this point in time?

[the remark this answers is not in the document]
xin-hedera: We don’t have such record.

a new token allowance row is only inserted when amount_granted changes, aka, from user submitted crypto approve allowance tx.

when a user spends the allowance, the changes to the remaining amount is accumulated and the result is *updated* to the existing row.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/CryptoAllowanceRepository.java
```diff
@@ -19,8 +19,69 @@
...
 import java.util.List;
+import org.springframework.data.jpa.repository.Query;
 import org.springframework.data.repository.CrudRepository;
 
 public interface CryptoAllowanceRepository extends CrudRepository<CryptoAllowance, AbstractCryptoAllowance.Id> {
     List<CryptoAllowance> findByOwner(long owner);
+
+    /**
+     * Retrieves the most recent state of the crypto allowances by their owner id up to a given block timestamp.
+     * It takes into account the crypto transfers that happened up to the given block timestamp, sums them up
```
IvanKavaldzhiev, line 30: Тhe transfers included should be up to the end of the block (`block_end`).

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenAllowanceRepository.java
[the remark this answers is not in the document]
bilyana-gospodinova: The query is updated to take into account the token transfers.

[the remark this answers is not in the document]
bilyana-gospodinova: It seems that due to citus distribution limitations I can't make a direct join between the contract_result and token_transfer tables so I divided the subquery for that in two separate subqueries.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenAllowanceRepository.java
```diff
@@ -34,4 +35,98 @@ public interface TokenAllowanceRepository extends CrudRepository<TokenAllowance,
...
+                            ),
+                            (
+                                select amount
+                                from transfers tr
+                                where tr.token_id = ta.token_id
+                                    and tr.payer_account_id = ta.spender
+                            ),
+                            0
+                        ) as amount
+                        from token_allowances ta
```
xin-hedera, line 126: the main issue in the logic is, the token transfer using allowance from `token_transfer` table assuming `payer_account_id` is the `spender` and those with corresponding contract call results so `contract_result.sender_id` is the `spender` are not a one or the other relationship,  which however is assumed in the coalesce:

`coalesce(amount from contract_call_transfers, amount from token_transfer, 0)`

The other obvious issue is for a token transfer `(account_id = a, consensus_timestamp = c, payer_account_id = p, token_id = t, is_approval = true)`, when it's from a contract all with `contract_call.sender_id = s` and `s <> p`, we may end up counting the transfer amount twice

- to the token allowance `(owner = a, spender = p, token = t)`
- and the token allowance `(owner = a, spender = s, token = t)`

With all the complexity in this query as well as potential performance issues,  I think it's worth to add a `spender` column to `token_transfer` and let importer set the correct value, @steven-sheehy ?

xin-hedera (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenAllowanceRepository.java
bilyana-gospodinova: Comment addressed. I added also another junit test for this case.

steven-sheehy: Can consider that for a future optimization. We might also consider ways to pre-calculating the per timestamp allowance amount on the token_transfer.

steven-sheehy (dismissed):

IvanKavaldzhiev (approved):
LGTM

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenAllowanceRepository.java
```diff
@@ -34,4 +35,104 @@ public interface TokenAllowanceRepository extends CrudRepository<TokenAllowance,
...
+                                from transfers tr
+                                where tr.token_id = ta.token_id
+                                    and tr.payer_account_id = ta.spender
+                                    and tr.consensus_timestamp not in (
+                                        select consensus_timestamp
+                                        from contract_call_transfers
+                                    )
+                            ),
+                            0
+                        ) as amount
```
xin-hedera, line 131: the `coalesce` will fail hard in real environment.

coalesce expects the individual values to be a single column from a single row.

however, here we have `contract_call_transfers` coming from a query with `group by cr.sender_id, tt.token_id, tt.consensus_timestamp`, obviously the subquery in the coalesce can return multiple rows

similarly, the second coalesce has a subquery on CTE `transfers` as a result of a query with `group by tt.token_id, tt.payer_account_id, tt.consensus_timestamp`.

In both places, `sum(amount)` is needed.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenAllowanceRepository.java
```diff
@@ -34,4 +35,104 @@ public interface TokenAllowanceRepository extends CrudRepository<TokenAllowance,
...
+                            and consensus_timestamp > lower(ta.timestamp_range)
+                        group by tt.token_id, tt.payer_account_id, tt.consensus_timestamp
+                    ), contract_results_filtered as (
+                        select sender_id, consensus_timestamp
+                        from contract_result cr
+                        where cr.consensus_timestamp <= :blockTimestamp
+                            and cr.consensus_timestamp in (
+                                select consensus_timestamp
+                                from token_transfer
+                            )
```
xin-hedera, line 95: remove the `contract_results_filtered` CTE, it's only used in `contract_call_transfers` and the where condition already takes care of `consensus_timestamp < :blockTimestamp and cr.consensus_timestamp = tt.consensus_timestamp`.

It may actually improve performance since in `contract_call_transfers` the timestamp is transitively limited in the range `(lower(ta.timestamp_rangge), blocktimestamp]`

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/repository/TokenAllowanceRepositoryTest.java
```diff
@@ -47,4 +50,487 @@ void findByOwner() {
...
+                .tokenTransfer()
+                .customize(t -> t.isApproval(true)
+                        .amount(amountForTransfer)
+                        .payerAccountId(EntityId.of(senderId))
+                        .id(TokenTransfer.Id.builder()
+                                .tokenId(EntityId.of(tokenId))
+                                .accountId(EntityId.of(ownerId))
+                                .consensusTimestamp(tokenTransfer1Timestamp)
+                                .build()))
+                .persist();
```
xin-hedera, line 503: it doesn't match the data produced by consensus nodes, there shouldn't be two identical rows in `token_transfer` table. (we used to have `(consensus_timestamp, token_id, account_id)` as the primary key however we relaxed it later). Please change them to be two token transfers using allowance at different consensus timestamp.

this is also why the coalesce issue I pointed out didn't get exposed.

xin-hedera (changes requested):

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/repository/TokenAllowanceRepositoryTest.java
xin-hedera: update: I obviously put the comment to the wrong test case, it should apply to `findByOwnerAndTimestampWithTransferMultipleEntries`

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenAllowanceRepository.java
xin-hedera: update: I may have missed the fact that the CTE is due to citus limit. If so, please disregard the change request.

bilyana-gospodinova: Exactly, the CTE is due to the citus limit.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/repository/TokenAllowanceRepositoryTest.java
bilyana-gospodinova: Nice catch. Thank you! Comment addressed.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenAllowanceRepository.java
bilyana-gospodinova: Comment addressed.

PR #7212, opened 2023-11-17
Fix incorrect key to look up entity in entity state

### LINKED ISSUE #7210: Entity stake period start may set incorrectly on staking reward transfer
### Description

In `SqlEntityListener.onStakingRewardTransfer`, the entity which gets staking reward payout is looked up by wrong key, which would certainly cause a not found current entity state. When a staking reward is triggered by a entity (crypto or contract) update transaction which also changes staking settings, the entity's stake period start will be set to `today - 1` while in fact it should be `today`. This would then cause pending reward calculation to include an extra day's worth staking reward. 

https://github.com/hashgraph/hedera-mirror-node/blob/a14bf059d2bb818091726ff6665b0532cc1189e3/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/entity/sql/SqlEntityListener.java#L418

### Steps to reproduce

1. configure an account to stake to node `X`, check the entity state in db, assume its stake period start is set to `T`
2. two days later, change the account's staked node to `Y` with a `CryptoUpdate` transaction
3. verify there is a staking reward payout, also verify the stake period start is set to day `T+1`. note the correct stake period start should be `T+2` instead

### Additional context

_No response_

### Hedera network

other

### Version

v0.93.0-SNAPSHOT

### Operating system

None

xin-hedera:
**Description**:

This PR fixes the bug in `SqlEntityListener.onStakingRewardTransfer`

- Use correct key to look up entity in state
- Add integration test for two staking reward transfers in same record file cross UTC day boundary
- Change `DomainUtils.id()` to always return non reserved entity id

**Related issue(s)**:

Fixes #7210

**Notes for reviewer**:

It turns out the bug won't cause incorrect staking metadata and incorrect pending rewards calculation.

Even though due to the incorrect key used, `onStakingRewardTransfer` will always create an entity update object with stakePeriodStart to today - 1. In case the staking reward transfer is triggered by an entity update transaction changing staking settings, which would also change stake period start, though to a different value (today), the concern in the ticket was importer may incorrectly set stakePeriodStart to today - 1. However the logic in `mergeEntity` will actually preserve the value `today` for consecutive entity updates with the same lower timestamp.

And there is an existing integration test to validate `today` is set as the new stake period start:

https://github.com/hashgraph/hedera-mirror-node/blob/ee649c0a7ebee27cc8d72c2f105fbeb03eb8aa2e/hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/record/entity/EntityRecordItemListenerCryptoTest.java#L878C10-L878C65

**Checklist**


hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/record/entity/EntityRecordItemListenerCryptoTest.java
```diff
@@ -1736,6 +1736,96 @@ void persistTransactionRecordBytes(boolean persist) {
...
+        // 5 ns before the next UTC midnight
+        var beforeNextMidnight = TestUtils.asStartOfEpochDay(currentDay + 1).minusNanos(5);
+        // for simplicity, use a ConsensusSubmitMessage transaction to trigger the staking reward payout
+        var consensusSubmitMessage1 = recordItemBuilder
+                .consensusSubmitMessage()
+                .record(r -> r.setConsensusTimestamp(Utility.instantToTimestamp(beforeNextMidnight))
+                        .addPaidStakingRewards(accountAmount(account.getNum(), 200))
+                        .setTransferList(TransferList.newBuilder()
+                                .addAccountAmounts(accountAmount(account.getNum(), 100))
+                                .addAccountAmounts(accountAmount(2, 20))
```
xin-hedera, line 1764: `DomainUtils.id()` generates id starting from 1 caused issue here, when `account.getNum()` happens to be 2, the assertions of the account's balance change in the `then` block fails miserably.

Thus I made the change to `Domainutils.id()` to always return value starting from 1001.

steven-sheehy (approved):
LGTM

jascks (approved):
LGTM

PR #7247, opened 2023-11-23
HIP-584: Integrate aliases into stacked state

### LINKED ISSUE #6857: HIP-584: Integrate aliases collections into the in-memory stacked state
### Problem

Currently we don't utilize the stacked frame design for keeping track of newly created or deleted aliases. Instead, we keep track of them in separate collections and don't benefit from the already existing internal logic for the stacked state frame design.

### Solution

Integrate the `CachingStateFrame` to also keep track of aliases via a new type of Accessor dedicated for aliases. Remove redundant logic from MirrorEvmContractAliases and optimize the code base.

### Alternatives

_No response_

IvanKavaldzhiev:
**Description**:

This PR aims to clean up the code base and mainly reduce the complexity of maintaining different maps with aliases. All of the business logic is handled automatically by the in-memory state (`Store`) by keeping additional `account` entry for a new alias.

As a side effect the size of the `ThreadLocal` context is reduced which would further increase performance.

In addition redundant arguments for some of the pricing logic methods were removed, since now we can integrate them directly as beans via injection.
**Related issue(s)**:

Fixes #6857 

**Notes for reviewer**:

**Checklist**


bilyana-gospodinova (approved):
LGTM

steven-sheehy (approved):
LGTM

PR #7283, opened 2023-11-29
Use deduped token balance data in historical token balance distribution REST API

### LINKED ISSUE #7281: Change historical token balance rest api /tokens/{id}/balances?timestamp=XXX to use deduped table
### Problem

Since the `token_balance` table is now deduped and partitioned by timestamp, we need to adapt the rest api `/tokens/{id}/balances?timestamp=XXX` to correctly get the balance info.

### Solution

- use the same logic in historical balance rest api to find the correct snapshot timestamp
- get the balance distribution of a specific token at the snapshot timestamp
- add new index to improve performance if needed

### Alternatives

_No response_

xin-hedera:
**Description**:

This PR changes `/tokens/{id}/balances?timestamp=XXX` to look up a token's balance distribution from the deduped `token_balance` table

- Change historical token balance distribution SQL query to use deduped `token_balance` table
- Refactor some functions in balances.js for better testability
- Change `balances.getAccountBalanceTimestampRange` to include the optimization logic

**Related issue(s)**:

Fixes #7281

**Notes for reviewer**:

**Checklist**


hedera-mirror-rest/__tests__/balances.test.js
```diff
@@ -320,3 +324,98 @@ describe('Balances tests', () => {
...
   testutils.testBadParams(request, server, api, 'order', testutils.badParamsList());
 });
+
+describe('getOptimizedTimestampRange', () => {
+  let clock;
+  // Set now, so Date.now() returns the expected epoch millis
+  const now = Date.parse('2022-10-15T08:10:15.333Z');
+
+  beforeAll(() => {
+    clock = sinon.useFakeTimers({now});
```
xin-hedera, line 334: use sinon fake timers, so Date.now stays at the value we want in order to test `getOptimizedTimestampRange`

hedera-mirror-rest/__tests__/specs/accounts/{id}/timestamp.json
```diff
@@ -285,18 +285,9 @@
       }
     ],
     "balance": {
-      "timestamp": "1234567890.300000010",
```
xin-hedera, line 288: the response has changed because now `balances.getAccountBalanceTimestampRange` includes the timestamp range optimization and then the logic falls back to the balance values in `entity` table.

hedera-mirror-rest/__tests__/specs/tokens/{id}/balances/all-params.json
```diff
@@ -206,55 +216,23 @@
             "balance": 190190
           }
         ]
-      },
-      {
-        "timestamp": 1566560003000000000,
```
xin-hedera, line 211: - account `0.1.20` public key is changed to match what's in the request url
- the request URL is changed to also include account `0.1.20`, i.e., `account.id=lt:0.1.21`
- the balance info for `0.1.20` at `1566560003.000000000` is removed to validate that when it's deduped, the SQL query gets the balance from the previous snapshot for `0.1.20`

hedera-mirror-rest/__tests__/tokens.test.js
```diff
@@ -404,76 +402,6 @@ describe('token extractSqlFromTokenBalancesRequest tests', () => {
         limit: defaultLimit,
       },
     },
-    {
```
xin-hedera, line 407: can no longer do unit test with `timestamp=XXX` filter because that'll query the database

hedera-mirror-rest/utils.js
```diff
@@ -922,7 +922,7 @@ const getFirstDayOfMonth = (secNs) => {
     return null;
   }
 
-  if (typeof secNs !== 'BigInt') {
+  if (typeof secNs !== 'bigint') {
```
xin-hedera, line 925: bug, because the `typeof` output of a BigInt is `bigint`, not `BigInt`

jnels124 (approved):
LGTM

edwin-greene (approved):
Looks good to me!

steven-sheehy (approved):

PR #7297, opened 2023-11-30
Fix incorrect timestamp in historical balances response

### LINKED ISSUE #7296: Balances REST api may return incorrect historical balance timestamp
### Description

The SQL query in balances REST api will pick the consensus timestamp tied to the account's balance, however, it's wrong for historical balance info.

When balance tables are deduped, an account's balance info may be retrieved from a row with timestamp t1, while the snapshot timestamp determined from the treasury account `0.0.2` is at t2, and we should always use t2 as the timestamp.

### Steps to reproduce

Query `/balances?timestamp=T&limit=1`, if at the chosen snapshot timestamp `t1`, the account's balance snapshot is at `t2`, the response timestamp will be `t2`, instead of `t1`. 

### Additional context

_No response_

### Hedera network

testnet

### Version

v0.93.0

### Operating system

None

xin-hedera:
**Description**:

- use the timestamp from treasury account 0.0.2

**Related issue(s)**:

Fixes #7296 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-rest/__tests__/specs/balances/timestamp.json
+++ b/hedera-mirror-rest/__tests__/specs/balances/timestamp.json
@@ -43,6 +43,11 @@
         "balance": 60
       },
       {
+        "timestamp": 1566560001000000000,
+        "id": 7,
+        "balance": 770
+      },
+      {
         "timestamp": 1566560003000000000,
         "id": 2,
         "balance": 2
@@ -67,11 +72,6 @@
         "timestamp": 1566560003000000000,
         "id": 6,
         "balance": 666
-      },
-      {
-        "timestamp": 1566560003000000000,
-        "id": 7,
-        "balance": 777
       },
       {
         "timestamp": 1566560007000000000,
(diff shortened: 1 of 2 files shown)


hedera-mirror-rest/__tests__/specs/balances/timestamp.json
```diff
@@ -96,44 +96,65 @@
...
+        "links": {
+          "next": null
+        }
+      }
+    },
+    {
+      "url": "/api/v1/balances?timestamp=1566560004.000000000&limit=1",
+      "responseStatus": 200,
+      "responseJson": {
+        "timestamp": "1566560003.000000000",
```
xin-hedera, line 146: this new test case with url `/api/v1/balances?timestamp=1566560004.000000000&limit=1` can reproduce the issue.

running against 0.93.0. will produce `"timestamp":  "1566560001.000000000"` instead

mgoelswirlds (approved):
Looks good.

steven-sheehy (approved):

PR #7314, opened 2023-12-04
HIP-584 Historical: Load EVM Configuration based on block number

### LINKED ISSUE #7051: HIP-584: Historical Support - Load EVM Configuration based on block number
### Problem

For historical support we want to be able to dynamically populate in memory cache keeping block mapping  to **EVM** version and supported precompiles.
So that when we have **eth_call** with historical block we can check cache and load evm configuration corresponding to the how the state of the **evm** config was at the time of that block.

### Solution

Based on design doc 

We can enhance the current web3  **application.yml** with the following.
image

Then in MirrorNodeEvmProperties we can define 2 collections:
    1. `Map<String, Long> systemContractToBlockNumber`
    2. `Map<String, Long> evmVersionToBlockNumber`

Once we have added these two collections:
On application startup we can use ApplicationRunner or CommandLineRunner interfaces
both of those can execute a logic that executes immediately after the spring boot application is 
started.

In the run method of the interface we will fill those collection `systemContractToBlockNumber` 

and `evmVersionToBlockNumber`

After we fill them with the data in application.yml we can then build a map of ranges in MirrorNodeEvmProperties (could be in another class)

`Map<Long, EvmConfig> blockToConfiguration`, where `EvmConfig` is a record:
`record EvmConfig(String evmVersion, List<String> systemContracts)`

example `blockToConfiguration`:

1. 0 (block) → 0.30.0 → Empty list of contracts. 
2. 1000(block) → 0.34.0 → HTS, ERC
3. 10000(block) → 0.38.0 → HTS, ERC, PRNG, Exchange rate

user creates eth_call and passes block <1000 → we need to load configuration 1

user creates eth_call and passes block <10000 → we need to load configuration 2 

user creates eth_call and passes block ≥10000 → we need to load configuration 3

once we know which configuration we need we need to load the evm config based on it.

MirrorNodeEvmProperties, PrecompileMapper etc.

**IMPORTANT NOTE:** 
After research it looks like all HTS, ERC, exchange rate and PRNG precompiles were introduced before lowest
EVM version 30. This means that we won't need yaml configs for precompiles we can only use evm_versions.
Since all version 30, 34, 38 will support those precompiles.

### Alternatives

_No response_

kselveliev:
**Description**:

This PR implements the following feature:

When a historical call is made:

Based on the block number passed in the request body we need to load 
proper evm version during that block snapshot.

In order to do that the first thing we can to do is to create a yaml config.
Example yaml config:
image

This yml config is filled with values from mainnet.

Explanation of yaml:

**evm_version**: represents a list of evm version and in which block they appeared for the first time.
For example evm version 0.34 appeared in block 44029066

Use case: 

For example: A user passed block 40000000, we then read from our evm versions 
we see that for that block we need to load evm version 0.30.

Example 2: A used passed block 47117794, we read from our evm versions and we 
see that for that block we need to load evm version 0.34.

If we do not define yaml configs evm versions will be taken from **HederaNetwork** enum.

**IMPORTANT NOTE:**

During a call with historical block, custom evm loading is triggered in **HederaEvmTxProcessor**.java.

During **HederaEvmTxProcessor.process()** we load the proper evm version config in 

**getMessageProcessor**() method. 

**Classed removed in PR**

Removed **HTSPrecompiledContractAdapter**.java

Removed **MirrorHTSPrecompiledContract**.java

Added unit tests for the newly create classes:

**Related issue(s)**:

Fixes #7051

**Checklist**


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmMessageCallProcessorErcHts.java
```diff
@@ -0,0 +1,120 @@
...
+import java.util.Optional;
+import org.hyperledger.besu.evm.EVM;
+import org.hyperledger.besu.evm.frame.ExceptionalHaltReason;
+import org.hyperledger.besu.evm.frame.MessageFrame;
+import org.hyperledger.besu.evm.precompile.MainnetPrecompiledContracts;
+import org.hyperledger.besu.evm.precompile.PrecompileContractRegistry;
+import org.hyperledger.besu.evm.tracing.OperationTracer;
+
+@Named
+public class MirrorEvmMessageCallProcessorErcHts extends HederaEvmMessageCallProcessor {
```
IvanKavaldzhiev, line 47: We can keep only this processor for both `HTS` and `ERC` precompiles and have special checks in the `PrecompileMapper` and `HTSPrecompiledContract` for the ERC mappings.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmMessageCallProcessorErc.java
```diff
@@ -0,0 +1,120 @@
...
+import java.util.Optional;
+import org.hyperledger.besu.evm.EVM;
+import org.hyperledger.besu.evm.frame.ExceptionalHaltReason;
+import org.hyperledger.besu.evm.frame.MessageFrame;
+import org.hyperledger.besu.evm.precompile.MainnetPrecompiledContracts;
+import org.hyperledger.besu.evm.precompile.PrecompileContractRegistry;
+import org.hyperledger.besu.evm.tracing.OperationTracer;
+
+@Named
+public class MirrorEvmMessageCallProcessorErc extends HederaEvmMessageCallProcessor {
```
IvanKavaldzhiev, line 47: Probably we can use only `MirrorEvmMessageCallProcessorErcHTS`

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmMessageCallProcessorErcHtsExchangeRate.java
```diff
@@ -0,0 +1,118 @@
...
+import java.util.Optional;
+import org.hyperledger.besu.evm.EVM;
+import org.hyperledger.besu.evm.frame.ExceptionalHaltReason;
+import org.hyperledger.besu.evm.frame.MessageFrame;
+import org.hyperledger.besu.evm.precompile.MainnetPrecompiledContracts;
+import org.hyperledger.besu.evm.precompile.PrecompileContractRegistry;
+import org.hyperledger.besu.evm.tracing.OperationTracer;
+
+@Named
+public class MirrorEvmMessageCallProcessorErcHtsExchangeRate extends HederaEvmMessageCallProcessor {
```
IvanKavaldzhiev, line 47: We can extract a parent class - `MirrorEvmMessageCallProcessor` that contains the common methods to reduce code duplication. Then the different child invariants would just have different constructors.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/config/EvmConfiguration.java
```diff
@@ -164,16 +180,37 @@ MessageCallProcessor messageCallProcessor(final EVM evm, final PrecompileContrac
...
-                EVM_VERSION_0_30, () -> contractCreationProcessor, EVM_VERSION_0_34, () -> contractCreationProcessor);
+
+        Map<String, Provider<ContractCreationProcessor>> processorsMap = new HashMap<>();
+        processorsMap.put(EVM_VERSION_0_30_ERC, () -> contractCreationProcessor);
+        processorsMap.put(EVM_VERSION_0_30_ERC_HTS, () -> contractCreationProcessor);
+        processorsMap.put(EVM_VERSION_0_30_HTS_ERC_EXCHANGE_RATE, () -> contractCreationProcessor);
+        processorsMap.put(EVM_VERSION_0_30, () -> contractCreationProcessor);
+        processorsMap.put(EVM_VERSION_0_34, () -> contractCreationProcessor);
+        processorsMap.put(EVM_VERSION_0_38, () -> contractCreationProcessor);
+        processorsMap.put(EVM_VERSION_0_45, () -> contractCreationProcessor);
```
IvanKavaldzhiev, line 191: For now we can use only one instance of `contractCreationProcessor` and return it for each evm version type. In this way we can save memory.

We can define only a single bean with `Provider<ContractCreationProcessor>` type and if we need to pass around a Map, we can construct it in `HederaEvmTxProcessor`.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/config/PrecompilesHolderErc.java
```diff
@@ -0,0 +1,38 @@
...
+import com.hedera.services.store.contracts.precompile.ERCPrecompiledContract;
+import jakarta.inject.Named;
+import java.util.HashMap;
+import java.util.Map;
+import lombok.Getter;
+import org.hyperledger.besu.evm.precompile.PrecompiledContract;
+
+@Named
+@Getter
+public class PrecompilesHolderErc {
```
IvanKavaldzhiev, line 30: We can remove this class.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/config/PrecompilesHolderErcHts.java
```diff
@@ -0,0 +1,38 @@
...
+
+@Named
+@Getter
+public class PrecompilesHolderErcHts {
+
+    public final Map<String, PrecompiledContract> hederaPrecompiles;
+
+    PrecompilesHolderErcHts(final ErcHtsPrecompiledContract ercHtsPrecompiledContract) {
+        hederaPrecompiles = new HashMap<>();
+        hederaPrecompiles.put(EVM_HTS_PRECOMPILED_CONTRACT_ADDRESS, ercHtsPrecompiledContract);
```
IvanKavaldzhiev, line 36: We should use directly `HTSPrecompiledContract`.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/config/ServicesConfiguration.java
```diff
@@ -717,15 +731,47 @@ HTSPrecompiledContract htsPrecompiledContract(
...
                 mirrorNodeEvmProperties,
                 precompileMapper,
-                evmHTSPrecompiledContract,
+                store,
+                tokenAccessor,
+                precompilePricingUtils);
+    }
+
+    @Bean
+    ERCPrecompiledContract ercPrecompiledContract(
```
IvanKavaldzhiev, line 747: We should remove this bean.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/config/ServicesConfiguration.java
```diff
@@ -717,15 +731,47 @@ HTSPrecompiledContract htsPrecompiledContract(
...
+                evmInfrastructureFactory,
+                mirrorNodeEvmProperties,
+                precompileMapperErc,
+                store,
+                tokenAccessor,
+                precompilePricingUtils);
+    }
+
+    @Bean
+    ErcHtsPrecompiledContract ercHtsPrecompiledContract(
```
IvanKavaldzhiev, line 764: Use the already existing HTSPrecompiledContract instead.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
```diff
@@ -141,6 +152,15 @@ public class MirrorNodeEvmProperties implements EvmProperties {
     @DurationMin(seconds = 100)
     private Duration rateLimit = Duration.ofSeconds(100L);
 
+    @Getter
+    private Map<String, Long> systemPrecompiles = new HashMap<>();
+
+    @Getter
+    private TreeMap<String, Long> evmVersions = new TreeMap<>();
```
IvanKavaldzhiev, line 159: evmVersions -> evmVersionToBlockNumber

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
```diff
@@ -141,6 +152,15 @@ public class MirrorNodeEvmProperties implements EvmProperties {
...
+
+    @Getter
+    private Map<String, Set<String>> evmPrecompiles;
```
IvanKavaldzhiev, line 162: evmPrecompiles -> evmVersionToPrecompiles

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
```diff
@@ -212,4 +236,95 @@ public enum HederaNetwork {
...
+     *
+     * <p>The method matches the provided set of precompiles against known combinations,
+     * appending the relevant precompile names to the EVM version. If the set of precompiles
+     * matches a known combination, a specific version string key is returned. If the set does not match
+     * any predefined combination, a default version string is returned.
+     *
+     * <p>The method considers the following precompile combinations:
+     * <ul>
+     *   <li>Set containing only ERC</li>
+     *   <li>Set containing ERC and HTS</li>
```
IvanKavaldzhiev, line 296: It would be more correct to switch the order and all list items to start with HTS since this is the first group of precompiles that we have started supporting.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
```diff
@@ -212,4 +236,95 @@ public enum HederaNetwork {
...
+        Set<String> precompileSet = new HashSet<>(precompiles);
+
+        // we have different combinations for ERC, HTS, EXCHANGE_RATE AND PRNG only for v0.30 evm
+        // the higher evm version support them all
+        if (!evmVersion.equals(EVM_VERSION_0_30)) {
+            return evmVersion;
+        }
+
+        if (precompileSet.equals(Set.of(ERC))) {
+            return evmVersion + "-" + ERC;
```
IvanKavaldzhiev, line 318: evmVersion + "-" + HTS

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
```diff
@@ -212,4 +236,95 @@ public enum HederaNetwork {
...
+    public static String getEvmWithPrecompiles(String evmVersion, Set<String> precompiles) {
+        Set<String> precompileSet = new HashSet<>(precompiles);
+
+        // we have different combinations for ERC, HTS, EXCHANGE_RATE AND PRNG only for v0.30 evm
+        // the higher evm version support them all
+        if (!evmVersion.equals(EVM_VERSION_0_30)) {
+            return evmVersion;
+        }
+
+        if (precompileSet.equals(Set.of(ERC))) {
```
IvanKavaldzhiev, line 317: Set.of(HTS)

hedera-mirror-web3/src/main/java/com/hedera/node/app/service/evm/contracts/execution/HederaEvmTxProcessor.java
```diff
@@ -190,11 +192,19 @@ protected MessageFrame buildInitialFrame(
...
     }
 
-    private AbstractMessageProcessor getMessageProcessor(final MessageFrame.Type type, final String evmVersion) {
+    private AbstractMessageProcessor getMessageProcessor(final MessageFrame.Type type, String evmVersion) {
+
+        if (ContractCallContext.get().useHistorical()) {
+            MirrorNodeEvmProperties evmProperties = ((MirrorNodeEvmProperties) dynamicProperties);
+            Set<String> precompilesAvailable = evmProperties.getPrecompilesAvailableAtBlock(
+                    ContractCallContext.get().getRecordFile().getIndex(), evmVersion);
+            evmVersion = MirrorNodeEvmProperties.getEvmWithPrecompiles(evmVersion, precompilesAvailable);
```
IvanKavaldzhiev, line 205: Why do we need the evm version here to be appended with precompiles?

hedera-mirror-web3/src/main/java/com/hedera/services/store/contracts/precompile/ERCPrecompiledContract.java
```diff
@@ -0,0 +1,156 @@
...
+import com.hedera.services.store.contracts.precompile.codec.FunctionParam;
+import com.hedera.services.store.contracts.precompile.impl.ApprovePrecompile;
+import com.hedera.services.store.contracts.precompile.utils.PrecompilePricingUtils;
+import com.hedera.services.store.models.Id;
+import com.hedera.services.store.models.NftId;
+import com.hedera.services.utils.EntityIdUtils;
+import java.util.function.UnaryOperator;
+import org.apache.tuweni.bytes.Bytes;
+
+public class ERCPrecompiledContract extends PrecompiledContractBase {
```
IvanKavaldzhiev, line 39: Shouldn't make such deviations in the copied code from services. We can utilize the `HTSPrecompiledContract`.

hedera-mirror-web3/src/main/java/com/hedera/services/store/contracts/precompile/HTSPrecompiledContract.java
```diff
@@ -49,213 +32,36 @@
...
  */
-public class HTSPrecompiledContract implements HTSPrecompiledContractAdapter {
-
-    public static final TupleType redirectType = TupleType.parse("(int32,bytes)");
-
-    public static final PrecompileContractResult INVALID_DELEGATE = new PrecompileContractResult(
-            null, true, MessageFrame.State.COMPLETED_FAILED, Optional.of(ExceptionalHaltReason.PRECOMPILE_ERROR));
-    private static final Logger log = LogManager.getLogger(HTSPrecompiledContract.class);
-    private static final Bytes STATIC_CALL_REVERT_REASON = Bytes.of("HTS precompiles are not static".getBytes());
+public class HTSPrecompiledContract extends PrecompiledContractBase {
```
IvanKavaldzhiev, line 51: Based on my other comments, I believe we should revert all changes in this file.

hedera-mirror-web3/src/main/java/com/hedera/services/store/contracts/precompile/PrecompileMapper.java
```diff
@@ -25,86 +26,7 @@ public class PrecompileMapper {
```
IvanKavaldzhiev, line 26: Should keep this `PrecompileMapper` as it is and reuse it.

hedera-mirror-web3/src/main/java/com/hedera/services/store/contracts/precompile/PrecompiledContractBase.java
```diff
@@ -0,0 +1,296 @@
...
+import java.util.function.UnaryOperator;
+import org.apache.commons.lang3.tuple.Pair;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.apache.tuweni.bytes.Bytes;
+import org.hyperledger.besu.datatypes.Address;
+import org.hyperledger.besu.evm.frame.ExceptionalHaltReason;
+import org.hyperledger.besu.evm.frame.MessageFrame;
+
+public abstract class PrecompiledContractBase extends EvmHTSPrecompiledContract {
```
IvanKavaldzhiev, line 60: We can get this code back to `HTSPrecompiledContract`.

hedera-mirror-web3/src/main/java/com/hedera/services/store/contracts/precompile/utils/PrecompileMapperUtils.java
```diff
@@ -0,0 +1,120 @@
...
+package com.hedera.services.store.contracts.precompile.utils;
+
+import com.hedera.services.store.contracts.precompile.AbiConstants;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class PrecompileMapperUtils {
+
+    public static final Set<Integer> ERC_PRECOMPILE_SELECTORS = Set.of(
```
IvanKavaldzhiev, line 26: We can use this `Set` for excluding ERC precompile execution in `MirrorEvmMessageCallProcessor` if we have a block that does not support them.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmMessageCallProcessorErcHts.java
IvanKavaldzhiev: We can follow the same logic for inclusion/exclusion for the HRC precompiles.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmMessageCallProcessorErc.java
kselveliev: Comment addressed, we use MirrorEvmMessageCallProcessorHtsErc now.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/config/PrecompilesHolderErc.java
kselveliev: Comment addressed. This class is removed.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/config/PrecompilesHolderErcHts.java
kselveliev: Comment addressed, we now only use HTSPrecompiledContract

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/config/ServicesConfiguration.java
kselveliev: Comment addressed.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
kselveliev: MirrorNodeEvmProperties is bound to the yaml config, if we want to do that name change we will need to change naming in yaml file. system_precompiles in yaml is read as systemPrecompiles in java class.

image

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
```diff
@@ -141,6 +152,15 @@ public class MirrorNodeEvmProperties implements EvmProperties {
     @DurationMin(seconds = 100)
     private Duration rateLimit = Duration.ofSeconds(100L);
 
+    @Getter
+    private Map<String, Long> systemPrecompiles = new HashMap<>();
```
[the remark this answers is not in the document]
kselveliev: Same as [above](https://github.com/hashgraph/hedera-mirror-node/pull/7314#discussion_r1417425401)

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
```diff
@@ -212,4 +236,95 @@ public enum HederaNetwork {
...
+     * Returns a string key representing the EVM version with specified precompiles.
+     *
+     * <p>The method matches the provided set of precompiles against known combinations,
+     * appending the relevant precompile names to the EVM version. If the set of precompiles
+     * matches a known combination, a specific version string key is returned. If the set does not match
+     * any predefined combination, a default version string is returned.
+     *
+     * <p>The method considers the following precompile combinations:
+     * <ul>
+     *   <li>Set containing only ERC</li>
```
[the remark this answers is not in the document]
kselveliev: Comment addressed.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
kselveliev: Comment addressed. After the latest refactoring the combinations of ERC and HTS changes are removed since they are both introduced before evm 0.30

hedera-mirror-web3/src/main/java/com/hedera/services/store/contracts/precompile/ERCPrecompiledContract.java
kselveliev: Comment addressed. We are only using HTSPrecompiledContract now.

hedera-mirror-web3/src/main/java/com/hedera/services/store/contracts/precompile/PrecompileMapper.java
kselveliev: Comment addressed, we are keeping the PrecompileMapper.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
kselveliev: Comment addressed. This javadoc is changed after refactoring.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/config/ServicesConfiguration.java
kselveliev: Comment addressed. We are only using HTSPrecompiledContract now.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
kselveliev: Comment addressed. This combination is removed after the latest refactoring.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
kselveliev: Comment addressed

hedera-mirror-web3/src/main/java/com/hedera/services/store/contracts/precompile/HTSPrecompiledContract.java
kselveliev: Comment addressed. I reverted the changes in HTSPrecompiledContract

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmMessageCallProcessorExchangeRate.java
```diff
@@ -0,0 +1,39 @@
...
+
+@Named
+public class MirrorEvmMessageCallProcessorExchangeRate extends MirrorEvmMessageCallProcessor {
+
+    public MirrorEvmMessageCallProcessorExchangeRate(
+            final AbstractAutoCreationLogic autoCreationLogic,
+            final EntityAddressSequencer entityAddressSequencer,
+            final EVM evm,
+            final PrecompileContractRegistry precompiles,
+            final PrecompilesHolderHtsErcExchangeRate precompilesHolderHtsErc,
```
IvanKavaldzhiev, line 35: precompilesHolderHtsErc -> precompilesHolderHtsErcExchangeRate

kselveliev: Comment addressed

hedera-mirror-web3/src/main/java/com/hedera/node/app/service/evm/contracts/execution/HederaEvmTxProcessor.java
kselveliev: This code was removed after latest refactoring

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
```diff
@@ -141,6 +145,9 @@ public class MirrorNodeEvmProperties implements EvmProperties {
     @DurationMin(seconds = 100)
     private Duration rateLimit = Duration.ofSeconds(100L);
 
+    @Getter
+    private TreeMap<String, Long> evmVersions = new TreeMap<>();
```
steven-sheehy, line 149: Invert the key and value then you can use `evmVersions.floorEntry(blockNum)` to efficiently find the value instead of looping.

```suggestion
    private NavigableMap<Long, String> evmVersions = new TreeMap<>();
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
steven-sheehy: Prefer `evmVersions`

hedera-mirror-web3/src/main/resources/application.yml
```diff
@@ -9,6 +9,12 @@ hedera:
         sslMode: DISABLE
         statementTimeout: 10000
         username: mirror_web3
+      evm:
+        evm_versions:
+          v0.30:
+          v0.34:
+          v0.38:
+          v0.45:
```
steven-sheehy, line 17: We should omit empty config from yaml. Move mainnet config to java. These are the values for mainnet which we should populate on HederaNetwork:
```
mirror_node=> select hapi_version_minor, min(consensus_end), max(consensus_end) from record_file where hapi_version_major = 0 and hapi_version_minor in (30, 34, 38, 45) group by hapi_version_minor;
 hapi_version_minor |         min         |         max         
--------------------+---------------------+---------------------
                 30 | 1666368347153861003 | 1670605199011072003
                 34 | 1675962001984524003 | 1680282000110196003
                 38 | 1686243920984236003 | 1689091200717666721
(3 rows)
```

hedera-mirror-web3/src/test/resources/application.yml
```diff
@@ -7,6 +7,12 @@ hedera:
         name: ${embedded.postgresql.schema}
         username: ${embedded.postgresql.user}
         password: ${embedded.postgresql.password}
+      evm:
+        evm_versions:
+          v0.30: 1000
+          v0.34: 2000
+          v0.38: 3000
+          v0.45: 4000
```
steven-sheehy, line 15: Why do we need these values here? For tests that specifically test it they should set it in `MirrorNodeEvmProperties`.

hedera-mirror-web3/src/main/java/com/hedera/services/store/contracts/precompile/HTSPrecompiledContract.java
```diff
@@ -71,20 +70,15 @@
...
- * consuming libraries like smart-contract-service it would be easier to delete the code base inside com.hedera.services
- * package.
+ * This class is a modified thread-safe copy of HTSPrecompiledContract from hedera-services repo.
  * <p>
  * Differences with the original class: 1. Use abstraction for the state by introducing {@link Store} interface. 2. Use
  * workaround to execute read only precompiles via calling ViewExecutor and RedirectViewExecutors, thus removing the
  * need of having separate precompile classes. 3. All stateful fields are extracted into {@link ContractCallContext} and the class is converted to a singleton bean
  */
-public class HTSPrecompiledContract implements HTSPrecompiledContractAdapter {
+public class HTSPrecompiledContract extends EvmHTSPrecompiledContract {
```
steven-sheehy, line 81: In the future, please put refactoring unrelated to the ticket like this in separate PRs so it's easier to review.

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/config/EvmConfiguration.java
```diff
@@ -157,18 +161,21 @@ CacheManager cacheManagerRecordFileEarliest() {
     }
 
     @Bean
-    Map<String, Provider<ContractCreationProcessor>> contractCreationProcessors(
+    public Provider<ContractCreationProcessor> contractCreationProcessorProvider(
```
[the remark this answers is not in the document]
kselveliev: You're right, I've missed that. We don't need to increase scope. I reverted that change

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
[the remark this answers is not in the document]
kselveliev: Comment addressed. Using NavigableMap now.

kselveliev: Comment addressed. I've inverted the key and value for better search.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
```diff
@@ -212,4 +223,30 @@ public enum HederaNetwork {
...
+     *
+     * @param blockNumber The block number to check.
+     * @return The key of the suitable EVM version for the given block number, or null if none found.
+     */
+    public String getEvmVersionForBlock(Long blockNumber) {
+        if (CollectionUtils.isEmpty(evmVersions)) {
+            // In case we don't have any specified evm versions in yml configuration point to default version
+            return EVM_VERSION;
+        }
+        String closestVersion = evmVersions.firstEntry().getKey(); // Default to the lowest version
```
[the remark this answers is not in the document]
kselveliev: If we consider evm version 30 as block 0 and after inverting the key and value as suggested above. Just doing evmVersions.floorEntry(blockNumber); should do the job. I've done that now.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
```diff
@@ -212,4 +223,30 @@ public enum HederaNetwork {
...
+
+    /**
+     * Determines the suitable EVM version based on the given block number. Returns the highest version whose block
+     * number is less than or equal to the input block number.
+     *
+     * @param blockNumber The block number to check.
+     * @return The key of the suitable EVM version for the given block number, or null if none found.
+     */
+    public String getEvmVersionForBlock(Long blockNumber) {
+        if (CollectionUtils.isEmpty(evmVersions)) {
```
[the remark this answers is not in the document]
kselveliev: I added a hardcoded evm versions map in HederaNetwork for mainnet.
Now it first checks if we have yaml config values - if yes use them.
If not  -> use HederaNetwork hardcoded values 
if HederaNetwork values are empty use default version.

hedera-mirror-web3/src/main/resources/application.yml
kselveliev: Comment addressed. I've omited the yaml configs and populated the values in HederaNetwork for mainnet. Since I do no see any result for 45 version I did not add it in the HederaNetwork evm versions map.

hedera-mirror-web3/src/test/resources/application.yml
kselveliev: You are right, we don't need that. I've removed this config.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
```diff
@@ -141,6 +153,9 @@ public class MirrorNodeEvmProperties implements EvmProperties {
     @DurationMin(seconds = 100)
     private Duration rateLimit = Duration.ofSeconds(100L);
 
+    @Getter
+    private NavigableMap<Long, String> evmVersions = new TreeMap<>();
```
steven-sheehy, line 157: nit: Sort inline with other fields

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
```diff
@@ -204,12 +223,52 @@ public long exchangeRateGasReq() {
...
+        NavigableMap<Long, String> evmVersionsMap = this.evmVersions;
+        if (CollectionUtils.isEmpty(evmVersionsMap)) {
+            // In case we don't have any specified evm versions in yml configuration use HederaNetwork values
+            evmVersionsMap = network.evmVersions;
+        }
+
+        if (CollectionUtils.isEmpty(evmVersionsMap)) {
+            // In case we don't have any specified evm versions in yml or HederaNetwork return default evm version
+            return EVM_VERSION;
+        }
```
steven-sheehy, line 266: It would be cleaner to separate getting the map and looking up the value from the map into separate methods. Below I override `getEvmVersions()` so it always has a non-null/non-empty map.

```java
@Override
public NavigableMap<Long, String> getEvmVersions() {
    if (CollectionUtils.isNotEmpty(evmVersions)) {
        return evmVersions;
    }

    if (CollectionUtils.isNotEmpty(network.evmVersions)) {
        return network.evmVersions;
    }

    return Map.of(0L, EVM_VERSION); // Can extract to a constant
}

String getEvmVersionForBlock(long blockNumber) {
    return getEvmVersions().floorEntry(blockNumber);
}
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/config/EvmConfiguration.java
```diff
@@ -82,6 +83,12 @@ public class EvmConfiguration {
     public static final String CACHE_NAME_TOKEN_ALLOWANCE = "tokenAllowance";
     public static final String EVM_VERSION_0_30 = "v0.30";
     public static final String EVM_VERSION_0_34 = "v0.34";
+    public static final String EVM_VERSION_0_38 = "v0.38";
+    public static final String EVM_VERSION_0_45 = "v0.45";
+    public static final Long GENESIS_BLOCK = 0L;
+    public static final Long EVM_VERSION_34_START_BLOCK = 44029066L;
+    public static final Long EVM_VERSION_38_START_BLOCK = 49117794L;
```
steven-sheehy, line 90: Don't think these need to be static class constants or are relevant to this class. Should just embed inline in HederaNetwork. Tests that validate offsets should set a custom evm version map.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
```diff
@@ -204,12 +223,52 @@ public long exchangeRateGasReq() {
...
 
         private final byte[] ledgerId;
         private final Bytes32 chainId;
+        private final NavigableMap<Long, String> evmVersions;
+
+        private static NavigableMap<Long, String> mainnetEvmVersionsMap() {
+            NavigableMap<Long, String> evmVersionsMap = new TreeMap<>();
+            evmVersionsMap.put(GENESIS_BLOCK, EVM_VERSION_0_30);
+            evmVersionsMap.put(EVM_VERSION_34_START_BLOCK, EVM_VERSION_0_34);
+            evmVersionsMap.put(EVM_VERSION_38_START_BLOCK, EVM_VERSION_0_38);
```
steven-sheehy, line 239: Prefer constants inline here for easier readability.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmPropertiesTest.java
```diff
@@ -51,4 +76,39 @@ void correctPropertiesEvaluation() {
...
+
+    @ParameterizedTest
+    @MethodSource("blockNumberToEvmVersionProvider")
+    void getEvmVersionForBlockFromConfig(Long blockNumber, String expectedEvmVersion) {
+        // given
+        NavigableMap<Long, String> evmVersions = new TreeMap<>();
+        evmVersions.put(GENESIS_BLOCK, EVM_VERSION_30);
+        evmVersions.put(EVM_VERSION_34_START_BLOCK, EVM_VERSION_34);
+        evmVersions.put(EVM_VERSION_38_START_BLOCK, EVM_VERSION_38);
+        properties.setEvmVersions(evmVersions);
```
steven-sheehy, line 88: I don't think we should test with a repeat of the config for getEvmVersionForBlockFromHederaNetwork. Use a different map that you create to test.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmPropertiesTest.java
```diff
@@ -51,4 +76,39 @@ void correctPropertiesEvaluation() {
...
+
+    private static Stream<Arguments> blockNumberToEvmVersionProvider() {
+        return Stream.of(
+                Arguments.of(0L, EVM_VERSION_30),
+                Arguments.of(1L, EVM_VERSION_30),
+                Arguments.of(EVM_VERSION_34_START_BLOCK - 1, EVM_VERSION_30),
+                Arguments.of(EVM_VERSION_34_START_BLOCK, EVM_VERSION_34),
+                Arguments.of(EVM_VERSION_38_START_BLOCK - 1, EVM_VERSION_34),
+                Arguments.of(EVM_VERSION_38_START_BLOCK, EVM_VERSION_38),
+                Arguments.of(EVM_VERSION_38_START_BLOCK + 1000, EVM_VERSION_38));
```
steven-sheehy, line 112: Instead of this hardcoded approach, we can iterate over map entries and for each value try with key, key - 1, and key + 1. For key-1 we assert value.compareTo(currentValue) < 0.

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
steven-sheehy: Note: getEvmVersionForBlock should be package private instead of public

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
```diff
@@ -204,12 +223,52 @@ public long exchangeRateGasReq() {
...
+
+        if (CollectionUtils.isEmpty(evmVersionsMap)) {
+            // In case we don't have any specified evm versions in yml or HederaNetwork return default evm version
+            return EVM_VERSION;
+        }
+        Entry<Long, String> evmEntry = evmVersionsMap.floorEntry(blockNumber);
+        if (evmEntry != null) {
+            return evmEntry.getValue();
+        } else {
+            return EVM_VERSION; // Return default version if no entry matches the block number
```
steven-sheehy, line 271: Needs test coverage

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
```diff
@@ -204,12 +223,52 @@ public long exchangeRateGasReq() {
...
         private final byte[] ledgerId;
         private final Bytes32 chainId;
+        private final NavigableMap<Long, String> evmVersions;
+
+        private static NavigableMap<Long, String> mainnetEvmVersionsMap() {
+            NavigableMap<Long, String> evmVersionsMap = new TreeMap<>();
+            evmVersionsMap.put(GENESIS_BLOCK, EVM_VERSION_0_30);
+            evmVersionsMap.put(EVM_VERSION_34_START_BLOCK, EVM_VERSION_0_34);
+            evmVersionsMap.put(EVM_VERSION_38_START_BLOCK, EVM_VERSION_0_38);
+            return evmVersionsMap;
```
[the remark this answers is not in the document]
kselveliev: Comment addressed. Now returning unmodifiableNavigableMap.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
kselveliev: Not entirely sure what you meant by this. I just moved the evmVersions map next to evmVersion variable.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
kselveliev: Comment addressed. I applied the suggestion. Just did not use the @Override annotation since I encountered problems with lombok if I used it. Also had to add if else logic in getEvmVersionForBlock since getEvmVersions().floorEntry(blockNumber); returns an entry not a string.

kselveliev: Comment addressed. I just did not use override since I had problems with lombok using it. And in getEvmVersionForBlock I had to add if else logic since getEvmVersions().floorEntry(blockNumber); returns an entry and we want the value of the entry.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/config/EvmConfiguration.java
kselveliev: Comment addressed. I removed the constants from EvmConfiguration.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
kselveliev: Comment addressed.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmPropertiesTest.java
kselveliev: I believe here you meant the getEvmVersionForBlockFromConfig method where we can load a custom map.
I added a custom map test for it.
For the HederaNetwork enum I believe it's harder to test with custom map since it is an enum.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmPropertiesTest.java
kselveliev: I modified the provider so that it test with key-1, key, and key + 1 for all entries in evmVersions

steven-sheehy (approved):
LGTM

steven-sheehy:
Just needs @IvanKavaldzhiev approval since he had some comments, then can merge.

IvanKavaldzhiev (approved):
LGTM

PR #7327, opened 2023-12-05
HIP-584: Provide proper estimate on contract deploy

### LINKED ISSUE #7249: Inaccurate Gas Estimation for Contract Deployment on Hedera Network
### Description

I am experiencing an issue with gas estimation when deploying a contract to the Hedera network using Hardhat and ethers.js. The `eth_estimateGas` method returns a gas value of 61060, but deploying the contract with this gas limit results in a failure due to insufficient gas. Presumably the hedera relay returns the gas estimation value as is from the mirror node, hence why I am reporting the issue here.

### Steps to reproduce

1. Set up a Hardhat environment configured for the Hedera network.
2. Use ethers.js with the following contract deployment code:
    ```javascript
    import {
      abi as UV3MATH_ABI,
      bytecode as UV3MATH_BYTECODE,
    } from "UV3Math.json";
    
    import { ethers } from "hardhat";
    
    const deployer = (await ethers.getSigners())[0];

    const uV3MathLibraryFactory = new ethers.ContractFactory(UV3MATH_ABI, UV3MATH_BYTECODE, deployer);
    const deployTx = uV3MathLibraryFactory.getDeployTransaction();
    const estimatedGas = await deployer.estimateGas(deployTx); // returns 61_060
    console.log("Estimated Gas:", estimatedGas.toString());

    const uV3MathLibrary = await uV3MathLibraryFactory.deploy({ gasLimit: estimatedGas });
    await uV3MathLibrary.deployTransaction.wait(blockConfirmations);
    ```
3. Observe the estimated gas and attempt the contract deployment.
4. Observe the deployment transaction fails with the following error:
```
ProviderError: [Request ID: 327cb8b3-a1d8-4779-a247-9ac7fa0555e8] Transaction gas limit provided '61060' is insufficient of intrinsic gas required '76232'
```

### Additional context

I have attach the ABI and bytecode file for the contract (`UV3Math.json`) to this issue.

If you want to replicate the issue in hardhat using the same deps versions:
"@nomiclabs/hardhat-ethers": "^2.2.1",
"ethers": "^5.7.2",
"hardhat": "^2.19.1",

[UV3Math.json](https://github.com/hashgraph/hedera-mirror-node/files/13452911/UV3Math.json)

### Hedera network

testnet

### Version

mirror-node@0.92.0

### Operating system

None

IvanKavaldzhiev:
**Description**:

Currently when we deploy a contract on consensus node we don't take into account the size of the init bytecode. Mirror node copied this logic and kept the same behaviour.

The `json-rpc-relay`, however, is taking into account the input size and performs a pre-check on contract deploy. When we try to deploy a contract using the estimate from the mirror node sometimes we hit an issue, since the pre-check on relay side has a higher expected gas value than what comes in as an estimate from the mirror-node.

This PR fixes the gas estimate on mirror-node side by taking into account the init bytecode size. In this way we would get an estimate that is between 5%-20% higher than what the EVM would return as used gas and this value should pass the `json-rpc-relay` pre-check phase.
**Related issue(s)**:

Fixes #7249 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-web3/src/main/java/com/hedera/node/app/service/evm/contracts/execution/HederaEvmTxProcessor.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/node/app/service/evm/contracts/execution/HederaEvmTxProcessor.java
@@ -108,7 +108,7 @@ public class HederaEvmTxProcessor {
             final Address mirrorReceiver,
             final boolean contractCreation) {
         final var blockValues = blockMetaSource.computeBlockValues(gasLimit);
-        final var intrinsicGas = gasCalculator.transactionIntrinsicGasCost(Bytes.EMPTY, contractCreation);
+        final var intrinsicGas = gasCalculator.transactionIntrinsicGasCost(payload, contractCreation);
         final var gasAvailable = gasLimit - intrinsicGas;
 
         final var valueAsWei = Wei.of(value);


georgi-l95 (approved):
LG, it fixes the issue with gas miscalculation between relay and mirror-node.

bilyana-gospodinova (approved):
LGTM

steven-sheehy (approved):

PR #7336, opened 2023-12-05
Improve transaction queries in v2

### LINKED ISSUE #6933: Improve transaction queries in v2
### Problem

The listed end points all seem to be suffering from the same performance issue in v2.

Appears they are struggling with the transactions query
```
with timestampFilter as (
      SELECT coalesce(t.consensus_timestamp, ctl.consensus_timestamp, ttl.consensus_timestamp) AS consensus_timestamp
      FROM (select consensus_timestamp
    from transaction as t
    WHERE t.payer_account_id = 1833905
      and (t.nonce = 0 or
  t.parent_consensus_timestamp is not null)
    order by t.consensus_timestamp desc
    limit 25 ) AS t
      FULL OUTER JOIN (
    SELECT DISTINCT ctl.consensus_timestamp AS consensus_timestamp
    FROM crypto_transfer AS ctl

   WHERE ctl.entity_id = 1833905
    ORDER BY ctl.consensus_timestamp desc
    limit 25 ) AS ctl
      ON t.consensus_timestamp = ctl.consensus_timestamp
      FULL OUTER JOIN (
    SELECT DISTINCT ttl.consensus_timestamp AS consensus_timestamp
    FROM token_transfer AS ttl

    WHERE ttl.account_id = 1833905
    ORDER BY ttl.consensus_timestamp desc
    limit 25 ) AS ttl
      ON coalesce(t.consensus_timestamp, ctl.consensus_timestamp) = ttl.consensus_timestamp
      limit 25 ),
      timestamp_range as (select min(consensus_timestamp) as min, max(consensus_timestamp) as max from timestampFilter),
      tlist as (select t.charged_tx_fee,t.consensus_timestamp,t.entity_id,t.max_fee,t.memo,t.nft_transfer,t.node_account_id,t.nonce,t.parent_consensus_timestamp,t.payer_account_id,t.result,t.scheduled,t.transaction_bytes,t.transaction_hash,t.type,t.valid_duration_seconds,t.valid_start_ns,t.index
                    from timestamp_range tr, transaction t
                    join timestampFilter tf on t.consensus_timestamp =
        tf.consensus_timestamp
        WHERE t.consensus_timestamp >= tr.min
        AND t.consensus_timestamp <= tr.max
                    order by t.consensus_timestamp desc
                    ),
c_list as (
      select jsonb_agg(jsonb_build_object(
    'amount', amount,
    'entity_id', ctr.entity_id,
    'is_approval', is_approval
  ) order by ctr.entity_id, amount) as ctr_list, ctr.consensus_timestamp
      from timestamp_range tr,  crypto_transfer ctr
      join tlist on ctr.consensus_timestamp = tlist.consensus_timestamp
      and ctr.payer_account_id = tlist.payer_account_id
      WHERE ctr.consensus_timestamp >= tr.min
      AND ctr.consensus_timestamp <= tr.max group by ctr.consensus_timestamp
  ),
t_list as (
    select jsonb_agg(jsonb_build_object(
    'account_id', account_id,
    'amount', amount,
    'token_id', token_id,
    'is_approval', is_approval
  ) order by token_id, account_id) as ttr_list, tk_tr.consensus_timestamp
    from timestamp_range tr,  token_transfer tk_tr
    join tlist on tk_tr.consensus_timestamp = tlist.consensus_timestamp
    and tk_tr.payer_account_id = tlist.payer_account_id
    WHERE tk_tr.consensus_timestamp >= tr.min
      AND tk_tr.consensus_timestamp <= tr.max group by tk_tr.consensus_timestamp
  ),
transfer_list as (
    select coalesce(t.consensus_timestamp, ctrl.consensus_timestamp, ttrl.consensus_timestamp) AS consensus_timestamp,
      ctrl.ctr_list,
      ttrl.ttr_list,
      t.charged_tx_fee,t.entity_id,t.max_fee,t.memo,t.nft_transfer,t.node_account_id,t.nonce,t.parent_consensus_timestamp,t.payer_account_id,t.result,t.scheduled,t.transaction_bytes,t.transaction_hash,t.type,t.valid_duration_seconds,t.valid_start_ns,t.index
    from tlist t
    full outer join c_list ctrl on t.consensus_timestamp = ctrl.consensus_timestamp
    full outer join t_list ttrl on t.consensus_timestamp = ttrl.consensus_timestamp
  )
    SELECT
    t.charged_tx_fee,
t.consensus_timestamp,
t.entity_id,
t.max_fee,
t.memo,
t.nft_transfer,
t.node_account_id,
t.nonce,
t.parent_consensus_timestamp,
t.payer_account_id,
t.result,
t.scheduled,
t.transaction_bytes,
t.transaction_hash,
t.type,
t.valid_duration_seconds,
t.valid_start_ns,
t.index,
t.ctr_list AS crypto_transfer_list,
t.ttr_list AS token_transfer_list

    FROM transfer_list t
    ORDER BY t.consensus_timestamp desc
```

1. /accounts/{id}
2. /accounts/{id}?timestamp=lte:{timestamp}
3. /transactions
4. /transactions?account.id={accountId}
5. /transactions?transactionType={transactionType} (this one has additional complexities due to the type filter - may warrant a separate issue after the completion of this one)

Note:
the `/accounts/{id}?timestamp=lte:{timestamp}` has additional issues with the timestamp query
```
with latest_token_balance as (select account_id, balance, token_id, consensus_timestamp
                                       from token_balance)

                select e.alias,
e.auto_renew_period,
e.created_timestamp,
e.decline_reward,
e.deleted,
e.ethereum_nonce,
e.evm_address,
e.expiration_timestamp,
e.id,
e.key,
e.max_automatic_token_associations,
e.memo,
e.receiver_sig_required,
e.staked_account_id,
e.staked_node_id,
e.stake_period_start,
e.type,
e.timestamp_range,
(case when es.pending_reward is null then 0
         when e.decline_reward is true or coalesce(e.staked_node_id, -1) = -1 then 0
         when e.stake_period_start >= es.end_stake_period then 0
         else es.pending_reward
    end) as pending_reward,
(select json_agg(jsonb_build_object('token_id', token_id, 'balance', balance)) ::jsonb
                        from (select token_id, balance
                              from latest_token_balance
                              where account_id = 1833905 and consensus_timestamp  =  1675441800763240000
                              order by token_id asc
        limit 1000) as account_token_balance) as token_balances,
(case
            when upper(e.timestamp_range) is null
              then COALESCE(ab.consensus_timestamp, (select max(consensus_end) from record_file))
            else COALESCE(ab.consensus_timestamp, upper(e.timestamp_range))
          end) as consensus_timestamp,
COALESCE(ab.balance, e.balance) as balance
                from (SELECT *
                      from entity e
                      where e.type in ('ACCOUNT', 'CONTRACT') and e.id = 1833905 and e.timestamp_range && '[,1675441982600000000]'
                      UNION ALL
                      SELECT *
                      from entity_history e
                      where e.type in ('ACCOUNT', 'CONTRACT') and e.id = 1833905 and e.timestamp_range && '[,1675441982600000000]'
                      order by timestamp_range desc limit 1) e
                         left join entity_stake es on es.id = e.id
                         left join account_balance ab on ab.account_id = e.id and ab.consensus_timestamp = 1675441800763240000 where (es.id = 1833905 OR es.id IS NULL)
```

### Solution

Investigate query bottle necks and optimize accordingly

### Alternatives

_No response_

jascks:
**Description**:
NOTE: This PR is still preliminary in development. It's purpose now is to get eyes on the work done thus far and the direction taken. Integration tests are failing on the expect fields in the SQL query. I'm looking into that. In addition, tests with `config.query.bindTimestampRange` set to true need to be developed.

This PR significantly changes how transaction results are retrieved from the database. Both the `/accounts` and `/transactions` endpoints utilize common code (in `transactions.js`) to achieve this. The once large query used (including an inner and outer query construction) has been replaced by two separate, serial queries. The first retrieves the relevant consensus timestamps and payer IDs from the `transaction`, `crypto_transfer` and `token_transfer` tables. The second query is built based on these retrieved values.

In the V2/Citus future, those three tables are distributed (sharded) by the payer ID, and partitioned by the consensus timestamp. Identifying the payer IDs and bounding the timestamp range is important to achieve the best performance. Therefore, timestamp bounds are computed based on what has been provided by the client. If an upper and lower bound are not clearly identified, then they are calculated, to limit the number of partitions accessed.

Note that for multiple equals queries, the values are now ORed rather than ANDed. In the latter case, specifying two or more different equals values resulted in zero transactions being returned. Being ORed, transactions may now be returned.

This PR improves  V2 performance for the following REST API endpoints:
1. /accounts/{id}
2. /accounts/{id}?timestamp=lte:{timestamp}
3. /transactions
4. /transactions?account.id={accountId}
5. /transactions?transactionType={transactionType} (this one has additional complexities due to the type filter - may warrant a separate issue after the completion of this one)

Note:
the `/accounts/{id}?timestamp=lte:{timestamp}` has additional issues with the timestamp query

**Related issue(s)**:

Fixes #6933

**Notes for reviewer**:

**Checklist**


hedera-mirror-rest/transactions.js
```diff
@@ -567,86 +524,191 @@ const getTransactionsInnerQuery = function (
...
+ *
+ * @param {{Object}[]} timestampQueryRows the results of the relevant timestamps query, which is rows
+ * consisting of consensus_timestamp and payer_account_id attributes. At least one row is required.
+ * @param {String} order Sorting order
+ * @param {Number} limit Max number of rows
+ * @return {Promise} Promise returning the transaction summary results for the provided transaction timestamp rows
+ */
+const getTransactionsSummary = async (timestampQueryRows, order, limit) => {
+  // Nothing with which to query
+  if (_.isNil(timestampQueryRows) || timestampQueryRows.length === 0) {
```
edwin-greene, line 549: This should work to simplify the undefined and empty check:
```suggestion
  if (_.isEmpty(timestampQueryRows)) {
```

jascks: Thanks! That willl be in the next push.

hedera-mirror-rest/utils.js
```diff
@@ -660,6 +660,41 @@ const parsePublicKeyQueryParam = (parsedQueryParams, columnName) => {
...
 };
 
+/**
+ * If enabled in config, ensure the returned timestamp range is fully bound; contains both a begin
+ * and end timestamp value. The provided Range is not modified. If changes are made a copy is returned.
+ *
+ * @param {Range} tsRange timestamp range, typically based on query parameters
+ * @return {Range} fully bound timestamp range
+ */
+const bindTimestampRange = function (tsRange) {
```
xin-hedera, line 670: we generally use arrow function

```suggestion
const bindTimestampRange = (tsRange) => {
```

hedera-mirror-rest/utils.js
```diff
@@ -660,6 +660,41 @@ const parsePublicKeyQueryParam = (parsedQueryParams, columnName) => {
...
+  const {bindTimestampRange, maxTimestampRangeNs} = config.query;
+  if (!bindTimestampRange) {
+    return tsRange;
+  }
+
+  let queryTsRange;
+  if (tsRange) {
+    if (tsRange.begin && tsRange.end) {
+      return tsRange; // Already fully bound
+    }
```
xin-hedera, line 680: Isn't the function supposed to limit the range to at most `maxTimestampRangeNs`? Then why would here return the range as is if both `begin` and `end` are not null and not zero?

hedera-mirror-rest/utils.js
```diff
@@ -660,6 +660,41 @@ const parsePublicKeyQueryParam = (parsedQueryParams, columnName) => {
...
+  }
+
+  if (queryTsRange.end) {
+    queryTsRange.begin = queryTsRange.end - maxTimestampRangeNs;
+  } else if (queryTsRange.begin) {
+    queryTsRange.end = queryTsRange.begin + maxTimestampRangeNs;
+  } else {
+    queryTsRange.end = BigInt(Date.now()) * constants.NANOSECONDS_PER_MILLISECOND;
+    queryTsRange.begin = queryTsRange.end - maxTimestampRangeNs;
+  }
```
xin-hedera, line 694: advancing `begin` or shrinking `end` depends on the order. In addition, you shouldn't change it if the user specified range is less than or equal to `maxTimestampRangeNs`.

hedera-mirror-rest/__tests__/utils.test.js
```diff
@@ -1358,6 +1359,58 @@ describe('Utils test - utils.parseTransactionTypeParam', () => {
...
 
+describe('Utils test - utils.bindTimestampRange', () => {
+  let clock;
+  // Set now, so Date.now() returns the expected epoch millis
+  const now = Date.parse('2022-10-15T08:10:15.333Z');
+  const nowNs = BigInt(now) * constants.NANOSECONDS_PER_MILLISECOND;
+
+  beforeAll(() => {
+    clock = sinon.useFakeTimers({now});
+  });
```
xin-hedera, line 1370: why would you need sinon fake timers? `bindTimestampRange` doesn't rely on Date.now(), right?

xin-hedera (changes requested):

hedera-mirror-rest/utils.js
jascks: 👍

hedera-mirror-rest/utils.js
jascks: Back when I first wrote this, I was only concerned about the range being bound. In this case it is bound by the client, though the range could be wide and involve numerous partitions. Please see follow up in below comment.

hedera-mirror-rest/utils.js
jascks: Here the range was not fully bound by the client or else the function would have already returned. So, begin or end or both is not defined. There is no shrinking going on per se. But, in regards to the comment above, where both ends are defined by the client, it sounds like shrinking needs to occur, based on the order and `maxTimestampRangeNs`.

xin-hedera: that's not true because

- we should still bound the range even it's a closed range by the user since it can be really large
- there is bug on line 678 - 680 for a range `[begin, end]` when both are not null and at least one is 0, simply because `(0 && any value)` evaluates to false

hedera-mirror-rest/utils.js
jascks: Limit in place.

hedera-mirror-rest/utils.js
jascks: Okay, I think I have resolved both bullet points now.

hedera-mirror-rest/utils.js
```diff
@@ -660,6 +660,47 @@ const parsePublicKeyQueryParam = (parsedQueryParams, columnName) => {
...
+        if (order === 'desc') {
+          queryTsRange.begin = queryTsRange.end - maxTimestampRangeNs;
+        } else {
+          queryTsRange.end = queryTsRange.begin + maxTimestampRangeNs;
+        }
+      }
+    }
+  }
+
+  return queryTsRange;
```
xin-hedera, line 701: This has become even more complicated due to many nested if.. else.. conditionals.

I still need to review the changes in `transactions.js` to have a full picture. However I believe it can be simplified if we require `tsRange` has both begin and end.

To achieve that, we can default tsRange to `[0, MAX_LONG]` and change the values to reflect what's passed in the request URL.

In the function, the algorithm then becomes:

```
if (end === MAX_LONG) {
  end = NOW;
}

if (end - begin <= maxTimestampRangeNs) {
  return tsRange;
}

if (order == 'desc') {
  begin = end - maxTimestampRangeNs;
} else {
  end = begin + maxTimestampRangeNs;
}

return tsRange;
```

xin-hedera: follow-up, there's obvious one-off problem in the algorithm in my comment in case both ends of the range are inclusive, i.e., for a range in the form `[begin, end]` the number of discrete timestamp values should be `end - begin + 1`

hedera-mirror-rest/transactions.js
jascks: Done

hedera-mirror-rest/utils.js
jascks: Done

hedera-mirror-rest/__tests__/transactions.test.js
```diff
@@ -201,84 +205,134 @@ const validateOrder = function (transactions, order) {
...
+      {field: 'consensus_timestamp', operator: '>=', value: Number(`${timeOneHourAgo}000000000`)},
+      {field: 'limit', operator: '=', value: limit},
+      {field: 'order', operator: '=', value: 'desc'},
+    ],
+    checks2: [
+      {field: 'consensus_timestamp', operator: '>=', value: `${timeNow - limit + 1}000000000`},
+      {field: 'consensus_timestamp', operator: '<=', value: `${timeNow}000000000`},
+      {field: 'payer_account_id', operator: 'in', value: [0, 1, 2]},
+      {
+        field: 'consensus_timestamp',
```
jascks, line 239: Oops. Was trying to cause test to fail which it wouldn't. Ignore for now.

hedera-mirror-rest/__tests__/transactions.test.js
```diff
@@ -201,84 +205,134 @@ const validateOrder = function (transactions, order) {
...
     checkFunctions: [
-      {func: validateTsRange, args: [0, timeNow]},
+      {func: validateTsRange, args: [0, timeNowMs]},
       {func: validateFields, args: []},
     ],
   },
   accountid_lowerlimit: {
     urlparam: 'account.id=gte:0.0.1111',
-    checks: [{field: 'entity_id', operator: '>=', value: 1111}],
+    checks1: [{field: 'entity_id', operator: '>=', value: 1111}],
```
jascks, line 273: While I created checks1 and checks2, these tests have not yet been modified otherwise.

jascks:
This PR has been superseded by #7503. The comments in this PR are still relevant, but work continues in the new PR.

PR #7344, opened 2023-12-07
Add topic message lookup to topic and sequence requests

### LINKED ISSUE #7157: Topic Message By Id and Sequence number Endpoints Show Regression in V2
### Problem

`/topics/{id}/messages/{sequenceNumber}` show large performance decrease in V2 when compared to V1

### Solution

When topic_message table is partitioned by timestamp range, we need to utilize topic_message_lookup table to narrow down the number of partitions to query for better performance.

- Set `hedera.mirror.importer.parser.record.entity.persist.topicMessageLookups=true` to enable topic_message_lookup and run the migration to backfill the table from existing data in topic_message table
- In Rest API 
    - For `/topics/{id}/messages/{sequenceNumber}`, the query to find the timestamp range is similar to `/topics/{id}/messages` , just need to use the range include operation, e.g., sequence_number_range @> $2 in the following query `select timestamp_range
from topic_message_lookup
where topic_id = $1 and sequence_number_range && $2
order by sequence_number_range`
   - Limit the number of timestamp partitions by adding the timestamp ranges to the where conditions

Refer to [6976](https://github.com/hashgraph/hedera-mirror-node/issues/6976)
  

### Alternatives

_No response_

edwin-greene:
**Description**:
When topic message lookup is configured, the topic and sequence request REST API will now use the topic_message_lookup table, this increases the performance of the query with partitioned data.

**Related issue(s)**:

Fixes #7157 

**Notes for reviewer**:
K6 results from testing with Citus: 
```
| topicsIdMessagesSequence | /topics/{id}/messages/{sequenceNumber} | 1500 | 100.00 | 4323.67/s | 4323.67/s | 297.57ms | No | |
```

**Checklist**


hedera-mirror-rest/topicmessage.js
```diff
@@ -101,14 +101,23 @@ const getMessageByTopicAndSequenceRequest = async (req, res) => {
...
-  const pgSqlParams = [topicId.getEncodedId(), seqNum];
+  let pgSqlQuery;
+  let pgSqlParams;
+  if (config.query.topicMessageLookup) {
+    const query = await getQueryWithTopicMessageLookupTimestamps(topicId, seqNum);
+    if (query === null) {
+      return null;
+    }
+    pgSqlQuery = query.pgSqlQuery;
+    pgSqlParams = query.pgSqlParams;
```
xin-hedera, line 112: prefer object destructuring syntax

```suggestion
    {pgSqlQuery,  pgSqlParams} = query;
```

hedera-mirror-rest/topicmessage.js
```diff
@@ -400,6 +409,39 @@ const getMessages = async (pgSqlQuery, pgSqlParams, preQueryHint) => {
   return rows.map((row) => new TopicMessage(row));
 };
 
+const getQueryWithTopicMessageLookupTimestamps = async (topicId, seqNum) => {
+  // Get the timestamp range from the topic message lookup table
+  const sequenceNumber = BigInt(seqNum);
```
xin-hedera, line 414: unnecessary since you have an explicit type conversion in the SQL

```suggestion
  const sequenceNumber = BigInt(seqNum);
```

hedera-mirror-rest/topicmessage.js
```diff
@@ -400,6 +409,39 @@ const getMessages = async (pgSqlQuery, pgSqlParams, preQueryHint) => {
...
+  const topicMessageLookupQuery = `select lower(timestamp_range) as timestamp_start, upper(timestamp_range) as timestamp_end
+                    from ${TopicMessageLookup.tableName}
+                    where topic_id = $1 and sequence_number_range @> $2::int8
+                    order by ${TopicMessageLookup.SEQUENCE_NUMBER_RANGE}`;
+  const pgObject = utils.buildPgSqlObject(topicMessageLookupQuery, [topicId.getEncodedId(), sequenceNumber]);
+  if (pgObject === null) {
+    return null;
+  }
```
xin-hedera, line 422: why do we need these lines?

should just call `pool.queryQuietly` with  the query and params

hedera-mirror-rest/topicmessage.js
```diff
@@ -400,6 +409,39 @@ const getMessages = async (pgSqlQuery, pgSqlParams, preQueryHint) => {
   return rows.map((row) => new TopicMessage(row));
 };
 
+const getQueryWithTopicMessageLookupTimestamps = async (topicId, seqNum) => {
+  // Get the timestamp range from the topic message lookup table
+  const sequenceNumber = BigInt(seqNum);
+  const topicMessageLookupQuery = `select lower(timestamp_range) as timestamp_start, upper(timestamp_range) as timestamp_end
+                    from ${TopicMessageLookup.tableName}
+                    where topic_id = $1 and sequence_number_range @> $2::int8
```
xin-hedera, line 417: nit, for consistency, we use bigint everywhere
```suggestion
                    where topic_id = $1 and sequence_number_range @> $2::bigint
```

hedera-mirror-rest/topicmessage.js
```diff
@@ -400,6 +409,39 @@ const getMessages = async (pgSqlQuery, pgSqlParams, preQueryHint) => {
   return rows.map((row) => new TopicMessage(row));
 };
 
+const getQueryWithTopicMessageLookupTimestamps = async (topicId, seqNum) => {
+  // Get the timestamp range from the topic message lookup table
+  const sequenceNumber = BigInt(seqNum);
+  const topicMessageLookupQuery = `select lower(timestamp_range) as timestamp_start, upper(timestamp_range) as timestamp_end
```
xin-hedera, line 415:
```suggestion
  const topicMessageLookupQuery = `select timestamp_range
```

and the postgres driver will decode it into a `pg-range.Range` object

hedera-mirror-rest/topicmessage.js
```diff
@@ -400,6 +409,39 @@ const getMessages = async (pgSqlQuery, pgSqlParams, preQueryHint) => {
...
+                    from ${TopicMessageLookup.tableName}
+                    where topic_id = $1 and sequence_number_range @> $2::int8
+                    order by ${TopicMessageLookup.SEQUENCE_NUMBER_RANGE}`;
+  const pgObject = utils.buildPgSqlObject(topicMessageLookupQuery, [topicId.getEncodedId(), sequenceNumber]);
+  if (pgObject === null) {
+    return null;
+  }
+
+  const {rows} = await pool.queryQuietly(pgObject.query, pgObject.params);
+  if (_.isNil(rows)) {
```
xin-hedera, line 425: rows won't be nil, if no result found, the `rows.length` will be 0

https://node-postgres.com/apis/result

hedera-mirror-rest/topicmessage.js
```diff
@@ -400,6 +409,39 @@ const getMessages = async (pgSqlQuery, pgSqlParams, preQueryHint) => {
...
+    return null;
+  } else if (rows.length === 0) {
+    throw new NotFoundError();
```
xin-hedera, line 428: missing test case coverage here

hedera-mirror-rest/topicmessage.js
```diff
@@ -400,6 +409,39 @@ const getMessages = async (pgSqlQuery, pgSqlParams, preQueryHint) => {
...
+  // Use the timestamp range as constraints for the topic message query
+  const timestampStart = _.first(rows).timestamp_start;
+  const timestampEnd = _.first(rows).timestamp_end;
+  const pgSqlQuery = `select *
+                      from ${TopicMessage.tableName}
+                      where ${TopicMessage.TOPIC_ID} = $1
+                        and ${TopicMessage.SEQUENCE_NUMBER} = $2
+                        and ((${TopicMessage.CONSENSUS_TIMESTAMP} >= $3
+                            and ${TopicMessage.CONSENSUS_TIMESTAMP} < $4))
+                      limit 1`;
```
xin-hedera, line 440: remove the limit, redundant

hedera-mirror-rest/topicmessage.js
edwin-greene: Thanks I have updated this.

hedera-mirror-rest/topicmessage.js
edwin-greene: Thanks, I have removed the nil check.

hedera-mirror-rest/topicmessage.js
edwin-greene: Got it, I have removed the limit.

hedera-mirror-rest/__tests__/specs/topics/{id}/messages/{id}/idseq-not-found-topic-message-lookup-enabled.json
```diff
@@ -0,0 +1,45 @@
...
+    },
+    "accounts": [
+      {
+        "num": 7,
+        "type": "TOPIC"
+      }
+    ],
+    "balances": [],
+    "transactions": [],
+    "cryptotransfers": [],
```
xin-hedera, line 17: nit, remove these empty domain objects in every new file.

xin-hedera (dismissed):
LGTM

xin-hedera (approved):

PR #7346, opened 2023-12-07
Update contract state endpoint to support slot parameter in any order

### LINKED ISSUE #7240: Contract state API query parameters are order dependent
### Description

The order of the query parameters in the `/api/v1/contracts/{id}/state` matters when we should be able to pass them in arbitrary order with the same results. Also, passing any operator other than `=`, `lte`, or `eq` is ignored and doesn't impact the query results since everything is converted to `<=`. We should check to see if we can support `lt`, `gt`, and `gte` and either add support or throw an error if we can't.

### Steps to reproduce

Doesn't work:
https://mainnet-public.mirrornode.hedera.com/api/v1/contracts/0.0.3964808/state?slot=0x01&timestamp=1700603449.165031003

Works:
https://mainnet-public.mirrornode.hedera.com/api/v1/contracts/0.0.3964808/state?timestamp=1700603449.165031003&slot=0x01

### Additional context

_No response_

### Hedera network

mainnet

### Version

0.92.0

### Operating system

None

edwin-greene:
**Description**:

- The contract state REST API now supports the slot parameter in any order.
- The timestamp parameter now supports the lt/gt/gte operators and explicitly does not support the ne operator.

**Related issue(s)**:

Fixes #7240 

**Notes for reviewer**:

**Checklist**


hedera-mirror-rest/controllers/contractController.js
```diff
@@ -866,31 +866,34 @@ class ContractController extends BaseController {
...
         default:
           break;
       }
     }
 
+    const slotInValues = [];
+    for (const slotFilter of slotFilters) {
+      // If a timestamp filter is present the slot value needs additional conversion
+      // because there is an inconsistency between the column slot in contract_state and contract_state_change.
+      const slot = utils.formatSlot(slotFilter.value, timestampPresent);
```
xin-hedera, line 884: believe this will break existing logic.

Put aside that there is bug in how slot was formatted.

The original logic supports when `timestamp` appears first and slot has odd number of digits, e.g., `0x3`, `0x103`.

The change you made is going to break it.

I think you should fix `formatSlot` so when leftPad is `true` it does what's originally done by
```
  formatSlot(formatSlot(slot), true);
```

xin-hedera (changes requested):

hedera-mirror-rest/controllers/contractController.js
edwin-greene: Thanks. I updated `utils.formatSlot`, it should now work with odd numbered values.

The original call to formatSlot should be functionally equivalent to the current version, so I haven't changed that. Here is the original: 

```
let slot = utils.formatSlot(filter.value);
if (timestamp) {
    slot = utils.formatSlot(filter.value, true);
}
```

xin-hedera: I think you missed my last comment about the `leftPad` flag is inverted. i.e., when `leftPad` is false, the function pads slot to a full 64 character hex string

edwin-greene: Thanks I see now, I have updated it make more sense :)

xin-hedera (approved):
LGTM

steven-sheehy (approved):
LGTM

xin-hedera (approved):
LGTM

Issue #7193, opened 2023-11-10
Add in-equivalence tests 


yiliev0:
### Problem

Currently we don't have tests that verify system accounts called via different opcodes like

- [ ] - direct calls
- [ ] - callcode
- [ ] - call
- [ ] - delegatecall
- [ ] - staticcall
- [x] - selfdestruct
- [x] - balance
- [x] - extcodecopy
- [x] - extcodehash
- [x] - extcodesize
- [ ] - HTS transfers

### Solution

Add tests that would cover all opcodes with system accounts

bilyana-gospodinova:
Closing this task as we decided to close all existing PRs.

PR #7365, opened 2023-12-11
HIP-584: Add utility adapter

IvanKavaldzhiev:
**Description**:

Add a utility class, which can make the same contract call against consensus node or mirror node and extract the result from their corresponding return types. Then, the result is validated following the same rules and expected result.
**Related issue(s)**: #7193 

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/NodeAdapter.java
+++ b/hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/NodeAdapter.java
@@ -1,0 +1,298 @@
+/*
+ * Copyright (C) 2023 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hedera.mirror.test.e2e.acceptance.steps;
+
+import static com.hedera.mirror.test.e2e.acceptance.config.AcceptanceTestProperties.EquivalenceNodeConfiguration.CONSENSUS;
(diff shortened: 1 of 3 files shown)


hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/NodeAdapter.java
```diff
@@ -0,0 +1,298 @@
...
+                    contractToCall, method, expectedResult, address, isForSystemAccount);
+        } else if (MIRROR.equals(acceptanceTestProperties.getEquivalenceNodeConfiguration())) {
+            validateMirrorContractCallNumeric(
+                    contractToCall.contractId().toSolidityAddress(), method, expectedResult, address);
+        } else {
+            validateConsensusContractCallQueryNumeric(
+                    contractToCall, method, expectedResult, address, isForSystemAccount);
+            validateMirrorContractCallNumeric(
+                    contractToCall.contractId().toSolidityAddress(), method, expectedResult, address);
+        }
```
steven-sheehy, line 129: I'm not a fan of this entire `NodeAdapter` class. We should not have multiple places that check config and either call mirror node or consensus node. There should be one place. This class also should not handle result conversion and simply return a normalized response like a `String` result. It also should not do any assertions on the result.

Basically, a NodeAdapter should just be responsible for normalizing the request and response for either type and calling the appropriate backend based upon the config. If you need to do both it can be via cucumber parameterization.

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/NodeAdapter.java
```diff
@@ -0,0 +1,298 @@
...
+import java.math.BigInteger;
+import org.springframework.beans.factory.annotation.Autowired;
+
+/**
+ * A utility class that provides methods to call contract functions against both consensus and mirror nodes. There are utility methods
+ * that extract results of the same type from the different response types from the different nodes. The end result from both nodes is validated against the same rules and expected result.
+ * Thus, we can check if both nodes have equivalent behaviour.
+ * */
+@Named
+public class NodeAdapter extends AbstractFeature {
```
steven-sheehy, line 44: What about instead of having this complicated approach just have a config that makes the existing `MirrorNodeClient.contractsCall()` call either HAPI or MAPI for all requests? Or do we not expect all calls to necessarily work via HAPI?

steven-sheehy (changes requested):

PR #7373, opened 2023-12-12
Update config limit max test

### LINKED ISSUE #7315: Flaky balances.test.js
### Description

The REST API tests are periodically [failing](https://github.com/hashgraph/hedera-mirror-node/actions/runs/7088437172/job/19290935653) in CI due to `balances.test.js`.

### Steps to reproduce

`./gradlew :rest:build`

### Additional context

_No response_

### Hedera network

other

### Version

main

### Operating system

None

edwin-greene:
**Description**:
Update the config limit max test; this could remove the possibility of a problem with the limits of other tests being set too high.

**Related issue(s)**:

Fixes #7315 

**Notes for reviewer**:
Setting the limit max to below 99 yields errors identical to those seen in the build logs for this issue. Perhaps there is a problem with the rest test infrastructure where the custom config from `config.test.js` is sometimes in use for other tests. The custom config reduced the max limit to 30, so that would cause the other tests to fail. This change assures that even if the custom config is used the tests will not fail.

The REST tests have been ran through the CI build process 5 times in this PR and have not been able to recreate the problem, so it could be that this change solves it.

**Checklist**

### DIFF
--- a/hedera-mirror-rest/__tests__/config.test.js
+++ b/hedera-mirror-rest/__tests__/config.test.js
@@ -30,7 +30,7 @@ const custom = {
         response: {
           compression: false,
           limit: {
-            max: 30,
+            max: 101,
           },
         },
         shard: 1,


jascks (approved):
LGTM

steven-sheehy (approved):
LGTM. We should think of ways to change `config.test.js` to not write to the standard locations to impact the traditional application.yml. Maybe some test specific env variable.

PR #7379, opened 2023-12-13
HIP-584: Add utility adapter for sending calls to different nodes for acceptance tests

IvanKavaldzhiev:
**Description**:

Add a utility class, which can make the same contract call against consensus node or mirror node and extract the result from their corresponding return types. Then, the result is validated following the same rules and expected result.

**Related issue(s)**: #7193 

Fixes #

**Notes for reviewer**:

It seems when we want to test a negative scenario with calls to the consensus node, we face a problem, since we are using a spring retry mechanism to send calls to consensus via the Java SDK. If the first request fails, we then attempt to send the same transaction object (having the same TransactionID) and eventually we receive DUPLICATE_TRANSACTION pre-check failures, which is normal, since we send the same transaction to the consensus node. 

We should enhance the retry mechanism, so that each new calls have a different timestamp and stop retrying if we have concrete error like INVALID_SOLIDITY_ADDRESS or CONTRACT_REVERT_EXECUTED. It makes sense to retry if we get status BUSY for example.

This would be handled in a new PR.
**Checklist**


hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/CallFeature.java
```diff
@@ -196,9 +192,13 @@ public void htsIsFrozen(String tokenName) {
...
-        var data = encodeData(PRECOMPILE, HTS_IS_FROZEN_SELECTOR, asAddress(tokenId), asAddress(contractClient));
-        var response = callContract(data, precompileContractAddress);
-
+        var response = callContract(
+                true,
+                StringUtils.EMPTY,
+                PRECOMPILE,
+                HTS_IS_FROZEN_SELECTOR,
+                asAddress(tokenId),
+                asAddress(contractClient));
```
steven-sheehy, line 201: Why do we need to change places to use the network adapter if they are hardcoded to mirror node?

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/CallFeature.java
```diff
@@ -160,9 +154,13 @@ public void ierc721MetadataTokenBalanceOf(String tokenName) {
...
-        var data = encodeData(ERC, IERC721_TOKEN_BALANCE_OF_SELECTOR, asAddress(tokenId), asAddress(contractClient));
-        var response = callContract(data, ercContractAddress);
-
+        var response = callContract(
+                true,
+                StringUtils.EMPTY,
+                ERC,
+                IERC721_TOKEN_BALANCE_OF_SELECTOR,
+                asAddress(tokenId),
+                asAddress(contractClient));
```
steven-sheehy, line 163: This is now a longer, harder to understand method signature. Prefer the separate varags encodeData and the shorter callContract(String, String). If we need more customization prefer a builder approach. Could use `ContractCallRequestBuilder` as the object to pass info to `NetworkAdapter` then provide helper methods to provide partially built objects.

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/CallFeature.java
IvanKavaldzhiev: I removed the hardcoded enum and instead now use cucumber parameters that are passed inside the feature definition. In this way it would be easier to re-execute the same call twice on different nodes, since we have a new call for each parameters row in the Examples section.

steven-sheehy: Yes, but this example is not that equivalence feature where the node enum is in the file. So why change how it calls mirror node to use the adapter?

hedera-mirror-test/src/test/resources/features/contract/equivalence.feature
```diff
@@ -7,33 +7,53 @@ Feature: in-equivalence tests
...
-      | "0.0.361" |
-      | "0.0.750" |
-      | "0.0.999" |
+      | account   | node        |
+      | "0.0.0"   | "consensus" |
+      | "0.0.0"   | "mirror"    |
+      | "0.0.1"   | "consensus" |
+      | "0.0.1"   | "mirror"    |
+      | "0.0.2"   | "consensus" |
+      | "0.0.2"   | "mirror"    |
```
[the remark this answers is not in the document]
IvanKavaldzhiev: I reduced the number of test values

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/response/ContractCallResponse.java
```diff
@@ -18,6 +18,7 @@
 
 import static com.hedera.mirror.test.e2e.acceptance.util.TestUtil.hexToAscii;
 
+import io.micrometer.common.util.StringUtils;
```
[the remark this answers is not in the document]
IvanKavaldzhiev: Fixed

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/client/NetworkAdapter.java
```diff
@@ -0,0 +1,184 @@
...
+            return mirrorClient.contractsCall(contractCallRequestBody);
+        } else {
+            final var gas = contractClient
+                    .getSdkClient()
+                    .getAcceptanceTestProperties()
+                    .getFeatureProperties()
+                    .getMaxContractFunctionGas();
+
+            final var result = contractClient.executeContractQuery(
+                    deployedContract.contractId(), method.getSelector(), gas, convertParametersForSDK(arguments));
```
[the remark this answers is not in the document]
IvanKavaldzhiev: I adapted the logic to directly set encoded bytes

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/CallFeature.java
IvanKavaldzhiev: I removed the new changes from all features except equivalence.feature

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/client/NetworkAdapter.java
```diff
@@ -0,0 +1,165 @@
...
+                    .getMaxContractFunctionGas();
+
+            final var decodedData = Strings.decode(data);
+            final var result = contractClient.executeContractQuery(
+                    deployedContract.contractId(), method.getSelector(), gas, decodedData);
+            return convertResponseFromConsensusNode(contractResource, method, result.asBytes());
+        }
+    }
+
+    public String encodeData(ContractResource resource, SelectorInterface method, Object... args) {
```
steven-sheehy, line 88: I still don't see why we're refactoring all encoding to move it to this `NetworkAdapter`, especially as part of this PR. I actually like it in the `AbstractFeature` and think we should revert and keep this PR very small and focused on just adding the logic to call either MAPI or HAPI and only for equivalence feature.

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/client/NetworkAdapter.java
```diff
@@ -0,0 +1,127 @@
...
+
+    public com.esaulpaugh.headlong.abi.Address asLongZeroHeadlongAddress(final ContractId contractID) {
+        return Address.wrap(Address.toChecksumAddress(BigInteger.valueOf(contractID.num)));
+    }
+
+    public com.esaulpaugh.headlong.abi.Address asLongZeroHeadlongAddress(final AccountId accountId) {
+        return Address.wrap(Address.toChecksumAddress(BigInteger.valueOf(accountId.num)));
+    }
+
+    private ContractCallResponse convertResponseFromConsensusNode(
```
steven-sheehy, line 90: Can you add a code comment why this extra conversion is necessary and we can't just hex encode the result? Does the relay have to do a similar conversion when calling consensus nodes?

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/EquivalenceFeature.java
```diff
@@ -82,59 +86,69 @@ public void selfDestructAndSetBeneficiary(String beneficiary) {
...
         }
     }
 
-    @Then("I execute balance opcode against a contract with balance")
-    public void balanceOfContract() {
-        var parameters = new ContractFunctionParameters().addAddress(equivalenceDestructContractSolidityAddress);
-        var functionResult = executeContractCallQuery(deployedEquivalenceCall, "getBalance", parameters);
-        assertEquals(new BigInteger("10000"), functionResult.getInt256(0));
+    @Then("I execute balance opcode against a contract with balance with call to {string}")
+    public void balanceOfContract(String node) {
```
steven-sheehy, line 103: It would be better to use an enum with automatic conversion using `@ParameterType`. See `AcceptanceTest` and `AccountNameEnum` mapping.

steven-sheehy (approved):
LGTM

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/client/NetworkAdapter.java
```diff
@@ -0,0 +1,127 @@
...
+                    .getMaxContractFunctionGas();
+
+            final var decodedData = Strings.decode(data);
+            final var result = contractClient.executeContractQuery(
+                    deployedContract.contractId(), method.getSelector(), gas, decodedData);
+            return convertResponseFromConsensusNode(contractResource, method, result.asBytes());
+        }
+    }
+
+    public com.esaulpaugh.headlong.abi.Address asLongZeroHeadlongAddress(final ContractId contractID) {
```
yiliev0, line 82: I believe we can use any of the `TestUtil.asAddress` methods, they should be doing the same thing. If not, we can enhance them or add an additional override method to accept the needed parameters.

IvanKavaldzhiev: Resolved

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/EquivalenceFeature.java
IvanKavaldzhiev: Fixed

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/client/NetworkAdapter.java
IvanKavaldzhiev: I added a javadoc. I couldn't find an easier way to get around this headlong decoding. I could research a bit more how SDKs handle the response and if we find an easier way, it could be adapter in a separate PR.

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/EquivalenceFeature.java
```diff
@@ -82,59 +87,67 @@ public void selfDestructAndSetBeneficiary(String beneficiary) {
         }
     }
 
-    @Then("I execute balance opcode to system account {string} address would return 0")
-    public void balanceOfAddress(String address) {
-        var accountId = new AccountId(extractAccountNumber(address)).toSolidityAddress();
-        var parameters = new ContractFunctionParameters().addAddress(accountId);
-        var functionResult = executeContractCallQuery(deployedEquivalenceCall, "getBalance", parameters);
+    @Then("I execute balance opcode to system account {string} address would return 0 with call to {string}")
+    public void balanceOfAddress(String address, String node) {
```
steven-sheehy, line 91: I meant use `@ParameterType` for automatic conversion. But can adapt in a follow up.
```suggestion
    @Then("I execute balance opcode to system account {string} address would return 0 with call to {node}")
    public void balanceOfAddress(String address, NodeNameEnum node) {
```

steven-sheehy (approved):

PR #7382, opened 2023-12-14
Re-implement temp table parser logic to be v2 compatible

### LINKED ISSUE #7007: Importer show ingestion performance regression on some tables with Citus
### Description

Importer paired with Citus database shows ingestion performance regression during a recent test at 10K TPS.

```
2023-10-04T01:25:08.437+0000 INFO scheduling-6 c.h.m.i.p.b.BatchInserter Copied 4237 rows to contract_log table in 269.7 ms
2023-10-04T01:25:08.483+0000 INFO scheduling-6 c.h.m.i.p.b.BatchInserter Copied 769 rows to contract_result table in 45.82 ms
2023-10-04T01:25:08.766+0000 INFO scheduling-6 c.h.m.i.p.b.BatchInserter Copied 80936 rows to crypto_transfer table in 283.0 ms
2023-10-04T01:25:08.801+0000 INFO scheduling-6 c.h.m.i.p.b.BatchInserter Copied 1 rows to custom_fee_temp table in 34.88 ms
2023-10-04T01:25:08.933+0000 INFO scheduling-6 c.h.m.i.p.b.BatchInserter Copied 5563 rows to topic_message table in 130.9 ms
2023-10-04T01:25:09.207+0000 INFO scheduling-6 c.h.m.i.p.b.BatchInserter Copied 19242 rows to transaction table in 274.2 ms
2023-10-04T01:25:09.449+0000 INFO scheduling-6 c.h.m.i.p.b.BatchInserter Copied 3275 rows to entity_temp table in 242.1 ms
2023-10-04T01:25:09.484+0000 INFO scheduling-6 c.h.m.i.p.b.BatchInserter Copied 1 rows to token_temp table in 34.76 ms
2023-10-04T01:25:10.442+0000 INFO scheduling-6 c.h.m.i.p.b.BatchInserter Copied 7364 rows to token_account_temp table in 957.3 ms
2023-10-04T01:25:12.763+0000 INFO scheduling-6 c.h.m.i.p.b.BatchInserter Copied 2233 rows to nft_temp table in 2.320 s
2023-10-04T01:25:12.789+0000 INFO scheduling-6 c.h.m.i.p.b.BatchInserter Copied 93 rows to staking_reward_transfer table in 25.54 ms
2023-10-04T01:25:12.825+0000 INFO scheduling-6 c.h.m.i.p.b.BatchInserter Copied 3815 rows to token_transfer table in 36.54 ms
2023-10-04T01:25:12.825+0000 INFO scheduling-6 c.h.m.i.p.r.e.s.SqlEntityListener Completed batch inserts in 4.658 s
2023-10-04T01:25:12.828+0000 INFO scheduling-6 c.h.m.i.p.r.e.s.SqlEntityListener Processed 1 sidecars
2023-10-04T01:25:12.828+0000 INFO scheduling-6 c.h.m.i.p.r.RecordFileParser Successfully processed 19242 items from 2023-10-03T07_02_12.003404395Z.rcd.gz in 4.872 s
2023-10-04T01:25:13.338+0000 INFO scheduling-6 c.h.m.i.p.b.BatchInserter Copied 3966 rows to contract_log table in 255.0 ms
2023-10-04T01:25:13.380+0000 INFO scheduling-6 c.h.m.i.p.b.BatchInserter Copied 713 rows to contract_result table in 42.09 ms
2023-10-04T01:25:13.641+0000 INFO scheduling-6 c.h.m.i.p.b.BatchInserter Copied 75712 rows to crypto_transfer table in 260.2 ms
2023-10-04T01:25:13.726+0000 INFO scheduling-6 c.h.m.i.p.b.BatchInserter Copied 5186 rows to topic_message table in 85.49 ms
2023-10-04T01:25:14.076+0000 INFO scheduling-6 c.h.m.i.p.b.BatchInserter Copied 17994 rows to transaction table in 349.4 ms
2023-10-04T01:25:14.325+0000 INFO scheduling-6 c.h.m.i.p.b.BatchInserter Copied 3188 rows to entity_temp table in 248.5 ms
2023-10-04T01:25:14.786+0000 INFO scheduling-6 c.h.m.i.p.b.BatchInserter Copied 7007 rows to token_account_temp table in 461.1 ms
2023-10-04T01:25:17.621+0000 INFO scheduling-6 c.h.m.i.p.b.BatchInserter Copied 2161 rows to nft_temp table in 2.835 s
2023-10-04T01:25:17.646+0000 INFO scheduling-6 c.h.m.i.p.b.BatchInserter Copied 88 rows to staking_reward_transfer table in 23.95 ms
2023-10-04T01:25:17.681+0000 INFO scheduling-6 c.h.m.i.p.b.BatchInserter Copied 3422 rows to token_transfer table in 35.05 ms
2023-10-04T01:25:17.681+0000 INFO scheduling-6 c.h.m.i.p.r.e.s.SqlEntityListener Completed batch inserts in 4.599 s
2023-10-04T01:25:17.684+0000 INFO scheduling-6 c.h.m.i.p.r.e.s.SqlEntityListener Processed 1 sidecars
2023-10-04T01:25:17.685+0000 INFO scheduling-6 c.h.m.i.p.r.RecordFileParser Successfully processed 17994 items from 2023-10-03T07_02_14.004366299Z.rcd.gz in 4.790 s
```

Notably, upsert ~2000 rows to `nft` table took 2+ seconds, other tables including `contract_log`, `entity`, and `token_account` also show high insert / upsert duration though not as bad as `nft`.

### Steps to reproduce

1. Create a citus database cluster with citus 2.1 and PostgreSQL 16, 1 coordinator and 3 workers
2. Download the record stream files [here](https://drive.google.com/file/d/1vH-R9XJ21bpMeILkGeYvwQYvm442NLhn/view?usp=drive_link) and extract the tarball to a directory structure per importer LocalStreamFileProvider's requirement
3. Run importer with LocalStreamFileProvider and v2 schema to ingest record stream files from performance network with `startDate: 2023-10-03T07:00:00Z`
4. Monitor the importer log for parse duration at 10K tps 

### Additional context

_No response_

### Hedera network

other

### Version

v0.88.0

### Operating system

None

jnels124:
**Description**:
* convert session temp tables to real tables
* add repeatable migration to regenerate "temp" upsert tables
* change dashboard variable values metric to use parser instead of downloader

**Related issue(s)**:

Fixes #7007 

**Notes for reviewer**:

**Checklist**


edwin-greene (dismissed):
Looks good to me

charts/hedera-mirror-common/dashboards/hedera-mirror-importer.json
```diff
@@ -5665,15 +5665,15 @@
           "type": "prometheus",
           "uid": "${prometheus}"
         },
-        "definition": "label_values(hedera_mirror_download_request_seconds_count{application=\"$application\",cluster=~\"$cluster\",namespace=~\"$namespace\",pod=~\"$pod\"}, type)",
```
jnels124, line 5668: just a small change to make metrics work when using local stream file provider

charts/hedera-mirror/templates/secret-passwords.yaml
```diff
@@ -132,15 +139,20 @@ stringData:
...
-    alter default privileges in schema {{ $dbSchema }} grant select on tables to readonly;
+    grant select on all tables in schema {{ $dbSchema }}, {{ $tempSchema }} to readonly;
+    grant usage on schema {{ $dbSchema }}, {{ $tempSchema }} to readonly;
+    alter default privileges in schema {{ $dbSchema }}, {{ $tempSchema }} grant select on tables to readonly;
 
     -- Grant readwrite privileges
-    grant insert, update, delete on all tables in schema {{ $dbSchema }} to readwrite;
-    alter default privileges in schema {{ $dbSchema }} grant insert, update, delete on tables to readwrite;
+    grant insert, update, delete on all tables in schema {{ $dbSchema }}, {{ $tempSchema }} to readwrite;
+    alter default privileges in schema {{ $dbSchema }}, {{ $tempSchema }} grant insert, update, delete on tables to readwrite;
```
steven-sheehy, line 155: Believe we should adjust the search path somewhere for Citus. Or is that handled somewhere else?

steven-sheehy (changes requested):

charts/hedera-mirror/templates/configmap-init.yaml
```diff
@@ -57,30 +60,39 @@ data:
...
-    alter default privileges in schema :dbSchema grant usage on sequences to readwrite;
+    grant insert, update, delete on all tables in schema :dbSchema, :tempSchema to readwrite;
+    grant usage on all sequences in schema :dbSchema, :tempSchema to readwrite;
+    alter default privileges in schema :dbSchema, :tempSchema grant insert, update, delete on tables to readwrite;
+    alter default privileges in schema :dbSchema, :tempSchema grant usage on sequences to readwrite;
 
     -- Alter search path
     \connect postgres postgres
-    alter database :dbName set search_path = :dbSchema, public;
+    alter database :dbName set search_path = :dbSchema, public, :tempSchema;
```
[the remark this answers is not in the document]
jnels124: I will update the doc

hedera-mirror-importer/src/main/resources/db/migration/v2/R__02_temp_table_distribution.sql
```diff
@@ -0,0 +1,15 @@
+---- ${flyway:timestamp} trigger to always run
+
+select create_distributed_table('${tempSchema}.contract_state_temp', 'contract_id', colocate_with => 'contract_state');
```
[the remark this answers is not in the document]
jnels124: it is not idempotent. However, the repeatable migration that runs before this drops the tables and recreates them so this will work.

hedera-mirror-importer/src/main/resources/db/migration/common/R__01_temp_tables.sql
```diff
@@ -0,0 +1,109 @@
+---- ${flyway:timestamp} trigger to always run
```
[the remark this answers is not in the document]
jnels124: Yes it will run on every startup. That is a good idea. We can have every versioned migration set a placeholder that is its version number. If it changes, then it will trigger these to run

charts/hedera-mirror/templates/secret-passwords.yaml
jnels124: we set it up above on line 92

hedera-mirror-importer/src/main/resources/db/migration/common/R__01_temp_tables.sql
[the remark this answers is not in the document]
steven-sheehy: I meant more do a query for latest versioned migration and compare against classpath migrations

hedera-mirror-importer/src/main/resources/db/migration/common/R__01_temp_tables.sql
```diff
@@ -0,0 +1,109 @@
...
+
+drop table if exists ${tempSchema}.contract_state_temp;
+drop table if exists ${tempSchema}.crypto_allowance_temp;
+drop table if exists ${tempSchema}.custom_fee_temp;
+drop table if exists ${tempSchema}.dissociate_token_transfer;
+drop table if exists ${tempSchema}.entity_temp;
+drop table if exists ${tempSchema}.entity_stake_temp;
```
xin-hedera, line 8: perhaps skip `entity_stake_temp`

- it's completely unused even created here, since there is no change to the SQLs in EntityStakeRepository, a temp table under schema `public` is still created and used
- normally, entity stake calculation is done once per day, and we have plans to reduce the number of rows to insert / update

Suggestion is create a ticket so we don't forget, we can revisit if it becomes a performance issue.

xin-hedera (changes requested):

hedera-mirror-importer/src/main/resources/db/migration/v2/R__02_temp_table_distribution.sql
```diff
@@ -0,0 +1,15 @@
...
+select create_distributed_table('${tempSchema}.crypto_allowance_temp', 'owner', colocate_with => 'crypto_allowance');
```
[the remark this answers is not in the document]
jnels124: we could do that but it doesn't make sense in one case. The dissociate token transfer table is distributed differently than token transfer would be.

charts/hedera-mirror/templates/configmap-init.yaml
```diff
@@ -36,7 +36,9 @@ data:
       --set "rosettaPassword=${HEDERA_MIRROR_ROSETTA_DB_PASSWORD}" \
       --set "rosettaUsername=${HEDERA_MIRROR_ROSETTA_DB_USERNAME}" \
       --set "web3Password=${HEDERA_MIRROR_WEB3_DB_PASSWORD}" \
-      --set "web3Username=${HEDERA_MIRROR_WEB3_DB_USERNAME}" <<__SQL__
+      --set "web3Username=${HEDERA_MIRROR_WEB3_DB_USERNAME}" \
+      --set "tempSchema=${HEDERA_MIRROR_IMPORTER_DB_TEMPSCHEMA}" \
+      --set "tempSchemaAdminRole=${HEDERA_MIRROR_IMPORTER_DB_TEMPSCHEMA}_admin" <<__SQL__
```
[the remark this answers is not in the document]
jnels124: in the bash script files, having the variable set this way makes the sql easier. Can't really do something like `create role :tempSchema_admin` there is not syntax i can find that will allow me to reuse the tempSchema var with additional text appended.

hedera-mirror-importer/src/main/resources/db/migration/common/R__01_temp_tables.sql
jnels124: I have removed the entity_stake_temp table creation for now since you are correct that is not being used as is. Will create a ticket to determine if this logic should be updated to use the temp schema.

hedera-mirror-importer/src/main/resources/db/migration/common/R__01_temp_tables.sql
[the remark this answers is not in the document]
jnels124: created a follow up ticket to handle this.

hedera-mirror-importer/src/main/resources/db/scripts/init-temp-schema.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
+set -e
+
+psql -d "user=postgres connect_timeout=3" \
+  --set ON_ERROR_STOP=1 \
+  --set "dbName=${DB_NAME:-mirror_node}" \
+  --set "dbSchema=${DB_SCHEMA:-public}" \
+  --set "importerUsername=${IMPORTER_USERNAME:-mirror_importer}" \
+  --set "ownerUsername=${OWNER_USERNAME:-mirror_node}" \
+  --set "tempSchema=${HEDERA_MIRROR_IMPORTER_DB_TEMPSCHEMA:-temporary}"  <<__SQL__
```
jnels124, line 10: will correct this name

hedera-mirror-importer/src/main/resources/db/migration/v2/R__02_temp_table_distribution.sql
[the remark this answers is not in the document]
steven-sheehy: Could just do Xin's suggestion and do that one special case as a one off

hedera-mirror-importer/src/main/resources/db/scripts/init-temp-schema.sh
```diff
@@ -0,0 +1,58 @@
...
+\connect :dbName
+create or replace function create_temp_schema(tempSchema varchar, dbSchema varchar, dbName varchar,
```
steven-sheehy, line 12: Why does this need to be a function using execute and formatted sql? It's a shell script so should be able to use `--set` and string templates. Also, the if schema exists check can move outside as a separate psql so you don't need conditional logic in the SQL itself.

steven-sheehy (changes requested):

hedera-mirror-importer/src/main/resources/db/scripts/init-temp-schema.sh
jnels124: I need to be able to do a query to see if the schema exists first. Originally tried to implement in anonymous block but unfortunately you are not able to access any params passed from the command line within a block

hedera-mirror-importer/src/main/resources/application.yml
```diff
@@ -13,6 +13,7 @@ hedera:
         restPassword: mirror_api_pass
         restUsername: mirror_api
         schema: public
+        tempSchema: temporary
```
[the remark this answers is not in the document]
jnels124: will do. Put it here mostly to make overriding from charts more consistent with other DB vars

docs/database/README.md
```diff
@@ -207,6 +207,9 @@ transactions in the balance and record streams. These issues should only appear
 * Solution: Fixed in Hedera Services [v0.34.2](https://github.com/hashgraph/hedera-services/releases/tag/v0.34.2) on
   February 17, 2023. Fixed in Mirror Node in v0.74.3 by a migration that adds the missing transactions.
 
+## Manual Steps Required For 0.96.0 Upgrade
+* For existing operators, upgrading to 0.96.0 requires manual upgrades to be applied to the database. You must execute the script located [here](/hedera-mirror-importer/src/main/resources/db/scripts/init-temp-schema.sh)
```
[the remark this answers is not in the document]
jnels124: I will be more clear

hedera-mirror-importer/src/main/resources/db/migration/v2/R__02_temp_table_distribution.sql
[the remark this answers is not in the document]
jnels124: will add to be a task to the existing follow up ticket

steven-sheehy (dismissed):
LGTM

charts/hedera-mirror/templates/configmap-init.yaml
```diff
@@ -57,30 +59,39 @@ data:
...
 
     -- Grant readwrite privileges
-    grant insert, update, delete on all tables in schema :dbSchema to readwrite;
-    grant usage on all sequences in schema :dbSchema to readwrite;
-    alter default privileges in schema :dbSchema grant insert, update, delete on tables to readwrite;
-    alter default privileges in schema :dbSchema grant usage on sequences to readwrite;
+    grant insert, update, delete on all tables in schema :dbSchema, :tempSchema to readwrite;
+    grant usage on all sequences in schema :dbSchema, :tempSchema to readwrite;
+    alter default privileges in schema :dbSchema, :tempSchema grant insert, update, delete on tables to readwrite;
+    alter default privileges in schema :dbSchema, :tempSchema grant usage on sequences to readwrite;
```
xin-hedera, line 90: I don't think these are needed either. The permission to access tempSchema should come from `authorization temporary_admin`, i.e., the temp schema's owner is `temporary_admin`, and `mirror_importer` is member of the role `temporary_admin`.

For that, instead of the current role hierarchy, I think we should do

- create role temporary_admin
- grant role temporary_admin to read_write

docs/database/README.md
```diff
@@ -207,6 +207,15 @@ transactions in the balance and record streams. These issues should only appear
 * Solution: Fixed in Hedera Services [v0.34.2](https://github.com/hashgraph/hedera-services/releases/tag/v0.34.2) on
   February 17, 2023. Fixed in Mirror Node in v0.74.3 by a migration that adds the missing transactions.
 
+## Breaking Schema Changes Introduced in 0.96.0
+
+In version 0.96.0 a new schema was introduced to handle processing of upsertable entities. This change doesn't require any
+manual steps for new operators that use one of our init scripts or helm charts to configure the database. However, existing operators 
+upgrading to versions greater than 0.96.0 are required to configure this new schema
```
xin-hedera, line 214: change it to `upgrading to 0.96.0 or a later version from an earlier version are required and configure and run this ... once`?

hedera-mirror-importer/src/main/resources/db/scripts/init.sh
```diff
@@ -46,7 +46,8 @@ psql -d "user=postgres connect_timeout=3" \
   --set "rosettaPassword=${ROSETTA_PASSWORD:-mirror_rosetta_pass}" \
   --set "rosettaUsername=${ROSETTA_USERNAME:-mirror_rosetta}" \
   --set "web3Password=${WEB3_PASSWORD:-mirror_web3_pass}" \
-  --set "web3Username=${WEB3_USERNAME:-mirror_web3}" <<__SQL__
+  --set "web3Username=${WEB3_USERNAME:-mirror_web3}" \
+  --set "tempSchema=${HEDERA_MIRROR_IMPORTER_DB_TEMPSCHEMA:-temporary}" <<__SQL__
```
xin-hedera, line 50: I think only the init script in charts needs the `HEDERA_MIRROR_IMPORTER_` prefix, right?
 ```suggestion
   --set "tempSchema=${DB_TEMPSCHEMA:-temporary}" <<__SQL__
```

File: hedera-mirror-importer/src/main/resources/db/scripts/init-temp-schema.sh [hunk 16]
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+set -e
+TEMP_SCHEMA="${HEDERA_MIRROR_IMPORTER_DB_TEMPSCHEMA:-temporary}"
```
xin-hedera, line 3: keep it consistent with init.sh

```suggestion
TEMP_SCHEMA="${DB_TEMPSCHEMA:-temporary}"
```

hedera-mirror-importer/src/main/resources/db/scripts/init-temp-schema.sh
```diff
@@ -0,0 +1,51 @@
...
+SCHEMA_EXISTS="$(psql -d "user=postgres connect_timeout=3" \
+                  -XAt \
+                  -c "select exists (select schema_name from information_schema.schemata where schema_name = '$TEMP_SCHEMA')")"
```
xin-hedera, line 6: missing dbname=
```suggestion
DB_NAME="${DB_NAME:-mirror_node}"
SCHEMA_EXISTS="$(psql -d "user=postgres connect_timeout=3 dbname=${DB_NAME}" \
                  -XAt \
                  -c "select exists (select schema_name from information_schema.schemata where schema_name = '$TEMP_SCHEMA')")"
```

xin-hedera (changes requested):

hedera-mirror-importer/src/main/resources/db/scripts/init-temp-schema.sh
jnels124: it connects to db below but will move here instead since we dont need to connect to multiple dbs

xin-hedera: The problem is the query actually runs in postgres database and thus the information schema info looked up in that db will not contain the temporary schema regardless if that name exists in mirror_node database, right?

hedera-mirror-importer/src/main/resources/db/scripts/init.sh
jnels124: I will make consistent. You are right that only the chart values need to do this.

charts/hedera-mirror/templates/configmap-init.yaml
jnels124: we can do that assuming we want every user now and down the line with readwrite (now it is only mirror_node/mirror_importer) to have the ownership abilities over the temp schema. Not a huge concern there but may not want that for all users with this role down the line

charts/hedera-mirror/templates/configmap-init.yaml
```diff
@@ -57,30 +59,39 @@ data:
...
-    grant select on all tables in schema :dbSchema to readonly;
-    grant select on all sequences in schema :dbSchema to readonly;
-    grant usage on schema :dbSchema to readonly;
-    alter default privileges in schema :dbSchema grant select on tables to readonly;
-    alter default privileges in schema :dbSchema grant select on sequences to readonly;
+    grant select on all tables in schema :dbSchema, :tempSchema to readonly;
+    grant select on all sequences in schema :dbSchema, :tempSchema to readonly;
+    grant usage on schema :dbSchema, :tempSchema to readonly;
+    alter default privileges in schema :dbSchema, :tempSchema grant select on tables to readonly;
+    alter default privileges in schema :dbSchema, :tempSchema grant select on sequences to readonly;
```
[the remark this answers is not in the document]
jnels124: I put them in there just in case it is helpful when running some analyze type querries on upsert statements. Having this allows to safely connect to the pods in various environments with read only to perform such querries

charts/hedera-mirror/templates/configmap-init.yaml
xin-hedera: That makes sense. Perhaps then we don’t need extra permission  grants of the temp schema to readwrite at all, because mirror_importer is a member of temporary_admin role?

hedera-mirror-importer/src/main/resources/db/scripts/init-temp-schema.sh
jnels124: Yes you are correct it causes an issue in the case of trying to restart. Restarting from a main version to this works no problem. Restarting from this version to this version doesn't. The schema exist lookup always returns false because we never actually create the schema in that database as you mentioned.

charts/hedera-mirror/templates/configmap-init.yaml
jnels124: Yes if we keep the temporary_admin role we can still get rid of the extra grants on the readwrite user for the new schema. However, vote to still keep the readonly grants for this schema.

xin-hedera (approved):
LGTM

steven-sheehy (approved):

PR #7415, opened 2023-12-18
HIP-584 Historical: fix slot parameter in `MirrorEntityAccess`

0xivanov:
**Description**:

After conducting both manual and integration testing, it was determined that the `contract_state_change table` is capable of storing values for the slot column ranging from 0 bytes to 32 bytes. This is in contrast to the `contract_state` table, where the slot is consistently represented as a 32-byte array.

`contract_state_change`:  

image

`contract_state`:  

image

Solution:
Trim the leading zeros from the key if we are doing historical call.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
@@ -119,7 +119,8 @@ public class MirrorEntityAccess implements HederaEvmEntityAccess {
         }
 
         return store.getHistoricalTimestamp()
-                .map(t -> contractStateRepository.findStorageByBlockTimestamp(entityId, key.toArrayUnsafe(), t))
+                .map(t -> contractStateRepository.findStorageByBlockTimestamp(
+                        entityId, trimLeadingZeros(key.toArrayUnsafe()), t))
                 .orElseGet(() -> contractStateRepository.findStorage(entityId, key.toArrayUnsafe()))
                 .map(Bytes::wrap)
                 .orElse(Bytes.EMPTY);
@@ -146,5 +147,28 @@ public class MirrorEntityAccess implements HederaEvmEntityAccess {
             entityId = store.getToken(address, OnMissing.DONT_THROW).getEntityId();
         }
         return entityId;
+    }
+
+    static byte[] trimLeadingZeros(byte[] inputArray) {
+        int leadingZeros = 0;
+        int length = inputArray.length;
+
+        // Count the number of leading zeros
+        for (int i = 0; i < length; i++) {
+            if (inputArray[i] == 0) {
+                leadingZeros++;
+            } else {
+                break;
+            }
+        }
+
+        // Create a new array with the trimmed length
+        int trimmedLength = length - leadingZeros;
+        byte[] trimmedArray = new byte[trimmedLength];
+
+        // Copy the non-zero elements to the new array
+        System.arraycopy(inputArray, leadingZeros, trimmedArray, 0, trimmedLength);
+
+        return trimmedArray;
     }
 }
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
@@ -120,7 +120,7 @@ public class MirrorEntityAccess implements HederaEvmEntityAccess {
 
         return store.getHistoricalTimestamp()
                 .map(t -> contractStateRepository.findStorageByBlockTimestamp(
-                        entityId, trimLeadingZeros(key.toArrayUnsafe()), t))
+                        entityId, key.trimLeadingZeros().toArrayUnsafe(), t))
                 .orElseGet(() -> contractStateRepository.findStorage(entityId, key.toArrayUnsafe()))
                 .map(Bytes::wrap)
                 .orElse(Bytes.EMPTY);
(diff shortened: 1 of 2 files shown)


hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccessTest.java
```diff
@@ -269,4 +281,27 @@ void fetchCodeIfPresentReturnsNull() {
         final var result = mirrorEntityAccess.fetchCodeIfPresent(ADDRESS);
         assertThat(result).isNull();
     }
+
+    @Test
+    void testTrimLeadingZeros() {
```
kselveliev, line 286: Can make this test parametrized

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
```diff
@@ -119,7 +119,8 @@ public Bytes getStorage(final Address address, final Bytes key) {
         }
 
         return store.getHistoricalTimestamp()
-                .map(t -> contractStateRepository.findStorageByBlockTimestamp(entityId, key.toArrayUnsafe(), t))
+                .map(t -> contractStateRepository.findStorageByBlockTimestamp(
+                        entityId, trimLeadingZeros(key.toArrayUnsafe()), t))
```
steven-sheehy, line 123:
```suggestion
                        entityId, key.trimLeadingZeros().toArrayUnsafe(), t))
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
```diff
@@ -119,7 +119,8 @@ public Bytes getStorage(final Address address, final Bytes key) {
...
                 .orElseGet(() -> contractStateRepository.findStorage(entityId, key.toArrayUnsafe()))
                 .map(Bytes::wrap)
```
steven-sheehy, line 125: The slot values are also not zero prefixed. Does the EVM library already pad them appropriately? Is there a test for this?

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccessTest.java
steven-sheehy: This test should go away since we can use the tuweni library for it. Instead, we should add a test that uses `getStorage()`.

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
0xivanov: Nice catch, did not notice there is such method

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
0xivanov: From what i observed in `contract_state` slots are zero prefixed. We've implemented an acceptance test, which involves modifying the contract state and subsequently retrieving the updated state. Additionally, there is an integration test named `stateChangeWorksWithDynamicEthCall` that performs a similar process. They both read from `contract_state` table and the passed key matches the bytes in the db.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccessTest.java
0xivanov: Removed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
steven-sheehy: Yes, but this `map` runs for both contract_state and contract_state_change. In the latter slot values are not prefixed, hence my question.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccessTest.java
```diff
@@ -225,8 +225,18 @@ void getAliasForAccountWithEmptyOne() {
...
+        final var result = UInt256.fromBytes(mirrorEntityAccess.getStorage(ADDRESS, BYTES));
+        assertThat(result).isEqualTo(UInt256.fromHexString(HEX));
+    }
+
+    @Test
+    void getStorageHistorical() {
         when(store.getHistoricalTimestamp()).thenReturn(timestamp);
-        when(contractStateRepository.findStorageByBlockTimestamp(ENTITY_ID, BYTES.toArrayUnsafe(), timestamp.get()))
+        when(contractStateRepository.findStorageByBlockTimestamp(
+                        ENTITY_ID, BYTES.trimLeadingZeros().toArrayUnsafe(), timestamp.get()))
```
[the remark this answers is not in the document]
0xivanov: Fixed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/MirrorEntityAccess.java
0xivanov: It seems I may have misinterpreted your message. Both `contract_state_change` and `contract_state` are non-zero prefixed for `value_read`/`value_written` and `value`, respectively. In addressing the question, I believe there's no need to pad historical values. Could you please provide clarification, as I might still be misunderstanding something?

steven-sheehy: I think you are right, I just wanted to be completely sure. It seems like the EVM must be left padding it somewhere.

steven-sheehy (approved):
LGTM

PR #7418, opened 2023-12-19
HIP-584: Add tests for historical data

### LINKED ISSUE #7048: Add tests for historical data support
### Problem

As the development of historical data support is underway, it is essential to incorporate relevant tests to ensure its functionality. The tests to be added should focus on:

- Contract storage 
- Account balance
- Attempting to hit non existing account in a block
- Executing "Finalised, pending, safe" statuses for blocks

### Solution

Setup a new test suite that would cover historical data support.

### Alternatives

_No response_

### LINKED ISSUE #7055: HIP-584: Historical Support- Provide acceptance tests proving that historical block support for system contract related calls are working properly.
### Problem

After extending all the db queries for system contracts and being able to load evm configuration based on block number
we need to create **Acceptance** tests proving that historical block support is working.

### Solution

Create **Acceptance** tests to check all scenarios.
Different block number passed so that we are able to test with all of the possible evm versions.
Hit all possible precompiles to check if all of the extended db queries are working properly.

### Alternatives

_No response_

yiliev0:
**Description**: 

This PR covers acceptance tests for historical data support with various precompile functions.

**Related issue(s)**: 

Fixes #7048 
Fixes #7055

**Notes for reviewer**:  acceptance tests will be fully operational once the PR #7505 (bug fix) is merged into the codebase.

**Checklist**


hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/HistoricalFeature.java
```diff
@@ -0,0 +1,965 @@
...
+    }
+
+    @Then("the mirror node REST API should return status {int} for the contracts creation")
+    public void verifyMirrorAPIResponse(int status) {
+        if (networkTransactionResponse != null) {
+            verifyMirrorTransactionsResponse(mirrorClient, status);
+        }
+    }
+
+    @RetryAsserts
```
0xivanov, line 140: Do we need the retry?

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/HistoricalFeature.java
```diff
@@ -0,0 +1,965 @@
...
+        var historicalResponse = callContract(initialBlockNumber, data, estimatePrecompileContractSolidityAddress);
+        assertEquals(response, historicalResponse);
+
+        // recreate the deleted token
+        var tokenResponse = tokenClient.getToken(tokenName);
+        networkTransactionResponse = tokenResponse.response();
+        verifyMirrorTransactionsResponse(mirrorClient, 200);
+    }
+
+    @RetryAsserts
```
0xivanov, line 756: Do we need the retry?

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/HistoricalFeature.java
```diff
@@ -0,0 +1,965 @@
...
+
+        var data = encodeData(PRECOMPILE, GET_TOKEN_INFORMATION, asAddress(tokenResponse.tokenId()));
+        assertEthCallReturnsBadRequest(previousBlock, data, estimatePrecompileContractSolidityAddress);
+    }
+
+    private String getLastBlockNumber() {
+        return mirrorClient.getBlocks().getBlocks().getFirst().getNumber().toString();
+    }
+
+    private void waitForNextBlock() throws InterruptedException {
```
0xivanov, line 907: add some inline comments explaining the logic here

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/HistoricalFeature.java
yiliev0: Removed.

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/HistoricalFeature.java
yiliev0: Removed.

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/HistoricalFeature.java
yiliev0: Done.

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/client/TokenClient.java
```diff
@@ -595,6 +595,8 @@ public enum TokenNameEnum {
                 TokenType.FUNGIBLE_COMMON,
                 TokenKycStatus.KycNotApplicable,
                 TokenFreezeStatus.FreezeNotApplicable),
+        FUNGIBLEHISTORICAL(
```
IvanKavaldzhiev, line 598: FUNGIBLEHISTORICAL -> FUNGIBLE_HISTORICAL

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/client/TokenClient.java
```diff
@@ -631,6 +633,11 @@ public enum TokenNameEnum {
                 TokenType.NON_FUNGIBLE_UNIQUE,
                 TokenKycStatus.KycNotApplicable,
                 TokenFreezeStatus.FreezeNotApplicable),
+        NFTHISTORICAL(
```
IvanKavaldzhiev, line 636: NFTHISTORICAL -> NFT_HISTORICAL

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/HistoricalFeature.java
```diff
@@ -0,0 +1,967 @@
...
+
+    @Given("I successfully create erc contract")
+    public void createNewErcContract() throws IOException {
+        deployedErcContract = getContract(ERC);
+        ercContractSolidityAddress = deployedErcContract.contractId().toSolidityAddress();
+    }
+
+    @Given("I successfully create precompile contract")
+    public void createNewPrecompileContract() throws IOException {
+        deployedPrecompileContract = getContract(PRECOMPILE);
```
IvanKavaldzhiev, line 125: This step might be a bit misleading, since we create 3 accounts apart from creating the precompile contract. Probably the creation of these accounts could be extracted in a new step?

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/client/TokenClient.java
0xivanov: Cucumber detects the enums in .feature files only if the enum is one word, we already have token `ALLOWANCEFUNGIBLE` which is written this way so i thinks it's ok. The other option is to copy a lot of the steps where the change is only the token used, but this would cause a lot of unnecessary code, especially in the historical.feature

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/HistoricalFeature.java
```diff
@@ -0,0 +1,967 @@
...
+        var initialFreezeStatus = response.getResultAsBoolean();
+        waitForNextBlock();
+        networkTransactionResponse = tokenClient.freeze(tokenId, receiverAccountId.getAccountId());
+        verifyMirrorTransactionsResponse(mirrorClient, 200);
+        var historicalFreezeStatus = callContract(initialBlockNumber, data, precompileContractSolidityAddress)
+                .getResultAsBoolean();
+        assertEquals(initialFreezeStatus, historicalFreezeStatus);
+
+        // reverting to old state
+        networkTransactionResponse = tokenClient.unfreeze(tokenId, receiverAccountId.getAccountId());
```
IvanKavaldzhiev, line 569: Could there be a moment where the `freezing` of the token could cause other acceptance tests ran in parallel to fail, if they are executed for the `same` token before the `unfreeze` operation?

yiliev0: The tokens in this test suite are used only here and won't mess with other tests running at the same time.

IvanKavaldzhiev (approved):
LGTM

0xivanov (dismissed):
LGTM

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/HistoricalFeature.java
yiliev0: Done.

steven-sheehy (changes requested):

hedera-mirror-test/src/test/resources/features/contract/historical.feature
```diff
@@ -0,0 +1,83 @@
+@contractbase @fullsuite @historical @call @web3 @acceptance
+Feature: Historical Feature
+
+  Scenario Outline: Validate Historical Data
+    Given I successfully create estimateGas contract
+    Given I successfully create erc contract
+    Given I successfully create precompile contract
+    Given I successfully create estimate precompile contract
```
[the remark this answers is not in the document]
yiliev0: Hey @steven-sheehy ,
Thank you for your suggestion. If I understand correctly, you're proposing that we repurpose the methods in call.feature for both current (latest=true) and historical (latest=false) scenarios, essentially by toggling a latest flag.

While this approach sounds efficient, particularly for scenarios that don't involve state changes (like token and balance updates or deletions), it might not be entirely feasible for our current test scenarios. A significant portion of our tests actively involves state modifications, which is crucial for thoroughly testing of historical support. These tests are designed to "mess up" the state, so to speak, to ensure that the calls return proper historical data. For example, tests that involve actions like wiping or burning tokens have been vital for uncovering issues where the token supply wasn't accurately reflected in historical calls post-modification.

Therefore, while we could integrate the existing historical support implementations into call.feature, it might not substantially reduce costs due to the state-changing nature of most of our tests. This isn't to say that we can't look into optimising our test suite and cost of execution - I would appreciate further suggestions.
However, we need to consider how these changes might affect the depth and quality of our testing.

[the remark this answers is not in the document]
IvanKavaldzhiev: I agree with @yiliev0. In these test, we need to deterministically change state and wait for a block, before we validate that historical `eth_call` is working properly. If we re-execute `call.feature` with a historical block after the executions with latest block have passed, there is a chance that nothing has changed for the given token between the two runs and the returned result would be the same for both `latest` and `historical` block. In this way there is a chance that we miss to catch a bug. It would be more obvious to change token's state and validate that we can still fetch the state value before the update with historical `eth_call`.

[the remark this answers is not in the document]
IvanKavaldzhiev: What if we integrate at least part of these historical `eth_call` scenarios inside TokenFeature and validate that prior to the operations included in TokenFeature, we read the old state of the token being modified? This could potentially save us a lot of the gas costs in these historical scenarios.

We can leave the most specific cases in the `HistoricalFeature`. Additionally, I think we can test multiple historical `eth_call` operations via one HAPI operation to consensus. For example we can delete a token and check:

- getTokenDefaultFreezeStatus
- getTokenDefaultKYCStatus
- getTokenType
- getTokenExpiryInfo

Otherwise, all 4 operations would recreate and delete a given token while we can do these checks at once. 

There might be other operations that might be grouped together and checked with one HAPI call.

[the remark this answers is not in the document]
yiliev0: Hey @IvanKavaldzhiev,

Thank you for your suggestions. I agree that consolidating operations like `getTokenDefaultFreezeStatus`, `getTokenDefaultKYCStatus`, `getTokenType`, and `getTokenExpiryInfo` into a single test would be a strategic move. By initiating the delete/create token process just once, we could indeed significantly reduce the associated costs. This approach makes it more efficient without compromising the quality of our test results.

Regarding the integration of historical `eth_call` scenarios with the `TokenFeature` tests, I appreciate the thought behind this idea. However, I believe that keeping these as separate entities might be more beneficial in the long run. The primary reason is that they are designed to verify different aspects of our system. Historical feature tests and token feature tests each have distinct objectives and scopes. By merging them, we risk creating tests that could fail for multiple unrelated reasons, potentially complicating the process of pinpointing the exact cause of any issues. Clear separation ensures that each test remains focused and that any failures are easier to diagnose and address.

Perhaps, we can look more into optimising the current behaviour, same as the suggestion you've proposed related to the `getToken` operations?

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/HistoricalFeature.java
```diff
@@ -0,0 +1,967 @@
...
+        int currentBlockNumber = Integer.parseInt(getLastBlockNumber());
+        int waitTime = 250;
+        int maxCycles = 8; // 8 cycles * 250 ms = 2 seconds per block
+        int totalTimeWaited = 0; // keep track of the total time waited
+
+        // The cycle below serves to check for a new block via the block API.
+        // If a new block is found, we break the loop early; otherwise, we complete the cycle,
+        // which amounts to a 2-second wait, aligning with the expected time for a new block.
+        for (int i = 0; i < maxCycles; i++) {
+            Thread.sleep(waitTime);
```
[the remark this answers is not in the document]
yiliev0: Replaced this method with awaitility. Thanks for suggestion!

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/HistoricalFeature.java
```diff
@@ -0,0 +1,967 @@
...
+    private ContractCallResponse deleteTokenAndGetHistoricalResponse(
+            TokenId tokenId, String blockNumber, String data, String solidityAddress) {
+        networkTransactionResponse = tokenClient.delete(admin, tokenId);
+        verifyMirrorTransactionsResponse(mirrorClient, 200);
+        return callContract(blockNumber, data, solidityAddress);
+    }
+
+    @Getter
+    @RequiredArgsConstructor
+    enum ContractMethods implements SelectorInterface {
```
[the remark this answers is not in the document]
yiliev0: Removed all duplicated methods.

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/HistoricalFeature.java
```diff
@@ -0,0 +1,970 @@
...
+        var initialResponse =
+                callContract(data, estimateContractSolidityAddress).getResultAsNumber();
+        var initialBlockNumber = getLastBlockNumber();
+        waitForNextBlock();
+        networkTransactionResponse =
+                accountClient.sendCryptoTransfer(receiverAccountId.getAccountId(), Hbar.fromTinybars(50000000), null);
+        verifyMirrorTransactionsResponse(mirrorClient, 200);
+        var response = callContract(initialBlockNumber, data, estimateContractSolidityAddress)
+                .getResultAsNumber();
+        assertEquals(initialResponse, response);
```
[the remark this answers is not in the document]
yiliev0: Replaced all occurrences of `assertEquals` with `assertThat`

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/HistoricalFeature.java
```diff
@@ -0,0 +1,970 @@
...
+
+    @Then("I successfully update the contract storage and get the initial value via historical data")
+    public void getHistoricalContractStorage() throws InterruptedException {
+        var data = encodeData(ESTIMATE_GAS, GET_COUNTER);
+        var initialResponse = callContract(data, estimateContractSolidityAddress);
+        // the block number where contract storage variable is still with initial value
+        var initialBlockNumber = getLastBlockNumber();
+
+        // executing the contract storage update after 2 blocks
+        waitForNextBlock();
```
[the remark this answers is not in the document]
yiliev0: I think I might have mislead you with this comment. We need to wait for next block, so we can make all the state modifications in the next block, including the executeContract in this test case. After that all HAPI methods are using `verifyMirrorTransactionsResponse`

hedera-mirror-test/src/test/resources/features/contract/historical.feature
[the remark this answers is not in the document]
steven-sheehy: > If we re-execute call.feature with a historical block after the executions with latest block have passed, there is a chance that nothing has changed for the given token between the two runs

My suggestion was to run call feature with latest=true, store the results, then run with latest=false which will change state again, then verify call against stored results. I guess this would only work if you ensured the state change operation differed each time (like transfer a tinybar for get balance).

It's not the hbar costs that concern me since `@acceptance` doesn't run on mainnet. It's more the execution time and maintenance burden costs. Do we know how much time and costs this change adds to the overall tests? Perhaps it's not that bad in practice and we can keep the tests as is.

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/HistoricalFeature.java
```diff
@@ -0,0 +1,935 @@
...
+        var result = contractClient.executeContract(
+                deployedEstimatePrecompileContract.contractId(), 3000000, "updateTokenInfoExternal", updateData, null);
+        networkTransactionResponse = result.networkTransactionResponse();
+        verifyMirrorTransactionsResponse(mirrorClient, 200);
+
+        // Collect and compare historical responses
+        for (BigInteger keyValue : tokenKeyValues) {
+            var data = encodeData(ESTIMATE_PRECOMPILE, GET_TOKEN_KEY, asAddress(tokenId.toSolidityAddress()), keyValue);
+            var historicalResponse = callContract(initialBlockNumber, data, estimatePrecompileContractSolidityAddress);
+            assertThat(currentResponses.get(keyValue)).isEqualTo(historicalResponse);
```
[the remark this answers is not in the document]
yiliev0: Fixed, thanks for heads up.

hedera-mirror-test/src/test/resources/features/contract/historical.feature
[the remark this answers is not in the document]
yiliev0: Hey @steven-sheehy, 
After the latest commit where I grouped some tests, it's the following:

- Acceptance tests without Historical feature against the integration env -> 8 min & 337 Hbars
- Acceptance tests with Historical feature against the integration env ->  12 min & 443 Hbars

steven-sheehy (approved):
LGTM. Still not a fan of everything but let's get it in and iterate. In particular, the waiting for next block causes a lot of unnecessary delay. Even within the historical feature we can structure the tests such that all of the state change operations occur then we wait one time for the next block before comparing initial vs latest response. Would require storing the initial response in a map.

IvanKavaldzhiev (approved):
LGTM! Agree with Steven that we can further optimise the tests structure and performance in follow up PRs.

yiliev0:
Resolved conflicts. Seems reasonable to optimise it in future PRs. We can communicate internally and find a feasible solution.

steven-sheehy (approved):

PR #7419, opened 2023-12-19
HIP-584 Historical: Fix historical crypto balance

0xivanov:
**Description**:
This PR includes the following fixes:

- get historical entity balance from `AccountBalanceRepository`
- fix `findHistoricalAccountBalanceUpToTimestamp` in edge case where `balance_snapshot` query does not match anything in the db. Currently reproducible in acceptance tests:
1. create account with initial balance 50
2. make crypto transfer +50 
3. make historical call with timestamp before the crypto transfer, expected balance is 50, but returns 0

I have observed that the state in db after these operations is the following:
`account_balance` - no entry for this accountId
`crypto_transfers` - 2 entries for this accountId, one for initial balance (50) and one for the crypto transfer (50)

`balance_snapshot` matches nothing because there is no entry for this accountId in `account_balance` before the given timestamp.
This causes the statement `ct.consensus_timestamp > s.consensus_timestamp` to be false and cannot match any crypto transfers, always returning 0 due to the `coalesce` at the end of the query.
NB. `account_balance` entry  gets persisted ~8 minutes after the account creation

Current solution:
If `balance_snapshot` does not match anything, find crypto transfers that occured after the account creation and before the given timestamp:
```sql
ct.consensus_timestamp >= coalesce((select consensus_timestamp from balance_snapshot), accountCreatedTimestamp)
```
Changed `>` to `>=` in order to match the initial balance transfer (in this case 50) if there is one.

**Related issue(s)**:

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/AccountDatabaseAccessor.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/AccountDatabaseAccessor.java
@@ -121,7 +121,8 @@ public class AccountDatabaseAccessor extends DatabaseAccessor<Object, Account> {
 
     private Long getAccountBalance(Entity entity, final Optional<Long> timestamp) {
         return timestamp
-                .map(t -> accountBalanceRepository.findHistoricalAccountBalanceUpToTimestamp(entity.getId(), t))
+                .map(t -> accountBalanceRepository.findHistoricalAccountBalanceUpToTimestamp(
+                        entity.getId(), t, entity.getCreatedTimestamp()))
                 .orElseGet(() -> Optional.ofNullable(entity.getBalance()))
                 .orElse(0L);
     }
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/AccountDatabaseAccessor.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/AccountDatabaseAccessor.java
@@ -119,10 +119,20 @@ public class AccountDatabaseAccessor extends DatabaseAccessor<Object, Account> {
                 .orElseGet(() -> nftRepository.countByAccountIdNotDeleted(accountId));
     }
 
+    /**
+     * Determines account balance based on block context.
+     *
+     * Non-historical Call:
+     * Get the balance from entity.getBalance()
+     * Historical Call:
+     * If the entity creation is after the passed timestamp - return 0L (the entity was not created)
+     * Else get the balance from the historical query `findHistoricalAccountBalanceUpToTimestamp`
+     */
     private Long getAccountBalance(Entity entity, final Optional<Long> timestamp) {
         return timestamp
-                .map(t -> accountBalanceRepository.findHistoricalAccountBalanceUpToTimestamp(
-                        entity.getId(), t, entity.getCreatedTimestamp()))
+                .map(t -> t >= entity.getCreatedTimestamp()
+                        ? accountBalanceRepository.findHistoricalAccountBalanceUpToTimestamp(entity.getId(), t)
+                        : Optional.of(0L))
                 .orElseGet(() -> Optional.ofNullable(entity.getBalance()))
                 .orElse(0L);
     }
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/AccountDatabaseAccessor.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/AccountDatabaseAccessor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2023 Hedera Hashgraph, LLC
+ * Copyright (C) 2023-2024 Hedera Hashgraph, LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -63,6 +63,8 @@ public class AccountDatabaseAccessor extends DatabaseAccessor<Object, Account> {
     private static final BinaryOperator<Long> NO_DUPLICATE_MERGE_FUNCTION = (v1, v2) -> {
         throw new IllegalStateException(String.format("Duplicate key for values %s and %s", v1, v2));
     };
+    private static final Optional<Long> ZERO_BALANCE = Optional.of(0L);
+
     private final EntityDatabaseAccessor entityDatabaseAccessor;
     private final NftAllowanceRepository nftAllowanceRepository;
     private final NftRepository nftRepository;
@@ -130,9 +132,14 @@ public class AccountDatabaseAccessor extends DatabaseAccessor<Object, Account> {
      */
     private Long getAccountBalance(Entity entity, final Optional<Long> timestamp) {
         return timestamp
-                .map(t -> t >= entity.getCreatedTimestamp()
-                        ? accountBalanceRepository.findHistoricalAccountBalanceUpToTimestamp(entity.getId(), t)
-                        : Optional.of(0L))
+                .map(t -> {
+                    Long createdTimestamp = entity.getCreatedTimestamp();
+                    if (createdTimestamp == null || t >= createdTimestamp) {
+                        return accountBalanceRepository.findHistoricalAccountBalanceUpToTimestamp(entity.getId(), t);
+                    } else {
+                        return ZERO_BALANCE;
+                    }
+                })
                 .orElseGet(() -> Optional.ofNullable(entity.getBalance()))
                 .orElse(0L);
     }
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/AccountDatabaseAccessor.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/AccountDatabaseAccessor.java
@@ -29,6 +29,7 @@ import com.hedera.mirror.common.domain.entity.Entity;
 import com.hedera.mirror.common.domain.entity.EntityId;
 import com.hedera.mirror.web3.evm.exception.WrongTypeException;
 import com.hedera.mirror.web3.evm.store.DatabaseBackedStateFrame.DatabaseAccessIncorrectKeyTypeException;
+import com.hedera.mirror.web3.repository.AccountBalanceRepository;
 import com.hedera.mirror.web3.repository.CryptoAllowanceRepository;
 import com.hedera.mirror.web3.repository.NftAllowanceRepository;
 import com.hedera.mirror.web3.repository.NftRepository;
@@ -68,6 +69,7 @@ public class AccountDatabaseAccessor extends DatabaseAccessor<Object, Account> {
     private final TokenAllowanceRepository tokenAllowanceRepository;
     private final CryptoAllowanceRepository cryptoAllowanceRepository;
     private final TokenAccountRepository tokenAccountRepository;
+    private final AccountBalanceRepository accountBalanceRepository;
 
     @Override
     public @NonNull Optional<Account> get(@NonNull Object key, final Optional<Long> timestamp) {
@@ -91,7 +93,7 @@ public class AccountDatabaseAccessor extends DatabaseAccessor<Object, Account> {
                 entity.getId(),
                 new Id(entity.getShard(), entity.getRealm(), entity.getNum()),
                 TimeUnit.SECONDS.convert(entity.getEffectiveExpiration(), TimeUnit.NANOSECONDS),
-                Optional.ofNullable(entity.getBalance()).orElse(0L),
+                getAccountBalance(entity, timestamp),
                 Optional.ofNullable(entity.getDeleted()).orElse(false),
                 getOwnedNfts(entity.getId(), timestamp),
                 Optional.ofNullable(entity.getAutoRenewPeriod()).orElse(DEFAULT_AUTO_RENEW_PERIOD),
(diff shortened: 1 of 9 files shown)


hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/repository/AccountBalanceRepositoryTest.java
```diff
@@ -101,21 +102,65 @@ void shouldNotIncludeBalanceAfterTimestampFilter() {
...
+                .customize(b -> b.amount(TRANSFER_AMOUNT)
+                        .entityId(accountId)
+                        .consensusTimestamp(secondTransfer.getConsensusTimestamp() + 2L))
+                .persist();
+        timestampBetweenTheTransfers = secondTransfer.getConsensusTimestamp() + 1L;
+
+        assertThat(accountBalanceRepository.findHistoricalAccountBalanceUpToTimestamp(
+                        accountId, timestampBetweenTheTransfers, accountCreationTimestamp))
+                .get()
+                .isEqualTo(TRANSFER_AMOUNT * 2);
```
IvanKavaldzhiev, line 151: If the balance we are searching for is in-between transfers, shouldn't the final amount be `15L + TRANSFER_AMOUNT`?

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/repository/AccountBalanceRepositoryTest.java
```diff
@@ -101,21 +102,65 @@ void shouldNotIncludeBalanceAfterTimestampFilter() {
...
+                .customize(b -> b.amount(TRANSFER_AMOUNT)
+                        .entityId(accountId)
+                        .consensusTimestamp(firstTransfer.getConsensusTimestamp() + 2L))
+                .persist();
+        long timestampBetweenTheTransfers = firstTransfer.getConsensusTimestamp() + 1L;
+
+        assertThat(accountBalanceRepository.findHistoricalAccountBalanceUpToTimestamp(
+                        accountId, timestampBetweenTheTransfers, accountCreationTimestamp))
+                .get()
+                .isEqualTo(TRANSFER_AMOUNT);
```
IvanKavaldzhiev, line 138: Shouldn't it be 15L + TRANSFER_AMOUNT?

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/repository/AccountBalanceRepositoryTest.java
IvanKavaldzhiev: Actually you are reusing  `timestampBetweenTheTransfers` which is a bit confusing. You can declare a new field - `timestampAfterSecondTransfer`. Then shouldn't the final balance be `15L + TRANSFER_AMOUNT*2`?

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/repository/AccountBalanceRepositoryTest.java
0xivanov: The setting of balance to the entity is misleading. When we create account with balance 15 ,for example, entry in `account_balance` gets persisted for this initial amount. In real environment when we create account the `entity.balance` should match the first `account_balance` for this account id.

0xivanov: notice that `firstTransfer` and `account` have the same timestamp.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/repository/AccountBalanceRepositoryTest.java
0xivanov: please check out my comments https://github.com/hashgraph/hedera-mirror-node/pull/7419#discussion_r1431451860
also, changed the tests to be more comprehendable

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/AccountBalanceRepository.java
```diff
@@ -64,22 +65,24 @@ public interface AccountBalanceRepository extends CrudRepository<AccountBalance,
...
+                    ),
+                    change as (
                         select sum(amount) as amount
-                        from crypto_transfer as ct, balance_snapshot as s
-                        where
+                        from crypto_transfer as ct
+                            where
                             ct.entity_id = ?1 and
-                            ct.consensus_timestamp > s.consensus_timestamp and
+                            ct.consensus_timestamp >= coalesce((select consensus_timestamp from balance_snapshot), ?3) and
```
xin-hedera, line 79: because the account_balance table is deduped and partitioned monthly, a more performant query is:

1. find the latest balance snapshot timestamp of treasury account `0.0.2` before `blockTimestamp`, it works because the design is we never dedup treasury account's balance info, and there will be a row for the account in every snapshot. Let's call it `balanceSnapshotTimestamp`

3. find `accountId`'s latest balance in the range `(balanceSnapshotTimestamp - 31 days, balanceSnapshotTimestamp]`

the equivalent SQL is

```sql
with balance_timestamp as (
  select consensus_timestamp
  from account_balance
  where account_id = 2 and
    consensus_timestamp > ?1 - 2678400000000000 and
    consensus_timestamp <= ?1
  limit 1
), balance_snapshot as (
  select balance
  from account_balance as ab, balance_timestamp as bt
  where account_id = ?1 and
    ab.consensus_timestamp > bt.consensus_timestamp - 2678400000000000 and
    ab.consensus_timestamp < bt.consensus_timestamp
  order by ab.consensus_timestamp desc
  limit 1
), change as (
  select sum(amount) as amount
  from crypto_transfer as ct, balance_timestamp as bt
  where ct.entity_id = ?1 and
      ct.consensus_timestamp > bt.consensus_timestamp and
      ct.consensus_timestamp <= ?2 and
      (ct.errata is null or ct.errata <> 'DELETE')
)
select coalesce((select balance from balance_snapshot), 0) + coalesce((select amount from change), 0)
```

In addition, `accouontCreatedTimestamp` is not needed.

xin-hedera (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/AccountBalanceRepository.java
0xivanov: Can you clarify this scenario?
We have `account_balance` entry with timestamp `1703060978991589711`, id `1136` and balance `50,000,000` (let's call it "myBalance")
We have `account_balance` entry with id `2` with the same timestamp  -`1703060978991589711`
We pass timestamp `1703060978991589711` as parameter.
The first part of the query would match and return timestamp `1703060978991589711` due to `<=`.
The second part won't match "myBalance" at timestamp `1703060978991589711` because we have `ab.consensus_timestamp < bt.consensus_timestamp` and we would go further back in time to search for another entry for id `1136`.
Should't we match "myBalance" in this scenario?
using `ab.consensus_timestamp <= bt.consensus_timestamp`

0xivanov: So how i think this `account_balance` system works: 
every X minutes, Y number of `account_balance` entries are dumped for the accounts involved in `crypto_transfers`. This process includes entries for those accounts, and it always includes an entry for `0.0.2`. This cycle then repeats, with the timestamp of `0.0.2` marking the beginning of a new cycle and the end of the previous one.
@xin-hedera please correct me if I'm wrong.
If what i'm saying is true then before the first cycle of `account_balance` entries this table is empty. Can we use `coalesce` in the `crypto_transfer` part of the query for this case?

0xivanov: I was contemplating today and devised the following SQL query (?1 representing accountId, and ?2 representing blockTimestamp):

```sql
with balance_timestamp as (
  select consensus_timestamp
  from account_balance
  where account_id = 2 and
    consensus_timestamp > ?2 - 2678400000000000 and
    consensus_timestamp <= ?2
  limit 1
), balance_snapshot as (
  select balance
  from account_balance as ab, balance_timestamp as bt
  where account_id = ?1 and
    ab.consensus_timestamp > bt.consensus_timestamp - 2678400000000000 and
    ab.consensus_timestamp <= bt.consensus_timestamp
  order by ab.consensus_timestamp desc
  limit 1
)
, change as (
  select sum(amount) as amount
  from crypto_transfer as ct
  where ct.entity_id = ?1 and
      ct.consensus_timestamp > coalesce((select consensus_timestamp from balance_snapshot), 0) and
      ct.consensus_timestamp <= ?2 and
      (ct.errata is null or ct.errata <> 'DELETE')
)
select coalesce((select balance from balance_snapshot), 0) + coalesce((select amount from change), 0)
```

xin-hedera: yes you are correct, sorry about the mistakes in the query in my feedback

- ?1 in the first CTE should be ?2
- in `balance_snapshot` CTE, you are right, it should be ab.consensus_timestamp <= bt.consensus_timestamp`

0xivanov: Modified the query

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/AccountDatabaseAccessor.java
```diff
@@ -117,6 +119,14 @@ private long getOwnedNfts(Long accountId, final Optional<Long> timestamp) {
                 .orElseGet(() -> nftRepository.countByAccountIdNotDeleted(accountId));
     }
 
+    private Long getAccountBalance(Entity entity, final Optional<Long> timestamp) {
+        return timestamp
+                .map(t -> accountBalanceRepository.findHistoricalAccountBalanceUpToTimestamp(
```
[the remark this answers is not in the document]
0xivanov: Added this check

kselveliev (approved):
LGTM

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/AccountBalanceRepository.java
```diff
@@ -61,22 +61,29 @@ public interface AccountBalanceRepository extends CrudRepository<AccountBalance,
...
                     """
-                    with balance_snapshot as (
-                        select balance, consensus_timestamp
+                    with balance_timestamp as (
+                        select consensus_timestamp
                         from account_balance
-                        where
-                            account_id = ?1 and
+                        where account_id = 2 and
+                            consensus_timestamp > ?2 - 2678400000000000 and
```
0xivanov, line 68: Enhance the javadoc

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/AccountDatabaseAccessor.java
```diff
@@ -117,6 +119,24 @@ private long getOwnedNfts(Long accountId, final Optional<Long> timestamp) {
...
+     *
+     * Non-historical Call:
+     * Get the balance from entity.getBalance()
+     * Historical Call:
+     * If the entity creation is after the passed timestamp - return 0L (the entity was not created)
+     * Else get the balance from the historical query `findHistoricalAccountBalanceUpToTimestamp`
+     */
+    private Long getAccountBalance(Entity entity, final Optional<Long> timestamp) {
+        return timestamp
+                .map(t -> t >= entity.getCreatedTimestamp()
```
0xivanov, line 133: Add test for this case

xin-hedera (dismissed):
LGTM

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/AccountDatabaseAccessor.java
steven-sheehy, line 133: Handle the scenario where createdTimestamp can be null. Add a test.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/AccountDatabaseAccessor.java
```diff
@@ -117,6 +119,24 @@ private long getOwnedNfts(Long accountId, final Optional<Long> timestamp) {
...
+                        ? accountBalanceRepository.findHistoricalAccountBalanceUpToTimestamp(entity.getId(), t)
+                        : Optional.of(0L))
```
steven-sheehy, line 135: Store as a static constant

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/AccountDatabaseAccessor.java
0xivanov: added zero balance constant

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/AccountDatabaseAccessor.java
0xivanov: I presume we retrieve the balance from the `accountBalanceRepository `  when the `createdTimestamp` is null. I have implemented the modification as described earlier. Added unit test.

steven-sheehy (approved):
LGTM

PR #7444, opened 2023-12-22
HIP-584: Properly extract error messages when making ContractCallQuery

IvanKavaldzhiev:
**Description**:

It seems when we want to test a negative scenario with contract call queries to the consensus node, we face a problem, since we are using a spring retry mechanism to send calls to consensus via the Java SDK. If the first request fails, we then attempt to send the same transaction object (having the same TransactionID) and eventually we receive DUPLICATE_TRANSACTION pre-check failures, which is normal, since we send the same transaction to the consensus node.

This PR changes the behaviour by sending a single query only and extract the error reason like INVALID_SOLIDITY_ADDRESS or CONTRACT_REVERT_EXECUTED.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


steven-sheehy (approved):
LGTM. I will let you merge

Issue #7054, opened 2023-10-13
HIP-584: Historical Support- Provide integration tests proving that historical block support for system contract related calls are working properly.


kselveliev:
### Problem

After extending all the db queries for system contracts and being able to load evm configuration based on block number
we need to create integration tests proving that historical block support is working.

### Solution

Create integration tests to check all scenarios.
Different block number passed so that we are able to test with all of the possible evm versions.
Hit all possible precompiles to check if all of the extended db queries are working properly.

PR #7470, opened 2024-01-03
HIP-584 Historical: Add integration tests for nested calls - part 3

bilyana-gospodinova:
**Description**:
In this PR three integration tests are added in order to help with the historical support testing. Each of them contains a function call which fails and a hardcoded value is returned after that. With this we aim to test if the expected hardcoded value is returned and that the failing function call does not make the whole transaction to fail. The tests are three in order to test the three main flows:
- get token info test - in this case we don't have a custom precompile in the code takes the shortcut for view functions
- HTS get approved test - we have a custom precompile in this case and the code does not take the shortcut for view functions
- mint nft token - a modification precompile. In this PR there is a fix in case an exception is thrown so that the frame would revert and the whole transaction would not fail.

Fixes https://github.com/hashgraph/hedera-mirror-node/issues/7054


hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallNestedCallsTest.java
```diff
@@ -80,6 +85,59 @@ void nestedReadOnlyTokenFunctionsTestEthEstimateGas(NestedEthCallContractFunctio
...
+
+        final var successfulResponse = functionEncodeDecoder.encodedResultFor(
+                "nestedHtsGetApprovedAndHardcodedResult", NESTED_CALLS_ABI_PATH, "hardcodedResult");
+
+        assertThat(contractCallService.processCall(serviceParameters)).isEqualTo(successfulResponse);
+    }
+
+    @ParameterizedTest
+    @MethodSource("blockArgumentsProvider")
+    void failedCallMintTokenWithHardcodedResult(final Address inputAddress, final BlockType blockNumber) {
```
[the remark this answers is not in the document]
bilyana-gospodinova: Comment addressed.

steven-sheehy (approved):

steven-sheehy (approved):
LGTM

kselveliev (approved):
LGTM

hedera-mirror-web3/src/main/java/com/hedera/services/store/contracts/precompile/HTSPrecompiledContract.java
```diff
@@ -148,9 +148,18 @@ public Pair<Long, Bytes> computeCosted(
             return handleReadsFromDynamicContext(input, frame);
         }
 
-        final var result = computePrecompile(input, frame);
+        Bytes result = null;
+        try {
+            result = computePrecompile(input, frame).getOutput();
+        } catch (final InvalidTransactionException e) {
```
IvanKavaldzhiev, line 154: It seems this exception is caught inside `computePrecompile` in the services implementation. Currently in our web3 sibling we catch only `NoSuchElementException`, but I believe we should catch `InvalidTransactionException` as well inside the method.

bilyana-gospodinova: I checked the code in services but there are cases (such as mint NFT) where the `InvalidTransactionException` is thrown from:
```
precompileContext.setGasRequirement(precompileContext
                ...
                .getGasRequirement(
                        ...);
```
and more specifically from `OpUsageCtxHelper#metaForTokenMint`.

so it would not be enough to add the exception in the catch block for `NoSuchElementException`. I added the catch clause outside `computePrecompile` just to be sure that we catch the `InvalidTransactionException` in all possible scenarios since we don't have test scenarios for all possible nested calls cases and there might be some implementation that is hidden like the one above and that might throw an exception from `prepareComputation` as well.

IvanKavaldzhiev (approved):
LGTM

PR #7476, opened 2024-01-04
HIP-584 Historical: Add integration tests for custom fees and HTS calls - part 4

bilyana-gospodinova:
**Description**:
In this PR test cases for the previously failing custom fees tests and tests for HTS_GET_APPROVED, HTS_ALLOWANCE, HTS_IS_APPROVED_FOR_ALL are added. The latter are the only scenarios where an exception is not thrown when the data is not found for the passed historical block. The returned result is zeroes. The reason for this inconsistency is because the top frame in the stack overrides the output with zeroes and this behaviour comes from the Besu EVM. These cases should be improved in the future to behave like all the other scenarios where an exception is thrown.

Fixes https://github.com/hashgraph/hedera-mirror-node/issues/7054

### DIFF
--- a/hedera-mirror-web3/src/main/java/com/hedera/services/store/contracts/precompile/impl/AllowancePrecompile.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/services/store/contracts/precompile/impl/AllowancePrecompile.java
@@ -70,7 +70,7 @@ public class AllowancePrecompile extends AbstractReadOnlyPrecompile {
         final var updater = (HederaEvmStackedWorldStateUpdater) frame.getWorldUpdater();
         final var inputData = frame.getInputData();
         final var wrapper = decodeTokenAllowance(inputData);
-        final var store = ((HederaEvmStackedWorldStateUpdater) frame.getWorldUpdater()).getStore();
+        final var store = updater.getStore();
         final var account = store.getAccount(addressFromBytes(wrapper.owner()), OnMissing.THROW);
         validateFalseOrRevert(account.isEmptyAccount(), INVALID_ACCOUNT_ID);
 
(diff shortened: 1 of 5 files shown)


hedera-mirror-web3/src/main/java/com/hedera/services/store/contracts/precompile/impl/AllowancePrecompile.java
```diff
@@ -67,6 +70,10 @@ public RunResult run(MessageFrame frame, TransactionBody transactionBody) {
         final var updater = (HederaEvmStackedWorldStateUpdater) frame.getWorldUpdater();
         final var inputData = frame.getInputData();
         final var wrapper = decodeTokenAllowance(inputData);
+        final var store = ((HederaEvmStackedWorldStateUpdater) frame.getWorldUpdater()).getStore();
```
kselveliev, line 73: I believe here we can just use final var store = updater.getStore();

hedera-mirror-web3/src/main/java/com/hedera/services/store/contracts/precompile/impl/IsApprovedForAllPrecompile.java
```diff
@@ -67,6 +70,10 @@ public RunResult run(MessageFrame frame, TransactionBody transactionBody) {
         final var updater = (HederaEvmStackedWorldStateUpdater) frame.getWorldUpdater();
         final var inputData = frame.getInputData();
         final var wrapper = decodeIsApprovedForAll(inputData);
+        final var store = ((HederaEvmStackedWorldStateUpdater) frame.getWorldUpdater()).getStore();
```
kselveliev, line 73: same as above comment

hedera-mirror-web3/src/main/java/com/hedera/services/store/contracts/precompile/impl/AllowancePrecompile.java
bilyana-gospodinova: Comment addressed.

steven-sheehy (approved):
LGTM

kselveliev (approved):
LGTM

PR #7477, opened 2024-01-04
HIP-584: Reduce code duplication in ContractCallTestSetup

IvanKavaldzhiev:
**Description**:

This PR extracts duplicated code in ContractCallTestSetup in helper methods. Additionally, all test constants are moved to a separate class.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServicePrecompileTest.java
+++ b/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServicePrecompileTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2023 Hedera Hashgraph, LLC
+ * Copyright (C) 2023-2024 Hedera Hashgraph, LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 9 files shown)


hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallTestSetup.java
```diff
@@ -58,374 +106,23 @@
...
-                                            .build()))))
-            .build();
-
-    protected static Key keyWithContractId = Key.newBuilder()
-            .setContractID(contractIdFromEvmAddress(PRECOMPILE_TEST_CONTRACT_ADDRESS.toArrayUnsafe()))
-            .build();
-    protected static Key keyWithDelegatableContractId = Key.newBuilder()
-            .setDelegatableContractId(contractIdFromEvmAddress(PRECOMPILE_TEST_CONTRACT_ADDRESS.toArrayUnsafe()))
-            .build();
+    //    protected static final long expiry = 1_234_567_890L;
```
0xivanov, line 125: Can this be removed?

IvanKavaldzhiev: Removed

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallTestConstants.java
```diff
@@ -0,0 +1,624 @@
...
+import java.util.Arrays;
+import java.util.List;
+import java.util.function.ToLongFunction;
+import org.apache.tuweni.bytes.Bytes;
+import org.bouncycastle.util.encoders.Hex;
+import org.hyperledger.besu.datatypes.Address;
+
+public class ContractCallTestConstants {
+
+    public static final long expiry = 1_234_567_890L;
```
kselveliev, line 62: should be uppercase since it is a constant

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallTestConstants.java
```diff
@@ -0,0 +1,624 @@
...
+import com.hederahashgraph.api.proto.java.TransactionFeeSchedule;
+import java.math.BigInteger;
+import java.util.Arrays;
+import java.util.List;
+import java.util.function.ToLongFunction;
+import org.apache.tuweni.bytes.Bytes;
+import org.bouncycastle.util.encoders.Hex;
+import org.hyperledger.besu.datatypes.Address;
+
+public class ContractCallTestConstants {
```
kselveliev, line 60: I don't think we need public access of the class and since I believe pretty much all of the public constants and methods in this class are accessed in this package only, we can set scope to package level instead of public

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallTestSetup.java
```diff
@@ -1220,114 +686,78 @@ private void ercContractTokenPersist(
...
         final var autoRenewEntityId = fromEvmAddress(AUTO_RENEW_ACCOUNT_ADDRESS.toArrayUnsafe());
 
-        domainBuilder
-                .entity()
-                .customize(e -> e.id(autoRenewEntityId.getId())
-                        .num(autoRenewEntityId.getNum())
-                        .evmAddress(null)
-                        .alias(toEvmAddress(autoRenewEntityId)))
-                .persist();
+        persistEntity(autoRenewEntityId, null, toEvmAddress(autoRenewEntityId), ACCOUNT, 0L, new byte[0]);
```
bilyana-gospodinova, line 744: Balance == 0 is not the same as in the previous implementation. Before the changes there was a balance by default. This probably does not affect the tests at the moment. Just something I noticed.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallTestSetup.java
```diff
@@ -1692,50 +992,29 @@ private void precompileContractPersist() {
...
-                .contractState()
-                .customize(c -> c.contractId(contractEntityId.getId())
-                        .slot(Bytes.fromHexString("0x0000000000000000000000000000000000000000000000000000000000000000")
-                                .toArrayUnsafe())
-                        .value(Bytes.fromHexString("0x4746573740000000000000000000000000000000000000000000000000000000")
-                                .toArrayUnsafe()))
-                .persist();
-
-        domainBuilder.recordFile().customize(f -> f.bytes(contractBytes)).persist();
+        persistEntity(contractEntityId, contractEvmAddress, new byte[0], CONTRACT, 1500L, new byte[0]);
```
bilyana-gospodinova, line 995: Do we need to create new byte[0] objects everywhere? Can't we pass null and check in `persistEntity` if the value is != null instead of checking for array length?

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServicePrecompileTest.java
```diff
@@ -16,6 +16,51 @@
 
 package com.hedera.mirror.web3.service;
 
+import static com.hedera.mirror.web3.service.ContractCallTestConstants.AUTO_RENEW_ACCOUNT_ADDRESS;
+import static com.hedera.mirror.web3.service.ContractCallTestConstants.ECDSA_KEY;
+import static com.hedera.mirror.web3.service.ContractCallTestConstants.ED25519_KEY;
+import static com.hedera.mirror.web3.service.ContractCallTestConstants.FIXED_FEE_WRAPPER;
+import static com.hedera.mirror.web3.service.ContractCallTestConstants.FRACTIONAL_FEE_WRAPPER;
+import static com.hedera.mirror.web3.service.ContractCallTestConstants.FROZEN_FUNGIBLE_TOKEN_ADDRESS;
```
steven-sheehy, line 24: I don't see how moving constants from the parent class `ContractCallTestSetup` to `ContractCallTestConstants` helps improve code readability or reduce duplication. It actually causes the opposite since now there are a lot more lines for import statements and they're duplicate import statements in different sub-classes. It also increases the scope of those constants to global when we should be trying to reduce variable scope. I think we should revert this change and go back to not having a constants class.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallTestSetup.java
```diff
@@ -1220,114 +686,78 @@ private void ercContractTokenPersist(
...
-                        .deleted(false))
-                .persist();
+
+        persistEntity(
+                spenderEntityId,
+                NOT_ASSOCIATED_SPENDER_ALIAS.toArray(),
+                NOT_ASSOCIATED_SPENDER_PUBLIC_KEY.toByteArray(),
+                ACCOUNT,
+                0L,
+                new byte[0]);
```
steven-sheehy, line 696: This is about the same amount of lines but way less readable. I find that methods with more than 3 parameters unreadable and a builder based approach to always be preferable. I'm okay with a little code duplication in tests if it makes the tests easier to understand and maintain. With this change I don't know what these parameters mean anymore when looking at things in a PR and now we have to provide parameters that aren't even applicable to this test case like `new byte[0]`. Same comment for other persist method.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallTestSetup.java
```diff
@@ -1343,40 +773,18 @@ private EntityId fungibleTokenPersist(
...
-                        .decimals(12)
-                        .wipeKey(key)
-                        .freezeKey(key)
-                        .pauseStatus(pauseStatus)
-                        .pauseKey(key)
-                        .supplyKey(key)
-                        .symbol("HBAR"))
-                .persist();
+        persistEntityToken(tokenEntityId, autoRenewEntityId, tokenEvmAddress, 1500L, tokenExpiration, key);
+        persistToken(
```
steven-sheehy, line 777: If we find that we're creating similar tokens in multiple test cases, the way to reduce duplication is to not create one or more long persist methods with many parameters, but to just return the default token builder. Something like:

```java
protected TokenBuilder getFungibleToken(EntityId tokenEntityId) {
    return domainBuilder.token()
         .customize(t -> t.tokenId(tokenEntityId.getId())
                        .treasuryAccountId(EntityId.of(0, 0, treasuryId.getId()))
                        ...
}

@Test
void myTest() {
    var token = getFungibleToken(tokenEntityId)
                    .freezeKey(key)
                    ...
                    .persist();
```

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallTestSetup.java
IvanKavaldzhiev: My initial idea was to reduce the size of the ContractCallTestSetup class itself, however you are right that some code readability was sacrificed. We can try to delete the overall number of tokens and accounts and naturally reduce the class size.  Soon it would get very hard to maintain.

I will put this PR in Draft for the moment until we implement a better approach.

PR #7481, opened 2024-01-05
Optimize list topic messages of specific topic id for v2

### LINKED ISSUE #7320: Topic Message Endpoint - Use timestamp param for V2
### Problem

Currently for V2 we are ignoring the timestamp parameter for the `/topics/{id}/messages` endpoint.
We need to handle the cases for timestamp parameter along with the existing sequence parameter.

### Solution

- TBD (Maybe generate a Dynamic sql based on the parameters)

### Alternatives

_No response_

xin-hedera:
**Description**:

This PR implements performance improvement for listing topic messages of a specific topic id using data in `topic_message_lookup` table

- rework the logic to query topic_message_lookup for timestamp ranges
- add support of consensus timestamp filtering
- fix the topic_message_lookup query not utilizing the btree_gist index on(topic_id, sequence_number_range)
- enable testing matrix for topic message integration specs depending on topic message lookup table
- remove duplicate topic message integration specs
- fix bug in integration test module which may skip test specs when multiple have matrix enabled
- add integration domain ops support of auto generating topic message lookup entries

**Related issue(s)**:

Fixes #7320 

**Notes for reviewer**:

k6 performance tests show much better performance in v2. The results are mixed when comparing against v1.

URL | Pass Rate | Avg. Req Duration | RPS  | v1 RPS | Diff   |
-----|-----------|-------------------|------|--------|------|
/topics/1693742/messages?limit=100 | 100% | 848ms | 1494/s | 1043/s | 143% |
/topics/1693742/messages?sequencenumber=5098305126 | 100% | 358ms | 3439/s | 3561/s | 96.6% |

The v1 RPS data is from release v0.96.0-rc1 report.

**Checklist**


hedera-mirror-rest/__tests__/integration/template.js
```diff
@@ -66,28 +65,32 @@ const walk = (dir, files = []) => {
...
-      .filter((f) => f.endsWith('.json'))
-      .map(async (f) => {
-        const specText = fs.readFileSync(f, 'utf8');
-        const spec = JSONParse(specText);
-        spec.name = path.basename(f);
-        const key = path.dirname(f).replace(specPath, '');
-        const specs = specMap[key] || [];
-        if (spec.matrix) {
-          const apply = (await import(path.join(modulePath, spec.matrix))).default;
-          specs.push(...apply(spec));
```
xin-hedera, line 82: because the `import(...)` function is asynchronous, where there are multiple specs with matrix, line 79 will create multiple empty array as `specs`, and then at line 86, the specs produced by the async execution which completes at last gets to override `specMap[key]`

hedera-mirror-rest/__tests__/integration/template.js
```diff
@@ -193,12 +196,12 @@ describe(`API specification tests - ${groupSpecPath}`, () => {
   };
 
   const specSetupSteps = async (spec) => {
+    overrideConfig(spec.config);
```
xin-hedera, line 199: override config at the beginning, so `query.topicMessageLookup` is consistent during setup and test case execution.

hedera-mirror-rest/topicmessage.js
```diff
@@ -169,266 +165,216 @@ const getTopicMessages = async (req, res) => {
...
+    // If the range size doesn't exceed the limit, use it as is. Note explicitly casting topic id to bigint is
+    // required to utilize the btree gist index on (topic_id, sequence_number_range)
+    query = `select timestamp_range
+      from topic_message_lookup
+      where topic_id = $1::bigint and sequence_number_range && $2::int8range and timestamp_range && $3::int8range
+      order by sequence_number_range`;
+  } else {
+    // Depending on the order, first query the table to find the bound of the sequence number range and build the actual
+    // range from the bound and the limit, then query the table with the actual range
+    // params.push(isOrderAsc ? sequenceNumberRange.upper : sequenceNumberRange.lower);
```
xin-hedera, line 285: commented code, will remove

hedera-mirror-rest/jest.config.js
```diff
@@ -31,6 +31,7 @@ const config = {
   reporters: ['jest-standard-reporter', 'jest-junit'],
   setupFilesAfterEnv: ['./__tests__/jestSetup.js'],
   testEnvironment: 'node',
+  testPathIgnorePatterns: ['/build/', '/node_modules/'],
```
xin-hedera, line 34: ignore files in `builder` folder, sometimes spotless may leave files in `build` folder with the same directory structure of the sources, so for example, if `utils.test.js` is left there, it's in path `build/spotlessJavascript/__tests__/utils.test.js`. We have to tell jest to exclude `build` otherwise `npm test` will fail.

hedera-mirror-rest/topicmessage.js
```diff
@@ -169,266 +165,215 @@ const getTopicMessages = async (req, res) => {
...
-                    where topic_id = $1 and sequence_number_range @> $2::bigint
-                    order by ${TopicMessageLookup.SEQUENCE_NUMBER_RANGE}`;
-  const {rows} = await pool.queryQuietly(topicMessageLookupQuery, [topicId.getEncodedId(), seqNum]);
+const getTopicMessageTimestampRange = async (topicId, seqNum) => {
+  // Get the timestamp range from the topic message lookup table. Note explicitly casting topic id to bigint is required
+  // to utilize the btree gist index on (topic_id, sequence_number_range)
+  const query = `select timestamp_range
+                    from topic_message_lookup
+                    where topic_id = $1::bigint and sequence_number_range @> $2::bigint
+                    order by sequence_number_range`;
```
jascks, line 361: Should there be a `limit 1` here since only `rows[0]` is utilized below?  Perhaps the `where` clause and the gist index results in at most one row already?

xin-hedera: not necessary as you already explained. If the query returns more than 1 row, we've got incorrect data in the table. Because that can only be true when for the sample topic id, there are multiple sequence number ranges containing the discrete sequence number.

jascks: Okay. Thank you for the explanation. 👍

steven-sheehy (approved):
LGTM

jascks (approved):
LGTM

mgoelswirlds (dismissed):
LGTM

jascks (approved):
LGTM again

jnels124 (approved):
LGTM

steven-sheehy (approved):

PR #7483, opened 2024-01-05
fix: Use balance timestamp in /tokens/{id}/balances

### LINKED ISSUE #7442: Use balance timestamp in `/tokens/{id}/balances`
### Problem

Currently the token balances endpoint returns the max record file timestamp. It would be more accurate and more performant to use the max balance timestamp from the list.

### Solution

* Update the `/api/v1/tokens/{id}/balances` to not join on `record_file` and set the timestamp to the max `token_account.balance_timestamp` of the entries in the response.
* Update the tables in `/docs/rest/README.md`

### Alternatives

_No response_

stefan-stefanooov:
**Description**:

- Update the `/api/v1/tokens/{id}/balances` to not join on `record_file` and set the `timestamp` to the max `token_account.balance_timestamp` of the entries in the response.
- Update the tables in `README.md`

**Related issue(s)**:

Fixes #7442


hedera-mirror-rest/tokens.js
```diff
@@ -573,7 +573,7 @@ const extractSqlFromTokenBalancesRequest = async (tokenId, filters) => {
     query = `select 
         ti.account_id,
         ti.balance,
-        (select max(consensus_end) from record_file) as consensus_timestamp
+        (select max(balance_timestamp) from token_account) as consensus_timestamp
```
steven-sheehy, line 576: This would query max timestamp N times. Need to either use a window function or move max logic to javascript based upon results.

steven-sheehy: You're right, it is not an uncorrelated sub-query. But it does not meet the requirements of the ticket without being related:

```
set the timestamp to the max token_account.balance_timestamp of the entries in the response.
```

Here if the table has 10 million entries and the query returns the earliest 25, the balance timestamp should be the max of the 25 and not that latest 10 millionth entry.

xin-hedera: just want to add that getting the max balance_timestamp from all rows in the table would be a performance issue as well.

stefan-stefanooov: Oh yes, you both are totally right :) 
I'll change it asap.

hedera-mirror-rest/__tests__/specs/accounts/balance-range.json
```diff
@@ -32,25 +32,29 @@
         "token_id": "0.0.99999",
         "account_id": "0.0.18",
         "balance": 20,
-        "created_timestamp": "2300"
+        "created_timestamp": "2300",
+        "balance_timestamp": 2344
```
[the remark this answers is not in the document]
stefan-stefanooov: There are several files with the same name in the project. I guess I opened them by mistake.
Fixed

hedera-mirror-rest/__tests__/specs/tokens/{id}/balances/limit.json
```diff
@@ -44,52 +44,62 @@
...
-        "balance": 350
+        "balance": 350,
+        "balance_timestamp": 5431
       },
       {
         "token_id": "0.20.1",
         "account_id": "0.0.23",
-        "balance": 450
+        "balance": 450,
+        "balance_timestamp": 5432
```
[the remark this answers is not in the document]
stefan-stefanooov: Nice idea. It is done.

hedera-mirror-rest/tokens.js
stefan-stefanooov: @steven-sheehy, @xin-hedera 
I tried to use a pure window function but it didn't work. So I moved to a `Common Table Expression`.
Hope you like it that way.

xin-hedera (approved):
LGTM

steven-sheehy (approved):
LGTM

PR #7487, opened 2024-01-05
Fix Rest API cache-control response header 

### LINKED ISSUE #7486: Rest API incorrect cache control configuration for some endpoints
### Description

We have custom cache-control for several rest api endpoints in application.yml, however, there are two issues:

- the path pattern for transaction by id or hash has changed from `/api/v1/transactions/:transactionId` to `/api/v1/transactioins/:transactionIdOrHash`
- the actual path pattern for `/api/v1/network/exchangerate` and `/api/v1/network/fees` in req.route.path are `/exchangerate` and `/fees` respectively

### Steps to reproduce

run the rest server locally, check for the `cache-control` header in response for different endpoints, compare it against to what's set in application.yml

### Additional context

_No response_

### Hedera network

other

### Version

v0.97.0-SNAPSHOT

### Operating system

None

xin-hedera:
**Description**:

This PR fixes the incorrect cache-control response header configuration

- Fix cache-control path pattern for transaction by id or hash endpoint
- Fix cache-control path pattern for network exchangerate and fees endpoints
- Change swagger metrics authentication default to false and remove override in helm charts

**Related issue(s)**:

Fixes #7486 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-rest/config/application.yml
+++ b/hedera-mirror-rest/config/application.yml
@@ -27,7 +27,7 @@ hedera:
       metrics:
         enabled: true
         config:
-          authentication: true
+          authentication: false
           durationBuckets: [25, 100, 250, 500]
           password: password
           requestSizeBuckets: []
@@ -67,15 +67,14 @@ hedera:
       response:
         compression: true
         headers:
-          default: { "Cache-Control": "public, max-age=1" }
+          default: { "cache-control": "public, max-age=1" }
           path:
-            /api/v1/network/fees: { "Cache-Control": "public, max-age=60" }
-            /api/v1/network/exchangerate:
-              { "Cache-Control": "public, max-age=60" }
-            /api/v1/transactions/:transactionId:
-              { "Cache-Control": "public, max-age=300" }
+            /fees: { "cache-control": "public, max-age=60" }
+            /exchangerate: { "cache-control": "public, max-age=60" }
+            /api/v1/transactions/:transactionIdOrHash:
+              { "cache-control": "public, max-age=300" }
             /api/v1/transactions/:transactionId/stateproof:
-              { "Cache-Control": "public, max-age=600" }
+              { "cache-control": "public, max-age=600" }
         includeHostInLink: false
         limit:
           default: 25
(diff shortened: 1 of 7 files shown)


hedera-mirror-rest/config/application.yml
```diff
@@ -67,15 +67,14 @@ hedera:
       response:
         compression: true
         headers:
-          default: { "Cache-Control": "public, max-age=1" }
+          default: { "cache-control": "public, max-age=1" }
```
xin-hedera, line 70: change to lower case to match what's in express response

hedera-mirror-rest/config/application.yml
```diff
@@ -67,15 +67,14 @@ hedera:
...
-              { "Cache-Control": "public, max-age=60" }
-            /api/v1/transactions/:transactionId:
-              { "Cache-Control": "public, max-age=300" }
+            /api/v1/transactions/:transactionIdOrHash:
+              { "cache-control": "public, max-age=300" }
             /api/v1/transactions/:transactionId/stateproof:
-              { "Cache-Control": "public, max-age=600" }
+              { "cache-control": "public, max-age=600" }
+            /exchangerate: { "cache-control": "public, max-age=60" }
+            /fees: { "cache-control": "public, max-age=60" }
```
xin-hedera, line 77: for handlers mounted under a Router, `req.route.path` only contains the part in the scope of the router. For example, the router is mounted on to `/trunk`, and a handler then is mounted at `/branch` onto the router, inside the handler, `req.route.path` is just `/branch`.

Per expressjs doc, `req.baseUrl` is supposed to be `/trunk`, however it looks like awaitjs/express breaks it.

As a result, the PR resorts to use the path inside the router for exchangerate and fees now.

edwin-greene (approved):
Looks good

hedera-mirror-rest/config/application.yml
xin-hedera: turns out it's not awaitjs/express.  `req.baseUrl` being empty in the responseHandler is expected. Create a follow-up ticket here: https://github.com/hashgraph/hedera-mirror-node/issues/7489

PR #7501, opened 2024-01-08
Replace TimePartitionService Cacheable annotation with caches

### LINKED ISSUE #7393: Performance issue in cached TimePartitionService methods
### Description

`TimePartitionService`'s methods are cached using the handy spring `@Cacheable` annotation, in order to avoid the performance overhead of querying the database to find the partition information.

However, when it's used in `TopicMessageLookupEntityListener.onTopicMessage()` for every topic message, there is huge performance degradation. Some research finds that even though it's counter-intuitive, the overhead introduced by the `@Cacheable` annotation during runtime is the main contributor, per the blog [here](https://dzone.com/articles/cacheable-overhead-spring-0).

Per the blog, switching to manual caching logic in the methods should resolve the issue.

### Steps to reproduce

Run importer with v2 schema, topic message lookup enabled, and 10K TPS HCS, monitor the parser duration.

### Additional context

_No response_

### Hedera network

other

### Version

v0.95.0

### Operating system

None

edwin-greene:
**Description**:
This replaces the `@Cacheable` annotation in `TimePartitionServiceImpl` with caches. In testing this has shown to improve performance.

**Related issue(s)**:

Fixes #7393 

**Notes for reviewer**:
Multiple runs of a unit test where ran comparing the performance of the unaltered @Cacheable methods and the non-Cacheable methods. The unit test was of a loop that performed 100 million identical requests to the methods. The non-Cacheable methods demonstrated a marked performance improvement, performing roughly twice as fast for getOverlappingTimePartitions and four times faster for getTimePartitions: 

@Cacheable measurements:
2024-01-08T19:37:18.006Z  INFO Test worker c.h.m.i.d.TimePartitionServiceTest getOverlappingTimePartitions took 41282 ms for 100,000,000 iterations
2024-01-08T19:40:32.382Z  INFO Test worker c.h.m.i.d.TimePartitionServiceTest getTimePartitions took 33759 ms for 100,000,000 iterations

Non @Cacheable measurements:
2024-01-08T19:15:34.405Z  INFO Test worker c.h.m.i.d.TimePartitionServiceTest getOverlappingTimePartitions took 21869 ms for 100,000,000 iterations
2024-01-08T19:28:26.697Z  INFO Test worker c.h.m.i.d.TimePartitionServiceTest getTimePartitions took 7704 ms for 100,000,000 iterations

**Checklist**

### DIFF
--- a/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/db/TimePartitionServiceImpl.java
+++ b/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/db/TimePartitionServiceImpl.java
@@ -26,9 +26,8 @@ import jakarta.inject.Named;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
-import java.util.Optional;
-import java.util.concurrent.Callable;
 import lombok.CustomLog;
+import org.apache.commons.lang3.StringUtils;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.cache.Cache;
 import org.springframework.cache.CacheManager;
@@ -60,15 +59,25 @@ public class TimePartitionServiceImpl implements TimePartitionService {
 
     @Override
     public List<TimePartition> getOverlappingTimePartitions(String tableName, long fromTimestamp, long toTimestamp) {
-        var overlappingTimePartitions = cacheTimePartitionOverlapLookup(
-                tableName, () -> Optional.of(queryForOverlappingTimePartitions(tableName, fromTimestamp, toTimestamp)));
-        return overlappingTimePartitions.orElse(Collections.emptyList());
+        String cacheKey = StringUtils.joinWith("-", tableName, fromTimestamp, toTimestamp);
+        try {
+            return cacheTimePartitionOverlap.get(
+                    cacheKey, () -> queryForOverlappingTimePartitions(tableName, fromTimestamp, toTimestamp));
+        } catch (Cache.ValueRetrievalException e) {
+            Utility.handleRecoverableError(
+                    "Error looking up timePartition {} from cacheTimePartitionOverlap", cacheKey, e);
+            return Collections.emptyList();
+        }
     }
 
     @Override
     public List<TimePartition> getTimePartitions(String tableName) {
-        var cachedValue = cacheTimePartitionLookup(tableName, () -> Optional.of(queryForTimePartitions(tableName)));
-        return cachedValue.orElse(Collections.emptyList());
+        try {
+            return cacheTimePartition.get(tableName, () -> queryForTimePartitions(tableName));
+        } catch (Cache.ValueRetrievalException e) {
+            Utility.handleRecoverableError("Error looking up timePartition {} from cacheTimePartition", tableName, e);
+            return Collections.emptyList();
+        }
     }
 
     private List<TimePartition> queryForOverlappingTimePartitions(
@@ -128,27 +137,6 @@ public class TimePartitionServiceImpl implements TimePartitionService {
         } catch (Exception e) {
             log.warn("Unable to query time partitions for table {}", tableName, e);
             return Collections.emptyList();
-        }
-    }
-
-    private Optional<List<TimePartition>> cacheTimePartitionOverlapLookup(
-            String tableName, Callable<Optional<List<TimePartition>>> loader) {
-        try {
-            return cacheTimePartitionOverlap.get(tableName, loader);
-        } catch (Cache.ValueRetrievalException e) {
-            Utility.handleRecoverableError(
-                    "Error looking up timePartition {} from cacheTimePartitionOverlap", tableName, e);
-            return Optional.empty();
-        }
-    }
-
-    private Optional<List<TimePartition>> cacheTimePartitionLookup(
-            String tableName, Callable<Optional<List<TimePartition>>> loader) {
-        try {
-            return cacheTimePartition.get(tableName, loader);
-        } catch (Cache.ValueRetrievalException e) {
-            Utility.handleRecoverableError("Error looking up timePartition {} from cacheTimePartition", tableName, e);
-            return Optional.empty();
         }
     }
 }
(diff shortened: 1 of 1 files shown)


hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/db/TimePartitionServiceImpl.java
```diff
@@ -21,33 +21,58 @@
...
+            JdbcTemplate jdbcTemplate) {
+        this.cacheTimePartitionOverlap = cacheManagerOverlapTimePartition.getCache(CACHE_NAME);
+        this.cacheTimePartition = cacheManagerTimePartition.getCache(CACHE_NAME);
+        this.jdbcTemplate = jdbcTemplate;
+    }
 
-    @Cacheable(cacheManager = CACHE_TIME_PARTITION_OVERLAP, cacheNames = CACHE_NAME)
     @Override
     public List<TimePartition> getOverlappingTimePartitions(String tableName, long fromTimestamp, long toTimestamp) {
+        var overlappingTimePartitions = cacheTimePartitionOverlapLookup(
```
xin-hedera, line 63: The cache key should include `tableName`, `fromTimestamp`, and `toTimestamp`.

Remove the function `cacheTimePartitionOverlapLookup` and pull the logic here.

Remove the `Optional` I don't see why it's needed.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/db/TimePartitionServiceImpl.java
```diff
@@ -21,33 +21,58 @@
...
+                tableName, () -> Optional.of(queryForOverlappingTimePartitions(tableName, fromTimestamp, toTimestamp)));
+        return overlappingTimePartitions.orElse(Collections.emptyList());
+    }
+
+    @Override
+    public List<TimePartition> getTimePartitions(String tableName) {
+        var cachedValue = cacheTimePartitionLookup(tableName, () -> Optional.of(queryForTimePartitions(tableName)));
```
xin-hedera, line 70: similarly, remove the function `cacheTimePartitionLookup` and pull up the logic here.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/db/TimePartitionServiceImpl.java
edwin-greene: Sounds good it is now updated.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/db/TimePartitionServiceImpl.java
edwin-greene: Got it, this has been updated.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/db/TimePartitionServiceImpl.java
```diff
@@ -21,33 +21,67 @@
...
 import org.springframework.jdbc.core.RowMapper;
 
 @CustomLog
 @Named
-@RequiredArgsConstructor
 public class TimePartitionServiceImpl implements TimePartitionService {
 
+    private final Cache cacheTimePartitionOverlap;
+    private final Cache cacheTimePartition;
+    private final JdbcTemplate jdbcTemplate;
```
xin-hedera, line 43: move these so they are after the `private static final` variables

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/db/TimePartitionServiceImpl.java
```diff
@@ -21,33 +21,67 @@
...
+            JdbcTemplate jdbcTemplate) {
+        this.cacheTimePartitionOverlap = cacheManagerOverlapTimePartition.getCache(CACHE_NAME);
+        this.cacheTimePartition = cacheManagerTimePartition.getCache(CACHE_NAME);
+        this.jdbcTemplate = jdbcTemplate;
+    }
 
-    @Cacheable(cacheManager = CACHE_TIME_PARTITION_OVERLAP, cacheNames = CACHE_NAME)
     @Override
     public List<TimePartition> getOverlappingTimePartitions(String tableName, long fromTimestamp, long toTimestamp) {
+        String cacheKey = StringUtils.joinWith("-", tableName, fromTimestamp, toTimestamp);
```
xin-hedera, line 62: IntelliJ profiler shows `StringUtils.joinWith` is the hotspot, this should be much faster
```suggestion
        String cacheKey = tableName + "-" +  fromTimestamp + "-" + toTimestamp;
```

xin-hedera (changes requested):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/db/TimePartitionServiceImpl.java
edwin-greene: Got it, they are now moved.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/db/TimePartitionServiceImpl.java
edwin-greene: Thanks! I have updated the concatenation.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/db/TimePartitionServiceImpl.java
```diff
@@ -43,11 +44,44 @@ public class TimePartitionServiceImpl implements TimePartitionService {
...
+            return Collections.emptyList();
+        }
+    }
+
+    @Override
+    public List<TimePartition> getTimePartitions(String tableName) {
+        try {
+            return cacheTimePartition.get(tableName, () -> queryForTimePartitions(tableName));
+        } catch (Cache.ValueRetrievalException e) {
+            Utility.handleRecoverableError("Error looking up timePartition {} from cacheTimePartition", tableName, e);
```
steven-sheehy, line 78: When would this occur? It seems like the both this and get overlapping call `queryForTimePartitions` which already catches any exceptions. So do we need this extra catch here and above?

edwin-greene: Thanks. It should not occur. I have removed both catches.

steven-sheehy (approved):
LGTM

xin-hedera (approved):
LGTM

PR #7515, opened 2024-01-10
fix: Fix null pointer error on create fungible token with inherit account key

### LINKED ISSUE #7508: Create fungible token with KeyValueType.INHERIT_ACCOUNT_KEY fails with null pointer exception
### Description

Contracts created with ethereum transaction do not inherit sender key and by default will have empty admin key.

Contracts created via hapi transaction will inherit this key.

Currently createFungibleToken function fails with null pointer error when it is invoked from ethereum transaction deployed contract which has no admin key, `setAllInheritedKeysTo` fails when the admin key is null and the function is called with keys with `KeyValueType.INHERIT_ACCOUNT_KEY`.

### Steps to reproduce

Deploy EstimatePrecompileContract.sol from test module via a hardhat project. 
Call this create fungible token function to reproduce the scenario.

The **to** in the request body of **contracts/call** should be the address of the deployed via hardhat EstimatePrecompileContract.

image

### Additional context

_No response_

### Hedera network

other

### Version

v0.96.0

### Operating system

None

kselveliev:
**Description**:

Currently **createFungibleToken** function fails with null pointer error 
when it is invoked from ethereum transaction deployed contract which has no admin key.
([Ethereum transaction deployed contract does not inherit sender key](https://github.com/hashgraph/hedera-services/pull/4110))

Function tested is - **Then I call estimateGas with CreateFungibleToken function** from test module.

`TokenCreateWrapper.setAllInheritedKeysTo() ` fails when the admin key is null and the function is called with keys with `KeyValueType.INHERIT_ACCOUNT_KEY`.

**Modifications done in this **PR****

Modify `JKey.convertJKey` method similar to hedera-services.

In case the key is missing return empty key
**hedera-services** logic - >

```
if (jkey.isEmpty()) {
     return jkey.convertJKeyEmpty();
}
```

Added similar logic in **mirror-node**
```
if (jkey == null) {
    return createEmptyKey();
}
```

Services obtains the key from ledger with empty value while mirror-node
store returns null in case the key is missing which is why the above conditions 
are not exactly the same.

Modified **JKey.java**

Added unit tests in **TokenCreatePrecompileTest** and **JKeyTest**

**Related issue(s)**:

Fixes #7508

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-web3/src/main/java/com/hedera/services/jproto/JKey.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/services/jproto/JKey.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2023 Hedera Hashgraph, LLC
+ * Copyright (C) 2023-2024 Hedera Hashgraph, LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


steven-sheehy (approved):
LGTM

PR #7523, opened 2024-01-12
HIP-584 Historical: Set temp contract address on estimate contract create

bilyana-gospodinova:
**Description**:
When there is a call to `address(this)` in a contract constructor the call for estimate gas on contract deploy fails as the address function is not evaluated correctly. This PR modifies the code so that we set a contract address in the message frame and this way when the contract constructor code is called in `AbstractMessageProcessor#codeExecute` the address is resolved correctly.

Fixes https://github.com/hashgraph/hedera-mirror-node/issues/7249


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmTxProcessorImpl.java
```diff
@@ -105,10 +110,11 @@ public HederaEvmTransactionProcessingResult execute(CallServiceParameters params
     protected MessageFrame buildInitialFrame(
             final MessageFrame.Builder baseInitialFrame, final Address to, final Bytes payload, long value) {
         if (Address.ZERO.equals(to)) {
+            var contractAddress = EntityIdUtils.asTypedEvmAddress(entityAddressSequencer.getNewContractId(to));
```
0xivanov, line 113: The issue is when we call `address(this)` when creating a contract. What will happen if we are outside contract constructor and the address `to` is not specified? Would this logic still create temp account? Can this break something in this scenario?

bilyana-gospodinova: The fix will not affect this scenario. I will push some integration tests for the constructor case and for this case.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallEvmCodesTest.java
```diff
@@ -211,4 +249,23 @@ private CallServiceParameters serviceParametersForEvmCodes(final Bytes callData)
...
+    private CallServiceParameters serviceParametersForAddressThis(final Bytes callData) {
+        final var sender = new HederaEvmAccount(SENDER_ADDRESS);
+        if (!areEntitiesPersisted) {
+            persistEntities();
+        }
+
+        return CallServiceParameters.builder()
+                .sender(sender)
+                .value(0L)
+                .receiver(Address.ZERO)
```
IvanKavaldzhiev, line 262: Usually at contract deploy, we don't set a receiver at all. Will these tests work without setting a receiver?

bilyana-gospodinova: No, we get a NullPointerException if the receiver is not set here in the test. Otherwise, in a real scenario when the "to" field is not set, in ContractCallRequest#constructServiceParameters the receiver is set to Address.ZERO so it is basically the same.

IvanKavaldzhiev (approved):
LGTM

steven-sheehy (approved):
LGTM

PR #7546, opened 2024-01-16
Fix `eth_call` with `extcodehash` on system account

### LINKED ISSUE #7517: eth_call for extcodehash against a system accounts returns 500
### Description

When performing eth_call for extcodehash for system account(0.0.356) it returns status 500 internal server error with:
`2024-01-11T10:39:43.978Z ERROR reactor-http-nio-4 c.h.m.w.c.ContractController Generic error:  java.lang.NullPointerException: Cannot invoke "org.hyperledger.besu.evm.account.Account.isEmpty()" because "account" is null`
It is expected to return default result similar to the consensus node.

### Steps to reproduce

1. Deploy the contract with the following function:
```
function getCodeHash(address _address) external view returns (bytes32) {
        bytes32 codehash;
        assembly {
            codehash := extcodehash(_address)
        }
        return codehash;
    }
```

2. Execute eth_call with extcodehash against system account 0.0.356

### Additional context

_No response_

### Hedera network

other

### Version

0.96.0

### Operating system

None

0xivanov:
**Description**:
This PR fixes null pointer caused by the default `ExtCodeHashOperation` implementation for system accounts.
Some of the system accounts do not exist in the db, so the world updater returns null for them.

WIP: added `SystemAccount` class and more logic to the get method in the WorldUpdater in order to prevent returning null for system accounts. This actually fixes the issue without the code hash override, but I am not sure which one is preferred, considering the fact that we will start consuming the smart contract library in the future. 

**Related issue(s)**:

Fixes #7517

**Notes for reviewer**:

**Checklist**


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/account/SystemAccount.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+package com.hedera.mirror.web3.evm.account;
+
+import org.hyperledger.besu.datatypes.Address;
+import org.hyperledger.besu.datatypes.Hash;
+import org.hyperledger.besu.datatypes.Wei;
+import org.hyperledger.besu.evm.fluent.SimpleAccount;
+
+public class SystemAccount extends SimpleAccount {
```
0xivanov, line 24: This class be useful in the future for manipulating different operations for system accounts

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/HederaEvmStackedWorldStateUpdater.java
```diff
@@ -83,6 +83,18 @@ public MutableAccount createAccount(Address address, long nonce, Wei balance) {
 
     @Override
     public Account get(final Address address) {
+        // entities from 1-1000 are system accounts/contracts
+        // some of them are missing in the db
+        // we create dummy empty account or SystemAccount
+        // since some of the system accounts/contracts are not present in the db
+        // and most operations with them are forbidden
+        if (isSystemAccount(address)) {
```
0xivanov, line 91: Might be useful to prevent null pointers when interacting with system accounts

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/operations/CustomExtCodeHashOperation.java
```diff
@@ -0,0 +1,61 @@
...
+import org.hyperledger.besu.evm.EVM;
+import org.hyperledger.besu.evm.frame.ExceptionalHaltReason;
+import org.hyperledger.besu.evm.frame.MessageFrame;
+import org.hyperledger.besu.evm.gascalculator.GasCalculator;
+import org.hyperledger.besu.evm.internal.FixedStack;
+import org.hyperledger.besu.evm.internal.Words;
+import org.hyperledger.besu.evm.operation.ExtCodeHashOperation;
+
+@Named
+public class CustomExtCodeHashOperation extends ExtCodeHashOperation {
```
IvanKavaldzhiev, line 34: We shouldn't copy classes from the modularized services code base, since the consensus node still run on the regular mono code. What I saw is we have a similar override in the mono code - `HederaExtCodeHashOperationV038` and we can copy it, instead. 

This approach is compliant with the future integration of `hedera-smart-contract-service`, since we will then use an override of this operation out of the box.

0xivanov: copied HederaExtCodeHashOperationV038

bilyana-gospodinova (approved):
LGTM

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/config/EvmConfiguration.java
```diff
@@ -238,6 +241,7 @@ EVM evm038(final HederaPrngSeedOperation prngSeedOperation) {
                 mirrorNodeEvmProperties,
                 prngSeedOperation,
                 hederaBlockHashOperation,
+                hederaExtCodeHashOperationV038,
```
0xivanov, line 244: We might want to keep this only for v38, but the default impl does not check for system accounts, so the additional logic MUST be kept

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/AbstractEvmStackedLedgerUpdater.java
```diff
@@ -73,4 +73,8 @@ public void commit() {
         }
         store.commit();
     }
+
+    public static boolean isSystemAccount(Address address) {
```
0xivanov, line 77: This is implemented as bean in https://github.com/hashgraph/hedera-mirror-node/pull/7558 so we merge it first

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/HederaEvmStackedWorldStateUpdater.java
```diff
@@ -83,6 +83,18 @@ public MutableAccount createAccount(Address address, long nonce, Wei balance) {
 
     @Override
     public Account get(final Address address) {
+        // entities from 1-1000 are system accounts/contracts
+        // some of them are missing in the db
```
steven-sheehy, line 87: They are not missing per se, they don't exist in either mirror node or consensus nodes. They are reserved for future use but not created. Below is the range of entities not existing in mirror node and I've confirmed with services that it matches their state as well.

350-399
751-799
802-899

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/HederaEvmStackedWorldStateUpdater.java
steven-sheehy, line 91: Disagree with this approach since it differs from services. Using the appropriate extcodehash operation should resolve this issue and this change is unnecessary and could cause a regression. Specifically for pre-0.38 it should throw an error and not return 0. Should revert changes to this class and remove SystemAccount class.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/config/EvmConfiguration.java
```diff
@@ -300,10 +305,10 @@ gasCalculator, mirrorNodeEvmProperties, createOperationExternalizer()),
                         new HederaEvmCreateOperation(gasCalculator, createOperationExternalizer()),
                         new HederaEvmSLoadOperation(gasCalculator),
                         new HederaExtCodeCopyOperation(gasCalculator, validator),
-                        new HederaExtCodeHashOperation(gasCalculator, validator),
```
steven-sheehy, line 303: We need to support the pre-0.38 behavior in HederaExtCodeHashOperation and the post 0.38 hederaExtCodeHashOperationV038.

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/HederaEvmStackedWorldStateUpdater.java
0xivanov: For pre-v38 we expect to revert with `INVALID_SOLIDITY_ADDRESS`?
In services we have the following addressValidator for v0.34:
```java
 @Provides
    @Singleton
    @V_0_34
    static BiPredicate<Address, MessageFrame> provideAddressValidator(
            final Map<String, PrecompiledContract> precompiledContractMap) {
        final var precompiles = precompiledContractMap.keySet().stream()
                .map(Address::fromHexString)
                .collect(Collectors.toSet());
        return (address, frame) ->
                precompiles.contains(address) || frame.getWorldUpdater().get(address) != null;
    }
```
With this validator and `HederaExtCodeHashOperation` as bean we have the following behaviour:
This would return `true` if we pass precompile address `0x167` because it's a precompile address. This causes null pointer in `HederaExtCodeHashOperation`.
The setup I think is the same as in services for v0.34...

steven-sheehy: You're right, it's actually a bug with using extcodehash on a precompile that's present in <v0.38 that would result in a FAIL_INVALID. Good catch.

So we have to either add the null check to the operation like the v038 and the besu operation does or ensure we never return null like you're doing. If we modify the operation then we have to copy it. So probably your SystemAccount approach is better.

steven-sheehy: Still think we should register both HederaExtCodeHashOperation or HederaExtCodeHashOperationV038 depending upon the version since the behavior between the two is drifting apart.

0xivanov: If we want to revert, different addressValidator needs to be injected to the pre-v0.38 operation. The `SystemAccount` approach would result in returning 0.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/config/EvmConfiguration.java
0xivanov: HederaExtCodeHashOperation added

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/operations/HederaExtCodeHashOperationV038.java
```diff
@@ -0,0 +1,81 @@
...
+ * This class is copied from hedera-services
+ *
+ * Hedera adapted version of the {@link ExtCodeHashOperation}.
+ *
+ * <p>Performs an existence check on the requested {@link Address} Halts the execution of the EVM
+ * transaction with {@link HederaExceptionalHaltReason#INVALID_SOLIDITY_ADDRESS} if the account does
+ * not exist, or it is deleted.
+ */
+@Named
+public class HederaExtCodeHashOperationV038 extends ExtCodeHashOperation {
```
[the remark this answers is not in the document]
0xivanov: Removed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/AbstractLedgerWorldUpdater.java
```diff
@@ -56,4 +56,8 @@ public void deleteAccount(Address address) {
     protected Collection<Address> getDeletedAccounts() {
         return deletedAccounts;
     }
+
+    public static boolean isSystemAccount(Address address) {
+        return address.numberOfLeadingZeroBytes() >= 18 && address.getInt(16) < 1001;
```
steven-sheehy, line 61: This logic duplicates strictSystemAccountDetector and AbstractEvmStackedLedgerUpdater but differs slightly. This does not compare unsigned and allows 1000, which I believe is incorrect. We should consolidate the two.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/AbstractEvmStackedLedgerUpdater.java
```diff
@@ -73,4 +73,8 @@ public void commit() {
...
+        return address.numberOfLeadingZeroBytes() >= 18 && address.getInt(16) < 1001;
```
steven-sheehy, line 78: This logic duplicates strictSystemAccountDetector and AbstractLedgerUpdater but differs slightly. This does not compare unsigned and allows 1000, which I believe is incorrect. We should consolidate the two.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/HederaEvmStackedWorldStateUpdater.java
steven-sheehy: I'm fine with returning zero as that is what we do post 0.38 and pre 0.38 is buggy.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/AbstractEvmStackedLedgerUpdater.java
0xivanov: This is not used actually, i forgot to delete it.

0xivanov: Are you Ok with the usage of `preV38AddressValidator` to fix the pre-v.38 version of the operation? This way .38 is different than the rest and this is the expected behaviour I believe.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/AbstractLedgerWorldUpdater.java
0xivanov: Removed

.github/workflows/rosetta.yml
```diff
@@ -4,8 +4,8 @@ on:
   schedule:
     - cron: "0 1 * * *" # Daily at 1:00 AM
   push:
-    branches: [ main ]
-    tags: [ v* ]
+    branches: [main]
+    tags: [v*]
```
IvanKavaldzhiev, line 8: Probably not expected change in this PR

IvanKavaldzhiev (approved):
LGTM

.github/workflows/rosetta.yml
0xivanov: probably spotless, will revert this change

steven-sheehy (approved):
LGTM

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/AbstractEvmStackedLedgerUpdater.java
steven-sheehy: Yes

PR #7554, opened 2024-01-16
Improve staking reward calculation performance

### LINKED ISSUE #7478: Improve pending rewards calculation performance
### Problem

As the number of accounts / contracts grows in the network, the pending rewards calculation task which runs around UTC midnight daily takes substantially longer to complete. The main factor looks to be the task writes way more data than needed, i.e., the task should only update pending rewards for accounts / contracts which have staked to a node and opted in stacking rewards for the just completed period.

### Solution

Change the task to only calculate and update pending rewards for accounts / contract which have staked to a node and opted in stacking rewards for the just completed period.

Need to pay attention to
- Indirect staking in which account `A` stakes to account `B` and `B` stakes to a node and opts in staking rewards. Though there should be no update for `A` wrt staking rewards, its balance should be added to `B`'s as its staking total
- Handle the situation where an account `X` previously staked to a node and received staking rewards, later it changed its staking settings so no longer receives staking rewards

### Alternatives

_No response_

edwin-greene:
**Description**:
Update the entity stake calculation sql to increase performance of the entity stake reward calculation.

**Related issue(s)**:

Fixes #7478 

**Notes for reviewer**:
This ticket includes a migration that removes unnecessary rows from `entity_stake` and `entity_stake_history`. Testing with mainnet data, the migration completes in approximately 15 minutes.

With this update the `entity_stake` table will only be updated with entities that have staking rewards. Entities that have no rewards will now be filtered out of the staking rewards calculation. Going forward, the `entity_stake` table will no longer be updated with the rewards for the current staking period for all entities (which included entities with 0 rewards), but will rather only have an updated row for entities that have a staking reward. This means that the data for some entities will be for past staking periods.

For example: 

Prior to this update each row contained the staking rewards for the current staking period even if there was no reward: 
```
id: 1000, end_stake_period: 2, pending_rewards: 10, timestamp_range: [10, 20)
id: 2000, end_stake_period: 2, pending_rewards: 0, timestamp_range: [10, 20)
```
With this change old staking periods will be present if an entity had earned a reward in the past but did not earn a reward in the current staking period:
```
id: 1000, end_stake_period: 2, pending_rewards: 10, timestamp_range: [10, 20)
id: 2000, end_stake_period: 1, pending_rewards: 5, timestamp_range: [0, 10)
```

**Checklist**


hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/repository/EntityStakeRepository.java
```diff
@@ -69,25 +68,31 @@ select lower(timestamp_range) as timestamp from entity_history where id = 800
...
             id,
             staked_account_id,
             staked_node_id,
             stake_period_start
           from entity, end_period
-          where deleted is not true and type in ('ACCOUNT', 'CONTRACT') and timestamp_range @> end_period.consensus_timestamp
+          where id = 800 or
+            (deleted is not true and type in ('ACCOUNT', 'CONTRACT') and timestamp_range @> end_period.consensus_timestamp
+                and
+              ((decline_reward is false and (staked_node_id <> -1 or staked_account_id <> 0)) or (decline_reward is true and staked_account_id <> 0))
```
xin-hedera, line 79: rewrite the condition so it's more stragithforward
```suggestion
              (staked_account_id <> 0 or (decline_reward is false and staked_node_id <> -1))
```

do the same for the `entity_history` part

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/repository/EntityStakeRepository.java
```diff
@@ -69,25 +68,31 @@ select lower(timestamp_range) as timestamp from entity_history where id = 800
...
             select
               distinct on (id)
-              decline_reward,
               id,
               staked_account_id,
               staked_node_id,
               stake_period_start
             from entity_history, end_period
-            where deleted is not true and type in ('ACCOUNT', 'CONTRACT') and timestamp_range @> end_period.consensus_timestamp
+            where id = 800 or
```
xin-hedera, line 91: can't include 800 in both subqueries (`entity` and `entity_history`), we may end up having two rows for 800 since we do a `union all`

change it to
```suggestion
            where id <> 800 and
```

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/repository/EntityStakeRepository.java
```diff
@@ -222,15 +224,13 @@ select staked_account_id, sum(balance) as staked_to_me
               where staked_account_id <> 0
               group by staked_account_id
             )
-            insert into entity_stake_temp (decline_reward_start, end_stake_period, id, pending_reward,
+            insert into entity_stake_temp (end_stake_period, id, pending_reward,
```
xin-hedera, line 227: when inserting into `entity_stake_temp`, skip entities with indirect staking, i.e., whose staked_account_id <> 0

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/repository/EntityStakeRepository.java
edwin-greene: Thanks I have updated this.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/repository/EntityStakeRepository.java
edwin-greene: Got it thanks!

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/repository/EntityStakeRepository.java
edwin-greene: I have updated this and added a where clause to not include the indirect staking.

hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/entity/AbstractEntityStake.java
```diff
@@ -33,8 +33,6 @@
 public abstract class AbstractEntityStake implements History {
 
     // The *Start properties are the entity's staking state at the beginning of the endStakePeriod + 1 staking period
-    private boolean declineRewardStart;
```
xin-hedera, line 36: move the comment on line 35 to line 44

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/repository/EntityStakeRepository.java
```diff
@@ -242,19 +244,19 @@ or coalesce(staked_node_id_start, -1) = -1
...
+              (case when ess.staked_node_id = -1 then 0
                     else ess.balance + coalesce(ps.staked_to_me, 0)
                end) as stake_total_start,
               int8range(ep.consensus_timestamp, null) as timestamp_range
             from entity_state_start ess
               left join ending_period_stake_state on entity_id = ess.id
               left join proxy_staking ps on ps.staked_account_id = ess.id,
-              ending_period ep;
+              ending_period ep
+              where ess.staked_account_id = 0;
```
xin-hedera, line 255: make it explicit
```suggestion
              where ess.staked_node_id <> -1;
```

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.92.1__update_entity_stake.sql
```diff
@@ -0,0 +1,4 @@
+alter table if exists entity_stake
+    drop column if exists decline_reward_start;
+alter table if exists entity_stake_history
+    drop column if exists decline_reward_start;
```
xin-hedera, line 4: drop rows `where decline_reward_start is true`
otherwise since the column is dropped, for entities which staked to a node and opted out rewards, if before the next period completes, it changes to opt in rewards, the calculator will accrue rewards in the just completed period while in fact it shouldn't.

hedera-mirror-rest/__tests__/specs/accounts/no-params.json
```diff
@@ -89,28 +89,24 @@
     ],
     "entityStakes": [
       {
-        "decline_reward_start": false,
```
xin-hedera, line 92: update one spec file to test the scenario that the entity's `entity_stake` row reflects its pending rewards at least 2 periods ago since the entity has changed staking settings to

- stake to nothing
- stake to an account
- stake to a node and opt out of rewards

The response should show the entity has 0 pending reward even though the row in `entity_stake` has non-zero value.

xin-hedera (changes requested):

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.92.1__update_entity_stake.sql
edwin-greene: Thanks I have updated the migration to drop those rows.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/repository/EntityStakeRepository.java
edwin-greene: Got it, I have updated it.

hedera-mirror-rest/__tests__/specs/accounts/no-params.json
edwin-greene: Thanks, I have added those cases to `no-params.json`

hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/entity/AbstractEntityStake.java
edwin-greene: Sounds good I have moved it.

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.92.1__update_entity_stake.sql
```diff
@@ -0,0 +1,6 @@
+delete from entity_stake where decline_reward_start is true;
+
```
steven-sheehy, line 2: Should we do any cleanup of existing rows to reduce unnecessary data?

edwin-greene: Thanks I have updated the migration to remove unnecessary rows from entity_stake and entity_stake_history.

steven-sheehy (approved):
LGTM

xin-hedera (approved):
LGTM

PR #7569, opened 2024-01-19
Support RFC5988 next link

### LINKED ISSUE #5879: Support RFC5988 next link
### Problem

As a client, I'd like to use a standardized pagination mechanism like [RFC5988](https://datatracker.ietf.org/doc/html/rfc5988) next links.

### Solution

Inside a middleware, inspect the payload for a `links.next` and set it in the Link response header: 

`Link: <${links.next}>; rel="next"`

### Alternatives

_No response_

stefan-stefanooov:
**Description**:

Inside a middleware, inspect the payload for a `links.next` and set it in the `Link` response header:

`Link: <${links.next}>; rel="next"`

**Related issue(s)**:

Fixes #5879

### DIFF
--- a/hedera-mirror-rest/middleware/responseHandler.js
+++ b/hedera-mirror-rest/middleware/responseHandler.js
@@ -23,8 +23,10 @@ const {
   response: {headers},
 } = config;
 
-const contentTypeHeader = 'Content-Type';
-const applicationJson = 'application/json; charset=utf-8';
+const CONTENT_TYPE_HEADER = 'Content-Type';
+const APPLICATION_JSON = 'application/json; charset=utf-8';
+const LINK_NEXT_HEADER = 'Link';
+const linkNextHeaderValue = (linksNext) => `${linksNext}; rel="next"`
 
 // Response middleware that pulls response data passed through request and sets in response.
 // Next param is required to ensure express maps to this middleware and can also be used to pass onto future middleware
@@ -38,11 +40,16 @@ const responseHandler = async (req, res, next) => {
     res.set(headers.path[req.route.path]);
 
     const code = res.locals.statusCode;
-    const contentType = res.locals[responseContentType] || applicationJson;
+    const contentType = res.locals[responseContentType] || APPLICATION_JSON;
+    const linksNext = res.locals.responseData.links?.next;
     res.status(code);
-    res.set(contentTypeHeader, contentType);
+    res.set(CONTENT_TYPE_HEADER, contentType);
 
-    if (contentType === applicationJson) {
+    if (linksNext) {
+      res.set(LINK_NEXT_HEADER, linkNextHeaderValue(linksNext));
+    }
+
+    if (contentType === APPLICATION_JSON) {
       res.send(JSONStringify(responseData));
     } else {
       res.send(responseData);
--- a/hedera-mirror-rest/middleware/responseHandler.js
+++ b/hedera-mirror-rest/middleware/responseHandler.js
@@ -26,7 +26,7 @@ const {
 const CONTENT_TYPE_HEADER = 'Content-Type';
 const APPLICATION_JSON = 'application/json; charset=utf-8';
 const LINK_NEXT_HEADER = 'Link';
-const linkNextHeaderValue = (linksNext) => `${linksNext}; rel="next"`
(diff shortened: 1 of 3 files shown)


hedera-mirror-rest/middleware/responseHandler.js
```diff
@@ -23,8 +23,10 @@ const {
   response: {headers},
 } = config;
 
-const contentTypeHeader = 'Content-Type';
-const applicationJson = 'application/json; charset=utf-8';
+const CONTENT_TYPE_HEADER = 'Content-Type';
+const APPLICATION_JSON = 'application/json; charset=utf-8';
+const LINK_NEXT_HEADER = 'Link';
+const linkNextHeaderValue = (linksNext) => `${linksNext}; rel="next"`
```
steven-sheehy, line 29: Reading the spec again and it looks like we missed the angled brackets. Please double check the RFC.

```suggestion
const linkNextHeaderValue = (linksNext) => `<${linksNext}>; rel="next"`
```

hedera-mirror-rest/middleware/responseHandler.js
```diff
@@ -38,11 +40,16 @@ const responseHandler = async (req, res, next) => {
     res.set(headers.path[req.route.path]);
 
     const code = res.locals.statusCode;
-    const contentType = res.locals[responseContentType] || applicationJson;
+    const contentType = res.locals[responseContentType] || APPLICATION_JSON;
+    const linksNext = res.locals.responseData.links?.next;
```
steven-sheehy, line 44: Please add tests to `responseHandler.test.js` to verify existence and non-existence.

hedera-mirror-rest/middleware/responseHandler.js
```diff
@@ -38,11 +40,16 @@ const responseHandler = async (req, res, next) => {
...
     res.status(code);
-    res.set(contentTypeHeader, contentType);
+    res.set(CONTENT_TYPE_HEADER, contentType);
 
-    if (contentType === applicationJson) {
+    if (linksNext) {
+      res.set(LINK_NEXT_HEADER, linkNextHeaderValue(linksNext));
```
steven-sheehy, line 49: We have a `responseContentType` in some specs to verify that response header. We should convert that field to an optional response headers map to assert in the json like in `responseHeaders: {"Content-Type": "application/json", "Link": "..."}`. Then update `template.js` to assert the new `responseHeaders`. Please update an existing spec in `__tests__/specs/` that has a next link already populated and set it in `responseHeaders`.

steven-sheehy (changes requested):

hedera-mirror-rest/middleware/responseHandler.js
stefan-stefanooov: Sure thing.

hedera-mirror-rest/middleware/responseHandler.js
stefan-stefanooov: @steven-sheehy (@xin-hedera)
Hello
The above comments are taken care of. 

Regarding this one. 
There is already a functionality, like the one you described, implemented.
You could add 

> "responseHeaders": {
    "Link": "</api/v1/accounts?account.id=lt:0.0.21&account.id=gt:0.0.20&limit=3>; rel=\"next\""
  }

to any `.json` test. And it will be forwarded to the template. [REF](https://github.com/hashgraph/hedera-mirror-node/blob/170b2f86c3a164ddd8ec6d54c7a6d5881238dffe/hedera-mirror-rest/__tests__/integration/template.js#L343)
In here we are matching the passed object to the `headers` property of the `supertest` library response object.

But still we are not getting the headers form `supertest` and I am not sure why.
When doing a real world test and in the unit tests I can see/test the headers.

Could you have a look and advice ? 

PS: 
For example I have added a `responseHeaders` object in `accounts-range.json` and uploaded it here.
The test fails.

xin-hedera: change the key to all lowercase in the spec file, i.e., `Link` -> `link`

stefan-stefanooov: Yeah. worked! Thanks @xin-hedera 
It's ready for review

steven-sheehy (approved):
LGTM

xin-hedera (approved):
LGTM

PR #7593, opened 2024-01-22
Fix failing selfdestruct estimate gas test

### LINKED ISSUE #7592: Estimate gas selfdestruct operation acceptance test fails
### Problem

After the introduction of custom self destruct operations in mirror-node acceptance test fails -> 
I call estimateGas with function that performs self destruct

### Solution

Add from parameter in request body of the acceptance test call for self destruct

### Alternatives

_No response_

kselveliev:
**Description**:
The self destruct estimate gas test is failing since currently estimate gas tests do not pass from value in request body.
The newly added custom self destruct operation makes validations on the from address which causes the test to fail.
Passing a valid from address in request body fixes the issue. 

Modified classes:
Added new method in `AbstractEstimateFeature` -> `validateGasEstimationWithSender` which now passes a from address
Modified self destruct test to use the new method 

**Related issue(s)**:

Fixes #7592

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/AbstractEstimateFeature.java
+++ b/hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/AbstractEstimateFeature.java
@@ -81,6 +81,32 @@ abstract class AbstractEstimateFeature extends AbstractFeature {
     }
 
     /**
+     * Validates the gas estimation for a specific contract call, including the sender's address.
+     * <p>
+     * This method estimates the gas cost for a given contract call considering the sender's address, and then checks
+     * whether the actual gas used falls within an acceptable deviation range. It utilizes the provided call endpoint
+     * to perform the contract call and then compares the estimated gas with the actual gas used.
+     *
+     * @param data            The function signature and method data of the contract call.
+     * @param actualGasUsed   The actual gas amount that was used for the call.
+     * @param solidityAddress The address of the solidity contract.
+     * @throws AssertionError If the actual gas used is not within the acceptable deviation range.
+     */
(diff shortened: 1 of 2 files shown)


0xivanov (dismissed):

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/AbstractEstimateFeature.java
```diff
@@ -80,6 +80,32 @@ protected void validateGasEstimation(String data, ContractMethodInterface actual
...
+                .data(data)
+                .to(solidityAddress)
+                .from(contractClient.getClientAddress())
+                .estimate(true)
+                .build();
+        ContractCallResponse msgSenderResponse = mirrorClient.contractsCall(contractCallRequestBody);
+        int estimatedGas = msgSenderResponse.getResultAsNumber().intValue();
+
+        assertWithinDeviation(actualGasUsed.getActualGas(), estimatedGas, lowerDeviation, upperDeviation);
+    }
```
IvanKavaldzhiev, line 107: To avoid code duplication we can define a private method with `Optional<Address>` argument for the sender, and this method could have all of the duplicated logic with a check whether a sender account was passed - if so we set it to the builder, if not - we don't. Then the original `validateGasEstimation` could pass `Optional.empty()` and the new .`validateGasEstimationWithSender` could pass `contractClient.getClientAddress()`.

kselveliev: Comment addressed

steven-sheehy (approved):
LGTM

PR #7645, opened 2024-01-30
Bump hedera-evm version to 0.46 and besu evm version to 23.10.2 

### LINKED ISSUE #7378: Update the Besu EVM library to version 23.10.2
### Problem

Current version of Besu EVM on mirror node 23.10.0 version

### Solution

Update Besu EVM on mirror node to 23.10.2 version

### Alternatives

_No response_

kselveliev:
**Description**:
This PR bumps the besu-evm library to 23.10.2
It also bumps hedera-evm library to 0.46 since 0.46 has a dependency on changes in 23.10.2 besu-evm.
Updating besu-evm to 23.10.2 won't work with hedera-evm lower than 0.46.
Some changes are required in integration tests particularly for historical data to support the new changes in the libraries.

**Modifications**:

Modified `HederaExtCodeHashOperationV038` operation, now takes `EvmProperties` in the constructor.

Modified `MirrorNodeEvmProperties` has 3 new added methods that need to be ovverriden, these will be used in evm version 46, now they have default false values.

Modified  `HederaEvmStackedWorldStateUpdater` to override `getOrCreate` method. The new hedera-evm 0.46 version `AbstractLedgerEvmWorldUpdater.getOrCreate` method returns null so we need to override it. Previous version `getOrCreate` used the default `getOrCreate` method in `WorldUpdater` interface. Now we override in HederaEvmStackedWorldStateUpdater and copy the code from `WorldUpdater.getOrCreate`.

Modified `ContractCallEvmCodesTest.getBlockHashAfterTheLatestReturnsZero`. Changed the test addresses since the test previously passed existing addresses that were created via `ContractCallTestSetup`. Added new test addresses that are not created in `ContractCallTestSetup`.
This is needed because of new changes introduced in hedera-evm 0.46 -> `AbstractCodeCache.getIfPresent` - has new changes 
image
Now using `entityAccess.isUsable(address)` now checks database for the address parameter.
Previously any non system contract addresses would pass in test without this check.

Modified `ContractCallServiceTest.pureCallWithCustomBlock` changed method name and switched order of given conditions - documented in code.

Modified `ContractCallTestSetup.persistEntities`. This change is required for the new check `entityAccess.isUsable(address)` described above in `AbstractCodeCache.getIfPresent`. Since now `entityAccess.isUsable(address)` will internally check the store for the address and subsequently the db.
We need to set the contract entities to have timestamp range of earlier evm versions so that they could be found during historical tests. 

Added new `recordFileEvm38Latest` this is needed for the HederaBlockHashOperation integration test.

**Related issue(s)**:

Fixes #7378 

**Notes for reviewer**:

**Checklist**


hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallTestSetup.java
```diff
@@ -1126,6 +1126,7 @@ protected void persistEntities() {
         contractAllowancesPersist(senderEntityId, REDIRECT_CONTRACT_ADDRESS, tokenTreasuryEntityId, nftEntityId3);
         exchangeRatesPersist();
         feeSchedulesPersist();
+        historicalBlocksPersist();
```
bilyana-gospodinova, line 1129: I am not sure we need this change. The historical blocks should have earlier timestamps - let's say consensus start = 100 and consensus end = 200. All of the other entities for the latest blocks tests should have later timestamps (200+) than these historical blocks and all of the historical data uses the consensus start and end timestamps of one of the historical blocks.

After the current changes the timestamps for the historical blocks would be later than the persisted entities for the latest blocks. The tests might pass this way but we are now further away from the live scenario and the real chronological flow. So now the tests work because the queries find the historical data as latest and not because it is historical.

My suggestion is to create a historical token treasury entity with which to persist the contracts, persist the historical blocks, persist the historical entities, persist the latest entities so that we can simulate as much as possible the chronological flow.

steven-sheehy (approved):
LGTM

IvanKavaldzhiev:
The update to V0.46 version would be in a separate PR, since we need additional code changes. I will remove the links to these issues.

Nana-EC:
@kselveliev does this PR need to adopt changes to reflect those here https://github.com/hashgraph/hedera-services/pull/11172
If not maybe it's in a different PR but we should utilize Journaled updated in the config for EVM 0.46 also

kselveliev:
> @kselveliev does this PR need to adopt changes to reflect those here [hashgraph/hedera-services#11172](https://github.com/hashgraph/hedera-services/pull/11172) If not maybe it's in a different PR but we should utilize Journaled updated in the config for EVM 0.46 also

Correct, I thought the journaled update was needed only for evm 0.46, but now that I see code in services it is also used in 0.30, 0.34 and 0.38. I've added it in my latest commit.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallTestSetup.java
kselveliev: Comment addressed.
I moved the historical methods in the beginning and they are create first before the latest data(which is the correct flow). The problem actually was that isUsable tried to search for the ERC_CONTRACT_ADDRESS, PRNG_CONTRACT_ADDRESS etc. in db, but it was unable to find them because they were created later. Now i just modified the contract creations and set them the proper timestamp range so that they could be found during historical tests.

IvanKavaldzhiev (approved):
LGTM

steven-sheehy (approved):

PR #7648, opened 2024-01-30
Drop topic message primary index

### LINKED ISSUE #7573: Drop `topic_message_pkey` index
### Problem

The size of the topic_message table keeps increasing and we can optimize it by deleting a redundant index.

### Solution

* Drop `topic_message_pkey` index
* Change `GrpcRepository.findLatest()` to join on transaction to get topic ID
* Change `/api/v1/topics/messages/{timestamp}` to join on transaction to get topic ID
* Run k6 test for `/api/v1/topics/messages/{timestamp}` to verify no regression

### Alternatives

_No response_

xin-hedera:
**Description**:

This PR drops the topic_mesage table primary index

- Drop topic_message_pkey
- Change rest api to find topic id from transaction table and then query topic_message table
- Change grpc TopicMessageRepository.findLatest to first find consensus submit message transactions and then get topic messages

**Related issue(s)**:

Fixes #7573

**Notes for reviewer**:

RPS for v1 /api/v1/topics/messages/{timestamp} dropped about 30%

**Checklist**


hedera-mirror-grpc/src/main/java/com/hedera/mirror/grpc/repository/TopicMessageRepositoryCustomImpl.java
```diff
@@ -73,4 +78,30 @@ public Stream<TopicMessage> findByFilter(TopicMessageFilter filter) {
 
         return typedQuery.getResultList().stream(); // getResultStream()'s cursor doesn't work with reactive streams
     }
+
+    @Override
+    public Stream<TopicMessage> findLatest(long consensusTimestamp, int limit) {
```
steven-sheehy, line 83: Why do we have to use a criteria API query and have two queries instead of a single native query with a CTE or join?

hedera-mirror-grpc/src/main/java/com/hedera/mirror/grpc/repository/TopicMessageRepositoryCustom.java
```diff
@@ -25,4 +25,7 @@ public interface TopicMessageRepositoryCustom {
 
     @Transactional(readOnly = true)
     Stream<TopicMessage> findByFilter(TopicMessageFilter filter);
+
+    @Transactional(readOnly = true)
+    Stream<TopicMessage> findLatest(long consensusTimestamp, int limit);
```
steven-sheehy, line 30: Think a list would be less overhead than stream since we're not paging through the results.
```suggestion
    List<TopicMessage> findLatest(long consensusTimestamp, int limit);
```

hedera-mirror-rest/topicmessage.js
```diff
@@ -78,10 +79,18 @@ const getMessageByConsensusTimestamp = async (req, res) => {
...
 
-  const query = `select *
-                      from ${TopicMessage.tableName}
-                      where ${TopicMessage.CONSENSUS_TIMESTAMP} = $1`;
-  const params = [consensusTimestamp];
+  let query = `select consensus_timestamp, entity_id, type
+    from transaction
+    where consensus_timestamp = $1 and result = 22`;
+  const {rows} = await pool.queryQuietly(query, [consensusTimestamp]);
+  if (rows.length !== 1 || rows[0].type !== consensusSubmitMessageType) {
```
steven-sheehy, line 86: Why not filter by type in the query?

hedera-mirror-grpc/src/main/java/com/hedera/mirror/grpc/repository/TopicMessageRepositoryCustomImpl.java
xin-hedera: Performance issue in v2. A join or a cte would cause all shard query for both tables. With two separate queries, only transaction table query goes to all shards.

steven-sheehy: Yeah, but since this is not enabled didn't think we needed to optimize it at this time. I do think that if we want to keep it separate we can use two static repository methods and join in the service layer instead of using the slower, dynamic criteria API.

steven-sheehy (approved):
LGTM

edwin-greene (approved):
Looks good

PR #7650, opened 2024-01-31
fix: Fix concurrency issue in message call processors

### LINKED ISSUE #7572: Web3 service Intermittently return mixed up values with `/contracts/call` when making requests in quick succession
### Description

When making two separate calls to `/api/v1/contracts/call` rapidly, sometimes the results from one call is returned as the result of the second call. In the below examples, we're invoking the `decimals()` method on the ERC-20 token contract, so we expect a static unchanging value to be returned each time, but occasionally this would not be true.

### Steps to reproduce

Option-1: Directly
The following python script runs two threads: 
- each repeatedly making a post request to `/api/v1/contracts/call`
- Thread-1 expects an '8' returned, and Thread-2 expects an '6'. 
- At random times Thread-1 could receive a '6', and vice versa.

Python script: https://gist.github.com/thusaras-vc/879e4fe7fd022e3aa299710714b07d17

Option-2: Through JSON RPC Relay
- Same as above, but this one targets the JSON RPC Relay (and is written in typescript)
- From the [JSON RPC list](https://docs.hedera.com/hedera/tutorials/more-tutorials/json-rpc-connections), this has not been reproducible against the Hashio endpoint, but it happens for Arkhia and ValidationCloud endpoints

TypeScript script: https://gist.github.com/thusaras-vc/710cf6416d9dc2ab83c2fd401de6809d

### Additional context

```bash
...
...
SAUCE matched 6 vs 6
WHBAR matched 8 vs 8
WHBAR matched 8 vs 8
SAUCE matched 6 vs 6
SAUCE matched 6 vs 6
WHBAR matched 8 vs 8
SAUCE mismatch decimal 8 vs 6!
```

### Hedera network

mainnet

### Version

0.96.0

### Operating system

Linux

kselveliev:
**Description**:
Currently making simultaneous calls of decimals function with multiple threads against erc tokens with different decimals will result in mixed up results. If one token has 10 decimals the other has 8 decimals, when we expect the result to be 10 it might turn out to be 8 and vice versa. This is because our MessageCallProcessors - `MirrorEvmMessageCallProcessor` and `MirrorEvmMessageCallProcessorV30` are singleton beans that extend hedera-evm `HederaEvmMessageCallProcessor` which maintains state with 
```
protected long gasRequirement;
protected Bytes output;
```
Multiple threads calling `executeHederaPrecompile()` will modify simultaneously shared  `gasRequirement` and `output` fields and mix up the results.

This Pr modifies :
`MirrorEvmMessageCallProcessor` override `executeHederaPrecompile()` define `gasRequirement` and `output` as local variables within the executeHederaPrecompile to ensure thread safety.
 
`MirrorEvmMessageCallProcessorV30` override `executeHederaPrecompile()` define `gasRequirement` and `output` as local variables within the executeHederaPrecompile to ensure thread safety.

**Related issue(s)**:

Fixes #7572 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmMessageCallProcessorV30.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmMessageCallProcessorV30.java
@@ -16,13 +16,26 @@
 
 package com.hedera.mirror.web3.evm.contracts.execution;
 
+import static org.apache.tuweni.bytes.Bytes.EMPTY;
+import static org.hyperledger.besu.evm.frame.ExceptionalHaltReason.INSUFFICIENT_GAS;
+import static org.hyperledger.besu.evm.frame.MessageFrame.State.COMPLETED_SUCCESS;
+import static org.hyperledger.besu.evm.frame.MessageFrame.State.EXCEPTIONAL_HALT;
+import static org.hyperledger.besu.evm.frame.MessageFrame.State.REVERT;
+
 import com.hedera.mirror.web3.evm.config.PrecompiledContractProvider;
 import com.hedera.node.app.service.evm.contracts.execution.HederaEvmMessageCallProcessor;
+import com.hedera.node.app.service.evm.store.contracts.AbstractLedgerEvmWorldUpdater;
+import com.hedera.node.app.service.evm.store.contracts.precompile.EvmHTSPrecompiledContract;
 import com.hedera.services.contracts.gascalculator.GasCalculatorHederaV22;
 import jakarta.inject.Named;
+import java.util.Optional;
+import org.apache.tuweni.bytes.Bytes;
 import org.hyperledger.besu.evm.EVM;
+import org.hyperledger.besu.evm.frame.MessageFrame;
 import org.hyperledger.besu.evm.precompile.MainnetPrecompiledContracts;
 import org.hyperledger.besu.evm.precompile.PrecompileContractRegistry;
+import org.hyperledger.besu.evm.precompile.PrecompiledContract;
+import org.hyperledger.besu.evm.tracing.OperationTracer;
 
 @Named
 public class MirrorEvmMessageCallProcessorV30 extends HederaEvmMessageCallProcessor {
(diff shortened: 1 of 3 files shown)


kselveliev:
Not entirely sure if this issue should be handled here or could even be changed in hedera-evm in services.
What do you think @steven-sheehy

steven-sheehy:
> Not entirely sure if this issue should be handled here or could even be changed in hedera-evm in services. What do you think @steven-sheehy

Yes, I think ultimately we want this fixed upstream. Talked to Michael T and he said making it stateless sounds reasonable and seems like it's not used in modularized. It looks like we should split `executeHederaPrecompile()` into two methods with the first part returning the gas and output as `Pair<Long, Bytes>` and the second part doing the tracing and frame update. Maybe for now we can copy the `HederaEvmMessageCallProcessor` and make that change locally until we get the updated evm library with the upstream fix.

IvanKavaldzhiev:
I'm not sure it's worth to modify `HederaEvmMessageCallProcessor` inside `hedera-evm`, which would be soon deprecated. At a given point we would replace this library with modules from the modularized repo. Also, this behaviour shouldn't be an issue in services, since transactions there are executed synchronously one by one.

IvanKavaldzhiev (dismissed):
LGTM

kselveliev:
> > Not entirely sure if this issue should be handled here or could even be changed in hedera-evm in services. What do you think @steven-sheehy
> 
> Yes, I think ultimately we want this fixed upstream. Talked to Michael T and he said making it stateless sounds reasonable and seems like it's not used in modularized. It looks like we should split `executeHederaPrecompile()` into two methods with the first part returning the gas and output as `Pair<Long, Bytes>` and the second part doing the tracing and frame update. Maybe for now we can copy the `HederaEvmMessageCallProcessor` and make that change locally until we get the updated evm library with the upstream fix.

I refactored the code, split the `executeHederaPrecompile()` into 2 methods one calculating gas and output and the other doing the tracing and frame update. Extracted common methods in a base abstract class to avoid code duplication

0xivanov (dismissed):
LGTM

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmMessageCallProcessorV30.java
```diff
@@ -35,4 +39,15 @@ public MirrorEvmMessageCallProcessorV30(
         super(v30, precompiles, precompilesHolder.getHederaPrecompiles());
         MainnetPrecompiledContracts.populateForIstanbul(precompiles, gasCalculator);
     }
+
+    @Override
+    protected void executeHederaPrecompile(
```
steven-sheehy, line 44: Is this method identical in each sub class? If so can move to parent

kselveliev: Identical indeed. I extracted it in parent class.

steven-sheehy (approved):
LGTM

PR #7651, opened 2024-01-31
Update rest test docker image to include init.sql

### LINKED ISSUE #7561: Use init.sql in Rest Module
### Problem

Currently all modules except for the rest module use the init.sql to configure the database for test setup. We should be able to apply a similar configuration to the postgres test container in the rest node module

### Solution

Update the container creation in `globalSetup.js` to include the sql script at the location required by the postgres container for db init.

```
dockerDb = await new PostgreSqlContainer(image)
      .withCopyFilesToContainer([{source: '../../hedera-mirror-common/src/test/resources/init.sql', target: '/docker-entrypoint-initdb.d/init.sql'}])
      .start();
```

### Alternatives

_No response_

edwin-greene:
**Description**:
Updates the rest test docker image to include init.sql

**Related issue(s)**:

Fixes #7561 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-rest/__tests__/globalSetup.js
+++ b/hedera-mirror-rest/__tests__/globalSetup.js
@@ -20,7 +20,7 @@ import path from 'path';
 import {PostgreSqlContainer} from '@testcontainers/postgresql';
 
 const dbName = 'mirror_node';
-const dbNamePrefix = 'INTEGRATION_DATABASE_URL';
+const dockerNamePrefix = 'INTEGRATION_DATABASE_URL';
 const v1DatabaseImage = 'postgres:14-alpine';
 const v2DatabaseImage = 'gcr.io/mirrornode/citus:12.1.1';
 
@@ -40,13 +40,13 @@ const createDbContainers = async (maxWorkers) => {
     'resources',
     'init.sql'
   );
-  const filesToCopy = {
+  const initSqlCopy = {
     source: initSqlPath,
     target: '/docker-entrypoint-initdb.d/init.sql',
   };
   for (let i = 1; i <= maxWorkers; i++) {
     const dockerDb = await new PostgreSqlContainer(image)
-      .withCopyFilesToContainer([filesToCopy])
+      .withCopyFilesToContainer([initSqlCopy])
       .withDatabase(dbName)
       .withPassword('mirror_node_pass')
       .withUsername('mirror_node')
@@ -63,16 +63,15 @@ const createDbContainers = async (maxWorkers) => {
 
 const getDatabases = () => dockerDbs;
 const getDatabaseName = () => dbName;
-
 const getReadOnlyUser = () => 'mirror_rest_java';
 const getReadOnlyPassword = () => 'mirror_rest_java_pass';
 
 const setJestEnvironment = (dockerDb, workerId) => {
-  const ownerUri = `${dbNamePrefix}_${workerId}`;
+  const ownerUri = `${dockerNamePrefix}_${workerId}`;
   process.env[ownerUri] = dockerDb.getConnectionUri();
 };
 
-const getOwnerConnectionUri = (workerId) => process.env[`${dbNamePrefix}_${workerId}`];
+const getOwnerConnectionUri = (workerId) => process.env[`${dockerNamePrefix}_${workerId}`];
 
 export default async function (globalConfig) {
   await createDbContainers(globalConfig.maxWorkers);
--- a/hedera-mirror-rest/__tests__/globalSetup.js
+++ b/hedera-mirror-rest/__tests__/globalSetup.js
@@ -30,16 +30,7 @@ let dockerDbs = [];
 
 const createDbContainers = async (maxWorkers) => {
   const image = isV2Schema() ? v2DatabaseImage : v1DatabaseImage;
-  const initSqlPath = path.join(
-    '..',
-    '..',
-    'hedera-mirror-node',
-    'hedera-mirror-common',
-    'src',
-    'test',
-    'resources',
-    'init.sql'
-  );
+  const initSqlPath = path.join('..', 'hedera-mirror-common', 'src', 'test', 'resources', 'init.sql');
   const initSqlCopy = {
     source: initSqlPath,
     target: '/docker-entrypoint-initdb.d/init.sql',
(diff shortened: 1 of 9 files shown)


hedera-mirror-rest/__tests__/globalSetup.js
```diff
@@ -30,7 +30,17 @@ let dockerDb;
 
 const createDbContainer = async (maxWorkers) => {
   const image = isV2Schema() ? v2DatabaseImage : v1DatabaseImage;
-  dockerDb = await new PostgreSqlContainer(image).start();
+  const initSqlPath = path.resolve('../../hedera-mirror-node/hedera-mirror-common/src/test/resources/init.sql');
```
steven-sheehy, line 33: Use path.join for cross platform compatibility

```suggestion
  const initSqlPath = path.join('..', '..', 'hedera-mirror-node', 'hedera-mirror-common', 'src', 'test', 'resources', 'init.sql');
```

xin-hedera, line 33: umm, doubt we should even have such a ticket.

First of all, the changes here actually don't affect rest api integration tests *much*. The only thing that takes effect when running the integration tests is the username is now `mirror_node`.

The database setup is very different in rest api integration tests than subprojects. In globalSetup.js

- a postgresql container starts
- multiple databases are created, one for each jest worker
- the database connection URL is exported as an env variable

Then, in an integration test module, inside `setupIntegrationTest`, `createPool` will create the global `pool` which points to the database associated with the jest worker.

Doing so, enables us to conserve resources with a single postgresql container while running multiple integration test modules in different jest workers in parallel.

Unless we want to change how it's setup, I don't see the value of using the common init.sql script here.

xin-hedera (changes requested):

hedera-mirror-rest/__tests__/globalSetup.js
steven-sheehy: Is it possible to configure the test setup to use db owner but change the regular user to use `mirror_rest`?

xin-hedera: that'll require some changes (exporting and using owner user and password in addition when running migrations) because the other optimization is the database migrations are done lazily by `flywayMigrate` called by `setupIntegrationTest`. This way the migrations won't run if for a certain jest worker all it does is just unit tests.

edwin-greene: Thanks I have updated it to use the read only user for the Rest tests.

hedera-mirror-rest/__tests__/globalTeardown.js
```diff
@@ -24,6 +24,6 @@ export default async function () {
     fs.rmSync(tmpDir, {force: true, recursive: true});
     console.log(`Removed temp directory ${tmpDir}`);
   }
-  await getDatabase().stop();
+  await getDatabases().forEach((d) => d.stop());
```
xin-hedera, line 27:
```suggestion
  await Promise.all(getDatabases().forEach((d) => d.stop()));
```

hedera-mirror-rest/__tests__/globalSetup.js
```diff
@@ -17,70 +17,64 @@
...
   process.env.MIGRATION_TMP_DIR = tmpDir;
   console.info(`Created temp directory ${tmpDir} for migration status`);
 };
 
-const getDatabase = () => dockerDb;
-const getDatabaseName = () => getDatabaseNameForWorker(process.env.JEST_WORKER_ID);
+const getDatabases = () => dockerDbs;
+const getDatabaseName = () => dbName;
+const getReadOnlyUser = () => 'mirror_rest_java';
+const getReadOnlyPassword = () => 'mirror_rest_java_pass';
```
xin-hedera, line 67:
```suggestion
const getReadOnlyUser = () => 'mirror_rest';
const getReadOnlyPassword = () => 'mirror_rest_pass';
```

hedera-mirror-rest/__tests__/integrationDbOps.js
```diff
@@ -93,7 +103,8 @@ const flywayMigrate = async () => {
 
   const workerId = process.env.JEST_WORKER_ID;
   logger.info(`Using flyway CLI to construct schema for jest worker ${workerId}`);
-  const dbConnectionParams = extractDbConnectionParams(process.env.INTEGRATION_DATABASE_URL);
+  const connectionUri = getOwnerConnectionUri(workerId);
+  const dbConnectionParams = extractDbConnectionParams(connectionUri);
   const apiUsername = `${defaultDbConfig.username}_${workerId}`;
```
xin-hedera, line 108: remove the `apiUsername` hack which is needed when we use multiple databases hosted by a single server.

xin-hedera (changes requested):

hedera-mirror-rest/__tests__/globalSetup.js
edwin-greene: Thanks I have updated it to use path.join

hedera-mirror-rest/__tests__/globalSetup.js
```diff
@@ -17,70 +17,64 @@
...
-    database: dockerDb.getDatabase(),
-    host: dockerDb.getHost(),
-    password: dockerDb.getPassword(),
-    port: dockerDb.getPort(),
-    sslmode: 'DISABLE',
-    user: dockerDb.getUsername(),
+  const initSqlPath = path.join(
+    '..',
+    '..',
+    'hedera-mirror-node',
```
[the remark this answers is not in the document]
edwin-greene: Thanks for spotting this, I have removed the unnecessary directories

hedera-mirror-rest/__tests__/globalTeardown.js
edwin-greene: Sounds good I have added the Promise.all

hedera-mirror-rest/__tests__/globalSetup.js
edwin-greene: Thanks I have made this change, I needed to add that user to init.sql as well.

hedera-mirror-rest/__tests__/globalTeardown.js
xin-hedera: umm, I missed it's `forEach()`, to use Promise.all to chain all the promises, we need to change to `map()`

xin-hedera (approved):
LGTM

steven-sheehy (approved):
LGTM

PR #7657, opened 2024-02-01
HIP-859 Add gas consumed field in ContractResult

### LINKED ISSUE #7543: Add `gas_consumed` field in `contract_result`
# Overview of Package

### Definition

Manually calculate the `gasConsumed` value by iterating over all sidecar actions files for a given transaction during mirror node import and save this value in `contract_result` table. Each sidecar action file consists of the gas value that was consumed by the EVM for the given sub-action, without using the refund policy, that Hedera implies. If we sum up all of the gas values on top of initial intrinsic gas (21 000), then we would get accurate value for the gasConsumed

### **Abstract**

Add additional information for the users about the exact gas that was consumed by the EVM during transaction execution.

### **Motivation**

Currently the only value for gas amount we have in `contract_result` table, which is `gasUsed`, gives us information for the gas the user was charged with. It can greatly differ from the actual amount that was used by the EVM, since we have a 20% refund policy on consensus nodes.

That means if the EVM needs 32 500 gas, but the user is passing 400 000 gas, he will be charged the max(32500, 320000), so it would be 320 000.

In this way we lose track of the actual gas that was charged by the EVM. This information might be useful for enhancing hashscan experience and add additional meta data to REST APIs response.

### **Rationale**

A new logic should be added in `hedera-mirror-importer` in the place where we import sidecar action files coming from the record stream. It should manually sum the `gasUsed` value from these files and the result should be saved in a new `gas_consumed` column in the `contract_result` table. That is possible, since the `gasUsed` file in the sidecars contain the actually consumed gas by the EVM, without taking into account the refund policy described above.

### User stories

1. As a user, I want to open my transaction in hashscan and inspect the actually consumed gas by the EVM, so that I have an understanding of how heavy my transaction was for computation by the EVM.

### **Specification**

1. Add a new column `gas_consumed` in `contract_result` table
2. Enhance importer logic with calculation of the sum of the `gasUsed` field of the transaction's sidecar action files, which are being imported
3. Create and implement a test plan. It should cover test coverage of the `hedera-mirror-importer` itself, as well tests in the `hedera-mirror-rest` module

### **Backwards compatibility**

We’re adding a new field in DB and would enhance the response body of some REST APIs. This shouldn't be a breaking change.

### **Security Implications**

No implications.

### **References/Dependencies**

### LINKED ISSUE #7586: HIP-859: Add a new column `gas_consumed` in `contract_result` table
The new field `gas_consumed` in the `contract_result` table in the mirror node will contain the consumed gas by the EVM.

### LINKED ISSUE #7587: HIP-859: Enhance importer logic for gas_consumed
Manually calculate the gasConsumed value by iterating over all sidecar actions files for a given transaction during mirror node import and save this value in contract_result table. Each sidecar action file consists of the gas value that was consumed by the EVM for the given sub-action, without using the refund policy, that Hedera implies. If we sum up all of the gas values on top of initial intrinsic gas (21 000), then we would get accurate value for the gasConsumed.

Example:

1. The transaction is a ContractCall

gas_consumed = intrinsic_gas + sum(contract_action_gas_used)

2. The transaction is a ContractCreate

gas_consumed = intrinsic_gas + cost(contract_init_bytecode) + sum(contract_action_gas_used)

### LINKED ISSUE #7588: HIP-859: Extend REST APIs with `gas_consumed`
The following REST APIs will be extended with the new field:

1. GET /api/v1/contracts/results
2. GET /api/v1/contracts/results/{transactionIdOrHash}
3. GET /api/v1/contracts/{contractIdOrAddress}/results 
4. GET /api/v1/contracts/{contractIdOrAddress}/results/{timestamp}

### LINKED ISSUE #7589: HIP-859: Historical data migration
For historical contract results we have the following options:

1. For the contract transactions that were executed after we started exporting sidecars from consensus nodes, we can make a DB migration in the mirror node and populate the gasConsumed field for them.
2. For the contract transactions that were executed before the introduction of sidecars or were executed after this period, but have missing sidecar records, we can put a null value for the gasConsumed field. In this case it would mean the value is not applicable and missing.

MrValioBg:
**Description**:

HIP-859 Add gas consumed field in ContractResult

**Related issue(s)**:

Fixes #7543 
Fixes #7589 
Fixes #7588 
Fixes #7587 
Fixes #7586 

**Notes for reviewer**:

Check comments in PR

**Checklist**


steven-sheehy:
There's 5 sub-tasks of https://github.com/hashgraph/hedera-mirror-node/issues/7543 but this PR looks like it's targeting to fix the first 3. Thus it should mark this as fixing each one. Also, as I [mentioned](https://github.com/hashgraph/hedera-mirror-node/issues/7586#issuecomment-1904241844) we should have one migration for both DDL and historical migration so it should probably fix the first 4 issues.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/GasConsumedMigration.java
```diff
@@ -0,0 +1,141 @@
...
+    public GasConsumedMigration(final @Owner JdbcTemplate jdbcTemplate, final SidecarProperties sidecarProperties) {
+        this.jdbcTemplate = new NamedParameterJdbcTemplate(jdbcTemplate);
+        this.sidecarProperties = sidecarProperties;
+    }
+
+    @Override
+    protected void doMigrate() throws IOException {
+        jdbcTemplate.getJdbcOperations().execute(ADD_GAS_CONSUMED_COLUMN);
+
+        if (!sidecarProperties.isEnabled()) {
```
MrValioBg, line 89: Seeking opinions on this @steven-sheehy  
I am think it makes more sense that we should execute the DDL even if sidecars are disabled.

Otherwise, if we opt to execute the DDL for the gas_consumed field only when sidecars are enabled, it might raise several issues:
- We would need to adjust the REST module to handle this case, and it does not sounds conceptually well to return field as null thats not a column in the database or having it appear only under certain conditions. - Documenting the behaviour would be challenging and would make integration testing harder.

However if we want to do it in one migration as [mentioned ](https://github.com/hashgraph/hedera-mirror-node/issues/7586#issuecomment-1904241844), even if every time we have the DDL for gas_consumed get executed, another issue arises:
If the migration is initially executed with disabled sidecars, it wouldn't execute a second time, thus not altering the historical data in contract_result table.

Possible solutions include: 
- Using two migrations. 
- Making it a repeatable migration.

MrValioBg: Note: Will include integration tests in the PR for the migration after the behaviour is specified.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/util/GasCalculatorHelper.java
```diff
@@ -0,0 +1,55 @@
...
+
+    /**
+     * Adds the intrinsic gas cost to the gas consumed by the EVM.
+     * In case of contract deployment, the init code is used to
+     * calculate the intrinsic gas cost. Otherwise, it's fixed at 21_000.
+     *
+     * @param gas The gas consumed by the EVM for the transaction
+     * @param initByteCode The init code of the contract
+     */
+    public static long addIntrinsicGas(final long gas, final ByteString initByteCode) {
```
xin-hedera, line 38: I don't see why we need to pass `gas` to the function, can remove `gas` and rename the function to `getIntrinsicGas` or `calculateIntrinsicGas`

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/GasConsumedMigration.java
```diff
@@ -0,0 +1,129 @@
...
+import org.flywaydb.core.api.MigrationVersion;
+import org.springframework.context.annotation.Lazy;
+import org.springframework.dao.EmptyResultDataAccessException;
+import org.springframework.jdbc.core.JdbcTemplate;
+import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
+
+@Named
+public class GasConsumedMigration extends AbstractJavaMigration {
+
+    private static final String ADD_GAS_CONSUMED_COLUMN =
```
xin-hedera, line 35: unless I missed some requirements, this should be a pure SQL migration instead of a Java based flyway migration.

The bottom line is the DDL should be a pure SQL migration, otherwise integration tests of non-importer modules (e.g., rest api) will fail since we currently don't have a mechanism to run DDL in java migration there.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/GasConsumedMigration.java
xin-hedera: we should add the column regardless of importer config

hedera-mirror-rest/__tests__/integrationDomainOps.js
```diff
@@ -1077,6 +1077,7 @@ const contractResultDefaults = {
   function_result: null,
   gas_limit: 1000,
   gas_used: null,
+  gas_consumed: null,
```
xin-hedera, line 1080: the failures in rest integration tests for `gas_consumed` does not esit are caused by the fact the DDL adding `gas_consumed` is in a java migration as I pointed out

xin-hedera (changes requested):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/util/GasCalculatorHelper.java
```diff
@@ -0,0 +1,55 @@
...
+        if (initByteCode == null) {
+            return gas + TX_BASE_COST;
+        }
+
+        int zeros = 0;
+        for (int i = 0; i < initByteCode.size(); i++) {
+            if (initByteCode.byteAt(i) == 0) {
+                ++zeros;
```
steven-sheehy, line 46: This should short circuit the loop on first non-zero byte. Prefer to use a standard util like ArrayUtils.indexOf() or similar.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/util/GasCalculatorHelper.java
```diff
@@ -0,0 +1,55 @@
...
+            }
+        }
+        final int nonZeros = initByteCode.size() - zeros;
+
+        long costForByteCode = TX_BASE_COST + TX_DATA_ZERO_COST * zeros + ISTANBUL_TX_DATA_NON_ZERO_COST * nonZeros;
```
steven-sheehy, line 51: Is it possible that any initcode from the past would not use istanbul prices? That is, do we need any different price depending upon timestamp?

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/ContractResultServiceImpl.java
```diff
@@ -428,7 +437,9 @@ private ByteString processSidecarRecords(RecordItem recordItem) {
                     stopwatch);
         }
 
-        return failedInitcode;
+        totalGasUsed = GasCalculatorHelper.addIntrinsicGas(totalGasUsed, contractDeployment);
+
+        return new SidecarProcessingResult(failedInitcode, totalGasUsed);
```
steven-sheehy, line 442: Since we don't want a Java migration, it might be more maintainable to not have a separate static util and instead have the intrinsic gas calculation moved to `sidecarProcessingResult.getIntrinsicGas()`.

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/domain/ContractResultServiceImplIntegrationTest.java
```diff
@@ -346,6 +346,84 @@ void processContractActionMissingCaller() {
...
+        // When
+        process(recordItem);
+
+        // Then
+        final var contractResult = contractResultRepository
+                .findById(recordItem.getConsensusTimestamp())
+                .orElse(null);
+        assertThat(contractResult).isNotNull();
+        assertThat(contractResult.getGasConsumed()).isNotNull();
+        assertThat(contractResult.getGasConsumed()).isEqualTo(21150L);
```
steven-sheehy, line 375: Take advantage of assertj's chaining. Same comment for other tests.

```suggestion
        assertThat(contractResultRepository.findById(recordItem.getConsensusTimestamp())
            .isNotNull()
            .extracting(ContractResult::getGasConsumed)
            .isNotNull()
            .isEqualTo(21150L);
```

steven-sheehy (changes requested):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/util/GasCalculatorHelper.java
MrValioBg: The HIP for sidecars, [HIP-513](https://hips.hedera.com/hip/hip-513), is included in release v36, and even before that, we used Istanbul pricing, so we wouldn't need to factor in different pricing. The class GasCalculatorHederaV22 has been there since 01.2022, while the HIP was proposed in 06.2022.

hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/contract/ContractResult.java
```diff
@@ -68,6 +68,8 @@ public class ContractResult implements Persistable<Long> {
 
     private Long gasUsed;
 
+    private Long gasConsumed;
```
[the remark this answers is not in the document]
MrValioBg: Done.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/GasConsumedMigration.java
```diff
@@ -0,0 +1,129 @@
...
+            long entityId = rs.getLong("entity_id");
+
+            long gasConsumed = calculateTotalGasUsed(consensusTimestamp);
+            var initByteCode = fetchInitCode(entityId);
+
+            gasConsumed = GasCalculatorHelper.addIntrinsicGas(gasConsumed, initByteCode);
+
+            jdbcTemplate.update(
+                    UPDATE_CONTRACT_RESULT_SQL,
+                    Map.of("consensusTimestamp", consensusTimestamp, "gasConsumed", gasConsumed));
```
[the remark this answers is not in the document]
MrValioBg: Hmm, how we could test with similar amount of data, from where we could get it?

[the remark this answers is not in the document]
steven-sheehy: Can just generate random data yourself. It doesn't have to be exact, just a good amount of data. Need to time the process and also use explain analyze to confirm the queries are optimal.

hedera-mirror-rest/model/contractResult.js
```diff
@@ -42,6 +42,7 @@ class ContractResult {
   static FUNCTION_RESULT = 'function_result';
   static GAS_LIMIT = 'gas_limit';
   static GAS_USED = 'gas_used';
+  static GAS_CONSUMED = 'gas_consumed';
```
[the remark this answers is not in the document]
MrValioBg: Done.

hedera-mirror-rest/api/v1/openapi.yml
```diff
@@ -2076,7 +2076,13 @@ components:
           example: "0x4a817c800"
         gas_used:
           description: The units of gas used to execute contract
-          example: 1000
+          example: 80000
+          format: int64
+          nullable: true
+          type: integer
+        gas_consumed:
```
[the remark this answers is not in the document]
MrValioBg: Done.

hedera-mirror-rest/viewmodel/contractResultViewModel.js
```diff
@@ -42,6 +42,7 @@ class ContractResultViewModel {
     this.function_parameters = toHexString(contractResult.functionParameters, true);
     this.gas_limit = contractResult.gasLimit;
     this.gas_used = contractResult.gasUsed;
+    this.gas_consumed = contractResult.gasConsumed;
```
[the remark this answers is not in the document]
MrValioBg: Done.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/util/GasCalculatorHelper.java
MrValioBg: Agreed, it would look cleaner to just get the intrinsicgas

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/util/GasCalculatorHelper.java
MrValioBg: I am not sure how it can short circuit, since it gets the byte a the exact index and only tracks zeros, though byteAt seems to not be recommended for iterating, but rather selecting bytes at a random index, so I changed it to a more performant approach.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/ContractResultServiceImpl.java
```diff
@@ -482,4 +491,44 @@ private boolean shouldPersistCreatedContractIDs(RecordItem recordItem) {
...
+                if (b == 0) {
+                    ++zeros;
+                }
+            }
+            final int nonZeros = bytes.length - zeros;
+
+            final long costForByteCode =
+                    TX_BASE_COST + TX_DATA_ZERO_COST * zeros + ISTANBUL_TX_DATA_NON_ZERO_COST * nonZeros;
+
+            return costForByteCode + TX_CREATE_EXTRA;
```
IvanKavaldzhiev, line 531: It seems we would add `TX_CREATE_EXTRA` always, even if we calculate the `gasConsumed` for a contract call?

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.94.0__add_gas_consumed_field.sql
```diff
@@ -0,0 +1,44 @@
...
+     ct.consensus_timestamp,
+     ct.entity_id,
+     COALESCE(SUM(ca.gas_used), 0) AS total_gas_used,
+     c.initcode
+   FROM
+     contract_transaction ct
+     LEFT JOIN contract_action ca ON ct.consensus_timestamp = ca.consensus_timestamp
+     LEFT JOIN contract c ON ct.entity_id = c.id
+   GROUP BY
+     ct.consensus_timestamp, ct.entity_id, c.initcode
```
xin-hedera, line 34: this will add gas consumed by contract deployment to all contract transactions for the same contract id

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.94.0__add_gas_consumed_field.sql
```diff
@@ -0,0 +1,44 @@
...
+ )
+ UPDATE contract_result cr
+ SET gas_consumed = gu.total_gas_used + 21000 +
+   (CASE
+     WHEN gu.initcode IS NOT NULL THEN
+       32000 + count_zero_bytes(gu.initcode) -- 32_000 EXTRA gas cost for contract creation
+     ELSE 0
+   END)
```
xin-hedera, line 42: prefer a function `get_intrinsic_gas` which implements the same logic in Java 

```
set gas_consumed = gu.total_gas_used + get_intrinsic_gas(gu.initcode)
```

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.94.0__add_gas_consumed_field.sql
```diff
@@ -0,0 +1,44 @@
...
+
+ -- Update the contract_result table with gas_consumed values, by summing up the gas used by contract actions and adding the gas cost for contract creation
+ WITH gas_usage AS (
+   SELECT
+     ct.consensus_timestamp,
+     ct.entity_id,
+     COALESCE(SUM(ca.gas_used), 0) AS total_gas_used,
+     c.initcode
+   FROM
+     contract_transaction ct
```
xin-hedera, line 30: shouldn't use `contract_transaction`, contract_action.entity_id isn't necessarily a contract (@jnels124, correct me if I'm wrong), use `contract_result` and `contract_result.entity_id` instead

xin-hedera (changes requested):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/ContractResultServiceImpl.java
xin-hedera: for contract call, the init byte code is null,  so the function only returns `TX_BASE_COST`

steven-sheehy (changes requested):

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.94.0__add_gas_consumed_field.sql
```diff
@@ -0,0 +1,44 @@
+ ALTER TABLE contract_result
+ ADD COLUMN IF NOT EXISTS gas_consumed bigint NULL;
```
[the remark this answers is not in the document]
MrValioBg: Got it. Will transform it to lower case.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/ContractResultServiceImpl.java
```diff
@@ -246,8 +246,8 @@ private void processContractResult(
         contractResult.setTransactionResult(transaction.getResult());
         transactionHandler.updateContractResult(contractResult, recordItem);
 
-        if (sidecarFailedInitcode != null && contractResult.getFailedInitcode() == null) {
-            contractResult.setFailedInitcode(DomainUtils.toBytes(sidecarFailedInitcode));
+        if (sidecarProcessingResult.failedInitByteCode() != null && contractResult.getFailedInitcode() == null) {
+            contractResult.setFailedInitcode(DomainUtils.toBytes(sidecarProcessingResult.failedInitByteCode()));
```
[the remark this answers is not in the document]
MrValioBg: Done, adjusted logic for SidecarProcessingResult & gas consumed calculation

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/ContractResultServiceImpl.java
```diff
@@ -482,4 +491,44 @@ private boolean shouldPersistCreatedContractIDs(RecordItem recordItem) {
...
+        private static final long TX_CREATE_EXTRA = 32_000L;
+
+        /**
+         * Returns the intrinsic gas cost for a contract transaction.
+         * In case of contract deployment, the init code is used to
+         * calculate the intrinsic gas cost. Otherwise, it's fixed at 21_000.
+         *
+         * @param initByteCode The init code of the contract
+         */
+        static long getIntrinsicGas(final ByteString initByteCode) {
```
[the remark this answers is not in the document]
MrValioBg: Got it. I've improved the logic.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/ContractResultServiceImpl.java
```diff
@@ -482,4 +491,44 @@ private boolean shouldPersistCreatedContractIDs(RecordItem recordItem) {
...
                 && recordItem.getHapiVersion().isLessThan(RecordFile.HAPI_VERSION_0_23_0);
     }
+
+    /**
+     * Record representing the result of processing sidecar records.
+     *
+     * @param failedInitByteCode The init code of the contract if the contract creation failed
+     * @param gasConsumed The gas consumed by the EVM for the transaction.
+     */
+    private record SidecarProcessingResult(ByteString failedInitByteCode, Long gasConsumed) {
```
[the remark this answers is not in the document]
MrValioBg: Done, we now only do it once in the constructor.

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.94.0__add_gas_consumed_field.sql
```diff
@@ -0,0 +1,44 @@
...
+
+ -- Function for counting zero and non-zero bytes
+ CREATE OR REPLACE FUNCTION count_zero_bytes(bytea) RETURNS bigint AS $$
```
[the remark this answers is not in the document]
MrValioBg: Got it.

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.94.0__add_gas_consumed_field.sql
MrValioBg: I've refactored the query and pushed the changes. But we have a possible issue raising up.
The contractbytecodes sidecars seem to not be related to any timestamp but just be in the contract table.

If we use the contract_result table, we should verify that a certain result was made after the sidecars were introduced, otherwise we cannot calculate the gas for it.
To do this, in the updated query I've adjusted the logic so it only changes the gas_consumed field  if a transaction has a contract_action sidecar.

It should be possible to have a transaction that has only ContractBytecode without contract actions, right?
If we have that case then we cannot calculate the gas_consumed for those transactions, because we cannot skip the contract actions check, otherwise it may result in modifying transactions that does not have sidecars.

hedera-mirror-rest/api/v1/openapi.yml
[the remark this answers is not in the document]
MrValioBg: What do you refer to as spec tests?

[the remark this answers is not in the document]
steven-sheehy: The results APIs in `hedera-mirror-rest/__tests__/specs/contracts` should have a gas_consumed field with a non-null value for at least some of the specs for each endpoint.

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.94.0__add_gas_consumed_field.sql
MrValioBg: Note: For the cases of contract creates the sum of the contract_action_gas_used is optional and only applicable if the contract perform some additional actions in its constructor.

I think this note in the HIP also concerns this topic.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/ContractResultServiceImpl.java
```diff
@@ -428,7 +438,7 @@ private ByteString processSidecarRecords(RecordItem recordItem) {
                     stopwatch);
         }
 
-        return failedInitcode;
+        return new SidecarProcessingResult(DomainUtils.toBytes(initByteCode), totalGasUsed == 0 ? null : totalGasUsed);
```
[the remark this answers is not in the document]
IvanKavaldzhiev: We want to populate the gasConsumed field only for the top level (main) transaction. The gasConsumed for its nested transactions would be accumulated in this top level value.

For child contract creates we would have a sidecar action file connected to the top level transaction, which would be of action type `CREATE`/`CREATE2`.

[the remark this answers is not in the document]
xin-hedera: That makes sense. Follow up question, does the gas used of the contract action corresponding to the child contract create include the amount consumed by child contract deployment?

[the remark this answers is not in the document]
xin-hedera: However the example in the HIP seems to match my concern:


```
3. The top-level transaction is a ContractCall to a method that have an inner call to another contract

gas_consumed = intrinsic_gas + cost(contract_init_bytecode_for_top_level_create) + sum(contract_action_gas_used_for_top_level_create) + cost(contract_init_bytecode_for_inner_create) + sum(contract_action_gas_used_for_inner_create)

```

Note the init bytecode of the inner create isn't available to the top level contract creates `RecordItem` inside `ContractResultServiceImpl.process()`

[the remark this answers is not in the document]
IvanKavaldzhiev: Yes the `gasUsed ` of the action related to child contract create contains the exact gas that was consumed including initbytecode size calculation. The `gasUsed `field was saved in the `HederaOperationTracer ` after the `CREATE ` operation has taken place and for the internal create calculation the `CustomGasCalculator ` is used, since its invoked by the HederaCreateOperation and HederaCreate2Operation custom opcode overrides.

I think your example from the HIP is mixed. For case 3 we have:
gas_consumed = intrinsic_gas + sum(contract_action_gas_used_for_top_level_call) + sum(contract_action_gas_used_for_inner_call)

steven-sheehy (changes requested):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/ContractResultServiceImpl.java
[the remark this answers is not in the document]
xin-hedera: thanks for the clarification. Yes I mixed the example and the formula, the formula is for example 4 not 3 in the HIP.

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.94.0__add_gas_consumed_field.sql
xin-hedera: afaik, child contract create falls into the category of having byte code sidecar while there is no contract action sidecar record. Per the logic in importer `ContractResultServiceImpl`, gas_consumed in its contract result will be `null`, so the SQL migration is correct.

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.94.0__add_gas_consumed_field.sql
```diff
@@ -0,0 +1,71 @@
...
+    from
+        contract_action
+    group by
+        consensus_timestamp
+), contract_transaction_info as (
+    select
+        cr.consensus_timestamp,
+        cr.contract_id,
+        -- Determine if this is a contract create or call transaction
+        (cr.contract_id = any(cr.created_contract_ids)) as is_creation,
```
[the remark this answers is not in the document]
MrValioBg: Done, I've gone ahead and changes the query to use join with the entity table for finding the relevant create operations.

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.94.0__add_gas_consumed_field.sql
MrValioBg: I tried executing an contract create transaction towards the local node, [only with a basic contract]([https://github.com/hashgraph/hedera-mirror-node/blob/main/hedera-mirror-test/src/t[…]mirror/test/e2e/acceptance/steps/EstimatePrecompileFeature.java](https://github.com/hashgraph/hedera-mirror-node/blob/main/hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/EstimatePrecompileFeature.java#L214)), that does not have anything in its constructor. 

Also placed a breakpoint in the new logic in the importer's ContractResultServiceImpl.

This resulted in having a contract action, not only contract bytecode sidecar, so we probably wouldn't have a case where we have only contract bytecode sidecar, with 0 contract actions.

hedera-mirror-rest/api/v1/openapi.yml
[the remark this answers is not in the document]
MrValioBg: Done.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/GasConsumedMigration.java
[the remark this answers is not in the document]
MrValioBg: Okay, working on this, will let you know, when there is an update.

xin-hedera (changes requested):

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.94.0__add_gas_consumed_field.sql
```diff
@@ -0,0 +1,73 @@
...
+    join contract_action_gas_usage cagu on cti.consensus_timestamp = cagu.consensus_timestamp
+)
+update contract_result cr
+set gas_consumed = gc.total_gas_used +
+                   get_intrinsic_gas(gc.initcode, gc.is_creation)
+from gas_calculation gc
+where cr.consensus_timestamp = gc.consensus_timestamp;
+
+drop function if exists bytes_gas_calculation(bytea);
+drop function if exists get_intrinsic_gas(bytea);
```
xin-hedera, line 73:
```suggestion
drop function if exists get_intrinsic_gas(bytea, boolean);
```

xin-hedera (changes requested):

xin-hedera:
The migration took about 30 minutes to complete in mainnet-staging-na.

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/migration/GasConsumedMigrationTest.java
```diff
@@ -0,0 +1,153 @@
...
+
+@EnabledIfV1
+@Import(DisablePartitionMaintenanceConfiguration.class)
+@RequiredArgsConstructor
+@Tag("migration")
+@TestPropertySource(properties = "spring.flyway.target=1.94.0")
+public class GasConsumedMigrationTest extends ImporterIntegrationTest {
+
+    private final JdbcTemplate jdbcTemplate;
+
```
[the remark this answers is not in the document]
MrValioBg: Done.

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.94.0__add_gas_consumed_field.sql
MrValioBg: Done.

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/migration/GasConsumedMigrationTest.java
```diff
@@ -0,0 +1,153 @@
...
+import org.springframework.jdbc.core.JdbcTemplate;
+import org.springframework.test.context.TestPropertySource;
+import org.springframework.transaction.support.TransactionTemplate;
+import org.springframework.util.StreamUtils;
+
+@EnabledIfV1
+@Import(DisablePartitionMaintenanceConfiguration.class)
+@RequiredArgsConstructor
+@Tag("migration")
+@TestPropertySource(properties = "spring.flyway.target=1.94.0")
```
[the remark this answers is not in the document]
MrValioBg: Adjusted.

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.94.0__add_gas_consumed_field.sql
```diff
@@ -0,0 +1,73 @@
...
+$$ language plpgsql immutable;
+
+-- Function to get intrinsic gas for contract creation
+create or replace function get_intrinsic_gas(initcode bytea, is_creation boolean) returns bigint as $$
+begin
+  -- Always include the base base cost of 21000 gas
+  if is_creation then
+    return 21000 + 32000 + bytes_gas_calculation(initcode); -- Calculate extra gas cost for contract creation
+  else
+    return 21000;
```
IvanKavaldzhiev, line 29: We miss the extra cost of contract call function parameters byte length

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/ContractResultServiceImpl.java
IvanKavaldzhiev: For contract calls we also charge for the parameters input size and we should have `costForByteCode` in both scenarios. Only the `TX_CREATE_EXTRA` differs.

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.94.0__add_gas_consumed_field.sql
```diff
@@ -0,0 +1,73 @@
...
+  end if;
+end;
+$$ language plpgsql immutable;
+
+-- Update contract_result with gas_consumed values only if there are corresponding entries in contract_action
+with contract_action_gas_usage as (
+    select
+        consensus_timestamp,
+        sum(gas_used) as total_gas_used
```
IvanKavaldzhiev, line 38: We should sum only the `gas_used` of the top level sidecar action. The migration script and all queries should be adapter accordingly - to filter out only the sidecar action with depth 0.

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.94.0__add_gas_consumed_field.sql
MrValioBg: Done, its adjusted

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/ContractResultServiceImpl.java
```diff
@@ -405,13 +413,15 @@ private ByteString processSidecarRecords(RecordItem recordItem) {
             } else if (sidecarRecord.hasActions()) {
                 var actions = sidecarRecord.getActions();
                 for (int actionIndex = 0; actionIndex < actions.getContractActionsCount(); actionIndex++) {
-                    processContractAction(actions.getContractActions(actionIndex), actionIndex, recordItem);
+                    final var action = actions.getContractActions(actionIndex);
+                    totalGasUsed = totalGasUsed + action.getGasUsed();
```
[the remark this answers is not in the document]
MrValioBg: Yes, we can filter using call depth indeed. Its done.

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.94.0__add_gas_consumed_field.sql
MrValioBg: Done, adjusted accordingly.

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/migration/GasConsumedMigrationTest.java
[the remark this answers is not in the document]
MrValioBg: After merge from the latest main, had to adjust this back to 1.94.0, as now the script before the one the test class is testing is 1.94.0

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/GasConsumedMigration.java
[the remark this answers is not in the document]
MrValioBg: I managed to populate the database with 1 Million contract_result records and this resulted in the query with the latest updates being completed for aprox 2minutes. 

Maybe we can run again in  mainnet-staging-na, as @xin-hedera did before, so we can have better estimation.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/ContractResultServiceImpl.java
```diff
@@ -384,19 +392,23 @@ private void processCreatedContractEntity(RecordItem recordItem, EntityId contra
...
+            return new SidecarProcessingResult(null, false, null);
         }
 
         var contractBytecodes = new ArrayList<ContractBytecode>();
         int migrationCount = 0;
         var stopwatch = Stopwatch.createStarted();
 
-        for (var sidecarRecord : sidecarRecords) {
-            boolean migration = sidecarRecord.getMigration();
+        long topLevelActionSidecarGasUsed = 0;
```
xin-hedera, line 406: nit

```suggestion
        Long topLevelActionSidecarGasUsed = null;
```

then the ternary expression at the end is not needed

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/ContractResultServiceImpl.java
```diff
@@ -428,7 +444,17 @@ private ByteString processSidecarRecords(RecordItem recordItem) {
                     stopwatch);
         }
 
-        return failedInitcode;
+        // If the payloadBytes is not set, yet
+        // it is a contract call transaction, and we should use the function parameters
+        if (payloadBytes == null) {
+            isContractCreation = false;
+            payloadBytes = transactionBody.getContractCall().getFunctionParameters();
```
xin-hedera, line 451: shouldn't we use function parameters in the transaction record's contract call result? If it's a contract call wrapped in an ethereum transaction, this will return empty string, right?

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.95.0__add_gas_consumed_field.sql
```diff
@@ -0,0 +1,70 @@
...
+        cti.payload,
+        cagu.gas_used
+    from
+        contract_transaction_info cti
+    -- Filter out only contract transactions with contract actions sidecars
+    join contract_action_gas_usage cagu on cti.consensus_timestamp = cagu.consensus_timestamp
+)
+update contract_result cr
+set gas_consumed = gc.gas_used +
+                   get_intrinsic_gas(gc.payload, gc.is_creation)
```
xin-hedera, line 65: sidecar is an optional feature in importer and mirrornode operators can choose to only process certain types of sidecars, so it's possible for a mirronode to process contract action sidecars but not bytecode, then `initcode` can be null.

```suggestion
                   get_intrinsic_gas(coalesce(gc.payload, ''::bytea), gc.is_creation)
```

xin-hedera (changes requested):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/ContractResultServiceImpl.java
MrValioBg: Good suggestion. Applied it.

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.95.0__add_gas_consumed_field.sql
MrValioBg: That's true. I've adjusted it, but in the case where the ContractBytecode sidecar does not exist, do we still want to populate the gas_consumed field for CREATE operations? As this would result in displaying an invalid gas_consumed value?

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/ContractResultServiceImpl.java
MrValioBg: I've adjusted the logic to factor the calculation for the payload of bytes in the case of contract call

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.95.0__add_gas_consumed_field.sql
xin-hedera: We can leave it as is. It's the operator's choice to configure importer per the requirement and thus many things are expected to break as a result.

If we decide to convert the backfill part into a Java migration, we can check the importer configuration and skip it if contract action / contract bytecode sidecars are not enabled.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/ContractResultServiceImpl.java
MrValioBg: Debugging while calling balanceOf operation the `recordItem.getTransactionRecord().getContractCallResult().getFunctionParameters()` returns empty result, because it is only populated when the paired TransactionBody with the RecordItem is not a contract call or contract create. So for HAPI transactions we have to use the current one.

For ethereum transactions we indeed have to use the `recordItem.getTransactionRecord().getContractCallResult().getFunctionParameters()` 
I've refactored the code accordingly.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/ContractResultServiceImpl.java
```diff
@@ -428,7 +446,20 @@ private ByteString processSidecarRecords(RecordItem recordItem) {
                     stopwatch);
         }
 
-        return failedInitcode;
+        // For a contract call we need to use the function parameters from the transaction body
+        if (payloadBytes == null && !isContractCreation) {
+            if (transactionBody.hasEthereumTransaction()) {
```
xin-hedera, line 451: should we change the logic to be more generic? i.e.

```
if transactionBody.hasContractCall()
  get function parameters from the contract call transaction body
else
  get function parameters from the call result in the transaction record
```

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.95.0__add_gas_consumed_field.sql
```diff
@@ -0,0 +1,70 @@
+alter table if exists contract_result
```
xin-hedera, line 1: the version should be V1.94.1, because migration V1.94.0 is targeted for the same milestone / release.

MrValioBg: That makes sense. Understood & I've changed the version.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/ContractResultServiceImpl.java
MrValioBg: Done, I've changed it.

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.94.1__add_gas_consumed_field.sql
```diff
@@ -0,0 +1,48 @@
...
+        end as zero_acount
+    from
+        contract_transaction_info cti
+    -- Filter out only contract transactions with contract actions sidecars
+    join contract_action_gas_usage cagu on cti.consensus_timestamp = cagu.consensus_timestamp
+)
+update contract_result cr
+set gas_consumed = gc.gas_used + creation_cost + 21000 + zero_acount * 4 + (count - zero_acount) * 16
+from gas_calculation gc
+where cr.consensus_timestamp = gc.consensus_timestamp;
```
xin-hedera, line 48: now the migration took less than 10 minutes in mainnet-staging:

UPDATE 3492164
Time: 566801.285 ms (09:26.801)

xin-hedera (approved):
LGTM

steven-sheehy (approved):
LGTM

PR #7689, opened 2024-02-06
Support batch stream file ingestion

### LINKED ISSUE #7316: Batch record file ingest
### Problem

When ingesting historical files, it would speed things up to persist multiple record files in one transaction to reduce database round trips.

### Solution

TBD

### Alternatives

_No response_

steven-sheehy:
**Description**:

* Add a `StreamFileParser.parse(List<StreamFile>)`
* Change parser last stream file lookup from a query to a cached value
* Change `StreamFileData` to lazy load bytes from files
* Change `TopicMessageLookupListener` to use `ParserContext` instead of internal state
* Move `RecordFile` persistence from `SqlEntityListener` to `RecordFileParser`
* Remove unused `onStart()` and `onError()` methods from `StreamFileListener`
* Update release checklist

**Related issue(s)**:

Fixes #7316

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/AbstractStreamFileParser.java
+++ b/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/AbstractStreamFileParser.java
@@ -16,6 +16,7 @@
 
 package com.hedera.mirror.importer.parser;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Stopwatch;
 import com.hedera.mirror.common.domain.StreamFile;
 import com.hedera.mirror.importer.exception.HashMismatchException;
@@ -24,6 +25,9 @@ import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.Timer;
 import java.time.Duration;
 import java.time.Instant;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicReference;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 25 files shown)


hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/AbstractStreamFileParser.java
```diff
@@ -100,32 +118,88 @@ public void parse(T streamFile) {
...
+        boolean success = true;
+        T streamFile = null;
+
+        try {
+            for (int i = 0; i < size; ++i) {
+                streamFile = streamFiles.get(i);
+
+                if (!shouldParse(previous, streamFile)) {
+                    streamFile.clear();
+                    return;
```
jascks, line 138: Should this be a `continue`? Or all all remaining stream files in the `List` assumed to be `!shouldParse` based on consensus start?

steven-sheehy: Possibly. This `shouldParse` will return false and trigger this code only when parsing is disabled or if it receives a timestamp before the last processed timestamp. If parsing is disabled, it will discard all files anyway so doesn't matter continue vs return. If timestamp is before last then we should probably continue instead of return. Though I'm not sure in which scenario we could ever send a list of files where we've processed some but not all of the files in a batch. I'll change to continue to be safe.

jascks (approved):
LGTM

jnels124 (approved):
LGTM

edwin-greene (approved):
Looks good to me

xin-hedera (approved):
LGTM

PR #7701, opened 2024-02-08
Set optimal ZFS + Postgres Settings for V2

### LINKED ISSUE #7613: Investigate Optimal ZFS Postgres Settings
### Problem

There are potential optimizations we can benefit from by appropriate configuration of ZFS and postgres

- [x] Investigate L2ARC cache. Adding a ssd drive to each node and configured as an L2 cache for ZFS can help improve performance
- [x] Configure optimal settings for the ARC cache. By default ZFS will attempt to use half the system RAM (but will give it up to other processes). This causes conflict with the helm chart configurations as we don't take into account this RAM usage and we should probably guarantee arc size as well as RAM needed for stackgres containers
- [x] configure page size of postgres optimally for zfs record size.
- [x] Performance gains with smaller zfs record size compared to compression losses 
- [x] Investigate need for a separate SLOG device
* when testing slog device note that zfs will skip its use if logbias is set to throughput. Tests should be ran without slog device and logbias=throughput vs with slog device with logbias=latency

### Solution

Run 10k ingest and k6 tests to test these configurations and compare results

### Alternatives

_No response_

jnels124:
**Description**:

* Add ability to add NVME L2 ARC
* Configure ZFS ARC to 2GB by default
* Configure ZFS primary cache to only cache metadata
* Configure ZFS to store extended attributes in inodes
* Increase shared buffers to 9GB
* change suggested instance type for worker/coordinator node pools

**Related issue(s)**:

Fixes #7613

**Notes for reviewer**:

**Checklist**


charts/hedera-mirror-common/values.yaml
```diff
@@ -421,20 +421,24 @@ zfs:
...
   enabled: false
   init:
+    arcSizeGb: 2
     diskPrefix: citus
     image:
       pullPolicy: IfNotPresent
       registry: gcr.io
       repository: google.com/cloudsdktool/google-cloud-cli
       tag: slim
+    nvmeDeviceId: ""
```
jnels124, line 431: Thought of maybe just making this a simple boolean to enable l2 arc cache but did it this way to be more flexible with configurations that may use additional ssd's in their deployment

charts/hedera-mirror-common/templates/zfs/configmap-init.yaml
```diff
@@ -192,7 +192,44 @@ data:
           echo "Unable to create pool. Manual intervention necessary"
           exit 1
       fi
+    
+      L2_ARC_NVME_DEVICE_ID="${L2_ARC_NVME_DEVICE_ID:-}"
```
steven-sheehy, line 196: This line is superfluous since if statement will check both that it exists and is nonzero.

docs/database/zfs.md
```diff
@@ -54,8 +63,7 @@ Before performing these steps, ensure that your kubectl is pointing to the corre
 `kubectl config get-contexts`.
 
 1. `helm repo add zfs https://openebs.github.io/zfs-localpv`
-2. `helm dependency build charts/hedera-mirror-common`
-3. `helm upgrade --install mirror charts/hedera-mirror-common --create-namespace -n common --set zfs.enabled=true,stackgres.enabled=true`
+2. `helm dependency build charts/hedera-mirror-common` 3. `helm upgrade --install mirror charts/hedera-mirror-common --create-namespace -n common --set zfs.enabled=true,stackgres.enabled=true`
```
steven-sheehy, line 66: Think a newline was accidentally deleted

docs/database/zfs.md
```diff
@@ -42,8 +42,17 @@ NETWORK="default"
...
+```
+
+If you plan to use an nvme attached ssd for an L2 ARC, be sure to add the below options to the creation of the coordinator and worker node pools.
+Note that having an L2 ARC requires some memory overhead in the primary ARC and you may need to increase its size in the helm chart (`zfs.init.arcSizeGb`).
+You should also set `zfs.init.l2ArcNvmeDeviceId` in the helm chart. If only attaching a single local ssd, this value should be `0`. See [here](https://cloud.google.com/kubernetes-engine/docs/how-to/persistent-volumes/local-ssd-raw#node-pool)
+for more information on the device id.
+
+```shell
+--local-nvme-ssd-block count=1
 ```
```
steven-sheehy, line 56: What's this? Did you intend to show the full gcloud command to create the NVME?

jnels124: This goes with the description above saying to add this option when creating the  worker/coordinator node pools.

steven-sheehy (approved):
LGTM

xin-hedera (approved):
LGTM

PR #7665, opened 2024-02-02
Added the absolute path to test.jar

nirbosl:
**Description**: 
I've added the absolute path to test.jar in order to allow the default image entrypoint to run from within 3rd party frameworks that may use a different working-directory, such as Jenkins, without the need to override the entrypoint.

- Modified ENTRYPOINT in the Dockerfile to include the absolute path for `test.jar`

**Related issue(s)**: https://github.com/swirlds/infrastructure/pull/5430

**Notes for reviewer**:
Adding the absolute path should change nothing whenever the container starts in a plain/vanilla k8s or docker environments, it will however remove the need I have to manually override the entrypoint in order to add a `cd /app` command when I attempt to start the container from within a Jenkins pipeline using a jenkins agent pod.
Apparently, once included as a container in the agent pod, Jenkins mounts its agent home directory and changes the working-dir in that container to `/home/jenkins/agent`

PR #7704, opened 2024-02-08
Expose the entrypoint command using an env var named ENTRYPOINT_COMMAND

nirbosl:
**Description**:
This PR is a follow-up and addition for the change made in #7665 
Jenkins appears to be implicitly overriding the entrypoint/cmd of any container(s) used by its execution agent.
In order to get around that, I've exposed an environment variable in the acceptance container named ENTRYPOINT_COMMAND, which would allow me to grab its value (that should match the entrypoint command itself) inside the jenkins pipeline and simply run it there, without making any customizations or overrides to the command.

- Added ENTRYPOINT_COMMAND environment variable to the acceptance-tests Dockerfile

**Related issue(s)**: https://github.com/swirlds/infrastructure/pull/5430

**Notes for reviewer**:

### DIFF
--- a/hedera-mirror-test/Dockerfile
+++ b/hedera-mirror-test/Dockerfile
@@ -8,4 +8,5 @@ RUN apt-get update \
     && rm -rf /var/lib/apt/lists/*
 USER 1000:1000
 
+ENV ENTRYPOINT_COMMAND="java -jar /app/test.jar"
 ENTRYPOINT ["java", "-jar", "/app/test.jar"]


steven-sheehy (approved):

jascks (approved):
LGTM

xin-hedera:
can you just use docker inspect to find the entrypoint?

for example:

```
$ docker inspect -f '{{.Config.Entrypoint}}' gcr.io/mirrornode/hedera-mirror-test:0.97.1
[java -jar test.jar]
```

nirbosl:
> @nirbosl
> 
> can you just use docker inspect to find the entrypoint?
> 
> for example:
> 
> ```
> $ docker inspect -f '{{.Config.Entrypoint}}' gcr.io/mirrornode/hedera-mirror-test:0.97.1
> [java -jar test.jar]
> ```

I spent all of yesterday looking for a way to do this.
We're running the container in a K8s pod, kubectl does not have access to the image metadata.

I could, in theory, mount docker.sock from the k8s node to the container, then install docker and run an inspect from there, however that poses a very big security vulnerability, as anyone with access to that container would be able to start any container, whenever they want with no restriction using our nodes.

There is another way I could use docker inspect to get it, however it increases the complexity and adds more moving parts (that are not required if we simply add this env var to the container), and that's to use a DIND docker image instead of the current vanilla-alpine in my initContainer, run a docker pull in the initContainer for the image I intend to launch for acceptance, docker inspect the entrypoint command and save it in the pod's shared mount, alongside application.yml that I generate and place there, then once the initContainer completes and acceptance starts, it will run the command saved in the shared mount.

Given that the entrypoint is likely to change once in a blue moon, and while being less dynamic - it is simpler and faster to just use that env var in the container, in my pipeline I'll just execute it using `bash -c $(echo $ENTRYPOINT_COMMAND)`

jascks:
I don't know all the utility of Dockerfiles, but is there any way to derive `ENTRYPOINT` from `ENTRYPOINT_COMMAND`, say:
```
ENTRYPOINT ENTRYPOINT_COMMAND.split(" ")
```
Just so the two don't get out of sync.

nirbosl:
A possible way to do that would be:

ENTRYPOINT["/bin/sh", "-c", "$ENTRYPOINT_COMMAND"]

The above is psuedo-code, however should be fairly accurate if my memory serves.

If we want to take that route, I'd like to test it first to make sure this works as we expect it to, and since it changes the way the command is started, you should also test that this form doesn't break anything on your end where this container is used.

xin-hedera (approved):
LGTM

PR #7737, opened 2024-02-13
Adds sender id to contract result service

### LINKED ISSUE #6831: `from` field is not populated correctly for reverted transaction (e.g. contract call)
### Description

When a user sends a transaction via the JSON RPC relay, the payer of tx is the relayer but the sender should be the signer. That case works as expected for valid transactions. But if the transaction is reverted, as we can see on [these lines in hedera-mirror-node-importer](https://github.com/hashgraph/hedera-mirror-node/blob/main/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/ContractResultServiceImpl.java#L137), the `senderId` equals the `payerId`, which in the user-relayer context is wrong. The proper `senderId` is set only for successful transactions, [here](https://github.com/hashgraph/hedera-mirror-node/blob/main/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/ContractResultServiceImpl.java#L273).

### Steps to reproduce

1. Make a contract call to a non-existing contract
2. Get the tx hash
3. Observe the `from` value in the exposed rest endpoint `/api/v1/contracts/results/<txHash>`

### Additional context

_No response_

### Hedera network

other

### Version

latest

### Operating system

Linux

konstantinabl:
**Description**:

**Related issue(s)**:

Fixes #6831 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/ContractResultServiceImpl.java
+++ b/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/ContractResultServiceImpl.java
@@ -43,6 +43,7 @@ import com.hederahashgraph.api.proto.java.ContractFunctionResult;
 import com.hederahashgraph.api.proto.java.ContractID;
 import com.hederahashgraph.api.proto.java.ResponseCodeEnum;
 import com.hederahashgraph.api.proto.java.TransactionBody;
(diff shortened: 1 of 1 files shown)


konstantinabl:
Issue appears to be fixed and not longer valid

PR #7738, opened 2024-02-13
Improve token account upsert performance in v2

### LINKED ISSUE #7669: Token Account Upserter V2 Performance
### Problem

While ingesting mainnet data in v2, token account upserter is to slow. It averages about 100ms to insert a single or very few rows.

### Solution

The problem is caused by the join on token. Without the join on token, the time to insert drops to 20-30ms  We need a way to optimize to get rid of the join or construct in an efficient manner. 

### Alternatives

_No response_

xin-hedera:
**Description**:

This PR optimizes token account upsert performance in v2

- Change some token_account columns to be nullable, and remove default values
- Remove token table join in token_account upsert SQL, switch to using upsert template
- Set TokenAccount.associated to true in partial token account updates where applicable
- Cache token's freeze default and kyc default in rest api on demand
- Set token account's freeze status and kyc status based on cache if missing
- Cache token in get single token api
- Change web3 to set freeze status and kyc status properly based on the related token
- Start postgresql containers for rest integration test in parallel
- Close owner pool in rest integration test to fix occasional stall and error log

**Related issue(s)**:

Fixes #7669 

**Notes for reviewer**:

**Checklist**


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenAccountRepository.java
```diff
@@ -27,12 +27,43 @@
...
+                             end
+                      end as freeze_status,
+                      case when kyc_status is not null then kyc_status
+                           when t.token_id is not null then
+                             case when t.kyc_key is null then 0
+                                  else 2
+                             end
+                      end as kyc_status
+                    from token_account as ta
+                    left join (select *from token where token_id = :#{#id.tokenId}) as t on true
```
xin-hedera, line 62: will add a space
```suggestion
                    left join (select * from token where token_id = :#{#id.tokenId}) as t on true
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenAccountRepository.java
```diff
@@ -27,12 +27,43 @@
...
+                             end
+                      end as freeze_status,
+                      case when kyc_status is not null then kyc_status
+                           when t.token_id is not null then
+                             case when t.kyc_key is null then 0
+                                  else 2
+                             end
+                      end as kyc_status
+                    from token_account as ta
+                    left join (select * from token where token_id = :#{#id.tokenId}) as t on true
```
xin-hedera, line 62: left join the single token row instead of the token table to make citus happy. Otherwise the query will fail since the two tables in the join have different distribute columns

hedera-mirror-rest/model/cachedToken.js
```diff
@@ -0,0 +1,52 @@
...
+      this.freezeDefault = token.freeze_default;
+    }
+
+    if (token.kyc_key !== undefined) {
+      this.kycDefault = token.kyc_key ? tokenKycValues['REVOKED'] : tokenKycValues['NOT_APPLICABLE'];
+    } else {
+      this.kycDefault = token.kyc_default;
+    }
+
+    this.tokenId = token.token_id;
```
steven-sheehy, line 48: Why do we need tokenId if it's the key in the map?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenAccountRepository.java
```diff
@@ -27,12 +27,43 @@
...
+                      ta.balance_timestamp,
+                      ta.created_timestamp,
+                      ta.timestamp_range,
+                      ta.token_id,
+                      case when freeze_status is not null then freeze_status
+                           when t.token_id is not null then
+                             case when t.freeze_key is null then 0
+                                  when t.freeze_default then 1
+                                  else 2
+                             end
```
steven-sheehy, line 53: Consider creating a `token.freeze_status` and `token.kyc_status` to centralize this logic.

hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/token/AbstractToken.java
```diff
@@ -56,12 +56,18 @@ public class AbstractToken implements History {
     @ToString.Exclude
     private byte[] freezeKey;
 
+    @Enumerated(EnumType.ORDINAL)
+    private TokenFreezeStatusEnum freezeStatus;
```
xin-hedera, line 60: not sure why we don't have db enum type for token freeze status enum and token kyc status enum. Here the two columns are consistent with the counterparts in token_account table, i.e., use ordinal as the value

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenAccountRepository.java
xin-hedera: removed. it was there as not much harm to have some redundancy.

steven-sheehy (approved):
LGTM

edwin-greene (approved):
Looks good to me

PR #7744, opened 2024-02-14
Fix local stream provider regression

### LINKED ISSUE #7740: Importer can't process sidecars from local directory
### Description

When the `importer` of the mirror node is set to read from a local directory instead of downloading record streams from S3 processing of sidecar records fails as the importer expects them to be at a different location.

### Steps to reproduce

1. Run `hedera-local-node` with an image tag of `main` or `0.99.0-SNAPSHOT`
2. Create a transaction that generates a sidecar (run local node smoke test or other `cd test/smoke` `npx hardhat test` in local-node repo)
3. Check logs

### Additional context

I've debugged the issue further and when the processed record stream is a sidecar its `filePath`  is an absolute path, so when prefixed with the `basePath` we get essentially a doubled prefix and therefore the file can't be read from that path. The same problem is not observed with regular recordstream files.

Issue seems to have been introduced with this refactor
https://github.com/hashgraph/hedera-mirror-node/commit/094701aa7918a46d6c5cf4df4b1c5f0d923b7151#diff-a7984887961011ee3508fff318d4f90335c987f1a91c3b7e3d980c304b7c7738R52-R55

### Hedera network

other

### Version

v.0.99.0-SNAPSHOT

### Operating system

macOS

steven-sheehy:
**Description**:

* Fix regression in `LocalStreamFileProvider` that causes duplicate prefixes
* Remove unused code in acceptance tests

**Related issue(s)**:

Fixes #7740

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/StreamFileData.java
+++ b/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/StreamFileData.java
@@ -75,6 +75,10 @@ public class StreamFileData {
         return readStreamFileData(file, StreamFilename.from(file.getPath(), File.separator));
     }
 
+    public static StreamFileData from(@NonNull File file, StreamFilename streamFilename) {
+        return readStreamFileData(file, streamFilename);
+    }
+
     public static StreamFileData from(@NonNull Path basePath, @NonNull StreamFilename streamFilename) {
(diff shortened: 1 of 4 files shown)


jascks (approved):
LGTM

isavov (approved):

xin-hedera (approved):
LGTM

PR #7753, opened 2024-02-15
Add a NFT allowance service

### LINKED ISSUE #7500: Add service layer for NFT allowances in the rest-java module

mgoelswirlds:
Adding NFT allowances service layer and tests.

**Description**:
This PR creates the NFT allowances service.

This PR modifies 
- Adds NftAllowanceService to create the service layer

**Related issue(s)**:

Fixes #7500

**Notes for reviewer**:

**Checklist**


hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/NftAllowanceRequest.java
```diff
@@ -0,0 +1,45 @@
...
+
+@Data
+@Builder
+public class NftAllowanceRequest {
+
+    @NotNull
+    private Long accountId;
+
+    @Builder.Default
+    private Integer limit = 25;
```
xin-hedera, line 32:
```suggestion
    private int limit = 25;
```

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/NftAllowanceRequest.java
```diff
@@ -0,0 +1,45 @@
...
+
+    @Builder.Default
+    private Sort.Direction order = Sort.Direction.ASC;
+
+    @Builder.Default
+    private Boolean owner = true;
```
xin-hedera, line 38:
```suggestion
    private boolean owner = true;
```
believe it's easier to understand by renaming it to `byOwner`, to match the repository function name

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/NftAllowanceServiceImpl.java
```diff
@@ -0,0 +1,52 @@
...
+        var accountId = request.getAccountId();
+        var limit = request.getLimit();
+        var order = request.getOrder();
+        var spenderId = request.getSpenderId();
+        var tokenId = request.getTokenId();
+
+        Pageable pageable;
+        // Set the value depending on the owner flag
+        if (request.getOwner()) {
+            pageable = PageRequest.of(0, limit, Sort.by(order, "owner").and(Sort.by(order, "token_id")));
```
xin-hedera, line 45: isn't the logic of ordering by `owner` column or `spender` column inverted?

when owner=true, we should have `order by spender, token_id`, and when owner=false, it should be `order by owner, token_id`

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/NftAllowanceRequest.java
```diff
@@ -0,0 +1,45 @@
...
+
+    @Builder.Default
+    private Long tokenId = 0L;
+
+    @Builder.Default
+    private Long spenderId = 0L;
```
xin-hedera, line 44: use primitive long and remove `@Builder.Default`

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/NftAllowanceRequest.java
```diff
@@ -0,0 +1,45 @@
...
+import lombok.Data;
+import org.jetbrains.annotations.NotNull;
+import org.springframework.data.domain.Sort;
+
+@Data
+@Builder
+public class NftAllowanceRequest {
+
+    @NotNull
+    private Long accountId;
```
xin-hedera, line 29: rename it to `owner` or `ownerId` and make it a primitive long.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/NftAllowanceServiceImpl.java
```diff
@@ -0,0 +1,52 @@
...
+
+@Named
+@RequiredArgsConstructor
+public class NftAllowanceServiceImpl implements NftAllowanceService {
+
+    private final NftAllowanceRepository repository;
+
+    public List<NftAllowance> getNftAllowances(NftAllowanceRequest request) {
+
+        var accountId = request.getAccountId();
```
xin-hedera, line 36: with the changes requested for the domain class, this should be `owner` or `ownerId`, easier to understand and consistent with the repository method siguature

xin-hedera (changes requested):

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/NftAllowanceServiceImpl.java
mgoelswirlds: Yes. You are correct. I should add a test for the order as well.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/NftAllowanceServiceImpl.java
```diff
@@ -0,0 +1,52 @@
...
+        var ownerId = request.getOwnerId();
+        var limit = request.getLimit();
+        var order = request.getOrder();
+        var spenderId = request.getSpenderId();
+        var tokenId = request.getTokenId();
+
+        Pageable pageable;
+        // Set the value depending on the owner flag
+        if (request.isOwner()) {
+            pageable = PageRequest.of(0, limit, Sort.by(order, "spender").and(Sort.by(order, "token_id")));
```
steven-sheehy, line 45: Can reduce scope if you move it inside if blocks
```suggestion
        // Set the value depending on the owner flag
        if (request.isOwner()) {
            var pageable = PageRequest.of(0, limit, Sort.by(order, "spender").and(Sort.by(order, "token_id")));
```

steven-sheehy, line 45: It seems like we would only ever have 4 possible `Sort` objects. We should create 4 static `Sort` objects as an optimization.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/NftAllowanceService.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+package com.hedera.mirror.restjava.service;
+
+import com.hedera.mirror.common.domain.entity.NftAllowance;
+import java.util.List;
+
+public interface NftAllowanceService {
+
+    List<NftAllowance> getNftAllowances(NftAllowanceRequest request);
```
steven-sheehy, line 24: Prefer higher level interfaces.
```suggestion
    Collection<NftAllowance> getNftAllowances(NftAllowanceRequest request);
```

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/NftAllowanceRequest.java
```diff
@@ -0,0 +1,41 @@
...
+
+    @Builder.Default
+    private Sort.Direction order = Sort.Direction.ASC;
+
+    @Builder.Default
+    private boolean isOwner = true;
+
+    private long tokenId;
+
+    private long spenderId;
```
steven-sheehy, line 40: nit: sort fields

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/NftAllowanceRequest.java
```diff
@@ -0,0 +1,41 @@
...
+
+import lombok.Builder;
+import lombok.Data;
+import org.springframework.data.domain.Sort;
+
+@Data
+@Builder
+public class NftAllowanceRequest {
+
+    private long ownerId;
```
steven-sheehy, line 27: This should be accountId and we should drop spenderId, right? Unless you're checking the owner flag in the controller layer and only populating the appropriate fields.

steven-sheehy (changes requested):

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/NftAllowanceRequest.java
xin-hedera: Perhaps the latter. I provided the feedback based on that the controller layer will set ownerId / spenderId from the values in the account id path parameter and the account id query parameter depending on the owner flag.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/NftAllowanceServiceImpl.java
```diff
@@ -0,0 +1,52 @@
...
+            return repository.findByOwnerAndFilterBySpenderAndToken(ownerId, spenderId, tokenId, pageable);
+        } else {
+            pageable = PageRequest.of(0, limit, Sort.by(order, "owner").and(Sort.by(order, "token_id")));
+            return repository.findBySpenderAndFilterByOwnerAndToken(spenderId, ownerId, tokenId, pageable);
+        }
```
xin-hedera, line 50: quite some query patterns are not supported, I don't think this is by design. for example:

- `/accounts/{id}/allowances/nfts?account.id=lte:100&token.id=lte:200`
- `/accounts/{id}/allowances/nfts?account.id=100&token.id=200'

Is it out of the scope of the ticket?

mgoelswirlds: Yes we considered just supporting 1 or 2 query patterns for now. Since the repository layer will be change to dynamically support all query patters.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/NftAllowanceRequest.java
mgoelswirlds: I will be populating the appropriate fields in the controller layer. i.e. the accountId from the path will be set to the owner /spender ID  and the account.id will be set to the other ID based on the flag .

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/NftAllowanceServiceImpl.java
mgoelswirlds: I believe we will have 6 possible `Sort` objects. Will create static objects.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/NftAllowanceServiceImpl.java
xin-hedera: I would imagine the code at all layers will look very different at the end with full support of the design, so I'm not sure about the approach here...

xin-hedera: also want to point out the existing repository method is wrong, for `account.id=gt:X&token.id=gt:Y`, the repository method will only return nft allowances with spender > X and token_id > Y. However, by the design, it should return rows with spender = X + 1 and token_id > Y and any rows with spender > X + 1.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/NftAllowanceServiceImpl.java
```diff
@@ -0,0 +1,82 @@
...
+        var limit = request.getLimit();
+        var order = request.getOrder();
+        var spenderId = request.getSpenderId();
+        var tokenId = request.getTokenId();
+        var tokenIdOperator = request.getAccountIdOperator();
+
+        //  LT,LTE,EQ,NE are not supported right now. Default is GT.
+        tokenId = getUpdatedEntityId(tokenIdOperator, tokenId);
+
+        // Set the value depending on the owner flag   99
```
steven-sheehy, line 59: What's 99?

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/NftAllowanceServiceImpl.java
```diff
@@ -0,0 +1,82 @@
...
+
+@Named
+@RequiredArgsConstructor
+public class NftAllowanceServiceImpl implements NftAllowanceService {
+
+    public static final String TOKEN_ID = "token_id";
+    public static final String OWNER = "owner";
+    public static final String SPENDER = "spender";
+    private static final Sort OWNER_TOKEN_ASC_ORDER =
+            Sort.by(Sort.Direction.ASC, OWNER).and(Sort.by(Sort.Direction.ASC, TOKEN_ID));
```
steven-sheehy, line 36: A helper method and static imports would clean it up.

```suggestion
    private static final Sort OWNER_TOKEN_ASC_ORDER = sort(ASC, OWNER, TOKEN_ID);
```

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/NftAllowanceServiceImpl.java
```diff
@@ -0,0 +1,82 @@
...
+
+            ownerId = getUpdatedEntityId(accountIdOperator, ownerId);
+            var pageable =
+                    PageRequest.of(0, limit, order.isAscending() ? OWNER_TOKEN_ASC_ORDER : OWNER_TOKEN_DESC_ORDER);
+            return repository.findBySpenderAndFilterByOwnerAndToken(spenderId, ownerId, tokenId, pageable);
+        }
+    }
+
+    private static long getUpdatedEntityId(RangeOperator accountIdOperator, long entityId) {
+        if (accountIdOperator.equals(RangeOperator.GTE)) {
```
steven-sheehy, line 77: Can use type safe comparison with enums.

```suggestion
        if (accountIdOperator == RangeOperator.GTE) {
```

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/Utils.java
```diff
@@ -0,0 +1,36 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hedera.mirror.restjava.common;
+
+import org.springframework.data.domain.Sort;
+
+public class Utils {
```
steven-sheehy, line 21: Nothing in this class seems reusable in any other class. Would be more appropriate back in the service. But can keep it here for now to avoid churn.

steven-sheehy (dismissed):
Unblocking

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/Utils.java
mgoelswirlds: Addressing it here. Since I need another approval for merging.

xin-hedera:
Need to DCO sign the commits

mgoelswirlds:
> Need to DCO sign the commits

Done

steven-sheehy (approved):

xin-hedera (approved):
LGTM

PR #7756, opened 2024-02-15
Allow a configurable contract call data size limit

### LINKED ISSUE #7692: Support configuration of web3 module contract call request data size limit
### Problem

Currently the mirror node hard codes the request data size limit supported by the `ContractController` in the web3 module
https://github.com/hashgraph/hedera-mirror-node/blob/main/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/viewmodel/ContractCallRequest.java#L41

This doesn't allow for local experimentation with tools that may go beyond this limit but also doesn't allow for Mirror Node operators to configure their limit higher or lower if so desired.

### Solution

Make this value configureable so that it can be modified without code change.

Potentially add a `hedera.mirror.web3.call.dataSizeLimit` or `hedera.mirror.web3.call.dataSizeLimit` (whatever is appropriate) that can be configured to set the desired limit.
Default can remain `24576 * 2` i.e. 24KiB

### Alternatives

_No response_

jascks:
**Description**:
Currently the mirror node hard codes the request data size limit supported by the `ContractController` in the web3 module. This doesn't allow for local experimentation with tools that may go beyond this limit but also doesn't allow for Mirror Node operators to configure their limit higher or lower if so desired.

This PR: 
* Defines the `hedera.mirror.web3.evm.maxDataSize` property for defining the maximum allowed data size.
*  Moves validation from `ContractCallRequest` annotation based to utilizing the property value and validating from within `ContractController`.
* Updates `configuration.md` to explain property value given using Spring Boot `DataSize`.
* Tweaks unit tests.

**Related issue(s)**:

Fixes #7692

**Notes for reviewer**:
* This PR is in draft mode as I'm working on more testing.
* Steven and Ivan discussed using a single property and applying it to both contract create and call.

**Checklist**


IvanKavaldzhiev (approved):
LGTM!

steven-sheehy (dismissed):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/ContractController.java
```diff
@@ -113,6 +119,18 @@ private CallServiceParameters constructServiceParameters(ContractCallRequest req
...
 
+    /*
+     * Contract data is represented as hexidecimal digits defined as characters in
+     * a String. So it takes two characters to represent one byte, and the configured max
+     * data size in bytes is doubled for validation of the data length within the request object.
+     */
+    private void validateContractDataSize(final ContractCallRequest request) {
+        var contractDataSizeValidator = new HexValidator(0L, evmProperties.getMaxDataSize().toBytes() * 2L, false);
+        if (!contractDataSizeValidator.isValid(request.getData(), null)) {
+            throw new InvalidParametersException("data field " + HexValidator.MESSAGE);
```
Nana-EC, line 130: Q: would it be valuable to expose the calculated size and size limit in the message so consumers know the threshold.
For example this would allow the relay to inform wallets and many ETH tools that initiate the query of the issue and provide a potential action path for developers. 
Without it devs are stuck with this message and no path to action.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/controller/ContractControllerTest.java
```diff
@@ -230,7 +236,7 @@ void callInvalidValue() {
     @Test
     void exceedingDataCallSizeOnEstimate() {
         final var request = request();
-        request.setData("0x" + BYTES.repeat(4000));
+        request.setData("0x" + BYTES.repeat(20000));
```
Nana-EC, line 239: Q: Is 20k here larger than the 24kb limit set above because of the powers of 2 noted earlier?
If not it may not be clear to devs reading this.
Could specify the size here in a similar fashion using `DataSize.ofKiloBytes(x)`
Point is to make it clear how the data set compares to the data limit in the same units

Nana-EC (commented):
Left some questions to prompt potential improvements

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
```diff
@@ -110,6 +113,10 @@ public class MirrorNodeEvmProperties implements EvmProperties {
     @Min(1)
     private int maxBatchSizeWipe = 10;
 
+    @Getter
+    @DataSizeUnit(DataUnit.KILOBYTES)
```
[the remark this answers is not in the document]
jascks: Are you saying you want the default unit to be simply number of bytes?

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/controller/ContractControllerTest.java
```diff
@@ -68,9 +70,13 @@ class ContractControllerTest {
     @MockBean
     private Bucket bucket;
 
+    @MockBean
+    private MirrorNodeEvmProperties evmProperties;
```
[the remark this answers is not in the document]
jascks: Well, it certainly works well getting that bean into the application context to be injected into the controller. I've still not found a way to push the POJO that way. `@Bean` cannot be used in test classes.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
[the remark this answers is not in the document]
jascks: Done.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/controller/ContractControllerTest.java
jascks: Given that both the call and create data sizes are now the same, I adopted this test scenario to fit. But, `BYTES` is the `String` "6080" (4 hex digits), and those 4 characters are repeated 20,000 times, and the  "0x" is prepended, and we have 80,002 characters, representing 40,000 bytes of data, which is larger than 24KiB.

I can work to make this more clear; something along the lines of repeating two hex digits representing a single byte. Then their repetition is more directly reflected in the results.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/ContractController.java
jascks: I must say that error message is sub-optimal.  For the data size check added to the controller, there are two likely failures: invalid hex string characters and too many characters.  That is, I cannot just assume the failure was due to length, and therefore opted to make the validator description message more encompassing.

Perhaps `HexValidator.isValid()` can be more helpful when provided a real `ConstraintValidatorContext` in terms of being more precise as well as providing additional developer context. Taking a look.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/controller/ContractControllerTest.java
jascks: Please take another look at this test class. I just pushed a change that should make this more obvious about what is going one.

edwin-greene (dismissed):
Looks good

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/ContractController.java
jascks: Okay, if the problem is that the data size is exceeded, the error message indicates the size of the provided data, and the limit that was violated.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/validation/HexValidator.java
```diff
@@ -39,16 +50,25 @@ public void initialize(Hex hex) {
...
         int prefixLength = value.startsWith(HEX_PREFIX) ? HEX_PREFIX.length() : 0;
         int length = value.length() - prefixLength;
-        return length >= minLength && length <= maxLength;
+        boolean isValidLength = length >= minLength && length <= maxLength;
+
+        if (throwOnInvalid && !isValidLength) {
+            throw new InvalidParametersException("data field length of %d characters violates limits of min %d or max %d"
+                    .formatted(length, minLength, maxLength));
+        }
+        return isValidLength;
```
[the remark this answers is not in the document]
steven-sheehy: Actually, we can also consider removing the validation and use the `Bytes.fromHexString()` with a destination size 2nd param in `constructServiceParameters()` to do the validation. It already prints nice error messages too.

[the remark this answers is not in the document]
jascks: Yeah, it was starting to feel a little dirty. Thank your for the suggestions. I will follow up.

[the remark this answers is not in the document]
jascks: When using `Bytes.fromHexString()`, if `destinationSize` is larger than the bytes represented by the hex string, the returned result is left padded with zeros. I don't see an alternative available in the API. Thus, everything will be padded to the configured size in bytes.

While `trimLeadingZeros()` can be called on the result, we won't know whether or not leading zeros may have been present in the original hex string, in which case we're modifying the data if we trim them out. If `destinationSize` is too small, then `IllegalArgumentException` is thrown. So, we can detect that scenario at least.

Since the `String` hexidecimal payload has already be deserialized into memory, perhaps the lazy getter with the regex is the way to go. If that matches okay, then use `Bytes.fromHexString()` as we do now.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/controller/ContractControllerTest.java
jascks: This has been updated.

Nana-EC (approved):

xin-hedera (approved):
LGTM

jnels124 (approved):
LGTM

steven-sheehy (approved):

PR #7781, opened 2024-02-20
Remove entity id based range partition

### LINKED ISSUE #7671: Test V2 With different partition Configurations
### Problem

Currently the v2 schema distributes and partitions several tables on entity id. This will cause sparse partitions on the other entity tables (i.e contract, token, nft, etc). We should test performance with and without this partitioning.

### Solution

- [ ] Run 10k ingest testing with current partitioning schema
- [ ] Run K6 tests
- [ ] Alter schema to make larger id partitions 100,000,000 instead of 1,000,0000 (this should produce a single partitions) run k6 and 10k ingest tests again
- [ ] Alter schema to not do entity id related partitions and run k6 and 10k ingest tests again

### Alternatives

_No response_

xin-hedera:
**Description**:

This PR removes entity id based range table partition in v2 for better db write / read performance.

- Remove entity id based range partition for tables in v2
- Remove related range partition maintenance UDFs in v2

**Related issue(s)**:

Fixes #7671 

**Notes for reviewer**:

**Checklist**


hedera-mirror-importer/src/main/resources/application.yml
```diff
@@ -110,9 +110,6 @@ spring:
       db-name: ${hedera.mirror.importer.db.name}
       db-user: ${hedera.mirror.importer.db.username}
       hashShardCount: 6 # Higher shard count here causes a performance issue due to the nature of how hash indexes work
-      idPartitionSize: 1000000
-      maxEntityId: 50000000
```
jnels124, line 114: These Values can also be removed from the `CommonTestConfiguration`

jnels124 (changes requested):

hedera-mirror-importer/src/main/resources/application.yml
xin-hedera: thanks, removed

steven-sheehy (approved):
LGTM

jnels124 (approved):
LGTM

Issue #7741, opened 2024-02-14
Executing eth_call with balance opcode against system account should return 0


yiliev0:
### Description

Executing `balance` opcode against a system account in the range 0.0.0-0.0.750 returns balance if the account is initialised and it has balance(check 0.0.1 on testnet). It should return 0 following the consensus node EVM behaviour.

### Steps to reproduce

1. Execute eth_call against Testnet with the following request body:
```
{
    "data": "f8b2cb4f0000000000000000000000000000000000000000000000000000000000000001",
    "estimate": false,
    "to": "0000000000000000000000000000000000359976"
}
```

Expected result: The eth_call should return 0 as balance
Actual result: The eth_call returns 21000 as balance.

### Hedera network

testnet

### Version

0.98.0-rc1

### Operating system

None

PR #7788, opened 2024-02-22
Balance of system accounts should return zero

bilyana-gospodinova:
**Description**:
This PR fixes a bug where executing balance opcode against a system account in the range 0.0.0-0.0.750 returns the actual balance instead of zero.

Fixes https://github.com/hashgraph/hedera-mirror-node/issues/7741


steven-sheehy (approved):
LGTM, but we should wait for evm 0.46 PR to merge first since this balance operation 38 should be configured for 46 as well.

IvanKavaldzhiev (approved):
LGTM

steven-sheehy:
> LGTM, but we should wait for evm 0.46 PR to merge first since this balance operation 38 should be configured for 46 as well.

Seems EVM 0.46 PR may take some time so merged this and added a [comment](https://github.com/hashgraph/hedera-mirror-node/pull/7749#discussion_r1500813372) to that PR instead to track.

PR #7815, opened 2024-02-26
Improve v2 performance for token account balance migration

### LINKED ISSUE #7672: TokenAccountBalanceMigration performs poorly in v2
### Problem

This migration struggles to run in v2

### Solution

Improve migration to run in a reasonable amount of time in v2

### Alternatives

_No response_

edwin-greene:
**Description**:
Increases the performance of the token account balance migration with v2

- Fix token balance snapshot CTE to use deduped balance info
- Split the query to reduce the number of timestamp partitions to scan

**Related issue(s)**:

Fixes #7672 

**Notes for reviewer**:
Tested on the Citus cluster with a timestamp range returning ~4 million token balance rows. The migration (without inserting, just selecting the rows to be inserted) completed in 10 to 15 seconds. 

**Checklist**


hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/TokenAccountBalanceMigration.java
```diff
@@ -90,13 +103,29 @@ public String getDescription() {
...
 
     @Override
     protected void doMigrate() {
+        var count = new AtomicInteger();
         var stopwatch = Stopwatch.createStarted();
-        int count = jdbcOperations.update(UPDATE_TOKEN_ACCOUNT_SQL);
+        transactionTemplate.executeWithoutResult(s -> {
+            try {
+                var timestampInfo = jdbcTemplate.queryForObject(
+                        GET_TIMESTAMP_INFO_SQL, Collections.emptyMap(), new DataClassRowMapper<>(TimestampInfo.class));
```
xin-hedera, line 116: can you create an abstract base class, move  GET_TIMESTAMP_INFO_SQL and the inner record `TimestampInfo` over so the code becomes shared by both`InitializeEntityBalanceMigration` and this migration?

edwin-greene: Sounds good, I have updated it with an abstract class that combines the shared functionality.

steven-sheehy (approved):
LGTM

xin-hedera (approved):
LGTM

mgoelswirlds (approved):
LGTM

PR #7818, opened 2024-02-27
Update expected gas in estimateGas acceptance tests

bilyana-gospodinova:
**Description**:
After the upgrade of Besu the gas cost for some operations is now different and in this PR we update the expected gas usage for the different test cases.

In this PR we fix also some failing tests such as:
- I call estimateGas with function balance of address
- I call estimateGas with CreateNFT function
- I verify the response from non existing account


hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/HistoricalFeature.java
```diff
@@ -853,6 +859,18 @@ private void waitForNextBlock() {
...
+        try {
+            Awaitility.await()
+                    .atMost(3, TimeUnit.SECONDS)
+                    .pollInterval(1, TimeUnit.SECONDS)
+                    .ignoreExceptions()
+                    .until(() -> mirrorClient.getAccountDetailsUsingEvmAddress(accountId).getEvmAddress() != null);
+        } catch (ConditionTimeoutException e) {
+            log.info("The account could not be imported in the mirror node for 3 seconds.");
+        }
+    }
```
steven-sheehy, line 872: I don't think this awaitility approach is necessary. `mirrorClient.getAccountDetailsUsingEvmAddress` and indeed every `MirrorNodeClient` REST API call has a retry spec that's configured from `RestPollingProperties`. Should instead just mirrorclient directly.

bilyana-gospodinova: This is what I initially thought too. However, it seems that `MirrorNodeClient.getAccountDetailsUsingEvmAddress` call is very fast and it gets called 20 times and the retries are exhausted before the account is actually imported in the mirror node so overall the test fails. At least locally on my environment. This is why I took this approach - in order to avoid some flaky behavior.

yiliev0 (approved):
LGTM

steven-sheehy (approved):
LGTM

PR #7827, opened 2024-02-29
Update Rest test containers to be created in integration tests

### LINKED ISSUE #7698: Rest tests intermittently fail in CI
### Description

The Rest tests sometimes fail in CI, when reran they succeed. It seems that the tests have become flaky because of the recent change to have a docker db container for each Jest worker. See https://github.com/hashgraph/hedera-mirror-node/issues/7561

### Steps to reproduce

1. See rest tests in CI. 

### Additional context

```
> Task :rest:test
(Use `node --trace-warnings ...` to show where the warning was created)
(node:2942) ExperimentalWarning: VM Modules is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
PASS __tests__/utils.test.js (4696 ms)
(node:2943) ExperimentalWarning: VM Modules is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
PASS __tests__/controllers/contractController.test.js (5643 ms)
PASS __tests__/tokens.test.js (1417 ms)
PASS __tests__/utilsFilters.test.js (1187 ms)
PASS __tests__/transactions.test.js (1116 ms)
PASS __tests__/controllers/accountController.test.js (1540 ms)
ERROR: Unable to obtain connection from database (jdbc:postgresql://localhost:[32](https://github.com/hashgraph/hedera-mirror-node/actions/runs/7802341291/job/21279568149#step:5:33)770/mirror_node) for user 'mirror_node': Connection to localhost:32770 refused. Check that the hostname and port are correct and that the postmaster is accepting TCP/IP connections.
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
SQL State  : 08001
Error Code : 0
Message    : Connection to localhost:32770 refused. Check that the hostname and port are correct and that the postmaster is accepting TCP/IP connections.

Caused by: org.postgresql.util.PSQLException: Connection to localhost:32770 refused. Check that the hostname and port are correct and that the postmaster is accepting TCP/IP connections.
Caused by: java.net.ConnectException: Connection refused
```

### Hedera network

other

### Version

main

### Operating system

None

### LINKED ISSUE #7766: Create PostgreSQL containers in rest integration test on demand
### Problem

In rest api test suite, one postgresql container is created for each jest worker, regardless of whether the worker will run any integration test suite or not. We should change the strategy to create the containers on demand only when the jest worker runs the first integration test suite. This will likely cut the total time when running local with many workers, but not much for CI (only 2 workers).

### Solution

- Remove postgresql container creation in globalSetup.js
- Create and start postgresql container on demand in integration test beforeAll, with testcontainers reuse flag, https://node.testcontainers.org/features/containers/#reusing-a-container
- Write the container's id to a per jest work file into a temp directory created by and shared with the jest parent process (visible to globalSetup/globalTeardown). To get the container id, use `StartedTestContainer.getId()`
- in globalTeardown, iterate through the per jest worker container id file, and given `id`, stop the container:
```javascript
import {getContainerRuntimeClient} from 'testcontainers';

const client = await getContainerRuntimeClient();
const container = client.container.getById(id);
await client.container.stop(container);
```

### Alternatives

Keep it as is

edwin-greene:
**Description**:

Updates the Rest test docker container creation to be called by integration tests and allows the docker containers to be reused. 

**Related issue(s)**:

Fixes #7766
Fixes #7698

**Notes for reviewer**:

**Checklist**


hedera-mirror-rest/__tests__/globalTeardown.js
```diff
@@ -14,16 +14,10 @@
...
 import log4js from 'log4js';
 
 export default async function () {
-  const tmpDir = process.env.MIGRATION_TMP_DIR;
-  if (tmpDir) {
-    fs.rmSync(tmpDir, {force: true, recursive: true});
-    console.log(`Removed temp directory ${tmpDir}`);
-  }
-  await Promise.all(getDatabases().map((d) => d.stop()));
+  await integrationDbOps.stopDockerContainers();
```
xin-hedera, line 21: i don't think this can work, they are run in different OS processes so every process has its own, different copy of `dockerContainerIds` in module `integrationDbOps`. what you end up doing here in globalTeardown is stop nothing

hedera-mirror-rest/__tests__/integrationDbOps.js
```diff
@@ -20,19 +20,21 @@ import os from 'os';
...
 
 import config from '../config';
-import {
-  getOwnerConnectionUri,
-  getDatabaseName,
-  getReadOnlyUser,
-  getReadOnlyPassword,
-  getDatabases,
-} from './globalSetup';
-import {getModuleDirname, isV2Schema} from './testutils';
```
xin-hedera, line 30: should import and use the function `isV2Schema` from `testutils`

hedera-mirror-rest/__tests__/integrationDbOps.js
```diff
@@ -74,36 +78,51 @@ const cleanUp = async () => {
...
+    .withCopyFilesToContainer([initSqlCopy])
+    .withDatabase(dbName)
+    .withLabels({workerId: workerId})
+    .withPassword('mirror_node_pass')
+    .withReuse()
+    .withUsername('mirror_node')
+    .start();
+  logger.info(`Started PostgreSQL container with image ${image}`);
+
+  fs.appendFileSync(containerTempFile, dockerDb.getId() + '\n');
```
xin-hedera, line 99: potentially multiple jest workers (OS processes) writing to the same file at the same time?

should change to just touch a file with the container id as its filename using `fs.closeSync(fs.openSync(filepath, 'w')`

edwin-greene: Excellent thanks I have made that update.

hedera-mirror-rest/__tests__/integrationDbOps.js
edwin-greene: Got it, fixed.

xin-hedera (dismissed):
LGTM

hedera-mirror-rest/__tests__/globalTeardown.js
```diff
@@ -15,15 +15,19 @@
...
-  const tmpDir = process.env.MIGRATION_TMP_DIR;
-  if (tmpDir) {
-    fs.rmSync(tmpDir, {force: true, recursive: true});
-    console.log(`Removed temp directory ${tmpDir}`);
+  const containerIds = fs.readdirSync(process.env.CONTAINER_TEMP_DIR);
+  const client = await getContainerRuntimeClient();
+  for (const id of containerIds) {
+    const container = client.container.getById(id);
+    await client.container.stop(container);
+    await client.container.remove(container);
```
steven-sheehy, line 27: Instead of creating a temp dir, writing the container IDs there, then reading those back to remove we could reuse the existing state in docker. Can just generate a global string like `mirror-rest: ${parentProcessId}` and attach it as a label to each container when starting. `parentProcessId` can be the actual parent process ID or some random integer generated by the parent. There doesn't seem to be a way to list by label, but we can do multiple `fetchByLabel()` until it deletes them all.

```suggestion
  const client = await getContainerRuntimeClient();
  while (true) {
    const container = client.container.fetchByLabel("mirror-rest", parentProcessId);
    if (!container) {
      break;
    }
    await client.container.stop(container);
    await client.container.remove(container);
```

hedera-mirror-rest/__tests__/integrationDbOps.js
```diff
@@ -74,36 +75,51 @@ const cleanUp = async () => {
...
+  const initSqlPath = path.join('..', 'hedera-mirror-common', 'src', 'test', 'resources', 'init.sql');
+  const initSqlCopy = {
+    source: initSqlPath,
+    target: '/docker-entrypoint-initdb.d/init.sql',
+  };
+
+  const dockerDb = await new PostgreSqlContainer(image)
+    .withCopyFilesToContainer([initSqlCopy])
+    .withDatabase(dbName)
+    .withLabels({workerId: workerId})
```
steven-sheehy, line 89: This label seems unused?

steven-sheehy (changes requested):

hedera-mirror-rest/__tests__/integrationDbOps.js
edwin-greene: The label seems to fix a problem with `withReuse()`, without the label the Jest workers seem to randomly reuse a container that is already in use by another Jest worker running another integration test.

steven-sheehy: Maybe add a comment to that effect

xin-hedera: testcontainer `withReuse()` says it'll reuse a container with same config when available. Apparently, it's done in a way by querying e.g. docker runtime to determine config equivalence. the labels are part of the config, so without adding the unique worker id as a label, for example, worker 2 will grab worker 1's container if it's already there

hedera-mirror-rest/__tests__/globalTeardown.js
edwin-greene: Sounds good, I have implemented this.

hedera-mirror-rest/__tests__/integrationDbOps.js
edwin-greene: Got it, I have added comments.

steven-sheehy (approved):
LGTM

xin-hedera (approved):
LGTM

PR #7837, opened 2024-03-01
Add an NFT allowance controller

### LINKED ISSUE #7760: HIP-857 Add controller layer to the NFT allowance API in the rest-java module
Implement the controller layer and establish an end-to-end static flow for the NFT allowance API.

mgoelswirlds:
**Description**:

This PR contains the controller layer for the NFT allowance REST API in the rest-java module.

* Add a Controller for the NFT allowance REST API
* Add mapstruct config for DTO to model mapping

**Related issue(s)**:

Fixes #7760

**Notes for reviewer**:

**Checklist**


steven-sheehy:
I get these failures when ran locally. Seems like it's a flaky test and might not be directly related to this PR.

```
NftAllowanceServiceTest > getNftAllowancesForOrderDesc(boolean) > [1] owner=true FAILED
    org.opentest4j.AssertionFailedError at NftAllowanceServiceTest.java:95

NftAllowanceServiceTest > getNftAllowancesForOrderDesc(boolean) > [2] owner=false FAILED
    org.opentest4j.AssertionFailedError at NftAllowanceServiceTest.java:95
```

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/RestUtils.java
```diff
@@ -0,0 +1,51 @@
...
+        return new Filter(paramName, RangeOperator.valueOf(splitVal[0].toUpperCase()), splitVal[1]);
+    }
+
+    public static Long getAccountNum(Filter filter) {
+        return filter != null ? parseIdFromString(filter.getValue())[2] : null;
+    }
+
+    public static Integer validateLimit(Integer limit) {
+        return limit > MAX_LIMIT ? MAX_LIMIT : limit;
+    }
```
steven-sheehy, line 50: Remove in favor of annotation validation

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/controller/AccountController.java
```diff
@@ -0,0 +1,88 @@
...
+            @RequestParam("account.id") @Valid Optional<String> accountIdQueryParam,
+            @RequestParam("owner") @Valid Optional<Boolean> owner,
+            @RequestParam("token.id") @Valid Optional<String> tokenId,
+            @RequestParam("limit") @Valid Optional<Integer> limit,
+            @RequestParam("order") @Valid Optional<String> order) {
+
+        
+        Filter filter = getFilter(FilterKey.ACCOUNT_ID, accountIdQueryParam.orElse(null));
+        Filter tokenIdFilter = getFilter(FilterKey.TOKEN_ID, tokenId.orElse(null));
+        Long accountPathParam = parseIdFromString(accountId)[2];
```
steven-sheehy, line 57: These `Filter` and `FilterKey` classes seem to serve no real purpose. They look to be intended for some generic filter use but then we turn around and change them into a non-generic `NftAllowanceRequest`. We should remove these classes in favor of proper validation and type conversion by Spring.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/controller/AccountController.java
```diff
@@ -0,0 +1,88 @@
...
+import com.hedera.mirror.restjava.mapper.NftAllowanceMapper;
+import com.hedera.mirror.restjava.service.NftAllowanceRequest;
+import com.hedera.mirror.restjava.service.NftAllowanceRequest.NftAllowanceRequestBuilder;
+import com.hedera.mirror.restjava.service.NftAllowanceService;
+import jakarta.validation.Valid;
+import java.util.Optional;
+import lombok.CustomLog;
+import lombok.RequiredArgsConstructor;
+import org.springframework.data.domain.Sort;
+import org.springframework.web.bind.annotation.*;
```
steven-sheehy, line 34: We should not use wildcard imports in mirror node

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/controller/AccountController.java
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        if (tokenIdFilter != null) {
+            requestBuilder.tokenId(getAccountNum(tokenIdFilter)).tokenIdOperator(tokenIdFilter.getOperator());
+        }
+
+        var serviceResponse = service.getNftAllowances(requestBuilder.build());
+
+        NftAllowancesResponse response = new NftAllowancesResponse();
+        response.setAllowances(accountMapper.map(serviceResponse));
```
steven-sheehy, line 85: There's no next link populated

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/controller/AccountController.java
```diff
@@ -0,0 +1,88 @@
...
+        if (isOwner != null) {
+            requestBuilder =
+                    isOwner ? requestBuilder.ownerId(accountPathParam) : requestBuilder.spenderId(accountPathParam);
+        }
+
+        if (tokenIdFilter != null) {
+            requestBuilder.tokenId(getAccountNum(tokenIdFilter)).tokenIdOperator(tokenIdFilter.getOperator());
+        }
+
+        var serviceResponse = service.getNftAllowances(requestBuilder.build());
```
steven-sheehy, line 82: There's zero `@ExceptionHandler` or error test cases.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/controller/AccountControllerTest.java
```diff
@@ -0,0 +1,86 @@
...
+                    ownerId, Optional.empty(), Optional.of(owner), Optional.empty(), Optional.of(2), Optional.empty());
+        } else {
+            response = accountController.getNftAllowancesByAccountId(
+                    spenderId,
+                    Optional.empty(),
+                    Optional.of(owner),
+                    Optional.empty(),
+                    Optional.of(2),
+                    Optional.empty());
+        }
```
steven-sheehy, line 60: There's no point in using parameterized tests if the parameterization just causes conditional logic in the test. Better to separate into different tests in that case.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/controller/AccountControllerTest.java
```diff
@@ -0,0 +1,86 @@
...
+    @ValueSource(booleans = {true, false})
+    void testWithNoQueryParams(boolean owner) {
+        var allowance = domainBuilder.nftAllowance().persist();
+        var ownerId = String.valueOf(allowance.getOwner());
+        var spenderId = String.valueOf(allowance.getSpender());
+        NftAllowancesResponse response;
+
+        if (owner) {
+            response = accountController.getNftAllowancesByAccountId(
+                    ownerId, Optional.empty(), Optional.of(owner), Optional.empty(), Optional.of(2), Optional.empty());
```
steven-sheehy, line 51: This test class is testing at the wrong level. We want the controller to use Spring's `RestClient` to call the HTTP endpoint to ensure all levels of the http stack are verified. Don't extend `RestJavaIntegrationTest` and use `@WebMvcTest` instead

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/controller/AccountControllerTest.java
```diff
@@ -0,0 +1,86 @@
...
+import lombok.RequiredArgsConstructor;
+import lombok.SneakyThrows;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+import org.springframework.test.context.junit.jupiter.SpringExtension;
+
+@ExtendWith(SpringExtension.class)
+@RequiredArgsConstructor
+class AccountControllerTest extends RestJavaIntegrationTest {
```
steven-sheehy, line 35: This test class is testing at the wrong level. We want the controller to use Spring's `RestClient` to call the HTTP endpoint to ensure all levels of the http stack are verified. Don't extend `RestJavaIntegrationTest` and use `@WebMvcTest` instead.

```suggestion
@ExtendWith(SpringExtension.class)
@RequiredArgsConstructor
@WebMvcTest(controllers = AccountController.class)
class AccountControllerTest {
```

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/mapper/NftAllowanceMapper.java
```diff
@@ -0,0 +1,43 @@
...
+
+import com.hedera.mirror.common.domain.entity.NftAllowance;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import org.mapstruct.Mapper;
+
+@Mapper(config = EntityMapper.class)
+public interface NftAllowanceMapper {
```
steven-sheehy, line 27: Needs a test that verifies all fields

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/mapper/NftAllowanceMapper.java
```diff
@@ -0,0 +1,43 @@
+/*
+ * Copyright (C) 2019-2024 Hedera Hashgraph, LLC
```
steven-sheehy, line 2: Please check all new files.

```suggestion
 * Copyright (C) 2024 Hedera Hashgraph, LLC
```

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/mapper/CommonMapper.java
```diff
@@ -0,0 +1,64 @@
...
+
+        var eid = com.hedera.mirror.common.domain.entity.EntityId.of(source);
+        return mapEntityId(eid);
+    }
+
+    default String mapEntityId(com.hedera.mirror.common.domain.entity.EntityId source) {
+        if (source == null) {
+            return null;
+        }
+        return source.getShard() + "." + source.getRealm() + "." + source.getNum();
```
steven-sheehy, line 41:
```suggestion
        return source != null ? source.toString() : null;
```

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/mapper/CommonMapper.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    default Instant mapInstant(Long source) {
```
steven-sheehy, line 44: Where is `Instant` used? The conversion should be from `Long` to `String` where `String` is `seconds.nanos`. `Instant.toString()` would not produce `seconds.nanos`.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/controller/AccountControllerTest.java
```diff
@@ -0,0 +1,86 @@
...
+
+    private final AccountController accountController;
+    private final NftAllowanceMapper mapper;
+
+    @SneakyThrows
+    @ParameterizedTest
+    @ValueSource(booleans = {true, false})
+    void testWithNoQueryParams(boolean owner) {
```
steven-sheehy, line 43: There's a huge quantity of test cases missing. We need negative and positive tests for every query parameter. We need negative tests for non-2xx scenarios.

steven-sheehy (changes requested):

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/controller/AccountController.java
```diff
@@ -0,0 +1,88 @@
...
+
+    private final NftAllowanceService service;
+    private final NftAllowanceMapper accountMapper;
+
+    @GetMapping(value = "/{accountId}/allowances/nfts")
+    NftAllowancesResponse getNftAllowancesByAccountId(
+            @PathVariable("accountId") @Valid String accountId,
+            @RequestParam("account.id") @Valid Optional<String> accountIdQueryParam,
+            @RequestParam("owner") @Valid Optional<Boolean> owner,
+            @RequestParam("token.id") @Valid Optional<String> tokenId,
```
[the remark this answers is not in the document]
mgoelswirlds: Changed this to EntityIdRangeParameter

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/controller/AccountControllerTest.java
mgoelswirlds: Using RestClient with an actual @SpringBootTest which brings up a server.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/mapper/CommonMapperTest.java
```diff
@@ -38,7 +37,7 @@ void mapEntityId() {
 
     @Test
     void mapEntityIdLong() {
-        assertNull(commonMapper.mapEntityId((Long) null));
+        // assertNull(commonMapper.mapEntityId((Long) null));
```
mgoelswirlds, line 40: Commenting out this flaky test.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/controller/AllowancesController.java
```diff
@@ -0,0 +1,171 @@
...
+    private final NftAllowanceMapper accountMapper;
+
+    @GetMapping(value = "/nfts")
+    NftAllowancesResponse getNftAllowancesByAccountId(
+            @PathVariable EntityIdRangeParameter id,
+            @RequestParam(name = ACCOUNT_ID, required = false) EntityIdRangeParameter accountId,
+            @RequestParam(defaultValue = "true") boolean owner,
+            @RequestParam(name = TOKEN_ID, required = false) EntityIdRangeParameter tokenId,
+            @RequestParam(defaultValue = "" + DEFAULT_LIMIT) @Min(MIN_LIMIT) @Max(MAX_LIMIT) int limit,
+            @RequestParam(defaultValue = "asc") String order) {
```
mgoelswirlds, line 76: Had to change this to String, since the order case with Sort.Direction was causing an  issue with parameter conversion. SO manually did that in the method.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/EntityIdRangeParameter.java
```diff
@@ -0,0 +1,58 @@
...
+            return EMPTY;
+        }
+
+        String[] splitVal = entityIdRangeParam.split(":");
+
+        if (splitVal.length == 1) {
+            // No operator specified. Just use "eq:"
+            return validateId(splitVal, 0, RangeOperator.EQ);
+        }
+        return validateId(splitVal, 1, RangeOperator.valueOf(splitVal[0].toUpperCase()));
```
steven-sheehy, line 40: `RangeOperator.valueOf()` will return an error like `No enum constant RangeOperator.BLAH`. This leaks class details. It also throws an IllegalArgumentException which currently doesn't have an `@ExceptionHandler`. It would be better if the error said `Invalid range operator blah. Valid values: lt, gt...`. We should encapsulate this in a `RangeOperator.of(String)` that also hides the upper casing.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/controller/AllowancesController.java
```diff
@@ -0,0 +1,171 @@
...
+        try {
+            orderDirection = Sort.Direction.fromString(order);
+        } catch (IllegalArgumentException e) {
+            throw new InvalidParametersException("order parameter must have a valid direction");
+        }
+        NftAllowanceRequestBuilder requestBuilder =
+                NftAllowanceRequest.builder().limit(limit).order(orderDirection).isOwner(owner);
+
+        if (accountId == null && tokenId != null) {
+            throw new InvalidParametersException("token.id parameter must have account.id present");
```
steven-sheehy, line 89: Where do we reject NE range operator?

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/Utils.java
```diff
@@ -0,0 +1,192 @@
...
+            return false;
+        }
+        if (isAMatch(ENTITY_ID_REGEX, id) || isAMatch(ENCODED_ENTITY_ID_REGEX, id)) {
+            return true;
+        }
+        return isValidEvmAddress(id, Constants.EvmAddressType.ANY);
+    }
+
+    private static boolean isAMatch(String regex, String id) {
+        Pattern partitionedPattern = Pattern.compile(regex);
```
steven-sheehy, line 62: Should statically compile Pattern

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/Utils.java
```diff
@@ -0,0 +1,192 @@
...
+
+    public static final Map<String, RangeOperator> OPERATOR_PATTERNS = Map.of(
+            "asc", RangeOperator.GTE,
+            "desc", RangeOperator.LTE);
+
+    public static boolean isValidEntityIdPattern(String id) {
+        if (StringUtils.isBlank(id)) {
+            return false;
+        }
+        if (isAMatch(ENTITY_ID_REGEX, id) || isAMatch(ENCODED_ENTITY_ID_REGEX, id)) {
```
steven-sheehy, line 55: We do not support encoded entity id as input to any API. We support a non-encoded `num` without realm or shard.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/Utils.java
```diff
@@ -0,0 +1,192 @@
...
+
+    private static boolean isValidEvmAddress(String address, Constants.EvmAddressType evmAddressType) {
+
+        if (evmAddressType == Constants.EvmAddressType.ANY) {
+            return isAMatch(EVM_ADDRESS_REGEX, address) || isAMatch(EVM_ADDRESS_SHARD_REALM_REGEX, address);
+        }
+        if (evmAddressType == Constants.EvmAddressType.NO_SHARD_REALM) {
+            return isAMatch(EVM_ADDRESS_REGEX, address);
+        }
+        return isAMatch(EVM_ADDRESS_SHARD_REALM_REGEX, address);
```
steven-sheehy, line 75: We should not have so many regex invocations. There should be a single pre-compiled regex for path id then we check which group is populated to determine evm address vs alias vs id.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/Utils.java
```diff
@@ -0,0 +1,192 @@
...
+
+    public static long[] parseFromEvmAddress(String evmAddress) {
+        // extract shard from index 0->8, realm from 8->23, num from 24->40 and parse from hex to decimal
+        var hexDigits = evmAddress.replace("0x", "");
+        return new long[] {
+            Long.parseLong(hexDigits.substring(0, 8), 16), // shard
+            Long.parseLong(hexDigits.substring(8, 24), 16), // realm
+            Long.parseLong(hexDigits.substring(24, 40), 16)
+        };
+    }
```
steven-sheehy, line 86: It's incorrect to parse a 40 byte evm address into a long array.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/Utils.java
```diff
@@ -0,0 +1,192 @@
...
+            var shardRealmNum = parseFromEvmAddress(numOrEvmAddress);
+            var shard = shardRealmNum[0];
+            var realm = shardRealmNum[1];
+            var num = shardRealmNum[2];
+            if (shard > MAX_SHARD || realm > MAX_REALM || num > MAX_NUM) {
+                // non-parsable evm address. get id num from evm address here
+            } else {
+                if (parts.size() == 3
+                        && ((Long.parseLong(parts.getFirst()) != shard) || Long.parseLong(parts.get(1)) != realm)) {
+                    throw new InvalidParametersException(" Id '%s' cannot be parsed".formatted(id));
```
steven-sheehy, line 117: All entity id parsing should go away. We should only have a single regex to extract out shard, realm, num with first two having defaults if not present. Then call EntityId.of().

steven-sheehy (changes requested):

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/mapper/EntityMapper.java
```diff
@@ -0,0 +1,30 @@
...
+import com.hedera.mirror.rest.model.NftAllowance;
+import org.mapstruct.MapperConfig;
+import org.mapstruct.Mapping;
+
+@MapperConfig(mappingInheritanceStrategy = AUTO_INHERIT_FROM_CONFIG, uses = CommonMapper.class)
+public interface EntityMapper<T extends NftAllowance> {
+
+    @Mapping(source = "timestampRange", target = "timestamp")
+    T map(com.hedera.mirror.common.domain.entity.NftAllowance source);
+}
```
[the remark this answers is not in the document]
mgoelswirlds: renaming this to RestJavaMapperConfig to avoid confusion with mapstruct MapperConfig

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/controller/AllowancesController.java
```diff
@@ -0,0 +1,171 @@
...
+
+    @ResponseStatus(value = HttpStatus.NOT_FOUND, reason = "Not found")
+    @ExceptionHandler(NoSuchElementException.class)
+    void notFound() {
+        // Currently no scenario is throwing this error
+    }
+
+    @ExceptionHandler
+    @ResponseStatus(BAD_REQUEST)
+    private Error inputValidationError(final InvalidInputException e) {
```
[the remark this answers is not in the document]
mgoelswirlds: Since this is a GET call HttpMediaTypeNotSupportedException will not be thrown. Instead I think I should handle `HttpRequestMethodNotSupportedException`

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/Constants.java
```diff
@@ -0,0 +1,45 @@
...
+import lombok.experimental.UtilityClass;
+
+@UtilityClass
+public class Constants {
+
+    public static final int MAX_LIMIT = 100;
+    public static final String DEFAULT_LIMIT = "25";
+    public static final String ENTITY_ID_REGEX = "^(((\\d{1,5}\\.){1,2}\\d{1,10})|(\\d{1,19}))$";
+
+    public static final String EVM_ADDRESS_REGEX = "^(((0x)?[A-Fa-f0-9]{40})|((\\d{1,10}\\.){0,2}[A-Fa-f0-9]{40}))$";
```
mgoelswirlds, line 30: The regex needed to be split into 2 because of the increased complexity as reported by Sonar.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/controller/AllowancesController.java
```diff
@@ -0,0 +1,171 @@
...
+
+        if (accountId == null && tokenId != null) {
+            throw new InvalidParametersException("token.id parameter must have account.id present");
+        }
+
+        // Setting both owner and spender Id to the account.id query parameter value.
+        if (accountId != null && accountId.getValue() != null) {
+            requestBuilder
+                    .spenderId(accountId.value().getId())
+                    .ownerId(accountId.value().getId())
```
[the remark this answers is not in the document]
mgoelswirlds: This has been moved to the service layer, but we need to get long to send it to the repo.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/controller/AllowancesController.java
```diff
@@ -0,0 +1,171 @@
...
+            next = buildNextLink(owner, orderDirection, limit, response, serviceResponse);
+        }
+        response.links(new Links().next(next));
+        return response;
+    }
+
+    @ResponseStatus(value = HttpStatus.NOT_FOUND, reason = "Not found")
+    @ExceptionHandler(NoSuchElementException.class)
+    void notFound() {
+        // Currently no scenario is throwing this error
```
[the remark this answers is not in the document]
mgoelswirlds: Will be part of the evm address ticket.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/Utils.java
```diff
@@ -0,0 +1,136 @@
...
+            var next = getNextParamQueries(orderString, lastValues, included);
+            if (next.isBlank()) {
+                return null;
+            }
+            // remove the '/' at the end of req.path
+            var path = url.endsWith("/") ? url.substring(0, url.length() - 1) : url;
+            paginationLink.append(path);
+            paginationLink.append("?limit=").append(limit);
+            paginationLink.append("&order=").append(orderString);
+            paginationLink.append(next);
```
steven-sheehy, line 118: This is all quite brittle and is missing edge cases. For example, if `owner=false` was in the url it would not get propagated to the next link. Add a test to verify that has every possible query param and has a next link.

Instead, we should take the approach were we iterate over the `request.getParameterMap()` and construct the next link from it, swapping out values only for token.id and account.id if applicable.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/NftAllowanceServiceImpl.java
```diff
@@ -41,39 +44,55 @@ public class NftAllowanceServiceImpl implements NftAllowanceService {
...
+                    accountId.value().getId(), ownerId, tokenId, pageable);
+        }
+    }
+
+    private static long getId(EntityIdRangeParameter idParam) {
+        long id = 0;
+        if (idParam != null) {
+            id = getUpdatedEntityId(idParam);
+        }
+        return id;
```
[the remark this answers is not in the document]
mgoelswirlds: These are the optional query params.  As you mentioned this will need to change with dynamic queries but until then the static queries will return all values with ids > 0 .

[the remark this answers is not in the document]
steven-sheehy: Yes, but if I omit token.id the query will put `token_id = 0` and return zero results. Better to throw an error if optional param not sent but query currently requires it.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/EntityIdParameter.java
```diff
@@ -0,0 +1,34 @@
...
+public record EntityIdParameter(EntityId value) {
+    public static final EntityIdParameter EMPTY = new EntityIdParameter(null);
+
+    public static EntityIdParameter valueOf(String entityIdParam) {
+
+        if (entityIdParam == null || entityIdParam.isBlank()) {
+            return EMPTY;
+        }
+
+        return new EntityIdParameter(parseId(entityIdParam));
```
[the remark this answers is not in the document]
mgoelswirlds: Since we would need the parseId code in near future in a common space , I would prefer to leave it in the utils.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/controller/AllowancesController.java
```diff
@@ -0,0 +1,115 @@
...
+                        response.getAllowances().get(serviceResponse.size() - 1).getOwner());
+            }
+            included.put(ACCOUNT_ID, true);
+            lastValues.put(
+                    TOKEN_ID,
+                    response.getAllowances().get(serviceResponse.size() - 1).getTokenId());
+            included.put(TOKEN_ID, false);
+        }
+        return Utils.getPaginationLink(false, lastValues, included, order);
+    }
```
steven-sheehy, line 114: Can be greatly simplified, I believe. We might be able to get rid of included map if we make it a linked map and assume last item is not inclusive?

```suggestion
        var last = CollectionUtils.lastElement(response.getAllowances());
        String next = null;

        if (last != null && serviceResponse.size() == limit) {
            long accountId = owner ? last.getSpender() : last.getOwner();
            var included = Map.of(ACCOUNT_ID, true, TOKEN_ID, false);
            var lastValues = Map.of(ACCOUNT_ID, accountId, TOKEN_ID, last.getTokenId());
            next = Utils.getPaginationLink(false, lastValues, included, order)
        }
```

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/controller/AllowancesControllerIntegrationTest.java
```diff
@@ -0,0 +1,250 @@
...
+        var nextLink = "/api/v1/accounts/%s/allowances/nfts?limit=1&order=asc&account.id=gte:%s&token.id=gt:%s"
+                .formatted(
+                        allowance1.getOwner(),
+                        EntityId.of(allowance2.getSpender()),
+                        EntityId.of(allowance2.getTokenId()));
+
+        // Performing the GET operation
+        var result = restClient
+                .get()
+                .uri(uriParams, allowance1.getOwner(), EntityId.of(allowance1.getSpender()))
```
steven-sheehy, line 140: nit: Could pull `.get()` up to setup.

```suggestion
        var result = restClient.uri(uriParams, allowance1.getOwner(), EntityId.of(allowance1.getSpender()))
```

steven-sheehy (changes requested):

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/Utils.java
```diff
@@ -0,0 +1,153 @@
...
+        var path = url.endsWith("/") ? url.substring(0, url.length() - 1) : url;
+        paginationLink.append(path);
+        if (!paramsMap.isEmpty() || !lastValues.isEmpty()) {
+            paginationLink.append("?");
+        }
+        // add all params from the existing url
+        int count = 0; // This is to check for first parameter
+        for (Map.Entry<String, String[]> param : paramsMap.entrySet()) {
+            if (!lastValues.containsKey(param.getKey())) {
+                if (count != 0) {
```
mgoelswirlds, line 125: Maintaining the request parameter order here is not very straightforward since the paramMap  is not a LinkedHashMap. I will fix this as part of a follow-up ticket.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/Utils.java
```diff
@@ -0,0 +1,151 @@
...
+        long shard = 0;
+        long realm = 0;
+        String numOrEvmAddress;
+        if (matcher.matches()) {
+
+            if (matcher.group(4) != null) {
+                realm = Long.parseLong(matcher.group(5));
+                if (matcher.group(2) == null) {
+                    // get the system shard value from properties
+                    shard = 0;
```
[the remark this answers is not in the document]
mgoelswirlds: https://github.com/hashgraph/hedera-mirror-node/issues/7858

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/controller/AllowancesControllerIntegrationTest.java
mgoelswirlds: Don't believe this can be done. The client needs a method.

hedera-mirror-rest-java/build.gradle.kts
```diff
@@ -42,3 +45,17 @@ dependencies {
...
+            "-Amapstruct.defaultComponentModel=jakarta",
+            "-Amapstruct.defaultInjectionStrategy=constructor",
+            "-Amapstruct.disableBuilders=true"
+        )
+    )
+}
+
+java.sourceSets["main"].java {
+    srcDir("build/generate-resources/main/com/hedera/mirror/rest/model")
+}
```
xin-hedera, line 61: is this needed? the directory is already added to main sources set in openapi-convetions

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/mapper/CommonMapper.java
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+        if (source.hasUpperBound()) {
+            target.setTo(mapTimestamp(source.upperEndpoint()));
+        }
+
+        return target;
+    }
+
+    default String mapTimestamp(long timestamp) {
```
xin-hedera, line 60:
```suggestion
    default String mapTimestamp(Long timestamp) {
```

and return `null` if `timestamp` is null, then simplify `mapRange` to remove the `if (source.has???Bound())` check

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/NftAllowanceServiceImpl.java
```diff
@@ -41,39 +42,59 @@ public class NftAllowanceServiceImpl implements NftAllowanceService {
...
             var pageable =
                     PageRequest.of(0, limit, order.isAscending() ? SPENDER_TOKEN_ASC_ORDER : SPENDER_TOKEN_DESC_ORDER);
-            return repository.findByOwnerAndFilterBySpenderAndToken(ownerId, spenderId, tokenId, pageable);
+            return repository.findByOwnerAndFilterBySpenderAndToken(
+                    accountId.value().getId(), spenderId, tokenId, pageable);
 
         } else {
-
-            ownerId = getUpdatedEntityId(accountIdOperator, ownerId);
+            var ownerId = verifyAndGetRangeId(ownerOrSpenderId);
```
xin-hedera, line 65: nit, should call `verifyAndGetRangeId(ownerOrSpenderId)` just once

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/OrderConverter.java
```diff
@@ -0,0 +1,31 @@
...
+import org.springframework.boot.context.properties.ConfigurationPropertiesBinding;
+import org.springframework.core.convert.converter.Converter;
+import org.springframework.data.domain.Sort;
+
+@Named
+@ConfigurationPropertiesBinding
+public class OrderConverter implements Converter<String, Sort.Direction> {
+    @Override
+    public Sort.Direction convert(String order) {
+        return Sort.Direction.valueOf(order.toUpperCase());
```
xin-hedera, line 29: the static method `fromString` is case-insensitive. Also do we really need a custom converter? OK to keep the converter though.

```suggestion
        return Sort.Direction.fromString(order);
```

hedera-mirror-rest-java/build.gradle.kts
mgoelswirlds: Just verified. It is not needed anymore.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/mapper/CommonMapper.java
mgoelswirlds: Can't do this because of method ambiguity while mapping using mapstruct.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/Utils.java
```diff
@@ -0,0 +1,178 @@
...
+        }
+        HttpServletRequest request = servletRequestAttributes.getRequest();
+        StringBuilder paginationLink = new StringBuilder();
+
+        if (lastValues == null || lastValues.isEmpty()) {
+            return null;
+        }
+
+        if (!isEnd) {
+            paginationLink.append(generateLinkFromRequestParams(lastValues, order, request));
```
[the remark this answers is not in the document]
mgoelswirlds: Agreed missed this in refactoring.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/Utils.java
```diff
@@ -0,0 +1,178 @@
...
+
+        // remove the '/' at the end of req.path
+        var path = url.endsWith("/") ? url.substring(0, url.length() - 1) : url;
+        link.append(path);
+
+        if (!paramsMap.isEmpty() || !lastValues.isEmpty()) {
+            link.append("?");
+        }
+
+        // add all params from the existing url
```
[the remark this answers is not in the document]
mgoelswirlds: Will create a followup.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/mapper/CommonMapper.java
```diff
@@ -0,0 +1,75 @@
...
+        if (timestamp == 0) {
+            return "0.0";
+        }
+        var timestampString = String.valueOf(timestamp);
+        StringBuilder sb = new StringBuilder();
+        // appending 0's to the timestamp to fulfill nanoseconds requirement
+        while (sb.length() < (NANO_DIGITS - timestampString.length() - 1)) {
+            sb.append('0');
+        }
+        sb.append(timestampString);
```
xin-hedera, line 70: use `StringUtils.leftPad(value, 10, '0')`

xin-hedera (changes requested):

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/mapper/CommonMapperTest.java
```diff
@@ -0,0 +1,79 @@
...
+                .usingRecursiveComparison()
+                .isEqualTo(range);
+
+        // test 4
+        range.setFrom("1586567700.453054000");
+        range.setTo("1586567700.453054000");
+        assertThat(commonMapper.mapRange(Range.openClosed(1586567700453054000L, 1586567700453054000L)))
+                .usingRecursiveComparison()
+                .isEqualTo(range);
+    }
```
xin-hedera, line 78: still need to test corner cases, e.g., given `[1, 100)`, expect `["0.000000001", "0.000000100")`. the current implementation would certainly fail, because it does right padding instead of left padding.

xin-hedera (approved):
LGTM

steven-sheehy (approved):
LGTM

PR #7856, opened 2024-03-04
Update OpenAPI to specify Unix timestamp

### LINKED ISSUE #7754: Update timestamp description on Open API to be more clear on expected format
### Problem

When working with the swagger API documentation to get the state of a smart contract, I came across needing a timestamp in seconds.nanoseconds format. When going to hash scan, the timestamp was given in local time (which is valid), but nowhere could I find an easy way to convert it to what the mirror node was expecting. It took me a bit of asking around to figure out i could give a UNIX timestamp. 
image (3)

### Solution

Other devs may experience a similar thing as me and was hoping we could update the swagger api docs where there is a timestamp section to be a bit more clear on what the mirror node expects the timestamp in. Reading seconds.nanoseconds format is not very clear. In addition, adding a tip on how to get a timestamp in that format would be great.

### Alternatives

_No response_

edwin-greene:
**Description**:

Updates the OpenAPI specification to specify that timestamps are Unix timestamps

**Related issue(s)**:

Fixes #7754 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-rest/api/v1/openapi.yml
+++ b/hedera-mirror-rest/api/v1/openapi.yml
@@ -3226,10 +3226,12 @@ components:
       items:
         $ref: "#/components/schemas/StakingRewardTransfer"
     Timestamp:
+      description: A Unix timestamp in seconds.nanoseconds format
       type: string
       example: "1586567700.453054000"
       pattern: '^\d{1,10}(\.\d{1,9})?$'
     TimestampNullable:
+      description: A Unix timestamp in seconds.nanoseconds format
       type: string
       example: "1586567700.453054000"
       pattern: '^\d{1,10}(\.\d{1,9})?$'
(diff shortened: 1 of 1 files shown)


hedera-mirror-rest/api/v1/openapi.yml
```diff
@@ -4424,7 +4426,7 @@ components:
         type: string
         pattern: ^((eq|gt|gte|lt|lte):)?\d{1,19}?$
     timestampQueryParam:
-      description: The consensus timestamp in seconds.nanoseconds format with an optional comparison operator
+      description: The consensus timestamp as a Unix timestamp in seconds.nanoseconds format with an optional comparison operator
```
steven-sheehy, line 4429: Same comment for other similar areas.

```suggestion
      description: The consensus timestamp as a Unix timestamp in seconds.nanoseconds format with an optional comparison operator. See [unixtimestamp.com](https://www.unixtimestamp.com/) for a simple way to convert a date to the 'seconds' part of the Unix time.
```

edwin-greene: Got it, I have added the link

steven-sheehy (approved):
LGTM

PR #7870, opened 2024-03-05
Update Rest API tokens endpoint to accept multiple token ids

### LINKED ISSUE #7810: Make `/tokens` support queries for multiple tokens
### Problem

Our GC member would like be able to query multiple tokens via `/tokens`. Currently the endpoint only supports querying 1 token at a time. 

### Solution

Make `/tokens` supports multiple token ids similar to `/accounts` 

https://testnet.mirrornode.hedera.com/api/v1/accounts?&account.id=0.0.6347&account.id=3053753
https://testnet.mirrornode.hedera.com/api/v1/tokens?token.id=0.0.3608138&token.id=0.0.3608140

### Alternatives

_No response_

edwin-greene:
**Description**:
Update Rest API tokens endpoint to accept multiple token id query parameters.

**Related issue(s)**:

Fixes #7810 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-rest/__tests__/specs/tokens/multiple-ids.json
+++ b/hedera-mirror-rest/__tests__/specs/tokens/multiple-ids.json
@@ -1,5 +1,5 @@
 {
-  "description": "Token api call for all tokens within a given token id range",
+  "description": "Token api call with multiple token id query parameters",
   "setup": {
     "entities": [
       {
--- a/hedera-mirror-rest/__tests__/specs/tokens/multiple-ids.json
+++ b/hedera-mirror-rest/__tests__/specs/tokens/multiple-ids.json
@@ -52,29 +52,59 @@
       }
     ]
   },
-  "urls": ["/api/v1/tokens?token.id=0.0.1135&token.id=0.7.25301"],
-  "responseStatus": 200,
-  "responseJson": {
-    "tokens": [
-      {
-        "token_id": "0.0.1135",
-        "symbol": "ORIGINALRDKSE",
-        "admin_key": null,
-        "decimals": 2,
-        "name": "Original",
-        "type": "FUNGIBLE_COMMON"
-      },
-      {
-        "token_id": "0.7.25301",
-        "symbol": "MIRRORTOKEN",
-        "admin_key": null,
-        "decimals": 3,
-        "name": "Mirror",
-        "type": "FUNGIBLE_COMMON"
+  "tests": [
+    {
+      "urls": [
+        "/api/v1/tokens?token.id=0.0.1135&token.id=0.7.25301",
+        "/api/v1/tokens?token.id=0.0.1135&token.id=eq:0.7.25301",
+        "/api/v1/tokens?token.id=0.0.1135&token.id=eq:0.7.25301&token.id=gt:1134&token.id=lte:0.7.25301",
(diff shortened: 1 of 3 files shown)


jascks (dismissed):
Short and sweet. LGTM.

hedera-mirror-rest/tokens.js
```diff
@@ -172,9 +173,18 @@ const extractSqlFromTokenRequest = (query, params, filters, conditions) => {
       continue;
     }
 
+    if (columnKey === sqlQueryColumns.TOKEN_ID && filter.operator === utils.opsMap.eq) {
```
steven-sheehy, line 176: What happens if there's a mix of eq and range operators?

hedera-mirror-rest/__tests__/specs/tokens/multiple-ids.json
```diff
@@ -0,0 +1,80 @@
...
+      },
+      {
+        "token_id": "0.23.45678",
+        "symbol": "HEDERACOIN",
+        "created_timestamp": "1234567890000000004",
+        "type": "FUNGIBLE_COMMON"
+      }
+    ]
+  },
+  "urls": ["/api/v1/tokens?token.id=0.0.1135&token.id=0.7.25301"],
```
steven-sheehy, line 55: Show one with eq.

```suggestion
  "urls": ["/api/v1/tokens?token.id=0.0.1135&token.id=eq:0.7.25301"],
```

hedera-mirror-rest/tokens.js
```diff
@@ -172,9 +173,18 @@ const extractSqlFromTokenRequest = (query, params, filters, conditions) => {
...
+      inClause.push(`$${params.push(filter.value)}`);
+      continue;
+    }
+
     conditions.push(`${filterColumnMap[filter.key]}${filter.operator}$${params.push(filter.value)}`);
   }
 
+  if (inClause.length > 0) {
+    conditions.push(`${sqlQueryColumns.TOKEN_ID} in (${inClause})`);
```
xin-hedera, line 185: We use `in ($1, $2, $3, ... $n)` in many places, however it's better to change to `= any($1)`, where the value for `$1` is just the array of all values.

hedera-mirror-rest/tokens.js
edwin-greene: I added this type of scenario to the spec test and to the tokens.test.js. The behavior at the moment is that the conditions are all `anded` together in the sql statement. 

This looks to be the same behavior as the accounts endpoint, for example `/api/v1/accounts?account.id=0.0.7&account.id=0.0.9&account.id=lt:0.0.9` only returns account 0.0.7.

hedera-mirror-rest/__tests__/specs/tokens/multiple-ids.json
edwin-greene: Thanks I have added that

hedera-mirror-rest/tokens.js
edwin-greene: Sounds good I have converted it to use `= any($`

xin-hedera (approved):
LGTM

jascks (approved):
LGTM

steven-sheehy (approved):
LGTM

mgoelswirlds (approved):

Issue #7690, opened 2024-02-06
HIP-857 Add dynamic queries to the Nft Allowance API in the rest-java module


mgoelswirlds:
Add dynamic queries to the rest-java module for the allowances/nfts query.

PR #7871, opened 2024-03-05
Integrate Jooq into rest-java to support dynamic SQL queries

xin-hedera:
**Description**:

- Add `NftAllowanceRepositoryCustom` to dynamically generate SQL and fetch results
- Add a custom generic `DomainRecordMapper` to map jooq `Record` objects to domain objects 
- Add gradle build tasks and flows to make jooq codegen to generate meta classes from tables in a live database
- Customize jooq configuration to not prefix everything with schema
- Workaround dependency issue introduced by gradle docker plugin
- Fix hibernate custom json formatter lost in some modules' integration tests

**Related issue(s)**:

Relates to #7690 

**Notes for reviewer**:

**Checklist**


hedera-mirror-rest-java/build.gradle.kts
```diff
@@ -42,3 +48,125 @@ dependencies {
...
+                }
+            }
+        }
+    }
+}
+
+java.sourceSets["main"].java { srcDir(jooqTargetDir) }
+
+/** Build service for providing database container. */
+abstract class PostgresService : BuildService<PostgresService.Params>, AutoCloseable {
```
xin-hedera, line 141: the service make it easy to share the postgresql container, also stops the container at the end

hedera-mirror-rest-java/build.gradle.kts
```diff
@@ -42,3 +48,125 @@ dependencies {
...
+    }
+
+    private var container: PostgreSQLContainer<Nothing>
+
+    init {
+        val initScript =
+            Transferable.of(
+                File(
+                        parameters.getRootDir().get() +
+                            "/hedera-mirror-importer/src/main/resources/db/scripts/init.sh"
```
xin-hedera, line 153: this has to be absolute path otherwise running the task from a different location will fail due to file not found when trying to read it

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/config/JooqCustomConfiguration.java
```diff
@@ -0,0 +1,30 @@
...
+import org.springframework.boot.autoconfigure.jooq.DefaultConfigurationCustomizer;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+class JooqCustomConfiguration {
+
+    @Bean
+    public DefaultConfigurationCustomizer configurationCustomizer() {
+        return c -> c.settings().withRenderSchema(false);
```
xin-hedera, line 28: all our tables are in the same schema, so let jooq know we don't want it to add the `public.` prefix everywhere

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/repository/JooqRepositoryTest.java
```diff
@@ -0,0 +1,73 @@
...
+
+        // then
+        assertThat(actual).containsExactly(expected1, expected2);
+
+        // when
+        query = dslContext
+                .selectFrom(NFT_ALLOWANCE)
+                .where(NFT_ALLOWANCE.OWNER.eq(expected1.getOwner()))
+                .getQuery();
+        var actualNftAllowance = jooqRepository.getEntity(query, NftAllowance.class);
```
xin-hedera, line 57: this will create a parameterized SQL query, e.g., with `where nft_allowance.ower = ?`

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/config/JooqCustomConfiguration.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+package com.hedera.mirror.restjava.config;
+
+import org.springframework.boot.autoconfigure.jooq.DefaultConfigurationCustomizer;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+class JooqCustomConfiguration {
```
xin-hedera, line 24: can't name the class `JooqConfiguration` since then the bean will name clash with the one brought in by `spring-boot-starter-jooq`

edwin-greene (approved):
Looks good to me

hedera-mirror-rest-java/build.gradle.kts
```diff
@@ -14,10 +14,15 @@
...
 
+import org.testcontainers.containers.PostgreSQLContainer
+import org.testcontainers.images.builder.Transferable
+
 description = "Hedera Mirror Node Rest Java"
 
 plugins {
     id("openapi-conventions")
+    id("org.jooq.jooq-codegen-gradle") version "3.19.5"
+    id("org.flywaydb.flyway") version "10.8.1"
```
jascks, line 25: It seems `graphql`, `grpc` and `web3` are getting by well enough using Spring Data repositories? Would there be a benefit to dropping this stuff into `jooq-conventions.gradle.kts` so it can be more easily leveraged elsewhere? Perhaps premature optimization at this time.

buildSrc/build.gradle.kts
```diff
@@ -22,6 +22,13 @@ repositories {
 }
 
 dependencies {
+    val dockerJavaVersion = "3.3.5"
+
+    // Add docker-java dependencies before gradle-docker-plugin to avoid the docker-java jars
+    // embedded in the plugin being used by testcontainers-postgresql
```
steven-sheehy, line 28: The order of specifying dependencies is not important, so this can be specified in sorted order. However, these are not direct dependencies but overrides of transitive dependencies. Move these to a [constraints](https://docs.gradle.org/current/userguide/dependency_constraints.html#sec:adding-constraints-transitive-deps) block and add a `because` description.

hedera-mirror-rest-java/build.gradle.kts
```diff
@@ -42,3 +48,125 @@ dependencies {
...
+}
+
+jooq {
+    configuration {
+        generator {
+            database {
+                excludes =
+                    """
+                    account_balance_old
+                    | account_balance_p.*
```
steven-sheehy, line 73: Instead of explicitly listing partitioned tables we could use a more generic regex like `.*_p\d+_\d+$`

hedera-mirror-rest-java/build.gradle.kts
```diff
@@ -42,3 +48,125 @@ dependencies {
...
+                """
+                includes = ".*"
+                inputSchema = dbSchema
+                isIncludeRoutines = false
+                isIncludeUDTs = false
+                name = "org.jooq.meta.postgres.PostgresDatabase"
+            }
+            target {
+                directory = jooqTargetDir
+                packageName = "com.hedera.mirror.restjava.model.jooq"
```
steven-sheehy, line 86: It's probably strange to have the jooq models in a sub directory of the view-model. Maybe move view model to `model/view` and keep jook in `model/jooq` OR to `viewmodel` and make it just `jooq`.

hedera-mirror-rest-java/build.gradle.kts
```diff
@@ -42,3 +48,125 @@ dependencies {
...
+            "schema" to dbSchema,
+            "tempSchema" to "temporary",
+            "topicRunningHashV2AddedTimestamp" to "0"
+        )
+    user = dbUser
+
+    usesService(dbContainerProvider)
+
+    doFirst { url = getJdbcUrl(dbContainerProvider.get().getContainer()) }
+}
```
steven-sheehy, line 118: Since kotlin can read Java, wondering if there's some way to directly use `CommonTestConfiguration.flywayProperties()` here so we don't repeat ourselves.

hedera-mirror-rest-java/build.gradle.kts
```diff
@@ -42,3 +48,125 @@ dependencies {
...
+                        parameters.getRootDir().get() +
+                            "/hedera-mirror-importer/src/main/resources/db/scripts/init.sh"
+                    )
+                    .readBytes()
+            )
+        container =
+            PostgreSQLContainer<Nothing>("postgres:14-alpine").apply {
+                withCopyToContainer(initScript, "/docker-entrypoint-initdb.d/init.sh")
+                withUsername("postgres")
+                start()
```
steven-sheehy, line 161: Again, this is literally the same java code as `CommonTestConfiguration.postgresql()` so we should find a way to reuse.

steven-sheehy (changes requested):

hedera-mirror-rest-java/build.gradle.kts
xin-hedera: agree. I tend to not make it a convention because at least now the flyway plugin and jooq plugin aren't used by other modules.

If we do see the needs of utilizing jooq in other modules, we still don't have to make it a convention, because the module to host jooq generated meta classes will then be the common module and we only need to have these tasks and configs in common instead.

hedera-mirror-rest-java/build.gradle.kts
xin-hedera: sure

hedera-mirror-rest-java/build.gradle.kts
xin-hedera: I don't think that's possible. We'll create a circular dependency most likely, as then buildSrc will depend on common module, however common module in reverse depends on buildSrc

buildSrc/build.gradle.kts
```diff
@@ -22,6 +22,13 @@ repositories {
...
+    implementation("com.github.docker-java:docker-java-api:$dockerJavaVersion")
```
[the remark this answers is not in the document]
xin-hedera: doesn't work that way, I guess the reason is this is not a transitive dependency problem.

buildSrc/build.gradle.kts
xin-hedera: moving to constraints doesn't work.

The underlying problem is not docker gradle plugin bringing the jars by transitive dependency, instead, the plugin embeds the docker-java jars directly in its distribution without shading. Though I didn't check the jars on the build system classpath after overriding with the above two, it's the magic which makes both testcontainers and docker plugin work.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/historicalbalance/HistoricalBalanceService.java
```diff
@@ -199,16 +199,17 @@ private Optional<Long> getMaxConsensusTimestamp(long timestamp) {
...
-                .map(AccountBalanceFile::getConsensusTimestamp)
-                .or(() -> recordFileRepository
-                        .findFirst()
-                        .map(RecordFile::getConsensusEnd)
-                        .map(timestamp ->
-                                timestamp + properties.getInitialDelay().toNanos()))
-                .filter(lastTimestamp -> consensusEnd - lastTimestamp
-                        >= properties.getMinFrequency().toNanos())
-                .isPresent();
+        return properties.isEnabled()
```
xin-hedera, line 202: just spotless reformat

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/downloader/BatchStreamFileNotifierTest.java
```diff
@@ -188,9 +188,9 @@ void queueCapacityReached() {
         var sleep = Duration.ofMillis(500L);
 
         doAnswer(invocation -> {
-            Uninterruptibles.sleepUninterruptibly(sleep);
-            return null;
-        })
+                    Uninterruptibles.sleepUninterruptibly(sleep);
```
xin-hedera, line 191: just spotless reformat

buildSrc/src/main/kotlin/jooq-conventions.gradle.kts
```diff
@@ -0,0 +1,152 @@
...
+                """
+                includes = ".*"
+                inputSchema = dbSchema
+                isIncludeRoutines = false
+                isIncludeUDTs = false
+                name = "org.jooq.meta.postgres.PostgresDatabase"
+            }
+            target {
+                directory = jooqTargetDir
+                packageName = "com.hedera.mirror.restjava.jooq.domain"
```
xin-hedera, line 68: The package for the generated meta classes is changed

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/repository/NftAllowanceRepositoryTest.java
```diff
@@ -90,4 +113,174 @@ void findByOwnerAndFilterBySpenderGtAndTokenGtOwnerNotPresent() {
                         nftAllowance.getOwner() + 1, nftAllowance.getSpender(), nftAllowance.getTokenId(), pageable))
                 .isEmpty();
     }
+
+    @MethodSource("provideFindAllArguments")
```
xin-hedera, line 117: The complexity of building the test case scenarios is in the `provideFindAllArguments` function. There are 3 owners, 3 spenders,  and 3 tokens created in the setup, and thus 27 nft allowances in total.

The test cases utilize indexes of owners / spenders / tokens to create filters and expected result in a relatively readable way.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/repository/NftAllowanceRepositoryTest.java
```diff
@@ -90,4 +113,174 @@ void findByOwnerAndFilterBySpenderGtAndTokenGtOwnerNotPresent() {
...
+        var expectedNftAllowances = expected.stream().map(nftAllowances::get).toList();
+
+        // when, then
+        assertThat(nftAllowanceRepository.findAll(byOwner, filters, limit, order))
+                .containsExactlyElementsOf(expectedNftAllowances);
+    }
+
+    @Test
+    void findAllThrowInvalidFilterException() {
+        var emptyFilters = Collections.<Filter<?>>emptyList();
```
xin-hedera, line 133: solar complains about possible `RuntimeException` thrown by `Collections.emptyList()` if directly passed to `findAll`, so split it here.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/repository/NftAllowanceRepositoryCustomImpl.java
```diff
@@ -0,0 +1,154 @@
...
+        } else if (operator == RangeOperator.LT || operator == RangeOperator.LTE) {
+            value -= 1L;
+        }
+
+        var filter = new Filter<>(primarySortFilter.field(), operator, value, Long.class);
+        return primaryCondition.and(makeCondition(filter));
+    }
+
+    @SuppressWarnings({"java:S1905", "rawtypes", "unchecked"})
+    private static Condition makeCondition(Filter<?> filter) {
```
xin-hedera, line 140: this function deserves to be in a utility class

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/repository/NftAllowanceRepositoryCustomImpl.java
```diff
@@ -0,0 +1,153 @@
...
+        var baseCondition = getBaseCondition(primaryCondition, primarySortFilter, tokenFilter);
+        var secondaryCondition = getSecondaryCondition(primaryCondition, primarySortFilter, tokenFilter);
+        var condition = secondaryCondition != null ? baseCondition.or(secondaryCondition) : baseCondition;
+
+        return dslContext
+                .selectFrom(NFT_ALLOWANCE)
+                .where(condition)
+                .orderBy(SORT_ORDERS.get(new OrderSpec(byOwner, order)))
+                .limit(limit)
+                .fetchInto(NftAllowance.class);
```
xin-hedera, line 90: went back to use jooq's fluent API because after all the conditions complexity is not as bad as I thought.


In case the query needs to be more dynamic thus naturally more complicated (e.g., unions, CTEs, and etc), we should use [QueryParts](https://www.jooq.org/doc/3.19/manual/sql-building/queryparts/)

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/jooq/DomainRecordMapper.java
```diff
@@ -0,0 +1,184 @@
...
+    private final org.jooq.Field<?>[] recordFields;
+    private final Map<String, Setter> setters;
+
+    public DomainRecordMapper(RecordType<R> recordType, Class<? extends E> type) {
+        try {
+            defaultConstructor = LOOKUP.findConstructor(type, MethodType.methodType(void.class));
+            recordFields = recordType.fields();
+            var fieldNames =
+                    Arrays.stream(recordFields).map(org.jooq.Field::getName).collect(Collectors.toSet());
+            setters = getSetters(fieldNames, type);
```
steven-sheehy, line 69: The `EntityMetadataRegistry` already does much of this logic and I would like to see it reused here as well since it abstracts the setters behind lambdas. That said, it's late in the cycle so we can open a ticket to follow up.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/jooq/DomainRecordMapper.java
```diff
@@ -0,0 +1,184 @@
...
+    @Override
+    @SuppressWarnings("unchecked")
+    public @Nullable E map(R source) {
+        try {
+            var entity = (E) defaultConstructor.invoke();
+            for (int i = 0; i < recordFields.length; i++) {
+                var field = recordFields[i];
+                var setter = setters.get(field.getName());
+                if (setter == null) {
+                    continue;
```
steven-sheehy, line 85: Should this throw an error or are there valid scenarios where it can be missing?

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/jooq/DomainRecordMapper.java
```diff
@@ -0,0 +1,184 @@
...
+            } else {
+                // TokenFreezeStatus and TokenKycStatus are stored as smallint instead of pg enum in the database
+                return targetType.getEnumConstants()[((Number) source).shortValue()];
+            }
+        } else if (targetType == Range.class && sourceType == LongRange.class) {
+            return LongRangeConverter.INSTANCE.convert((LongRange) source);
+        } else if (targetType == EntityId.class && sourceType == Long.class) {
+            return EntityId.of((Long) source);
+        } else if (collectionType != null && sourceType == JSONB.class) {
+            return OBJECT_MAPPER.readValue(((JSONB) source).data(), collectionType);
```
steven-sheehy, line 121: This hardcoded mapping is a bit brittle. We should replace this with a single call to Spring's `ConversionService` for types that aren't the same on both side. Then we can register custom converters like `LongRangeConverter` and take advantage of the built-in converters. Could be a follow up ticket.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/repository/NftAllowanceRepositoryCustomImpl.java
```diff
@@ -0,0 +1,153 @@
...
+import java.util.List;
+import java.util.Map;
+import lombok.CustomLog;
+import lombok.RequiredArgsConstructor;
+import org.jooq.Condition;
+import org.jooq.DSLContext;
+import org.jooq.Field;
+import org.jooq.SortField;
+
+@CustomLog
```
steven-sheehy, line 38: unused

steven-sheehy (changes requested):
Still reviewing repository

hedera-mirror-rest-java/build.gradle.kts
```diff
@@ -18,18 +18,23 @@ description = "Hedera Mirror Node Rest Java"
...
     implementation(platform("org.springframework.cloud:spring-cloud-dependencies"))
     implementation(project(":common"))
     implementation("io.github.mweirauch:micrometer-jvm-extras")
     implementation("io.micrometer:micrometer-registry-prometheus")
     implementation("jakarta.inject:jakarta.inject-api")
+    implementation("org.jooq:jooq-postgres-extensions:$jooqVersion")
     implementation("org.springframework:spring-context-support")
     implementation("org.springframework.boot:spring-boot-actuator-autoconfigure")
     implementation("org.springframework.boot:spring-boot-configuration-processor")
+    implementation("org.springframework.boot:spring-boot-starter-jooq")
```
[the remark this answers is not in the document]
xin-hedera: will do

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/Order.java
```diff
@@ -0,0 +1,22 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hedera.mirror.restjava.common;
+
+public enum Order {
+    ASC,
+    DESC
+}
```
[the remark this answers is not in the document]
xin-hedera: will switch to it if no issue found

build.gradle.kts
```diff
@@ -31,6 +31,7 @@ plugins {
 // Can't use typed variable syntax due to Dependabot limitations
 extra.apply {
     set("grpcVersion", "1.62.2")
+    set("jooq.version", "3.19.6")
```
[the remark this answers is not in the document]
xin-hedera: the official jooq gradle plugin is only available from 3.19.x, and then I want to keep the version consistent between the plugin and our compile time dependencies. will add a note

buildSrc/build.gradle.kts
```diff
@@ -30,14 +39,20 @@ dependencies {
...
     implementation("io.freefair.gradle:lombok-plugin:8.6")
     implementation("io.spring.gradle:dependency-management-plugin:1.1.4")
     implementation("org.apache.commons:commons-compress:1.26.1")
+    implementation("org.flywaydb:flyway-database-postgresql:$flywayVersion")
     implementation("org.gradle:test-retry-gradle-plugin:1.5.8")
+    implementation("org.jooq:jooq-codegen-gradle:$jooqVersion")
+    implementation("org.jooq:jooq-meta:$jooqVersion")
     implementation("org.openapitools:openapi-generator-gradle-plugin:7.1.0")
     implementation("org.owasp:dependency-check-gradle:8.4.3")
+    implementation("org.postgresql:postgresql:42.7.2")
```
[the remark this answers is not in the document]
xin-hedera: I don't think so, can give it a try

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/exception/MismatchTypeException.java
```diff
@@ -0,0 +1,32 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hedera.mirror.restjava.exception;
+
+import java.io.Serial;
+
+@SuppressWarnings("java:S110")
+public class MismatchTypeException extends RestException {
```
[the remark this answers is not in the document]
xin-hedera: sure

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/jooq/DomainRecordMapper.java
xin-hedera: good catch. This should never happen, so I'll remove it. I think I was considering the inverted scenario, that we do have some transient field / some setters without corresponding column in our domain entity class

buildSrc/build.gradle.kts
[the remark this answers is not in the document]
xin-hedera: removed

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/jooq/DomainRecordMapper.java
xin-hedera: will create a ticket

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/repository/NftAllowanceRepositoryCustomImpl.java
xin-hedera: removed

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/repository/NftAllowanceRepositoryCustomImpl.java
```diff
@@ -0,0 +1,158 @@
...
+            new OrderSpec(false, Direction.ASC), List.of(NFT_ALLOWANCE.OWNER.asc(), NFT_ALLOWANCE.TOKEN_ID.asc()),
+            new OrderSpec(false, Direction.DESC), List.of(NFT_ALLOWANCE.OWNER.desc(), NFT_ALLOWANCE.TOKEN_ID.desc()));
+
+    private final DSLContext dslContext;
+
+    @NotNull
+    @Override
+    public Collection<NftAllowance> findAll(
+            boolean byOwner, @NotNull List<Filter<?>> filters, int limit, @NotNull Direction order) {
+        Filter<?> approvedForAllFilter = null;
```
xin-hedera, line 53: obviously forgot to support approved_for_all, now added.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/repository/NftAllowanceRepositoryCustomImpl.java
```diff
@@ -0,0 +1,158 @@
...
+            new OrderSpec(true, Direction.DESC), List.of(NFT_ALLOWANCE.SPENDER.desc(), NFT_ALLOWANCE.TOKEN_ID.desc()),
+            new OrderSpec(false, Direction.ASC), List.of(NFT_ALLOWANCE.OWNER.asc(), NFT_ALLOWANCE.TOKEN_ID.asc()),
+            new OrderSpec(false, Direction.DESC), List.of(NFT_ALLOWANCE.OWNER.desc(), NFT_ALLOWANCE.TOKEN_ID.desc()));
+
+    private final DSLContext dslContext;
+
+    @NotNull
+    @Override
+    public Collection<NftAllowance> findAll(
+            boolean byOwner, @NotNull List<Filter<?>> filters, int limit, @NotNull Direction order) {
```
steven-sheehy, line 52: There's a bit of inefficiency in that we convert from flattened variables in controller to `NftAllowanceRequest` for service then to this filters list for repository. We might have to revisit this once all the layers are in place and consolidate to one level of mapping.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/repository/NftAllowanceRepositoryCustomImpl.java
steven-sheehy: It might be better and reusable to define this mapping within the `RangeOperator` enum. It could return some lambda to apply to the field and value.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/Filter.java
```diff
@@ -0,0 +1,29 @@
...
+import com.hedera.mirror.restjava.exception.MismatchedTypeException;
+import org.jooq.Field;
+
+public record Filter<T>(Field<?> field, RangeOperator operator, T value, Class<T> type) {
+
+    public Filter {
+        if (field.getType() != type) {
+            throw new MismatchedTypeException(field.getType(), type);
+        }
+    }
```
steven-sheehy, line 28: Is the extra type parameter really needed? Doesn't the Field and the `T value` have the necessary type info?

```suggestion
public record Filter<T>(Field<?> field, RangeOperator operator, T value) {
```

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/repository/NftAllowanceRepositoryCustomImpl.java
```diff
@@ -0,0 +1,158 @@
...
+
+    private Condition getSecondaryCondition(
+            Condition commonCondition, Filter<?> primarySortFilter, Filter<?> tokenFilter) {
+        // No secondary condition if there is no token filter, or the primary sort filter's operator is EQ. Note that
+        // it's guaranteed that there must be a primary sort filter when token filter exists
+        if (tokenFilter == null || primarySortFilter.operator() == RangeOperator.EQ) {
+            return null;
+        }
+
+        var value = (Long) primarySortFilter.value();
```
steven-sheehy, line 131: It doesn't seem to provide value to have the generics. We should declare the parameters with `Filter<Long>` or consider removing the generics.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/repository/NftAllowanceRepositoryCustomImpl.java
```diff
@@ -0,0 +1,158 @@
...
+
+        return dslContext
+                .selectFrom(NFT_ALLOWANCE)
+                .where(condition)
+                .orderBy(SORT_ORDERS.get(new OrderSpec(byOwner, order)))
+                .limit(limit)
+                .fetchInto(NftAllowance.class);
+    }
+
+    private Condition getBaseCondition(Condition commonCondition, Filter<?> primarySortFilter, Filter<?> tokenFilter) {
```
steven-sheehy, line 98: This logic might be good to have in a generic `JooqRepository` in the future.

steven-sheehy (changes requested):

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/repository/NftAllowanceRepositoryCustomImpl.java
xin-hedera: I think the NftAllownceServiceRequest will change a lot or go away.

We may need a factory class to provide lookup of http query parameter + endpoint to jooq table fields.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/Filter.java
xin-hedera: Yes. The other way to get the class type during runtime of reflection, of cause we dont want to pay the huge cost as this is frequently instantiated

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/repository/NftAllowanceRepositoryCustomImpl.java
xin-hedera: the generics in Filter provides the type safety so when initiated, the value for the field has matching type.

We either cast the value to `Long` here, or cast the filter to `Filter<Long>` beforehand (because the method has `List<<Filter<?>>`

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/repository/NftAllowanceRepositoryCustomImpl.java
mgoelswirlds: As a follow-up , we would need to change the service layer to integrate the controller and refactor NftAllownceServiceRequest. Will create a followup ticket.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/exception/RestException.java
```diff
@@ -0,0 +1,34 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hedera.mirror.restjava.exception;
+
+import com.hedera.mirror.common.exception.MirrorNodeException;
+import java.io.Serial;
+
+abstract class RestException extends MirrorNodeException {
```
mgoelswirlds, line 22: I have this same class in my PR. So depending on which PR goes in first the other can delete it.

steven-sheehy (approved):
LGTM

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/repository/NftAllowanceRepositoryCustomImpl.java
mgoelswirlds: I don't believe we would need the approved_for_all filter.

mgoelswirlds (approved):
LGTM

PR #7883, opened 2024-03-08
Adapt gas_consumed migration query to get failed initcode from contract_result for failed contract create transactions

### LINKED ISSUE #7882: Migration query for gas_consumed doesnt use failed initcode for failed create transactions
### Description

Migration query for gas_consumed doesnt use failed initcode for failed create transactions, because it seeks for created contract in the contract table, thus not factoring the initcode intrinsic gas calculation properly.

### Steps to reproduce

Run a create transaction with bad contract:
```
contract DeployDummyContract{
    DummyContract dummyContract;
    constructor(bool flag) payable {
        dummyContract = new DummyContract(flag);
    }
}
contract DummyContract {
    constructor(bool shouldFail) payable {
        if (shouldFail) {
            revert("DummyContract deployment failed intentionally.");
        }
    }
}
```

### Additional context

_No response_

### Hedera network

other

### Version

0.100.0

### Operating system

None

MrValioBg:
**Description**:
This PR adapts migration query for gas_consumed to check if failed_initcode column is not null, this way we handle the case where if a contract create transaction is failed and we haven't populated the contract table we will still factor the initcode in the intrinsic gas calculation.

We have to cherry pick the fix after merge.

**Related issue(s)**:

Fixes #7882 

**Notes for reviewer**:

**Checklist**


hedera-mirror-importer/src/main/resources/db/migration/v1/V1.94.1__add_gas_consumed_field.sql
```diff
@@ -18,12 +18,17 @@ with contract_action_gas_usage as (
         e.created_timestamp,
         -- Correctly assign `payload` based on whether it's a contract creation or call.
         coalesce(
-        case
-            when cr.consensus_timestamp = e.created_timestamp then c.initcode
-            else cr.function_parameters
-        end, ''::bytea) as payload,
+            case
+                when cr.consensus_timestamp = e.created_timestamp then c.initcode
+                when cr.failed_initcode is not null then cr.failed_initcode
```
steven-sheehy, line 23: You can't modify migrations as we've already deployed this to environments and made available in tags, so it's already ran and Flyway will fail if the checksum differs.

You'll have to create a new migration. We should optimize the new migration to just loop over rows with failed_initcode is not null to reduce runtime.

steven-sheehy (changes requested):

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.94.1__add_gas_consumed_field.sql
xin-hedera: perhaps renaming it to 1.94.1.1 is an easier approach. The changes in the migration are idempotent and for preprod rerunning the backfill logic should take long except mainnet staging.

MrValioBg: I've gone ahead and renamed the migration as well as changed the source path for the sql file in GasConsumedMigrationTest.java

xin-hedera (approved):
LGTM

steven-sheehy (approved):
LGTM

Issue #7895, opened 2024-03-11
Transfer of token to hollow account fails with INVALID_ACCOUNT_ID


bilyana-gospodinova:
### Description

On transferFromNFT or transferFrom HTS call to a hollow account the transaction fails as the account could not be found in the store. The same call is successful in services. The difference is in the `DecodingFacade`. There is a method called `convertLeftPaddedAddressToAccountId` that has a third parameter (called `exists`) and in this case the account is created from the alias but in the mirror node the account remains the default one and in `HederaTokenStore#checkAccountUsability` an `INVALID_ACCOUNT_ID` error is thrown.

### Steps to reproduce

Execute acceptance tests:
- Then I call precompile with transferFrom "NFTEQUIVALENCE" token to an EVM address
- Then I call precompile with transferFrom "FUNGIBLE" token to an EVM address

### Hedera network

testnet

### Version

v0.99.0

### Operating system

None

PR #7908, opened 2024-03-15
07895: TransferFrom token to hollow account fails

bilyana-gospodinova:
**Description**:
On `transferFromNFT` or `transferFrom` HTS call to a hollow account the transaction fails as the account could not be found in the store. This PR adds the missing implementation from services that creates an account in case it does not exist. This change broke some of the other in-equivalence tests for system accounts so another missing implementation was added - `TransferPrecompile#revertIfReceiverIsSystemAccount`. After these changes an error is thrown when a transfer is attempted to a system account. Some of the integration tests are adjusted as well - a new system account is defined in order to be used for the system account related tests and the old owner account is now used in the regular cases. 

Fixes https://github.com/hashgraph/hedera-mirror-node/issues/7895

### DIFF
--- a/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallTestSetup.java
+++ b/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallTestSetup.java
@@ -146,7 +146,7 @@ public class ContractCallTestSetup extends Web3IntegrationTest {
     // Account addresses
     protected static final Address AUTO_RENEW_ACCOUNT_ADDRESS = toAddress(EntityId.of(0, 0, 740));
     protected static final Address AUTO_RENEW_ACCOUNT_ADDRESS_HISTORICAL = toAddress(EntityId.of(0, 0, 1078));
-    protected static final Address SPENDER_ADDRESS = toAddress(EntityId.of(0, 0, 741));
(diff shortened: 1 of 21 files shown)


hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallTestSetup.java
```diff
@@ -1856,6 +1857,21 @@ private EntityId ownerEntityPersistHistorical() {
         return ownerEntityId;
     }
 
+    @Nullable
+    private EntityId systemAccountEntityPersist() {
+        final var systemAccountEntityId = fromEvmAddress(SYSTEM_ACCOUNT_ADDRESS.toArrayUnsafe());
+
+        domainBuilder
+                .entity()
+                .customize(e -> e.id(systemAccountEntityId.getId())
```
IvanKavaldzhiev, line 1866: I don't think system accounts should have their own balance.

bilyana-gospodinova:
This PR has duplicate logic as https://github.com/hashgraph/hedera-mirror-node/pull/8200. This one here needs to be updated to contain only the changes related to the system accounts in-equivalence acceptance tests.

bilyana-gospodinova:
Closing the PR as discussed. The tests are no longer needed.

PR #7953, opened 2024-03-22
Lazy load nested items during contract call

### LINKED ISSUE #7903: Lazy load EVM entity objects
### Problem

When invoking a simple precompile it eagerly loads the `Account` and its nested info like NFT count, allowances, etc. even if it doesn't require the information to return a result.

### Solution

Change EVM objects returned from `DatabaseAccessor`'s like `Account`, `Token`, etc. to not eagerly query for nested information. These objects should be enhanced to be constructed with `Supplier` for nested info and the getters should invoke the supplier when needed. Guava's `Suppliers.memoize()` can be used to wrap the supplier to cache it in case it's invoked multiple times.

### Alternatives

_No response_

IvanKavaldzhiev:
**Description**:

This PR enhances db accessors' performance by lazy loading some of the nested fields, which are not always directly used when fetching the models.
**Related issue(s)**:

Fixes #7903 

**Notes for reviewer**:

To be tested with acceptance and K6 tests
**Checklist**


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/contract/HederaEvmStackedWorldStateUpdater.java
```diff
@@ -127,17 +129,17 @@ private void persistAccount(Address address, long nonce, Wei balance) {
...
                 null,
                 0,
-                Collections.emptySortedMap(),
-                Collections.emptySortedMap(),
-                Collections.emptySortedSet(),
-                0,
-                0,
+                TreeMap::new,
+                TreeMap::new,
+                TreeSet::new,
```
steven-sheehy, line 140: Prefer zero allocation collections unless needed.
```suggestion
                Collections::emptySortedMap,
                Collections::emptySortedMap,
                Collections::emptySortedSet,
```

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/Account.java
```diff
@@ -148,17 +150,17 @@ public Account(Long entityId, Id id, long balance) {
                 entityId,
                 id,
                 0L,
-                balance,
+                Suppliers.memoize(() -> balance),
```
steven-sheehy, line 153:
```suggestion
                () -> balance,
```

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/Account.java
```diff
@@ -148,17 +150,17 @@ public Account(Long entityId, Id id, long balance) {
...
                 null,
                 0,
-                new TreeMap<>(),
-                new TreeMap<>(),
-                new TreeSet<>(),
-                0,
-                0,
+                () -> Collections.unmodifiableSortedMap(new TreeMap<>()),
+                () -> Collections.unmodifiableSortedMap(new TreeMap<>()),
+                () -> Collections.unmodifiableSortedSet(new TreeSet<>()),
```
steven-sheehy, line 161:
```suggestion
                Collections::emptySortedMap,
                Collections::emptySortedMap,
                Collections::emptySortedSet,
```

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/Account.java
```diff
@@ -176,17 +178,17 @@ public Account(ByteString alias, Long entityId, Id id, long balance) {
                 entityId,
                 id,
                 0L,
-                balance,
+                Suppliers.memoize(() -> balance),
```
steven-sheehy, line 181:
```suggestion
                () -> balance,
```

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/Account.java
```diff
@@ -176,17 +178,17 @@ public Account(ByteString alias, Long entityId, Id id, long balance) {
...
                 null,
                 0,
-                new TreeMap<>(),
-                new TreeMap<>(),
-                new TreeSet<>(),
-                0,
-                0,
+                () -> Collections.unmodifiableSortedMap(new TreeMap<>()),
+                () -> Collections.unmodifiableSortedMap(new TreeMap<>()),
+                () -> Collections.unmodifiableSortedSet(new TreeSet<>()),
```
steven-sheehy, line 189:
```suggestion
                Collections::emptySortedMap,
                Collections::emptySortedMap,
                Collections::emptySortedSet,
```

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/Account.java
```diff
@@ -223,7 +225,7 @@ private Account createNewAccountWithNewOwnedNfts(final Account oldAccount, final
                 oldAccount.expiry,
                 oldAccount.balance,
                 oldAccount.deleted,
-                ownedNfts,
+                Suppliers.memoize(() -> ownedNfts),
```
steven-sheehy, line 228:
```suggestion
                () -> ownedNfts,
```

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/Account.java
```diff
@@ -262,7 +264,7 @@ private Account createNewAccountWithNumAssociations(final Account oldAccount, fi
                 oldAccount.cryptoAllowances,
                 oldAccount.fungibleTokenAllowances,
                 oldAccount.approveForAllNfts,
-                numAssociations,
+                Suppliers.memoize(() -> numAssociations),
```
steven-sheehy, line 267:
```suggestion
                () -> numAssociations,
```

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/Account.java
```diff
@@ -295,7 +297,7 @@ private Account createNewAccountWithNewPositiveBalances(Account oldAccount, int
                 oldAccount.fungibleTokenAllowances,
                 oldAccount.approveForAllNfts,
                 oldAccount.numAssociations,
-                newNumPositiveBalances,
+                Suppliers.memoize(() -> newNumPositiveBalances),
```
steven-sheehy, line 300:
```suggestion
                () -> newNumPositiveBalances,
```

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/Account.java
```diff
@@ -414,7 +416,7 @@ private Account createNewAccountWithNewBalance(Account oldAccount, long newBalan
                 oldAccount.entityId,
                 oldAccount.id,
                 oldAccount.expiry,
-                newBalance,
+                Suppliers.memoize(() -> newBalance),
```
steven-sheehy, line 419:
```suggestion
                () -> newBalance,
```

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/Account.java
```diff
@@ -453,7 +455,7 @@ private Account createNewAccountWithNewCryptoAllowances(
                 oldAccount.autoRenewSecs,
                 oldAccount.proxy,
                 oldAccount.autoAssociationMetadata,
-                cryptoAllowances,
+                Suppliers.memoize(() -> cryptoAllowances),
```
steven-sheehy, line 458:
```suggestion
                () -> cryptoAllowances,
```

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/Account.java
```diff
@@ -487,7 +489,7 @@ private Account createNewAccountWithNewFungibleTokenAllowances(
                 oldAccount.proxy,
                 oldAccount.autoAssociationMetadata,
                 oldAccount.cryptoAllowances,
-                fungibleTokenAllowances,
+                Suppliers.memoize(() -> fungibleTokenAllowances),
```
steven-sheehy, line 492:
```suggestion
                () -> fungibleTokenAllowances,
```

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/Account.java
```diff
@@ -521,7 +523,7 @@ private Account createNewAccountWithNewApproveForAllNfts(
                 oldAccount.autoAssociationMetadata,
                 oldAccount.cryptoAllowances,
                 oldAccount.fungibleTokenAllowances,
-                newApproveForAllNfts,
+                Suppliers.memoize(() -> newApproveForAllNfts),
```
steven-sheehy, line 526:
```suggestion
                () -> newApproveForAllNfts,
```

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/Account.java
```diff
@@ -667,77 +669,114 @@ public Account decrementUsedAutomaticAssociations() {
...
+    public Long getBalance() {
+        return balance != null && balance.get() != null ? balance.get() : 0L;
+    }
+
+    public Long getOwnedNfts() {
+        return ownedNfts != null ? ownedNfts.get() : 0L;
+    }
+
+    public SortedMap<EntityNum, Long> getCryptoAllowances() {
+        return cryptoAllowances != null ? cryptoAllowances.get() : Collections.unmodifiableSortedMap(new TreeMap<>());
```
steven-sheehy, line 681:
```suggestion
        return cryptoAllowances != null ? cryptoAllowances.get() : Collections.emptySortedMap();
```

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/Account.java
```diff
@@ -667,77 +669,114 @@ public Account decrementUsedAutomaticAssociations() {
...
+    }
+
+    public SortedMap<FcTokenAllowanceId, Long> getFungibleTokenAllowances() {
+        return fungibleTokenAllowances != null
+                ? fungibleTokenAllowances.get()
+                : Collections.unmodifiableSortedMap(new TreeMap<>());
```
steven-sheehy, line 687:
```suggestion
                : Collections.emptySortedMap();
```

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/Account.java
```diff
@@ -667,77 +669,114 @@ public Account decrementUsedAutomaticAssociations() {
...
+    }
+
+    public SortedSet<FcTokenAllowanceId> getApproveForAllNfts() {
+        return approveForAllNfts != null ? approveForAllNfts.get() : Collections.unmodifiableSortedSet(new TreeSet<>());
```
steven-sheehy, line 691:
```suggestion
        return approveForAllNfts != null ? approveForAllNfts.get() : Collections.emptySortedSet();
```

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/TokenRelationship.java
```diff
@@ -117,7 +118,7 @@ private TokenRelationship createNewTokenRelationshipWithNewBalance(
         return new TokenRelationship(
                 oldTokenRel.token,
                 oldTokenRel.account,
-                balance,
+                Suppliers.memoize(() -> balance),
```
steven-sheehy, line 121:
```suggestion
                (() -> balance,
```

hedera-mirror-web3/src/main/java/com/hedera/services/txns/crypto/AbstractAutoCreationLogic.java
```diff
@@ -109,17 +111,17 @@ public Pair<ResponseCodeEnum, Long> create(
...
                 null,
                 0,
-                Collections.emptySortedMap(),
-                Collections.emptySortedMap(),
-                Collections.emptySortedSet(),
-                0,
-                0,
+                () -> Collections.unmodifiableSortedMap(new TreeMap<>()),
+                () -> Collections.unmodifiableSortedMap(new TreeMap<>()),
+                () -> Collections.unmodifiableSortedSet(new TreeSet<>()),
```
steven-sheehy, line 122:
```suggestion
                Collections::emptySortedMap,
                Collections::emptySortedMap,
                Collections::emptySortedSet,
```

hedera-mirror-web3/src/test/java/com/hedera/services/txn/token/DissociateLogicTest.java
```diff
@@ -260,17 +261,17 @@ void verifyUpdagtedNumPositiveBalance() {
...
                 0L,
                 null,
                 3,
                 null,
                 null,
                 null,
-                3,
-                3,
+                Suppliers.ofInstance(3),
+                Suppliers.ofInstance(3),
```
steven-sheehy, line 274: Why are we using this `Suppliers.ofInstance` in tests instead of `() -> 3`?

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/Token.java
```diff
@@ -271,8 +272,8 @@ public static Token fromGrpcOpAndMeta(
                 feeScheduleKey.orElse(null),
                 pauseKey.orElse(null),
                 op.getFreezeDefault(),
-                treasury,
-                autoRenewAccount,
+                Suppliers.memoize(() -> treasury),
+                Suppliers.memoize(() -> autoRenewAccount),
```
steven-sheehy, line 276:
```suggestion
                () -> treasury,
                () -> autoRenewAccount,
```

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/Token.java
```diff
@@ -285,7 +286,7 @@ public static Token fromGrpcOpAndMeta(
                 op.getDecimals(),
                 op.getAutoRenewPeriod().getSeconds(),
                 0,
-                Collections.emptyList());
+                () -> Collections.unmodifiableList(new ArrayList<>()));
```
steven-sheehy, line 289:
```suggestion
                Collections::emptyList);
```

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/Token.java
```diff
@@ -409,7 +410,7 @@ private Token createNewTokenWithNewTotalSupply(Token oldToken, long totalSupply)
                 true,
                 oldToken.type,
                 oldToken.supplyType,
-                totalSupply,
+                Suppliers.memoize(() -> totalSupply),
```
steven-sheehy, line 413:
```suggestion
                () -> totalSupply,
```

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/Token.java
```diff
@@ -464,7 +465,7 @@ private Token createNewTokenWithNewTreasury(Token oldToken, Account treasury) {
                 oldToken.feeScheduleKey,
                 oldToken.pauseKey,
                 oldToken.frozenByDefault,
-                treasury,
+                Suppliers.memoize(() -> treasury),
```
steven-sheehy, line 468:
```suggestion
                () -> treasury,
```

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/Token.java
```diff
@@ -824,7 +825,7 @@ private Token createNewTokenWithTreasury(Token oldToken, Account treasury) {
...
```
steven-sheehy, line 828:
```suggestion
                () -> treasury,
```

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/Token.java
```diff
@@ -1049,7 +1050,7 @@ private Token createNewTokenWithAutoRenewAccount(Token oldToken, Account autoRen
                 oldToken.pauseKey,
                 oldToken.frozenByDefault,
                 oldToken.treasury,
-                autoRenewAccount,
+                Suppliers.memoize(() -> autoRenewAccount),
```
steven-sheehy, line 1053:
```suggestion
                () -> autoRenewAccount,
```

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/Token.java
```diff
@@ -1840,7 +1843,9 @@ public Token setLastUsedSerialNumber(final long lastUsedSerialNumber) {
     }
 
     public List<CustomFee> getCustomFees() {
-        return customFees;
+        return customFees != null && customFees.get() != null
+                ? customFees.get()
+                : Collections.unmodifiableList(new ArrayList<>());
```
steven-sheehy, line 1848:
```suggestion
                : Collections.emptyList();
```

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/store/accessor/TokenDatabaseAccessorTest.java
```diff
@@ -215,9 +215,6 @@ void getAutoRenewHistorical() {
         when(autorenewEntity.getRealm()).thenReturn(12L);
         when(autorenewEntity.getNum()).thenReturn(13L);
         when(autorenewEntity.getBalance()).thenReturn(14L);
-        when(entityRepository.findActiveByIdAndTimestamp(
-                        databaseToken.getTreasuryAccountId().getId(), timestamp.get()))
-                .thenReturn(Optional.of(autorenewEntity));
```
steven-sheehy, line 220: We should have tests in all the accessor tests that verify a) Not loaded for top level and b) is lazy loaded for all nested items.

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/Account.java
IvanKavaldzhiev: I believe it would be a good idea to return unmodifiable collections in the getters, so that we don't introduce side effects of changing a model field, which should be immutable as a whole and replaced entirely in the stacked state in case of update.

hedera-mirror-web3/src/test/java/com/hedera/services/txn/token/DissociateLogicTest.java
IvanKavaldzhiev: Forgot to change it to a lambda.

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/Account.java
IvanKavaldzhiev: But we should also wrap the existing `cryptoAllowances` collection that might have been set, so probably we can have an `unmodifiable` definition over the whole expression, as I refactored the code? If we don't - we can set a normal collection that can remain modifiable.

hedera-mirror-web3/src/main/java/com/hedera/services/store/models/Token.java
```diff
@@ -1840,7 +1842,8 @@ public Token setLastUsedSerialNumber(final long lastUsedSerialNumber) {
     }
 
     public List<CustomFee> getCustomFees() {
-        return customFees;
+        return Collections.unmodifiableList(
+                customFees != null && customFees.get() != null ? customFees.get() : new ArrayList<>());
```
steven-sheehy, line 1846: `customFees.get()` should never be null per `TokenDatabaseAccessor.getCustomFees()`.

```suggestion
        return customFees != null ? Collections.unmodifiableList(customFees.get()) : Collections.emptyList();
```

steven-sheehy (changes requested):

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/store/contract/HederaEvmWorldStateTest.java
```diff
@@ -208,17 +209,17 @@ void commitsNewlyCreatedAccountToStackedStateFrames() {
                 0L,
                 Id.fromGrpcAccount(accountIdFromEvmAddress(address.toArrayUnsafe())),
                 0L,
-                123L,
+                Suppliers.ofInstance(123L),
                 false,
-                0L,
+                Suppliers.ofInstance(0L),
```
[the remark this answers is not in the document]
IvanKavaldzhiev: I refactored all occurrences

steven-sheehy (dismissed):
LGTM

kselveliev (dismissed):
LGTM

hedera-mirror-web3/src/main/java/com/google/common/base/Suppliers.java
```diff
@@ -0,0 +1,181 @@
...
+            try {
+                if (delegate != SUCCESSFULLY_COMPUTED) {
+                    T t = delegate.get();
+                    value = t;
+                    delegate = (Supplier<T>) SUCCESSFULLY_COMPUTED;
+                    return t;
+                }
+            } finally {
+                lock.unlock();
+            }
```
steven-sheehy, line 167: This is slower than before since it always locks instead of only locking when it's not present. I also don't think we need to guarantee that the delegate only runs once so we can get away with something simpler. Finally, we should not override Guava's class since Guava doesn't sound like they're ever going to change their implementation. I suggest this approach in our own util class like `Web3Utils`:

```java
    public static <T> Supplier<T> memoize(Supplier<T> delegate) {
        final var value = new AtomicReference<T>();
        return () -> {
            final var previousValue = value.get();
            if (previousValue != null) {
                return previousValue;
            }

            final var newValue = delegate.get();
            value.set(newValue);
            return newValue;
        };
    }
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/utils/Suppliers.java
```diff
@@ -0,0 +1,96 @@
...
+            return newValue;
+        };
+    }
+
+    /** Returns a supplier that always supplies {@code instance}. */
+    public static <T extends @Nullable Object> Supplier<T> ofInstance(T instance) {
+        return new SupplierOfInstance<>(instance);
+    }
+
+    private static class SupplierOfInstance<T extends @Nullable Object> implements Supplier<T>, Serializable {
```
steven-sheehy, line 63: Remove this method and class since they are unused and we should be using the guava version.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/AccountDatabaseAccessor.java
```diff
@@ -141,52 +143,66 @@ private Long getAccountBalance(Entity entity, final Optional<Long> timestamp) {
...
 
-    private TokenAccountBalances getNumberOfAllAndPositiveBalanceTokenAssociations(
+    private Supplier<Integer> getAllBalanceTokenAssociations(long accountId, final Optional<Long> timestamp) {
+        return Suppliers.memoize(() -> getNumberOfAllAndPositiveBalanceTokenAssociations(accountId, timestamp)
+                .get()
+                .all());
+    }
+
+    private Supplier<Integer> getPositiveBalanceTokenAssociations(long accountId, final Optional<Long> timestamp) {
+        return Suppliers.memoize(() -> getNumberOfAllAndPositiveBalanceTokenAssociations(accountId, timestamp)
```
steven-sheehy, line 200: This doesn't share the `getNumberOfAllAndPositiveBalanceTokenAssociations` supplier across these two methods so it's still run twice. We should remove these two methods and do the following in `accountFromEntity`:

```java
var tokenAccountBalances = getNumberOfAllAndPositiveBalanceTokenAssociations(account, timestamp);
var all = () -> tokenAccountBalances.get().all();
var positive = () -> tokenAccountBalances.get().positive();
```

steven-sheehy (changes requested):

steven-sheehy (approved):
LGTM 🥂

steven-sheehy (approved):

PR #7984, opened 2024-03-29
Fix address(this) resolution when mirror address is passed in the call

### LINKED ISSUE #7857: Inconsistency in contract address resolution with `address(this)`
##  **Description:**
A member from SaucerSwap has reported an inconsistency in contract address resolution with `address(this)` method when a contract is created using create2.

`Using the "contracts/call" endpoint on the mirror node, we called a view function on a contract that uses address(this).  This contract was created by another contract and has a create2 address/salt, so the evm address is not in the long zero format.  What we found was that address(this) in the contract will be whatever is specified as the 'to' parameter in the http request.  This could lead to some unintended consequences if calls rely on address(this) to hash something or as a key into a key/value store.
`

### Smart contracts in question: 
- AddrCheck contract code:
```solidity
pragma solidity ^0.8.0;

contract AddrCheck {
    function getAddressThis() public view returns(address) {
        return address(this);
    }
}
```

- Factory contract code:
```solidity
pragma solidity ^0.8.0;

import "./AddrCheck.sol";

contract Factory {    
    function makeContract() public returns(address) {
        return address(new AddrCheck{salt: keccak256(abi.encodePacked(msg.sender, block.timestamp))}());;
    }
}
```

In the contracts above, `Factory` creates an `AddrCheck` contract with a keccak hash for a salt value. We can then call the view function `getAddressThis()` like in the curl requests below (function selector `0x0a83b9a6`) and see that the result is the long zero address or the create2 address depending on the input of the `to` field. In the CURL requests below, the `to` field points at the `AddrCheck` contract which is created by `create2` in the `Factory` contract. The `to` fields in the 2 requests have two different addresses but both addresses are reserved for the same contract instance.

### **Example CURL Requests:**
```bash
# Request 1 - `to` field contains long zero address => result returns long zero address 
curl -X 'POST' \
  'https://testnet.mirrornode.hedera.com/api/v1/contracts/call' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "block": "latest",
  "data": "0x0a83b9a6",
  "estimate": false,
  "from": "0x000000000000000000000000000000000000cc5e",
  "gas": 15000000,
  "gasPrice": 100000000,
  "to": "0x000000000000000000000000000000000037a4a6"
}'
{"result":"0x000000000000000000000000000000000000000000000000000000000037a4a6"}

# Request 2 - `to` field contains create2 address => result returns create2 address 
curl -X 'POST' \
  'https://testnet.mirrornode.hedera.com/api/v1/contracts/call' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "block": "latest",
  "data": "0x0a83b9a6",
  "estimate": false,
  "from": "0x000000000000000000000000000000000000cc5e",
  "gas": 15000000,
  "gasPrice": 100000000,
  "to": "0x5357c6838dfd31de4d434974c4584b5162c571e2"
}'
{"result":"0x0000000000000000000000005357c6838dfd31de4d434974c4584b5162c571e2"}
```

It's worth notice that when we call this endpoint `contracts/{contractIdOrAddress}` with the same long zero address, the `evm_address` attribute in the returned object contains the create2 address instead of the long zero address. Here is a call to `contracts/{contractIdOrAddress}` with the long zero address

```bash
curl -X 'GET' \
  'https://testnet.mirrornode.hedera.com/api/v1/contracts/0x000000000000000000000000000000000037a4a6' \
  -H 'accept: application/json'

{
  "admin_key": null,
  "auto_renew_account": null,
  "auto_renew_period": 7776000,
  "contract_id": "0.0.3646630",
  "created_timestamp": "1709319580.503908004",
  "deleted": false,
  "evm_address": "0x5357c6838dfd31de4d434974c4584b5162c571e2", // this is the create2 address
  "expiration_timestamp": "1717095580.503908004",
  "file_id": null,
  "max_automatic_token_associations": 0,
  "memo": "",
  "nonce": 1,
  "obtainer_id": null,
  "permanent_removal": null,
  "proxy_account_id": null,
  "timestamp": {
    "from": "1709319580.503908004",
    "to": null
  },
  "bytecode": "0x60806040...",
  "runtime_bytecode": "0x60806040..."
}
```

### **More Findings with the Relay:**
When we submit an `eth_call` call with the same parameters, we receive the same behavior. The `result` will return whichever defined in the `to` field.

```bash
# Request 1 - `to` field contains long zero address => result returns long zero address 
curl --location 'https://testnet.hashio.io/api' \
--header 'Content-Type: application/json' \
--data '{
    "id": "2",
    "jsonrpc": "2.0",
    "method": "eth_call",
    "params": [
        {
            "to": "0x000000000000000000000000000000000037a4a6",
            "data": "0x0a83b9a6"
        },
        "latest"
    ]
}'
{"result":"0x000000000000000000000000000000000000000000000000000000000037a4a6","jsonrpc":"2.0","id":"2"}

# Request 2 - `to` field contains create2 address => result returns create2 address 
curl --location 'https://testnet.hashio.io/api' \
--header 'Content-Type: application/json' \
--data '{
    "id": "2",
    "jsonrpc": "2.0",
    "method": "eth_call",
    "params": [
        {
            "to": "0x5357c6838dfd31de4d434974c4584b5162c571e2",
            "data": "0x0a83b9a6"
        },
        "latest"
    ]
}'
{"result":"0x0000000000000000000000005357c6838dfd31de4d434974c4584b5162c571e2","jsonrpc":"2.0","id":"2"}

```

However, when we submit an `eth_sendRawRequest` call with the `to` field contains the long zero address, we would receive a transaction hash in return. We can then make a call to the`contracts/result/{transactionIdOrHash}` endpoint with the transaction hash, then observe the `address` & `call_result` fields contains the create2 address.

```bash
#### Signs the transaction
const tx = {
        to: '0x000000000000000000000000000000000037a4a6',
        data: '0x0a83b9a6',
        value: 0,
        gasLimit: 50000,
        chainId: Number(CHAIN_ID),
        nonce: await relay.getAccountNonce(accounts[1].address),
        gasPrice: gasPrice,
        maxPriorityFeePerGas: gasPrice,
        maxFeePerGas: gasPrice,
        type: 2,
 };
const signedTx = await accounts[1].wallet.signTransaction(tx);
console.log(signedTx) // 0x02f8728201288085b0f387b00085b0f387b00082c35094000000000000000000000000000000000037a4a680840a83b9a6c001a0b119d12297376d595d47d8c6d8ffaf85e30c3489effa80eacd32d26545ba2a2fa043cb9758641fdc2bcaad45752a855bd6f6dbdd9587973952ffc44198d6eb134f
```

#### Submit the `eth_sendRawTransaction` call

```
curl --location 'https://testnet.hashio.io/api' \                                                                              ─╯
--header 'Content-Type: application/json' \
--data '{
    "id": "2",
    "jsonrpc": "2.0",
    "method": "eth_sendRawTransaction",
    "params": [ "0x02f8728201288085b0f387b00085b0f387b00082c35094000000000000000000000000000000000037a4a680840a83b9a6c080a0f1c811545742e9e4922c73f899b1bd4de0b2e0c02e61ee4921b78216c180fa3ba0068b8543afcfd2782701e3a6e93594dbf459e813a1576ffc98f6455f760c98e7",
        "latest"
    ]
}'
{"result":"0x2f1776b7f63f48bbb61031842b0ec81423f7388440d9c5d19f5538eb239b7e86","jsonrpc":"2.0","id":"2"}
```

#### Make a call to `contracts/result/{transactionIdOrHash}` endpoint
```
curl -X 'GET' \
  'https://testnet.mirrornode.hedera.com/api/v1/contracts/results/0x2f1776b7f63f48bbb61031842b0ec81423f7388440d9c5d19f5538eb239b7e86?nonce=1' \
  -H 'accept: application/json'

{
    "address": "0x5357c6838dfd31de4d434974c4584b5162c571e2", /// this is the create2 address
    "amount": 0,
    "bloom": "0x00...",
    "call_result": "0x0000000000000000000000005357c6838dfd31de4d434974c4584b5162c571e2", /// this contains the create2 address
    "contract_id": "0.0.3646630",
    "created_contract_ids": [],
    "error_message": null,
    "from": "0x000000000000000000000000000000000037baf1",
    "function_parameters": "0x0a83b9a6",
    "gas_limit": 50000,
    "gas_used": 40000,
    "timestamp": "1709508080.511153073",
    "to": "0x000000000000000000000000000000000037a4a6",
    "hash": "0x2f1776b7f63f48bbb61031842b0ec81423f7388440d9c5d19f5538eb239b7e86",
    "block_hash": "0x9e36b57443041c0631ac8b67f00406737d8320c53cd0885653ab6d96f461fb2784eb54fddb7960ec64f947aabbefa116",
    "block_number": 1185673,
    "logs": [],
    "result": "SUCCESS",
    "transaction_index": 1,
    "state_changes": [],
    "status": "0x1",
    "failed_initcode": null,
    "access_list": "0x",
    "block_gas_used": 40000,
    "chain_id": "0x128",
    "gas_price": "0x",
    "max_fee_per_gas": "0x4c",
    "max_priority_fee_per_gas": "0x4c",
    "r": "0xf1c811545742e9e4922c73f899b1bd4de0b2e0c02e61ee4921b78216c180fa3b",
    "s": "0x068b8543afcfd2782701e3a6e93594dbf459e813a1576ffc98f6455f760c98e7",
    "type": 2,
    "v": 0,
    "nonce": 0
}
```

With these findings, it sounds like the consensus node will return the consistent value. Meanwhile the web3 module on the mirror node has the issue. It's worth notice that this is simply just an assumption. 

### Steps to reproduce

Please follow along with the CURL requests mentioned in the Description. All the requests happen within the `testnet` context.

### Additional context

_No response_

### Hedera network

mainnet, testnet, previewnet, other

### Version

latest

### Operating system

None

kselveliev:
There is an inconsistency on call address(this) function depending on the **to** parameter in a contract call request.
Address(this) will return whatever is specified as the 'to' parameter in the http request.
If a contract is created via ethereum transaction on hedera it will have a canonical ethereum address. It will also have the long zero address.
So in our contract call request we can put in the **to** parameter either the canonical evm address or the long zero and it should return the same response.
Currently if we call address(this) of a contract that has a canonical evm address but we pass it's long zero address in the contract call request it will return long zero address in the response. But we actually want the evm address returned.

This **PR** adds a check to see if a long zero address in passed in the request and will make the following verification:
If the address is long zero verify that there might be a canonical evm address.
If there exists a canonical evm address use it instead of the long zero.
Otherwise use the long zero.

Services uses similar approach -> 
```
if (relayerId == null) {
            result = evmTxProcessor.execute(
                    sender, receiver.canonicalAddress(), op.getGas(), op.getAmount(), callData, txnCtx.consensusTime());
}
```

 
This PR modifies:
`MirrorEvmTxProcessorImpl` adds a new method to determine the correct **to** address.
Added integration test for address function in `ContractCallDynamicCallsTest`

**Related issue(s)**:

Fixes #7857 

**Notes for reviewer**:

**Checklist**


hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallDynamicCallsTest.java
```diff
@@ -245,7 +258,8 @@ enum DynamicCallsContractFunctions {
                 },
                 null),
         GRANT_KYC_REVOKE_KYC_FUNGIBLE("grantKycRevokeKyc", new Object[] {FUNGIBLE_TOKEN_ADDRESS, SENDER_ALIAS}, null),
-        GRANT_KYC_REVOKE_KYC_NFT("grantKycRevokeKyc", new Object[] {NFT_ADDRESS, SENDER_ALIAS}, null);
+        GRANT_KYC_REVOKE_KYC_NFT("grantKycRevokeKyc", new Object[] {NFT_ADDRESS, SENDER_ALIAS}, null),
+        ADDRESS_THIS("getAddressThis", null, null);
```
IvanKavaldzhiev, line 262: We need more test cases like passing mirror address and alias address and checking the response is correct

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmTxProcessorImpl.java
```diff
@@ -148,4 +152,22 @@ protected MessageFrame buildInitialFrame(
...
         }
     }
+
+    private Address determineReceiverAddress(Address receiver) {
+        if (receiver == null || receiver.equals(Address.ZERO)) {
+            return receiver;
+        }
+        // if address is mirror address check for canonical evm address presence
+        if (aliasManager.resolveForEvm(receiver).equals(receiver)) {
+            Address canonical = tokenAccessor.canonicalAddress(receiver);
```
steven-sheehy, line 162: Why are we using `TokenAccessor` to lookup an account's canonical address instead of `AccountAccessor.canonicalAddress()`?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmTxProcessorImpl.java
```diff
@@ -148,4 +152,22 @@ protected MessageFrame buildInitialFrame(
                     .build();
         }
     }
+
+    private Address determineReceiverAddress(Address receiver) {
+        if (receiver == null || receiver.equals(Address.ZERO)) {
+            return receiver;
+        }
+        // if address is mirror address check for canonical evm address presence
+        if (aliasManager.resolveForEvm(receiver).equals(receiver)) {
```
steven-sheehy, line 161: Why can't we use the simpler and more efficient `aliasManager.isMirror(receiver)`?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmTxProcessorImpl.java
```diff
@@ -148,4 +152,22 @@ protected MessageFrame buildInitialFrame(
...
+            Address canonical = tokenAccessor.canonicalAddress(receiver);
+            if (null == canonical || Address.ZERO.equals(canonical)) {
+                return receiver;
+            } else {
+                return canonical;
+            }
+        } else {
+            // the address is a canonical evm address, return it
+            return receiver;
```
steven-sheehy, line 170: Invert this logic so we can avoid the extra nesting and return early if it's not a mirror.

```java
if (!aliasManager.isMirror(address)) {
    return reciever;
}

// rest of logic
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmTxProcessorImpl.java
```diff
@@ -148,4 +152,22 @@ protected MessageFrame buildInitialFrame(
                     .build();
         }
     }
+
+    private Address determineReceiverAddress(Address receiver) {
+        if (receiver == null || receiver.equals(Address.ZERO)) {
+            return receiver;
+        }
```
steven-sheehy, line 159: Should push this check into `aliasManager.isMirror()`

steven-sheehy (changes requested):

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallDynamicCallsTest.java
kselveliev: You are right. I've added a test where we pass a mirror address and we have also and evm address present to verify that we return evm address. I've also added a test where we pass mirror address but we do not have evm address and we verify that we return mirror address.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmTxProcessorImpl.java
```diff
@@ -148,4 +152,22 @@ protected MessageFrame buildInitialFrame(
...
+        // if address is mirror address check for canonical evm address presence
+        if (aliasManager.resolveForEvm(receiver).equals(receiver)) {
+            Address canonical = tokenAccessor.canonicalAddress(receiver);
+            if (null == canonical || Address.ZERO.equals(canonical)) {
```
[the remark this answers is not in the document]
kselveliev: Comment addressed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmTxProcessorImpl.java
kselveliev: From what i’ve been testing
` accountAccessor.canonicalAddress(0x00000000000000000000000000000000000004e7)` will still
return 0x00000000000000000000000000000000000004e7 even if there is also an evm address.

it will first execute this check internally 
```
if (aliases.isInUse(addressOrAlias)) {
    return addressOrAlias;
} 
```
this will result into true and return 0x00000000000000000000000000000000000004e7 when we 
want the evm address. This is why I avoided it.
 Could be observed in this test -> testGetAddressThisWithLongZeroRecipientThatHasEvmAlias

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmTxProcessorImpl.java
kselveliev: Comment addressed. Added this if in the first check.
```
 if (receiver == null || receiver.equals(Address.ZERO) || !aliasManager.isMirror(receiver)) {
            return receiver;
}
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmTxProcessorImpl.java
kselveliev: Got a bit scared to do it since we will need to override isMirror method from HederaEvmContractAliases. And I don't know how this might affect other code. I can try to do it if you want.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmTxProcessorImpl.java
kselveliev: I think we can. If we add `!aliasManager.isMirror(receiver)` in the first if we can even omit this if check.

IvanKavaldzhiev (approved):
LGTM

PR #8006, opened 2024-04-01
Add an importer duration metric

### LINKED ISSUE #8004: Add an overall importer duration metric
### Problem

The importer has a number of duration and latency metrics but would benefit from a metric that tracks the overall time to process a particular stream file.

### Solution

Add a `hedera.mirror.importer.duration` metric that tracks the time from listing the file until it's committed to the database.

### Alternatives

_No response_

steven-sheehy:
**Description**:

* Add a `hedera.mirror.importer.duration` metric
* Change `record_file.load_start` and `record_file.load_end` from seconds to milliseconds granularity
* Change record batch `queueCapacity` from `1` to `10`
* Change record batch `frequency` from `100ms` to `20ms`
* Update `HAPI Compatibility Matrix` and add an item to the feature checklist

**Related issue(s)**:

Fixes #8004

**Notes for reviewer**:

It would take too long to migrate existing load start/end to milliseconds and not provide any additional data.

**Checklist**


hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/reader/record/ProtoRecordFileReader.java
```diff
@@ -55,7 +54,7 @@ public class ProtoRecordFileReader implements RecordFileReader {
     @Override
     public RecordFile read(StreamFileData streamFileData) {
         var filename = streamFileData.getFilename();
-        var loadStart = Instant.now().getEpochSecond();
+        var loadStart = streamFileData.getStreamFilename().getTimestamp();
```
steven-sheehy, line 57: Instead of using now, use a timestamp from streamfilename. This ensures we track the earliest we've become aware of this stream file.

jascks: Ah, I was going around in circles on `loadStart`'s value. The duration starts at this _awareness_, rather than when the importer actually processes the file.

jascks (approved):
LGTM

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/reader/record/ProtoRecordFileReader.java
xin-hedera: however, during catch up or even the importer is slightly lagged behind, the duration will be artificially inflated to the point it's not useful at all, right?

jnels124: I think it is true there is inflation if there is waiting on the blocking queue. Becomes potentially more inflated with queue capacity > 1 because we can download multiple files that may be sitting there waiting exponentially longer in worst case. Think it is still useful to see though.

jnels124 (approved):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/reader/record/ProtoRecordFileReader.java
steven-sheehy: It's debatable. The same thing can be said by starting it at de-serialization like before since we now have a queue capacity of 10. The notifier buffer will skew the results and cause it to increase just the same as the download buffer will.

I think the steady state is most important to measure. We want to track the time a stream file interacts with the importer and when it's listed in s3 is when we first are aware of it. If we only track once it's de-serialized or verified then we're missing parts of the stack that could have inefficiencies. We can use additional metrics to subtract from this new overall metric if we want to account for historical lag.

edwin-greene (approved):
Looks good

xin-hedera (approved):
LGTM

Issue #7822, opened 2024-02-28
Verify mirror node and services are in sync


bilyana-gospodinova:
Check that the copied code from services that is located under "com.hedera.services" is in sync with the latest code in the services repo.

bilyana-gospodinova:
The results are described in: https://www.notion.so/limechain/Verify-mirror-node-and-services-are-in-sync-c1a9752154d6440a9302f97170ba71fe?pvs=4

PR #8021, opened 2024-04-04
Remove exception in TxnResourceUsageEstimator - usageGiven

zhpetkov:
**Description**:
Remove throws an exception in `TxnResourceUsageEstimator` - `usageGiven` and make needed changes because this affects all classes that implement it. The changes are part of verify the mirror node and services are in sync

**Related issue(s)**: #7822

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-web3/src/main/java/com/hedera/services/fees/calculation/TxnResourceUsageEstimator.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/services/fees/calculation/TxnResourceUsageEstimator.java
@@ -42,7 +42,6 @@ public interface TxnResourceUsageEstimator {
      * @param txn      the txn in question
      * @param sigUsage the signature usage
      * @return the estimated resource usage
-     * @throws Exception            if the txn is malformed
      * @throws NullPointerException or analogous if the estimator does not apply to the txn
      */
     @SuppressWarnings("java:S112")
--- a/hedera-mirror-web3/src/main/java/com/hedera/services/fees/calculation/TxnResourceUsageEstimator.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/services/fees/calculation/TxnResourceUsageEstimator.java
@@ -46,5 +46,5 @@ public interface TxnResourceUsageEstimator {
      * @throws NullPointerException or analogous if the estimator does not apply to the txn
      */
     @SuppressWarnings("java:S112")
-    FeeData usageGiven(TransactionBody txn, SigValueObj sigUsage) throws Exception;
+    FeeData usageGiven(TransactionBody txn, SigValueObj sigUsage);
 }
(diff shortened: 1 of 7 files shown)


hedera-mirror-web3/src/main/java/com/hedera/services/fees/calculation/TxnResourceUsageEstimator.java
```diff
@@ -46,5 +46,5 @@ public interface TxnResourceUsageEstimator {
      * @throws NullPointerException or analogous if the estimator does not apply to the txn
      */
     @SuppressWarnings("java:S112")
-    FeeData usageGiven(TransactionBody txn, SigValueObj sigUsage) throws Exception;
```
bilyana-gospodinova, line 49: The javadoc that says an `Exception` can be thrown from this method needs to be updated. 

The point for this task was to verify that there are no differences in the exception handling between services and the mirror node - for both checked and unchecked exceptions. What happens if a runtime exception is thrown (it is mentioned that a `NullPointerException` can be thrown)? I don't see a test in the mirror node for that case. Please add accordingly. 

Can you please clarify if you checked for differences in the code flow in case of an exception - in both services and the mirror node? Is there a case when the mirror node returns a different response from the one in services?

FYI - I found this PR in services when this method is modified to not throw an exception - https://github.com/hashgraph/hedera-services/pull/9139. It might help and give some hints from where to start in case you haven't made the above mentioned checks.

bilyana-gospodinova (changes requested):

hedera-mirror-web3/src/test/java/com/hedera/services/fees/calculation/UsageBasedFeeCalculatorTest.java
```diff
@@ -212,15 +212,12 @@ void estimatesFutureGasPriceInTinybars() {
     }
 
     @Test
-    void failsWithIseGivenApplicableButUnusableCalculator() {
+    void failsWithNpeGivenApplicableButUnusableCalculator() {
```
bilyana-gospodinova, line 215: Please add a test for the case when `UsageBasedFeeCalculator#getTxnUsageEstimator` throws a `NoSuchElementException` and verify that the behavior is the same as in services.

hedera-mirror-web3/src/main/java/com/hedera/services/fees/calculation/TxnResourceUsageEstimator.java
zhpetkov: The last commit update javadoc. I removed exceptions when the txn is malformed, because it is removed in hedera-services. I replaced `IllegalArgumentException` with `NullPointerException`. I checked exception handling in `UsageBasedFeeCalculator`, `TokenAssociateResourceUsage`, `TokenDeleteResourceUsage`, `TokenDissociateResourceUsage` (in mirror node there isn’t `usageGiven` calculate the fee for token associate transaction in modular code), `TokenUpdateResourceUsage`(in mirror node there isn’t `usageGiven` calculate the fee for token associate transaction in modular code).

steven-sheehy (approved):

bilyana-gospodinova (approved):
LGTM but there are 2 Sonar issues.

kselveliev (approved):

PR #8022, opened 2024-04-05
Fix balance opcode for 0.0.999 acceptance test

### LINKED ISSUE #8019: Balance opcode for "0.0.999" acceptance test fails
### Description

The `balance opcode to system account "0.0.999" address would return 0` fails against mainnet returns a non-zero balance.

### Steps to reproduce

1. Configure acceptance test to point at mainnet
2. `./gradlew :test:acceptance --info -Dcucumber.filter.tags=@equivalence`

### Additional context

```

AcceptanceTest > Cucumber STANDARD_OUT
    2024-04-03T21:29:04.347Z  INFO ForkJoinPool-1-worker-6 c.h.m.t.e.a.s.EquivalenceFeature Executing step: I execute balance opcode to system account "0.0.999" address would return 0 
    2024-04-03T21:29:04.389Z  INFO ForkJoinPool-1-worker-6 c.h.m.t.e.a.c.MirrorNodeClient Calling contract: class ContractCallRequest {
        block: latest
        data: f8b2cb4f00000000000000000000000000000000000000000000000000000000000003e7
        estimate: false
        from: 0000000000000000000000000000000000506aed
        gas: 15000000
        gasPrice: 100000000
        to: 0000000000000000000000000000000000506ac6
        value: 0
    } 
    2024-04-03T21:29:04.454Z  INFO reactor-http-nio-3 c.h.m.t.e.a.c.MirrorNodeClient POST https://mainnet.mirrornode.hedera.com/api/v1/contracts/call: 200 OK 
    2024-04-03T21:29:04.497Z ERROR ForkJoinPool-1-worker-6 c.h.m.t.e.a.s.EquivalenceFeature Scenario failure: 
    I execute balance opcode to system account "0.0.999" address would return 0 <- 
    expected: 899379576524
     but was: 0
     org.opentest4j.AssertionFailedError: 
    expected: 899379576524
     but was: 0
        at java.base/jdk.internal.reflect.DirectConstructorHandleAccessor.newInstance(DirectConstructorHandleAccessor.java:62)
        at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:502)
        at com.hedera.mirror.test.e2e.acceptance.steps.EquivalenceFeature.balanceOfAddress(EquivalenceFeature.java:126)
        at ✽.I execute balance opcode to system account "0.0.999" address would return 0(classpath:features/contract/equivalence.feature:12)


AcceptanceTest > Cucumber > in-equivalence tests > Validate in-equivalence system accounts for balance, extcodesize, extcodecopy, extcodehash > Examples > com.hedera.mirror.test.e2e.acceptance.AcceptanceTest.Example #1.1 FAILED
    org.opentest4j.AssertionFailedError: 
    expected: 899379576524
     but was: 0
        at java.base@21.0.2/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:502)
        at app//com.hedera.mirror.test.e2e.acceptance.steps.EquivalenceFeature.balanceOfAddress(EquivalenceFeature.java:126)
        at ✽.I execute balance opcode to system account "0.0.999" address would return 0(classpath:features/contract/equivalence.feature:12)
```

### Hedera network

mainnet

### Version

v0.100.0

### Operating system

None

zhpetkov:
**Description**:
Add logic in `EquivalenceFeature` that expects `balance` to be greater than 0 when `accountId` is greater than 751 and checks account existence by `accountId`. 

**Related issue(s)**:

Fixes #8019 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/EquivalenceFeature.java
+++ b/hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/EquivalenceFeature.java
@@ -27,7 +27,6 @@ import static com.hedera.mirror.test.e2e.acceptance.steps.EquivalenceFeature.Con
 import static com.hedera.mirror.test.e2e.acceptance.steps.EquivalenceFeature.ContractMethods.GET_CODE_SIZE;
 import static com.hedera.mirror.test.e2e.acceptance.util.TestUtil.asAddress;
 import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
-import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import com.esaulpaugh.headlong.abi.TupleType;
 import com.hedera.hashgraph.sdk.AccountId;
@@ -117,19 +116,14 @@ public class EquivalenceFeature extends AbstractFeature {
         }
     }
 
-    @Then("I execute balance opcode to system account {string} address with id 1 to 750 would return 0")
+    @Then("I execute balance opcode to system account {string} address would return 0")
     public void balanceOfAddress(String address) {
         var nodeType = acceptanceTestProperties.getNodeType();
         final var accountId = new AccountId(extractAccountNumber(address));
         var data = encodeData(EQUIVALENCE_CALL, GET_BALANCE, asAddress(accountId));
         var functionResult =
                 callContract(nodeType, StringUtils.EMPTY, EQUIVALENCE_CALL, GET_BALANCE, data, BIG_INTEGER_TUPLE);
-
-        if (extractAccountNumber(address) > 750 && accountExists(accountId)) {
-            assertTrue(functionResult.getResultAsNumber().compareTo(BigInteger.ZERO) >= 0);
-        } else {
-            assertThat(functionResult.getResultAsNumber()).isEqualTo(BigInteger.ZERO);
-        }
+        assertThat(BigInteger.ZERO).isEqualTo(functionResult.getResultAsNumber());
     }
 
     @Then("I execute balance opcode against a contract with balance")
--- a/hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/EquivalenceFeature.java
+++ b/hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/EquivalenceFeature.java
@@ -117,7 +117,7 @@ public class EquivalenceFeature extends AbstractFeature {
         }
     }
 
-    @Then("I execute balance opcode to system account {string} address would return 0")
+    @Then("I execute balance opcode to system account {string} address with id 1 to 750 would return 0")
     public void balanceOfAddress(String address) {
         var nodeType = acceptanceTestProperties.getNodeType();
         final var accountId = new AccountId(extractAccountNumber(address));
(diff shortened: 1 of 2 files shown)


hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/EquivalenceFeature.java
```diff
@@ -123,7 +124,16 @@ public void balanceOfAddress(String address) {
         var data = encodeData(EQUIVALENCE_CALL, GET_BALANCE, asAddress(accountId));
         var functionResult =
                 callContract(nodeType, StringUtils.EMPTY, EQUIVALENCE_CALL, GET_BALANCE, data, BIG_INTEGER_TUPLE);
-        assertThat(BigInteger.ZERO).isEqualTo(functionResult.getResultAsNumber());
+
+        if (extractAccountNumber(address) < 751) {
+            assertThat(functionResult.getResultAsNumber()).isEqualTo(BigInteger.ZERO);
+        } else {
+            if (isAccountExists(accountId)) {
+                assertTrue(functionResult.getResultAsNumber().compareTo(BigInteger.ZERO) > 0);
```
yiliev0, line 132: There could be a case where the account exists(i.e. 0.0.999) but it has 0 hbar balance. In this case I think this assert might fail. Check for example previewnet -> https://hashscan.io/previewnet/account/0.0.999?ph=1&pt=1

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/EquivalenceFeature.java
```diff
@@ -195,4 +205,13 @@ enum ContractMethods implements SelectorInterface {
 
         private final String selector;
     }
+
+    private boolean isAccountExists(final AccountId accountId) {
```
kselveliev, line 209: Prefer `accountExists` instead of current name.

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/EquivalenceFeature.java
zhpetkov: I tried with this case and the test failed again. The problem is fixed only for mainnet because the balance of the account with ID 999 is bigger than 0 there. Maybe we have to remove the test for 999 because if its balance changes this will fail the test again in the future. I can send hbars to the account with ID 999 in Previewnet.

kselveliev: If the account exists and it's balance is 0 this is a valid case right? So can't we just do 
`assertTrue(functionResult.getResultAsNumber().compareTo(BigInteger.ZERO) >= 0);` and we should be good?

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/EquivalenceFeature.java
```diff
@@ -116,14 +117,19 @@ public void selfDestructAndSetBeneficiary(String beneficiary) {
...
         var data = encodeData(EQUIVALENCE_CALL, GET_BALANCE, asAddress(accountId));
         var functionResult =
                 callContract(nodeType, StringUtils.EMPTY, EQUIVALENCE_CALL, GET_BALANCE, data, BIG_INTEGER_TUPLE);
-        assertThat(BigInteger.ZERO).isEqualTo(functionResult.getResultAsNumber());
+
+        if (extractAccountNumber(address) > 750 && accountExists(accountId)) {
+            assertTrue(functionResult.getResultAsNumber().compareTo(BigInteger.ZERO) >= 0);
+        } else {
+            assertThat(functionResult.getResultAsNumber()).isEqualTo(BigInteger.ZERO);
+        }
```
steven-sheehy, line 132: Is this necessary to check? We already have an equivalence test `I execute balance opcode against a contract with balance` to verify balance opcode works for regular account/contract. Can't we just remove `0.0.999` from this test?

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/EquivalenceFeature.java
```diff
@@ -195,4 +195,13 @@ enum ContractMethods implements SelectorInterface {
 
         private final String selector;
     }
+
```
kselveliev, line 198: With the latest change this method becomes unneeded and can be removed

kselveliev (approved):

steven-sheehy (approved):
LGTM

PR #8026, opened 2024-04-05
Support EVM address and alias lookup in NFT allowance REST API

### LINKED ISSUE #7858: Nft Allowance Controller - Input Validation
### Problem

Validate inputs for the NFT allowance request

### Solution

-  Handle evm address and alias Input.
-  Check for accountId existence.
-  get systemShard from properties.
-  Maintaining the next link param order to be the same as the original link.

### Alternatives

_No response_

mgoelswirlds:
**Description**:
This PR add evm address and alias account Id validation.

This PR modifies 
- Add evm address and alias account ID validation
- Account ID existence check in an entityService
- Getting system shard value from properties and validating the input shard value against it.
-  Adding caching to the repository existsById api
-  Fixing flaky mapping tests

**Related issue(s)**:

Fixes #7858

**Notes for reviewer**:
There is some spotless reformatting due to merge from main which can be ignored.

**Checklist**


hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/EntityIdParameter.java
```diff
@@ -16,20 +16,173 @@
...
+import static com.hedera.mirror.restjava.common.Constants.HEX_PREFIX;
 
 import com.hedera.mirror.common.domain.entity.EntityId;
+import java.util.Arrays;
+import java.util.Objects;
+import java.util.regex.Pattern;
+import org.apache.commons.codec.DecoderException;
+import org.apache.commons.codec.binary.Hex;
+import org.apache.commons.lang3.ArrayUtils;
 import org.jooq.tools.StringUtils;
```
mgoelswirlds, line 30: Work to refactor this into 3 separate parameter classes, will be done in parallel, in a separate PR.
Functionality it will be the same.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/EntityServiceImpl.java
```diff
@@ -0,0 +1,65 @@
...
+@RequiredArgsConstructor
+public class EntityServiceImpl implements EntityService {
+
+    @Value("${hedera.mirror.restJava.shard}")
+    private final long systemShard;
+
+    private final EntityRepository entityRepository;
+
+    @Override
+    public long lookup(EntityIdParameter accountId) {
```
mgoelswirlds, line 37: Adding this lookup method here to be reused by other APIs in the fuure.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/EntityServiceImpl.java
```diff
@@ -0,0 +1,65 @@
...
+                throw new EntityNotFoundException("No account with a matching evm address found");
+            }
+            id = entityId.get();
+        } else if (accountId.type() == EntityIdType.EVMADDRESS && accountId.evmAddress() != null) {
+            var entityId = entityRepository.findByEvmAddress(accountId.evmAddress());
+            if (!entityId.isPresent()) {
+                throw new EntityNotFoundException("No account with a matching alias found");
+            }
+            id = entityId.get();
+        }
```
steven-sheehy, line 62: This should be a switch statement by enum. Checking if the num/alias/evmAddress is not null is redundant since that should be guaranteed by `EntityIdParameter.valueOf()`.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/EntityServiceImpl.java
```diff
@@ -0,0 +1,65 @@
...
+        return id;
```
steven-sheehy, line 63: Instead of returning zero, we should throw an error if one of these three types did not validate.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/EntityServiceImpl.java
```diff
@@ -0,0 +1,65 @@
...
+        if (accountId.type() == EntityIdType.NUM && accountId.num() != null) {
+            if (!entityRepository.existsById(accountId.num().getId())) {
+                throw new EntityNotFoundException("Id %s does not exist".formatted(accountId));
+            }
+            id = accountId.num().getId();
+        } else if (accountId.type() == EntityIdType.ALIAS && accountId.alias() != null) {
+            var entityId = entityRepository.findByAlias(accountId.alias());
+            if (!entityId.isPresent()) {
+                throw new EntityNotFoundException("No account with a matching evm address found");
+            }
```
steven-sheehy, line 54: Move this not found assertion outside each check so it doesn't repeat. Use generic message like "No account found for the given ID".

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/repository/EntityRepositoryTest.java
```diff
@@ -0,0 +1,64 @@
...
+@RequiredArgsConstructor
+public class EntityRepositoryTest extends RestJavaIntegrationTest {
+
+    private final EntityRepository entityRepository;
+
+    @Test
+    void exists() {
+        var entity = domainBuilder.entity().persist();
+        var entityDeleted =
+                domainBuilder.entity().customize((b) -> b.deleted(true)).persist();
```
steven-sheehy, line 34: For each existence query, add a test with deleted=null

steven-sheehy (changes requested):

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/repository/EntityRepository.java
```diff
@@ -0,0 +1,39 @@
...
+
+    @Cacheable(cacheNames = CACHE_NAME, cacheManager = ENTITY_CACHE, unless = "#result == null")
+    @Query(value = "select EXISTS(SELECT 1 from entity where id = ?1 and deleted <> true)", nativeQuery = true)
+    boolean existsById(Long entityId);
+
+    @Query(value = "select id from entity where alias = ?1 and deleted <> true", nativeQuery = true)
+    Optional<Long> findByAlias(byte[] alias);
+
+    @Query(value = "select id from entity where evm_address = ?1 and deleted <> true", nativeQuery = true)
+    Optional<Long> findByEvmAddress(byte[] evmAddress);
```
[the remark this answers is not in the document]
steven-sheehy: Actually, think caffeine doesn't work with arrays. Would have to wrap it in something so maybe can defer.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/EntityIdParameter.java
```diff
@@ -16,20 +16,173 @@
...
-public record EntityIdParameter(EntityId value) {
-    public static final EntityIdParameter EMPTY = new EntityIdParameter(null);
+public record EntityIdParameter(
+        EntityId num, byte[] evmAddress, byte[] alias, Long shard, Long realm, EntityIdType type) {
+    public static final EntityIdParameter EMPTY = new EntityIdParameter(null, null, null, 0L, 0L, null);
+
+    public static final String ACCOUNT_ALIAS_REGEX = "^((\\d{1,5})\\.)?((\\d{1,5})\\.)?([A-Z2-7]+)$";
+
+    public static final String ENTITY_ID_REGEX = "^((\\d{1,5})\\.)?((\\d{1,5})\\.)?(\\d{1,10})$";
+    public static final String EVM_ADDRESS_REGEX = "^(((0x)?|((\\d{1,5})\\.)?((\\d{1,5})\\.)?)([A-Fa-f0-9]{40}))$";
```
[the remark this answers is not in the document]
mgoelswirlds: This wasn't supported. Added a testcase. Handling `0.0xabcdef`, or `0.0.0xabcdef` now.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/EntityIdParameter.java
```diff
@@ -16,20 +16,173 @@
...
+                && type == that.type;
+    }
+
+    @Override
+    public int hashCode() {
+        int result = Objects.hash(num, type);
+        result = 31 * result + Arrays.hashCode(evmAddress);
+        result = 31 * result + Arrays.hashCode(alias);
+        return result;
+    }
```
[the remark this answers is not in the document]
mgoelswirlds: For arrays SonarLint complains about them not being considered. I am not adding it to a map or set anywhere.I will suppress it.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/EntityIdAliasParameter.java
```diff
@@ -0,0 +1,50 @@
...
+            // This gets the shard and realm value
+            realm = Long.parseLong(aliasMatcher.group(4));
+            shard = Long.parseLong(aliasMatcher.group(2));
+
+        } else if (aliasMatcher.group(1) != null) {
+            // This gets the realm value and shard will be null
+            realm = Long.parseLong(aliasMatcher.group(2));
+        }
+
+        return new EntityIdAliasParameter(new Base32().decode(aliasMatcher.group(5)), shard, realm);
```
xin-hedera, line 48: I don't see how to configure apache commons `Base32` to omit padding, our alias format is rfc4648 without padding (https://hips.hedera.com/hip/hip-32).

can try guava `BaseEncoding.base32().omitPadding()`

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/repository/EntityRepository.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.Optional;
+import org.springframework.cache.annotation.Cacheable;
+import org.springframework.data.jpa.repository.Query;
+import org.springframework.data.repository.CrudRepository;
+
+public interface EntityRepository extends CrudRepository<Entity, Long> {
+
+    @Cacheable(cacheNames = CACHE_NAME, cacheManager = ENTITY_CACHE, unless = "#result == null")
+    @Query(value = "select EXISTS(SELECT 1 from entity where id = ?1 and deleted <> true)", nativeQuery = true)
+    boolean existsById(Long entityId);
```
xin-hedera, line 32: we can't cache any of these since in rest java we don't have a way to evict the cache entry when `deleted` becomes `true` from `false`

xin-hedera (changes requested):

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/repository/EntityRepository.java
steven-sheehy: The cache has an expireAfterWrite=60s to purge older entries so that they're eventually consistent and show the correct deleted flag

xin-hedera: Still think 60s is too long, enough for an EVM address to be claimed by a different account id.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/common/EntityIdParameterTest.java
```diff
@@ -0,0 +1,134 @@
...
+    @ParameterizedTest
+    @ValueSource(strings = {"0.0.4294967296", "32768.65536.4294967296"})
+    @DisplayName("EntityId parse from string tests, negative cases for ID having valid format")
+    void testInvalidEntity(String input) {
+        assertThrows(InvalidEntityException.class, () -> EntityIdParameter.valueOf(input));
+    }
+
+    @Test
+    @DisplayName("EntityId parse from string tests")
+    void entityParseFromString() {
```
steven-sheehy, line 67: Prefer this test method be split into separate `valueOfId`, `valueOfAlias`, and `valueOfEvmAddress` methods each with parameterized inputs.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/EntityServiceImpl.java
```diff
@@ -0,0 +1,63 @@
...
+            throw new IllegalArgumentException("ID %s has invalid shard".formatted(accountId));
+        }
+
+        if (accountId.realm() != 0) {
+            throw new IllegalArgumentException("ID %s has invalid realm".formatted(accountId));
+        }
+
+        var id =
+                switch (accountId) {
+                    case EntityIdNumParameter(EntityId num) -> entityRepository.existsById(num.getId()) ? Optional.of(num.getId()) : Optional.empty();
```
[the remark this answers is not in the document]
mgoelswirlds: Since this  just returns a boolean, map wouldn't work and would prefer to leave this as is.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/EntityServiceImpl.java
```diff
@@ -0,0 +1,63 @@
...
+                    case EntityIdAliasParameter(byte[] alias, Long shard, Long realm) -> entityRepository.findByAlias(alias);
+                    case EntityIdEvmAddressParameter(byte[] evmAddress, Long shard, Long realm) -> entityRepository.findByEvmAddress(evmAddress);
```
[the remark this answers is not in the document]
mgoelswirlds: Doesn't work, all params are mandatory for the object.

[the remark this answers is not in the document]
steven-sheehy: Ok, then just `case EntityIdEvmAddressParameter p -> entityRepository.findByAlias(p.alias()`, etc would be preferable since it doesn't repeat all parameters.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/EntityIdEvmAddressParameter.java
```diff
@@ -0,0 +1,52 @@
...
+        Long realm = 0L;
+
+        if (evmMatcher.group(3) != null) {
+            // This gets the realm value
+            realm = Long.parseLong(evmMatcher.group(4));
+            shard = Long.parseLong(evmMatcher.group(2));
+
+        } else if (evmMatcher.group(1) != null) {
+            realm = Long.parseLong(evmMatcher.group(2));
+        }
```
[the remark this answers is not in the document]
mgoelswirlds: I cannot use this. I am checking for a particular group matching here. Just checking the count wouldn't be accurate.
Count `Returns the number of capturing groups in this matcher's pattern.`, in this case the `groupCount()` will always return 6 (which is the number of groups in the pattern) if the pattern matches overall. So before extracting the value for say realm , I need to ensure that group match exists there.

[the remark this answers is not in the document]
steven-sheehy: The point still remains that generating a string just to discard it is inefficient. So instead of check group(3) should check group(4) and store it in a variable for later conversion to long. Or can use `matcher.start(4) >= 0` which does the same check without the string generation.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/common/EntityIdParameterTest.java
```diff
@@ -0,0 +1,109 @@
...
+            "1.2.0000000100000000000000020000000000000003,0000000100000000000000020000000000000003",
+            "0x00007fff000000000000ffff00000000ffffffff,00007fff000000000000ffff00000000ffffffff",
+            "0.0.000000000000000000000000000000000186Fb1b,000000000000000000000000000000000186Fb1b",
+            "0.0x000000000000000000000000000000000186Fb1b,000000000000000000000000000000000186Fb1b",
+            "0.0.0x000000000000000000000000000000000186Fb1b,000000000000000000000000000000000186Fb1b",
+            "0.000000000000000000000000000000000186Fb1b,000000000000000000000000000000000186Fb1b",
+            "000000000000000000000000000000000186Fb1b,000000000000000000000000000000000186Fb1b",
+            "0x000000000000000000000000000000000186Fb1b,000000000000000000000000000000000186Fb1b"
+    })
+    void valueOfEvmAddress(String givenEvmAddress,String expectedEvmAddress) {
```
steven-sheehy, line 93: Add expected shard and realm parameters and verify them against actual.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/common/EntityIdParameterTest.java
```diff
@@ -0,0 +1,109 @@
...
+                        EntityIdParameter.valueOf(givenEvmAddress))
+                        .evmAddress());
+    }
+
+    @ParameterizedTest
+    @CsvSource({"AABBCC22,AABBCC22",
+            "0.AABBCC22,AABBCC22",
+            "0.1.AABBCC22,AABBCC22"
+    })
+    void valueOfAlias(String givenAlias, String expectedAlias) {
```
steven-sheehy, line 105: Add expected shard and realm parameters and verify them against actual.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/controller/AllowancesControllerIntegrationTest.java
```diff
@@ -72,7 +74,34 @@ void setup() {
...
+        Collection<NftAllowance> collection = List.of(allowance1, allowance2);
+
+        var result = restClient
+                .get()
+                .uri("", allowance1.getOwner())
+                .accept(MediaType.ALL)
+                .retrieve()
+                .body(NftAllowancesResponse.class);
+        assertThat(result.getAllowances()).isEqualTo(mapper.map(collection));
+        assertNull(result.getLinks().getNext());
```
steven-sheehy, line 95: Always use assertj. Check elsewhere.

```suggestion
        assertThat(result.getLinks().getNext()).isNull();
```

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/controller/AllowancesControllerIntegrationTest.java
```diff
@@ -72,7 +74,34 @@ void setup() {
...
+        var allowance2 = domainBuilder
+                .nftAllowance()
+                .customize(nfta -> nfta.owner(allowance1.getOwner()))
+                .persist();
+        Collection<NftAllowance> collection = List.of(allowance1, allowance2);
+
+        var result = restClient
+                .get()
+                .uri("", allowance1.getOwner())
+                .accept(MediaType.ALL)
```
steven-sheehy, line 91: Remove accept all header from all tests since in setup we already set an appropriate accept header.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/controller/AllowancesControllerIntegrationTest.java
```diff
@@ -72,7 +74,34 @@ void setup() {
 
     @Test
     void successWithNoQueryParamsShardRealmNumAccountId() {
```
steven-sheehy, line 76: Need a parameterized test for 404 not found with various ids

steven-sheehy, line 76: Since `AllowancesController` can in the future contain other allowances APIs, should prefix all tests with `nftAllowances`. Can also simplify names to not mention things that aren't present.

```suggestion
    void nftAllowancesEntityId() {
```

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/service/EntityServiceTest.java
```diff
@@ -0,0 +1,111 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+@RequiredArgsConstructor
+public class EntityServiceTest extends RestJavaIntegrationTest {
```
steven-sheehy, line 38: Change all test classes to package private.

```suggestion
class EntityServiceTest extends RestJavaIntegrationTest {
```

steven-sheehy (changes requested):

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/EntityIdNumParameter.java
```diff
@@ -0,0 +1,62 @@
...
+    public static EntityIdNumParameter valueOf(String id) {
+
+        var matcher = ENTITY_ID_PATTERN.matcher(id);
+
+        if (!matcher.matches()) {
+            return null;
+        }
+
+        long realm = 0;
+        long shard = 0;
```
steven-sheehy, line 46: Actually shard should default to `RestJavaProperties.getShard()` instead of zero. Not sure the best way to do this though. Same comment for other classes.

docs/configuration.md
```diff
@@ -587,15 +587,17 @@ to configure the application.
...
-| `hedera.mirror.restJava.db.sslMode`          | DISABLE               | The SSL level. Accepts either DISABLE, ALLOW, PREFER, REQUIRE, VERIFY_CA or VERIFY_FULL. |
-| `hedera.mirror.restJava.db.statementTimeout` | 10000                 | The number of milliseconds to wait before timing out a query statement                   |
-| `hedera.mirror.restJava.db.username`         | mirror_rest_java      | The username used to connect to the database                                             |
+| Name                                         | Default                                              | Description                                                                              |
+|----------------------------------------------|------------------------------------------------------|------------------------------------------------------------------------------------------|
+| `hedera.mirror.restJava.cache.entity`        | maximumSize=100000,expireAfterWrite=10s,recordStats  | The Caffeine cache specification for the entity.                                         |
```
xin-hedera, line 592: missing `hedera.mirror.restJava.cache.enabled`

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/EntityIdNumParameter.java
mgoelswirlds: Found a way to get the bean in a non spring class.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/controller/AllowancesControllerIntegrationTest.java
```diff
@@ -263,12 +334,34 @@ void failTokenIdPresentWithoutAccount() {
...
+                    "AABBCC22"
+            })
+    void nftAllowancesNotFoundException(String accountId) {
+        RestClient restClient = RestClient.create();
+
+        // Performing the GET operation
+        assertThrows(
+                HttpClientErrorException.NotFound.class,
+                () -> restClient.get().uri(callUri, accountId).retrieve().body(NftAllowancesResponse.class));
+    }
```
steven-sheehy, line 366: It's not sufficient to just check response code, we also need to verify error response body follows our expected format. Use assertj to verify additional exception details.

```suggestion
        assertThatThrownBy(() -> restClient.get()
                    .uri(uriParams, 1000)
                    .retrieve()
                    .body(NftAllowancesResponse.class)
                )
                .isInstanceOf(HttpClientErrorException.NotFound.class)
                .hasMessageContaining("some message")
                .asInstanceOf(InstanceOfAssertFactories.type(HttpClientErrorException.NotFound.class))
                .extracting(r -> r.getResponseBodyAs(Error.class).getStatus().getMessages().get(0))
                .returns("foo", ErrorStatusMessagesInner::getData)
                .returns("bar", ErrorStatusMessagesInner::getDetail)
                .returns("baz", ErrorStatusMessagesInner::getMessage);
```

We can put this in a helper method and use it to verify the other error cases.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/SpringApplicationContext.java
```diff
@@ -0,0 +1,25 @@
+package com.hedera.mirror.restjava.common;
+
+import org.springframework.beans.BeansException;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.ApplicationContextAware;
+import org.springframework.stereotype.Component;
+
+@Component
```
steven-sheehy, line 8: Missing license header and we prefer to use `@Named`

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/SpringApplicationContext.java
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    @Override
+    public void setApplicationContext(ApplicationContext context) throws BeansException {
+        setContext(context);
+    }
+
+    private static synchronized void setContext(ApplicationContext context) {
+        SpringApplicationContext.context = context;
+    }
```
steven-sheehy, line 24: nit: Could use a static final AtomicReference.

```suggestion
       CONTEXT.set(context);
    }
```

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/repository/EntityRepositoryTest.java
```diff
@@ -0,0 +1,74 @@
...
+package com.hedera.mirror.restjava.repository;
+
+import com.hedera.mirror.restjava.RestJavaIntegrationTest;
+import lombok.RequiredArgsConstructor;
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@RequiredArgsConstructor
+public class EntityRepositoryTest extends RestJavaIntegrationTest {
```
steven-sheehy, line 26:
```suggestion
class EntityRepositoryTest extends RestJavaIntegrationTest {
```

steven-sheehy (changes requested):

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/controller/AllowancesControllerIntegrationTest.java
mgoelswirlds: The error message extracted here is not the expected one.It is empty. I am working on getting this to work.

steven-sheehy: It may not know how to resolve the type from the exception. Might have to register a `defaultStatusHandler()` on the `RestClientBuilder` with proper message converters.

docs/configuration.md
```diff
@@ -587,15 +587,18 @@ to configure the application.
...
-| `hedera.mirror.restJava.db.statementTimeout` | 10000                 | The number of milliseconds to wait before timing out a query statement                   |
-| `hedera.mirror.restJava.db.username`         | mirror_rest_java      | The username used to connect to the database                                             |
+| Name                                         | Default                                               | Description                                                                              |
+|----------------------------------------------|-------------------------------------------------------|------------------------------------------------------------------------------------------|
+| `hedera.mirror.restJava.cache.entity`        | maximumSize=100000,expireAfterWrite=10s,recordStats   | The Caffeine cache specification for the entity.                                         |
+| `hedera.mirror.restJava.cache.enabled`       | true                                                  | Whether caching should be enabled at all.                                                |
```
xin-hedera, line 593: nit, order, this should be before `cache.entity`

steven-sheehy:
Should suppress the 2 override equals for arrays sonar warning.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/controller/AllowancesControllerIntegrationTest.java
mgoelswirlds: Currently only this method is returning an Error type in the service layer and not through the controller so I am checking for the detailed message only for the EntityNotFound exception.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/controller/AllowancesControllerIntegrationTest.java
```diff
@@ -239,36 +313,94 @@ void successWithEmptyNextLink() {
...
 
         // Performing the GET operation
-        assertThrows(
-                HttpClientErrorException.MethodNotAllowed.class,
-                () -> restClient.post().uri(callUri, "0.0.1000").retrieve().body(NftAllowancesResponse.class));
+        validateErrorMessage(uriParams, EntityId.of(allowance1.getOwner()).toString(), "Invalid range operator ne. This operator is not supported",HttpClientErrorException.BadRequest.class, "Bad Request");
     }
+
+    private void validateErrorMessage(String uriParams, String accountId, String expected, Class<? extends HttpClientErrorException> clazz, String description) {
+        assertThatThrownBy(() -> restClient.get().uri(uriParams, accountId).retrieve().body(NftAllowancesResponse.class)
```
steven-sheehy, line 395: Prefer to decouple test invocation from error validation so it's more reusable.

```suggestion
    private void validateError(ThrowingCallable callable, Class<? extends HttpClientErrorException> clazz, String message, String description) {
        assertThatThrownBy(callable)
```

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/controller/AllowancesControllerIntegrationTest.java
```diff
@@ -239,36 +313,94 @@ void successWithEmptyNextLink() {
...
-                .body(NftAllowancesResponse.class));
+                .body(NftAllowancesResponse.class)
+        )
+                .isInstanceOf(HttpClientErrorException.BadRequest.class)
+                .hasMessageContaining("Bad Request");
     }
 
     @Test
-    void failTokenIdPresentWithoutAccount() {
+    void nftAllowancesUnsupportedMediaException() {
```
steven-sheehy, line 333: Why are we not able to use the validateError helper method for these two to verify the structure of their response?

mgoelswirlds: That is because these 2 exceptions are generated by the controller either for the method or during parameter conversion and the Exception controller advice doesn't catch them so the Error type is not getting generated.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/controller/ExceptionControllerAdvice.java
```diff
@@ -43,33 +42,28 @@
...
     private ResponseEntity<Error> bindError(final BindException e) {
         return errorResponse(e.getMessage(), BAD_REQUEST);
     }
 
     @ExceptionHandler
-    @ResponseStatus(SERVICE_UNAVAILABLE)
     private ResponseEntity<Error> queryTimeout(final QueryTimeoutException e) {
         log.error("Query timed out: {}", e.getMessage());
-        return errorResponse(SERVICE_UNAVAILABLE.getReasonPhrase(), SERVICE_UNAVAILABLE);
+        return errorResponse(e.getMessage(), SERVICE_UNAVAILABLE);
```
steven-sheehy, line 66: Revert. We shouldn't expose 5xx error messages to clients.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/controller/GenericControllerAdvice.java
```diff
@@ -36,7 +35,6 @@
 public class GenericControllerAdvice extends ResponseEntityExceptionHandler {
 
     @ExceptionHandler
-    @ResponseStatus(INTERNAL_SERVER_ERROR)
     private ResponseEntity<Error> genericError(final Exception e) {
         log.error("Generic error: ", e);
         return errorResponse(e.getMessage(),INTERNAL_SERVER_ERROR);
```
steven-sheehy, line 40: Revert message change so we don't expose internal details for 5xx.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/controller/AllowancesControllerIntegrationTest.java
steven-sheehy: So that's a bug. `GenericControllerAdvice` should ensure all error responses use our json format. Can be follow up.

steven-sheehy (changes requested):

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/controller/AllowancesControllerIntegrationTest.java
mgoelswirlds: Will create a follow-up ticket.

steven-sheehy (approved):
LGTM

xin-hedera (approved):
LGTM

PR #8070, opened 2024-04-11
Optimize NftRepository queries in web3

### LINKED ISSUE #8064: Slow NFT related db queries cause web3 `/contract/call` to return 500
### Description

Users have reported for certain `/contract/call` web3 requests, they get 500 intermittently. The contract call loads NFT information and the corresponding db query times out frequently.

### Steps to reproduce

Run the `/contract/call` sample provided by the user.

### Additional context

_No response_

### Hedera network

mainnet

### Version

v0.101.0

### Operating system

None

xin-hedera:
**Description**:

This PR optimizes queries in web3 NftRepository

- Add a `btree_gist` index on `account_id, timestamp_range` and drop the timestamp_range index
- Change historical nft balance by account query to utilize the new index
- Change historical nft balance by account and token to utilize the new index
- Rewrite historical nft total supply query

**Related issue(s)**:

Fixes #8064 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-importer/src/main/resources/db/migration/v1/V1.96.0__optimize_nft_history_index.sql
+++ b/hedera-mirror-importer/src/main/resources/db/migration/v1/V1.96.0__optimize_nft_history_index.sql
@@ -1,0 +1,3 @@
+drop index if exists nft_history__timestamp_range;
+create index if not exists nft_history__account_timestamp_range
+  on nft_history using gist (account_id, timestamp_range);
--- a/hedera-mirror-importer/src/main/resources/db/migration/v2/V2.0.3__index_init.sql
+++ b/hedera-mirror-importer/src/main/resources/db/migration/v2/V2.0.3__index_init.sql
@@ -161,9 +161,10 @@ create index if not exists nft__allowance on nft (account_id, spender, token_id,
     where account_id is not null and spender is not null;
 
 -- nft history
+create index if not exists nft_history__account_timestamp_range
+    on nft_history using gist (account_id, timestamp_range);
 create index if not exists nft_history__token_serial_lower_timestamp
     on nft_history (token_id, serial_number, lower(timestamp_range));
-create index if not exists nft_history__timestamp_range on nft_history using gist (timestamp_range);
 
 -- nft_allowance
 alter table if exists nft_allowance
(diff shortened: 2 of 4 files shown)


hedera-mirror-importer/src/main/resources/db/migration/v1/V1.96.0__optimize_nft_history_index.sql
```diff
@@ -0,0 +1,3 @@
+drop index if exists nft_history__timestamp_range;
+create index if not exists nft_history__account_timestamp_range
+  on nft_history using gist (account_id, timestamp_range);
```
xin-hedera, line 3: `btree_gist` index gives us the best performance and it's much better than different btree indexes tested.

There may be concern if requiring `btree_gist` index will break other mirror node operators, I believe it should be fine as it's a trusted extension in default installation.

hedera-mirror-rosetta/test/db/db.go
```diff
@@ -91,41 +90,39 @@ func (d DbResource) GetGormDb() *gorm.DB {
...
-	name     string
-	username string
-	password string
+	endpoint      string
+	ownerPassword string
 }
 
 func (d dbParams) toDsn() string {
-	return fmt.Sprintf("postgres://%s:%s@%s/%s?sslmode=disable", d.username, d.password, d.endpoint, d.name)
+	return fmt.Sprintf("postgres://%s:%s@%s/%s?sslmode=disable", ownerUsername, d.ownerPassword, d.endpoint, dbName)
```
xin-hedera, line 98: resort to use `mirror_node` user in rosetta integration tests due to the cleanup script and deleting data in some test case setup.

ideally since now it's changed to use the same `init.sh` as other modules, we should really use `mirror_rosetta` which is readonly to run the integration tests. Will create a ticket as follow-up.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallTestSetup.java
```diff
@@ -2196,6 +2193,32 @@ private EntityId nftPersistHistorical(
                         .timestampRange(Range.openClosed(
                                 historicalBlock.lowerEndpoint() - 1, historicalBlock.upperEndpoint() + 1)))
                 .persist();
+
+        // nft table
+        domainBuilder
```
xin-hedera, line 2198: for any nft in the `nft_history` table, there must be the same nft by (token_id, serial) in `nft` table.

steven-sheehy (approved):
LGTM

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.96.0__optimize_nft_history_index.sql
steven-sheehy: Adding `breaking` label. Let's remember to mention in the release notes upgrading section.

edwin-greene (approved):
Looks good

Issue #8072, opened 2024-04-11
`contracts/call` errors out when transaction.to is set to null


quiet-node:
### Description

Currently,  when a request with `transaction.to=null` is sent to `contracts/call` endpoint, mirror-node throws an error with message `to field must not be empty`. `transaction.to` is set to null in the case of deploying a contract, and `contracts/call` should accept a transaction object with nullable `to` field. 

### Steps to reproduce

1. Run below cURL request
```bash
curl -X 'POST' \
  'https://testnet.mirrornode.hedera.com/api/v1/contracts/call' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "block": "latest",
  "data": "0x6080604052...",
  "estimate": false,
  "from": "0x9de0881b3110aA8cAD1dF3182B1eB6F14d1608a2",
  "gas": 15000000,
  "gasPrice": 100000000,
  "to": null,
  "value": 0
}'
```

2. Observe the output to see the error message like below
```bash
{"_status":{"messages":[{"message":"to field must not be empty","detail":"","data":""}]}}
```

### Hedera network

mainnet, testnet, previewnet, other

### Version

latest

### Operating system

None

PR #8100, opened 2024-04-15
Adapt contract deploy to work for eth call

bilyana-gospodinova:
**Description**:
The old validation of the "to" field on contract call made the requests to fail in the contract deploy scenario because the "to" field is empty in this case and only the estimate calls worked before. This PR makes the validation less restrictive and it is now applicable only for transfers.

Fixes https://github.com/hashgraph/hedera-mirror-node/issues/8072

### DIFF
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/ContractController.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/ContractController.java
@@ -94,7 +94,7 @@ class ContractController {
 
         /*When performing estimateGas with an empty "to" field, we set a default value of the zero address
         to avoid any potential NullPointerExceptions throughout the process.*/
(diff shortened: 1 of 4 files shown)


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/ContractController.java
```diff
@@ -94,7 +94,7 @@ private CallServiceParameters constructServiceParameters(ContractCallRequest req
 
         /*When performing estimateGas with an empty "to" field, we set a default value of the zero address
```
kselveliev, line 95: We need to slightly adjust the comment since it mentions `estimateGas` but we removed it now.

bilyana-gospodinova: Thanks. Updated.

steven-sheehy (approved):
LGTM

PR #8127, opened 2024-04-18
Add HIP-904 Frictionless Airdrops design

### LINKED ISSUE #8081: HIP-904: Frictionless Airdrops
### Problem

Mirror node needs to support [HIP-904](https://hips.hedera.com/hip/hip-904) Frictionless Aidrops.

### Solution

```[tasklist]
### Tasks
- [ ] #8082
- [ ] https://github.com/hashgraph/hedera-mirror-node/issues/8272
- [ ] https://github.com/hashgraph/hedera-mirror-node/issues/8273
- [ ] https://github.com/hashgraph/hedera-mirror-node/issues/8274
- [ ] https://github.com/hashgraph/hedera-mirror-node/issues/8833
- [ ] https://github.com/hashgraph/hedera-mirror-node/issues/8275
- [ ] https://github.com/hashgraph/hedera-mirror-node/issues/9033
- [ ] https://github.com/hashgraph/hedera-mirror-node/issues/8276
- [ ] https://github.com/hashgraph/hedera-mirror-node/issues/8282
- [ ] https://github.com/hashgraph/hedera-mirror-node/issues/8277
- [ ] https://github.com/hashgraph/hedera-mirror-node/issues/8278
- [ ] https://github.com/hashgraph/hedera-mirror-node/issues/9529
- [ ] https://github.com/hashgraph/hedera-mirror-node/issues/9378
- [ ] https://github.com/hashgraph/hedera-mirror-node/issues/9379
- [ ] https://github.com/hashgraph/hedera-mirror-node/issues/9530
```

### Alternatives

_No response_

### LINKED ISSUE #8082: HIP-904: Frictionless Airdrops Design
### Problem

We need to design how we will support [HIP-904](https://hips.hedera.com/hip/hip-904): Frictionless Airdrops.

### Solution

* Add a `docs/design/frictionless-airdrops.md` that outlines the impact to mirror node
* Create sub-tasks under the airdrop epic

### Alternatives

_No response_

edwin-greene:
**Description**:

Add the design doc for HIP-904 Friction-less Airdrops

**Related issue(s)**:

Fixes #8082

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/docs/design/frictionless-airdrops.md
+++ b/docs/design/frictionless-airdrops.md
@@ -1,0 +1,171 @@
+# HIP-336 Approval and Allowance
+
+## Purpose
+
+[HIP-904](https://hips.hedera.com/hip/hip-904) enables frictionless airdrops of both fungible and non-fungible tokens by removing the requirement to pre-associate tokens with the receiver’s account.
+
+## Goals
+
+- Ingest `TokenAirdrop` transactions and persist to the database
+- Ingest `PendingAirdrop` transactions and persist to the database
+- Ingest `TokenCancelAirdrop` and `TokenClaimAirdrop` transactions and remove entries from the database
+- Ingest `TokenReject` transactions and persist to the database
+- Expose Pending Airdrop information via the Java REST API
+
+## Non-Goals
+
+- Support for historical airdrop information
+- Enhance gRPC APIs with Pending Airdrop information
+- Enhance Web3 APIs with Pending Airdrop information
+
+## Architecture
+
+### Database
+
+#### Pending Airdrops
+
+```sql
+create table if not exists pending_airdrop
+(
+    account_id          bigint  not null,
+    amount              bigint
+    payer_account_id    bigint  not null,
+    serial_number       bigint
+    token_id            bigint  not null
+    primary key (account_id, payer_account_id, token_id)
+);
+
+create index if not exists pending__airdrop on pending_airdrop (account_id, payer_account_id);
+```
+
+### Importer
+
+#### Pending Airdrop Parsing
+
+When parsing pending airdrops,
+
+- Persist pending airdrops to the `pending_airdrop` table.
+- If a pending airdrop entry already exists, the new amount should be added to the existing amount.
+
+#### Domain
+
+- Add an `TokenAirdrop` domain object with the same fields as `CryptoTransfer`.
+- Add an `TokenCancelAirdrop` domain object which contains the fields `accountId`, `payerAccountId`, `tokenId`, and `serialNumber`.
+- Add an `TokenClaimAirdrop` domain object which contains the fields `accountId`, `payerAccountId`, `tokenId`, and `serialNumber`.
+- Add an `TokenReject` domain object which contains the fields `accountId`, `tokenId`, and `serialNumber`.
+- Add an `PendingAirdrop` domain object with the same fields as the schema.
+
+#### Entity Listener
+
+- Add `onCancelAirdrop` to handle deletes from the `pending_airdrop` table.
+- Add `onClaimAirdrop` to handle deletes from the `pending_airdrop` table.
+- Add `onPendingAirdrop` to handle inserts to the `pending_airdrop` table.
+- Add `onTokenReject` which will be similar to `onCryptoTransfer` except the balance will be zero.
+- Add `onTokenAirdrop` which will be similar to `onCryptoTransfer`.
+
+#### Transaction Handlers
+
+- Add `TokenAirdropTransactionHandler` which will be similar to `CryptoTransferTransactionHandler`
+- Add `TokenCancelAirdropTransactionHandler`
+- Add `TokenClaimAirdropTransactionHandler`
+- Add `TokenRejectTransactionHandler` which will be similar to `CryptoTransferTransactionHandler`
+- Add `PendingAirdropTransactionHandler`
+
+#### Batch Deletion
+
+- Add a batch deletion velocity template to be used by Claim airdrop and Cancel airdrop.
+- Add a `GenericDeleteQueryGenerator` to be used for batch deletion.
+
+### REST API
+
+- Add the new endpoints to Java REST.
+
+#### List of outstanding airdrops sent by senderIdOrEvmAddress which have not been claimed by recipients
+
+`/api/v1/accounts/{senderIdOrEvmAddress}/airdrops/outstanding`
+
+```json
+{
+  "airdrops": [
+    {
+      "amount": 333,
+      "receiver_id": "0.0.999",
+      "sender_id": "0.0.222",
+      "serial_number": null,
+      "token_id": "0.0.111"
+    },
+    {
+      "amount": 555,
+      "receiver_id": "0.0.999",
+      "sender_id": "0.0.222",
+      "serial_number": null,
+      "token_id": "0.0.444"
+    },
+    {
+      "amount": null,
+      "receiver_id": "0.0.999",
+      "sender_id": "0.0.222",
+      "serial_number": 888,
+      "token_id": "0.0.666"
+    }
+  ],
+  "links": {
+    "next": null
+  }
+}
+```
+
+Optional Filters
+
+- `limit` - The maximum number of airdrops to return in the response. Defaults to `25` with a max of `100`.
+- `order` - The direction to sort the items in the response. Can be `asc` or `desc` with a default of `asc`.
+- `receiver.id` - The receiver account the outstanding airdrop was intended for.
+- `serialnumber` - The specific serial number associated with airdrop. Supports `eq`, `gt`, `gte`, `lt`, and `lte` operators. Only one occurrence is allowed.
+- `token.id` - The token ID this airdrop is associated with. Supports `eq`, `gt`, `gte`, `lt`, and `lte` operators. Only one occurrence is allowed.
+
+#### List of pending airdrops that receiverIdOrEvmAddress has not yet claimed
+
+`/api/v1/accounts/{receiverIdOrEvmAddress}/airdrops/pending`
+
+```json
+{
+  "airdrops": [
+    {
+      "amount": 333,
+      "receiver_id": "0.0.999",
+      "sender_id": "0.0.222",
+      "serial_number": null,
+      "token_id": "0.0.111"
+    },
+    {
+      "amount": 555,
+      "receiver_id": "0.0.999",
+      "sender_id": "0.0.222",
+      "serial_number": null,
+      "token_id": "0.0.444"
+    },
+    {
+      "amount": null,
+      "receiver_id": "0.0.999",
+      "sender_id": "0.0.222",
+      "serial_number": 888,
+      "token_id": "0.0.666"
+    }
+  ],
+  "links": {
+    "next": null
+  }
+}
+```
+
+Optional Filters
+
+- `limit` - The maximum number of airdrops to return in the response. Defaults to `25` with a max of `100`.
+- `order` - The direction to sort the items in the response. Can be `asc` or `desc` with a default of `asc`.
+- `sender.id` - The sender account that initiated the pending airdrop.
+- `serialnumber` - The specific serial number associated with airdrop. Supports `eq`, `gt`, `gte`, `lt`, and `lte` operators. Only one occurrence is allowed.
+- `token.id` - The token ID this airdrop is associated with. Supports `eq`, `gt`, `gte`, `lt`, and `lte` operators. Only one occurrence is allowed.
+
+## Non-Functional Requirements
+
+- Ingest new transaction types at the same rate as consensus nodes
--- a/docs/design/frictionless-airdrops.md
+++ b/docs/design/frictionless-airdrops.md
@@ -6,15 +6,16 @@
 
 ## Goals
 
-- Ingest `TokenAirdrop` transactions and persist to the database
-- Ingest `PendingAirdrop` transactions and persist to the database
-- Ingest `TokenCancelAirdrop` and `TokenClaimAirdrop` transactions and remove entries from the database
-- Ingest `TokenReject` transactions and persist to the database
+- Ingest the following transactions and persist to the database:
+  - `PendingAirdrop`
+  - `TokenAirdrop`
+  - `TokenCancelAirdrop`
+  - `TokenClaimAirdrop`
+  - `TokenReject`
 - Expose Pending Airdrop information via the Java REST API
 
 ## Non-Goals
 
-- Support for historical airdrop information
 - Enhance gRPC APIs with Pending Airdrop information
 - Enhance Web3 APIs with Pending Airdrop information
 
@@ -25,17 +26,20 @@
 #### Pending Airdrops
 
 ```sql
-create table if not exists pending_airdrop
+create type airdrop_resolution_state as enum ('PENDING', 'CANCEL', 'CLAIM');
+
+create table if not exists token_pending_airdrop
 (
-    account_id          bigint  not null,
-    amount              bigint
-    payer_account_id    bigint  not null,
-    serial_number       bigint
-    token_id            bigint  not null
-    primary key (account_id, payer_account_id, token_id)
+    account_id          bigint                   not null,
+    amount              bigint,
+    resolution_state    airdrop_resolution_state not null default 'PENDING',
+    payer_account_id    bigint                   not null,
+    serial_number       bigint,
+    timestamp_range     int8range                not null,
+    token_id            bigint                   not null
 );
 
-create index if not exists pending__airdrop on pending_airdrop (account_id, payer_account_id);
+create index if not exists token_pending__airdrop on token_pending_airdrop (account_id, payer_account_id);
 ```
 
 ### Importer
@@ -44,37 +48,42 @@ create index if not exists pending__airdrop on pending_airdrop (account_id, paye
 
 When parsing pending airdrops,
 
-- Persist pending airdrops to the `pending_airdrop` table.
-- If a pending airdrop entry already exists, the new amount should be added to the existing amount.
+- Persist pending airdrops to the `token_pending_airdrop` table.
+- If an entry already exists for a fungible token, the timestamp range of the existing entry should be closed and a new entry should be added with `amount` set to the sum of the existing amount and the new amount.
 
 #### Domain
 
-- Add an `TokenAirdrop` domain object with the same fields as `CryptoTransfer`.
+- Add an `TokenAirdrop` domain object with the same fields as `accountId`, `amount`, `payerAccountId`, `tokenId`, and `serialNumber`.
 - Add an `TokenCancelAirdrop` domain object which contains the fields `accountId`, `payerAccountId`, `tokenId`, and `serialNumber`.
 - Add an `TokenClaimAirdrop` domain object which contains the fields `accountId`, `payerAccountId`, `tokenId`, and `serialNumber`.
 - Add an `TokenReject` domain object which contains the fields `accountId`, `tokenId`, and `serialNumber`.
-- Add an `PendingAirdrop` domain object with the same fields as the schema.
+- Add an `TokenPendingAirdrop` domain object with the same fields as the schema.
 
 #### Entity Listener
 
-- Add `onCancelAirdrop` to handle deletes from the `pending_airdrop` table.
-- Add `onClaimAirdrop` to handle deletes from the `pending_airdrop` table.
-- Add `onPendingAirdrop` to handle inserts to the `pending_airdrop` table.
+- Add `onCancelAirdrop` to handle setting the `resolution_state` and closing the `timestamp_range` on the `token_pending_airdrop` table.
+- Add `onClaimAirdrop` to handle setting the `resolution_state` flag and closing the `timestamp_range` on the `token_pending_airdrop` table.
+- Add `onTokenPendingAirdrop` to handle inserts to the `token_pending_airdrop` table.
 - Add `onTokenReject` which will be similar to `onCryptoTransfer` except the balance will be zero.
 - Add `onTokenAirdrop` which will be similar to `onCryptoTransfer`.
 
 #### Transaction Handlers
 
 - Add `TokenAirdropTransactionHandler` which will be similar to `CryptoTransferTransactionHandler`
-- Add `TokenCancelAirdropTransactionHandler`
-- Add `TokenClaimAirdropTransactionHandler`
+- Add `TokenCancelAirdropTransactionHandler` which will set the resolution state of the pending airdrop to `CANCEL`
+- Add `TokenClaimAirdropTransactionHandler` which will set the resolution state of the pending airdrop to `CLAIM`
+- Add `TokenPendingAirdropTransactionHandler`
 - Add `TokenRejectTransactionHandler` which will be similar to `CryptoTransferTransactionHandler`
-- Add `PendingAirdropTransactionHandler`
 
-#### Batch Deletion
+#### Support unlimited Max Automatic Token Associations
 
-- Add a batch deletion velocity template to be used by Claim airdrop and Cancel airdrop.
-- Add a `GenericDeleteQueryGenerator` to be used for batch deletion.
+- Add test(s) for the `-1` value for `maxAutomaticTokenAssociations`.
+
+### WEB3
+
+#### Support unlimited Max Automatic Token Associations
+
+- Update Web3 `CryptoOpsUsage` and `SyntheticTxnFactory` to handle `-1` values for `maxAutomaticTokenAssociations`.
 
 ### REST API
 
--- a/docs/design/frictionless-airdrops.md
+++ b/docs/design/frictionless-airdrops.md
@@ -1,4 +1,4 @@
-# HIP-336 Approval and Allowance
+# HIP-904 Frictionless Airdrops
 
 ## Purpose
 
@@ -7,7 +7,6 @@
 ## Goals
 
 - Ingest the following transactions and persist to the database:
-  - `PendingAirdrop`
   - `TokenAirdrop`
   - `TokenCancelAirdrop`
   - `TokenClaimAirdrop`
@@ -26,20 +25,28 @@
 #### Pending Airdrops
 
 ```sql
-create type airdrop_resolution_state as enum ('PENDING', 'CANCEL', 'CLAIM');
+create type airdrop_state as enum ('PENDING', 'CANCELLED', 'CLAIMED');
 
-create table if not exists token_pending_airdrop
+create table if not exists token_airdrop
 (
-    account_id          bigint                   not null,
     amount              bigint,
-    resolution_state    airdrop_resolution_state not null default 'PENDING',
-    payer_account_id    bigint                   not null,
+    receiver_account_id bigint         not null,
+    sender_account_id   bigint         not null,
     serial_number       bigint,
-    timestamp_range     int8range                not null,
-    token_id            bigint                   not null
+    state               airdrop_state  not null default 'PENDING',
+    timestamp_range     int8range      not null,
+    token_id            bigint         not null
 );
 
-create index if not exists token_pending__airdrop on token_pending_airdrop (account_id, payer_account_id);
+create index if not exists token__airdrop on token_airdrop (sender_account_id, receiver_account_id, token_id, serial_number);
+
+create table if not exists token_airdrop_history
+(
+  like token_airdrop including defaults
+);
+
+create index if not exists token_airdrop_history__token_serial_lower_timestamp
+  on token_airdrop_history using gist (timestamp_range);
 ```
 
 ### Importer
@@ -48,32 +55,29 @@ create index if not exists token_pending__airdrop on token_pending_airdrop (acco
 
 When parsing pending airdrops,
 
-- Persist pending airdrops to the `token_pending_airdrop` table.
-- If an entry already exists for a fungible token, the timestamp range of the existing entry should be closed and a new entry should be added with `amount` set to the sum of the existing amount and the new amount.
+- Persist airdrops to the `token_airdrop` table.
+- If a `TokenAirdrop` transaction occurs when an entry already exists for a fungible token in the `PENDING` state, an entry should be made in the `token_airdrop_history` table (just as with a change of state to `CANCELLED` or `CLAIMED`), and the `token_airdrop` table should be updated with `amount` set to the sum of the existing amount and the new amount.
 
 #### Domain
 
-- Add an `TokenAirdrop` domain object with the same fields as `accountId`, `amount`, `payerAccountId`, `tokenId`, and `serialNumber`.
-- Add an `TokenCancelAirdrop` domain object which contains the fields `accountId`, `payerAccountId`, `tokenId`, and `serialNumber`.
-- Add an `TokenClaimAirdrop` domain object which contains the fields `accountId`, `payerAccountId`, `tokenId`, and `serialNumber`.
-- Add an `TokenReject` domain object which contains the fields `accountId`, `tokenId`, and `serialNumber`.
-- Add an `TokenPendingAirdrop` domain object with the same fields as the schema.
+- Add an `TokenAirdrop` domain object with the same fields as the schema.
+- Add an `TokenCancelAirdrop` domain object which contains the fields `receiverAccountId`, `senderAccountId`, `serialNumber` and `tokenId`.
+- Add an `TokenClaimAirdrop` domain object which contains the fields `receiverAccountId`, `senderAccountId`, `serialNumber` and `tokenId`.
+- Add an `TokenReject` domain object which contains the fields `receiverAccountId`, `serialNumber` and `tokenId`.
 
 #### Entity Listener
 
-- Add `onCancelAirdrop` to handle setting the `resolution_state` and closing the `timestamp_range` on the `token_pending_airdrop` table.
-- Add `onClaimAirdrop` to handle setting the `resolution_state` flag and closing the `timestamp_range` on the `token_pending_airdrop` table.
-- Add `onTokenPendingAirdrop` to handle inserts to the `token_pending_airdrop` table.
-- Add `onTokenReject` which will be similar to `onCryptoTransfer` except the balance will be zero.
-- Add `onTokenAirdrop` which will be similar to `onCryptoTransfer`.
+- Add `onCancelAirdrop` to handle setting the `state` to `CANCELLED`. Updates the `token_airdrop` table and the `token_airdrop_history` table.
+- Add `onClaimAirdrop` to handle setting the `state` to `CLAIMED`. Updates the `token_airdrop` table and the `token_airdrop_history` table.
+- Add `onTokenAirdrop` to handle inserts to the `token_airdrop` table.
+- Add `onTokenReject` which will be similar to `onCryptoTransfer` except the balance will be zero. Does not affect the `token_airdrop` table.
 
 #### Transaction Handlers
 
-- Add `TokenAirdropTransactionHandler` which will be similar to `CryptoTransferTransactionHandler`
-- Add `TokenCancelAirdropTransactionHandler` which will set the resolution state of the pending airdrop to `CANCEL`
-- Add `TokenClaimAirdropTransactionHandler` which will set the resolution state of the pending airdrop to `CLAIM`
-- Add `TokenPendingAirdropTransactionHandler`
-- Add `TokenRejectTransactionHandler` which will be similar to `CryptoTransferTransactionHandler`
+- Add `TokenAirdropTransactionHandler` which will add a new entry to the `token_airdrop` table with the default state of `PENDING`.
+- Add `TokenCancelAirdropTransactionHandler` which will set the state of the airdrop to `CANCELLED`
+- Add `TokenClaimAirdropTransactionHandler` which will set the state of the airdrop to `CLAIMED`
+- Add `TokenRejectTransactionHandler`
 
 #### Support unlimited Max Automatic Token Associations
 
--- a/docs/design/frictionless-airdrops.md
+++ b/docs/design/frictionless-airdrops.md
@@ -61,23 +61,21 @@ When parsing pending airdrops,
 #### Domain
 
 - Add an `TokenAirdrop` domain object with the same fields as the schema.
-- Add an `TokenCancelAirdrop` domain object which contains the fields `receiverAccountId`, `senderAccountId`, `serialNumber` and `tokenId`.
-- Add an `TokenClaimAirdrop` domain object which contains the fields `receiverAccountId`, `senderAccountId`, `serialNumber` and `tokenId`.
-- Add an `TokenReject` domain object which contains the fields `receiverAccountId`, `serialNumber` and `tokenId`.
+- Add an `TokenCancelAirdrop` domain object with the same fields as the schema.
+- Add an `TokenClaimAirdrop` domain object with the same fields as the schema.
 
 #### Entity Listener
 
 - Add `onCancelAirdrop` to handle setting the `state` to `CANCELLED`. Updates the `token_airdrop` table and the `token_airdrop_history` table.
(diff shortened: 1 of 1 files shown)


docs/design/frictionless-airdrops.md
```diff
@@ -0,0 +1,171 @@
...
+
+## Purpose
+
+[HIP-904](https://hips.hedera.com/hip/hip-904) enables frictionless airdrops of both fungible and non-fungible tokens by removing the requirement to pre-associate tokens with the receiver’s account.
+
+## Goals
+
+- Ingest `TokenAirdrop` transactions and persist to the database
+- Ingest `PendingAirdrop` transactions and persist to the database
+- Ingest `TokenCancelAirdrop` and `TokenClaimAirdrop` transactions and remove entries from the database
```
edwin-greene, line 11: Instead of deletion, we could follow the model for entity, adding a deleted flag, and then adding a prune method to clean up deleted pending airdrops.

docs/design/frictionless-airdrops.md
```diff
@@ -0,0 +1,171 @@
...
+  "links": {
+    "next": null
+  }
+}
+```
+
+Optional Filters
+
+- `limit` - The maximum number of airdrops to return in the response. Defaults to `25` with a max of `100`.
+- `order` - The direction to sort the items by `token_id` in the response. Can be `asc` or `desc` with a default of `asc`.
```
edwin-greene, line 121: Since there is no consensus_timestamp, should we order on `token.id`?

docs/design/frictionless-airdrops.md
edwin-greene: After a conversation with Steven, will update this to not delete. Putting the PR into draft while I'm making those updates.

docs/design/frictionless-airdrops.md
```diff
@@ -0,0 +1,180 @@
...
+    resolution_state    airdrop_resolution_state not null default 'PENDING',
+    payer_account_id    bigint                   not null,
+    serial_number       bigint,
+    timestamp_range     int8range                not null,
+    token_id            bigint                   not null
+);
+
+create index if not exists token_pending__airdrop on token_pending_airdrop (account_id, payer_account_id);
+```
+
```
edwin-greene, line 44: The lifecycle of a pending airdrop always ends in either `cancel` or `claim`. Over time the number of non-pending airdrops will constantly increase.

As the rest api currently has no need for information about non-pending airdrops, the queries involved may eventually have performance problems as the bulk of the table will progressively be dominated by non-pending airdrops. To address this we may want to have a separate table for resolved airdrops, such that when an airdrop is canceled or claimed it is moved from the pending airdrops table to a resolved airdrops table.

steven-sheehy: We can also consider adding deleted=false condition to every airdrop index to reduce the index size.

docs/design/frictionless-airdrops.md
```diff
@@ -0,0 +1,180 @@
...
+- Add an `TokenReject` domain object which contains the fields `accountId`, `tokenId`, and `serialNumber`.
+- Add an `TokenPendingAirdrop` domain object with the same fields as the schema.
+
+#### Entity Listener
+
+- Add `onCancelAirdrop` to handle setting the `resolution_state` and closing the `timestamp_range` on the `token_pending_airdrop` table.
+- Add `onClaimAirdrop` to handle setting the `resolution_state` flag and closing the `timestamp_range` on the `token_pending_airdrop` table.
+- Add `onTokenPendingAirdrop` to handle inserts to the `token_pending_airdrop` table.
+- Add `onTokenReject` which will be similar to `onCryptoTransfer` except the balance will be zero.
+- Add `onTokenAirdrop` which will be similar to `onCryptoTransfer`.
```
edwin-greene, line 68: This treats the `TokenAirdrop` as a CryptoTransfer, with similar handling - just updating the balance/balanceTimestamp. But the "source" of a transfer, whether it was a normal transfer or a TokenAirdrop, is not persisted. So perhaps we should add a column to the token_transfer table that indicates the source?

docs/design/frictionless-airdrops.md
```diff
@@ -0,0 +1,180 @@
...
+- Enhance Web3 APIs with Pending Airdrop information
+
+## Architecture
+
+### Database
+
+#### Pending Airdrops
+
+```sql
+create type airdrop_resolution_state as enum ('PENDING', 'CANCEL', 'CLAIM');
```
xin-hedera, line 29: is this directly from HIP / proposed proto files? Shouldn't we use `CANCELLED` / `CLAIMED`  to describe states?

docs/design/frictionless-airdrops.md
```diff
@@ -0,0 +1,180 @@
...
+
+create table if not exists token_pending_airdrop
+(
+    account_id          bigint                   not null,
+    amount              bigint,
+    resolution_state    airdrop_resolution_state not null default 'PENDING',
+    payer_account_id    bigint                   not null,
+    serial_number       bigint,
+    timestamp_range     int8range                not null,
+    token_id            bigint                   not null
```
xin-hedera, line 39: shouldn't there be a `sender_id`?

docs/design/frictionless-airdrops.md
```diff
@@ -0,0 +1,180 @@
...
+);
```
xin-hedera, line 40: How do we update the table when a pending airdrop becomes either cancelled or claimed? We don't have a unique index, it's going to be slow at least.

Example is, account `A` airdrops NFT `N-1` and `N-2` to account `B`, when account `B` claims them, we need an index to update the state efficiently.

This can extend beyond past already finalized airdrops since it's perfectly fine for the same token being airdopped again after the previous one is cancelled / claimed.

docs/design/frictionless-airdrops.md
xin-hedera, line 68: I don't think we need both `onTokenAirdrop` and `onTokenPendingAirdrop`

immediately succeeded  token transfers from a `TokenAirdrop` transaction should be reflected in the transaction record's token transfer list and `EntityRecordItemListener` already handles it, however we may have to check closely if any logic needs to be updated, especially for approved transfers if any.

docs/design/frictionless-airdrops.md
```diff
@@ -0,0 +1,180 @@
...
+- Add `onTokenPendingAirdrop` to handle inserts to the `token_pending_airdrop` table.
+- Add `onTokenReject` which will be similar to `onCryptoTransfer` except the balance will be zero.
+- Add `onTokenAirdrop` which will be similar to `onCryptoTransfer`.
+
+#### Transaction Handlers
+
+- Add `TokenAirdropTransactionHandler` which will be similar to `CryptoTransferTransactionHandler`
+- Add `TokenCancelAirdropTransactionHandler` which will set the resolution state of the pending airdrop to `CANCEL`
+- Add `TokenClaimAirdropTransactionHandler` which will set the resolution state of the pending airdrop to `CLAIM`
+- Add `TokenPendingAirdropTransactionHandler`
```
xin-hedera, line 75: There is no `TokenPendingAirdropTransaction`

xin-hedera (changes requested):

docs/design/frictionless-airdrops.md
xin-hedera: update: aren't we supposed to have a history table as well? For the index, I think we need a unique index to make `on conflict ... do ...` work. There is still issue when updating fungible token airdrops as we don't know the exact amount, at least for `TokenCancelAirdrop` transaction

docs/design/frictionless-airdrops.md
xin-hedera: update: I don't think we should allow approved token transfers in `TokenAirdrop` transaction, nor did the HIP intend to allow it. However, we need to amend the HIP to call it out explicitly.

docs/design/frictionless-airdrops.md
xin-hedera: these are the changes needed to make upsert work with the table

- rename `account_id` to `receiver_account_id`
- add a not null bigint column `sender_account_id`
- create a unique index on `(sender_account_id, receiver_account_id, token_id, serial_number)`

We'll keep both `amount` and `serial_number`.

For fungible token, `serial_number` will be always `NULL`. That way, when we do update, on conflict ... update ... will work.

In addition, I'm not sure about how useful `pay_acount_id` is, we definitely should not create an index including the column. The payer account of the transactions which updating the same airdrop can change over the time, e.g., obviously, in the simplest use case, `TokenAirdrop` and `TokenClaimAirdrop` should have different payer account id.

docs/design/frictionless-airdrops.md
```diff
@@ -0,0 +1,180 @@
...
+
+- Update Web3 `CryptoOpsUsage` and `SyntheticTxnFactory` to handle `-1` values for `maxAutomaticTokenAssociations`.
+
+### REST API
+
+- Add the new endpoints to Java REST.
+
+#### List of outstanding airdrops sent by senderIdOrEvmAddress which have not been claimed by recipients
+
+`/api/v1/accounts/{senderIdOrEvmAddress}/airdrops/outstanding`
```
jascks, line 94: Should we just use `{idOrAliasOrEvmAddress}` here. We can say it's the sender's ID in the docs. Is an account alias valid for this API?

jascks: I was a little confused about the difference between _outstanding_ and _pending_, but it looks like it is really used in the URI path to indicate whether the provided account ID path parameter is that of the sender or the receiver.

We just added the NFT allowances API that utilizes the query parameter `ownerQueryParam` to do something similar regarding the owner and spender. I've not done a full survey of the existing API endpoints, but this is inconsistent.

docs/design/frictionless-airdrops.md
jascks: True, but it looks like `TokenAirdropTransactionHandler` is going to have to check the transaction record and process and `PendingAirdropRecord` to add any pending airdrops resulting from the transaction.

In addition, we need to note the comment in `TokenAirdropTransactionBody`:
```
 * If an airdrop would create a pending airdrop for a fungible/common token, and a pending airdrop
 * for the same sender, receiver, and token already exists, the existing pending airdrop
 * SHALL be updated to add the new amount to the existing airdrop, rather than creating a new
 * pending airdrop.
 ```

File: docs/design/frictionless-airdrops.md (hunk 5 above)
Comment (edwin-greene) [reply docs/design/frictionless-airdrops.md]: Sounds good I have updated them.

File: docs/design/frictionless-airdrops.md [hunk 10]
```diff
@@ -0,0 +1,180 @@
...
+## Architecture
+
+### Database
+
+#### Pending Airdrops
+
+```sql
+create type airdrop_resolution_state as enum ('PENDING', 'CANCEL', 'CLAIM');
+
+create table if not exists token_pending_airdrop
```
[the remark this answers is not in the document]
edwin-greene: Got it, fixed.

docs/design/frictionless-airdrops.md
```diff
@@ -0,0 +1,180 @@
...
+(
+    account_id          bigint                   not null,
+    amount              bigint,
+    resolution_state    airdrop_resolution_state not null default 'PENDING',
```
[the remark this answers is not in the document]
edwin-greene: Sounds good I have updated the name to `state`.

docs/design/frictionless-airdrops.md
edwin-greene: Thanks, I have updated it to add the `sender_id`.

docs/design/frictionless-airdrops.md
edwin-greene: Thanks I have updated it according to these suggestions. Do we want the index of the history table to be just on timestamp range?

docs/design/frictionless-airdrops.md
edwin-greene: For immediately successful token transfers, don't we need a way to distinguish them as "airdropped" vs normal token transfers?

docs/design/frictionless-airdrops.md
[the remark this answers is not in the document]
edwin-greene: Sounds good, I have updated it to be named `airdrop_state`.

docs/design/frictionless-airdrops.md
edwin-greene: It may be useful, but the HIP specified the rest api specifics and I think additional changes would need to go through the review process

docs/design/frictionless-airdrops.md
xin-hedera: I don't see the importance to mark it as being airdropped, it in fact looks more like a normal token transfer as the same can be done by a `CryptoTransfer` transaction when the receiver already has the corresponding token association or available auto token association slot.

docs/design/frictionless-airdrops.md
xin-hedera: I believe treating the path variable as receiver account id makes more sense as the most common query would be as a receiver, what my current airdrops are.

I'm sure we'll get requests to show airdrops by sender from users. However, as of now, we can limit it to receiver while leave it open for extension.

jascks: I see, so later we could add something like, say, `senderQueryParam`, and default to `false`. And when `true`, query by sender.

docs/design/frictionless-airdrops.md
xin-hedera: not sure about the index for the history table. potentially web3 may read the table for historical state.

docs/design/frictionless-airdrops.md
```diff
@@ -0,0 +1,184 @@
...
+When parsing pending airdrops,
+
+- Persist airdrops to the `token_airdrop` table.
+- If a `TokenAirdrop` transaction occurs when an entry already exists for a fungible token in the `PENDING` state, an entry should be made in the `token_airdrop_history` table (just as with a change of state to `CANCELLED` or `CLAIMED`), and the `token_airdrop` table should be updated with `amount` set to the sum of the existing amount and the new amount.
+
+#### Domain
+
+- Add an `TokenAirdrop` domain object with the same fields as the schema.
+- Add an `TokenCancelAirdrop` domain object which contains the fields `receiverAccountId`, `senderAccountId`, `serialNumber` and `tokenId`.
+- Add an `TokenClaimAirdrop` domain object which contains the fields `receiverAccountId`, `senderAccountId`, `serialNumber` and `tokenId`.
```
xin-hedera, line 65: anything updating the `token_airdrop` table should reuse the same `TokenAirdrop` domain class

docs/design/frictionless-airdrops.md
```diff
@@ -0,0 +1,184 @@
...
+- Add an `TokenReject` domain object which contains the fields `receiverAccountId`, `serialNumber` and `tokenId`.
```
xin-hedera, line 66: why do we need a `TokenReject` domain object? `TokenReject` transaction's effect is externalized as token transfer list in the transaction record, that is, any such rejected token is sent from the transaction payer to the token treasury.

docs/design/frictionless-airdrops.md
```diff
@@ -0,0 +1,184 @@
...
+
+#### Entity Listener
+
+- Add `onCancelAirdrop` to handle setting the `state` to `CANCELLED`. Updates the `token_airdrop` table and the `token_airdrop_history` table.
+- Add `onClaimAirdrop` to handle setting the `state` to `CLAIMED`. Updates the `token_airdrop` table and the `token_airdrop_history` table.
+- Add `onTokenAirdrop` to handle inserts to the `token_airdrop` table.
+- Add `onTokenReject` which will be similar to `onCryptoTransfer` except the balance will be zero. Does not affect the `token_airdrop` table.
```
xin-hedera, line 73: "balance will be zero" is only true for rejected fungible tokens. For NFTs, token reject can only send back 1 serial per `TokenReference` rejection in the transaction body so it would not necessarily drop NFT balance to 0.

edwin-greene: Thanks I have removed `onTokenReject`

docs/design/frictionless-airdrops.md
edwin-greene: I see, I have updated the design and removed the domain object for `TokenReject`

docs/design/frictionless-airdrops.md
edwin-greene: Thanks I have updated the design accordingly.

docs/design/frictionless-airdrops.md
```diff
@@ -0,0 +1,182 @@
...
+When parsing pending airdrops,
+
+- Persist airdrops to the `token_airdrop` table.
+- If a `TokenAirdrop` transaction occurs when an entry already exists for a fungible token in the `PENDING` state, an entry should be made in the `token_airdrop_history` table (just as with a change of state to `CANCELLED` or `CLAIMED`), and the `token_airdrop` table should be updated with `amount` set to the sum of the existing amount and the new amount.
+
+#### Domain
+
+- Add an `TokenAirdrop` domain object with the same fields as the schema.
+- Add an `TokenCancelAirdrop` domain object with the same fields as the schema.
+- Add an `TokenClaimAirdrop` domain object with the same fields as the schema.
```
xin-hedera, line 65: I assume you meant `domain class`, I don't see the benefit of adding different domain classes per transaction type, and it will make implementation to unnecessarily cast / convert a `TokenCancelAirdrop` / `TokenClaimAirdrop` to `TokenAirdrop` domain object, then apply the merge logic, and finally go into the upsert SQL to update `token_airdrop` / `token_airdop_hisitory` table

Instead we should just have a `TokenAirdrop` domain class

- for `TokenCancelAirdrop`, it fills in the fields in the unique index, the lower timestamp, and set the state to `cancelled`, then call `onTokenAirdrop` or `onCancelAirdrop` if you like that approach more
- for `TokenClaimAirdrop`, similar to above, only difference is state is set to `claimed`

xin-hedera (changes requested):

docs/design/frictionless-airdrops.md
```diff
@@ -0,0 +1,182 @@
...
+}
+```
+
+Optional Filters
+
+- `limit` - The maximum number of airdrops to return in the response. Defaults to `25` with a max of `100`.
+- `order` - The direction to sort the items by `token_id` in the response. Can be `asc` or `desc` with a default of `asc`.
+- `receiver.id` - The receiver account the outstanding airdrop was intended for.
+- `serialnumber` - The specific serial number associated with airdrop. Supports `eq`, `gt`, `gte`, `lt`, and `lte` operators. Only one occurrence is allowed.
+- `token.id` - The token ID this airdrop is associated with. Supports `eq`, `gt`, `gte`, `lt`, and `lte` operators. Only one occurrence is allowed.
```
xin-hedera, line 135: just a comment, so now we have an endpoint with 3 pagination params? To add to the complexity, for fungible tokens, serial number is `null`...

xin-hedera: perhaps I misunderstood it if we only want to support `eq` operator for `receiver.id`, if so, need to call it out.

docs/design/frictionless-airdrops.md
Nana-EC: > update: I don't think we should allow approved token transfers in TokenAirdrop transaction, nor did the HIP intend to allow it.

TokenAirdrop transaction will support the transferring of tokens from allowance.

Nana-EC: > For immediately successful token transfers, don't we need a way to distinguish them as "airdropped" vs normal token transfers?

No need for this a the ledger level

docs/design/frictionless-airdrops.md
mgoelswirlds: So depending on the decision of what the path parameter at this stage would be the query param would change to sender right? also would that also support only eq?

docs/design/frictionless-airdrops.md
```diff
@@ -0,0 +1,182 @@
...
+#### Domain
+
+- Add an `TokenAirdrop` domain object with the same fields as the schema.
+- Add an `TokenCancelAirdrop` domain object with the same fields as the schema.
+- Add an `TokenClaimAirdrop` domain object with the same fields as the schema.
+
+#### Entity Listener
+
+- Add `onCancelAirdrop` to handle setting the `state` to `CANCELLED`. Updates the `token_airdrop` table and the `token_airdrop_history` table.
+- Add `onClaimAirdrop` to handle setting the `state` to `CLAIMED`. Updates the `token_airdrop` table and the `token_airdrop_history` table.
```
steven-sheehy, line 70: These two should go away. There's just one onTokenAirdrop with different states passed in.

steven-sheehy (changes requested):

docs/design/frictionless-airdrops.md
```diff
@@ -0,0 +1,182 @@
...
+```
+
+### Importer
+
+#### Pending Airdrop Parsing
+
+When parsing pending airdrops,
+
+- Persist airdrops to the `token_airdrop` table.
+- If a `TokenAirdrop` transaction occurs when an entry already exists for a fungible token in the `PENDING` state, an entry should be made in the `token_airdrop_history` table (just as with a change of state to `CANCELLED` or `CLAIMED`), and the `token_airdrop` table should be updated with `amount` set to the sum of the existing amount and the new amount.
```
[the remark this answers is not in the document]
xin-hedera: it should be replaced, like token account association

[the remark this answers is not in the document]
edwin-greene: The way I understand it thus far for fungible tokens is that when the upsert occurs, if the existing entry is in the `PENDING` state, then the amount of the existing entry and the new airdrop should be summed and a history table row should be generated (with the amount from the first entry and timestamp range close off). 

If the existing entry is in any other state, then the amount should not be summed and a history table row should be generated in the same manner.

docs/design/frictionless-airdrops.md
edwin-greene: Thanks I have removed them

docs/design/frictionless-airdrops.md
edwin-greene: Thanks I have added eq and the other operators, the HIP did not specify so I am unsure if we would like to just support eq or match the others

docs/design/frictionless-airdrops.md
```diff
@@ -0,0 +1,180 @@
...
+
+#### Support unlimited Max Automatic Token Associations
+
+- Add test(s) for the `-1` value for `maxAutomaticTokenAssociations`.
+
+### WEB3
+
+#### Support unlimited Max Automatic Token Associations
+
+- Update Web3 `CryptoOpsUsage` and `SyntheticTxnFactory` to handle `-1` values for `maxAutomaticTokenAssociations`.
```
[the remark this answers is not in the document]
edwin-greene: Thanks I have updated it with that specification

docs/design/frictionless-airdrops.md
```diff
@@ -0,0 +1,180 @@
...
+
+### REST API
+
+- Add the new endpoints to Java REST.
```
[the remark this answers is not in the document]
edwin-greene: Got it I have fixed this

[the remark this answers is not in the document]
edwin-greene: Thanks I have added a sentence of explanation

docs/design/frictionless-airdrops.md
```diff
@@ -0,0 +1,180 @@
...
+
+#### List of outstanding airdrops sent by senderIdOrEvmAddress which have not been claimed by recipients
```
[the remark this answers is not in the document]
edwin-greene: Thanks I have added some more details here.

docs/design/frictionless-airdrops.md
```diff
@@ -0,0 +1,180 @@
...
+- Add `TokenCancelAirdropTransactionHandler` which will set the resolution state of the pending airdrop to `CANCEL`
+- Add `TokenClaimAirdropTransactionHandler` which will set the resolution state of the pending airdrop to `CLAIM`
+- Add `TokenPendingAirdropTransactionHandler`
+- Add `TokenRejectTransactionHandler` which will be similar to `CryptoTransferTransactionHandler`
+
+#### Support unlimited Max Automatic Token Associations
+
+- Add test(s) for the `-1` value for `maxAutomaticTokenAssociations`.
+
+### WEB3
```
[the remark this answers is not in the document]
edwin-greene: Got it, this is fixed.

docs/design/frictionless-airdrops.md
```diff
@@ -0,0 +1,182 @@
...
+    amount              bigint,
+    receiver_account_id bigint         not null,
+    sender_account_id   bigint         not null,
+    serial_number       bigint,
+    state               airdrop_state  not null default 'PENDING',
+    timestamp_range     int8range      not null,
+    token_id            bigint         not null
+);
+
+create index if not exists token__airdrop on token_airdrop (sender_account_id, receiver_account_id, token_id, serial_number);
```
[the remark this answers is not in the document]
edwin-greene: Thanks I have added another index, please double check that it looks correct

docs/design/frictionless-airdrops.md
edwin-greene: Got it I have removed them.

docs/design/frictionless-airdrops.md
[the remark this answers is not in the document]
steven-sheehy: My question was to spur changes to the document, not answering in comments to a PR.

docs/design/frictionless-airdrops.md
Nana-EC: Each product has some flexibility to improve the design.
The HIP is the first high level design. Design docs like this are certainly where if a param is missing it should be added and yes we can go back and update the HIP based on implementation.
So far as the overall goal and design of the HIP isn't changes there's some wiggle room.

I believe in this case the transaction consensus timestamp would be useful to capture and will likely be one criteria for which wallets and explorer will use to sort the list by.

edwin-greene: Thanks, sounds good I have added `consensus_timestamp` to the returned results.

docs/design/frictionless-airdrops.md
```diff
@@ -0,0 +1,180 @@
...
+```
+
+Optional Filters
+
+- `limit` - The maximum number of airdrops to return in the response. Defaults to `25` with a max of `100`.
+- `order` - The direction to sort the items by `token_id` in the response. Can be `asc` or `desc` with a default of `asc`.
+- `sender.id` - The sender account that initiated the pending airdrop.
+- `serialnumber` - The specific serial number associated with airdrop. Supports `eq`, `gt`, `gte`, `lt`, and `lte` operators. Only one occurrence is allowed.
+- `token.id` - The token ID this airdrop is associated with. Supports `eq`, `gt`, `gte`, `lt`, and `lte` operators. Only one occurrence is allowed.
+
```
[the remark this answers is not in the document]
edwin-greene: Got it, I have added that section

docs/design/frictionless-airdrops.md
steven-sheehy: Please revert this change. The payload structure is defined in the HIP and agreed upon by the committee and council. The timestamp range allows us the flexibility of adding a timestamp parameter in the future.

steven-sheehy: Actually, I spoke too soon. The HIP doesn't have any timestamp. We can add the field but add as `timestamp: {from: 1, to: null}` to be consistent. Need to also update the HIP.

edwin-greene: Thanks I have updated the timestamp part of the response. If it looks good as I have it, I will open a PR to add it to the HIP.

docs/design/frictionless-airdrops.md
```diff
@@ -0,0 +1,180 @@
...
+## Non-Functional Requirements
```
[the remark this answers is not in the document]
edwin-greene: I have added some detail to the Java Rest section to include consideration of performance on Citus. Do you think that is sufficient, or is there something more we should look into regarding the design of the indexes in consideration of the Citus partitioning?

docs/design/frictionless-airdrops.md
```diff
@@ -0,0 +1,211 @@
...
+  "links": {
+    "next": null
+  }
+}
+```
+
+Optional Filters
+
+- `limit` - The maximum number of airdrops to return in the response. Defaults to `25` with a max of `100`.
+- `order` - The direction to sort the items by `consensus_timestamp` in the response. Can be `asc` or `desc` with a default of `asc`.
```
steven-sheehy, line 206: The sort order is not by timestamp, but by the multi-column index it's using. And this depends upon whether it's pending or outstanding.

docs/design/frictionless-airdrops.md
```diff
@@ -0,0 +1,211 @@
...
+## Non-Functional Requirements
+
+- Ingest new transaction types at the same rate as consensus nodes
+
+## Acceptance Tests
+
+Add acceptance tests for the token feature that use an existing set of accounts and a token_id from the existing acceptance tests and performs the following:
+
+- Send two airdrops, a fungible and nft airdrop to an account that has been associated with the tokens and verify that the airdrops are transferred to the account and that the two REST APIs do not list the airdrops.
+- Reject the set of airdrops and verify that the account no longer has a balance of the fungible token and does not own the nft.
```
steven-sheehy, line 220: You can't reject them since they never went to pending and were directly transferred since they're previously associated. So we should add another scenario where we send an airdrop and sender rejects.

docs/design/frictionless-airdrops.md
```diff
@@ -0,0 +1,211 @@
...
+- Send a fungible and nft airdrop to an account that has not been associated with the tokens and has no open slots for automatic associations and verify that the airdrops are listed by the two REST APIs.
+- Cancel the pending airdrops and verify that they are no longer listed by the two REST APIs.
+- Send a fungible and nft airdrop to an account that has not been associated with the tokens and has no open slots for automatic associations and verify that the airdrop is listed in the two REST APIs.
+- Claim the airdrops and verify that they are no longer listed in the two REST APIs.
```
steven-sheehy, line 224: You should also verify they are listed in other APIs like `/accounts/{id}/nfts`, `/accounts/{id}/tokens`, and `/tokens/{id}/nfts/{serial}`.

Same comment for other scenarios where the transfer occurs.

docs/design/frictionless-airdrops.md
```diff
@@ -0,0 +1,229 @@
...
+  "links": {
+    "next": null
+  }
+}
+```
+
+Optional Filters
+
+- `limit` - The maximum number of airdrops to return in the response. Defaults to `25` with a max of `100`.
+- `order` - The direction to sort the items by `consensus_timestamp` in the response. Can be `asc` or `desc` with a default of `asc`.
```
steven-sheehy, line 146: Same comment

steven-sheehy (changes requested):

docs/design/frictionless-airdrops.md
edwin-greene: Thanks for catching that, I have corrected the sorting details.

docs/design/frictionless-airdrops.md
edwin-greene: Got it this is now fixed.

docs/design/frictionless-airdrops.md
edwin-greene: Great idea, I have added those endpoints to the verification for each of the different steps.

docs/design/frictionless-airdrops.md
edwin-greene: For tokens in `PENDING` state, my current understanding is that the receiver can only claim them and that they cannot reject them from that state. For tokens that the receiver already owns, they may return them to the token treasury with token reject.

steven-sheehy: > For tokens in PENDING state

For the scenario I commented on, I'm pretty sure it never goes to PENDING state. Since they're already associated it directly gets transferred.

docs/design/frictionless-airdrops.md
```diff
@@ -0,0 +1,229 @@
...
+    amount              bigint,
+    receiver_account_id bigint         not null,
+    sender_account_id   bigint         not null,
+    serial_number       bigint,
+    state               airdrop_state  not null default 'PENDING',
+    timestamp_range     int8range      not null,
+    token_id            bigint         not null
+);
+
+create index if not exists token_airdrop__sender_id on token_airdrop (sender_account_id, receiver_account_id, token_id, serial_number);
```
xin-hedera, line 41: This needs to be a unique index to make upset on conflict work

docs/design/frictionless-airdrops.md
```diff
@@ -0,0 +1,229 @@
...
+  - `TokenAirdrop`
+  - `TokenCancelAirdrop`
+  - `TokenClaimAirdrop`
+  - `TokenReject`
+- Expose Pending Airdrop information via the Java REST API
+
+## Non-Goals
+
+- Enhance gRPC APIs with Pending Airdrop information
+- Enhance Web3 APIs with Pending Airdrop information
```
Nana-EC, line 19: Note. The HIP calls out smart contract functions that will expose the HAPI logic.
Mirror node will not have to write much code to support it but it will have to add test coverage to ensure contracts/call supports the call and estimation flows.
We can leave that part for a design doc update in the future as HAPI features will be prioritizes before system contract implementation.
Feel free to note this in the design doc if valuable

docs/design/frictionless-airdrops.md
```diff
@@ -0,0 +1,229 @@
...
+```
+
+### Importer
+
+#### Pending Airdrop Parsing
+
+When parsing pending airdrops,
+
+- Persist airdrops to the `token_airdrop` table.
+- If a `TokenAirdrop` transaction occurs when an entry already exists then the existing entry should be moved in the `token_airdrop_history` table. If the existing entry is a fungible token in the state of `PENDING` then the new entry should be updated with `amount` set to the sum of the existing amount and the new amount. If the existing entry is in any other state, then the `amount` should be set to the amount of the new entry. Updates to NFT tokens that are `PENDING` will update the timestamp range and generate a history row.
```
Nana-EC, line 60:
```suggestion
- If a `TokenAirdrop` transaction occurs when an entry already exists i.e matching sender, receiver, token_id and serial_number (if an NFT), then the existing entry should be moved into the `token_airdrop_history` table. If the existing entry is a fungible token in the state of `PENDING` then the new entry should be updated with `amount` set to the sum of the existing amount and the new amount. If the existing entry is in any other state, then the `amount` should be set to the amount of the new entry. Updates to NFT tokens that are `PENDING` will update the timestamp range and generate a history row.
```

Nana-EC, line 60: Every update other than PENDING for a fungible token should not see an amount change since you either claim or cancel all fungible tokens. So the amount should not change
```suggestion
- If a `TokenAirdrop` transaction occurs when an entry already exists then the existing entry should be moved in the `token_airdrop_history` table. If the existing entry is a fungible token in the state of `PENDING` then the new entry should be updated with `amount` set to the sum of the existing amount and the new amount. Updates to NFT tokens that are `PENDING` will update the timestamp range and generate a history row.
```

Nana-EC, line 60: Is this not covered by the first sentence? If so remove it. If you want to clarify something else it needs more details since you already established a history row.
```suggestion
- If a `TokenAirdrop` transaction occurs when an entry already exists then the existing entry should be moved in the `token_airdrop_history` table. If the existing entry is a fungible token in the state of `PENDING` then the new entry should be updated with `amount` set to the sum of the existing amount and the new amount. If the existing entry is in any other state, then the `amount` should be set to the amount of the new entry. 
```

docs/design/frictionless-airdrops.md
Nana-EC: So I'm not clear that there's a need to ask "Show me the Claimed or Cancelled airdrops".
if there is i'd imagine querying for transactions of those types would be better than hitting the airdrop API.
In which case another consideration might be moving claimed and cancelled airdrops to the history row only. No need to have them in `token_airdrop`, this should reduce the bloat.

docs/design/frictionless-airdrops.md
Nana-EC: The goal of this API is to see for a given address what outstanding (not claimed or cancelled) airdrops they have.
So the `/airdrops/outstanding` path should always be in the context of the sender

docs/design/frictionless-airdrops.md
```diff
@@ -0,0 +1,229 @@
...
+- Reject the tokens and verify that related endpoints for the account (`/accounts/{id}/nfts`, `/accounts/{id}/tokens`) do not show the account as the owner of the tokens. Verify that the tokens have been returned to their treasury account (`/accounts/{treasury account id}/nfts`, `/accounts/{treasury account id}/tokens`, and `/tokens/{id}/nfts/{serial}`).
+- Send a fungible and nft airdrop to an account that has not been associated with the tokens and has no open slots for automatic associations and verify that the airdrops are listed by the two REST APIs. Verify that related endpoints for the account (`/accounts/{id}/nfts`, `/accounts/{id}/tokens`, and `/tokens/{id}/nfts/{serial}`) do not show the account as the owner of the tokens.
+- Cancel the pending airdrops and verify that they are no longer listed by the two REST APIs.
```
Nana-EC, line 222: Can add that the sender still owns the tokens and the receiver does not.

Nana-EC (changes requested):
Looking good.
Some suggestions for consideration and clarity.

docs/design/frictionless-airdrops.md
edwin-greene: Thanks I have added that as a verification step.

docs/design/frictionless-airdrops.md
edwin-greene: I have updated this to separate the behavior between nft and fungible tokens, it should be easier to read and include these clarifications.

edwin-greene: Thanks this sentence has been reworked into two.

edwin-greene: Thanks this sentence has been reworked into two.

docs/design/frictionless-airdrops.md
edwin-greene: Got it I have updated it to have the `unique` keyword

docs/design/frictionless-airdrops.md
steven-sheehy: That doesn't make sense either. Once they're claimed they are also transferred. The token reject transaction is for the sender to cancel pending airdrops that have not been claimed or transferred.

docs/design/frictionless-airdrops.md
```diff
@@ -0,0 +1,230 @@
...
+- Send a fungible and nft airdrop to an account that has not been associated with the tokens and has no open slots for automatic associations and verify that the airdrop is listed in the two REST APIs. Verify that related endpoints for the account (`/accounts/{id}/nfts`, `/accounts/{id}/tokens`, and `/tokens/{id}/nfts/{serial}`) do not show the account as the owner of the tokens.
+- Claim the airdrops and verify that they are no longer listed in the two REST APIs. Verify that related endpoints for the account (`/accounts/{id}/nfts`, `/accounts/{id}/tokens`, and `/tokens/{id}/nfts/{serial}`) show the account as the owner of the tokens.
+- Reject the tokens that were just claimed and verify that related endpoints for the account (`/accounts/{id}/nfts`, `/accounts/{id}/tokens` and `/tokens/{id}/nfts/{serial}`) do not show the account as the owner of the tokens. Verify that the tokens have been returned to their treasury account (`/accounts/{treasury account id}/nfts`, `/accounts/{treasury account id}/tokens`, and `/tokens/{id}/nfts/{serial}`).
```
steven-sheehy, line 225: You can't reject tokens that are already claimed. So you'll need to do another `Send a fungible and nft airdrop to an account that has not been associated` setup before rejecting.

steven-sheehy (approved):
LGTM

xin-hedera (approved):
LGTM

Nana-EC (approved):
LG

PR #8134, opened 2024-04-19
Speed up PostgreSQL to Citus migration

### LINKED ISSUE #7855: Research faster PostgreSQL to Citus migration approach
### Problem

The current approach to pg_dump from Cloud SQL to files then pg_restore those files to Citus takes a very long time.

### Solution

* Investigate feasibility of using [dimitri/pgcopydb](https://github.com/dimitri/pgcopydb)
* Explore Kafka as a sink
* Consider optimizations to current approach

### Alternatives

_No response_

jnels124:
**Description**:
* add ability to execute migrations in batches
* add ability to restart migration
* use direct pipe instead of download/upload of csv
* add config option to parallelize batch copies

**Related issue(s)**:

Fixes #7855 

**Notes for reviewer**:

**Checklist**


hedera-mirror-importer/src/main/resources/db/scripts/v2/migration.sh
```diff
@@ -1,69 +1,255 @@
...
+  trap "handleExit 'for table ${table} (BINARY)'" ERR
+
+  local tableExists="f"
+  tableExists=$(targetTableExists "${table}")
+  if [[ "${tableExists}" = "f" ]]; then
+    log "Skipping '${table}' table since it doesn't exist on the target"
+    return
+  fi
+
+  cat >querySource.sh <<EOF
```
jnels124, line 220: Unable to get the querySource function to be available via call from `FROM program`

hedera-mirror-importer/src/main/resources/db/scripts/v2/migration.sh
```diff
@@ -1,69 +1,255 @@
...
+  if [[ "${has_data}" = "t" ]]; then
+    log "Skipping '${table}' table since it contains existing data"
+    return
+  fi
+
+  if [[ "${table}" != "topic_message" && "${columns}" =~ $TS_COLUMN_REGEX ]]; then
+    where="where consensus_timestamp \<= ${MAX_TIMESTAMP}"
+  # handle address book.
+  elif [[ "${columns}" =~ \<consensus_timestamp_start\> ]]; then
+    where="where consensus_timestamp_end \<= ${MAX_TIMESTAMP}"
```
jnels124, line 145: not indexed but table is small

hedera-mirror-importer/src/main/resources/db/scripts/v2/migration.sh
```diff
@@ -1,69 +1,255 @@
...
+  LEAST(${MAX_TIMESTAMP}, (extract(epoch from (rfs.start_timestamp + INTERVAL '1 month')) * 1000000000)::bigint) as upper_bound
+  from record_file_stats rfs, total_records tr
+  order by 3 desc
+  "
+
+  local upperBound=$MAX_TIMESTAMP
+  local pctPerCursor=$(echo "scale=8; 100 / ${ASYNC_TABLE_SPLITS}" | bc)
+  local totalPct=0
+  local -n cursors=$1 # Init array to hold cursor definitions
+  local stats=$(PGPASSWORD="${SOURCE_DB_PASSWORD}" psql -q --csv -t -h "${SOURCE_DB_HOST}" -d "${SOURCE_DB_NAME}" -p "${SOURCE_DB_PORT}" -U "${SOURCE_DB_USER}" -c "${query}")
```
jnels124, line 108: unable to get `querySource` function to work correctly here. Attempting to asign the result of that call to stats and then redirect to the while doesn't work. It attempts to execute the text as a command. Can use assistance with the 'bashism' here.

hedera-mirror-importer/src/main/resources/db/scripts/v2/migration.sh
```diff
@@ -1,69 +1,250 @@
...
+
+  while IFS="," read -r pct_of_total lower_bound upper_bound; do
+    upperBound=$((upper_bound > upperBound ? upper_bound : upperBound))
+    local groupRange=$((upperBound - lower_bound))
+    local rangeToTake=$(echo "scale=0; ${pctPerCursor} * ${groupRange}" | bc)
+    totalPct=$(echo "(${totalPct} + ${pct_of_total})" | bc)
+    while (($(echo "${totalPct} >= ${pctPerCursor}" | bc) )); do
```
jnels124, line 120: correcting this calculation

hedera-mirror-importer/src/main/resources/db/scripts/v2/migration.sh
```diff
@@ -1,69 +1,277 @@
...
+  local columns="${2}"
+
+  local has_data=$(queryTarget "select exists (select * from ${table}  limit 1)")
+  if [[ "${has_data}" = "t" ]]; then
+    log "Skipping '${table}' table since it contains existing data"
+    return
+  fi
+
+  local where
+  if [[ "${table}" != "topic_message" && "${columns}" =~ $TS_COLUMN_REGEX ]]; then
```
jnels124, line 147: will remove topic message restriction here before merge

hedera-mirror-importer/src/main/resources/db/scripts/v2/migration.sh
```diff
@@ -125,14 +340,30 @@ log "Running Flyway migrate"
...
 
 TABLES=$(querySource "${TABLES_QUERY}" | tr " " "\n")
-COUNT=$(echo "${TABLES}" | wc -l)
-log "Migrating ${COUNT} tables from ${SOURCE_DB_HOST}:${SOURCE_DB_PORT} to ${TARGET_DB_HOST}:${TARGET_DB_PORT}"
-echo "${TABLES}" | xargs -n 1 -P "${CONCURRENCY}" -I {} bash -c 'copyTable "$@"' _ {}
+ASYNC_TABLES=$(echo "${ASYNC_TABLES//[\']/}" | tr "," "\n")
+
+declare tablesArray
+IFS=" " readarray -t tablesArray <<< "${TABLES}"
+tablesArray+=("topic_message")
```
jnels124, line 356: tables query already orders by size and this would be last in the list already. no need to specify here AND add to excluded tables

hedera-mirror-importer/src/main/resources/db/scripts/v2/migration.sh
```diff
@@ -1,69 +1,277 @@
...
+FLYWAY_DIR="/tmp/flyway"
 MIGRATIONS_DIR="${SCRIPTS_DIR}/../../migration/v2"
 export PATH="${FLYWAY_DIR}/:$PATH"
 source "${SCRIPTS_DIR}/migration.config"
 
+ASYNC_TABLE_SPLITS=${ASYNC_TABLE_SPLITS:-1000}
 export CONCURRENCY=${CONCURRENCY:-5}
-export MAX_TIMESTAMP="${MAX_TIMESTAMP}"
+export CONCURRENT_COPIES_PER_TABLE=${CONCURRENT_COPIES_PER_TABLE:-5}
+export MAX_TIMESTAMP="${MAX_TIMESTAMP:-$(querySource "select max(consensus_end) from record_file;")}"
```
jnels124, line 274: Should we store this value to a table? If stopped during the synchronous stage of the migration and restarted later we would have tables copying from different points (assuming the source has been ingesting files). Can also document in the readme to set this value if you are restarting the migration. (async tables are un-effected by this as they read from the status table.

hedera-mirror-importer/src/main/resources/db/scripts/v2/migration.sh
```diff
@@ -1,69 +1,278 @@
...
+
+  local where
+  if [[ "${columns}" =~ $TS_COLUMN_REGEX ]]; then
+    where="where consensus_timestamp \<= ${MAX_TIMESTAMP}"
+  # handle address book.
+  elif [[ "${columns}" =~ \<consensus_timestamp_start\> ]]; then
+    where="where consensus_timestamp_end \<= ${MAX_TIMESTAMP}"
+  fi
+
+  local sql="\\copy ${table}(${columns}) FROM PROGRAM 'echo copy \(select ${columns} from ${table} ${where}\) to STDOUT with BINARY|./querySource.sh' WITH BINARY;"
```
steven-sheehy, line 154: Why the need for querySource.sh? Can we not just run it in psql directly?

hedera-mirror-importer/src/main/resources/db/scripts/v2/migration.sh
```diff
@@ -1,69 +1,278 @@
...
+  columns="${columns#"${columns%%[![:space:]]*}"}" # Trim leading whitespace
+  echo "${columns}"
+}
+
+insertCursors() {
+  local hasTable="f"
+  local sqlTemp="/tmp/cursors-$$.csv"
+  local tableDDL="
+    create table if not exists async_migration_status (
+    the_table                varchar(100),
```
steven-sheehy, line 62:
```suggestion
    table_name                varchar(100),
```

hedera-mirror-importer/src/main/resources/db/scripts/v2/migration.sh
```diff
@@ -97,20 +307,25 @@ if [[ "${TARGET_DB_HEALTHY}" != "t" ]]; then
...
-wget -qO- ${FLYWAY_URL} | tar -xz && mv flyway-* flyway
+wget -qO- "${FLYWAY_URL}" | tar -xz && mv flyway-*/* .
 log "Flyway installed"
 
+CREATE_INDEXES_BEFORE_MIGRATION=${CREATE_INDEXES_BEFORE_MIGRATION:-true}
+if [[ "${CREATE_INDEXES_BEFORE_MIGRATION}" = "true" ]]; then
+  cp "${MIGRATIONS_DIR}/V2.0."* "${FLYWAY_DIR}"/sql/
+else
+  cp "${MIGRATIONS_DIR}/V2.0."[0-2]* "${FLYWAY_DIR}"/sql/
+fi
```
steven-sheehy, line 323: We should just always create before since we tested that was faster. Plus we'll soon have indexes created in new migrations.

steven-sheehy (changes requested):

hedera-mirror-importer/src/main/resources/db/scripts/v2/migration.sh
jnels124: We can probably not have the querySource.sh and instead say from program `psql -h source -c  "copy ... to stdout"`

hedera-mirror-importer/src/main/resources/db/scripts/v2/migration.sh
jnels124: Left this as an option because it does reduce the active time of the required time of connection between source and target. However not against removing it

docs/database/README.md
```diff
@@ -268,14 +268,19 @@ Following are the prerequisites and steps for migrating V1 data to V2.
...
-7. Run the [migration.sh](/hedera-mirror-importer/src/main/resources/db/scripts/v2/migration.sh) script.
-8. Update the mirror node configuration to point to the new Citus DB and start it.
+5. Stop the [Importer](/docs/importer/README.md) process. 
+6. Run the [migration.sh](/hedera-mirror-importer/src/main/resources/db/scripts/v2/migration.sh) script. Due to the time it will take to complete the migration, 
+   it is recommended to run the script in a way that doesn't require your terminal session to remain open.`: 
+   - run and redirect the output `./migration.sh > migrationout.txt 2> migration-error.txt &`
+   - Pause job `ctrl-z`
+   - Run in background `bg`
+   - Run `jobs` to get the correct job id 
+   - Disown the process `disown -h %1`. Replace 1 with the correct job id.
```
steven-sheehy, line 282: Isn't this simpler?

```suggestion
   `./migration.sh 2>&1 > migration.log & disown`
```

docs/database/README.md
```diff
@@ -268,14 +268,19 @@ Following are the prerequisites and steps for migrating V1 data to V2.
...
    [migration.config](/hedera-mirror-importer/src/main/resources/db/scripts/v2/migration.config).
-5. Get the correct version of [flyway](https://flywaydb.org/documentation/usage/commandline/) based on your OS and
+4. Get the correct version of [flyway](https://flywaydb.org/documentation/usage/commandline/) based on your OS and
    update it in the `FLYWAY_URL` field in the `migration.config` file. The default is set to the linux version.
-6. Stop the [Importer](/docs/importer/README.md) process.
-7. Run the [migration.sh](/hedera-mirror-importer/src/main/resources/db/scripts/v2/migration.sh) script.
-8. Update the mirror node configuration to point to the new Citus DB and start it.
+5. Stop the [Importer](/docs/importer/README.md) process. 
+6. Run the [migration.sh](/hedera-mirror-importer/src/main/resources/db/scripts/v2/migration.sh) script. Due to the time it will take to complete the migration, 
+   it is recommended to run the script in a way that doesn't require your terminal session to remain open.`:
```
steven-sheehy, line 277:
```suggestion
   it is recommended to run the script in a way that doesn't require your terminal session to remain open: 
```

docs/database/README.md
jnels124: Definitely simpler. I will try that out. I am not real familiar with the command and thought this was the simplest way.

jnels124: This worked fine. However, I left standard error and standard out as different log files. This version of the migration logs a bit more and it is preferable to make sure any errors are not lost in these logs.

steven-sheehy (approved):
LGTM

xin-hedera (approved):
LGTM

PR #8171, opened 2024-04-24
specify limits for each stackgres container

### LINKED ISSUE #8158: Specify Resource Limits For Each Container In Citus Pods
### Description

The scheduling of citus pods is inconsistent. This is because stackgres will configure different values for these containers across deployments even if the config is the same.

Need to prevent scheduling issues by explicitly setting the requests for each of the containers.

### Steps to reproduce

Deploy citus via stackgres various times and note the resources allocated to the individual containers on the citus pods.

### Additional context

_No response_

### Hedera network

other

### Version

v0.103.0-RC3

### Operating system

None

jnels124:
**Description**:
* specify cpu requests for all stackgres containers

**Related issue(s)**:

Fixes #8158 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/charts/hedera-mirror/values-prod.yaml
+++ b/charts/hedera-mirror/values-prod.yaml
@@ -132,7 +132,7 @@ stackgres:
       requests:
         containers:
           cluster-controller:
-            cpu: 300m
+            cpu: 500m
           envoy:
             cpu: 750m
           patroni:
@@ -154,7 +154,7 @@ stackgres:
       requests:
         containers:
           cluster-controller:
-            cpu: 300m
+            cpu: 500m
           envoy:
             cpu: 750m
           patroni:
--- a/charts/hedera-mirror/values-prod.yaml
+++ b/charts/hedera-mirror/values-prod.yaml
@@ -131,6 +131,14 @@ stackgres:
       memory: 24Gi
       requests:
         containers:
+          cluster-controller:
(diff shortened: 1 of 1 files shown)


charts/hedera-mirror/values-prod.yaml
```diff
@@ -131,6 +131,14 @@ stackgres:
       memory: 24Gi
       requests:
         containers:
+          cluster-controller:
+            cpu: 300m
+          envoy:
+            cpu: 750m
+          patroni:
+            cpu: 7350m
```
steven-sheehy, line 139: Does the resources.cpu=7350m need to be removed or adjusted?

jnels124: No it wouldn't need to be changed or removed. This is already the current behavior because our `dedicatedResources` config and specifying this value as the overall limit configured for the pod. However, after looking at metrics going to adjust these values here. Initial values were based off rough % used by stackgres. Our dashboards show we should allocate differently.

jnels124: From metrics appears we can configure envoy and pg bouncer to have much lower requests but left as is because don't want to go too low here without production like traffic

steven-sheehy (approved):

xin-hedera (approved):
LGTM

PR #8206, opened 2024-04-30
fix: Add a check so that prng and exchange rate precompiles won't acc…

### LINKED ISSUE #7636: Executing eth_call with value against PRNG/ExchangeRate precompile addresses returns success
### Description

Executing an internal call with value against PRNG/Exchange rate precompile addresses returns success. It should return 400 bad request as these addresses should not be able to receive hbars.

### Steps to reproduce

1. Create and deploy a contract with the following function:
```
    function makeCallWithAmount(address _to, bytes memory _data) external payable returns (bool success, bytes memory returnData) {
        (success, returnData) = _to.call{value: msg.value}(_data);
    }
```
2. Execute eth_call with the following body by replacing the `to` address with the contract address from step one.
```
{
    "from": "0x0000000000000000000000000000000000000002",
    "data": "0x91548228000000000000000000000000000000000000000000000000000000000000016900000000000000000000000000000000000000000000000000000000000000400000000000000000000000000000000000000000000000000000000000000004d83bf9a100000000000000000000000000000000000000000000000000000000",
    "estimate": false,
    "to": ,
    "value": 933333
}
```

**Expected result:** The endpoint should return 400 bad request with `INVALID_FEE_SUBMITTED`
**Actual result:** The endpoint returns 200 with valid response.

**Note:** This works for both PRNG and ExchangeRate precompiles

### Additional context

_No response_

### Hedera network

other

### Version

0.96.0

### Operating system

None

kselveliev:
**Description**:

Adds check in Prng and Exchange rate precompiles so that they are not able to accept value. Seems like services added the change only for prng but not for exchange rate so they need to add one more pr there for exchange rate.

After https://github.com/hashgraph/hedera-services/issues/11218 was fixed in services, we need to port the changes to mirror node.

Current pitfall of web3 is that even though with the new checks we throw `INVALID_FEE_SUBMITTED` status. The end result will be a `CONTRACT_REVERT_EXECUTED`. Future work will be a work package to improve error handling logic and return proper statuses.

* Add new check to `ExchangeRatePrecompiledContract` to not accept value
* Add new check to `PrngSystemPrecompiledContract` to not accept value
* Modify `PrngSystemPrecompiledContractTest` to work with the new changes.
* Modify `ContractCallSystemPrecompileTest` to add new integration tests to check that prng and exchange rate do not accept value, added new functions in solidity files to test that, and made a small refactoring of the test class

**Related issue(s)**:

Fixes #7636

**Checklist**


hedera-mirror-web3/src/test/resources/contracts/ExchangeRatePrecompile/ExchangeRatePrecompile.sol
```diff
@@ -1,22 +1,37 @@
...
+    }
+
+    function tinycentsToTinybarsPayable(uint256 tinycents) external returns (uint256 tinybars) {
+        (bool success, bytes memory result) = PRECOMPILE_ADDRESS.call{value : 1}(
+            abi.encodeWithSelector(TINYCENTS_TO_TINYBARS, tinycents));
+        require(success);
+        tinybars = abi.decode(result, (uint256));
+    }
+
+    function tinybarsToTinycentsPayable(uint256 tinybars) external returns (uint256 tinycents) {
```
bilyana-gospodinova, line 32: It is a little confusing that the names of the functions suggest that they are payable, the abi shows that they are not (and the payment is hardcoded) and then the tests suggest again that a payment is made but we pass no value and we need to check the contract in order to understand the flow. Maybe if we pass the payment to the test as value and then the value to the contract function it will be more easier to understand.

kselveliev: Comment addressed. Now in the new cases I'm passing value in the top level call. I've renamed the function names also. Initially I just copied the code from services and did not pay attention to the confusing naming.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallSystemPrecompileTest.java
```diff
@@ -92,6 +108,20 @@ void systemPrecompileFunctionsTestEthEstimateGas(final SystemContractFunctions c
...
     }
 
+    @ParameterizedTest
+    @EnumSource(ExchangeRateFunctionsWithValue.class)
+    void exchangeRatePrecompileFunctionsTestEthEstimateGasWithValueRevertExecution(
+            final ExchangeRateFunctionsWithValue contractFunc) {
+        final var functionHash = functionEncodeDecoder.functionHashFor(
+                contractFunc.name, EXCHANGE_RATE_PRECOMPILE_ABI_PATH, contractFunc.functionParameters);
+        final var serviceParameters = serviceParametersForExecution(
+                functionHash, EXCHANGE_RATE_PRECOMPILE_CONTRACT_ADDRESS, ETH_ESTIMATE_GAS, 0L, BlockType.LATEST);
```
bilyana-gospodinova, line 118: Actually, I meant this 0 value here as well. If in the contract we use `PRECOMPILE_ADDRESS.call{value: msg.value}`, we should not need to pass the value as a parameter to the function (unless we need specifically two different values) and this way we can have only 1 function that will cover the case when we pass value and when we don't. The same can be done for PRNG as well.

bilyana-gospodinova (changes requested):

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallSystemPrecompileTest.java
kselveliev: Now I understand. Comment addressed :).

bilyana-gospodinova (approved):
LGTM

xin-hedera (approved):
LGTM

PR #8236, opened 2024-05-07
Fix v2 Migration Logic

### LINKED ISSUE #8233: V2 Migration Issues
### Description

After running the migration on mainnet data, it was discovered that the data is not being split correctly amongst the chunks. Also, the migration fails to copy `file_data' as there is a consensus_timestamp with 0 saved to the database and there is no partition for this to go in v2. 

### Steps to reproduce

1. Run migration
2. After the async_migration_status table is created, look at the last range. Notice it is way off.

### Additional context

_No response_

### Hedera network

other

### Version

v0.104.0-RC2

### Operating system

None

jnels124:
**Description**:

* Bump spring-cloud-gcp-dependencies to 5.2.1
* Fix calculation to split work evenly across copy jobs
* Add a migration to remove file_data from bootstrap address_book 

**Related issue(s)**:

Fixes #8233

**Notes for reviewer**:

**Checklist**


hedera-mirror-importer/src/main/resources/db/scripts/v2/migration.sh
```diff
@@ -147,18 +147,21 @@ getCursorConditions() {
...
+  log "Will split data into ${pctPerCursor}% chunks"
   local stats=$(PGPASSWORD="${SOURCE_DB_PASSWORD}" psql -q --csv -t -h "${SOURCE_DB_HOST}" -d "${SOURCE_DB_NAME}" -p "${SOURCE_DB_PORT}" -U "${SOURCE_DB_USER}" -c "${query}")
-
   while IFS="," read -r pct_of_total lower_bound upper_bound; do
     upperBound=$((upper_bound > upperBound ? upper_bound : upperBound))
-    local groupRange=$((upperBound - lower_bound))
-    local rangeToTake=$(echo "scale=10; (${pctPerCursor} / 100) * ${groupRange}" | bc)
     totalPct=$(echo "(${totalPct} + ${pct_of_total})" | bc)
+    local groupRange=$((upperBound - lower_bound))
+    local rangeToTake=$(echo "scale=10; ${groupRange} / (${totalPct} / ${pctPerCursor})" | bc)
```
steven-sheehy, line 156: nit: I believe this syntax might be a bit simpler/efficient:

```suggestion
    local rangeToTake=$(bc -S 10 -e "${groupRange} / (${totalPct} / ${pctPerCursor}")
```

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.97.0__remove_bootstrap_addressbook.sql
```diff
@@ -0,0 +1,3 @@
+delete from file_data where consensus_timestamp=0 and entity_id=102;
+delete from address_book where start_consensus_timestamp=1;
+delete from address_book_entry where consensus_timestamp=1;
```
steven-sheehy, line 3: I'm not sure the impact on deleting these entries. Can we not just exclude these rows in the migration?

steven-sheehy (changes requested):

hedera-mirror-importer/src/main/resources/db/scripts/v2/migration.sh
```diff
@@ -147,18 +147,21 @@ getCursorConditions() {
...
   local -n cursors=$1 # Init array to hold cursor definitions
+  log "Will split data into ${pctPerCursor}% chunks"
   local stats=$(PGPASSWORD="${SOURCE_DB_PASSWORD}" psql -q --csv -t -h "${SOURCE_DB_HOST}" -d "${SOURCE_DB_NAME}" -p "${SOURCE_DB_PORT}" -U "${SOURCE_DB_USER}" -c "${query}")
-
   while IFS="," read -r pct_of_total lower_bound upper_bound; do
     upperBound=$((upper_bound > upperBound ? upper_bound : upperBound))
-    local groupRange=$((upperBound - lower_bound))
-    local rangeToTake=$(echo "scale=10; (${pctPerCursor} / 100) * ${groupRange}" | bc)
     totalPct=$(echo "(${totalPct} + ${pct_of_total})" | bc)
+    local groupRange=$((upperBound - lower_bound))
```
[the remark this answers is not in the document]
jnels124: the snake case vars are the db columns.

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.97.0__remove_bootstrap_addressbook.sql
jnels124: Yes after doing some more testing I am inclined to agree. With writing partial address books etc and starting stopping at various times there is some corner cases. We can safely delete the file_data which is the only thing causing an issue. Current code will not add this file_data entry.

steven-sheehy: I think we still need them on the target db. Can we exclude during migration and manually copy the difficult rows over via a psql command in the migration?

hedera-mirror-importer/src/main/resources/db/scripts/v2/migration.sh
jnels124: I can not make this syntax work and the man page doesn't specify these options.
```
NAME
       bc - An arbitrary precision calculator language

SYNTAX
       bc [ -hlwsqv ] [long-options] [  file ... ]

DESCRIPTION
       bc is a language that supports arbitrary precision numbers with interactive execution of statements.  There are some similarities in the syntax to the C programming language.  A standard math library is available by command line option.  If requested, the math library is defined before processing any files.  bc starts by processing code from all the files listed on the command line in the order listed.  After all files have been processed, bc reads from
       the standard input.  All code is executed as it is read.  (If a file contains a command to halt the processor, bc will never read from the standard input.)

       This version of bc contains several extensions beyond traditional bc implementations and the POSIX draft standard.  Command line options can cause these extensions to print a warning or to be rejected.  This document describes the language accepted by this processor.  Extensions will be identified as such.

   OPTIONS
       -h, --help
              Print the usage and exit.

       -i, --interactive
              Force interactive mode.

       -l, --mathlib
              Define the standard math library.

       -w, --warn
              Give warnings for extensions to POSIX bc.

       -s, --standard
              Process exactly the POSIX bc language.

       -q, --quiet
              Do not print the normal GNU bc welcome.

       -v, --version
              Print the version number and copyright and quit.

   NUMBERS
       The most basic element in bc is the number.  Numbers are arbitrary precision numbers.  This precision is both in the integer part and the fractional part.  All numbers are represented internally in decimal and all computation is done in decimal.  (This version truncates results from divide and multiply operations.)  There are two attributes of numbers, the length and the scale.  The length is the total number of decimal digits used by bc to represent  a
       number and the scale is the total number of decimal digits after the decimal point.  For example:
               .000001 has a length of 6 and scale of 6.
               1935.000 has a length of 7 and a scale of 3.
```

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.97.0__remove_bootstrap_addressbook.sql
jnels124: to add this row on the target we would need to create a default partition for`file_data`. To be clear without any changes, someone starting up a mirror node today will not have `file_data` with `consensus_timestamp=0`. At some point, previous code did add this to production data but not current code. However, current code will still save the `address_book` and `address_book_entry`'s.

I have updated the pr to only delete the `file_data`. This should produce a consistent db state if mirror node was stood up today and data ingested from genesis.

hedera-mirror-importer/src/main/resources/db/scripts/v2/migration.sh
steven-sheehy: Probably macos syntax differs from linux.

steven-sheehy (approved):
LGTM

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.97.0__remove_bootstrap_addressbook.sql
```diff
@@ -0,0 +1 @@
+delete from file_data where consensus_timestamp=0 and entity_id=102;
```
mgoelswirlds, line 1: A question : where are we ignoring saving the bootstrap address book?

jnels124: need to update that comment. We will save the address book and the entries. The existing code already doesn't save this `file_data` so the only change needed is to delete the existing "bad" row

mgoelswirlds (approved):
LGTM

edwin-greene (approved):
Looks good

jascks (approved):
LGTM

PR #8241, opened 2024-05-07
Implement opcode logger functionality

### LINKED ISSUE #8173: Implement opcode logger tracing logic
We are currently able to simulate transaction using the web3 module in the mirror-node. However we does not track and set anywhere the logs from executions in the form of opcodes.
As first step we should create a new list in the `ContractCallContext`, which is a list of objects with the following fields:
```
pc — the current program counter of the transaction, which represents the location of the current opcode being executed. A register that keeps track of the address of the next instruction to be executed in a program.
op — the current being executed, such as PUSH1, ADD, or SWAP1. A code that represents a specific operation or instruction that a computer or processor can understand and execute.
gas — the amount of gas remaining in the transaction at the current step of execution.
gasCost — the amount of gas consumed by the current opcode being executed.
depth — the current depth of the call stack for the transaction, reflecting the number of active frames representing nested function calls.
stack — the contents of the EVM stack at the current step of the transaction's execution. The stack is a data structure used by the EVM to manage and manipulate data during the execution of opcodes
memory — the contents of the memory at the current step of execution. A temporary data storage location that is available to smart contracts during execution.

```

We can reuse most of the flow used for calls, but we'd have to have our own implementation of the `tracePrecompileCall`, which will happen when we implement `OperationTracer` interface. Then for the current stack we can get the opcode using `currentFrame.getCurrentOperation().getOpcode()` and push inside the list we previously created in the ContractCallContext.

For Hedera Precompiles is more complex, as this approach won't work. If they are not successfull, we will populate wrong opcodes, therefore we'd need to check the sidecars for their result and set the opreration as `REVERT` at the right moment, in order to simulate the transaction 100% correctly.

konstantinabl:
**Description**:

The main goal of this PR is to add new tracer (OpcodeTracer) in the web3 module, in order to support opcodeLogger for debugging purposes.

Main files added:
* `OpcodeTracer` - tracer class, which has methods for tracing contract calls and uses the message frame to get the executed opcodes and save them
* `ContractionActionService` - service that queries the DB for relevant contract actions, used in the OpcodeTracer
* `EntityService` - service that finds an entity in the DB, used to find the evmAddress of an account

**Related issue(s)**:

Fixes #8173 

**Notes for reviewer**:


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/OpcodeTracer.java
```diff
@@ -0,0 +1,112 @@
...
+public class OpcodeTracer implements HederaEvmOperationTracer {
+
+    private OpcodeTracerOptions options;
+    private List<Opcode> opcodes;
+    private List<ContractAction> contractActions;
+
+    @Override
+    public void init(MessageFrame initialFrame) {
+        opcodes = new ArrayList<>();
+        ContractCallContext ctx = initialFrame.getContextVariable("ContractCallContext");
```
victor-yanev, line 31: Let's use `ContractCallContext.CONTEXT_NAME` here instead of "ContractCallContext"

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/OpcodeTracer.java
```diff
@@ -0,0 +1,112 @@
...
+
+    @Override
+    public void tracePostExecution(final MessageFrame frame, final Operation.OperationResult operationResult) {
+        final Optional<Bytes[]> memory = captureMemory(frame);
+        final Optional<Bytes[]> stackPostExecution = captureStack(frame);
+        final Optional<Map<UInt256, UInt256>> storage = captureStorage(frame);
+        OptionalLong gasCostOptional = operationResult.getGasCost() == 0 ? OptionalLong.empty() : OptionalLong.of(operationResult.getGasCost());
+        opcodes.add(new Opcode(
+                frame.getPC(),
+                Optional.of(frame.getCurrentOperation().getName()),
```
victor-yanev, line 44: If `frame.getCurrentOperation().getName()` is never going to be `null`, let's just directly use that instead of wrapping it in `Optional`. 

Usually, `Optional<T>` is used in cases where we want to explicitly show that it is possible for this field to be `null` and whoever uses it has is forced to handle the case of an empty optional in order to avoid a `NullPointerException` being thrown. For example, a method that looks up an entity from a database not find a match, in which case it would return `Optional.empty()` instead of `null`.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/OpcodeTracer.java
```diff
@@ -0,0 +1,112 @@
...
+        options = ctx.getOpcodeTracerOptions();
+        contractActions = ctx.getContractActions();
+    }
+
+    @Override
+    public void tracePostExecution(final MessageFrame frame, final Operation.OperationResult operationResult) {
+        final Optional<Bytes[]> memory = captureMemory(frame);
+        final Optional<Bytes[]> stackPostExecution = captureStack(frame);
+        final Optional<Map<UInt256, UInt256>> storage = captureStorage(frame);
+        OptionalLong gasCostOptional = operationResult.getGasCost() == 0 ? OptionalLong.empty() : OptionalLong.of(operationResult.getGasCost());
```
victor-yanev, line 41: Same for `gasCost`, let's just directly use `long` here instead of wrapping it in `OptionalLong`, unless we have a good reason to (e.g., if we have something special to execute when `gasCost == 0` and we want to force it)

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/OpcodeTracer.java
```diff
@@ -0,0 +1,112 @@
...
+
+    private Optional<Map<UInt256, UInt256>> captureStorage(final MessageFrame frame) {
+        if (!options.isStorage()) {
+            return Optional.empty();
+        }
+
+        try {
+            final Map<UInt256, UInt256> storageContents =
+                    new TreeMap<>(
+                            frame.getWorldUpdater().getAccount(frame.getRecipientAddress()).getUpdatedStorage());
```
victor-yanev, line 89: I see that it's possible for `WorldUpdater.getAccount` to return `null` if the account does not exist - let's add a null check here and log a warning in case it happens:

```java
final Address address = frame.getRecipientAddress()
final Account account = frame.getWorldUpdater().getAccount(address);

if (account == null) {
    log.warn("Failed to retrieve storage contents. Account not found in WorldUpdater");
    return Optional.of(new TreeMap<>());
}

return Optional.of(new TreeMap<>(account.getUpdatedStorage()));
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/OpcodeTracer.java
```diff
@@ -0,0 +1,112 @@
...
+
+            return Optional.of(storageContents);
+        } catch (final ModificationNotAllowedException e) {
+            return Optional.of(new TreeMap<>());
```
victor-yanev, line 93: Let's add a warning here for log traceability in case we catch a `ModificationNotAllowedException`:
```java
} catch (final ModificationNotAllowedException e) {
    log.warn(e.getMessage(), e);
    return Optional.of(new TreeMap<>());
}
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractActionService.java
```diff
@@ -0,0 +1,14 @@
...
+
+import java.util.List;
+
+public interface ContractActionService {
+
+    /**
+     * @param consensusTimestamp the consensus timestamp of a record
+     * @return the sidecar file associated with a record file
+     */
+    List<ContractAction> findContractActionByConsensusTimestamp(Long consensusTimestamp);
```
victor-yanev, line 13: We already get enough context that this is called from `ContractActionService`, we can shorten this:
`findContractActionByConsensusTimestamp` -> `findByConsensusTimestamp`.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractActionServiceImpl.java
```diff
@@ -0,0 +1,18 @@
...
+import java.util.List;
+
+@Service
+@RequiredArgsConstructor
+public class ContractActionServiceImpl implements ContractActionService{
+    private final ContractActionRepository contractActionRepository;
+    @Override
+    public List<ContractAction> findContractActionByConsensusTimestamp(Long consensusTimestamp) {
+        return contractActionRepository.findContractActionByTimestamp(consensusTimestamp);
+    }
```
victor-yanev, line 17: nit: let's fix the formatting here - add empty lines around the class, field and method declarations.

hedera-mirror-web3/src/main/java/com/hedera/node/app/service/evm/contracts/execution/HederaEvmTxProcessor.java
```diff
@@ -57,7 +59,9 @@ public class HederaEvmTxProcessor {
     protected final PricesAndFeesProvider livePricesSource;
     protected final Map<SemanticVersion, Provider<MessageCallProcessor>> mcps;
     protected final Map<SemanticVersion, Provider<ContractCreationProcessor>> ccps;
-    protected final HederaEvmOperationTracer tracer;
+
+    //TODO: do not import TRACER_TYPE from ProcessorImpl
```
victor-yanev, line 63: We can remove this TODO if it's not relevant anymore

hedera-mirror-web3/src/main/java/com/hedera/node/app/service/evm/contracts/execution/HederaEvmTxProcessor.java
```diff
@@ -210,4 +221,13 @@ private AbstractMessageProcessor getMessageProcessor(
...
     }
+
+    private HederaEvmOperationTracer getTracer(TracerType tracerType) {
+        return tracerMap.get(tracerType).get();
+    }
+
+    public enum TracerType {
+        OPCODE,
+        OPERATION
+    }
```
victor-yanev, line 232: Let's move this in a separate file `TracerType.java` in the same directory where `OpcodeTracer` and `MirrorOperationTracer` are.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/common/ContractCallContext.java
```diff
@@ -45,6 +49,15 @@ public class ContractCallContext {
...
     private CachingStateFrame<Object> stackBase;
 
+    @Setter
+    private OpcodeTracerOptions opcodeTracerOptions;
+
+    @Setter
+    private List<Opcode> opcodes = List.of();
+
+    @Setter
+    private List<ContractAction> contractActions;
```
victor-yanev, line 59: nit: We can default this to `List.of()`

victor-yanev (changes requested):
Nice job 🔥, this has been a tough task!

I've left a few suggestions below:

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/ContractActionRepository.java
```diff
@@ -0,0 +1,14 @@
...
+import org.springframework.data.jpa.repository.Query;
+import org.springframework.data.repository.CrudRepository;
+
+import java.util.List;
+import java.util.Optional;
+
+
+public interface ContractActionRepository extends CrudRepository<ContractAction, Long>{
+    @Query("select r from ContractAction r where r.consensusTimestamp = ?1")
+    List<ContractAction> findContractActionByTimestamp(long timestamp);
```
victor-yanev, line 13: nit: Same here - we can rename this to `findByConsensusTimestamp` or `findAllByConsensusTimestamp`.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractActionServiceImpl.java
```diff
@@ -0,0 +1,44 @@
...
+@RequiredArgsConstructor
+public class ContractActionServiceImpl implements ContractActionService{
+
+    private final ContractActionRepository contractActionRepository;
+    private final TransactionService transactionService;
+    private final EthereumTransactionService ethereumTransactionService;
+
+    @Override
+    public List<ContractAction> findFromTransaction(@NonNull @Valid TransactionIdOrHashParameter transactionIdOrHash) {
+        Assert.isTrue(transactionIdOrHash.isValid(), "Invalid transactionIdOrHash");
```
IvanKavaldzhiev, line 27: Isn't the @Valid annotation doing this check automatically?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
```diff
@@ -88,20 +97,9 @@ public String processCall(final CallServiceParameters params) {
                 if (params.isEstimate()) {
                     // eth_estimateGas initialization - historical timestamp is Optional.empty()
                     ctx.initializeStackFrames(store.getStackedStateFrames());
-                    result = estimateGas(params);
+                    result = estimateGas(params, ctx);
```
IvanKavaldzhiev, line 100: I think it would be cleaner if we have a dedicated ContractCallDebugService, which would be dedicated only for all existing and future debug related RPC calls. The workflow there is different - we replay an already executed transaction, while in ContractCallService we simulate a hypothetical transaction. So both services would have different responsibilities and needs.

If some components seem to be common for both services, they can be extracted in a parent Service class.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
```diff
@@ -88,20 +97,9 @@ public String processCall(final CallServiceParameters params) {
...
                 } else {
-                    BlockType block = params.getBlock();
-                    // if we have historical call then set corresponding file record
-                    if (block != BlockType.LATEST) {
```
IvanKavaldzhiev, line 95: In the case of `debug_traceTransaction` we won't have a defined block explicitly. We should find the `record_file` corresponding to the timestamp for the `transactionHash` that is sent as an input, so that the `ContractCallContext` contains proper data.

victor-yanev: This is already done when constructing the call service parameters, and the block number is passed to the ContractCallService.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
```diff
@@ -88,20 +97,9 @@ public String processCall(final CallServiceParameters params) {
...
-                        var recordFileOptional = recordFileService
-                                .findRecordFileByBlock(block)
-                                .orElseThrow(BlockNumberNotFoundException::new);
-                        ctx.setRecordFile(recordFileOptional);
-                    }
-                    // eth_call initialization - historical timestamp is Optional.of(recordFile.getConsensusEnd())
-                    // if the call is historical
-                    ctx.initializeStackFrames(store.getStackedStateFrames());
-                    final var ethCallTxnResult = doProcessCall(params, params.getGas(), true);
+                    final var ethCallTxnResult = callContract(params, TracerType.OPERATION, ctx);
```
IvanKavaldzhiev, line 102: That's a symptom of why we should split the logic in 2 different services. We shouldn't use any tracing in the case of `eth_call`, `eth_estimateGas`, so that we don't make this operations more CPU and memory intensive.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/TracerType.java
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hedera.mirror.web3.evm.contracts.execution.traceability;
+
+public enum TracerType {
+    OPCODE,
+    OPERATION
```
IvanKavaldzhiev, line 21: For `OPERATION` we use directly the `json-rpc-relay` implementation, which utilises the /actions endpoint. The REST API in the web3 module should be used only for `OPCODE` tracing. I think we should delete the `OPERATION` type.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
victor-yanev: I agree, but the other operation tracer was used previously as well, we haven’t introduced any changes to the other methods.

Previously, the `MirrorOperationTracer` was injected directly into the processor and now we inject a `Map<TracerType, HederaEvmTracer>` instead, and thats why we need to pass the tracer type here, in order to get the right tracer that we want to use.

If we don’t want to do any tracing in `eth_call` and `eth_estimateGas`, we can create a new `TracerType.NONE` that’s used by default and doesn’t invoke any tracers when processing the call.

What’s your opinion here?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/TracerType.java
victor-yanev: https://github.com/hashgraph/hedera-mirror-node/pull/8241#discussion_r1620414264

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/EntityServiceImpl.java
```diff
@@ -0,0 +1,18 @@
...
+import com.hedera.mirror.common.domain.entity.Entity;
+import com.hedera.mirror.web3.repository.EntityRepository;
+import java.util.Optional;
+import lombok.RequiredArgsConstructor;
+import org.springframework.stereotype.Service;
+
+@Service
+@RequiredArgsConstructor
+public class EntityServiceImpl implements EntityService {
+    private final EntityRepository entityRepository;
```
IvanKavaldzhiev, line 12: I think we can directly use the `EntityDatabaseAccessor` for the moment and delete this Service definition.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
konstantinabl: I agree
Will do.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
IvanKavaldzhiev: We should use `MirrorOperationTracer` for `eth_call` and `eth_estimateGas`, since it enables logging capabilities that can be turned on by a special flag.

I thought that the TRACER type is for the callTracer type that we already have in place. I agree with your approach. 🙌

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractActionServiceImpl.java
```diff
@@ -0,0 +1,30 @@
...
+import java.util.List;
+import java.util.Optional;
+import lombok.NonNull;
+import lombok.RequiredArgsConstructor;
+import org.springframework.stereotype.Service;
+import org.springframework.util.Assert;
+
+@Service
+@RequiredArgsConstructor
+public class ContractActionServiceImpl implements ContractActionService{
```
IvanKavaldzhiev, line 21: nit: ContractActionService{ -> ContractActionService {

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/BaseService.java
```diff
@@ -0,0 +1,100 @@
...
+import static com.hedera.mirror.web3.convert.BytesDecoder.maybeDecodeSolidityErrorStringToReadableMessage;
+import static com.hedera.mirror.web3.evm.exception.ResponseCodeUtil.getStatusOrDefault;
+import static com.hedera.mirror.web3.service.ContractCallService.GAS_METRIC;
+import static com.hedera.mirror.web3.service.model.BaseCallServiceParameters.CallType.ERROR;
+import static com.hedera.mirror.web3.service.model.BaseCallServiceParameters.CallType.ETH_DEBUG_TRACE_TRANSACTION;
+import static org.apache.logging.log4j.util.Strings.EMPTY;
+
+@Named
+@CustomLog
+public class BaseService {
```
IvanKavaldzhiev, line 29: This name is too generic. We should have a given common context for all of its successors.

Probably we can rename it to `BaseService` -> `ContractCallService`
And the existing `ContractCallService` could be renamed to `ContractExecutionService`

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallDebugService.java
```diff
@@ -0,0 +1,95 @@
...
+import jakarta.inject.Named;
+import lombok.CustomLog;
+
+import java.util.List;
+
+@CustomLog
+@Named
+public class ContractCallDebugService extends BaseService {
+
+    private final ContractActionService contractActionService;
```
IvanKavaldzhiev, line 28: If we need `contractActionRepository` only here, I think we can directly use the repository on this service level, instead of wrapping it in a service.

konstantinabl: Ok, what if another service needs it in the future? I mean its kind of an overkill now, but could be useful in the future

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/model/BaseCallServiceParameters.java
```diff
@@ -0,0 +1,26 @@
...
+public interface BaseCallServiceParameters {
+    HederaEvmAccount getSender();
+    Address getReceiver();
+    long getGas();
+    long getValue();
+    Bytes getCallData();
+    boolean isStatic();
+    boolean isEstimate();
+    CallType getCallType();
+    BlockType getBlock();
```
victor-yanev, line 18: Let's order these alphabetically (same for the classes that implement this interface).

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/model/ContractCallDebugServiceParameters.java
```diff
@@ -0,0 +1,25 @@
...
+import com.hedera.mirror.web3.viewmodel.BlockType;
+import com.hedera.node.app.service.evm.store.models.HederaEvmAccount;
+import lombok.Builder;
+import lombok.RequiredArgsConstructor;
+import lombok.Value;
+import org.apache.tuweni.bytes.Bytes;
+import org.hyperledger.besu.datatypes.Address;
+
+@Value
+@Builder
```
victor-yanev, line 12: It's used by the `OpcodeController` tests
```suggestion
@Builder(toBuilder = true)
```

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/controller/OpcodesControllerTest.java
```diff
@@ -740,7 +737,7 @@ RecordFileService recordFileService(final RecordFileRepository recordFileReposit
 
         @Bean
         OpcodeService opcodeService(final RecordFileService recordFileService,
-                                    final ContractCallService contractCallService,
+                                    final ContractDebugService contractExecutionService,
```
victor-yanev, line 740: `contractExecutionService` -> `contractDebugService`

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractExecutionServicePrecompileTest.java
```diff
@@ -45,7 +45,7 @@
 import org.junit.jupiter.params.provider.MethodSource;
 import org.junit.jupiter.params.provider.ValueSource;
 
-class ContractCallServicePrecompileTest extends ContractCallTestSetup {
+class ContractExecutionServicePrecompileTest extends ContractCallTestSetup {
```
victor-yanev, line 48: No need to rename this (same for the other test files which were renamed). Same reason as above.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallTestSetup.java
```diff
@@ -500,7 +501,7 @@ public class ContractCallTestSetup extends Web3IntegrationTest {
     protected FunctionEncodeDecoder functionEncodeDecoder;
 
     @Autowired
-    protected ContractCallService contractCallService;
+    protected ContractExecutionService contractExecutionService;
```
victor-yanev, line 504: No need to rename this field, it's cluttering the PR with empty changes making it harder to review. 

Just change the type.

hedera-mirror-web3/src/main/java/com/hedera/node/app/service/evm/contracts/execution/HederaEvmTxProcessor.java
```diff
@@ -153,6 +160,9 @@ public HederaEvmTransactionProcessingResult execute(
 
         tracer.finalizeOperation(initialFrame);
 
+        if (tracer instanceof OpcodeTracer opcodeTracer) {
+            contractCallContext.setOpcodes(opcodeTracer.getOpcodes());
+        }
```
victor-yanev, line 165: We can directly update the context inside `OpcodeTracer.finalizeOperation` instead of doing it here. In this way we delegate this to the tracer and we don't need to be doing `instanceof` custom checks.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
```diff
@@ -1,171 +1,57 @@
...
-    private final MeterProvider<Counter> gasUsedCounter;
-    private final Store store;
     private final MirrorEvmTxProcessor mirrorEvmTxProcessor;
-    private final RecordFileService recordFileService;
-    private final ThrottleProperties throttleProperties;
     private final Bucket gasLimitBucket;
+    private final ThrottleProperties throttleProperties;
+    private final Meter.MeterProvider<Counter> gasCounter;
 
     public ContractCallService(
```
victor-yanev, line 35:
```suggestion
    protected ContractCallService(
```

victor-yanev (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallDebugService.java
konstantinabl: changed it.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
```diff
@@ -190,19 +76,24 @@ private void restoreGasToBucket(HederaEvmTransactionProcessingResult result, lon
...
+            if (type == ETH_DEBUG_TRACE_TRANSACTION) {
+                log.warn("Transaction failed with status: {}, detail: {}, revertReason: {}",
+                        getStatusOrDefault(txnResult), detail, revertReason.toHexString());
+            } else {
+                throw new MirrorEvmTransactionException(getStatusOrDefault(txnResult), detail, revertReason.toHexString());
+            }
         } else {
             updateGasMetric(type, txnResult.getGasUsed(), 1);
         }
     }
```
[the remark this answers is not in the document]
konstantinabl: Won't this create a lot of complexity, since I need some info like _**detail**_ to throw the error or log a warning. I will possibly need to create a method handleError, which should be inherited in each class and pass the details from the parent. While like this, have the callType anyway :?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
```diff
@@ -190,19 +76,24 @@ private void restoreGasToBucket(HederaEvmTransactionProcessingResult result, lon
...
 
-    private void updateGasMetric(final CallType callType, final long gasUsed, final int iterations) {
-        gasUsedCounter
+    protected void updateGasMetric(final ContractCallDebugServiceParameters.CallType callType, final long gasUsed, final int iterations) {
```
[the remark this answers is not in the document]
konstantinabl: Its a mistake while refactoring

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
konstantinabl: You suggested it should be abstract
https://github.com/hashgraph/hedera-mirror-node/pull/8241#discussion_r1625623001
So it cannot be abstract and protected at the same time, right. Its a top-level class

konstantinabl: https://github.com/hashgraph/hedera-mirror-node/pull/8241#discussion_r1625623001
You suggested it should be abstract, so it cannot be protected as well, right? Its a top-level class

victor-yanev: An abstract class is a restricted class that cannot be used to create objects (to access it, it must be inherited from another class)

And that’s what we want for our base class, it doesn't make sense to create objects of the base class, we only want to use it as a template for the subclasses.

konstantinabl: my bad, i thought you suggested this change for the class not the constuctor :)

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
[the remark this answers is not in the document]
victor-yanev:
```java
    @Override
    protected void validateResult(final HederaEvmTransactionProcessingResult txnResult, final CallServiceParameters.CallType type) {
        try {
            super.validateResult(txnResult, type);
        } catch (MirrorEvmTransactionException e) {
            log.warn("Transaction failed with status: {}, detail: {}, revertReason: {}",
                    getStatusOrDefault(txnResult), e.getDetail(), e.getData());
        }
    }
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmTxProcessorImpl.java
```diff
@@ -90,7 +92,10 @@ public MirrorEvmTxProcessorImpl(
         this.tokenAccessor = tokenAccessor;
     }
 
-    public HederaEvmTransactionProcessingResult execute(CallServiceParameters params, long estimatedGas) {
+    public HederaEvmTransactionProcessingResult execute(final CallServiceParameters params,
+                                                        final long estimatedGas,
+                                                        final TracerType tracerType,
+                                                        final ContractCallContext ctx) {
```
victor-yanev, line 98:
```suggestion
    public HederaEvmTransactionProcessingResult execute(final CallServiceParameters params,
                                                        final long estimatedGas,
                                                        final ContractCallContext ctx) {
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
```diff
@@ -41,131 +16,86 @@
...
 import org.apache.tuweni.bytes.Bytes;
 
-@CustomLog
-@Named
-public class ContractCallService {
+import static com.hedera.mirror.web3.convert.BytesDecoder.maybeDecodeSolidityErrorStringToReadableMessage;
+import static com.hedera.mirror.web3.evm.exception.ResponseCodeUtil.getStatusOrDefault;
+import static com.hedera.mirror.web3.service.model.CallServiceParameters.CallType.ERROR;
+import static com.hedera.mirror.web3.service.model.CallServiceParameters.CallType.ETH_DEBUG_TRACE_TRANSACTION;
+import static org.apache.logging.log4j.util.Strings.EMPTY;
```
victor-yanev, line 26: Please copy these settings in your IDE to avoid cluttering the PR with changes related to import formatting:

image

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
```diff
@@ -41,131 +16,86 @@
...
-    private final MirrorEvmTxProcessor mirrorEvmTxProcessor;
     private final RecordFileService recordFileService;
-    private final ThrottleProperties throttleProperties;
-    private final Bucket gasLimitBucket;
 
-    public ContractCallService(
-            MeterRegistry meterRegistry,
-            BinaryGasEstimator binaryGasEstimator,
-            Store store,
+    protected final Store store;
```
victor-yanev, line 40: nit: Let's preserve the old order here:

```java
     private final MeterProvider<Counter> gasLimitCounter;
     private final MeterProvider<Counter> gasUsedCounter;
     private final Store store;
     private final MirrorEvmTxProcessor mirrorEvmTxProcessor;
     private final RecordFileService recordFileService;
     private final ThrottleProperties throttleProperties;
     private final Bucket gasLimitBucket;
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
```diff
@@ -41,131 +16,86 @@
...
                 .withRegistry(meterRegistry);
         this.gasUsedCounter = Counter.builder(GAS_USED_METRIC)
                 .description("The amount of gas consumed by the EVM")
                 .withRegistry(meterRegistry);
-        this.store = store;
-        this.mirrorEvmTxProcessor = mirrorEvmTxProcessor;
-        this.recordFileService = recordFileService;
-        this.throttleProperties = throttleProperties;
-        this.gasLimitBucket = gasLimitBucket;
-    }
```
victor-yanev, line 84: nit: same here

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
```diff
@@ -41,131 +16,86 @@
...
 
-    private HederaEvmTransactionProcessingResult doProcessCall(
-            CallServiceParameters params, long estimatedGas, boolean restoreGasToThrottleBucket) {
+    protected HederaEvmTransactionProcessingResult doProcessCall(CallServiceParameters params,
+                                                                 long estimatedGas,
+                                                                 boolean restoreGasToThrottleBucket,
+                                                                 ContractCallContext ctx) throws MirrorEvmTransactionException {
         try {
-            var result = mirrorEvmTxProcessor.execute(params, estimatedGas);
+            var result = mirrorEvmTxProcessor.execute(params, estimatedGas, params.getTracerType(), ctx);
```
victor-yanev, line 98: We already have the tracer type in the params:
```suggestion
            var result = mirrorEvmTxProcessor.execute(params, estimatedGas, ctx);
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
```diff
@@ -190,20 +120,26 @@ private void restoreGasToBucket(HederaEvmTransactionProcessingResult result, lon
         }
     }
 
-    private void validateResult(final HederaEvmTransactionProcessingResult txnResult, final CallType type) {
+    protected void validateResult(final HederaEvmTransactionProcessingResult txnResult, final ContractExecutionParameters.CallType type) {
```
victor-yanev, line 123: nit: use static import for `ContractCallParameters.CallType` and change this to:
```suggestion
    protected void validateResult(final HederaEvmTransactionProcessingResult txnResult, final CallType type) {
```

Same for all other occurrences.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractActionServiceImpl.java
```diff
@@ -0,0 +1,25 @@
...
+import jakarta.validation.Valid;
+import java.util.List;
+
+import lombok.NonNull;
+import lombok.RequiredArgsConstructor;
+import org.springframework.stereotype.Service;
+
+@Service
+@RequiredArgsConstructor
+public class ContractActionServiceImpl implements ContractActionService {
```
victor-yanev, line 16: Remove this

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractActionService.java
```diff
@@ -0,0 +1,17 @@
+package com.hedera.mirror.web3.service;
+
+import com.hedera.mirror.common.domain.contract.ContractAction;
+import com.hedera.mirror.web3.common.TransactionIdOrHashParameter;
+import com.hedera.mirror.web3.service.model.ContractDebugParameters;
+import jakarta.validation.Valid;
+import java.util.List;
+import lombok.NonNull;
+
+public interface ContractActionService {
```
victor-yanev, line 10: Remove this

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractExecutionService.java
```diff
@@ -0,0 +1,117 @@
...
+import java.util.Objects;
+import lombok.CustomLog;
+import org.apache.tuweni.bytes.Bytes;
+
+@CustomLog
+@Named
+public class ContractExecutionService extends ContractCallService {
+
+    static final String GAS_LIMIT_METRIC = "hedera.mirror.web3.call.gas.limit";
+    static final String GAS_USED_METRIC = "hedera.mirror.web3.call.gas.used";
```
victor-yanev, line 42: We should have this in our base class already. Please, remove these from here.

hedera-mirror-web3/src/main/java/com/hedera/node/app/service/evm/contracts/execution/HederaEvmTxProcessor.java
victor-yanev: nit: Reminder

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallTestSetup.java
victor-yanev: nit: reminder

victor-yanev (dismissed):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/OpcodeTracer.java
```diff
@@ -0,0 +1,161 @@
...
+        if (!options.isStack()) {
+            return Collections.emptyList();
+        }
+
+        final Bytes[] stackContents = new Bytes[frame.stackSize()];
+        for (int i = 0; i < stackContents.length; i++) {
+            // Record stack contents in reverse
+            stackContents[i] = frame.getStackItem(stackContents.length - i - 1);
+        }
+        return Arrays.asList(stackContents);
```
steven-sheehy, line 128: Why are we creating an array just to convert it to a list?

```suggestion
        int size = frame.stackSize();
        var stack = new ArrayList<Bytes>(size);

        for (int i = 0; i < size; ++i) {
            stack.add(frame.getStackItem(size - 1 - i));
        }

        return stack;
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/OpcodeTracer.java
```diff
@@ -0,0 +1,161 @@
...
+    private List<Bytes> captureMemory(final MessageFrame frame) {
+        if (!options.isMemory()) {
+            return Collections.emptyList();
+        }
+
+        final Bytes[] memoryContents = new Bytes[frame.memoryWordSize()];
+        for (int i = 0; i < memoryContents.length; i++) {
+            memoryContents[i] = frame.readMemory(i * 32L, 32);
+        }
+        return Arrays.asList(memoryContents);
```
steven-sheehy, line 115: Same comment. Don't use arrays

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/OpcodeTracer.java
```diff
@@ -0,0 +1,161 @@
...
+                frame.getCurrentOperation() != null
+                        ? frame.getCurrentOperation().getName()
+                        : StringUtils.EMPTY,
+                frame.getRemainingGas(),
+                output != null ? gasRequirement : 0L,
+                frame.getDepth(),
+                Collections.emptyList(),
+                Collections.emptyList(),
+                Collections.emptyMap(),
+                revertReason.map(Bytes::toString).orElse(null)));
```
steven-sheehy, line 98: This is brittle and prone to mapping mistakes. Add `@Builder` to Opcode and use builder methods to build. Same for other trace method

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/OpcodeTracer.java
```diff
@@ -0,0 +1,161 @@
...
+                storage,
+                frame.getRevertReason().map(Bytes::toString).orElse(null)));
+    }
+
+    @Override
+    public void tracePrecompileCall(final MessageFrame frame, final long gasRequirement, final Bytes output) {
+        final Optional<Bytes> revertReason =
+                frame.getRevertReason().isPresent() ? frame.getRevertReason() : getRevertReason(contractActions);
+
+        revertReason.ifPresent(bytes -> log.trace("Revert reason: {}", bytes.toHexString()));
```
steven-sheehy, line 85: This lines seems pretty pointless

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/OpcodeTracer.java
```diff
@@ -0,0 +1,161 @@
...
+                stack,
+                memory,
+                storage,
+                frame.getRevertReason().map(Bytes::toString).orElse(null)));
+    }
+
+    @Override
+    public void tracePrecompileCall(final MessageFrame frame, final long gasRequirement, final Bytes output) {
+        final Optional<Bytes> revertReason =
+                frame.getRevertReason().isPresent() ? frame.getRevertReason() : getRevertReason(contractActions);
```
steven-sheehy, line 83: See other comment, but another point is even if we somehow need contract actions this doesn't map it to a specific action by index. It returns a random action.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/model/ContractExecutionParameters.java
```diff
@@ -0,0 +1,40 @@
+/*
+ * Copyright (C) 2019-2024 Hedera Hashgraph, LLC
```
steven-sheehy, line 2:
```suggestion
 * Copyright (C) 2024 Hedera Hashgraph, LLC
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/model/ContractExecutionParameters.java
```diff
@@ -0,0 +1,40 @@
...
+    BlockType block;
+    Bytes callData;
+    CallType callType;
+    long gas;
+    boolean isEstimate;
+    boolean isStatic;
+    Address receiver;
+    HederaEvmAccount sender;
+    TracerType tracerType = TracerType.OPERATION;
+    long value;
```
steven-sheehy, line 39: nit: Add `private final` even if `@Value` adds it implicitly so it's clear.

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/ContractActionRepository.java
```diff
@@ -0,0 +1,26 @@
...
+
+package com.hedera.mirror.web3.repository;
+
+import com.hedera.mirror.common.domain.contract.ContractAction;
+import java.util.List;
+import org.springframework.data.repository.CrudRepository;
+
+public interface ContractActionRepository extends CrudRepository<ContractAction, Long> {
+
+    List<ContractAction> findAllByConsensusTimestamp(long consensusTimestamp);
```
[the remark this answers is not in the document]
konstantinabl: IMO findAll makes more sense, since we return a List of ContractActions, not a single one

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/model/ContractDebugParameters.java
```diff
@@ -0,0 +1,43 @@
...
+    CallType callType = CallType.ETH_DEBUG_TRACE_TRANSACTION;
+    long consensusTimestamp;
+    long gas;
+    boolean isEstimate = false;
+    boolean isStatic = false;
+    Address receiver;
+    HederaEvmAccount sender;
+    TracerType tracerType = TracerType.OPCODE;
+    long value;
+}
```
victor-yanev, line 43:
```suggestion
public class ContractDebugParameters implements CallServiceParameters {

     @NotNull
     BlockType block;

     @NotNull
     Bytes callData;

     @NotNull
     CallType callType = CallType.ETH_DEBUG_TRACE_TRANSACTION;

     @Positive
     long consensusTimestamp;

     @PositiveOrZero
     long gas;

     @AssertFalse
     boolean isEstimate = false;

     @AssertFalse
     boolean isStatic = false;

     @NotNull
     Address receiver;

     @NotNull
     HederaEvmAccount sender;

     @NotNull
     TracerType tracerType = TracerType.OPCODE;

     @PositiveOrZero
     long value;
 }
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractDebugService.java
```diff
@@ -0,0 +1,81 @@
...
+import com.hedera.node.app.service.evm.contracts.execution.HederaEvmTransactionProcessingResult;
+import io.github.bucket4j.Bucket;
+import io.micrometer.core.instrument.MeterRegistry;
+import jakarta.inject.Named;
+import java.util.List;
+import lombok.CustomLog;
+
+@CustomLog
+@Named
+public class ContractDebugService extends ContractCallService {
```
victor-yanev, line 41:
```suggestion
@CustomLog
@Named
@Validated
public class ContractDebugService extends ContractCallService {
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractDebugService.java
```diff
@@ -0,0 +1,81 @@
...
+            MirrorEvmTxProcessor mirrorEvmTxProcessor,
+            Bucket gasLimitBucket,
+            ThrottleProperties throttleProperties,
+            MeterRegistry meterRegistry) {
+        super(mirrorEvmTxProcessor, gasLimitBucket, throttleProperties, meterRegistry, recordFileService, store);
+        this.contractActionRepository = contractActionRepository;
+    }
+
+    public OpcodesProcessingResult processOpcodeCall(
+            final ContractDebugParameters params, final OpcodeTracerOptions opcodeTracerOptions) {
```
victor-yanev, line 57:
```suggestion
            final @Validate ContractDebugParameters params, final OpcodeTracerOptions opcodeTracerOptions) {
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/OpcodeServiceImpl.java
```diff
@@ -139,35 +139,34 @@ private OpcodesResponse buildOpcodesResponse(@NonNull OpcodesProcessingResult re
...
                 .map(recordFile -> BlockType.of(recordFile.getIndex().toString()))
                 .orElse(BlockType.LATEST);
 
-        return CallServiceParameters.builder()
+        return ContractDebugParameters.builder()
                 .sender(new HederaEvmAccount(getSenderAddress(contractResult)))
                 .receiver(getReceiverAddress(ethTransaction, contractResult))
                 .gas(getGasLimit(ethTransaction, contractResult))
                 .value(getValue(ethTransaction, contractResult).longValue())
                 .callData(getCallData(ethTransaction, contractResult))
```
victor-yanev, line 169:
```suggestion
                .callData(getCallData(ethTransaction, contractResult))
                .consensusTimestamp(consensusTimestamp)
```

victor-yanev (changes requested):
After adding `consensusTimestamp` to the `ContractDebugParameters`, we are not passing it in the `OpcodeService#buildCallServiceParameters` method and as a consequence the `contractActions` list is always empty in the `ContractCallContext`.

Let's add some validations to the `processOpcodeCall` method to catch such issues and any potential bugs in the future.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/model/ContractDebugParameters.java
steven-sheehy: It's quite inefficient and redundant to validate user input twice just because there was a temporary mapping bug in a PR. I think the main problem here is that you are doing integration tests in a separate PR so you're not catching things in this one.

victor-yanev: We are not validating user input here, but it’s more like a validation which checks that the debug service parameters that we have built are valid. 

If the validation fails here, it means that we have a bug in the logic and it’s an early sign that something is not okay. If we remove this validation the bug wont be caught anywhere and the method will silently continue despite receiving invalid input. It’s a good practice to have such checks instead of completely relying on tests to cover all edge cases.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/OpcodeTracer.java
konstantinabl: Fixed this. Since we cannot know on exactly which contract action we are currently just by checking the MessageFrame, I have implemented a counter, similar to actionStack in services, so we know which operation we are executing and can get the correct revertreason from the DB

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/OpcodeTracer.java
```diff
@@ -0,0 +1,161 @@
...
+
+    private static Optional<Bytes> getRevertReason(List<ContractAction> contractActions) {
+        if (CollectionUtils.isEmpty(contractActions)) {
+            return Optional.empty();
+        }
+        return contractActions.stream()
+                .filter(ContractAction::hasRevertReason)
+                .map(action -> Bytes.of(action.getResultData()))
+                .findFirst();
+    }
```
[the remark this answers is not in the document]
konstantinabl: Reverted hedera precompile calls are not kept in the message frame, so we are not able to retrieve them. 
Explained how they are retrieved now here -> https://github.com/hashgraph/hedera-mirror-node/pull/8241#discussion_r1643375879

victor-yanev (approved):
LGTM, left a few nits below:

victor-yanev:
Please, also make sure that we have covered all of the new logic in the `OpcodeTracerTest` file. (e.g., the logic with the indexing of the contract actions - we can mock several contract actions, with one being a reverting call to HTS -> make sure that we are indexing the correct contract action for the message frame)

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/OpcodeTracer.java
```diff
@@ -0,0 +1,176 @@
...
+public class OpcodeTracer implements HederaEvmOperationTracer {
+
+    @Override
+    public void traceContextEnter(final MessageFrame frame) {
+        getContext().incrementContractActionsCounter();
+    }
+
+    @Override
+    public void traceContextReEnter(final MessageFrame frame) {
+        getContext().decrementContractActionsCounter();
```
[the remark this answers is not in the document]
konstantinabl: I think you migt have been misled by the name of the variable for the counter, namely **_contractActionsCounter_**. We are not counting the contract_actions.index, but as we discussed previously, we need to know the index of the exact contract action that corresponds to our message frame, so we can take the revert reason, since in the frame we don't have revert reason set for Hedera precompiles. Why we can't use the tracePostExecution - because it tracks some additional intermediate frames, which are not added in contract_actions and the counter will not be correct. You can see in services they have a method called pushActionOfIntermediate. In general, we wanted to replicate the behaviour in services, so we have the same counter for the actions and can map one to the current frame. This is the idea of this variable, to track the index we are at currently. 
This is the solution we've come up with, and we have discussed this approach with @IvanKavaldzhiev, as you suggested. I hope my explanation is clear. If you have any further questions, let's discuss them!

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/common/ContractCallContext.java
```diff
@@ -32,6 +37,17 @@ public class ContractCallContext {
     public static final String CONTEXT_NAME = "ContractCallContext";
     private static final ScopedValue<ContractCallContext> SCOPED_VALUE = ScopedValue.newInstance();
 
+    @Setter
+    private List<ContractAction> contractActions = List.of();
+
+    @Setter
+    private int contractActionsCounter = 0;
```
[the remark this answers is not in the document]
konstantinabl: Please refer to my comment https://github.com/hashgraph/hedera-mirror-node/pull/8241#discussion_r1646094392

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/OpcodeTracer.java
```diff
@@ -0,0 +1,176 @@
...
+                .reason(frame.getRevertReason().map(Bytes::toString).orElse(null))
+                .build();
+
+        context.addOpcodes(opcode);
+    }
+
+    @Override
+    public void tracePrecompileCall(final MessageFrame frame, final long gasRequirement, final Bytes output) {
+        ContractCallContext context = getContext();
+        Optional<Bytes> revertReason = isCallToHederaTokenService(frame) ? getRevertReasonFromContractActions(context.getContractActions()) : frame.getRevertReason();
```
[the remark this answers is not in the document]
konstantinabl: Addressed this here -> https://github.com/hashgraph/hedera-mirror-node/pull/8241/files#diff-a52bbc27042361685a4dce7f46214370ba30aa442624dd3b6f1fb6063a77efa2R26
Are you okay with this approach?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/OpcodeTracer.java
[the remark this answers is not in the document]
IvanKavaldzhiev: `pushActionOfIntermediate` is a concept of the new `EvmActionTracer`, which is part of the `hedera-smart-contract-service` (from mod code). Web3 based its logic on the mono code base and we have copied some code base from there. Even though now services uses the mod code and we are still reusing the mono code - we should align the `OpcodeTracer` logic with the mono code to be consistent.

When we integrate the `hedera-smart-contract-service` into web3, then we can revisit the OpcodeTracer implementation.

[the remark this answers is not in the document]
IvanKavaldzhiev: I think we should follow the logic for `List<SolidityAction> allActions` inside `HederaTracer`. In this list we always add new sidecars that are then exported to mirror-node. As Steven mentioned, sidecar action's index is always incrementing, so I think we should follow the places where we add new `sidecar` in `allActions` in `HederaTracer` and in mirror-node's `OpcodeTracer` just increment this counter in the same places.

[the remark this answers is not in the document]
IvanKavaldzhiev: The only time we remove sidecars from `allActions` is removing the so-called `invalid` actions, which I'm not sure if will be applicable for the case of `debug_traceTransaction`, since we skip some business and validation logic for the sake of simulation. This could be further analyzed.

steven-sheehy:
Some code smells to address

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/OpcodeTracer.java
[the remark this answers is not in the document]
steven-sheehy: Yes, I was also looking at HederaTracer. Looks like actions are generated in tracePrecompileResult, traceAccountCreationResult, and tracePostExecution. And my other comment was if we have a list of actions (including precompiled actions) we generated then that list provides the counter and we don't need another counter.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/OpcodeTracer.java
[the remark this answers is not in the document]
steven-sheehy: Works for me

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/MirrorEvmTxProcessor.java
```diff
@@ -16,10 +16,12 @@
...
 
+import com.hedera.mirror.web3.common.ContractCallContext;
 import com.hedera.mirror.web3.service.model.CallServiceParameters;
 import com.hedera.node.app.service.evm.contracts.execution.HederaEvmTransactionProcessingResult;
 
 public interface MirrorEvmTxProcessor {
 
-    HederaEvmTransactionProcessingResult execute(CallServiceParameters params, long estimatedGas);
+    HederaEvmTransactionProcessingResult execute(
+            final CallServiceParameters params, final long estimatedGas, final ContractCallContext context);
```
steven-sheehy, line 26: Revert the change to this class and its subclass to pass context.

hedera-mirror-web3/src/main/java/com/hedera/node/app/service/evm/contracts/execution/HederaEvmTxProcessor.java
```diff
@@ -106,7 +108,9 @@ public HederaEvmTransactionProcessingResult execute(
             final Bytes payload,
             final boolean isStatic,
             final Address mirrorReceiver,
-            final boolean contractCreation) {
+            final boolean contractCreation,
+            final TracerType tracerType,
+            final ContractCallContext contractCallContext) {
```
steven-sheehy, line 113: Revert this change. Just call ContractCallContext.get() to retrieve inside execute(). We should try to keep copied service code as similar as possible to upstream.

steven-sheehy (changes requested):
Some merge conflicts as well.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/OpcodeTracer.java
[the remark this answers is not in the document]
konstantinabl: As we discussed we will change this logic in another PR after we merge this PR to main
An issue has been created https://github.com/hashgraph/hedera-mirror-node/issues/8639

steven-sheehy (approved):
LGTM

PR #8245, opened 2024-05-07
feat: Add new opcodes endpoint

### LINKED ISSUE #8174: Add new `/api/v1/contracts/results/{transactionIdOrHash}/opcodes` endpoint
As part of the work package to support tracing transactions and externalizing the executed opcodes, we'd need to add new endpoint on path `/api/v1/contracts/results/{transactionIdOrHash}/opcodes`, however important detail should be that this endpoint will be disabled by default. Can be used and enabled only on local or custom networks and won't be available on the official public ones.
Spring supports this, one example how we can achieve this behaviour is https://docs.spring.io/spring-boot/reference/features/developing-auto-configuration.html#features.developing-auto-configuration.condition-annotations.property-conditions

Current contract call route can be found [here](https://github.com/hashgraph/hedera-mirror-node/blob/356fe63e94cf9223527f1469bce70f7fceb37690/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/ContractController.java#L74). New route can be influenced by it, we should also add rate limiting and hash validation.

More info can be found [here](https://hips.hedera.com/hip/hip-801).

victor-yanev:
**Description**:

**Related issue(s)**:

Fixes #8174 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/ContractController.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/ContractController.java
@@ -16,49 +16,25 @@
 
 package com.hedera.mirror.web3.controller;
 
-import static com.hedera.mirror.web3.controller.ValidationErrorParser.extractValidationError;
-import static com.hedera.mirror.web3.service.model.CallServiceParameters.CallType.ETH_CALL;
-import static com.hedera.mirror.web3.service.model.CallServiceParameters.CallType.ETH_ESTIMATE_GAS;
-import static org.apache.tuweni.bytes.Bytes.EMPTY;
-import static org.springframework.http.HttpStatus.BAD_REQUEST;
-import static org.springframework.http.HttpStatus.INTERNAL_SERVER_ERROR;
-import static org.springframework.http.HttpStatus.NOT_FOUND;
-import static org.springframework.http.HttpStatus.NOT_IMPLEMENTED;
-import static org.springframework.http.HttpStatus.SERVICE_UNAVAILABLE;
-import static org.springframework.http.HttpStatus.TOO_MANY_REQUESTS;
-import static org.springframework.http.HttpStatus.UNSUPPORTED_MEDIA_TYPE;
-
 import com.hedera.mirror.web3.evm.properties.MirrorNodeEvmProperties;
-import com.hedera.mirror.web3.exception.EntityNotFoundException;
-import com.hedera.mirror.web3.exception.InvalidInputException;
 import com.hedera.mirror.web3.exception.InvalidParametersException;
-import com.hedera.mirror.web3.exception.MirrorEvmTransactionException;
 import com.hedera.mirror.web3.exception.RateLimitException;
 import com.hedera.mirror.web3.service.ContractCallService;
 import com.hedera.mirror.web3.service.model.CallServiceParameters;
 import com.hedera.mirror.web3.viewmodel.ContractCallRequest;
 import com.hedera.mirror.web3.viewmodel.ContractCallResponse;
-import com.hedera.mirror.web3.viewmodel.GenericErrorResponse;
 import com.hedera.node.app.service.evm.store.models.HederaEvmAccount;
 import io.github.bucket4j.Bucket;
 import jakarta.validation.Valid;
 import lombok.CustomLog;
 import lombok.RequiredArgsConstructor;
-import org.apache.commons.lang3.StringUtils;
 import org.apache.tuweni.bytes.Bytes;
 import org.hyperledger.besu.datatypes.Address;
-import org.springframework.dao.QueryTimeoutException;
-import org.springframework.http.converter.HttpMessageConversionException;
-import org.springframework.validation.BindException;
-import org.springframework.web.HttpMediaTypeNotSupportedException;
-import org.springframework.web.bind.annotation.CrossOrigin;
-import org.springframework.web.bind.annotation.ExceptionHandler;
-import org.springframework.web.bind.annotation.PostMapping;
-import org.springframework.web.bind.annotation.RequestBody;
-import org.springframework.web.bind.annotation.RequestMapping;
-import org.springframework.web.bind.annotation.ResponseStatus;
-import org.springframework.web.bind.annotation.RestController;
-import org.springframework.web.server.ServerWebInputException;
+import org.springframework.web.bind.annotation.*;
+
+import static com.hedera.mirror.web3.service.model.CallServiceParameters.CallType.ETH_CALL;
+import static com.hedera.mirror.web3.service.model.CallServiceParameters.CallType.ETH_ESTIMATE_GAS;
+import static org.apache.tuweni.bytes.Bytes.EMPTY;
 
 @CustomLog
 @RequestMapping("/api/v1/contracts")
@@ -145,94 +121,5 @@ class ContractController {
             throw new InvalidParametersException(
                     "gas field must be less than or equal to %d".formatted(evmProperties.getMaxGasLimit()));
         }
-    }
-
-    /**
-     * Temporary handler, intended for dealing with forthcoming features that are not yet available, such as the absence
-     * of a precompile for gas estimation.
-     **/
-    @ExceptionHandler
-    @ResponseStatus(NOT_IMPLEMENTED)
-    private GenericErrorResponse unsupportedOpResponse(final UnsupportedOperationException e) {
-        return errorResponse(e.getMessage());
-    }
-
-    @ExceptionHandler
-    @ResponseStatus(TOO_MANY_REQUESTS)
-    private GenericErrorResponse rateLimitError(final RateLimitException e) {
-        return errorResponse(e.getMessage());
-    }
-
-    @ExceptionHandler
-    @ResponseStatus(BAD_REQUEST)
-    private GenericErrorResponse validationError(final BindException e) {
-        final var errors = extractValidationError(e);
-        log.warn("Validation error: {}", errors);
-        return new GenericErrorResponse(errors);
-    }
-
-    @ExceptionHandler
-    @ResponseStatus(BAD_REQUEST)
-    private GenericErrorResponse inputValidationError(final InvalidInputException e) {
-        log.warn("Input validation error: {}", e.getMessage());
-        return new GenericErrorResponse(e.getMessage());
-    }
-
-    @ExceptionHandler
-    @ResponseStatus(BAD_REQUEST)
-    private GenericErrorResponse mirrorEvmTransactionException(final MirrorEvmTransactionException e) {
-        log.warn("Mirror EVM transaction error: {}", e.getMessage());
-        return errorResponse(e.getMessage(), e.getDetail(), e.getData());
-    }
-
-    @ExceptionHandler
-    @ResponseStatus(BAD_REQUEST)
-    private GenericErrorResponse invalidJson(final ServerWebInputException e) {
-        log.warn("Transaction body parsing error: {}", e.getMessage());
-        return errorResponse(e.getReason(), "Unable to parse JSON", StringUtils.EMPTY);
-    }
-
-    @ExceptionHandler
-    @ResponseStatus(BAD_REQUEST)
-    private GenericErrorResponse invalidJson(final HttpMessageConversionException e) {
-        log.warn("Transaction body parsing error: {}", e.getMessage());
-        return errorResponse("Unable to parse JSON", e.getMessage(), StringUtils.EMPTY);
-    }
-
-    @ExceptionHandler
-    @ResponseStatus(NOT_FOUND)
-    private GenericErrorResponse notFound(final EntityNotFoundException e) {
-        log.warn("Not found: {}", e.getMessage());
-        return errorResponse(e.getMessage());
-    }
-
-    @ExceptionHandler
-    @ResponseStatus(UNSUPPORTED_MEDIA_TYPE)
-    private GenericErrorResponse unsupportedMediaTypeError(final HttpMediaTypeNotSupportedException e) {
-        log.warn("Unsupported media type error: {}", e.getMessage());
-        return errorResponse(UNSUPPORTED_MEDIA_TYPE.getReasonPhrase(), e.getMessage(), StringUtils.EMPTY);
-    }
-
-    @ExceptionHandler
-    @ResponseStatus(INTERNAL_SERVER_ERROR)
-    private GenericErrorResponse genericError(final Exception e) {
-        log.error("Generic error: ", e);
-        return errorResponse(INTERNAL_SERVER_ERROR.getReasonPhrase());
-    }
-
-    @ExceptionHandler
-    @ResponseStatus(SERVICE_UNAVAILABLE)
-    private GenericErrorResponse queryTimeout(final QueryTimeoutException e) {
-        log.error("Query timed out: {}", e.getMessage());
-        return errorResponse(SERVICE_UNAVAILABLE.getReasonPhrase());
-    }
-
-    private GenericErrorResponse errorResponse(final String errorMessage) {
-        return new GenericErrorResponse(errorMessage);
-    }
-
-    private GenericErrorResponse errorResponse(
-            final String errorMessage, final String detailedErrorMessage, final String hexErrorMessage) {
-        return new GenericErrorResponse(errorMessage, detailedErrorMessage, hexErrorMessage);
     }
 }
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/OpcodesController.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/OpcodesController.java
@@ -1,0 +1,152 @@
+/*
+ * Copyright (C) 2024 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hedera.mirror.web3.controller;
+
+import com.google.protobuf.ByteString;
+import com.hedera.mirror.common.domain.transaction.EthereumTransaction;
+import com.hedera.mirror.web3.evm.utils.TransactionUtils;
+import com.hedera.mirror.web3.exception.RateLimitException;
+import com.hedera.mirror.web3.repository.EthereumTransactionRepository;
+import com.hedera.mirror.web3.service.ContractCallService;
+import com.hedera.mirror.web3.service.model.CallServiceParameters;
+import com.hedera.mirror.web3.viewmodel.BlockType;
+import com.hedera.mirror.web3.viewmodel.OpcodesResponse;
+import com.hedera.node.app.service.evm.store.models.HederaEvmAccount;
+import com.hedera.services.utils.EntityIdUtils;
+import com.hederahashgraph.api.proto.java.ContractID;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import io.github.bucket4j.Bucket;
+import lombok.CustomLog;
+import lombok.RequiredArgsConstructor;
+import org.apache.tuweni.bytes.Bytes;
+import org.hyperledger.besu.datatypes.Address;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.web.bind.annotation.*;
+
+import java.math.BigInteger;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+import static com.hedera.mirror.common.util.DomainUtils.convertToNanosMax;
+import static com.hedera.mirror.web3.evm.utils.TransactionUtils.isValidEthHash;
+import static com.hedera.mirror.web3.evm.utils.TransactionUtils.isValidTransactionId;
+import static com.hedera.mirror.web3.service.model.CallServiceParameters.CallType.ETH_CALL;
+
+@CustomLog
+@RestController
+@RequiredArgsConstructor
+@RequestMapping("/api/v1/contracts/results")
+@ConditionalOnProperty(prefix = "hedera.mirror.opcode.tracer", name = "enabled", havingValue = "true")
+class OpcodesController {
+
+    private final ContractCallService contractCallService;
+    private final Bucket bucket;
+    private final EthereumTransactionRepository transactionRepository;
+
+    @CrossOrigin(origins = "*")
+    @PostMapping(value = "/{transactionIdOrHash}/opcodes")
+    OpcodesResponse opcodes(@PathVariable String transactionIdOrHash,
+                            @RequestParam(required = false, defaultValue = "true") boolean stack,
+                            @RequestParam(required = false, defaultValue = "false") boolean memory,
+                            @RequestParam(required = false, defaultValue = "false") boolean storage) {
+        if (!bucket.tryConsume(1)) {
+            throw new RateLimitException("Rate limit exceeded.");
+        }
+
+        final var params = constructServiceParameters(transactionIdOrHash);
+        final var result = contractCallService.processOpcodeCall(params);
+
+        return OpcodesResponse.builder()
+                // TODO: Not sure if this is the correct way to get the contractId here
+                .contractId(result.transactionProcessingResult()
+                        .getRecipient()
+                        .map(EntityIdUtils::contractIdFromEvmAddress)
+                        .map(ContractID::toString)
+                        .orElse(null))
+                // TODO: Not sure if this is the correct way to get the address here
+                .address(result.transactionProcessingResult()
+                        .getRecipient()
+                        .map(Address::toHexString)
+                        .orElse(Address.ZERO.toHexString()))
+                .gas(result.transactionProcessingResult().getGasPrice())
+                .failed(!result.transactionProcessingResult().isSuccessful())
+                .returnValue(result.transactionProcessingResult().getOutput().toHexString())
+                .opcodes(result.opcodes().stream()
+                        .map(opcode -> OpcodesResponse.Opcode.builder()
+                                .pc(opcode.pc())
+                                .op(opcode.op())
+                                .gas(opcode.gas())
+                                .gasCost(opcode.gasCost())
+                                .depth(opcode.depth())
+                                .stack(stack ?
+                                        opcode.stack().stream()
+                                                .map(Bytes::toHexString)
+                                                .toList() :
+                                        List.of())
+                                .memory(memory ?
+                                        opcode.memory().stream()
+                                                .map(Bytes::toHexString)
+                                                .toList() :
+                                        List.of())
+                                .storage(storage ?
+                                        opcode.storage().entrySet().stream()
+                                                .collect(Collectors.toMap(
+                                                        Map.Entry::getKey,
+                                                        entry -> entry.getValue().toHexString())) :
+                                        Map.of())
+                                .reason(opcode.reason())
+                                .build())
+                        .toList())
+                .build();
+    }
+
+    private CallServiceParameters constructServiceParameters(@NonNull String transactionIdOrHash) {
+        EthereumTransaction ethTransaction;
+        if (isValidEthHash(transactionIdOrHash)) {
+            // TODO: Need to get the transaction by hash here (not sure if this is the correct way to do it)
+            final var transactionHash = ByteString.fromHex(transactionIdOrHash).toByteArray();
+            ethTransaction = transactionRepository
+                    .findByHash(transactionHash)
+                    .orElseThrow(() -> new IllegalArgumentException("Transaction not found"));
+        } else if (isValidTransactionId(transactionIdOrHash)) {
+            // TODO: Need to get the transaction by ID here (not sure if this is the correct way to do it)
+            final var transactionId = TransactionUtils.parseTransactionId(transactionIdOrHash);
+            final var transactionValidStart = Objects.requireNonNull(transactionId.transactionValidStart());
+            ethTransaction = transactionRepository
+                    .findById(convertToNanosMax(transactionValidStart.seconds(), transactionValidStart.nanos()))
+                    .orElseThrow(() -> new IllegalArgumentException("Transaction not found"));
+        } else {
+            throw new IllegalArgumentException("Invalid transaction ID or hash");
+        }
+
+        return CallServiceParameters.builder()
+                .isStatic(false)
+                .callType(ETH_CALL)
+                .isEstimate(false)
+                // TODO: Need to get block number somehow from the fetched transaction above
+                .block(BlockType.LATEST)
+                // TODO: Need to get sender address somehow from the fetched transaction above
+                .sender(new HederaEvmAccount(Address.ZERO))
+                .receiver(Address.fromHexString(Bytes.of(ethTransaction.getToAddress()).toHexString()))
+                .gas(ethTransaction.getGasLimit())
+                .value(new BigInteger(ethTransaction.getValue()).longValue())
+                .callData(Bytes.of(ethTransaction.getCallData()))
+                .build();
+    }
+}
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/EthereumTransactionRepository.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/EthereumTransactionRepository.java
@@ -1,0 +1,27 @@
+/*
+ * Copyright (C) 2024 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hedera.mirror.web3.repository;
+
+import com.hedera.mirror.common.domain.transaction.EthereumTransaction;
+import org.springframework.data.repository.CrudRepository;
+
+import java.util.Optional;
+
+public interface EthereumTransactionRepository extends CrudRepository<EthereumTransaction, Long> {
+
+    Optional<EthereumTransaction> findByHash(byte[] transactionHash);
+}
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/config/ControllerExceptionHandler.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/config/ControllerExceptionHandler.java
@@ -1,0 +1,130 @@
+/*
+ * Copyright (C) 2024 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hedera.mirror.web3.config;
+
+import com.hedera.mirror.web3.exception.EntityNotFoundException;
+import com.hedera.mirror.web3.exception.InvalidInputException;
+import com.hedera.mirror.web3.exception.MirrorEvmTransactionException;
+import com.hedera.mirror.web3.exception.RateLimitException;
+import com.hedera.mirror.web3.viewmodel.GenericErrorResponse;
+import lombok.CustomLog;
+import org.apache.commons.lang3.StringUtils;
+import org.springframework.dao.QueryTimeoutException;
+import org.springframework.http.converter.HttpMessageConversionException;
+import org.springframework.validation.BindException;
+import org.springframework.web.HttpMediaTypeNotSupportedException;
+import org.springframework.web.bind.annotation.ControllerAdvice;
+import org.springframework.web.bind.annotation.ExceptionHandler;
+import org.springframework.web.bind.annotation.ResponseStatus;
+import org.springframework.web.server.ServerWebInputException;
+
+import static com.hedera.mirror.web3.controller.ValidationErrorParser.extractValidationError;
+import static org.springframework.http.HttpStatus.*;
+
+@CustomLog
+@ControllerAdvice
+public class ControllerExceptionHandler {
+
+    /**
+     * Temporary handler, intended for dealing with forthcoming features that are not yet available, such as the absence
+     * of a precompile for gas estimation.
+     **/
+    @ExceptionHandler
+    @ResponseStatus(NOT_IMPLEMENTED)
+    private GenericErrorResponse unsupportedOpResponse(final UnsupportedOperationException e) {
+        return errorResponse(e.getMessage());
+    }
+
+    @ExceptionHandler
+    @ResponseStatus(TOO_MANY_REQUESTS)
+    private GenericErrorResponse rateLimitError(final RateLimitException e) {
+        return errorResponse(e.getMessage());
+    }
+
+    @ExceptionHandler
+    @ResponseStatus(BAD_REQUEST)
+    private GenericErrorResponse validationError(final BindException e) {
+        final var errors = extractValidationError(e);
+        log.warn("Validation error: {}", errors);
+        return new GenericErrorResponse(errors);
+    }
+
+    @ExceptionHandler
+    @ResponseStatus(BAD_REQUEST)
+    private GenericErrorResponse inputValidationError(final InvalidInputException e) {
+        log.warn("Input validation error: {}", e.getMessage());
+        return new GenericErrorResponse(e.getMessage());
+    }
+
+    @ExceptionHandler
+    @ResponseStatus(BAD_REQUEST)
+    private GenericErrorResponse mirrorEvmTransactionException(final MirrorEvmTransactionException e) {
+        log.warn("Mirror EVM transaction error: {}", e.getMessage());
+        return errorResponse(e.getMessage(), e.getDetail(), e.getData());
+    }
+
+    @ExceptionHandler
+    @ResponseStatus(BAD_REQUEST)
+    private GenericErrorResponse invalidJson(final ServerWebInputException e) {
+        log.warn("Transaction body parsing error: {}", e.getMessage());
+        return errorResponse(e.getReason(), "Unable to parse JSON", StringUtils.EMPTY);
+    }
+
+    @ExceptionHandler
+    @ResponseStatus(BAD_REQUEST)
+    private GenericErrorResponse invalidJson(final HttpMessageConversionException e) {
+        log.warn("Transaction body parsing error: {}", e.getMessage());
+        return errorResponse("Unable to parse JSON", e.getMessage(), StringUtils.EMPTY);
+    }
+
+    @ExceptionHandler
+    @ResponseStatus(NOT_FOUND)
+    private GenericErrorResponse notFound(final EntityNotFoundException e) {
+        log.warn("Not found: {}", e.getMessage());
+        return errorResponse(e.getMessage());
+    }
+
+    @ExceptionHandler
+    @ResponseStatus(UNSUPPORTED_MEDIA_TYPE)
+    private GenericErrorResponse unsupportedMediaTypeError(final HttpMediaTypeNotSupportedException e) {
+        log.warn("Unsupported media type error: {}", e.getMessage());
+        return errorResponse(UNSUPPORTED_MEDIA_TYPE.getReasonPhrase(), e.getMessage(), StringUtils.EMPTY);
+    }
+
+    @ExceptionHandler
+    @ResponseStatus(INTERNAL_SERVER_ERROR)
+    private GenericErrorResponse genericError(final Exception e) {
+        log.error("Generic error: ", e);
+        return errorResponse(INTERNAL_SERVER_ERROR.getReasonPhrase());
+    }
+
+    @ExceptionHandler
+    @ResponseStatus(SERVICE_UNAVAILABLE)
+    private GenericErrorResponse queryTimeout(final QueryTimeoutException e) {
+        log.error("Query timed out: {}", e.getMessage());
+        return errorResponse(SERVICE_UNAVAILABLE.getReasonPhrase());
+    }
+
+    private GenericErrorResponse errorResponse(final String errorMessage) {
+        return new GenericErrorResponse(errorMessage);
+    }
+
+    private GenericErrorResponse errorResponse(
+            final String errorMessage, final String detailedErrorMessage, final String hexErrorMessage) {
+        return new GenericErrorResponse(errorMessage, detailedErrorMessage, hexErrorMessage);
+    }
+}
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/utils/TransactionUtils.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/utils/TransactionUtils.java
@@ -1,0 +1,81 @@
+/*
+ * Copyright (C) 2024 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hedera.mirror.web3.evm.utils;
+
+import com.hedera.hapi.node.base.AccountID;
+import com.hedera.hapi.node.base.Timestamp;
+import com.hedera.hapi.node.base.TransactionID;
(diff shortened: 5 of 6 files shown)


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/OpcodesController.java
```diff
@@ -0,0 +1,152 @@
...
+                            @RequestParam(required = false, defaultValue = "false") boolean memory,
+                            @RequestParam(required = false, defaultValue = "false") boolean storage) {
+        if (!bucket.tryConsume(1)) {
+            throw new RateLimitException("Rate limit exceeded.");
+        }
+
+        final var params = constructServiceParameters(transactionIdOrHash);
+        final var result = contractCallService.processOpcodeCall(params);
+
+        return OpcodesResponse.builder()
```
steven-sheehy, line 75: You need to use an OpenAPI first development model whereby the openapi schema is updated first, then the model is generated from that. See the rest-java module for how that should be done. We were too lax with contracts/call so it doesn't follow that standard but we should do that going forward.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/OpcodesController.java
```diff
@@ -0,0 +1,152 @@
...
+@ConditionalOnProperty(prefix = "hedera.mirror.opcode.tracer", name = "enabled", havingValue = "true")
+class OpcodesController {
+
+    private final ContractCallService contractCallService;
+    private final Bucket bucket;
+    private final EthereumTransactionRepository transactionRepository;
+
+    @CrossOrigin(origins = "*")
+    @PostMapping(value = "/{transactionIdOrHash}/opcodes")
+    OpcodesResponse opcodes(@PathVariable String transactionIdOrHash,
```
steven-sheehy, line 64: We should use a strongly typed object that encapsulates the validation and parsing of transaction id or hash similar to `EntityIdRangeParameter` in `AllowancesController`.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/OpcodesController.java
```diff
@@ -0,0 +1,152 @@
...
+            ethTransaction = transactionRepository
+                    .findByHash(transactionHash)
+                    .orElseThrow(() -> new IllegalArgumentException("Transaction not found"));
+        } else if (isValidTransactionId(transactionIdOrHash)) {
+            // TODO: Need to get the transaction by ID here (not sure if this is the correct way to do it)
+            final var transactionId = TransactionUtils.parseTransactionId(transactionIdOrHash);
+            final var transactionValidStart = Objects.requireNonNull(transactionId.transactionValidStart());
+            ethTransaction = transactionRepository
+                    .findById(convertToNanosMax(transactionValidStart.seconds(), transactionValidStart.nanos()))
+                    .orElseThrow(() -> new IllegalArgumentException("Transaction not found"));
```
steven-sheehy, line 133: Repository calls belong in the service layer. Perhaps might need an alternative method on ContractCallService

victor-yanev: Done, I extracted some of the common logic in two new services: `RecordFileService` and `TransactionService`. 

Please, have a look and tell me if you have any concerns 😁.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/OpcodesController.java
victor-yanev: Done.

hedera-mirror-web3/src/main/resources/application.yml
```diff
@@ -5,7 +5,7 @@ hedera:
         host: 127.0.0.1
         name: mirror_node
         password: mirror_web3_pass
-        port: 5432
+        port: 5433
```
konstantinabl, line 8: shouldnt commit this :?

victor-yanev: Yes, I changed it locally but forgot to exclude it from the change list. 

Thanks for catching this!

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/OpcodesController.java
```diff
@@ -0,0 +1,136 @@
...
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.RestController;
+
+@CustomLog
+@RestController
+@RequiredArgsConstructor
+@RequestMapping("/api/v1/contracts/results")
+@ConditionalOnProperty(prefix = "hedera.mirror.opcode.tracer", name = "enabled", havingValue = "true")
```
IvanKavaldzhiev, line 48: I believe we should keep the default value to be "false", since the API is meant for local use only (at least for now).

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/EthereumTransactionRepository.java
```diff
@@ -0,0 +1,26 @@
...
+
+package com.hedera.mirror.web3.repository;
+
+import com.hedera.mirror.common.domain.transaction.EthereumTransaction;
+import java.util.Optional;
+import org.springframework.data.repository.CrudRepository;
+
+public interface EthereumTransactionRepository extends CrudRepository<EthereumTransaction, Long> {
+
+    Optional<EthereumTransaction> findByHash(byte[] transactionHash);
```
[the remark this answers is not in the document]
jnels124: we would need to lookup the ethereum_transaction by consensus_timestamp and payer_account_id. In v2 this table is distributed on the payer_account_id column. In v2 schema, queries that don't contain the distribution column will be sent to all shards to gather the result. If the distribution column is provided, it only has to do a single query to find the data.

hedera-mirror-rest/api/v1/openapi.yml
```diff
@@ -611,6 +611,40 @@ paths:
...
       tags:
         - contracts
+  /api/v1/contracts/results/{transactionIdOrHash}/opcodes:
+    get:
+      summary: Get the opcode traces for a historical transaction on the network with the given transaction ID or hash
+      description: |
+        Re-executes a transaction and returns a result containing detailed information for the execution, 
+        including all values from the {@code stack}, {@code memory} and {@code storage} 
+        and the entire trace of opcodes that were executed during the replay.
+      operationId: getContractOpcodesByTransactionIdOrHash
```
steven-sheehy, line 621:
```suggestion
      operationId: getContractOpcodes
```

hedera-mirror-rest/api/v1/openapi.yml
```diff
@@ -611,6 +611,40 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/OpcodesResponse"
+        400:
+          description: Validation error
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
```
steven-sheehy, line 639: We should have 404 for when transaction id or hash not found.

hedera-mirror-rest/api/v1/openapi.yml
```diff
@@ -1500,6 +1534,55 @@ components:
           $ref: "#/components/schemas/NftAllowances"
         links:
           $ref: "#/components/schemas/Links"
+    OpcodesResponse:
+      type: object
+      properties:
+        address:
+          type: string
+        contract_id:
+          type: string
```
steven-sheehy, line 1543:
```suggestion
          $ref: "#/components/schemas/EntityId"
```

hedera-mirror-rest/api/v1/openapi.yml
```diff
@@ -1500,6 +1534,55 @@ components:
           $ref: "#/components/schemas/NftAllowances"
         links:
           $ref: "#/components/schemas/Links"
+    OpcodesResponse:
+      type: object
```
steven-sheehy, line 1538: Add a required section to OpcodesResponse and Opcode and list the fields

hedera-mirror-rest/api/v1/openapi.yml
```diff
@@ -1500,6 +1534,55 @@ components:
...
+      properties:
+        address:
+          type: string
```
steven-sheehy, line 1541: Add an informative description to all primitive fields. Note in the description whether it is hex encoded or not.

hedera-mirror-rest/api/v1/openapi.yml
```diff
@@ -1500,6 +1534,55 @@ components:
           $ref: "#/components/schemas/NftAllowances"
         links:
           $ref: "#/components/schemas/Links"
+    OpcodesResponse:
+      type: object
+      properties:
+        address:
```
steven-sheehy, line 1540: Please add the same to all hex encoded fields.

```suggestion
        address:
          format: binary
```

hedera-mirror-rest/api/v1/openapi.yml
```diff
@@ -1500,6 +1534,55 @@ components:
...
+          format: int32
+        op:
+          type: string
+        gas:
+          type: integer
+          format: int64
+        gas_cost:
+          type: integer
+          format: int64
+        depth:
```
steven-sheehy, line 1569: Sort all fields by name in opcodes and opcodes response

steven-sheehy (changes requested):
Still reviewing

hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/transaction/Opcode.java
```diff
@@ -0,0 +1,32 @@
...
+ * limitations under the License.
+ */
+
+package com.hedera.mirror.common.domain.transaction;
+
+import java.util.List;
+import java.util.Map;
+import org.apache.tuweni.bytes.Bytes;
+
+public record Opcode(int pc,
```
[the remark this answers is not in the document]
victor-yanev: Moved it to `com.hedera.mirror.web3.evm.contracts.execution.traceability`

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/CallServiceParametersBuilderImpl.java
```diff
@@ -0,0 +1,199 @@
...
+
+        return buildFromRecordItem(recordItem, recordFile.getIndex());
+    }
+
+    private CallServiceParameters buildFromRecordItem(RecordItem recordItem, Long blockNumber) {
+        return CallServiceParameters.builder()
+                .sender(new HederaEvmAccount(getSenderAddress(recordItem)))
+                .receiver(getReceiverAddress(recordItem))
+                .gas(getGasLimit(recordItem))
+                .value(getValue(recordItem).longValue())
```
steven-sheehy, line 109: These and most other values can be retrieved from `EthereumTransaction`

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/ContractController.java
```diff
@@ -78,57 +52,18 @@ ContractCallResponse call(@RequestBody @Valid ContractCallRequest request) {
...
         } catch (InvalidParametersException e) {
             // The validation failed but no processing was made - restore the consumed gas back to the bucket.
             gasLimitBucket.addTokens(request.getGas());
             throw e;
         }
-
-        return new ContractCallResponse(result);
-    }
-
-    private CallServiceParameters constructServiceParameters(ContractCallRequest request) {
```
steven-sheehy, line 97: Undo this change. The controller should be responsible for mapping from the user input to the service parameters. It shouldn't live anywhere else as it's not reusable by any other class.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/CallServiceParametersBuilderImpl.java
```diff
@@ -0,0 +1,199 @@
...
+import lombok.RequiredArgsConstructor;
+import lombok.SneakyThrows;
+import org.apache.tuweni.bytes.Bytes;
+import org.hyperledger.besu.datatypes.Address;
+import org.springframework.stereotype.Service;
+import org.springframework.util.Assert;
+
+@Service
+@RequiredArgsConstructor
+public class CallServiceParametersBuilderImpl implements CallServiceParametersBuilder {
```
steven-sheehy, line 48: I am not a fan of this class. We should undo the change to ContractController to create its parameters. We should then either push this logic into OpcodesController or change this to OpcodeService and have it invoke contractcallservice as well.

hedera-mirror-common/src/test/java/com/hedera/mirror/common/domain/TransactionMocks.java
```diff
@@ -0,0 +1,358 @@
...
+import com.hederahashgraph.api.proto.java.TransactionRecord;
+import com.hederahashgraph.api.proto.java.TransferList;
+import java.util.List;
+import lombok.SneakyThrows;
+import lombok.experimental.UtilityClass;
+import org.apache.commons.lang3.RandomStringUtils;
+import org.jetbrains.annotations.NotNull;
+
+@UtilityClass
+public class TransactionMocks {
```
steven-sheehy, line 52: Please remove this class. We should not be crafting HAPI transactions to test anything in web3 module. Transaction and record bytes are not stored in our database anyway.

steven-sheehy (changes requested):
This PR is way too big and has a lot of issues, hence why it has taken a long time to review. It attempts to do too much refactoring at the same time when instead any refactoring should've been done upfront before any opcode work. It would've been more appropriate to break this work package down into smaller chunks like openapi, controller plus input validation, service layer, repository, etc.

Note: Please make sure you expand the feedback to address all 40+ items

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/controller/OpcodesControllerTest.java
```diff
@@ -0,0 +1,643 @@
...
+import org.springframework.test.web.servlet.MockMvc;
+import org.springframework.test.web.servlet.ResultMatcher;
+import org.springframework.test.web.servlet.request.MockHttpServletRequestBuilder;
+import org.springframework.transaction.support.TransactionOperations;
+import org.springframework.util.StringUtils;
+
+@ExtendWith(SpringExtension.class)
+@WebMvcTest(controllers = OpcodesController.class)
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+class OpcodesControllerTest {
```
steven-sheehy, line 111: Mock MVC tests are fine, but we also need some integration tests that test it E2E

hedera-mirror-rest/api/v1/openapi.yml
```diff
@@ -611,6 +611,49 @@ paths:
...
+              schema:
+                $ref: "#/components/schemas/OpcodesResponse"
+        400:
+          description: Validation error
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+        404:
+          description: Transaction or record file not found
```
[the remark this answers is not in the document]
victor-yanev: It's also possible that we return 404 if the record file cannot be found in the repository call, that's why I combined the message here.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/OpcodesController.java
```diff
@@ -0,0 +1,145 @@
...
+            @RequestParam(required = false, defaultValue = "true") boolean stack,
+            @RequestParam(required = false, defaultValue = "false") boolean memory,
+            @RequestParam(required = false, defaultValue = "false") boolean storage
+    ) {
+        if (!rateLimitBucket.tryConsume(1)) {
+            throw new RateLimitException("Rate limit exceeded.");
+        }
+
+        final var params = callServiceParametersBuilder.buildFromTransaction(transactionIdOrHash);
+        if (!gasLimitBucket.tryConsume(params.getGas())) {
```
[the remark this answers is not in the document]
victor-yanev: Do we really need gas throttle here, since this is a historical transaction that has already passed any checks?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/Opcode.java
```diff
@@ -0,0 +1,32 @@
...
+import org.apache.tuweni.bytes.Bytes;
+
+public record Opcode(int pc,
+                     String op,
+                     long gas,
+                     long gasCost,
+                     int depth,
+                     Optional<Bytes[]> stack,
+                     Optional<Bytes[]> memory,
+                     Optional<Map<Bytes, Bytes>> storage,
```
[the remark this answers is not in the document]
victor-yanev: We tried to make the response equivalent to the one in besu and geth, where `null` is returned in cases where we pass `disableStorage`, `disableMemory` or `disableStack` in the opcode tracer options. 

References:
- https://geth.ethereum.org/docs/developers/evm-tracing/built-in-tracers
- https://github.com/hyperledger/besu/blob/main/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/vm/DebugOperationTracer.java

hedera-mirror-common/src/test/java/com/hedera/mirror/common/domain/TransactionMocks.java
victor-yanev: Most of the logic that is done to build the `CallServiceParameters` is obtaining them via fetching those Transactions from the database, how are we going to test this without building any transactions in the set up of the tests?

If we don't want to be crafting transactions, what we can do is change the OpcodeControllerTest to a controller integration test, deploy a contract, do some contract call and test against the real transactions that are generated from these operations. 

Though, this would take a significant time to implement since we don't have any integration tests for controllers in the web3 module and we would need to set up a new base class for this type of controller tests. 

Needless to say, this would make the PR even bigger than it is already.

We can do this in another PR which will include integration tests for this controller and remove the opcode controller unit tests from this PR. What do you think?

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/controller/OpcodesControllerTest.java
victor-yanev: I totally agree, the QA team is currently implementing E2E cucumber tests which are calling this endpoint.

Also, if we decide to refactor these to integration tests (https://github.com/hashgraph/hedera-mirror-node/pull/8245#discussion_r1618582217 ), I suggest we do this in a separate PR to avoid piling up this PR with additional changes?

hedera-mirror-common/src/test/java/com/hedera/mirror/common/domain/TransactionMocks.java
victor-yanev: Another option is to move this logic to a new `OpcodeService` and add integration tests for the service and keep those controller tests as unit tests which mock a response from the service?

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/controller/OpcodesControllerTest.java
steven-sheehy: Disagree about e2e tests. We can have one e2e test in cucumber but everything else should be an integration test. See Google's testing pyramid.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/CallServiceParametersBuilderImpl.java
```diff
@@ -0,0 +1,199 @@
...
+                    .flatMap(ethereumTransactionService::findByConsensusTimestamp);
+        }
+
+        return buildFromTransaction(transaction, ethTransaction);
+    }
+
+    private CallServiceParameters buildFromTransaction(Optional<Transaction> transactionOpt,
+                                                       Optional<EthereumTransaction> ethTransactionOpt) throws InvalidProtocolBufferException {
+        final Transaction transaction = transactionOpt
+                .orElseThrow(() -> new EntityNotFoundException("Transaction not found"));
```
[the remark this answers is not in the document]
victor-yanev: Isn't it possible that we have an existing `Transaction` in the database for the given `transactionIdOrHash` parameter, but not an existing `EthereumTransaction`?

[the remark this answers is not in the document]
steven-sheehy: Then those aren't Ethereum transactions. Are we supporting contract cal and create as well? If so you'll have to get data from contract_result.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/CallServiceParametersBuilderImpl.java
```diff
@@ -0,0 +1,199 @@
...
+
+        final RecordFile recordFile = recordFileService
+                .findRecordFileForTimestamp(transaction.getConsensusTimestamp())
```
steven-sheehy, line 83: Please double check that whatever block we lookup is cached in repository since it's looked up again in call flow.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/controller/OpcodesControllerTest.java
steven-sheehy: Actually we should have zero e2e cucumber tests since this API is disabled by default and it will never run in CI or during a deploy if we do write them.

steven-sheehy (commented):
Just noticed the large gap in test coverage. Every repository and service should have a corresponding integration test (not mock).

hedera-mirror-common/src/test/java/com/hedera/mirror/common/domain/TransactionMocks.java
steven-sheehy: Believe you still need domain object setup for the mock based opcode controller test. You can just use DomainBuilder directly in test and let it auto generate as much of the fields as possible. Can add an opcode integration test as a follow up pr.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/OpcodesController.java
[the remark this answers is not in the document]
steven-sheehy: Gas throttle is to ensure process doesn't consume too many resources from concurrent requests, not to validate transaction. So we need it.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/ContractTransactionHashRepository.java
```diff
@@ -0,0 +1,28 @@
...
+
+import com.hedera.mirror.common.domain.contract.ContractTransactionHash;
+import java.util.Optional;
+import org.springframework.data.repository.CrudRepository;
+import org.springframework.stereotype.Repository;
+
+@Repository
+public interface ContractTransactionHashRepository extends CrudRepository<ContractTransactionHash, byte[]> {
+
+    Optional<ContractTransactionHash> findByHash(byte[] hash);
```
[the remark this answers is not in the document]
victor-yanev: But we have `hash` marked as `@jakarta.persistence.Id` - is it possible to have two contract transaction hashes with the same id?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/OpcodesController.java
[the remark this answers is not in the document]
victor-yanev: Okay, in this case, how can we estimate the throttle if we haven't fetched the transaction yet?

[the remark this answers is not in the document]
steven-sheehy: Good point. We'll have to either assume some default and refund it back until we get actual or just wait until after we get contract result gas.

steven-sheehy:
Need to also add this endpoint to the ingress routes in the web3 helm chart.

hedera-mirror-rest/api/v1/openapi.yml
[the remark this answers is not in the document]
victor-yanev: What about scenarios where the contract_result is missing, or the contract_transaction_hash?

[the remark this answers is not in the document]
victor-yanev: What kind of exception do we want to throw in those cases and what status should be returned?

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/controller/OpcodesControllerTest.java
victor-yanev: Synced this with the team that’s working on it, we will not be adding e2e tests for this endpoint in the mirror node.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/ContractTransactionHashRepository.java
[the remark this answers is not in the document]
victor-yanev: Actually I removed this custom method from here and just used the default `findById` method.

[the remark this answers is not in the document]
victor-yanev: Still, we should be fine considering we are querying by ID?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/RecordFileServiceImpl.java
```diff
@@ -0,0 +1,57 @@
...
+public class RecordFileServiceImpl implements RecordFileService {
+
+    private final RecordFileRepository recordFileRepository;
+
+    @Override
+    public Optional<RecordFile> findRecordFileByBlock(BlockType block) {
+        if (block == BlockType.EARLIEST) {
+            return recordFileRepository.findEarliest();
+        } else if (block == BlockType.LATEST) {
+            return recordFileRepository.findLatest();
```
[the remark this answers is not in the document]
victor-yanev: Will be covered in the PR for integration tests.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/OpcodesController.java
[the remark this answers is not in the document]
victor-yanev: Since this is going to be used only locally, should we be overcomplicating things with the gas throttle? I don’t really see how there could be so much traffic when it’s only going to be used locally by devs. What do you think?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/Opcode.java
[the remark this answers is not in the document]
victor-yanev: It’s also `null` in the specification of the HIP:
IMG_6659

victor-yanev:
> Just noticed the large gap in test coverage. Every repository and service should have a corresponding integration test (not mock).

Will be done in the PR for the integration tests.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/CallServiceParametersBuilderImpl.java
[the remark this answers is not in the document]
victor-yanev: And yes, I believe we also want to support regular transactions, not only ETH ones.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/Opcode.java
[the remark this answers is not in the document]
IvanKavaldzhiev: You are referring to the section for `JSON-RPC Relay Trace API` which refers to example outputs from the json-rpc-relay. The Java REST API in mirror-node is not JSON RPC compatible and we confront to typical Java REST behaviour.

We can return empty collections for all fields and the relay can translate them to null values.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/CallServiceParametersBuilderImpl.java
victor-yanev: I made this new repository method also to be cacheable

victor-yanev: FYI here is the cache configuration:
```java
    @Bean(CACHE_MANAGER_RECORD_FILE_TIMESTAMP)
    CacheManager cacheManagerRecordFileTimestamp() {
        final var caffeine = Caffeine.newBuilder()
                .expireAfterWrite(10, TimeUnit.MINUTES)
                .maximumSize(10000)
                .recordStats();
        final CaffeineCacheManager caffeineCacheManager = new CaffeineCacheManager();
        caffeineCacheManager.setCacheNames(Set.of(CACHE_NAME));
        caffeineCacheManager.setCaffeine(caffeine);
        return caffeineCacheManager;
    }
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/ContractResultRepository.java
```diff
@@ -0,0 +1,26 @@
...
+
+package com.hedera.mirror.web3.repository;
+
+import com.hedera.mirror.common.domain.contract.ContractResult;
+import org.springframework.data.repository.CrudRepository;
+import org.springframework.stereotype.Repository;
+
+@Repository
+public interface ContractResultRepository extends CrudRepository<ContractResult, Long> {
+
```
konstantinabl, line 25: This is empty

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/ContractTransactionHashRepository.java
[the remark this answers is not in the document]
steven-sheehy: >  is it possible to have two contract transaction hashes with the same id?

Yes, it's possible. A user could submit the exact same EthereumTransaction twice. There's no primary key on this table guaranteeing uniqueness, just an index on hash.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/OpcodesController.java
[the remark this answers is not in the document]
steven-sheehy: Let's see what @bilyana-gospodinova or @IvanKavaldzhiev think.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/ControllerExceptionHandler.java
```diff
@@ -0,0 +1,177 @@
...
+public class ControllerExceptionHandler {
+
+    /**
+     * Temporary handler, intended for dealing with forthcoming features that are not yet available, such as the absence
+     * of a precompile for gas estimation.
+     **/
+    @ExceptionHandler
+    @ResponseStatus(NOT_IMPLEMENTED)
+    private ResponseEntity<?> unsupportedOpResponse(final UnsupportedOperationException e) {
+        return new ResponseEntity<>(errorResponse(e.getMessage()), NOT_IMPLEMENTED);
```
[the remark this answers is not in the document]
steven-sheehy: Why are conversations marked resolved but no change was done? BTW, on mirror node it's the standard that reviewer will resolve their own discussion items.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/CallServiceParametersBuilderImpl.java
steven-sheehy: Adding this cache doesn't address my comment or provide much value. This cache is only used to find by timestamp which is new and only used in opcode flow, not call flow. My comment was to ensure that we don't look up `RecordFile` object twice for opcode flow, which it still does with this change.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/OpcodesController.java
```diff
@@ -0,0 +1,86 @@
...
+     *
+     * @param transactionIdOrHash The transaction ID or hash
+     * @param stack Include stack information
+     * @param memory Include memory information
+     * @param storage Include storage information
+     * @return {@link OpcodesResponse} containing the result of the transaction execution
+     */
+    @GetMapping(value = "/{transactionIdOrHash}/opcodes")
+    OpcodesResponse getContractOpcodes(
+            @PathVariable @Valid TransactionIdOrHashParameter transactionIdOrHash,
```
steven-sheehy, line 67: nit: Don't think this `@Valid` actually does anything since validation is done manually by `valueOf()` and not via JSR303 annotations.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/OpcodesController.java
```diff
@@ -0,0 +1,86 @@
...
+
+        final var params = opcodeService.buildCallServiceParameters(transactionIdOrHash);
+        if (!gasLimitBucket.tryConsume(params.getGas())) {
+            throw new RateLimitException("Rate limit exceeded.");
+        }
+
+        final var options = new OpcodeTracerOptions(stack, memory, storage);
+        final var result = contractCallService.processOpcodeCall(params, options);
+
+        return opcodeService.buildOpcodesResponse(result);
```
steven-sheehy, line 84: I think it's awkward that the opcode service only builds request and response and doesn't have responsibility to invoke the call service. Think we should encapsulate opcode flow entirely within opcode service.

```suggestion
        final var options = new OpcodeTracerOptions(stack, memory, storage);
        return opcodeService.opcodeCall(transactionIdOrHash, options);
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/RecordFileRepository.java
```diff
@@ -52,4 +53,8 @@ public interface RecordFileRepository extends PagingAndSortingRepository<RecordF
             unless = "#result == null")
     @Query(value = "select * from record_file order by consensus_end desc limit 1", nativeQuery = true)
     Optional<RecordFile> findLatest();
+
+    @Cacheable(cacheNames = CACHE_NAME, cacheManager = CACHE_MANAGER_RECORD_FILE_TIMESTAMP, unless = "#result == null")
+    @Query("select r from RecordFile r where r.consensusEnd >= ?1 order by r.consensusEnd asc limit 1")
+    Optional<RecordFile> findByTimestamp(long timestamp);
```
steven-sheehy, line 59: Should have a test

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TransactionRepository.java
```diff
@@ -0,0 +1,29 @@
...
+import com.hedera.mirror.common.domain.entity.EntityId;
+import com.hedera.mirror.common.domain.transaction.Transaction;
+import java.util.Optional;
+import org.springframework.data.repository.CrudRepository;
+import org.springframework.stereotype.Repository;
+
+@Repository
+public interface TransactionRepository extends CrudRepository<Transaction, Long> {
+
+    Optional<Transaction> findByPayerAccountIdAndValidStartNs(EntityId payerAccountId, long validStartNs);
```
steven-sheehy, line 28: Should have a test.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/common/TransactionIdParameterTest.java
```diff
@@ -0,0 +1,86 @@
...
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+import org.springframework.test.context.junit.jupiter.SpringExtension;
+
+@ExtendWith(SpringExtension.class)
+class TransactionIdParameterTest {
+
+    private static final Pattern TRANSACTION_ID_PATTERN = Pattern.compile("^(\\d+)\\.(\\d+)\\.(\\d+)-(\\d{1,19})-(\\d{1,9})$");
```
steven-sheehy, line 42: Should expose and use the Pattern in `TransactionIdParameter`

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/common/TransactionIdParameterTest.java
```diff
@@ -0,0 +1,86 @@
...
+        assertEquals(5, matcher.groupCount());
+
+        final var parameter = assertDoesNotThrow(() -> TransactionIdParameter.valueOf(transactionId));
+        assertNotNull(parameter);
+        assertInstanceOf(TransactionIdParameter.class, parameter);
+        assertEquals(parseLong(matcher.group(1)), parameter.payerAccountId().getShard());
+        assertEquals(parseLong(matcher.group(2)), parameter.payerAccountId().getRealm());
+        assertEquals(parseLong(matcher.group(3)), parameter.payerAccountId().getNum());
+        assertEquals(parseLong(matcher.group(4)), parameter.validStart().getEpochSecond());
+        assertEquals(parseInt(matcher.group(5)), parameter.validStart().getNano());
```
steven-sheehy, line 84: Mirror node standard is to use assertj's assertThat() for all assertions.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/common/TransactionIdParameterTest.java
```diff
@@ -0,0 +1,86 @@
...
+                assertNull(parameter);
+                return;
+            }
+            assertThrows(InvalidParametersException.class, () -> TransactionIdParameter.valueOf(transactionId));
+            return;
+        }
+
+        Matcher matcher = TRANSACTION_ID_PATTERN.matcher(transactionId);
+        assertTrue(matcher.matches());
+        assertEquals(5, matcher.groupCount());
```
steven-sheehy, line 75: I don't think re-matching via a duplicated regex to extract the groups provides value. It could just mean you have the same bug in both. Since there's only one valid entry you could just manually verify that the values match the 0.0.3-1234567890-123

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/common/TransactionIdOrHashParameterTest.java
```diff
@@ -0,0 +1,103 @@
...
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+import org.springframework.test.context.junit.jupiter.SpringExtension;
+
+@ExtendWith(SpringExtension.class)
+class TransactionIdOrHashParameterTest {
+
+    private static final Pattern TRANSACTION_ID_PATTERN = Pattern.compile("^(\\d+)\\.(\\d+)\\.(\\d+)-(\\d{1,19})-(\\d{1,9})$");
```
steven-sheehy, line 40: Same comments from TransactionIdParameterTest about duplicating and using this pattern to extract groups.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/OpcodeServiceImpl.java
```diff
@@ -0,0 +1,202 @@
...
+                .block(blockType)
+                .build();
+    }
+
+    private Address getSenderAddress(ContractResult contractResult) {
+        return entityDatabaseAccessor.evmAddressFromId(contractResult.getSenderId(), Optional.empty());
+    }
+
+    private Address getReceiverAddress(ContractResult contractResult) {
+        return entityDatabaseAccessor.evmAddressFromId(EntityId.of(contractResult.getContractId()), Optional.empty());
```
steven-sheehy, line 177: Use EthereumTransaction.toAddress if available

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/ControllerExceptionHandler.java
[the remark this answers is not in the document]
victor-yanev: I tried the other way, but it didn’t work. This worked so I went with this to avoid duplication.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/common/TransactionIdParameterTest.java
victor-yanev: I don’t agree, this would mask out bugs because if the pattern changes in the actual class it will also change in the tests and the bug will be present in both files, so it will be masked out. If we declare it here, we know for sure that if the pattern changes in the actual class, those tests would fail here and we will catch any potential bugs.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/OpcodesController.java
victor-yanev: Thanks, it’s a leftover from the old validations. I will remove it.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/common/TransactionIdParameterTest.java
victor-yanev: Sure, I can validate it directly against the value and remove this pattern matching in the assertions.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/CallServiceParametersBuilderImpl.java
victor-yanev: Good point, I thought the implementation would try to fetch the RecordFile by the consensus timestamp, but since we are already giving the block number it makes sense to also update the cache for the other repository method which takes it by its index.

victor-yanev: Another option is to directly pass the `RecordFile` (instead of the block number) in separate parameters (e.g., `CallDebugServiceParameters`) which will be used only for the opcode call, so that we don’t couple the two together.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/OpcodesController.java
steven-sheehy: Also, it's debatable but I think the mapping to view model response should be moved back to the controller.

victor-yanev: I don't have a strong opinion on this one I just moved it to the service because I thought it would be better to also cover it with the service integration tests because we are calling the database accessor service and we only have unit tests for the controller. But if you prefer, I can move this back to the controller.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/RecordFileServiceImpl.java
```diff
@@ -0,0 +1,57 @@
...
+import com.hedera.mirror.web3.exception.BlockNumberOutOfRangeException;
+import com.hedera.mirror.web3.repository.RecordFileRepository;
+import com.hedera.mirror.web3.viewmodel.BlockType;
+import java.util.Optional;
+import lombok.RequiredArgsConstructor;
+import org.springframework.stereotype.Service;
+
+@Service
+@RequiredArgsConstructor
+public class RecordFileServiceImpl implements RecordFileService {
```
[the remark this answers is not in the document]
victor-yanev: We need this to be public, because we are invoking the constructor to define the bean inside `com.hedera.mirror.web3.controller.OpcodeControllerTest.TestConfig` (another package):
```java
class OpcodeControllerTest {
    ...

    @TestConfiguration
    public static class TestConfig {
        ...
        
        @Bean
        RecordFileService recordFileService(final RecordFileRepository recordFileRepository) {
            return new RecordFileServiceImpl(recordFileRepository);
        }

        @Bean
        OpcodeService opcodeService(final RecordFileService recordFileService,
                                    final ContractTransactionHashRepository contractTransactionHashRepository,
                                    final EthereumTransactionRepository ethereumTransactionRepository,
                                    final TransactionRepository transactionRepository,
                                    final ContractResultRepository contractResultRepository,
                                    final EntityDatabaseAccessor entityDatabaseAccessor) {
            return new OpcodeServiceImpl(
                    recordFileService,
                    contractTransactionHashRepository,
                    ethereumTransactionRepository,
                    transactionRepository,
                    contractResultRepository,
                    entityDatabaseAccessor
            );
        }
    }
}
```

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/common/TransactionIdOrHashParameterTest.java
victor-yanev: https://github.com/hashgraph/hedera-mirror-node/pull/8245#discussion_r1623222865

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/common/TransactionIdParameterTest.java
xin-hedera: I'd ask why do we even test the regex pattern? That is implementation details. So I'd suggest to keep it private and remove whatever logic in test cases which validates the pattern

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/config/EvmConfiguration.java
```diff
@@ -164,6 +165,18 @@ CacheManager cacheManagerRecordFileIndex() {
         return caffeineCacheManager;
     }
 
+    @Bean(CACHE_MANAGER_RECORD_FILE_TIMESTAMP)
+    CacheManager cacheManagerRecordFileTimestamp() {
+        final var caffeine = Caffeine.newBuilder()
+                .expireAfterWrite(10, TimeUnit.MINUTES)
```
xin-hedera, line 171:
```suggestion
                .expireAfterAccess(10, TimeUnit.MINUTES)
```

perhaps also follow the pattern to make it a configurable string with a default value like other cache managers

hedera-mirror-web3/src/test/resources/config/application.yml
```diff
@@ -1,5 +1,8 @@
 hedera:
   mirror:
+    opcode:
```
xin-hedera, line 3: 1. why isn't it added to the main application.yml, instead here in test?
2. why isn't this under `hedera.mirror.web3`?
3. regardless, we need to update `configuration.md` to add the new property

xin-hedera (changes requested):

hedera-mirror-web3/src/test/resources/config/application.yml
victor-yanev: 1. In the main application, we want this property to be disabled by default - that’s why I have added this only to the tests
2. Thanks, I will move it under `hedera.mirror.web3`
3. Sure, I will also update the `configuration.md` file with the new property

xin-hedera: our convention is add it to main `application.yml` with default value, here it should be `false`. Override it in test application.yml if a different value is more convenient for testing purpose. So please also add it to main application.yml.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/common/TransactionIdParameterTest.java
xin-hedera: @victor-yanev bump it up

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/common/TransactionIdParameter.java
```diff
@@ -0,0 +1,54 @@
...
+
+    private static final Pattern TRANSACTION_ID_PATTERN = Pattern.compile("^(\\d+)\\.(\\d+)\\.(\\d+)-(\\d{1,19})-(\\d{1,9})$");
+
+    public static TransactionIdParameter valueOf(String transactionId) throws InvalidParametersException {
+        if (transactionId == null) {
+            return null;
+        }
+
+        Matcher matcher = TRANSACTION_ID_PATTERN.matcher(transactionId);
+        if (!matcher.matches() || matcher.groupCount() != 5) {
```
xin-hedera, line 35: Isn't `matcher.groupCount() != 5` redundant? This causes partial testing coverage btw.

victor-yanev: I know it's most likely redundant, but I kept it just in case as a sanity check before moving forward with the parsing. 

I can remove it if you prefer.

hedera-mirror-web3/src/test/resources/config/application.yml
victor-yanev: Done.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/common/TransactionIdParameterTest.java
victor-yanev: Done. It has been removed.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/OpcodesController.java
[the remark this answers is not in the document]
victor-yanev: CC: @xin-hedera

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/config/EvmConfiguration.java
victor-yanev: The expiration time is also hardcoded in all other cache manager beans here.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/common/TransactionIdParameter.java
victor-yanev: Done.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/OpcodesController.java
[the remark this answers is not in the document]
xin-hedera: agree we can leave it as a future task.

xin-hedera (dismissed):
LGTM

xin-hedera (approved):
LGTM

PR #8253, opened 2024-05-08
Trigger k6 rest testsuite when acceptance test pod is deleted

### LINKED ISSUE #8237: Auto-triggered k6 tests may impact acceptance tests
### Description

The testkube k6 performance test suite is set to kick off when the rest-test pod completes during a deployment. However there may be times, especially when the importer is lagged behind, the load put by the k6 performance tests will impact the acceptance tests job and cause it to fail.

### Steps to reproduce

Deploy new version to the staging environment, watch when the k6 tests kick off, the acceptance tests are still running.

### Additional context

_No response_

### Hedera network

other

### Version

v0.104.0-rc2

### Operating system

None

xin-hedera:
**Description**:

This PR change how k6 rest testsuite is triggered

- Move k6 rest testsuite trigger label to acceptance test pod

**Related issue(s)**:

Fixes #8237 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/charts/hedera-mirror/templates/_helpers.tpl
+++ b/charts/hedera-mirror/templates/_helpers.tpl
@@ -98,3 +98,11 @@ Create the name of the service account to use
 {{- default "default" .Values.serviceAccount.name -}}
 {{- end -}}
 {{- end -}}
+
+{{/*
+Acceptance Test Labels
+*/}}
+{{- define "hedera-mirror-acceptance.labels" -}}
+{{ include "hedera-mirror.labels" . }}
+test-trigger: rest-trigger
+{{- end -}}
--- a/charts/hedera-mirror-rest/templates/_helpers.tpl
+++ b/charts/hedera-mirror-rest/templates/_helpers.tpl
@@ -47,14 +47,6 @@ Labels
 {{- end -}}
 
 {{/*
-Rest Test Labels
-*/}}
-{{- define "hedera-mirror-rest-test.labels" -}}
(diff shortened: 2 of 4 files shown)


charts/hedera-mirror/templates/_helpers.tpl
```diff
@@ -98,3 +98,11 @@ Create the name of the service account to use
 {{- default "default" .Values.serviceAccount.name -}}
 {{- end -}}
 {{- end -}}
+
+{{/*
+Acceptance Test Labels
+*/}}
+{{- define "hedera-mirror-acceptance.labels" -}}
+{{ include "hedera-mirror.labels" . }}
+test-trigger: rest-trigger
```
steven-sheehy, line 107: The value rest-trigger no longer makes sense. We should rename it to something else here and in the matchLabels. Not sure if `true` needs to be quoted.

```suggestion
test-trigger: true
```

xin-hedera: it can be confusing, however, we can define the value `rest-trigger` as it means to trigger the rest testsuite, not that whoever has the label is related to rest service

steven-sheehy: Probably we should rename the rest suite in the future since it now contains web3. We can do that later though.

steven-sheehy (approved):

mgoelswirlds (approved):
LGTM

PR #8256, opened 2024-05-08
8255 add index to nft allowance table for spender centric queries

### LINKED ISSUE #8255: Add index to nft_allowance table to increase performance of spender centric queries
### Problem

Using the boolean `owner` query parameter, the account ID specified in the NFT allowances REST API can be indicated to be that of the token owner or the spender (via allowance). The current PK index on the `nft_allowance` table optimizes for the owner scenario, but performance for the spender scenario is  sub-optimal.

### Solution

Add an additional index to also optimize for spender centric queries. E.g.: `btree(spender, owner, token_id)`. Spender centric results are sorted by owner, and a token ID may or may not be specified as a query parameter.

- Add v1 and v2 migrations to define the index. The `nft_allowances` table in mainnet-staging has only about 17k rows, so the migration will be fast.
- Update related readme files.

### Alternatives

_No response_

jascks:
**Description**:

Add index to `nft_allowance` to support allowance queries based on spender account ID (`owner=false`). The existing PK index optimizes for the owner ID scenario, but performance for the spender scenario is  sub-optimal.

- Add v1 and v2 migrations to define the index.
- Update related readme files.

**Related issue(s)**:

Fixes #8255

**Notes for reviewer**:
The `nft_allowances` table in mainnet-staging has only about 17k rows, so the migration will be fast

**Checklist**


steven-sheehy (approved):
LGTM

xin-hedera (approved):
LGTM

mgoelswirlds (approved):
LGTM

PR #8261, opened 2024-05-09
Nft allowances Range additions

### LINKED ISSUE #8176: NFT allowance rest API should return only approved for all allowances
### Problem

Currently the NFT allowance API returns approve_for_all=true and false values. 

### Solution

- Modify the NFT allowance API to return values only for `approve_for_all=true`
Example :
**GET** `/api/v1/accounts/0.0.8488/allowances/nfts?limit=3&owner=false`

```
{
  "allowances": [
    {
      "owner": "0.0.1000",
      "spender": "0.0.8488",
      "token_id": "0.0.1033",
      "timestamp": {
        "from": "1633466229.96874612",
        "to": null
      }
    },
    {
      "owner": "0.0.1000",
      "spender": "0.0.8488",
      "token_id": "0.0.1034",
      "timestamp": {
        "from": "1633466229.96874618",
        "to": null
      }
    }
  ],
  "links": {
	   "next": "/api/v1/accounts/0.0.8488/allowances/nfts?limit=3&order=asc&account.id=gte:1000&token.id=gt:0.0.1034"
	}
}
```

### Alternatives

_No response_

mgoelswirlds:
**Description**:
This PR adds the support for greater and lower bound conditions for query parameters in the request.

This PR modifies 
* The service layer and the repository layer to handle greater and lower bound conditions for query parameters.
* Adds extra parameter validations as per the openapi spec
* Shortcuts invalid ranges

**Related issue(s)**:

Fixes #8176

**Notes for reviewer**:
- This PR will have changes to the next link depending on the Next link PR
- This PR does not add all validations to Controller tests since they are validated in services tests and these take longer to execute.

**Checklist**


hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/repository/NftAllowanceRepositoryCustomImpl.java
```diff
@@ -115,25 +148,27 @@ private Condition getBaseCondition(
...
-        // No secondary condition if there is no token parameter, or the primary sort parameter's operator is EQ. Note
-        // that
-        // it's guaranteed that there must be a primary sort parameter when token parameter exists
-        if (tokenParam == null || primarySortParam.operator() == RangeOperator.EQ) {
+
+        // No secondary condition if there is no primary parameter bound or no token parameter, or the primary sort
+        // parameter's operator is EQ. Note that it's guaranteed that there must be
+        // a primary sort parameter when token parameter exists except in case of optimized Range
+
+        if (primary == null || token == null || primary.operator() == RangeOperator.EQ) {
```
mgoelswirlds, line 160: Need to have a null check for primary because when we optimize a range for lte == gte , we make it lte, in that case param can be null.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/NftAllowanceServiceImpl.java
```diff
@@ -39,22 +44,96 @@ public Collection<NftAllowance> getNftAllowances(NftAllowanceRequest request) {
 
         checkOwnerSpenderParamValidity(ownerOrSpenderId, token);
 
-        verifyRangeId(token);
-        verifyRangeId(ownerOrSpenderId);
+        if (!CollectionUtils.isEmpty(ownerOrSpenderId)) {
+            var bounds = getRange(ownerOrSpenderId);
+
+            // Shortcutting when the range is invalid
+            if (bounds.lower > bounds.upper) {
```
mgoelswirlds, line 51: Returning empty result when range is invalid.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/NftAllowanceServiceImpl.java
```diff
@@ -39,22 +44,96 @@ public Collection<NftAllowance> getNftAllowances(NftAllowanceRequest request) {
...
+                return Collections.emptyList();
+            } else if (bounds.lower == bounds.upper) {
+                var optimizedRangeParam = new EntityIdRangeParameter(RangeOperator.LTE, EntityId.of(bounds.upper));
+                request.setOwnerOrSpenderId(List.of(optimizedRangeParam));
```
mgoelswirlds, line 55: When lower bound and upper bound are equal we optimize it to just be lower bound inclusive

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/NftAllowanceServiceImpl.java
```diff
@@ -39,22 +44,96 @@ public Collection<NftAllowance> getNftAllowances(NftAllowanceRequest request) {
...
+        }
 
         return repository.findAll(request, id);
     }
 
-    private static void verifyRangeId(EntityIdRangeParameter idParam) {
-        if (idParam != null && idParam.operator() == RangeOperator.NE) {
-            throw new IllegalArgumentException("Invalid range operator ne. This operator is not supported");
+    @SuppressWarnings({"java:S131"})
+    private Bound getRange(List<EntityIdRangeParameter> ownerOrSpenderIds) {
```
jascks, line 63: Should this method, along with `verifyRangeId()`, perhaps `checkOwnerSpenderParamValidity()`, along with the `Bound` record be pushed to the common `Utils` class? They seem generally useful for future API endpoints. The methods are already `static`.

mgoelswirlds: You have a good point as to they do not need to be static.I will change that part.
 But since they are very specific to the allowances API , I think it makes sense to keep them local here as they might not be used by any other API.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/repository/NftAllowanceRepositoryTest.java
```diff
@@ -226,42 +232,80 @@ private void populateNftRequestMap() {
...
+                        .tokenId(List.of(new EntityIdRangeParameter(RangeOperator.EQ, EntityId.of(tokenIds.get(1)))))
                         .limit(4)
                         .order(Direction.ASC)
                         .build(),
                 List.of(new Tuple(1, 0, 1)));
+        nftAllowanceRequests.put(
+                NftAllowanceRequest.builder()
+                        .isOwner(true)
+                        .accountId(new EntityIdNumParameter(EntityId.of(owners.get(1))))
+                        .ownerOrSpenderId(
```
mgoelswirlds, line 300: This is a corner case , which does follow the validation rules as per the spec , but eventually it ends up ignoring the `LTE` operator for token.id. 
Is that acceptable?

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/NftAllowanceServiceImpl.java
jascks: Okay.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/NftAllowanceRequest.java
```diff
@@ -37,7 +38,7 @@ public class NftAllowanceRequest {
     @Builder.Default
     private boolean isOwner = true;
 
-    private EntityIdRangeParameter ownerOrSpenderId;
+    private List<EntityIdRangeParameter> ownerOrSpenderId;
 
-    private EntityIdRangeParameter tokenId;
+    private List<EntityIdRangeParameter> tokenId;
```
edwin-greene, line 43: Nit, may be clearer if these were plural now: `ownerOrSpenderIds`, and `tokenIds`

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/repository/NftAllowanceRepositoryCustomImpl.java
```diff
@@ -64,10 +64,19 @@ public Collection<NftAllowance> findAll(NftAllowanceRequest request, EntityId ac
         var primarySortParam = request.getOwnerOrSpenderId();
         var tokenParam = request.getTokenId();
```
edwin-greene, line 65: Nit, these could be clearer as plurals, `primarySortParams` and `tokenParams`

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/repository/NftAllowanceRepositoryCustomImpl.java
```diff
@@ -64,10 +64,19 @@ public Collection<NftAllowance> findAll(NftAllowanceRequest request, EntityId ac
...
 
+        Bound primaryBounds = getBounds(primarySortParam);
+        Bound tokenBounds = getBounds(tokenParam);
```
edwin-greene, line 68: Nit these can be `var` instead of `Bound`

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/repository/NftAllowanceRepositoryCustomImpl.java
```diff
@@ -77,10 +86,33 @@ public Collection<NftAllowance> findAll(NftAllowanceRequest request, EntityId ac
                 .fetchInto(NftAllowance.class);
     }
 
-    private Condition getBaseCondition(
+    private Bound getBounds(List<EntityIdRangeParameter> primarySortParam) {
+        EntityIdRangeParameter lowerBoundParam = null;
+        EntityIdRangeParameter upperBoundParam = null;
+
+        if (primarySortParam != null) {
+            for (EntityIdRangeParameter param : primarySortParam) {
```
edwin-greene, line 94: Nit, can be `var` instead of `EntityIdRangeParameter`

jascks: Not when setting the value to null though.

jascks: Sorry @edwin-greene, looking at this again, you are quite correct for the for loop on line 94.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/repository/NftAllowanceRepositoryCustomImpl.java
```diff
@@ -115,25 +148,27 @@ private Condition getBaseCondition(
...
+    private Condition getMiddleCondition(
             EntityIdRangeParameter primarySortParam,
             EntityIdRangeParameter tokenParam,
             TableField<NftAllowanceRecord, Long> primarySortField) {
-        // No secondary condition if there is no token parameter, or the primary sort parameter's operator is EQ. Note
-        // that
-        // it's guaranteed that there must be a primary sort parameter when token parameter exists
-        if (tokenParam == null || primarySortParam.operator() == RangeOperator.EQ) {
+
+        // No secondary condition if there is no primary parameter bound or no token parameter, or the primary sort
```
jascks, line 156: Total nit. Since this was renamed, seems the nomenclature is now middle condition.

```java
// No middle condition if there is no...
```

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/NftAllowanceServiceImpl.java
```diff
@@ -33,28 +38,102 @@ public class NftAllowanceServiceImpl implements NftAllowanceService {
...
+            List<EntityIdRangeParameter> ownerOrSpenderParams, List<EntityIdRangeParameter> tokenParams) {
+
+        if (CollectionUtils.isEmpty(ownerOrSpenderParams)) {
+            if (!CollectionUtils.isEmpty(tokenParams)) {
+                throw new IllegalArgumentException("token.id parameter must have account.id present");
+            } else {
+                return;
+            }
+        }
+        var accountOperators = verifyRangeId(ownerOrSpenderParams);
```
xin-hedera, line 118: make `verifyRangeId` return a record object

`private record Operators(boolean hasEq, boolean hasGtGte, boolean hasLtLte) {};`

then the validations can be easier to read.

xin-hedera (changes requested):

xin-hedera:
How are we going to merge this without the next link fix?

Given the http request from the user, e.g., `/api/v1/200/allowances/nfts?account.id=gt:200&account.id=lt:800&order=asc`, the response may have a next link like `/api/v1/200/allowances/nfts?account.id=gt:200&account.id=lt:800&order=asc&account.id=gt:202`, which violates `max=2` constraint.

Do we ship it and tell users that's a known bug?

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/NftAllowanceServiceImpl.java
```diff
@@ -33,28 +38,102 @@ public class NftAllowanceServiceImpl implements NftAllowanceService {
...
+                case GT -> lowerBound = param.value().getId() + 1;
+                case LTE -> upperBound = param.value().getId();
+                case GTE -> lowerBound = param.value().getId();
+            }
+        }
+
+        return new Bound(lowerBound, upperBound);
+    }
+
+    private static List<RangeOperator> verifyRangeId(List<EntityIdRangeParameter> idParams) {
```
jascks, line 78: I'd consider returning `Set<RangeOperator>`  since it does not seem order is important and that `contains()` is invoked on this list of items. This is O(n/2) for a `List` and O(1) for a `Set`. Though, I understand `RangeOperator` has only six possible values and it's unlikely an API client would provide all of those, so the performance difference in reality is not likely to mementous.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/repository/NftAllowanceRepositoryCustomImpl.java
```diff
@@ -77,10 +86,33 @@ public Collection<NftAllowance> findAll(NftAllowanceRequest request, EntityId ac
                 .fetchInto(NftAllowance.class);
     }
 
-    private Condition getBaseCondition(
+    private Bound getBounds(List<EntityIdRangeParameter> primarySortParam) {
```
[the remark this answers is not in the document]
xin-hedera: Or create a new data class to pass the query filters from service layer to repo layer, and in the new class, instead of `EntityIdRangeParameters`, use explicit lower / upper bound for ownerOrSpender and tokenId. That can also eliminate the duplicate bound analysis in repository layer.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/NftAllowanceServiceImpl.java
mgoelswirlds: This surely will makes things radable , but I need to know the exact operators and count for certain validations , which would lead me to add all 5 operators to the record.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/NftAllowanceServiceImpl.java
```diff
@@ -33,28 +37,125 @@ public class NftAllowanceServiceImpl implements NftAllowanceService {
...
+                }
+                case RangeOperator.LTE -> {
+                    hasLte = true;
+                    countLtOperator++;
+                }
+                case RangeOperator.EQ -> {
+                    hasEq = true;
+                    countEqOperator++;
+                }
+            }
```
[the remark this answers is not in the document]
xin-hedera: in addition, for the validation here, there is no difference between lt or lte, same for gt or gte. instead of individual counters, you just need `countLtLte` and `countGtGte`.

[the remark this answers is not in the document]
mgoelswirlds: I had done that but the place where I face dan issue was for the following validations
`Single occurrence only supported.Requires the presence of an lte or eq account.id query`, here I specifically need to know lte and similarly gte for the other one.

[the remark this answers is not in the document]
xin-hedera: I see. you can add helper functions to the record class

```java
private record Operators(...) {
  public boolean hasGtGte() {
    return hasGt || hasGte;
  }
}
```

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/NftAllowanceServiceImpl.java
```diff
@@ -33,28 +40,166 @@ public class NftAllowanceServiceImpl implements NftAllowanceService {
...
+        }
+
+        if (countGtGte > 1 || countLtLte > 1 || countEq > 1) {
+            throw new IllegalArgumentException("Single occurrence only supported.");
+        }
+        if (countEq == 1 && (countGtGte != 0 || countLtLte != 0)) {
+            throw new IllegalArgumentException("Can't support both range and equal for this parameter.");
+        }
+        return new Operators(countEq > 0, hasGt, hasGte, hasLt, hasLte);
```
steven-sheehy, line 165: Again, think this logic would be better served to be encapsulated in `Bound` for reuse and to avoid creating this `Operators`. We should instead calculate an `EnumMap<RangeOperator, Integer>` in the `Bound` constructor that we can add a `boolean hasOperator(RangeOperator.EQ)` helper method to Bound to use for multi-parameter validation.

steven-sheehy (changes requested):

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/NftAllowanceServiceImpl.java
mgoelswirlds: The `/allowances/tokens` API returns an empty response.So keeping it consistent.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/NftAllowanceServiceImpl.java
```diff
@@ -33,28 +40,166 @@ public class NftAllowanceServiceImpl implements NftAllowanceService {
...
+                            || param.operator() == RangeOperator.EQ) {
+                        lowerBoundParam = param;
+                    } else if (param.operator() == RangeOperator.LT || param.operator() == RangeOperator.LTE) {
+                        upperBoundParam = param;
+                    }
+                }
+            }
+        }
+
+        return new Bound(lowerBoundParam, upperBoundParam);
```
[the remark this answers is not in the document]
mgoelswirlds: Just to clarify , are you suggesting to change Bound definition to be like
`public record Bound(List<EntityIdRangeParameter> params)`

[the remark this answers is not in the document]
steven-sheehy: Correct

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/NftAllowanceServiceImpl.java
```diff
@@ -33,28 +35,72 @@ public class NftAllowanceServiceImpl implements NftAllowanceService {
...
-        var token = request.getTokenId();
-        var id = entityService.lookup(request.getAccountId());
+        var ownerOrSpenderId = request.getOwnerOrSpenderIds();
+        var token = request.getTokenIds();
 
         checkOwnerSpenderParamValidity(ownerOrSpenderId, token);
 
-        verifyRangeId(token);
-        verifyRangeId(ownerOrSpenderId);
+        if (ownerOrSpenderId.getLowerBound() != null
```
mgoelswirlds, line 43: To be able to return an empty reponse as in the `allowances/tokens` endpoint , I had to keep this check here. If I move it to Bound I can throw an exception `Invalid range provided` or something, but that will be inconsistent with the other similar API.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/NftAllowanceServiceImpl.java
```diff
@@ -33,28 +40,166 @@ public class NftAllowanceServiceImpl implements NftAllowanceService {
...
+    }
+
+    @SuppressWarnings({"java:S131"})
+    private Bound getRange(List<EntityIdRangeParameter> idRangeParameters) {
+        EntityIdRangeParameter lowerBoundParam = null;
+        EntityIdRangeParameter upperBoundParam = null;
+
+        if (idRangeParameters != null) {
+            for (EntityIdRangeParameter param : idRangeParameters) {
+                if (param != null) {
```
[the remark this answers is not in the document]
mgoelswirlds: the param will not be null. So adjusted that.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/Bound.java
```diff
@@ -0,0 +1,99 @@
...
+            case RangeOperator.LT -> cardinality.put(
+                    RangeOperator.LT, cardinality.getOrDefault(RangeOperator.LT, 0) + 1);
+            case RangeOperator.LTE -> cardinality.put(
+                    RangeOperator.LTE, cardinality.getOrDefault(RangeOperator.LTE, 0) + 1);
+            case RangeOperator.EQ -> cardinality.put(
+                    RangeOperator.EQ, cardinality.getOrDefault(RangeOperator.EQ, 0) + 1);
+            case RangeOperator.NE -> cardinality.put(
+                    RangeOperator.NE, cardinality.getOrDefault(RangeOperator.NE, 0) + 1);
+        }
+    }
```
steven-sheehy, line 94: Don't really need this switch statement or method as a whole. Just put this inline in the for loop:

```java
cardinality.merge(param.operator(), 1, Math::addExact);
```

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/Bound.java
```diff
@@ -0,0 +1,99 @@
...
+@Getter
+public class Bound {
+
+    private EntityIdRangeParameter lower;
+    private EntityIdRangeParameter upper;
+    private final EnumMap<RangeOperator, Integer> cardinality = new EnumMap<>(RangeOperator.class);
+
+    public Bound(List<EntityIdRangeParameter> params, boolean primarySortField) {
+
+        if (params != null) {
```
steven-sheehy, line 35: switch to fail fast to reduce nesting

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/Bound.java
```diff
@@ -0,0 +1,99 @@
...
+package com.hedera.mirror.restjava.service;
+
+import com.hedera.mirror.common.domain.entity.EntityId;
+import com.hedera.mirror.restjava.common.EntityIdRangeParameter;
+import com.hedera.mirror.restjava.common.RangeOperator;
+import java.util.EnumMap;
+import java.util.List;
+import lombok.Getter;
+
+@Getter
```
steven-sheehy, line 26: We shouldn't expose cardinality map

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/NftAllowanceServiceImpl.java
```diff
@@ -33,28 +33,64 @@ public class NftAllowanceServiceImpl implements NftAllowanceService {
...
-            EntityIdRangeParameter ownerOrSpenderId, EntityIdRangeParameter token) {
-        if (ownerOrSpenderId == null && token != null) {
-            throw new IllegalArgumentException("token.id parameter must have account.id present");
+    private static int getCountGtGte(Bound ids) {
+        return ids.getCardinality(RangeOperator.GT) + ids.getCardinality(RangeOperator.GTE);
+    }
+
+    private static int getCountLtLte(Bound ids) {
+        return ids.getCardinality(RangeOperator.LT) + ids.getCardinality(RangeOperator.LTE);
+    }
```
steven-sheehy, line 77: Can add a generic `Bound.getCardinality(RangeOperator... operators)` or move `getCountLtLte()` to `Bound` and rename to `getLowerCardinality()`.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/Bound.java
```diff
@@ -0,0 +1,99 @@
...
+                } else if (param.hasUpperBound()) {
+                    upper = param;
+                }
+                populateCardinality(param);
+            }
+            long adjustedLower = adjustLowerBound();
+            long adjustedUpper = adjustUpperBound();
+
+            if (primarySortField && adjustedLower > adjustedUpper) {
+                throw new IllegalArgumentException("Invalid range provided");
```
steven-sheehy, line 49: Why does it matter if it's a primary sort field to validate range? It should be for all bounds.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/Bound.java
```diff
@@ -0,0 +1,99 @@
...
+import com.hedera.mirror.restjava.common.RangeOperator;
+import java.util.EnumMap;
+import java.util.List;
+import lombok.Getter;
+
+@Getter
+public class Bound {
+
+    private EntityIdRangeParameter lower;
+    private EntityIdRangeParameter upper;
```
steven-sheehy, line 30: Probably in the future we'll want to make `Bound` work generically for `RangeParameter`

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/NftAllowanceServiceImpl.java
```diff
@@ -33,28 +33,64 @@ public class NftAllowanceServiceImpl implements NftAllowanceService {
...
     }
 
-    private static void verifyRangeId(EntityIdRangeParameter idParam) {
-        if (idParam != null && idParam.operator() == RangeOperator.NE) {
-            throw new IllegalArgumentException("Invalid range operator ne. This operator is not supported");
+    private static void checkOwnerSpenderParamValidity(Bound ownerOrSpenderParams, Bound tokenParams) {
+
+        if (ownerOrSpenderParams.getLower() == null
+                && ownerOrSpenderParams.getUpper() == null
+                && (tokenParams.getLower() != null || tokenParams.getUpper() != null)) {
```
steven-sheehy, line 50: Add a helper to `Bound` to clean this up:

```java
public boolean isEmpty() {
    return lower == null && upper == null;
}
```

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/NftAllowanceServiceImpl.java
```diff
@@ -33,28 +33,64 @@ public class NftAllowanceServiceImpl implements NftAllowanceService {
...
+        }
+
+        if (singleOperatorCheck(ids)) {
+            throw new IllegalArgumentException("Single occurrence only supported.");
+        }
+
+        if (ids.getCardinality(RangeOperator.EQ) == 1 && (getCountGtGte(ids) != 0 || getCountLtLte(ids) != 0)) {
+            throw new IllegalArgumentException("Can't support both range and equal for this parameter.");
         }
     }
```
steven-sheehy, line 95: These single field validations can be moved into `Bound` helper methods that can be optionally called by various service layers. This allows for reuse and encapsulation.

```java
public void verifyUnsupported(RangeOperator unsupportedOperator) {...}
public void verifySingleOccurrence() {...}
public void verifyEqualOrRange() {...}
```

steven-sheehy (changes requested):

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/Bound.java
mgoelswirlds: This is the scenario for our current API where that is relevant:

```
for a specific owner, given we have (spender, token) pairs

(100, 200)
(100, 201)
(200, 50)


and for the request account.id >= 100 && token.id >= 200 & account.id <=200 & token.id <= 50) 
we should return all 3 rows
if we shortcut on token id, we return nothing, and that is wrong
```

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/Bound.java
mgoelswirlds: Yes. makes sense. Or to have different abstractions of Bound.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/Bound.java
steven-sheehy: I think you're confusing input validation with rows returned from the db. The scenarios you list are for the latter. For user input, we need to make sure tokenid lower <= upper.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/repository/NftAllowanceRepositoryCustomImpl.java
```diff
@@ -61,13 +61,15 @@ public Collection<NftAllowance> findAll(NftAllowanceRequest request, EntityId ac
...
 
         var commonCondition = getCondition(primaryField, RangeOperator.EQ, accountId.getId());
-        var baseCondition = getBaseCondition(primarySortParam, tokenParam, primarySortField);
-        var secondaryCondition = getSecondaryCondition(primarySortParam, tokenParam, primarySortField);
-        var condition = commonCondition.and(baseCondition.or(secondaryCondition));
+        var lowerCondition = getOuterBoundCondition(primaryBounds.getLower(), tokenBounds.getLower(), primarySortField);
+        var middleCondition = getMiddleCondition(primaryBounds.getLower(), tokenBounds.getLower(), primarySortField)
+                .and(getMiddleCondition(primaryBounds.getUpper(), tokenBounds.getUpper(), primarySortField));
+        var upperCondition = getOuterBoundCondition(primaryBounds.getUpper(), tokenBounds.getUpper(), primarySortField);
+        var condition = commonCondition.and(lowerCondition.or(middleCondition).or(upperCondition));
```
xin-hedera, line 72: the repository will return incorrect data for such requests

- `/accounts/1001/allowances/nfts?account.id=gte:2001&account.id=lte:2001&token.id=gte:3000`
- `/accounts/1001/allowances/nfts?account.id=gte:2001&account.id=lte:2001&token.id=3000`

For the second, I'm not sure if we should return 400 to indicate for `token.id=` the only allowed operator for `account.id` is =

This also indicates there are missing test coverage.

xin-hedera (changes requested):

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/Bound.java
xin-hedera: for multiple-column pagination, we do concat the values to form a single value to compare and order.

For example, given two `long` columns `A` (primary) and `B`, with values `a` and `b`, concat them `a << 64 | b`.

When we have `A=gt:a1&A=lt:a2&B=gt:b1&B=lt:b2`, where the range`(a1, a2)` contains more than 1 discrete values and `(b1, b2)` is effectively empty, we don't throw an error since for a concatenated value ` (a3 << 64 | b3) > (a1 << 64 | b1`, `b3 > b1` isn't necessarily true.

However, we should add logic to throw error when `(a1, a2)` contains a single value, and `b1 > b2`.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/Bound.java
```diff
@@ -0,0 +1,120 @@
...
+        }
+    }
+
+    public void verifySingleOccurrence() {
+        if (this.getCardinality(RangeOperator.GT, RangeOperator.GTE) > 1
+                || this.getCardinality(RangeOperator.LT, RangeOperator.LTE) > 1
+                || this.getCardinality(RangeOperator.EQ) > 1) {
+            throw new IllegalArgumentException("Single occurrence only supported.");
+        }
+    }
```
steven-sheehy, line 111: This is an ambiguous error presented to the user. We can refactor it to make it clearer which operator is at fault. We may also consider passing in the query parameter name into the constructor to provide more accurate error messages.

```suggestion
        verifySingleOccurrence(RangeOperator.EQ);
        verifySingleOccurrence(RangeOperator.GT, RangeOperator.GTE);
        verifySingleOccurrence(RangeOperator.LT, RangeOperator.LTE);
    }
    
    private void verifySingleOccurrence(RangeOperator... rangeOperators) {
        if (this.getCardinality(rangeOperators) > 1) {
            throw new IllegalArgumentException("Only one range operator from %s is allowed for the given parameter".formatted(Arrays.toString(rangeOperators));
        }
    }
```

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/Bound.java
```diff
@@ -0,0 +1,120 @@
...
+    public boolean isEmpty() {
+        return lower == null && upper == null;
+    }
+
+    public boolean hasLowerAndUpper() {
+        return lower != null && upper != null;
+    }
+
+    public boolean hasEqualBounds() {
+        return getLower().value().getId() == getUpper().value().getId();
```
steven-sheehy, line 96:
```suggestion
        return Objects.equals(getLower().value(), getUpper().value());
```

steven-sheehy (changes requested):

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/Bound.java
steven-sheehy: Xin's scenario still needs to be addressed

mgoelswirlds: Missed this. WIll add it now.

mgoelswirlds: I would need to add that in the service layer as not everyone is going to have multi column pagination.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/Bound.java
```diff
@@ -0,0 +1,126 @@
...
+    public boolean isEmpty() {
+        return lower == null && upper == null;
+    }
+
+    public boolean hasLowerAndUpper() {
+        return lower != null && upper != null;
+    }
+
+    public boolean hasEqualBounds() {
+        return Objects.equals(getLower().value(), getUpper().value());
```
xin-hedera, line 97: - why `Object.equals` instead of just comparing the primitive long?
- NPE if either lower is null or upper is null
- inconsistency: other places just use instance variable `lower` / `upper`, here it's the getters.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/repository/NftAllowanceRepositoryCustomImpl.java
```diff
@@ -61,13 +62,28 @@ public Collection<NftAllowance> findAll(NftAllowanceRequest request, EntityId ac
...
+        var lowerPrimary = primaryBounds.getLower();
+        var upperPrimary = primaryBounds.getUpper();
+
+        // This is for range shortcutting when the GTE and LTE values are same. With GT and LT the values will be
+        // shortcut in the controller.
+        if (primaryBounds.hasLowerAndUpper() && primaryBounds.hasEqualBounds()) {
+            lowerPrimary = new EntityIdRangeParameter(
+                    RangeOperator.EQ, primaryBounds.getLower().value());
+            upperPrimary = null;
+        }
```
xin-hedera, line 75: what if there is `account.id=gt:10&account.id=lt:12`? this also means the comment is inaccurate. there seems to be missing test coverage as well.

xin-hedera (changes requested):

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/Bound.java
steven-sheehy: > why Object.equals instead of just comparing the primitive long?

`value()` is `EntityId` so you mean extracting id out of it and comparing? That's violating encapsulation of `EntityId` and doesn't make it compatible if we make `Bound` work with the higher level interface `EntityIdParameter` in the future.

Agree with other points which were the case before switching to `Objects.equals()` as well.

xin-hedera: Agree however `Bound` is not a generic class,  both `lower` and `upper` are of type `EntityIdRangeParameter`

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/repository/NftAllowanceRepositoryCustomImpl.java
mgoelswirlds: I have addressed that.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/Bound.java
mgoelswirlds: Needed to change that part of the code to used adjusted lower and upper bounds

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/repository/NftAllowanceRepositoryCustomImpl.java
```diff
@@ -61,13 +62,26 @@ public Collection<NftAllowance> findAll(NftAllowanceRequest request, EntityId ac
         var primaryField = byOwner ? NFT_ALLOWANCE.OWNER : NFT_ALLOWANCE.SPENDER;
         var primarySortField = byOwner ? NFT_ALLOWANCE.SPENDER : NFT_ALLOWANCE.OWNER;
 
-        var primarySortParam = request.getOwnerOrSpenderId();
-        var tokenParam = request.getTokenId();
+        var primaryBounds = request.getOwnerOrSpenderIds();
```
xin-hedera, line 65: there are still corner cases for which the logic here would get incorrect nft allowances, I'll see if there's a better way without introducing too much complexity

xin-hedera (approved):
LGTM. in case of any minor issues, can address in a follow-up PR

steven-sheehy (approved):
LGTM

PR #8264, opened 2024-05-09
Add linkFactory for NftAllowances Java REST API

### LINKED ISSUE #7982: Create a generic LinkFactory to return next links for rest java APIs
### Problem

Currently the next link logic is not in a proper service. 
Also address the issue stated in https://github.com/hashgraph/hedera-mirror-node/pull/7837#discussion_r1546419083

### Solution

- Add a new `LinkFactory` to create and test general link creation.
- API specific logic will be in a separate extractor.

```java
public interface LinkFactory {
    <T> Links create(Iterable<T> items, Pageable pageable, ParameterExtractor<T> extractor);

    interface ParameterExtractor<T> {
        Function<T, String> extract(String name);
    }
}

ParameterExtractor<NftAllowance> extractor = param -> switch (param) {
    case "account.id":
        yield NftAllowance::getSpender;
    case "token.id":
        yield NftAllowance::getTokenId;
    default:
        yield null;
};

var links = linkFactory.create(allowances, pageable, extractor);
response.setAllowances(allowances);
response.setLinks(links);
```
### Alternatives

_No response_

### LINKED ISSUE #8185: Nft Allowance API - Next link generation correction
### Description

With the current implementation a scenario with `account.id=gt:…&token.id=gt:…&order=desc` would generate wrong result.

### Steps to reproduce

1. Provide `account.id=gt:…&token.id=gt:…&order=desc` for nft allowance rest api will generate incorrect Next Link.

### Additional context

Fix next link generation to handle the above case to return `account.id=gt:…&token.id=gt:…&account.id=lte:…&token.id=lt:…order=desc`

### Hedera network

other

### Version

v.0.103.0

### Operating system

None

edwin-greene:
**Description**:

* Add a new `LinkFactory` for generic next link generation
* Fix next link generation for `account.id=gt:…&token.id=gt:…&order=desc`
* Move API specific logic into a separate data structure passed to the factory

**Related issue(s)**:

Fixes #7982
Fixes #8185

**Notes for reviewer**:

**Checklist**


hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactory.java
```diff
@@ -0,0 +1,69 @@
...
+package com.hedera.mirror.restjava.common;
+
+import com.hedera.mirror.rest.model.Links;
+import java.util.Map.Entry;
+import java.util.function.Function;
+import org.springframework.data.domain.Sort.Direction;
+import org.springframework.web.util.UriComponentsBuilder;
+
+public interface LinkFactory {
+    <T> Links create(T item, ParameterExtractor<T> extractor);
```
edwin-greene, line 26: The issue has a `Pageable` parameter here, after going through some iterations I wasn't sure that it was needed.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/controller/AllowancesControllerTest.java
```diff
@@ -242,24 +295,28 @@ void nftAllowancesOrderDesc() {
...
+                        .formatted(
+                                allowance2.getOwner(),
+                                EntityId.of(allowance2.getSpender()),
+                                EntityId.of(allowance2.getTokenId()));
 
         // Then
-        assertThat(result.getAllowances()).isEqualTo(mapper.map(List.of(allowance2)));
-        assertThat(result.getLinks().getNext()).isEqualTo(nextLink);
+        assertThat(result.getAllowances()).isEqualTo(mapper.map(List.of(allowance3, allowance2)));
+        assertThat(result.getLinks().getNext()).isEqualTo(next);
```
edwin-greene, line 319: When the NFT Allowances Range PR is merged, this test can be expanded to include following the next link. At the moment the next link contains a lower bound and upper bound for the account.id and token.id parameters, which isn't supported by the AllowancesController (until the aforementioned PR is merged).

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/controller/AllowancesController.java
```diff
@@ -70,21 +84,17 @@ NftAllowancesResponse getNftAllowancesByAccountId(
...
-            lastValues.put(ACCOUNT_ID, lastAccountId);
-            lastValues.put(TOKEN_ID, last.getTokenId());
-            next = Utils.getPaginationLink(false, lastValues, order);
+        response.setAllowances(allowances);
+        if (allowances.size() == limit) {
+            var last = CollectionUtils.lastElement(allowances);
+            var links = linkFactory.create(last, extractor);
+            response.links(links);
+        } else {
+            response.links(new Links());
```
steven-sheehy, line 95: The link factory should handle the if/else internally and return the appropriate links object so that here we just receive the links and set it.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/controller/AllowancesLinkFactory.java
```diff
@@ -0,0 +1,98 @@
...
+import com.hedera.mirror.restjava.common.RangeOperator;
+import jakarta.inject.Named;
+import jakarta.servlet.http.HttpServletRequest;
+import lombok.RequiredArgsConstructor;
+import org.springframework.data.domain.Sort;
+import org.springframework.web.util.UriComponentsBuilder;
+
+@Named
+@RequiredArgsConstructor
+public class AllowancesLinkFactory implements LinkFactory {
```
steven-sheehy, line 36: I think you misunderstood the intent of the ticket. The ticket was to help make the link generation more generic and reusable so it can be used by topic ID API but this PR actual makes it less generic and reusable. The `LinkFactory` was intending to replace the generic static Utils with a generic bean that could generate links for any endpoint. We should not have endpoint specific link factories.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactory.java
```diff
@@ -0,0 +1,69 @@
...
+    interface ParameterExtractor<T> {
+        Function<T, String> extract(String name);
+    }
+
+    /**
+     * Add a query parameter to the UriComponentsBuilder
+     *
+     * @return true if the query parameter was added, false otherwise
+     */
+    default <T> boolean addQueryParam(
```
steven-sheehy, line 37: Remove this since it should be in the generic `LinkFactoryImpl`.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactory.java
steven-sheehy: You will need it to make this factory generic for all endpoints.

steven-sheehy (changes requested):

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactory.java
edwin-greene: Thanks I have updated this to be a generic linkFactory

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/controller/AllowancesController.java
edwin-greene: Got it, I have updated the linkFactory to make the determination of empty links rather than the controller

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/controller/AllowancesLinkFactory.java
edwin-greene: Thanks, this has now been updated to be a generic link factory

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactory.java
edwin-greene: Thanks, this has been moved/refactored.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
```diff
@@ -0,0 +1,140 @@
...
+    private final HttpServletRequest request;
+
+    public <T> Links create(Iterable<T> items, Pageable pageable, ParameterExtractor<T> extractor) {
+        var links = new Links();
+        if (items == null) {
+            return links;
+        }
+
+        var itemsList = StreamSupport.stream(items.spliterator(), false).toList();
+        var lastItem = CollectionUtils.lastElement(itemsList);
```
xin-hedera, line 49:
```suggestion
        var lastItem = Iterables.getLast(items);
```

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
```diff
@@ -0,0 +1,140 @@
...
+                    builder.queryParam(secondarySort.getProperty(), RangeOperator.GT + ":" + secondarySortValue);
+                }
+            }
+        }
+    }
+
+    private List<String> addQueryParam(
+            String lastParameterValue, Entry<String, String[]> entry, Direction order, boolean inclusive) {
+        var queryParams = new ArrayList<String>();
+        if (lastParameterValue == null) {
```
xin-hedera, line 116: if it's null (which should not happen at all), all existing values should be added as `k=v` query param to the URL

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
```diff
@@ -0,0 +1,140 @@
...
+        }
+
+        for (var value : entry.getValue()) {
+            if (order.isDescending() || value.equalsIgnoreCase("false") || value.equalsIgnoreCase("true")) {
+                queryParams.add(value);
+            } else {
+                var operator = getOperator(value, inclusive);
+                queryParams.add(operator.toString() + ":" + lastParameterValue);
+            }
+        }
```
xin-hedera, line 127: multiple issues here including the logic I completely can't understand

- why when order=desc, all the existing values from the request URL are added and the value extracted from the last item is just ignored?
- what's special about the string "false" and "true"?
- in the else branch, why the same last value is added multiple times?

I think the idea is for a key `k`,

- if there is value `v` extracted from the last item, when order is descending, the function should remove all `lt` / `lte` values, replace it with `lt` or `lte` `v`, and keep all other existing values; when the order is ascending, apply the opposite logic
- if there is no value extracted, copy all the existing values

xin-hedera (changes requested):

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
edwin-greene: It would be great to remove `StreamSupport`, but the link factory still seems to need a means of getting the size of the items.  If the `limit` is greater than the number of items passed in, the LinkFactory will return an empty link. To get the number of items to compare to the `limit`, it is using `StreamSupport` to convert the iterable to a list.

steven-sheehy: Just pass in a List instead of an Iterable

steven-sheehy (changes requested):

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactory.java
```diff
@@ -0,0 +1,31 @@
...
+
+import com.hedera.mirror.rest.model.Links;
+import java.util.function.Function;
+import org.springframework.data.domain.Pageable;
+
+public interface LinkFactory {
+    <T> Links create(Iterable<T> items, Pageable pageable, ParameterExtractor<T> extractor);
+
+    interface ParameterExtractor<T> {
+        Function<T, String> extract(String name);
```
xin-hedera, line 27: let's change this to

```suggestion
        String extract(T item, String name);
```

So you don't have to deal with the situation that there is a function to extract parameter `p`, however the extracted value is `null`.

With the new `extract` function, given the item, and parameter name `p`, if it returns `null`, just add all existing values for `p` to the next link.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
```diff
@@ -0,0 +1,140 @@
...
+        var links = new Links();
+        if (items == null) {
+            return links;
+        }
+
+        var itemsList = StreamSupport.stream(items.spliterator(), false).toList();
+        var lastItem = CollectionUtils.lastElement(itemsList);
+        var limit = pageable.getPageSize();
+        var paramsMap = request.getParameterMap();
+        if (lastItem == null || limit > itemsList.size() || paramsMap.isEmpty()) {
```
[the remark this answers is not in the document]
edwin-greene: Thanks I have fixed this.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
```diff
@@ -0,0 +1,140 @@
...
+                case LIMIT -> builder.queryParam(LIMIT, limit);
+                case ORDER -> builder.queryParam(ORDER, order.name().toLowerCase());
+                default -> {
+                    var extract = extractor.extract(key);
+                    if (extract != null) {
+                        var value = extract.apply(lastItem);
+                        boolean inclusive = extractor.isInclusive(key);
+                        var queryValues = addQueryParam(value, entry, order, inclusive);
+                        builder.queryParam(key, queryValues);
+                    }
```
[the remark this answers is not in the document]
edwin-greene: Got it, I have updated this

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
```diff
@@ -0,0 +1,140 @@
...
+                }
+            }
+        }
+
+        var secondarySort = sorts.hasNext() ? sorts.next() : null;
+        addRangeBoundaries(primarySort, secondarySort, extractor, lastItem, builder);
```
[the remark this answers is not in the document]
edwin-greene: Thanks I have simplified the sorting and the linkFactory no longer has a concept of anything other than a primarySort (to detemine if ASC or DESC) and a lastSort (to determine if inclusive or not). The linkFactory still iterates the sorting to determine if any sorts passed by the `Pageable` have extractable values that were not included in the parameters.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
```diff
@@ -0,0 +1,140 @@
...
+                            extractor.extract(secondarySort.getProperty()).apply(lastItem);
+                    builder.queryParam(secondarySort.getProperty(), RangeOperator.GT + ":" + secondarySortValue);
+                }
+            }
+        }
+    }
+
+    private List<String> addQueryParam(
+            String lastParameterValue, Entry<String, String[]> entry, Direction order, boolean inclusive) {
+        var queryParams = new ArrayList<String>();
```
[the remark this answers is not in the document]
edwin-greene: Thanks I have updated this.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
```diff
@@ -0,0 +1,140 @@
...
+                if (secondarySort != null && !queryParameters.contains(secondarySort.getProperty())) {
+                    var secondarySortValue =
+                            extractor.extract(secondarySort.getProperty()).apply(lastItem);
+                    builder.queryParam(secondarySort.getProperty(), RangeOperator.GT + ":" + secondarySortValue);
+                }
+            }
+        }
+    }
+
+    private List<String> addQueryParam(
```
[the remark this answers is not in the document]
edwin-greene: After some refactoring this method is no longer present.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
```diff
@@ -0,0 +1,140 @@
...
+
+@Named
+@RequiredArgsConstructor
+public class LinkFactoryImpl implements LinkFactory {
+
+    private final HttpServletRequest request;
+
+    public <T> Links create(Iterable<T> items, Pageable pageable, ParameterExtractor<T> extractor) {
+        var links = new Links();
+        if (items == null) {
```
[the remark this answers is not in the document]
edwin-greene: Thanks I have updated this to check for empty.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
edwin-greene: Thanks I have updated this to a List.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
edwin-greene: This method is no longer present after refactoring, but this should be fixed as indicated in the current version.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
edwin-greene: Thanks, I have refactored this section to be in line with this feedback.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
```diff
@@ -0,0 +1,140 @@
...
+import org.springframework.data.domain.Sort.Direction;
+import org.springframework.data.domain.Sort.Order;
+import org.springframework.util.CollectionUtils;
+import org.springframework.web.util.UriComponentsBuilder;
+
+@Named
+@RequiredArgsConstructor
+public class LinkFactoryImpl implements LinkFactory {
+
+    private final HttpServletRequest request;
```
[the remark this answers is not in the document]
edwin-greene: Got it, I have removed that part.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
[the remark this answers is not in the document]
edwin-greene: Thanks I have made this change.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactory.java
edwin-greene: Got it I have made this change.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/controller/AllowancesController.java
```diff
@@ -70,21 +74,41 @@ NftAllowancesResponse getNftAllowancesByAccountId(
...
+                    yield nftAllowance.getTokenId();
+                default:
+                    yield null;
+            };
+        }
+
+        @Override
+        public HttpServletRequest getRequest() {
+            return request;
+        }
```
edwin-greene, line 112: It seems that the extractor has need of the `paramMap` or at least knowledge of the presence of the `owner=true` query parameter. 

This could be one way of providing the `paramMap` to the extractor and also providing it to the `LinkFactoryImpl`, which could be more performant than calling `RequestContextHolder.getRequestAttributes()` in the LinkFactory.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
```diff
@@ -0,0 +1,207 @@
...
+import org.springframework.util.CollectionUtils;
+import org.springframework.web.util.UriComponentsBuilder;
+
+@Named
+@RequiredArgsConstructor
+public class LinkFactoryImpl implements LinkFactory {
+
+    public <T> Links create(List<T> items, Pageable pageable, ParameterExtractor<T> extractor) {
+        var links = new Links();
+        if (items == null || items.isEmpty() || pageable.isUnpaged()) {
```
xin-hedera, line 46:
```suggestion
        if (CollectionUtils.isEmpty(items)) {
```

I don't think we should check `isUnpaged` at all, the contract should be if no next link needed, just don't call `create`

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
```diff
@@ -0,0 +1,207 @@
...
+                    operator = inclusive ? RangeOperator.LTE : RangeOperator.LT;
+                } else {
+                    operator = inclusive ? RangeOperator.GTE : RangeOperator.GT;
+                }
+                builder.queryParam(sort.getProperty(), operator + ":" + lastValue);
+            }
+        });
+    }
+
+    private record RangeBound(RangeOperator operator, String value) implements RangeParameter<String> {
```
xin-hedera, line 173: such logic should reside in the controller

the `isInclusive` flag is used to decide
- if true and order desc, pick `LTE`
- if false and order desc, pick `LT`
- if true and order asc, pick `GTE`
- if false and order asc, pick `GT`

The factory should not further change the value returned by `extractor.extract`

xin-hedera (changes requested):

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
```diff
@@ -0,0 +1,207 @@
...
+    private <T> String createNextLink(T lastItem, int limit, Pageable pageable, ParameterExtractor<T> extractor) {
+        var sortOrders = pageable.getSort();
+        var primarySort = Iterables.getFirst(sortOrders, null);
+        var lastSort = Iterables.getLast(sortOrders, null);
+        var order = primarySort == null ? Direction.ASC : primarySort.getDirection();
+        var request = extractor.getRequest();
+        var builder = UriComponentsBuilder.fromPath(request.getRequestURI());
+        var paramsMap = request.getParameterMap();
+
+        for (var entry : paramsMap.entrySet()) {
```
xin-hedera, line 69: the order of the entries will be random, as a result the order of the `p=v` pairs in the generated next link will be random too.

We may consider sort it, or at least provide an assert utility function in the test cases to ignore the `p=v` order when validating the next url.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/controller/AllowancesController.java
edwin-greene: This has been removed

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
```diff
@@ -0,0 +1,207 @@
...
+    public <T> Links create(List<T> items, Pageable pageable, ParameterExtractor<T> extractor) {
+        var links = new Links();
+        if (items == null || items.isEmpty() || pageable.isUnpaged()) {
+            return links;
+        }
+
+        var limit = pageable.getPageSize();
+        if (limit > items.size()) {
+            return links;
+        }
```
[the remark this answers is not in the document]
edwin-greene: Thanks I have made these updates.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
```diff
@@ -0,0 +1,207 @@
...
+
+        var lastItem = CollectionUtils.lastElement(items);
+        var nextLink = createNextLink(lastItem, limit, pageable, extractor);
+        return links.next(nextLink);
+    }
+
+    private <T> String createNextLink(T lastItem, int limit, Pageable pageable, ParameterExtractor<T> extractor) {
+        var sortOrders = pageable.getSort();
+        var primarySort = Iterables.getFirst(sortOrders, null);
+        var lastSort = Iterables.getLast(sortOrders, null);
```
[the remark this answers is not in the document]
edwin-greene: Thanks, I have updated the main logic of the implementation to be similar to this.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
```diff
@@ -0,0 +1,207 @@
...
+        var order = primarySort == null ? Direction.ASC : primarySort.getDirection();
+        var request = extractor.getRequest();
+        var builder = UriComponentsBuilder.fromPath(request.getRequestURI());
+        var paramsMap = request.getParameterMap();
+
+        for (var entry : paramsMap.entrySet()) {
+            var key = entry.getKey();
+            switch (key) {
+                case LIMIT -> builder.queryParam(LIMIT, limit);
+                case ORDER -> builder.queryParam(ORDER, order.name().toLowerCase());
```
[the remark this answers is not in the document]
edwin-greene: Thanks I have removed this logic.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
edwin-greene: It makes good sense for the extractor alone to be able to determine the value so that nothing else needs to understand the specifics of the value in order to increment or decrement it. 

But it may be that the extractor will need more information to perform the task. Still researching.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
```diff
@@ -0,0 +1,197 @@
...
+            var upperBound = rangeBounds.stream()
+                    .filter(rangeBound -> rangeBound.operator() == RangeOperator.LT)
+                    .reduce((rangeBound1, rangeBound2) ->
+                            rangeBound1.value().compareTo(rangeBound2.value()) > 0 ? rangeBound1 : rangeBound2)
+                    .orElse(RangeBound.EMPTY);
+            addBoundToLink(upperBound, key, builder);
+        }
+
+        // Always add the last value to the link
+        builder.queryParam(key, lastValueOperator + ":" + lastValue);
```
xin-hedera, line 154: I still have a hard time to understand why we need such complicated logic here. Is there any corner case to handle?

For a pagination param `p`, given

- existing `values` from the request params map
- `order` from `pageable`

All we need to do is,

```
for each value in values
  parse and extract operator from value
  if ((order is asc and operator is gt/gte) or (order is desc and operator is lt/lte)) {
    continue; // essentially skip it since the new bound should come from the extracted value
  }

  add param / value pair to builder
```

So this function should only have `builder`, `entry`, and `order` as the arguments.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
```diff
@@ -0,0 +1,197 @@
...
+                addQueryParamToLink(entry, inclusive, builder, lastValue, order, key);
+            }
+        }
+
+        for (var entry : paginationParamsMap.entrySet()) {
+            var key = entry.getKey();
+            if (!paramsMap.containsKey(key)) {
+                boolean inclusive = extractor.isInclusive(key);
+                addPaginationValuesToLink(key, entry.getValue(), inclusive, builder, order);
+            }
```
xin-hedera, line 91: why whether the key exists in `paramsMap` matters?

this part should be just

```
  operator = if order is asc and inclusive then GTE
                     elif order is asc and !inclusive then GT
                     elif order is desc and inclusive then LTE
                     else LT
  add pair (key, operator + ":" +  value) to builder's param map
```

xin-hedera (changes requested):

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactory.java
```diff
@@ -0,0 +1,32 @@
...
+import java.util.Map;
+import org.springframework.data.domain.Pageable;
+
+public interface LinkFactory {
+    <T> Links create(List<T> items, Pageable pageable, ParameterExtractor<T> extractor);
+
+    interface ParameterExtractor<T> {
+        Map<String, String> extract(T item);
+
+        boolean isInclusive(String param);
```
edwin-greene, line 30: A suggestion Steven had was that we could remove `isInclusive` here. Instead of having it as a method in the interface, we could use the `Sort` inside `Pageable` to determine inclusivity. 

The `Sort` contains a list of sorts, and all but the last sort can be used to indicate `inclusive = true`, while the last sort indicates `inclusive = false`.

xin-hedera: That’s feasible however it should be the last is not inclusive and any non-last is inclusive. Special case is if only one, it should be non inclusive.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
edwin-greene: Got it, I have simplified this logic

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
edwin-greene: Thanks I have updated this.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/controller/AllowancesControllerTest.java
```diff
@@ -242,24 +297,28 @@ void nftAllowancesOrderDesc() {
...
                 .get()
                 .uri(uriParams, allowance1.getOwner())
                 .retrieve()
                 .body(NftAllowancesResponse.class);
+        var next =
+                "/api/v1/accounts/%s/allowances/nfts?account.id=gt:0.0.1000&account.id=lte:%s&owner=true&token.id=gt:0.0.1001&token.id=lt:%s&limit=2&order=desc"
+                        .formatted(
+                                allowance2.getOwner(),
+                                EntityId.of(allowance2.getSpender()),
+                                EntityId.of(allowance2.getTokenId()));
```
edwin-greene, line 317: Note: This part of the test verifies that the other issue [#8185](https://github.com/hashgraph/hedera-mirror-node/issues/8185) is fixed by this PR.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/RangeBound.java
```diff
@@ -0,0 +1,38 @@
...
+
+        value = value.toLowerCase();
+        var splitVal = value.split(":");
+        return switch (splitVal.length) {
+            case 1 -> new RangeBound(RangeOperator.EQ, value);
+            case 2 -> new RangeBound(RangeOperator.of(splitVal[0]), splitVal[1]);
+            default -> throw new IllegalArgumentException(
+                    "Invalid range %s. Should have format operator:value".formatted(value));
+        };
+    }
```
xin-hedera, line 37: we should be able to move the common logic to a factory, don't have to make the change in this PR

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/common/LinkFactoryTest.java
```diff
@@ -0,0 +1,222 @@
...
+                        .create(List.of(nftAllowance), pageable, extractor)
+                        .getNext())
+                .isEqualTo(expectedLink);
+    }
+
+    @DisplayName("Get pagination links unknown parameter")
+    @Test
+    void testUnknownParameter() {
+        Map<String, String[]> params = new LinkedHashMap<>();
+        params.put("limit", new String[] {"1"});
```
xin-hedera, line 153: remove `limit` here so we have one test case for when there's no `limit` query param in the request URL

```suggestion
```

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactory.java
steven-sheehy: At this point, best for a follow up.

steven-sheehy (dismissed):
LGTM

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/RangeBound.java
xin-hedera: still need to address this

steven-sheehy (approved):

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/RangeBound.java
xin-hedera: changing to factory may affect many other places. however still removed `RangeBound` and just switched to `String.startsWith`

xin-hedera (approved):
LGTM

steven-sheehy (approved):

PR #8267, opened 2024-05-10
Clear incorrect token metadata / metadata key and fix ingestion logic

### LINKED ISSUE #8263: Mirrornode may ingest and persist incorrect token metadata and metadata key
### Description

Token metadata and metadata key fields are added to protobuf 0.48.x, however the feature in services is only enabled starting from release 0.49.0.

Mirrornode generally implements and deploys features ahead of services release. However, in the case of token metadata & metadata key, services in 0.48.x would simply process token create & token update transactions as if those fields don't exist. This makes mirrornode to persist metadata and metadata key info into database while none would go into the ledger state, if ever such transactions were processed.

### Steps to reproduce

N/A

### Additional context

_No response_

### Hedera network

other

### Version

v0.101.0

### Operating system

None

xin-hedera:
**Description**:

This PR fixes the incorrect token metadata / metadata key, and the ingestion logic

- Only process token metadata / metadata key in `TokenCreate` and `TokenUpdate` transactions when HAPI version >= 0.49.0
- Add a db migration to clear incorrect token metadata / metadata key

**Related issue(s)**:

Fixes #8263 

**Notes for reviewer**:

The time consuming part of the migration is to query `record_file` table with filter on HAPI version. In total, the migration took about 2 minutes in mainnet, and about 4.5 minutes in testnet.

Unfortunately, we have to get it into release 104 and deploy it before mainnet 0.49.0 upgrade, otherwise there can still be incorrect metadata / metadata key in our db.

It's due to that we don't have the data in db to tell us what fields are updated by a token update, and what are not changed thus coalesced from the previous token state. Note, it's perfectly fine for a token update transaction to have the same metadata value set in the tx body.

**Checklist**

### DIFF
--- a/hedera-mirror-common/src/test/java/com/hedera/mirror/common/domain/DomainBuilder.java
+++ b/hedera-mirror-common/src/test/java/com/hedera/mirror/common/domain/DomainBuilder.java
@@ -1106,11 +1106,11 @@ public class DomainBuilder {
      * @param value The timestamp to reset to
      */
     public void resetTimestamp(long value) {
-        timestampOffset = value - timestamp();
+        timestampOffset = value - timestampNoOffset();
     }
 
     public long timestamp() {
-        return DomainUtils.convertToNanosMax(now.getEpochSecond(), now.getNano()) + number() + timestampOffset;
+        return timestampNoOffset() + timestampOffset;
     }
 
     private long tinybar() {
@@ -1122,6 +1122,10 @@ public class DomainBuilder {
                 .atStartOfDay()
                 .toLocalDate()
                 .toEpochDay();
+    }
+
+    private long timestampNoOffset() {
+        return DomainUtils.convertToNanosMax(now.getEpochSecond(), now.getNano()) + number();
     }
 
     private int transactionIndex() {
--- a/hedera-mirror-common/src/test/java/com/hedera/mirror/common/domain/DomainBuilder.java
+++ b/hedera-mirror-common/src/test/java/com/hedera/mirror/common/domain/DomainBuilder.java
@@ -145,6 +145,8 @@ public class DomainBuilder {
     private final Instant now = Instant.now();
     private final SecureRandom random = new SecureRandom();
 
+    private long timestampOffset = 0;
+
     // Intended for use by unit tests that don't need persistence
     public DomainBuilder() {
         this(null, null);
@@ -1099,8 +1101,16 @@ public class DomainBuilder {
         return RandomStringUtils.random(characters, "0123456789abcdef");
     }
 
+    /**
+     * Reset the timestamp, so next call of timestamp() will return value + 1
+     * @param value The timestamp to reset to
+     */
+    public void resetTimestamp(long value) {
+        timestampOffset = value - timestamp();
+    }
+
     public long timestamp() {
-        return DomainUtils.convertToNanosMax(now.getEpochSecond(), now.getNano()) + number();
+        return DomainUtils.convertToNanosMax(now.getEpochSecond(), now.getNano()) + number() + timestampOffset;
     }
 
     private long tinybar() {
--- a/hedera-mirror-importer/src/main/resources/db/migration/v1/V1.97.2__clear_token_metadata.sql
+++ b/hedera-mirror-importer/src/main/resources/db/migration/v1/V1.97.2__clear_token_metadata.sql
@@ -1,47 +1,10 @@
 -- clear token metadata and metadata_key accidentally set prior to services 0.49.0
-with last_hapi_version as (
-  select coalesce((
-    select (hapi_version_major * 2^32)::bigint + hapi_version_minor
+with timestamp_info as (
+    -- make it closed-open, if nothing found, it's int8range(0, 0) which is empty
+    -- 1709229598938101716 is the last 0.46.x transaction's timestamp in testnet, mainnet's is at a later time
+    select int8range(coalesce(min(consensus_start), 0), coalesce(max(consensus_end) + 1, 0)) as affected_range
     from record_file
-    order by consensus_end desc
-    limit 1
-  ), 0::bigint) as version
-), lower_bound as (
-  -- The exclusive lower bound when last version >= 0.47. It's not 0.48 because services release 0.48 reports HAPI
-  -- version 0.47 in record files. The value is either the last consensus timestamp of services release 0.46 or 1ns
-  -- before the first consensus timestamp in case the min version in record_file table is > 0.46
-  select coalesce((
-    select consensus_end
-    from record_file
-    where hapi_version_major = 0 and hapi_version_minor = 46
-    order by consensus_end desc
-    limit 1), (
-    select consensus_start - 1
-    from record_file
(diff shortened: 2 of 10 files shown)


hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/migration/ClearTokenMetadataMigrationTest.java
```diff
@@ -0,0 +1,283 @@
...
+        // token1 updated
+        createOrUpdateToken(t -> {}, token1, ++timestamp);
+
+        // corner case, token4 created at the last consensus timestamp of 0.47, with metadata / metadata key
+        timestamp = recordFiles.get(3).getConsensusEnd();
+        createOrUpdateToken(t -> {}, CREATE, timestamp);
+
+        // token5 created in 0.49, without metadata / metadata key
+        timestamp = recordFiles.get(4).getConsensusStart();
+        long token5 = createOrUpdateToken(t -> t.metadata(null).metadataKey(null), CREATE, timestamp);
```
jascks, line 133: On token creation, by default the metadata is an empty byte array, not null.

In the `token` table, the only way for that column to be null is for legacy tokens present prior to the application of `V1.95.0__add_token_metadata_and_key.sql`, which added the `metadata` and `metadata_key` columns with default null values.

For tokens created since via the 0.48 protobufs, the column value will be an empty byte array if metadata is not supplied by the HAPI client at the time.

xin-hedera: will change it to empty byte array

jascks (dismissed):
LGTM

jnels124 (commented):
Do we need a v2 migration? I guess this really only effects preview net which can be reset as we wish. I can hold off on starting the staging migration.

xin-hedera:
> Do we need a v2 migration? I guess this really only effects previewnet citus which can be reset as we wish. I can hold off on starting the staging migration until this has been applied

I think it's a no. doesn't really matter much for staging, the worst case is we can just manually fix the data in staging after the migration, or just leave it as is.

jnels124 (dismissed):

edwin-greene (dismissed):
Looks good

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.97.2__clear_token_metadata.sql
```diff
@@ -0,0 +1,59 @@
...
+    select consensus_end + 1
+    from record_file
+    where hapi_version_major = 0 and hapi_version_minor = 47
+    order by consensus_end desc
+    limit 1), (
+    select consensus_start
+    from record_file
+    order by consensus_end
+    limit 1
+  )) as timestamp
```
steven-sheehy, line 39: This is all quite complicated and prone to corner cases. This can be simplified if you just do a single query with some fixed lower bound to improve performance. For the fixed lower bound can take the the min 0.47 consensus_start from testnet and mainnet. This query takes 2s to run on mainnet:

```sql
select min(consensus_start) as min_consensus_start, max(consensus_end) as max_consensus_end
from record_file
where hapi_version_major = 0 and hapi_version_minor in (47, 48) and consensus_end >= 1709229731591879003;
```

steven-sheehy (changes requested):

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/domain/RecordItemBuilder.java
```diff
@@ -779,7 +780,8 @@ public Builder<TokenUpdateTransactionBody.Builder> tokenUpdate() {
                 .setToken(tokenId())
                 .setTreasury(accountId())
                 .setWipeKey(key());
-        return new Builder<>(TransactionType.TOKENUPDATE, transactionBody);
+        return new Builder<>(TransactionType.TOKENUPDATE, transactionBody)
+                .recordItem(r -> r.hapiVersion(HAPI_VERSION_0_49_0));
```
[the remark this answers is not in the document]
xin-hedera: better to do it in a follow up, since change the default may have cascading effect on other test cases. Just don't want to increase the scope especially it's just test cases

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.97.2__clear_token_metadata.sql
xin-hedera: agree on the complexity. However performance wise, it depends.

When data is hot, for mainnet, the simple query actually took longer, around 1067.534 ms, v.s. the complicated logic, which took 866.005 ms.

For testnet, the simpley query won with 618ms v.s. 1280ms.

Also, to make it work for all scenarios, for example, when hapi version major/minor/patch are all null prior to those fields were introduced in perhaps format v5, or there is simply no 0.47.0 record file in db, similar logic is still needed to create an `empty` timestamp range.

steven-sheehy: > agree on the complexity. However performance wise, it depends.

If they're about the same (difference measured in milliseconds), then we should prefer the simpler one.

> Also, to make it work for all scenarios, for example, when hapi version major/minor/patch are all null prior to those fields were introduced in perhaps format v5, or there is simply no 0.47.0 record file in db, similar logic is still needed to create an empty timestamp range.

A separate CTE could coalesce to empty range if needed.

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/domain/RecordItemBuilder.java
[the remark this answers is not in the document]
steven-sheehy: Sure

steven-sheehy (dismissed):
LGTM

hedera-mirror-common/src/test/java/com/hedera/mirror/common/domain/DomainBuilder.java
```diff
@@ -1099,8 +1101,16 @@ public String hash(int characters) {
...
+     * Reset the timestamp, so next call of timestamp() will return value + 1
+     * @param value The timestamp to reset to
+     */
+    public void resetTimestamp(long value) {
+        timestampOffset = value - timestamp();
+    }
+
     public long timestamp() {
-        return DomainUtils.convertToNanosMax(now.getEpochSecond(), now.getNano()) + number();
+        return DomainUtils.convertToNanosMax(now.getEpochSecond(), now.getNano()) + number() + timestampOffset;
```
xin-hedera, line 1113: I need to fix this, the math is wrong, when reset, the `timestamp()` call depends on the previous `timestampOffset`...

jascks (approved):
LGTM

steven-sheehy (approved):

jascks (approved):
LGTM

jnels124 (approved):
LGTM

PR #8317, opened 2024-05-15
Prevent ZFS components from restarting on every deploy

### LINKED ISSUE #8279: Prevent ZFS Components From Starting On Every Deploy
### Problem

Currently anytime the chart is updated, ZFS init pods restart even though they had no changes. The pod should only redeploy when there are changes.

### Solution

Determine why the zfs init pod restarts on every chart upgrade and fix 

### Alternatives

_No response_

jnels124:
**Description**:

* Exit ZFS init script on any error
* Update metadata labels on ZFS init daemonset to match selector labels

**Related issue(s)**:

Fixes #8279

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/charts/hedera-mirror-common/templates/zfs/configmap-init.yaml
+++ b/charts/hedera-mirror-common/templates/zfs/configmap-init.yaml
@@ -2,7 +2,7 @@
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  labels: {{ include "hedera-mirror-common.labels" . | nindent 4 }}
+  labels: {{ include "hedera-mirror-common.selectorLabels" . | nindent 4 }}
     app: zfs-init
   name: {{ .Release.Name }}-zfs-init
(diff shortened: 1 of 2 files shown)


steven-sheehy (approved):
LGTM

charts/hedera-mirror-common/templates/zfs/configmap-init.yaml
```diff
@@ -2,7 +2,7 @@
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  labels: {{ include "hedera-mirror-common.labels" . | nindent 4 }}
+  labels: {{ include "hedera-mirror-common.selectorLabels" . | nindent 4 }}
```
jnels124, line 5: This configmap is hashed and added as an annotation to init pods

xin-hedera (approved):
LGTM

PR #8352, opened 2024-05-23
Check support of negative one value for maxAutomaticTokenAssociations

### LINKED ISSUE #8273: HIP-904 Ensure Importer support for unlimited Max Automatic Token Associations
### Problem

Ensure support for unlimited Max Automatic Token Associations. Importer should already support them.

### Solution

Add test(s) for the `-1` value for `maxAutomaticTokenAssociations` in Importer.
Protobuf version 0.51 is the first version that will use -1 to indicate unlimited max automatic token associations. 

### Alternatives

_No response_

mgoelswirlds:
**Description**:
This PR checks for the support of `-1` for maxAutomaticTokenAssociations to support frictionless airdrops in the future.

This PR modifies 
* Adds tests to verify `-1` value works for maxAutomaticTokenAssociations.

**Related issue(s)**:

Fixes #8273

**Notes for reviewer**:
I looks like there aren't any special operations done with maxAutomaticTokenAssociations, which would be affected because of a -1 value.

**Checklist**

### DIFF
--- a/hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/record/entity/EntityRecordItemListenerContractTest.java
+++ b/hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/record/entity/EntityRecordItemListenerContractTest.java
@@ -138,6 +138,7 @@ class EntityRecordItemListenerContractTest extends AbstractEntityRecordItemListe
                     }
                     b.clearAutoRenewAccountId()
                             .setDeclineReward(true)
+                            .setMaxAutomaticTokenAssociations(-1)
                             .setStakedAccountId(AccountID.newBuilder()
                                     .setAccountNum(domainBuilder.id())
                                     .build());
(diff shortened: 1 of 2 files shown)


hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/record/entity/EntityRecordItemListenerContractTest.java
```diff
@@ -138,6 +138,7 @@ void contractCreate(boolean bytecodeSourceFileId, boolean hasAutoRenewAccount) {
                     }
                     b.clearAutoRenewAccountId()
                             .setDeclineReward(true)
+                            .setMaxAutomaticTokenAssociations(-1)
```
steven-sheehy, line 141: Still trying to verify if -1 even applies to contract create. For now, we should at least add a similar test case for crypto create and update

steven-sheehy: Looks like it can apply to contracts with a key, which is missing from the HIP/protos. So can keep this code but add one for account.

mgoelswirlds: Sounds good.

PR #8378, opened 2024-05-27
test: add abstract ControllerTest in hedera-mirror-node.web3

victor-yanev:
**Description**:
This PR adds an abstract `ControllerTest` class, which contains:
- common bean configurations
- common boilerplate code for the controller tests
- common tests for each controller:
  - `cors()`
  - `unsupportedMediaType()`
  - `malformedJsonBody()`
  - `exceedingRateLimit()`
  - `exceedingGasLimit()`
- common helper methods which are used in the controller tests:
  - `responseBody()`
  - `convertToJson()`
  - `numberErrorString()`

**Checklist**

### DIFF
--- a/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/controller/ControllerTest.java
+++ b/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/controller/ControllerTest.java
@@ -1,0 +1,211 @@
+/*
+ * Copyright (C) 2024 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hedera.mirror.web3.controller;
+
+import static org.mockito.ArgumentMatchers.anyLong;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.options;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.content;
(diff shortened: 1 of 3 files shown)


hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/controller/OpcodesControllerTest.java
```diff
@@ -142,303 +109,238 @@ class OpcodesControllerTest {
...
-        mockMvc.perform(opcodesRequest(transactionIdOrHash))
-                .andExpect(status().isNotFound())
-                .andExpect(responseBody(new GenericErrorResponse("Record file with transaction not found")));
-    }
+        @Override
+        protected MockHttpServletRequestBuilder customizeRequest(MockHttpServletRequestBuilder requestBuilder) {
+            return requestBuilder
+                    .queryParam("stack", String.valueOf(opcodeTracerOptions.isStack()))
+                    .queryParam("memory", String.valueOf(opcodeTracerOptions.isMemory()))
+                    .queryParam("storage", String.valueOf(opcodeTracerOptions.isStorage()));
```
steven-sheehy, line 150: Don't think we should map from this opcodeTracerOptions to the query params. We should just pass a URL or URL suffix like `"contracts/results/abcdef/opcodes?stack=true&memory=false&storage=true"`. This gives us the flexibility to test different combinations like missing one or all query parameters and using the default or setting stack=invalid_value for negative testing, etc. The above approach is too limiting and abstracted away from REST API testing which should involve URLs not POJOs.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/controller/ControllerTest.java
```diff
@@ -0,0 +1,261 @@
...
+            return performRequest(null);
+        }
+
+        /**
+         * @param requestBody the request body (if any) to include in the request
+         * @return the {@link ResultActions} for the request
+         */
+        @SneakyThrows
+        protected ResultActions performRequest(@Nullable Object requestBody) {
+            return mockMvc.perform(buildRequest(requestBody));
```
steven-sheehy, line 128: I'm not a fan of abstracting away how it calls a REST API in a controller test. I like to see the path parameters, query parameters and method all listed in the individual tests so it's self documenting. These helper methods are not as flexible as the Spring APIs so they make it harder in the future for custom behavior. So think we should remove these performRequest, customizeRequest and getParameters, buildRequest methods in favor of directly using spring.

Having a default request for testing cors, etc in EndpointTest does make sense.

steven-sheehy (changes requested):

PR #8382, opened 2024-05-27
Refactor `ContractCallNativePrecompileTest` in web3

### LINKED ISSUE #8380: Refactor `ContractCallNativePrecompileTest`
As the description of the WP suggests, refactor the mentioned integration tests in web3 module (`ContractCallNativePrecompileTest`, `ContractCallNestedCallsTest` and `ContractCallServiceERCTokenTest`) in a way to be self-sufficient enough and rely only on the provided data in this specific test. Essentially, if some of the data in separate tests inside the same integration test file is repeated, it could abstracted away in the integration test class, if appropriate

### LINKED ISSUE #8454: Refactor `ContractCallNativePrecompileTest`

petreze:
**Description**:
This PR aims to refactor `ContractCallNativePrecompileTest` integration test files by removing redundancies from `ContractCallTestSetup` and include only needed DB setup data and entities for the specific test cases in those files

**Related issue(s)**:

Fixes #8454 

**Notes for reviewer**:
We have some prerequisite logic in `ContractCallTestSetup` that might be useful to keep there as all tests might need it: `exchangeRatesPersist()` and `feeSchedulesPersist()`

**Checklist**


fchitakova (dismissed):

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServiceERCTokenTest.java
```diff
@@ -16,29 +16,112 @@
...
+            ByteString.fromHex("3a2102ff806fecbd31b4c377293cba8d2b78725965a4990e0ff1b1b29a1d2c61402310");
+    private static final Address SPENDER_ALIAS = Address.wrap(
+            Bytes.wrap(recoverAddressFromPubKey(SPENDER_PUBLIC_KEY.substring(2).toByteArray())));
+    private static final ByteString SPENDER_PUBLIC_KEY_HISTORICAL =
+            ByteString.fromHex("3a210398e17bcbd2926c4d8a31e32616b4754ac0a2fc71d7fb768e657db46202625f34");
+    private static final Address SPENDER_ALIAS_HISTORICAL = Address.wrap(Bytes.wrap(
+            recoverAddressFromPubKey(SPENDER_PUBLIC_KEY_HISTORICAL.substring(2).toByteArray())));
+    private static final Address OWNER_ADDRESS = toAddress(EntityId.of(0, 0, 1044));
+    private static final Address OWNER_ADDRESS_HISTORICAL = toAddress(EntityId.of(0, 0, 1065));
+    private static final Address HOLLOW_ACCOUNT_ALIAS = Address.wrap(Bytes.wrap(recoverAddressFromPubKey(
```
steven-sheehy, line 81: There's two aspects that make the tests on `main` hard to understand and maintain:

* Shared setup across test classes
* Shared setup across test methods within a class

Your changes help with the first but not the second. In general, we try to avoid static variables with identifiers that are used across test methods for the same reason we avoid it across test classes. We prefer to create our test input dynamically using `DomainBuilder`. If the defaults in `DomainBuilder` are not sufficient we can create helper reusable methods that use `DomainBuilder` to create them and link different entities together, but use dynamically generated IDs instead of static ones.

So here I think we should remove all these static identifiers and figure out a way to craft the data dynamically but in a reusable manner in case we want to share it with other test classes.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallNativePrecompileTest.java
```diff
@@ -221,6 +250,82 @@ void directCallToNativePrecompileBlake2f() {
...
+                .callData(callData)
+                .gas(15_000_000L)
+                .isStatic(false)
+                .callType(callType)
+                .isEstimate(ETH_ESTIMATE_GAS == callType)
+                .block(block)
+                .build();
+    }
+
+    private void feeSchedulesPersist() {
```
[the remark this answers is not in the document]
petreze: Agreed, this is the approach that is implemented in `ContractCallServiceERCTokenTest` as we got into the same conclusion

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServiceERCTokenTest.java
petreze: @steven-sheehy 
I entirely agree on both points and the end goal but how do you imagine the dynamic generation of addresses lets say. In every case, my understanding is that we would have to pass some num field to generate an `Address` or a specific string to generate a valid public key as `ByteString`. If we generate those entirely randomly, we could end up with some cases like:
- generating entities with the same `num` field which could end up messing some linkages between entities (accounts - token relations or something else)
- or generate `invalid` strings as public keys
I assume these would be a problem or I misunderstood something..?

steven-sheehy: > generating entities with the same num field which could end up messing some linkages between entities (accounts - token relations or something else)

That's why I said `create helper reusable methods that use DomainBuilder to create them and link different entities together`. Point is to minimize the hardcoding and customization of fields generated by DomainBuilder to what's truly necessary. Something like the below which can be used inline or wrapped in a helper function:

```java
    var account = domainBuilder.entity().customize(e -> e.type(ACCOUNT)).persist();
    var token = domainBuilder.entity().customize(e -> e.type(TOKEN)).persist();
    var tokenAccount = domainBuilder.tokenAccount().customize(ta -> ta.accountId(account.getId().tokenId(token.getId()).persist();
```

xin-hedera (approved):
LGTM

PR #8384, opened 2024-05-28
Token creation with `INHERIT` keys via smart contract throws `CONTRACT_REVERT_EXECUTED`

### LINKED ISSUE #8366: Token creation with `INHERIT` keys via smart contract throws `CONTRACT_REVERT_EXECUTED`
### Description

Executing [this contract call](https://github.com/hashgraph/hedera-smart-contracts/blob/main/contracts/hts-precompile/examples/token-create/TokenCreateContract.sol#L24) against services resolves to success but executing it in the mirror-node-web3 module (e.g. _eth_call_ or _eth_estimateGas_ from the relay) resolves to `CONTRACT_REVERT_EXECUTED`.

### Steps to reproduce

After some debugging, I've found out that [here](https://github.com/hashgraph/hedera-mirror-node/blob/main/hedera-mirror-web3/src/main/java/com/hedera/services/store/contracts/precompile/impl/TokenCreatePrecompile.java#L773) `account.key` is **null** which resolves to `INVALID_<TYPE>_KEY` afterward. Another concerning place could be the JKey resolution when the account has been fetched from the store.

### Additional context

_No response_

### Hedera network

other

### Version

0.104.0

### Operating system

Linux

natanasow:
**Description**:

Executing [this contract call](https://github.com/hashgraph/hedera-smart-contracts/blob/main/contracts/hts-precompile/examples/token-create/TokenCreateContract.sol#L24) against services resolves to success but executing it in the mirror-node-web3 module (e.g. _eth_call_ or _eth_estimateGas_ from the relay) resolves to `CONTRACT_REVERT_EXECUTED`.

**Related issue(s)**:

Fixes #8366 

**Steps to reproduce**:

After some debugging, I've found out that [here](https://github.com/hashgraph/hedera-mirror-node/blob/main/hedera-mirror-web3/src/main/java/com/hedera/services/store/contracts/precompile/impl/TokenCreatePrecompile.java#L773) `account.key` is null which resolves to `INVALID_<TYPE>_KEY` afterward. Another concerning place could be the JKey resolution when the account has been fetched from the store.

**Notes for reviewer**:

**Checklist**


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/store/accessor/AccountDatabaseAccessor.java
```diff
@@ -213,9 +215,18 @@ private Supplier<TokenAccountBalances> getNumberOfAllAndPositiveBalanceTokenAsso
...
         try {
-            return keyBytes == null ? null : asFcKeyUnchecked(Key.parseFrom(keyBytes));
+            if (keyBytes == null) {
+                if (isContract) {
+                    return new JContractIDKey(
+                            ContractID.newBuilder().setContractNum(entityId).build());
+                }
+                return null;
+            } else {
+                return asFcKeyUnchecked(Key.parseFrom(keyBytes));
```
steven-sheehy, line 227: It's incorrect to map `entity.getId()` to just the number of `ContractID` since id contains shard and realm encoded in it as well. Instead, pass entire `Entity` and create `ContractID` from separate realm, shard and num fields.

```suggestion
            var keyBytes = entity.getKey();
            if (keyBytes == null) {
                if (entity.getType() == EntityType.CONTRACT) {
                    return new JContractIDKey(                            ContractID.newBuilder()
                        .setContractNum(entity.getNum())
                        .setRealmNum(entity.getRealm())
                        .setShardNum(entity.getShard())
                        .build());
                }
                return null;
            } else {
                return asFcKeyUnchecked(Key.parseFrom(keyBytes));
```

steven-sheehy, line 227: Invert this logic to reduce nesting so that it's `if (keyBytes != null) {return asFcKeyUnchecked(Key.parseFrom(keyBytes));}

steven-sheehy (changes requested):

steven-sheehy (approved):
LGTM

PR #8388, opened 2024-05-28
Log additional details for query timeout exception

### LINKED ISSUE #8364: Log SQL Timeout Exceptions for Web3
### Problem

Some web3 calls result in a sql timeout. It is unclear what data is triggering this to occur. 

### Solution

We should log all sql timeout exceptions  and include the entire request object in logs if data field is not to big. 

### Alternatives

_No response_

jnels124:
**Description**:
* log additional details when queury timeout exception occurs 

**Related issue(s)**:

Fixes #8364

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/ContractController.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/ContractController.java
@@ -87,7 +87,7 @@ class ContractController {
             params = constructServiceParameters(request);
             result = contractCallService.processCall(params);
         } catch (QueryTimeoutException e) {
-            log.error("Query timed out: {}", e.getMessage() + " request: " + request);
+            log.error("Query timed out: {} request: {}", e.getMessage(), request);
             throw e;
         } catch (InvalidParametersException e) {
             // The validation failed but no processing was made - restore the consumed gas back to the bucket.
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/ContractController.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/ContractController.java
@@ -79,12 +79,11 @@ class ContractController {
         }
 
         String result;
-        CallServiceParameters params = null;
         try {
             validateContractData(request);
             validateContractMaxGasLimit(request);
 
-            params = constructServiceParameters(request);
+            final var params = constructServiceParameters(request);
             result = contractCallService.processCall(params);
         } catch (QueryTimeoutException e) {
             log.error("Query timed out: {} request: {}", e.getMessage(), request);
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/ContractController.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/ContractController.java
@@ -87,28 +87,8 @@ class ContractController {
             params = constructServiceParameters(request);
             result = contractCallService.processCall(params);
         } catch (QueryTimeoutException e) {
-            final var paramsString = new StringBuilder();
-
-            if (params != null) {
-                if (params.getSender() != null) {
-                    paramsString.append(" sender: ").append(params.getSender().canonicalAddress());
-                }
-                paramsString
-                        .append(" receiver: ")
-                        .append(params.getReceiver())
-                        .append(" gas: ")
-                        .append(params.getGas())
-                        .append(" value: ")
-                        .append(params.getValue());
-
-                if (params.getCallData() != null
-                        && params.getCallData().size()
-                                < ((evmProperties.getMaxDataSize().toBytes()) * .75)) {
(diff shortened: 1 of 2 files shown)


jascks (dismissed):
LGTM

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/ContractController.java
```diff
@@ -79,12 +79,36 @@ ContractCallResponse call(@RequestBody @Valid ContractCallRequest request) {
...
+                        .append(params.getGas())
+                        .append(" value: ")
+                        .append(params.getValue());
+
+                if (params.getCallData() != null
+                        && params.getCallData().size()
+                                < ((evmProperties.getMaxDataSize().toBytes()) * .75)) {
+                    paramsString.append(" data: ").append(params.getCallData().toHexString());
+                }
+            }
```
steven-sheehy, line 109: This is verbose and brittle. We should just log the `ContractCallRequest` directly. I don't think we need to truncate the call data since query time outs should be infrequent.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/ContractController.java
```diff
@@ -79,12 +79,36 @@ ContractCallResponse call(@RequestBody @Valid ContractCallRequest request) {
...
+
+            throw new QueryTimeoutException(e.getMessage() + " Params: " + paramsString);
```
steven-sheehy, line 111: Seems like we're creating this wrapper exception just so we can capture the request and use it to log in `ContractController.queryTimeout()`. Instead, we should just remove that log in the exception handler and log here but still throw.

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/ContractController.java
jnels124: I agree. When discussed in standup this was requested. I can remove the data size limitation.

steven-sheehy: Yeah, I waffled on the truncating. But if we were to truncate we would just `request.setData(truncate(request.getData())` then use its `toString()`.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/ContractController.java
jnels124: I can do that. Was hesitant to remove the log from the exception handler in case future code changes introduce a timeout exception outside of this block.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/ContractController.java
```diff
@@ -79,12 +79,16 @@ ContractCallResponse call(@RequestBody @Valid ContractCallRequest request) {
...
+        CallServiceParameters params = null;
         try {
             validateContractData(request);
             validateContractMaxGasLimit(request);
 
-            final var params = constructServiceParameters(request);
+            params = constructServiceParameters(request);
             result = contractCallService.processCall(params);
+        } catch (QueryTimeoutException e) {
+            log.error("Query timed out: {}", e.getMessage() + " request: " + request);
```
steven-sheehy, line 90:
```suggestion
            log.error("Query timed out: {}, request: {}", e.getMessage(), request);
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/ContractController.java
```diff
@@ -79,12 +79,16 @@ ContractCallResponse call(@RequestBody @Valid ContractCallRequest request) {
         }
 
         String result;
+        CallServiceParameters params = null;
```
steven-sheehy, line 82: Revert

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/ContractController.java
steven-sheehy: If we want to capture the request object in the handler we could use `webRequest.setAttribute("request", request, RequestAttributes.SCOPE_REQUEST);` in the controller to set it and retrieve in handler.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/ContractController.java
jnels124: ya caught that. thanks.

steven-sheehy (approved):
LGTM

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/ContractController.java
jnels124: caught that too you were too quick

PR #8424, opened 2024-05-31
Use web3j in acceptance tests

### LINKED ISSUE #8392: Use web3j in acceptance tests
### Problem

In https://github.com/hashgraph/hedera-mirror-node/pull/8123, consensus code node was copied to encode and sign Ethereum transactions. It would be preferable to use a library to perform such operations like web3j.

### Solution

* Add web3j library
* Change Ethereum transaction generation and signing to use web3j
* Remove copied `EthTxData` and `EthTxSigs` classes

### Alternatives

_No response_

JivkoKelchev:
**Description**:

* Add web3j library
* Change Ethereum transaction generation and signing to use web3j
* Remove copied EthTxData and EthTxSigs classes

**Related issue(s)**:

Fixes #8392

**Notes for reviewer**:

Not sure if web3j will be used anywhere else, so I added the dependency in `hedera-mirror-test` as `testImplementation`

**Checklist**


hedera-mirror-test/build.gradle.kts
```diff
@@ -58,6 +58,7 @@ dependencies {
     testImplementation("org.springframework.retry:spring-retry")
     testImplementation("org.apache.tuweni:tuweni-bytes")
     testImplementation("commons-codec:commons-codec")
+    testImplementation("org.web3j:core:4.12.0")
```
steven-sheehy, line 61: We don't put versions in child scripts, only in dependencies.constraints section in the root build.gradle.kts

JivkoKelchev: Done

steven-sheehy (approved):

PR #8425, opened 2024-05-31
Implement integration tests covering `OpcodeService`

victor-yanev:
**Description**:

This PR adds integration tests for `OpcodeService`.

- processOpcodeCall
  - callWithNestedEthCalls
  - callWithDifferentCombinationsOfTracerOptions
  - callWithTransactionNotFoundExceptionTest
  - callWithContractResultNotFoundExceptionTest
  - callWithContractTransactionHashNotFoundExceptionTest

**Checklist**


hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/OpcodeServiceTest.java
```diff
@@ -0,0 +1,415 @@
...
+import org.junit.jupiter.api.TestInstance;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.EnumSource;
+import org.junit.jupiter.params.provider.MethodSource;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Captor;
+import org.springframework.beans.factory.annotation.Autowired;
+
+class OpcodeServiceTest extends ContractCallTestSetup {
```
xin-hedera, line 68:
```suggestion
@RequiredArgsConstructor
class OpcodeServiceTest extends ContractCallTestSetup {
```

and

- remove `@Autowired` instance properties
- make them `final`

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/OpcodeServiceTest.java
```diff
@@ -0,0 +1,415 @@
...
+
+        @Captor
+        private ArgumentCaptor<ContractDebugParameters> serviceParametersCaptor;
+
+        @Captor
+        private ArgumentCaptor<OpcodeTracerOptions> tracerOptionsCaptor;
+
+        private final ResultCaptor<OpcodesProcessingResult> opcodesResultCaptor = new ResultCaptor<>(OpcodesProcessingResult.class);
+
+        private final AtomicReference<ContractDebugParameters> expectedServiceParameters = new AtomicReference<>();
```
xin-hedera, line 92: no need to use atomic reference

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/OpcodeServiceTest.java
```diff
@@ -0,0 +1,415 @@
...
+            exchangeRatesPersist();
+            feeSchedulesPersist();
+        }
+
+        @BeforeEach
+        void setUpArgumentCaptors() {
+            expectedServiceParameters.set(null);
+            doAnswer(opcodesResultCaptor)
+                    .when(contractDebugService)
+                    .processOpcodeCall(serviceParametersCaptor.capture(), tracerOptionsCaptor.capture());
```
xin-hedera, line 133: I'm not sure why spying the contractDebugService (including capturing its input and output) when doing integration testing of opCodeService. We should treat it as a black box.

The obvious issue aside from the methodology is bugs in contractDebugService may have a chance to escape, since instead of verifying the output from outside of opCodeService without even knowing it depends on contractDebugService, the test cases here directly verify the response is the same as captured from contractDebugService.

xin-hedera (changes requested):

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/OpcodeServiceTest.java
victor-yanev: First, we have to keep in mind that those tests are heavy to perform, especially when we are passing `stack=true, memory=true, storage=true`. Thats why we are doing this “capturing” to avoid executing the same contract call twice which could take quite a lot of time if we have to be also capturing the memory and stack contents after the execution in the EVM.

We have another test file with integration tests for `ContractDebugService` which tests things on a lower level and captures the result from the EVM processor, then compares that against the `OpcodesProcessingResult` that is returned from the call. We can also do this here, but we would essentially be duplicating the tests from the `ContractDebugServiceTest` file. 

As far as the `OpcodeService` is concerned, we just want to make sure that we are building and passing the correct `ContractDebugParameters` and `OpcodeTracerOptions` to the `ContractDebugService`. Do you agree?

xin-hedera: discussed offline. seems like it's a tradeoff we have to make.

steven-sheehy (approved):

steven-sheehy (approved):

PR #8458, opened 2024-06-03
test: 8449-refactor-contract-call-dynamic-calls-integration-test

fchitakova:
**Description**:

The idea of this change is to make the test independent of `ContractCallTestSetup` class and only reuse serviceParametersForExecution method from the base class.

**Related issue(s)**: https://github.com/hashgraph/hedera-mirror-node/issues/8449 

Fixes #

**Notes for reviewer**:

**Checklist**


hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallDynamicCallsTest.java
```diff
@@ -16,33 +16,131 @@
...
+            Bytes.wrap(recoverAddressFromPubKey(SENDER_PUBLIC_KEY.substring(2).toByteArray())));
+    private static final Address TREASURY_ADDRESS = toAddress(EntityId.of(0, 0, 743));
+    private static final ByteString NOT_ASSOCIATED_SPENDER_PUBLIC_KEY =
+            ByteString.fromHex("3a21033a514176466fa815ed481ffad09110a2d344f6c9b78c1d14afc351c3a51be33d");
+    private static final Address NOT_ASSOCIATED_SPENDER_ALIAS = Address.wrap(Bytes.wrap(recoverAddressFromPubKey(
+            NOT_ASSOCIATED_SPENDER_PUBLIC_KEY.substring(2).toByteArray())));
+    private static final Address OWNER_ADDRESS = toAddress(EntityId.of(0, 0, 1044));
+
+    // Token addresses
+    private static final Address RECEIVER_ADDRESS = toAddress(EntityId.of(0, 0, 1045));
```
xin-hedera, line 102: nit, just want to point out the shorter form `EntityId.of(1045)`

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallDynamicCallsTest.java
```diff
@@ -71,37 +171,537 @@ void dynamicCallsTestWithAliasSenderForEstimateGas(DynamicCallsContractFunctions
...
+        domainBuilder
+                .fileData()
+                .customize(f -> f.fileData(exchangeRatesSet.toByteArray())
+                        .entityId(exchangeRateEntityId)
+                        .consensusTimestamp(expiry))
+                .persist();
+    }
+
+    private void feeSchedulesPersist() {
+        var feeScheduleEntityId = EntityId.of(0L, 0L, 111L);
```
xin-hedera, line 669: this was made public static from a main java source file in the other PR, I believe. also applies to file 112

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallDynamicCallsTest.java
```diff
@@ -16,33 +16,131 @@
...
+    void setup() {
+        feeSchedulesPersist();
+        exchangeRatesPersist();
+    }
 
     @ParameterizedTest
     @EnumSource(DynamicCallsContractFunctions.class)
     void dynamicCallsTestWithAliasSenderForEthCall(DynamicCallsContractFunctions contractFunctions) {
+        persistDynamicCallTestEntities();
+        final var senderAddress = toAddress(EntityId.of(0, 0, 1043));
```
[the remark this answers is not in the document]
fchitakova: Done, thanks!

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallDynamicCallsTest.java
```diff
@@ -16,33 +16,131 @@
...
+import org.jetbrains.annotations.Nullable;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.EnumSource;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
 
-class ContractCallDynamicCallsTest extends ContractCallTestSetup {
+class ContractCallDynamicCallsTest extends Web3IntegrationTest {
```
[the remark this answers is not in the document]
fchitakova: Done, thanks

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallDynamicCallsTest.java
fchitakova: `EntityId.of` returns `EntityId`, but we need `Address` there

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallDynamicCallsTest.java
fchitakova: @xin-hedera , this Id is used only here, and the intend of this PR is to make:
1. the tests independent of other classes which creates their data 
2. every test to create its own data 

Also, every test case needs different fee schedules persisted so reusing it from other class in every this reverse the effort of this refactoring since FeeSchedules would contain data which are not needed in every test case

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallDynamicCallsTest.java
```diff
@@ -16,29 +16,122 @@
...
+    // Contract addresses
+    private static final Address DYNAMIC_ETH_CALLS_CONTRACT_ALIAS =
+            Address.fromHexString("0x742d35Cc6634C0532925a3b844Bc454e4438f44e");
+
+    // Account addresses
+    private static final ByteString SPENDER_PUBLIC_KEY =
+            ByteString.fromHex("3a2102ff806fecbd31b4c377293cba8d2b78725965a4990e0ff1b1b29a1d2c61402310");
+    private static final Address SPENDER_ALIAS = Address.wrap(
+            Bytes.wrap(recoverAddressFromPubKey(SPENDER_PUBLIC_KEY.substring(2).toByteArray())));
+    private static final ByteString SENDER_PUBLIC_KEY =
```
IvanKavaldzhiev, line 87: For this and all connected PRs - we should try to delete as much static fields like this, as possible. Instead, we should try to rely on dynamically generated keys, addresses, etc by the `DomainBuilder`. I think this can be done in a separate PR but will be important to not rely on a hardcoded value to have a more effective refactoring.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallDynamicCallsTest.java
fchitakova: Thanks, just applied this!

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallDynamicCallsTest.java
fchitakova: Thanks, we need to remove the enums in order to achieve this to generate addresses and ids dynamically, this is work in progress now.

PR #8462, opened 2024-06-03
Undo file_data time partition in V2

### LINKED ISSUE #8426: Web3 FileDataRepository.getFileAtTimestamp much slower in V2
### Description

`FileDataRepository.getFileAtTimestamp` is more than 10 times slower in V2 with citus than in V1 with postgresql. It's mainly caused by the monthly time partitions.

### Steps to reproduce

check the metrics in grafana

### Additional context

_No response_

### Hedera network

other

### Version

v0.107.0-SNAPSHOT

### Operating system

None

xin-hedera:
**Description**:

This PR undoes the `file_data` table time partition in V2

- Undo `file_data` table time partition
- Drop the unused `(consensus_timestamp, entity_id) primary key

**Related issue(s)**:

Fixes #8426 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-importer/src/main/resources/db/migration/v2/V2.2.0__undo_file_data_time_partition.sql
+++ b/hedera-mirror-importer/src/main/resources/db/migration/v2/V2.2.0__undo_file_data_time_partition.sql
@@ -1,0 +1,9 @@
+alter table if exists file_data rename to file_data_old;
+
+create table if not exists file_data (like file_data_old including defaults);
+select create_distributed_table('file_data', 'entity_id', colocate_with => 'entity');
+insert into file_data select * from file_data_old;
+
+drop table if exists file_data_old cascade;
+
+create index if not exists file_data__id_timestamp on file_data (entity_id, consensus_timestamp);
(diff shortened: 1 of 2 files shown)


hedera-mirror-importer/src/main/resources/db/migration/v2/V2.2.0__undo_file_data_time_partition.sql
```diff
@@ -0,0 +1,9 @@
+alter table if exists file_data rename to file_data_old;
+
+create table if not exists file_data (like file_data_old including defaults);
+select create_distributed_table('file_data', 'entity_id', colocate_with => 'entity');
+insert into file_data select * from file_data_old;
+
+drop table if exists file_data_old cascade;
+
+create index if not exists file_data__id_timestamp on file_data (entity_id, consensus_timestamp);
```
xin-hedera, line 9: there is a primary key `(consensus_timestamp, entity_id)`, believe it was there because `consensus_timestamp` is the table's primary key in V1.

however, in V2, since the table is sharded, it's not going to detect duplicate `consensus_timestamp` when the entity ids are different. In addition, we need to move away from being too strict and halting importer as the result.

jnels124 (approved):
LGTM

PR #8468, opened 2024-06-04
Add fallback to REST API fileService

### LINKED ISSUE #8419: Fallback to the previous system file if the current file is corrupted: REST API
### Description

System files are uploaded in chunks and sometimes errors can cause subsequent chunks to not get uploaded causing the file to be incomplete.

### Steps to reproduce

Upload a partial system file
Query that system file e.g. /api/v1/network/fees

### Additional context

_No response_

### Hedera network

mainnet

### Version

v0.104.0

### Operating system

None

edwin-greene:
**Description**:
Adds a fallback mechanism to the REST APIs for `api/v1/network/exchangerate` and `api/v1/network/fees`. If the timestamp requested version of the file cannot be parsed the api will automatically fall back to the most recent version that can be parsed.

- Add fallback in REST api for exchange and fee schedule
- Fix REST fileDataService not replacing all `timestamp` column references to avoid conflict
- Change web3 `FileDataRepository.getFileAtTimestamp` to return an Optional of `FileData`

**Related issue(s)**:

Fixes #8419 

**Notes for reviewer**:

**Checklist**


hedera-mirror-rest/service/fileDataService.js
```diff
@@ -18,6 +18,7 @@ import _ from 'lodash';
 
 import BaseService from './baseService';
 import {ExchangeRate, FileData, FeeSchedule} from '../model';
+import * as utils from '../utils.js';
```
xin-hedera, line 21:
```suggestion
import * as utils from '../utils';
```

hedera-mirror-rest/__tests__/specs/network/fees/fallbackMaxRetries.json
```diff
@@ -0,0 +1,105 @@
...
+        "file_data": "0a280a0a08541a061a04408888340a0a08061a061a04408",
+        "transaction_type": 19
+      }
+    ]
+  },
+  "urls": [
+    "/api/v1/network/fees?timestamp=1234567890.900800600",
+    "/api/v1/network/fees"
+  ],
+  "responseStatus": 500,
```
xin-hedera, line 95: We should return 404, not 500

hedera-mirror-rest/service/fileDataService.js
```diff
@@ -103,13 +104,48 @@ class FileDataService extends BaseService {
...
+
+      try {
+        return new resultConstructor(row);
+      } catch (error) {
+        caughtError = error;
+        logger.warn(
+          `Failed to load file data for file id ${fileEntityId} at ${filterQueries.timestamp}, failing back to previous file. Retry attempt ${retries}. Error: ${error}`
+        );
+
+        const fallBackTimestamp = row.consensus_timestamp - 1n;
```
xin-hedera, line 132: `consensus_timestamp` is the last not the first so this will not work when the file contents are from multiple file related transactions

hedera-mirror-rest/service/fileDataService.js
```diff
@@ -103,13 +104,48 @@ class FileDataService extends BaseService {
...
+        if (_.isNil(filterQueries.whereQuery[0]?.param)) {
+          // The query has no timestamp filters, add one to fall back to the previous file
+          filterQueries.whereQuery.push({
+            query: FileData.CONSENSUS_TIMESTAMP + utils.opsMap.lte,
+            param: fallBackTimestamp,
+          });
+        } else {
+          filterQueries.whereQuery[0].param = fallBackTimestamp;
```
xin-hedera, line 140: this is a hack and may or may not work depending on what kind of `timestamp=?` is in the request URL.

xin-hedera (changes requested):

hedera-mirror-rest/service/fileDataService.js
edwin-greene: Thanks I have updated this.

hedera-mirror-rest/__tests__/specs/network/fees/fallbackMaxRetries.json
edwin-greene: Got it ok I have fixed this.

hedera-mirror-rest/service/fileDataService.js
edwin-greene: Thanks for catching this. I have updated the query to return the minimum timestamp.

hedera-mirror-rest/service/fileDataService.js
edwin-greene: Excellent thanks, I have updated this.

hedera-mirror-rest/service/fileDataService.js
```diff
@@ -103,13 +105,35 @@ class FileDataService extends BaseService {
...
+        logger.warn(
+          `Failed to load file data for file id ${fileEntityId} at ${row.consensus_timestamp}, failing back to previous file. Retry attempt ${retries}. Error: ${error}`
+        );
+
+        currentFilterQueries.whereQuery = [
+          {
+            query: FileData.CONSENSUS_TIMESTAMP + utils.opsMap.lt,
+            param: row.fallback_timestamp,
+          },
+        ];
```
xin-hedera, line 132: just realized `filterQueries` is really weird: `{whereQuery: []}`

nevertheless, your change doesn't fully address my previous feedback. it's still replacing the whole `whereQuery` array.

what if the original `whereQuery` is

```js
[
  {
    query: 'consensus_timestamp > ',
    param: 100000,
  },
  {
    query: 'consensus_timestamp < ',
    param: 200000,
  }
]
```

The logic here would throw away the original conditions and it shouldn't, right?

xin-hedera (changes requested):

mgoelswirlds (approved):
LGTM

xin-hedera (approved):
LGTM

jascks (approved):
LGTM

PR #8472, opened 2024-06-04
convert topic message lookup to async migration

### LINKED ISSUE #8461: Convert TopicMessageLookupMigration to Async Migration
### Problem

The current migration is synchronous and blocks startup.

### Solution

Convert to be asynchronous

### Alternatives

_No response_

jnels124:
**Description**:
* convert topic message lookup migration to be asynchronous

**Related issue(s)**:

Fixes #8461

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/TopicMessageLookupMigration.java
+++ b/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/TopicMessageLookupMigration.java
@@ -25,8 +25,8 @@ import com.hedera.mirror.importer.exception.ParserException;
 import com.hedera.mirror.importer.parser.record.entity.EntityProperties;
 import com.hedera.mirror.importer.repository.RecordFileRepository;
 import jakarta.inject.Named;
-import java.util.Comparator;
 import java.util.LinkedList;
+import java.util.List;
 import java.util.Optional;
 import lombok.Getter;
 import org.apache.commons.lang3.BooleanUtils;
@@ -59,7 +59,7 @@ public class TopicMessageLookupMigration extends AsyncJavaMigration<String> {
 
     private final EntityProperties entityProperties;
     private final JdbcTemplate jdbcTemplate;
-    private final LinkedList<String> partitions = new LinkedList<>();
+    private final List<String> partitions = new LinkedList<>();
 
     private final RecordFileRepository recordFileRepository;
     private final TimePartitionService timePartitionService;
@@ -118,10 +118,8 @@ public class TopicMessageLookupMigration extends AsyncJavaMigration<String> {
         partitions.addAll(timePartitions.stream()
                 .filter(p -> p.getTimestampRange().lowerEndpoint() <= lastRecordFile.get()
                         && !p.getName().equals(activePartition))
-                .sorted(Comparator.comparing(p -> p.getTimestampRange().upperEndpoint()))
                 .map(TimePartition::getName)
-                .toList()
-                .reversed());
+                .toList());
 
         log.info("Migrating topic_message_lookup for partition {} synchronously", activePartition);
         getTransactionOperations().executeWithoutResult(status -> migratePartition(activePartition));
@@ -131,13 +129,13 @@ public class TopicMessageLookupMigration extends AsyncJavaMigration<String> {
 
     @Override
     protected String getInitial() {
-        return partitions.poll();
+        return partitions.isEmpty() ? null : partitions.removeLast();
     }
 
     @Override
     protected Optional<String> migratePartial(String partitionName) {
         migratePartition(partitionName);
-        return Optional.ofNullable(partitions.poll());
+        return Optional.ofNullable(getInitial());
     }
 
     private void migratePartition(String partitionName) {
--- a/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/TopicMessageLookupMigration.java
+++ b/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/TopicMessageLookupMigration.java
@@ -120,10 +120,7 @@ public class TopicMessageLookupMigration extends AsyncJavaMigration<String> {
                 .toList());
 
         log.info("Migrating topic_message_lookup for partition {} synchronously", activePartition.getName());
-        getTransactionOperations().executeWithoutResult(status -> {
-            migratePartition(activePartition.getName());
-            log.info("Migrated topic_message_lookup for partition {}", activePartition.getName());
-        });
+        getTransactionOperations().executeWithoutResult(status -> migratePartition(activePartition.getName()));
 
         return !partitions.isEmpty();
     }
@@ -144,6 +141,7 @@ public class TopicMessageLookupMigration extends AsyncJavaMigration<String> {
         if (BooleanUtils.isFalse(alreadyMigrated)) {
             var sql = String.format(UPDATE_PARTITION_SQL, partitionName);
             jdbcTemplate.update(sql);
+            log.info("Migrated topic_message_lookup for partition {}", partitionName);
         } else {
             log.info("Partition {} already migrated", partitionName);
         }
--- a/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/TopicMessageLookupMigration.java
+++ b/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/TopicMessageLookupMigration.java
@@ -16,17 +16,17 @@
 
 package com.hedera.mirror.importer.migration;
 
-import com.google.common.collect.Range;
+import com.hedera.mirror.common.domain.transaction.RecordFile;
 import com.hedera.mirror.importer.ImporterProperties;
 import com.hedera.mirror.importer.db.DBProperties;
 import com.hedera.mirror.importer.db.TimePartition;
 import com.hedera.mirror.importer.db.TimePartitionService;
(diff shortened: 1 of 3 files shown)


hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/TopicMessageLookupMigration.java
```diff
@@ -16,94 +16,135 @@
...
-                "Successfully backfilled topic_message_lookup table for {} time partitions in {}",
-                partitions.size(),
-                stopwatch);
+        var activePartition = timePartitions.stream()
+                .filter(p -> p.getTimestampRange().contains(lastRecordFile.get()))
+                .findFirst()
+                .orElseThrow(() -> new ParserException("No active partition found for topic_message"));
+
+        partitions.addAll(timePartitions.stream()
+                .filter(p -> p.getTimestampRange().lowerEndpoint() <= lastRecordFile.get()
```
xin-hedera, line 118: nit,  a simpler condition is `upper endpoint < consensus start of last record file`

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/TopicMessageLookupMigration.java
```diff
@@ -16,94 +16,135 @@
...
-                var sql = String.format(UPDATE_PARTITION_SQL, partition.getName());
-                jdbcTemplate.update(sql);
-            }
-        });
-
-        log.info(
-                "Successfully backfilled topic_message_lookup table for {} time partitions in {}",
-                partitions.size(),
-                stopwatch);
+        var activePartition = timePartitions.stream()
```
xin-hedera, line 112: there is flow in the last active partition detection: Since the migration is really slow, the active partition can change in between runs. For example, the migration is kicked off with 2024-05 as the last active partition, runs for a couple of days and importer gets restarted. The second time the migration runs, the active partition becomes 2024-06.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/TopicMessageLookupMigration.java
```diff
@@ -16,94 +16,135 @@
...
+
+        log.info("Migrating topic_message_lookup for partition {} synchronously", activePartition.getName());
+        getTransactionOperations().executeWithoutResult(status -> migratePartition(activePartition.getName()));
+
+        return !partitions.isEmpty();
+    }
+
+    @Override
+    protected String getInitial() {
+        return Objects.requireNonNull(partitions.poll()).getName();
```
xin-hedera, line 130: we should run the async part backwards

xin-hedera (changes requested):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/TopicMessageLookupMigration.java
jnels124: Because the active partition is populated synchronously at the start it doesn't matter right? The active partition will always be the partition that we are currently ingesting files for and will always be  populated up to `select max(consensus_end) from record_file` before any record_file processing happens.

So at the start of the migration if we have `consensus_end='2024-05-31 16:00:00' `

We start importer and the migration runs. The synchronous portion of the migration will write all entries for `topic_message` up to that point before record file processing occurs. 

If we do a deploy later say when `consensus_end='2024-06-03 16:00:00'`. Then there is no synchronous work to be performed by the migration because there will be entries for the active partition so it will be skipped.

jnels124: There is an issue though if the active record_file spans partitions. will fix that

xin-hedera: Ah, I think I missed the empty check in `migrationPartition`

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/TopicMessageLookupMigration.java
jnels124: sorted to process in desc order

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/TopicMessageLookupMigration.java
```diff
@@ -16,94 +16,140 @@
...
+        if (timePartitions.isEmpty()) {
+            log.info("Skip the migration since topic_message doesn't contain any partitions");
+            return false;
         }
 
-        var transactionManager = new DataSourceTransactionManager(Objects.requireNonNull(jdbcTemplate.getDataSource()));
-        var transactionTemplate = new TransactionTemplate(transactionManager);
-        transactionTemplate.executeWithoutResult(s -> {
-            jdbcTemplate.update(TRUNCATE_TABLE_SQL);
+        var activePartitions = timePartitions.stream()
```
jnels124, line 113: This could actually be a single partition. If boundary is crossed, we only care about the portion of the record file that is in the new partition for the synchronous part of the migration. Will make this change.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/TopicMessageLookupMigration.java
jnels124: I left this condition as is because it allows us to only process the portion of the record file that is going to be written to by the next one.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/TopicMessageLookupMigration.java
```diff
@@ -16,94 +16,139 @@
...
+                    group by topic_id;
+                    """;
+    private static final String PARTITION_NEEDS_MIGRATION_SQL =
+            """
+                    select exists (select 1 from topic_message_lookup where partition = ?)
+                    """;
 
     private final EntityProperties entityProperties;
     private final JdbcTemplate jdbcTemplate;
+    private final LinkedList<String> partitions = new LinkedList<>();
```
xin-hedera, line 62: nit
```suggestion
    private final List<String> partitions = new LinkedList<>();
```

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/TopicMessageLookupMigration.java
```diff
@@ -16,94 +16,139 @@
...
+        var activePartition = timePartitions.stream()
+                .filter(p -> p.getTimestampRange().contains(lastRecordFile.get()))
+                .map(TimePartition::getName)
+                .findFirst()
+                .orElseThrow(() -> new ParserException("No active partition found"));
+
+        partitions.addAll(timePartitions.stream()
+                .filter(p -> p.getTimestampRange().lowerEndpoint() <= lastRecordFile.get()
+                        && !p.getName().equals(activePartition))
+                .sorted(Comparator.comparing(p -> p.getTimestampRange().upperEndpoint()))
```
xin-hedera, line 121: `TimePartitionService.getTimePartitions` already returns the partitions sorted by the timestamp range in ascending order.

```suggestion
```

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/migration/TopicMessageLookupMigrationTest.java
```diff
@@ -218,5 +302,19 @@ void notPartitioned() {
     @SneakyThrows
     private void runMigration() {
         migration.doMigrate();
+        waitForCompletion();
+    }
+
+    protected void waitForCompletion() {
+        await().atMost(Duration.ofSeconds(5))
```
xin-hedera, line 309: perhaps increase the wait time to 30s  avoid any weirdness in github CI

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/migration/TopicMessageLookupMigrationTest.java
```diff
@@ -218,5 +302,19 @@ void notPartitioned() {
     @SneakyThrows
     private void runMigration() {
         migration.doMigrate();
+        waitForCompletion();
+    }
+
+    protected void waitForCompletion() {
```
xin-hedera, line 308: private?
```suggestion
    private void waitForCompletion() {
```

xin-hedera (changes requested):
just some nitpickings

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/migration/TopicMessageLookupMigrationTest.java
jnels124: oops yes definitely intended to make private

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/TopicMessageLookupMigration.java
jnels124: was using its deque implementation but not really necessary. will change

xin-hedera (approved):
LGTM

mgoelswirlds (approved):
LGTM

PR #8457, opened 2024-06-03
Refactor `SelfDestructOperationTest`

fchitakova:
**Description**:

The idea of this change is to make the test independent of ContractCallTestSetup class and only reuse `serviceParametersForExecution` method from the base class. 

**Related issue(s)**: https://github.com/hashgraph/hedera-mirror-node/issues/8448 

Fixes #

**Notes for reviewer**:

**Checklist**

- [x] Documented (Code comments, README, etc.)
- [X] Tested (unit, integration, etc.)

PR #8483, opened 2024-06-06
Refactor `ContractCallServiceERCTokenTest` in web3

### LINKED ISSUE #8456: Refactor `ContractCallServiceERCTokenTest `

petreze:
**Description**:
This PR aims to refactor `ContractCallServiceERCTokenTest` integration test files by removing redundancies from ContractCallTestSetup and include only needed DB setup data and entities for the specific test cases in those files

**Related issue(s)**:

Fixes #8456 

**Notes for reviewer**:
There are several methods from the base class that has to be identical in `ContractCallServiceERCTokenTest` and thats why IMO we can reuse them..

**Checklist**


hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallTestSetup.java
```diff
@@ -1703,7 +1702,7 @@ private void fileDataPersist() {
     }
 
     // Account persist
-    private void tokenAccountPersist(
+    protected void tokenAccountPersist(
```
IvanKavaldzhiev, line 1705: Can you try to use `DomainBuilder` to dynamically create the entities you need inside the test itself, instead of relying on a parent class. If we achieve this decoupling, it will be a lot easier to get rid of `ContractCallTestSetup`.

If you plan to use a step by step approach and achieve this in steps - then we should be fine.

petreze: In some cases we pass those specific hardcoded addresses as input parameters or expected results to the test enums. Also we have some common logic inside `ContractCallTestSetup` which would be useful to reuse..

steven-sheehy: I don't think this PR or any of the others provide value if we don't do what Ivan suggests. Besides coupling it is also inefficient because every test sets up all data even if it doesn't need it. So it increases test duration and makes it harder to determine what db changes are truly necessary for the specific test case. The problem is the use of enums forces us to use hardcoded values right now. If we can get the solidity gradle plugin to work then each test can create its minimal, dynamic input.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServiceERCTokenTest.java
```diff
@@ -202,15 +362,114 @@ void supportedErcModificationsRedirectPrecompileOperationsTest(final ErcContract
...
+    protected CallServiceParameters serviceParametersForExecution(
+            final Bytes callData,
+            final Address contractAddress,
+            final CallServiceParameters.CallType callType,
+            final long value,
+            final BlockType block) {
+        return serviceParametersForExecution(callData, contractAddress, callType, value, block, 15_000_000L);
+    }
+
+    protected CallServiceParameters serviceParametersForExecution(
```
steven-sheehy, line 384: Prefer to return the builder and customize in caller. This would avoid needing overloaded server parameter helper methods. See my comment in https://github.com/hashgraph/hedera-mirror-node/pull/8457#discussion_r1635304471

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallTestSetup.java
steven-sheehy: > If you plan to use a step by step approach and achieve this in steps - then we should be fine.

Also, agree with Ivan that this could be a stepping stone to push down global variables into sub-classes as a first step.

steven-sheehy (changes requested):

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServiceERCTokenTest.java
petreze: @steven-sheehy 
I agree entirely on returning the builder itself which will give a lot of flexibility but can we keep `contractAddress` and `callData` parameters as well as they are going to be set dynamically on every test. So after calling `serviceParametersForExecution()` instead of having `builder.callData().receiver()` we can pass those in the method call itself?

steven-sheehy: That's fine. Main purpose is to reduce very large parameter list. 1-3 params are okay if they are required.

steven-sheehy:
Closing in favor of the approach in #8385

PR #8488, opened 2024-06-06
Fix invalid next link when the last pagination param has eq operator

### LINKED ISSUE #8407: RestJava LinkFactory creates invalid next link when the last pagination param has eq operator
### Description

Given the request url `http://mirror-restjava/api/v1/accounts/99550/allowances/nfts?limit=2&account.id=gte:0.0.99700&token.id=0.0.99800`, and there are more than the `limit` number of matching results, the link factory will create such a next link:

```json
"links": {
  "next": "/api/v1/accounts/99550/allowances/nfts?limit=2&token.id=0.0.99800&token.id=gt:0.0.99800&account.id=gte:0.0.99701"
}
```

Following the link will get a 400 code with the body

```json
{
  "_status": {
    "messages": [
      {
        "data": null,
        "detail": null,
        "message": "Can't support both range and equal for token.id"
      }
    ]
  }
}
```

The correct next link should be `/api/v1/accounts/99550/allowances/nfts?limit=2&token.id=0.0.99800&account.id=gt:0.0.99701`.

In general, with the pagination parameters in the order from the left most to the right most, if the current pagination parameter's operator is `eq`, the link factory should copy the existing `p=v` pair from the request URL, and the operator for the pagination parameter before it should be exclusive.

### Steps to reproduce

check the description

### Additional context

_No response_

### Hedera network

other

### Version

v0.106.0-rc1

### Operating system

None

edwin-greene:
**Description**:

Fixes the next link generation and corrects the succeeding query parameter operators.

**Related issue(s)**:

Fixes #8407

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
+++ b/hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
@@ -21,10 +21,10 @@ import com.hedera.mirror.rest.model.Links;
 import jakarta.annotation.Nonnull;
 import jakarta.inject.Named;
 import jakarta.servlet.http.HttpServletRequest;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
-import java.util.TreeMap;
 import java.util.function.Function;
 import lombok.RequiredArgsConstructor;
 import org.springframework.data.domain.Pageable;
@@ -105,28 +105,29 @@ class LinkFactoryImpl implements LinkFactory {
         }
     }
 
+    @SuppressWarnings("java:S1125")
     private void addExtractedParamsToQueryParams(
             Sort sort,
             Map<String, String> paginationParamsMap,
             Direction order,
             LinkedMultiValueMap<String, String> queryParams) {
-        var sortEqMap = new TreeMap<String, Boolean>();
-        sort.map(Sort.Order::getProperty).forEach(s -> sortEqMap.put(s, containsEq(queryParams.get(s))));
+        var sortEqMap = new HashMap<String, Boolean>();
+        var sortList = sort.map(s -> {
+                    var property = s.getProperty();
+                    sortEqMap.put(property, containsEq(queryParams.get(property)));
+                    return property;
+                })
+                .toList();
 
-        var sortKeys = sortEqMap.keySet().toArray();
-        for (int i = 0; i < sortKeys.length; i++) {
-            var key = sortKeys[i].toString();
+        for (int i = 0; i < sortList.size(); i++) {
+            var key = sortList.get(i);
             if (queryParams.containsKey(key) && Boolean.TRUE.equals(sortEqMap.get(key))) {
                 // This query parameter has already been added with an eq
                 continue;
             }
 
-            var exclusive = true;
-            if (sortKeys.length > i + 1) {
-                var succeedingKey = sortKeys[i + 1];
-                exclusive = sortEqMap.get(succeedingKey);
-            }
-
+            int nextParamIndex = i + 1;
+            boolean exclusive = sortList.size() > nextParamIndex ? sortEqMap.get(sortList.get(nextParamIndex)) : true;
             var value = paginationParamsMap.get(key);
             queryParams.add(key, getOperator(order, exclusive) + ":" + value);
         }
@@ -148,11 +149,13 @@ class LinkFactoryImpl implements LinkFactory {
     }
 
     private static boolean containsEq(List<String> values) {
-        if (values != null) {
-            for (var value : values) {
-                if (hasEq(value)) {
-                    return true;
-                }
+        if (values == null) {
+            return false;
+        }
+
+        for (var value : values) {
+            if (hasEq(value)) {
+                return true;
             }
         }
 
--- a/hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
+++ b/hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
@@ -64,14 +64,22 @@ class LinkFactoryImpl implements LinkFactory {
         var sortOrders = pageable.getSort();
         var primarySort = Iterables.getFirst(sortOrders, null);
         var order = primarySort == null ? Direction.ASC : primarySort.getDirection();
-        var lastSort = Iterables.getLast(sortOrders, null);
-        // Only the last sort param's value is exclusive
-        var exclusiveParam = lastSort != null ? lastSort.getProperty() : null;
         var builder = UriComponentsBuilder.fromPath(request.getRequestURI());
         var paramsMap = request.getParameterMap();
         var paginationParamsMap = extractor.apply(lastItem);
         var linkedMultiValueMap = new LinkedMultiValueMap<String, String>();
 
+        addParamMapToLink(paramsMap, paginationParamsMap, order, linkedMultiValueMap);
+        addExtractedParamsToLink(sortOrders, paginationParamsMap, order, linkedMultiValueMap);
+        builder.queryParams(linkedMultiValueMap);
+        return builder.toUriString();
+    }
+
+    private void addParamMapToLink(
+            Map<String, String[]> paramsMap,
+            Map<String, String> paginationParamsMap,
+            Direction order,
+            LinkedMultiValueMap<String, String> linkedMultiValueMap) {
         for (var entry : paramsMap.entrySet()) {
             var key = entry.getKey();
             if (!paginationParamsMap.containsKey(key)) {
@@ -82,21 +90,6 @@ class LinkFactoryImpl implements LinkFactory {
                 addQueryParamToLink(entry, order, linkedMultiValueMap);
             }
         }
-
-        var sortKeys = sortOrders.map(Sort.Order::getProperty).toList();
-        for (var entry : paginationParamsMap.entrySet()) {
-            var key = entry.getKey();
-            if (linkedMultiValueMap.containsKey(key) && containsEq(linkedMultiValueMap.get(key))) {
-                // This query parameter has already been added with an eq
-                continue;
-            }
-
-            var exclusive = isExclusive(exclusiveParam, sortKeys, key, linkedMultiValueMap);
-            linkedMultiValueMap.add(key, getOperator(order, exclusive) + ":" + entry.getValue());
-        }
-
-        builder.queryParams(linkedMultiValueMap);
-        return builder.toUriString();
     }
 
     private void addQueryParamToLink(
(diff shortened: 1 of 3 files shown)


hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
```diff
@@ -62,39 +64,79 @@ private <T> String createNextLink(
         var sortOrders = pageable.getSort();
         var primarySort = Iterables.getFirst(sortOrders, null);
         var order = primarySort == null ? Direction.ASC : primarySort.getDirection();
-        var lastSort = Iterables.getLast(sortOrders, null);
-        // Only the last sort param's value is exclusive
-        var exclusiveParam = lastSort != null ? lastSort.getProperty() : null;
         var builder = UriComponentsBuilder.fromPath(request.getRequestURI());
         var paramsMap = request.getParameterMap();
         var paginationParamsMap = extractor.apply(lastItem);
+        var linkedMultiValueMap = new LinkedMultiValueMap<String, String>();
```
xin-hedera, line 70: this is really bad variable name.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
```diff
@@ -62,39 +64,79 @@ private <T> String createNextLink(
...
 
+        addParamMapToLink(paramsMap, paginationParamsMap, order, linkedMultiValueMap);
```
xin-hedera, line 72: the function name reads as add them to link, actually it adds items to a map. Adding to link is an operation which happens later.

```suggestion
        addParamMapToLink(paramsMap, paginationParamsMap, order, linkedMultiValueMap);
```

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
```diff
@@ -62,39 +64,79 @@ private <T> String createNextLink(
...
             if (isSameDirection(order, value)) {
                 continue;
             }
 
-            builder.queryParam(entry.getKey(), value);
+            linkedMultiValueMap.add(entry.getKey(), value);
+            if (hasEq(value)) {
+                // Only 1 eq is allowed, return without adding additional values
+                return;
+            }
```
xin-hedera, line 107: remove. this is invalid, and the factory should not assume rules enforced by controllers.

```suggestion
```

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
```diff
@@ -62,39 +64,79 @@ private <T> String createNextLink(
...
+            Map<String, String> paginationParamsMap,
+            Direction order,
+            LinkedMultiValueMap<String, String> linkedMultiValueMap) {
+        var sortKeys = sort.map(Sort.Order::getProperty).toList();
+        if (sortKeys.isEmpty()) {
+            for (var entry : paginationParamsMap.entrySet()) {
+                linkedMultiValueMap.add(entry.getKey(), entry.getValue());
+            }
+            return;
+        }
```
xin-hedera, line 122: if somehow no sort keys provided, run the other logic to determine whether the operator should be exclusive or not. We should not short cut here.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
```diff
@@ -62,39 +64,79 @@ private <T> String createNextLink(
...
+
+        for (int i = 0; i < sortKeys.size(); i++) {
+            var key = sortKeys.get(i);
+            if (linkedMultiValueMap.containsKey(key) && containsEq(linkedMultiValueMap.get(key))) {
```
xin-hedera, line 126: the containsEq logic runs twice for the same query parameter, instead, we should preprocess if for the sort columns and store the result for reuse.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
```diff
@@ -62,39 +64,79 @@ private <T> String createNextLink(
...
+                // This query parameter has already been added with an eq
+                continue;
+            }
+
+            var entry = paginationParamsMap.get(key);
```
xin-hedera, line 131:
```suggestion
            var value = paginationParamsMap.get(key);
```

since `entry` is misleading.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
```diff
@@ -112,4 +154,21 @@ private static boolean isSameDirection(Direction order, String value) {
             case DESC -> normalized.startsWith("lt:") || normalized.startsWith("lte:");
         };
     }
+
+    private static boolean containsEq(List<String> values) {
+        if (values != null) {
```
xin-hedera, line 159: no need to do null check
```suggestion
```

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
```diff
@@ -112,4 +154,21 @@ private static boolean isSameDirection(Direction order, String value) {
...
+                }
+            }
+        }
+
+        return false;
+    }
+
+    private static boolean hasEq(String value) {
+        var normalized = value.toLowerCase();
+        return normalized.startsWith("eq:") || (!normalized.startsWith("gt") && !normalized.startsWith("lt"));
```
xin-hedera, line 172: what if a value starts with `gt` or `lt` however it's not an operator?

xin-hedera (changes requested):

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
edwin-greene: Thanks I have updated these method names.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
edwin-greene: Sounds good I have removed it.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
edwin-greene: Got it, I have removed this section.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
edwin-greene: Thanks I have updated it to perform the `containsEq` for each sort prior to the rest.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
edwin-greene: Thanks I have updated it.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
edwin-greene: I have removed it.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
edwin-greene: Got it ok, I have updated it to include the ":" and each operator to validate that the string is believed to be an operator.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
xin-hedera: it's still here

xin-hedera: Since the logic has changed I see now null check is needed. Just a nit, short cut return false, so we have less nested logic.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
```diff
@@ -62,39 +65,70 @@ private <T> String createNextLink(
...
+            queryParams.add(entry.getKey(), value);
+        }
+    }
+
+    private void addExtractedParamsToQueryParams(
+            Sort sort,
+            Map<String, String> paginationParamsMap,
+            Direction order,
+            LinkedMultiValueMap<String, String> queryParams) {
+        var sortEqMap = new TreeMap<String, Boolean>();
```
xin-hedera, line 113: `TreeMap` sorts keys in its natural order, so it's not the order the pagination params.

Naive way is just use a for loop and create both a linkedList and map.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
```diff
@@ -62,39 +65,70 @@ private <T> String createNextLink(
...
+            if (queryParams.containsKey(key) && Boolean.TRUE.equals(sortEqMap.get(key))) {
+                // This query parameter has already been added with an eq
+                continue;
+            }
+
+            var exclusive = true;
+            if (sortKeys.length > i + 1) {
+                var succeedingKey = sortKeys[i + 1];
+                exclusive = sortEqMap.get(succeedingKey);
+            }
```
xin-hedera, line 128: shorten it to a ternary expression.

```suggestion
            boolean exclusion = sortKeys.size() > i + 1 ? sortEqMap.get(sortKeys.get(i + 1)) : true;
```

xin-hedera (changes requested):

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
edwin-greene: Thanks I originally mistook it for another item I removed. I have updated it to have the short cut.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
edwin-greene: Thanks!! I have fixed it so that the order will be correct, it won't use the map for iterating anymore.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/LinkFactoryImpl.java
edwin-greene: Thanks I have made this update

xin-hedera (approved):
LGTM

jnels124 (approved):
LGTM

PR #8489, opened 2024-06-07
fix: Handle empty `EthereumTransaction.toAddress` when building `CallServiceParameters`

victor-yanev:
For contract deploy transactions it is possible that the `toAddress` field of `EthereumTransaction` is an empty array.

With the current logic we only check for null values but not for empty arrays, so an exception will be thrown later when we try to get the receiver address: `An account address must be 20 bytes long, got 0`

This PR fixes the issue and also adds new test cases in `OpcodesControllerTest`.


xin-hedera (approved):
LGTM

PR #8507, opened 2024-06-10
fix: debug_traceTransaction with opcode logger returns failed for contract deploy

### LINKED ISSUE #8506: `debug_traceTransaction` with opcode logger returns failed for contract deployment
### Description

When we try to debug a contract create tx, we're trying to build the receiver's address [here](https://github.com/hashgraph/hedera-mirror-node/blob/8173-implement-opcode-logger-tracing-logic/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/OpcodeServiceImpl.java#L192). The adequate `to` value for contract deploy is `null` but [here](https://github.com/hashgraph/hedera-mirror-node/blob/8173-implement-opcode-logger-tracing-logic/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/OpcodeServiceImpl.java#L192) we populate it with the value from the contract result which is the address of newly created contract. After that population, we don't simulate a contract deployment but a simple contract call.

### Steps to reproduce

- create a contract deploy tx
- try to debug it and observe for `failed` field or executed opcodes

### Additional context

_No response_

### Hedera network

other

### Version

latest

### Operating system

None

victor-yanev:
**Description**:

When we try to debug a contract create tx, we're trying to build the receiver's address [here](https://github.com/hashgraph/hedera-mirror-node/blob/8173-implement-opcode-logger-tracing-logic/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/OpcodeServiceImpl.java#L192). 

The adequate `receiverAddress` value for contract deploy is `Address.ZERO`, but we populate it with the value from the `ContractResult` which is the address of newly created contract. 

After that population, we don't simulate a contract deployment but a simple contract call.

**Steps to reproduce:**
- create a contract deploy tx
- try to debug it and observe for failed field or executed opcodes

**Related issue(s)**:

Fixes #8506

### DIFF
--- a/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/utils/TransactionProviderEnum.java
+++ b/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/utils/TransactionProviderEnum.java
@@ -143,15 +143,15 @@ public enum TransactionProviderEnum {
             return domainBuilder.wrap(EthereumTransaction.builder(), () -> null);
         }
 
-        Address evmAddress;
+        final byte[] evmAddress;
         if (contractEvmAddress != null && contractEvmAddress.length == EVM_ADDRESS_LENGTH) {
-            evmAddress = Address.wrap(Bytes.wrap(contractEvmAddress));
+            evmAddress = contractEvmAddress;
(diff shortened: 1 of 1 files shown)


steven-sheehy (approved):
LGTM

PR #8535, opened 2024-06-12
Increase default memory for prometheus

### LINKED ISSUE #8530: Increase default memory for prometheus pod
### Problem

With additional metrics being generated, the prometheus pod is crashing due to OOM

### Solution

Increase default memory resources 

### Alternatives

_No response_

jnels124:
**Description**:
* bump mem limits/requests for prometheus

**Related issue(s)**:

Fixes #8530 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/charts/hedera-mirror-common/values.yaml
+++ b/charts/hedera-mirror-common/values.yaml
@@ -308,10 +308,10 @@ prometheus:
       resources:
         limits:
           cpu: 750m
-          memory: 2Gi
+          memory: 2750Mi
         requests:
           cpu: 250m
-          memory: 250Mi
+          memory: 1Gi
       retention: 60d
       ruleSelectorNilUsesHelmValues: false
       scrapeInterval: 30s


jnels124:
Screenshot 2024-06-12 at 11 46 58 AM

edwin-greene (approved):
Looks good

charts/hedera-mirror-common/values.yaml
```diff
@@ -308,10 +308,10 @@ prometheus:
       resources:
         limits:
           cpu: 750m
-          memory: 2Gi
```
xin-hedera, line 311: should we make this even higher?

charts/hedera-mirror-common/values.yaml
```diff
@@ -308,10 +308,10 @@ prometheus:
...
+          memory: 2750Mi
```
steven-sheehy, line 311: Prometheus was crashing in preprod when set to 4GB. Perhaps that environment is an outlier though since it contains many namespaces.

steven-sheehy (approved):

charts/hedera-mirror-common/values.yaml
jnels124: I left about a half a gig of wiggle room compared to its max value reached when setting it to 5Gi for the test period

charts/hedera-mirror-common/values.yaml
jnels124: Yes would make more sense to override for that cluster I think. Staging cluster usage is a better representation for establishing the defaults.

PR #8543, opened 2024-06-13
ci: Update madhead/semver-utils to use the step-security maintained version

### LINKED ISSUE #8542: Update semver-utils to use the step-security maintained semver-utils action
madhead/semver-utils should be updated to use step-security/semver-utils

rbarkerSL:
**Description**:

Replaces madhead/semver-utils with step-security/semver-utils

**Related issue(s)**:

Fixes #8542

### DIFF
--- a/.github/workflows/release-automation.yml
+++ b/.github/workflows/release-automation.yml
@@ -38,7 +38,7 @@ jobs:
 
       - name: Parse Version
         id: version_parser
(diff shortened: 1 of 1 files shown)


nathanklick (approved):

xin-hedera (commented):
Have we tested that it works the same way? Just wanna make sure it’s not going to cause release issue

rbarkerSL:
> Have we tested that it works the same way? Just wanna make sure it’s not going to cause release issue

Yes, we have tested this same change in several other repositories. It behaves correctly.

PR #8551, opened 2024-06-14
Implement cache control support in REST Java

### LINKED ISSUE #8346: RestJava feature parity - cache control support
### Problem

The JS REST API module supports generic cache-control header with config based fine tuning per endpoint, we need to implement the same in RestJava

### Solution

- add a function in `GenericControllerAdvice` to customize cache-control header in response 

### Alternatives

_No response_

jascks:
**Description**:

The JS REST API module supports generic `cache-control` header with config based fine tuning per endpoint, we need to implement the same in Rest Java.

The JS REST API can support any response headers, not just `cache-control`. REST Java now has the same capability for both default and path (controller request mapping) specific paths. Multiple headers may be configured, but only a single value per header.

- Adds `responseHeaders()` method to `GenericControllerAdvice`.
- Adds additional configuration handling to `RestJavaProperties`

**Related issue(s)**:

Fixes #8346

**Notes for reviewer**:
Please see my comments in this PR. Namely,  what should the initial `cache-control` headers look like for the two REST Java API endpoints?

**Checklist**


hedera-mirror-rest-java/src/main/resources/application.yml
```diff
@@ -9,6 +9,17 @@ hedera:
         sslMode: DISABLE
         statementTimeout: 10000
         username: mirror_rest_java
+      response:
+        headers:
+          defaults:
```
jascks, line 14: Curious what you think the appropriate configuration is we ship with. Neither of these endpoints is excessively taxing on the system/DB I believe.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/controller/ControllerTest.java
```diff
@@ -107,6 +116,26 @@ void cors() {
             assertThat(headers.getAccessControlAllowOrigin()).isEqualTo("*");
         }
 
+        @Test
+        void headers() {
```
steven-sheehy, line 120: We should have an assert in TopicControllerTest that verifies maxAge=5 and one in AllowancesControllerTest that verifies maxAge=1

steven-sheehy (changes requested):

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/controller/GenericControllerAdvice.java
```diff
@@ -41,19 +44,30 @@
...
 @Order(Ordered.HIGHEST_PRECEDENCE)
+@RequiredArgsConstructor
 class GenericControllerAdvice extends ResponseEntityExceptionHandler {
 
+    private final RestJavaProperties properties;
+
     @ModelAttribute
     @SuppressWarnings("java:S5122")
     private void corsHeader(HttpServletResponse response) {
         response.setHeader(HttpHeaders.ACCESS_CONTROL_ALLOW_ORIGIN, "*");
```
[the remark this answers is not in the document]
jascks: If there is a path override though, we need to add the access control header there as well. Defaults are replaced by path specific overrides, not merged.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/RestJavaProperties.java
```diff
@@ -28,4 +31,28 @@ public class RestJavaProperties {
...
+    @Validated
+    public static class ResponseHeadersConfig {
+        @NotNull
+        private Map<String, String> defaults = new HashMap<>();
+
+        @NotNull
+        private Map<String, Map<String, String>> path = new HashMap<>();
+
+        public Map<String, String> getHeadersForPath(String path) {
+            return path == null ? defaults : this.path.getOrDefault(path, defaults);
```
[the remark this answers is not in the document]
jascks: Done

hedera-mirror-rest-java/src/main/resources/application.yml
```diff
@@ -9,6 +9,17 @@ hedera:
...
+            "cache-control": "public, max-age=1"
+          path:
+            # Override default for individual controller request mappings. "[]" around the path preserves special characters.
+            # Multiple header name key/value pairs may be defined per path.
+            "[/api/v1/topics/{id}]":
+              "cache-control": "public, max-age=10"
```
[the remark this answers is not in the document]
jascks: Done

hedera-mirror-rest-java/src/main/resources/application.yml
```diff
@@ -9,6 +9,17 @@ hedera:
...
+            "[/api/v1/accounts/{id}/allowances/nfts]":
+              "cache-control": "public, max-age=20"
```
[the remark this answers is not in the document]
jascks: Done

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/controller/GenericControllerAdvice.java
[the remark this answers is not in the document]
jascks: Anyway, with that caveat, this can be implemented, but mistakes could be made. Still which to proceed?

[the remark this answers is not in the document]
steven-sheehy: That should be fixed then regardless of cors header. If you override header A you should not also have to override B-Z. We can add a `PostConstruct` on the properties and merge() defaults into path map.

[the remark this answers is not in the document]
jascks: Done. Default properties are inherited by each defined path, with the ability to override. Header names are treated in a case insensitive manner so that multiple headers of the same spelling but different case cannot be defined.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/controller/ControllerTest.java
jascks: This is not necessary. The properties are injected into `ControllerTest` in addition to the REST Java application. Therefore, given the test subclasses, what they return via `getUrl()` is used to look up the configured header names and values expected for the returned request mapping information. It then asserts those are present within the response headers returned by the API endpoint. Therefore, as `application.yml` is updated over time, existing tests will not break.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/RestJavaPropertiesTest.java
```diff
@@ -0,0 +1,79 @@
...
+package com.hedera.mirror.restjava;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.Map;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
+
+class RestJavaPropertiesTest {
```
jascks, line 26: Looks like I need to update this...

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/controller/ControllerTest.java
steven-sheehy: The assumes there's no bug in the logic in properties class and that there's no mistype in the application.yml. It's possible this test just asserts empty list equals empty list. We don't have to add it to every API in the future, but we should have an integration test at controller level that verifies both an overridden and a non-overriden property. Hence why I suggested above. A single line assert in one test is a small price to pay for confidence it provides us.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/RestJavaProperties.java
```diff
@@ -28,4 +36,47 @@ public class RestJavaProperties {
 
     @Min(0)
     private long shard = 0L;
+
+    @NotNull
+    private ResponseConfig response = new ResponseConfig();
```
steven-sheehy, line 41: nit: Sort before shard

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/RestJavaProperties.java
```diff
@@ -28,4 +36,47 @@ public class RestJavaProperties {
...
+        for (var pathHeaders : response.headers.path.entrySet()) {
+            var mergedHeaders = Stream.concat(
+                            response.headers.defaults.entrySet().stream(), pathHeaders.getValue().entrySet().stream())
+                    .collect(Collectors.toMap(
+                            Entry::getKey,
+                            Entry::getValue,
+                            (v1, v2) -> v2,
+                            () -> new TreeMap<>(String.CASE_INSENSITIVE_ORDER)));
+
+            response.headers.path.put(pathHeaders.getKey(), mergedHeaders);
```
steven-sheehy, line 58:
```suggestion
            pathHeaders.setValue(mergedHeaders);
```

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/RestJavaProperties.java
```diff
@@ -28,4 +36,47 @@ public class RestJavaProperties {
...
+    public static class ResponseHeadersConfig {
+        @NotNull
+        private Map<String, String> defaults = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
+
+        @NotNull
+        private Map<String, Map<String, String>> path = new HashMap<>();
+
+        public Map<String, String> getHeadersForPath(String apiPath) {
+            return apiPath == null ? defaults : path.getOrDefault(apiPath, defaults);
+        }
```
steven-sheehy, line 80: This method isn't really needed anymore, right? apiPath can just be null and return null from map. Defaults are already merged into this.path. So callers can just invoke .getPath().get(apiPath).

steven-sheehy (changes requested):

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/RestJavaProperties.java
jascks: But not all paths have headers defined (e.g. NFT allowances), in which case `.getPath().get(apiPath)` will return null and then the code will have to know to go with the default headers, and then access those. Something like:
```java
var responseHeaders = properties.getResponse().getHeaders().getPath().getOrDefault(requestMapping, properties.getResponse().getHeaders().getDefaults());
```

mgoelswirlds (changes requested):

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/RestJavaProperties.java
jascks: Done

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/RestJavaProperties.java
jascks: Done

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/RestJavaPropertiesTest.java
jascks: Done

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/controller/ControllerTest.java
jascks: Done

docs/configuration.md
```diff
@@ -600,16 +600,18 @@ to configure the application.
...
+| `hedera.mirror.restJava.db.statementTimeout`       | 10000                 | The number of milliseconds to wait before timing out a query statement                                                                                 |
+| `hedera.mirror.restJava.db.username`               | mirror_rest_java      | The username used to connect to the database                                                                                                           |
+| `hedera.mirror.restJava.response.headers.defaults` | See application.yml   | The default headers to add to every response. For each header, specify its `name: value`                                                               |
+| `hedera.mirror.restJava.response.headers.path`     | See application.yml   | Override default headers per path to add to every response. The key is the controller request mapping, then for each header, specify its `name: value` |
+| `hedera.mirror.restJava.shard`                     | 0                     | The default shard number that this mirror node participates in                                                                                         |
```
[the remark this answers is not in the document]
jascks: @mgoelswirlds , do you still see this formatting problem? It looks all nice and aligned now to me.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/RestJavaProperties.java
jascks: Admittedly, it is only `GenericControllerAdvice` that invokes this method in the real world. Keeping it somewhat ignorant about. how things operate behind the scene seems like a good idea still to me.

steven-sheehy (approved):
LGTM

mgoelswirlds (approved):
LGTM

PR #8552, opened 2024-06-14
Updated importer stream bucket read credentials in testnet-eu

nirbosl:
**Description**:
- Replaced `accessKey` and `secretKey` for both AWS and GCP buckets in `testnet-eu` and `testnet-citus`.

**Related issue(s)**:
https://github.com/swirlds/infrastructure/issues/6195

**Notes for reviewer**:
Moving away from date-versioned user account names to: `mirrornode-testnet-stream-ro` in order to avoid mistakes such as the deletion of users that owned the keys used by testnet importers.

**Checklist**

- [x] Keys verified by setting up a local importer and providing its config with the new keys.


jnels124 (commented):
Changing the sealed secret doesn't cause a deploy?

steven-sheehy:
> Changing the sealed secret doesn't cause a deploy?

It does cause a helmrelease reconciliation

steven-sheehy (approved):
LGTM

xin-hedera (approved):
LGTM

PR #8592, opened 2024-06-20
Increase delay duration to avoid pod scheduling issue

### LINKED ISSUE #8591: Potential HPA resource contention running k6 perf test suite
### Problem

Background: 

- In the k6 test suite, there is a delay 60 secs step between rest/restjava and restjava/web3
- HPAs for rest/restjava/web3 all have a default stabilization window of 300 secs for scaling down, which means, even though the current desired number of replicas would require the replicaset to scale down sharply, it would not happen or scale down to the lower number if there are previous high numbers in the past 300 secs window

 Combined with the max number of compute nodes for the node pool, it's possible that the ramp up stage for restjava/web3 will have a hard time to make the corresponding HPA to increase the restjava/web3 pods because there are still too many rest/restjava pods taking up resources in the 300s stabilization window.

### Solution

- increase the delay step duration
- or decrease HPA scale down stabilization window

### Alternatives

_No response_

xin-hedera:
**Description**:

This PR addresses the ramp up pod scheduling issue during k6 perf tests

- Increase delay duration between test steps 

**Related issue(s)**:

Fixes #8591 

**Notes for reviewer**:

As mentioned in the ticket, the default HPA scale down stabilization window is 300s. Setting the delay to 330s would give enough time to completely scale down the deployment tested against in the previous test step to its min replicas.

**Checklist**

### DIFF
--- a/charts/hedera-mirror-common/templates/test-suite-rest.yaml
+++ b/charts/hedera-mirror-common/templates/test-suite-rest.yaml
@@ -23,19 +23,19 @@ spec:
   {{- end }}
   steps:
     - execute:
-        - delay: 10s
+        - delay: {{ eq $idx 0 | ternary "10s" $.Values.testkube.test.delay }}
       stopOnFailure: false
     - execute:
         - test: test-rest-{{ $target.namespace }}
       stopOnFailure: false
     - execute:
-        - delay: 60s
+        - delay: {{ $.Values.testkube.test.delay }}
       stopOnFailure: false
     - execute:
         - test: test-rest-java-{{ $target.namespace }}
       stopOnFailure: false
     - execute:
-        - delay: 60s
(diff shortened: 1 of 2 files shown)


charts/hedera-mirror-common/templates/test-suite-rest.yaml
```diff
@@ -23,19 +23,19 @@ spec:
   {{- end }}
   steps:
     - execute:
-        - delay: 10s
+        - delay: {{ eq $idx 0 | ternary "10s" $.Values.testkube.test.delay }}
```
xin-hedera, line 26: the second test suite is triggered by completion of test-web3 in the first suite, so just for the second suite we need to set the initial delay longer as well.

jascks (approved):
LGTM

steven-sheehy (approved):

PR #8595, opened 2024-06-20
Update contract log query for Citus performance

### LINKED ISSUE #8573: Address performance Decrease on Contract Result Logs by Contract Id In V2
### Problem

```
2024-06-17T14:41:36.576Z ERROR dfeb6ef1 ::ffff:10.4.0.5 GET /api/v1/contracts/1077627/results/logs?limit=100 in 20473 ms: 503 Service unavailable DbError: timeout exceeded when trying to connect
    at Pool.queryQuietly (file:///home/node/app/utils.js:1413:13)
    at async RecordFileService.getRows (file:///home/node/app/service/baseService.js:69:20)
    at async RecordFileService.getRecordFileBlockDetailsFromTimestampArray (file:///home/node/app/service/recordFileService.js:107:18)
    at async ContractService.getContractLogs (file:///home/node/app/service/contractService.js:431:27)
    at async getContractLogsById (file:///home/node/app/controllers/contractController.js:766:18)
    at async wrappedMiddleware (/home/node/app/node_modules/@awaitjs/express/index.js:64:9)
```

### Solution

Address bottleneck and obtain performance comparable to v1

### Alternatives

_No response_

### LINKED ISSUE #8574: Citus Optimize Contract Results Logs List Endpoint
### Problem

`/contracts/results/logs?limit=100` is non performant in v2. We need to optimize this query.

Related to #8573

### Solution

determine bottleneck and address to be comparable to v1

### Alternatives

_No response_

edwin-greene:
**Description**:
Updates the contract log query to increase performance in V2

**Related issue(s)**:

Fixes #8573
Fixes #8574 

**Notes for reviewer**:
Prior explain analyze for query with data from mainnet-staging: 
`Execution Time: 2149.675 ms`
Explain analyze for the updated query:
`Execution Time: 4.447 ms`

**Checklist**


hedera-mirror-rest/service/recordFileService.js
```diff
@@ -100,20 +95,41 @@ class RecordFileService extends BaseService {
    * Retrieves the recordFiles containing the transactions of the given timestamps
    *
    * @param {(string|Number|BigInt)[]} timestamps consensus timestamp array
+   * @param {string} timestampOrder order of the timestamps, asc or desc
    * @return {Promise<Map>} A map from the consensus timestamp to its record file
    */
-  async getRecordFileBlockDetailsFromTimestampArray(timestamps) {
+  async getRecordFileBlockDetailsFromTimestampArray(timestamps, timestampOrder) {
```
xin-hedera, line 101: we can require the timestamps are always ordered, just it can be either asc or desc. We don't need an explicit parameter for the order, since we can just find it from the values in the `timestamps` array.

hedera-mirror-rest/service/recordFileService.js
```diff
@@ -100,20 +95,41 @@ class RecordFileService extends BaseService {
...
-      'getRecordFileBlockDetailsFromTimestampArray'
-    );
-
-    rows.forEach((row) => {
-      const recordFile = row.record_file ? new RecordFile(row.record_file) : null;
-      row.timestamps.forEach((timestamp) => recordFileMap.set(timestamp, recordFile));
-    });
+    if (!timestampOrder) {
+      timestampOrder = orderFilterValues.ASC;
+    }
```
xin-hedera, line 105: such logic is a great spot for bugs as callers can miss that if no order is passed, the function assumes it's asc.

hedera-mirror-rest/service/recordFileService.js
```diff
@@ -100,20 +95,41 @@ class RecordFileService extends BaseService {
...
+
+    let query = RecordFileService.recordFileBlockDetailsFromTimestampArrayQuery;
+    query += ` order by consensus_end ${timestampOrder}`;
+
+    const rows = await super.getRows(query, [timestamps], 'getRecordFileBlockDetailsFromTimestampArray');
+
+    rows
```
xin-hedera, line 112: it's better to not mix functional programming and normal for loop. so I'd prefer the following

```javascript
let index = 0;
for (const row in rows) {
  const recordFile = new RecordFile(row);
  const {consensusEnd, consensusStart} = recordFile;
  for (; index < timestamps.length; index++) {
    const timestamp = timestamps[index];
    if (consensusEnd >= timestamp && consensusStart <= timestamp) {
      recordFileMap.set(timestamp, recordFile);
    } else if ((timestampOrder == orderFilterValues.ASC && timestamp > consensusEnd) ||
            (timestampOrder == orderFilterValues.DESC && timestamp < consensusStart)) {
      break;
    }
  }
}
```

xin-hedera (changes requested):

hedera-mirror-rest/service/recordFileService.js
edwin-greene: Thanks I have removed the parameter and included logic to get the order from the timestamps

hedera-mirror-rest/service/recordFileService.js
edwin-greene: Thanks this part has been removed.

jnels124 (dismissed):
LGTM

hedera-mirror-rest/service/recordFileService.js
```diff
@@ -171,6 +181,17 @@ class RecordFileService extends BaseService {
...
     return row ? new RecordFile(row) : null;
   }
+
+  getTimestampOrder(timestamps) {
+    for (let i = 0; i < timestamps.length - 1; i++) {
+      if (timestamps[i] == timestamps[i + 1]) {
+        continue;
+      }
+      return timestamps[i] < timestamps[i + 1] ? orderFilterValues.ASC : orderFilterValues.DESC;
+    }
```
xin-hedera, line 191: not sure if timestamps can be empty, but not bad to be a bit defensive here. Also, we don't need a for loop.

```
if (timestamps.length === 0) return ASC;
const first =timestamps[0];
const last = timestamps[timestamps.length - 1];
return first > last ? DESC : ASC;
```

xin-hedera (changes requested):

hedera-mirror-rest/service/recordFileService.js
edwin-greene: Excellent thanks I have updated this.

xin-hedera (approved):
LGTM

steven-sheehy (approved):

PR #8650, opened 2024-06-26
Refactor the way contract action index is tracked in `OpcodeTracer`

### LINKED ISSUE #8639: Refactor the way contract action index is tracked in `OpcodeTracer`
### Problem

Currently the way that the `contractActionIndexOfCurrentFrame` is tracked inside `OpcodeTracer` is not exactly how it is done in the `HederaTracer`, where the list of contract actions is built and we are still using the mono code in the `web3` module, so we need to refactor this to be identical to what is done in `HederaTracer`. We need to copy the logic from there and increment the `contractActionIndexOfCurrentFrame` in the places where items are pushed to the `allActions` list in `HederaTracer`.

### Solution

Copy the way contract actions are added to the `allActions` list in [HederaTracer](https://github.com/hashgraph/hedera-services/blob/ce5f07658a088a968ca56a03614d034ca648e3e1/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/contracts/execution/traceability/HederaTracer.java)
- Looks like actions are generated in `init`, `tracePrecompileResult`, `traceAccountCreationResult`, and `tracePostExecution`
- We will need to add this interface - [HederaOperationTracer](https://github.com/hashgraph/hedera-services/blob/ce5f07658a088a968ca56a03614d034ca648e3e1/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/contracts/execution/traceability/HederaOperationTracer.java) 
- And also copy the needed logic from [CustomMessageCallProcessor](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/processors/CustomMessageCallProcessor.java#L239), which handles the precompile calls, updates the state of the message frames, sets an exceptional halt reason and also calls the tracePrecompileResult method from the [HederaOperationTracer](https://github.com/hashgraph/hedera-services/blob/ce5f07658a088a968ca56a03614d034ca648e3e1/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/contracts/execution/traceability/HederaOperationTracer.java)  interface.
- Test that there are no regressions in the `eth_call`, `eth_estimateGas` and `debug_traceTransaction` methods
- Verify that `OpcodeTracer` returns the correct revert reason for reverting HTS calls by taking the right contract action with the index that is tracked:
  - try to locally deploy a test contract which does different nested contract calls, some of which are reverting HTS calls, and then hit the `contract/results/{transactionIdOrHash}/opcodes` endpoint to replay the transaction and debug the `OpcodeTracer` to verify that the `contractActionIndexOfCurrentFrame` is correct when we hit the `tracePrecompileCall` method for the HTS calls

### Alternatives

_No response_

victor-yanev:
**Description**:

**Problem**

Currently the way that the `contractActionIndexOfCurrentFrame` is tracked inside `OpcodeTracer` is not exactly how it is done in the `HederaTracer`, where the list of contract actions is built and we are still using the mono code in the `web3` module, so we need to refactor this to be identical to what is done in `HederaTracer`. We need to copy the logic from there and increment the `contractActionIndexOfCurrentFrame` in the places where items are pushed to the `allActions` list in `HederaTracer`.

**This PR**

Copies the way contract actions are added to the `allActions` list in [HederaTracer](https://github.com/hashgraph/hedera-services/blob/ce5f07658a088a968ca56a03614d034ca648e3e1/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/contracts/execution/traceability/HederaTracer.java)

- Increments the index in `init`, `tracePrecompileResult`, `traceAccountCreationResult`, and `tracePostExecution`
- Adds the [HederaOperationTracer](https://github.com/hashgraph/hedera-services/blob/ce5f07658a088a968ca56a03614d034ca648e3e1/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/contracts/execution/traceability/HederaOperationTracer.java) interface
- Copies the needed logic from which handles the precompile calls, updates the state of the message frames, sets an exceptional halt reason and also calls the `tracePrecompileResult` method from the `HederaOperationTracer` interface.

**Manual testing done:**
- Verified that there are no regressions in the `eth_call`, `eth_estimateGas` and `debug_traceTransaction` methods
- Verified that `OpcodeTracer` returns the correct `reason` for reverting `HTS` calls from the appropriate contract action:
  - locally deployed a test contract which does different nested contract calls, some of which are reverting HTS calls
  - hit the `contract/results/{transactionIdOrHash}/opcodes` endpoint to replay the transaction
  - debugged the `OpcodeTracer` to verify that the `contractActionIndexOfCurrentFrame` is correct when we hit the `tracePrecompileCall` method for the HTS calls

**Related issue(s)**:

Fixes #8639

**Notes for reviewer**:

**Checklist**


IvanKavaldzhiev (dismissed):
LGTM

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/OpcodeTracer.java
```diff
@@ -155,24 +185,62 @@ private Map<Bytes, Bytes> captureStorage(final MessageFrame frame, OpcodeTracerO
...
+     * @param revertReason the revert reason
+     * @return the formatted revert reason
+     */
+    private Bytes formatRevertReason(final Bytes revertReason) {
+        if (revertReason == null || revertReason.isZero()) {
+            return Bytes.EMPTY;
+        }
+
+        Bytes trimmedReason = revertReason.trimLeadingZeros();
+        if (trimmedReason.size() <= 4) {
```
bilyana-gospodinova, line 236: Perhaps this value needs to be defined as a constant so that the code is more readable.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/convert/BytesDecoder.java
```diff
@@ -47,4 +47,16 @@ public static String maybeDecodeSolidityErrorStringToReadableMessage(final Bytes
         }
         return StringUtils.EMPTY;
     }
+
+    public static Bytes getAbiEncodedRevertReason(final Bytes revertReason) {
+        if (revertReason.toHexString().startsWith(ERROR_SIGNATURE)) {
```
bilyana-gospodinova, line 52: The "signature" name is a little confusing as the same terminology in services is related to keys, cryptography and verifications. Can you please rename the constant? My suggestion would be something like "REVERT_REASON_PREFIX" or "ERROR_STRING_PREFIX" or something like that.

victor-yanev: I think `ERROR_FUNCTION_SELECTOR` would be most correct since it decodes to `Error(string)`: https://besu.hyperledger.org/23.10.2/private-networks/how-to/send-transactions/revert-reason#revert-reason-format

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/convert/BytesDecoder.java
```diff
@@ -47,4 +47,16 @@ public static String maybeDecodeSolidityErrorStringToReadableMessage(final Bytes
...
+            return revertReason;
+        }
+
+        String revertReasonPlain = new String(revertReason.toArray());
+
+        return Bytes.concatenate(
```
bilyana-gospodinova, line 58: It seems to me that too many type transformations are being made. We call `formatRevertReason` in `OpcodeTracer#getRevertReasonFromContractActions`. There we have `action.getResultData()` as byte array. Then we transform it to Bytes, to hex string, to byte array, to String and we return it as Bytes. Can you please think if there is a way to reduce this amount?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/OpcodeTracer.java
victor-yanev: Changed it to use `Integer.BYTES`

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/convert/BytesDecoder.java
bilyana-gospodinova: Thanks, that's better. However, when the `revertReason` is null the Bytes concatenation in `getAbiEncodedRevertReason` method will produce the same result each time and I think we can simplify that by defining it as a constant and directly returning it. This way we can start with the null check and directly return this value and skip all the null checks after that.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/traceability/OpcodeTracer.java
```diff
@@ -155,24 +185,61 @@ private Map<Bytes, Bytes> captureStorage(final MessageFrame frame, OpcodeTracerO
...
+     * @param revertReason the revert reason
+     * @return the formatted revert reason
+     */
+    private Bytes formatRevertReason(final Bytes revertReason) {
+        if (revertReason == null || revertReason.isZero()) {
+            return Bytes.EMPTY;
+        }
+
+        Bytes trimmedReason = revertReason.trimLeadingZeros();
+        if (trimmedReason.size() <= Integer.BYTES) {
```
bilyana-gospodinova, line 236: Can you please add a comment what would this int contain?

victor-yanev: I thought it's evident enough from the line below:
```java
ResponseCodeEnum responseCode = ResponseCodeEnum.forNumber(trimmedReason.toInt());
```

bilyana-gospodinova: Yes, but you have to analyze the code in order to understand it. At first glance, it seems that the reason needs to be an int. However, it seems that it is fine if it isn't, because the reason can be formatted successfully on line 243 as well. If the reader clicks on the provided link at line 226, they will understand, but the link can be moved or become invalid at some point so my suggestion was to add a comment describing the reason format.

victor-yanev: Yeah, that actually covers an edge case where the revert reason in the contract actions contains a response code number in a plain string and:
1. it’s not in the valid format described in besu’s documentation
2. It contains only a number which doesn’t speak anything to the end user

so we convert this to an ABI-encoded string of the response code name, to give some relevant information, at least, if we hit this edge case.

victor-yanev: You are right, I will add a comment that explains this above those lines. Otherwise, its not really obvious why we are doing this.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/AbstractEvmMessageCallProcessor.java
```diff
@@ -37,11 +41,31 @@
...
+    @Override
+    public void start(final MessageFrame frame, final OperationTracer operationTracer) {
+        super.start(frame, operationTracer);
+
+        // potential precompile execution will be done after super.start(),
+        // so trace results here
+        final var contractAddress = frame.getContractAddress();
+        if ((isNativePrecompileCheck.test(contractAddress) || hederaPrecompiles.containsKey(contractAddress))
+                && operationTracer instanceof HederaOperationTracer hederaOperationTracer) {
+            hederaOperationTracer.tracePrecompileResult(
```
steven-sheehy, line 63: What's the advantage of casting tracer and using `tracePrecompileResult()` instead of just implementing `tracePrecompileCall()` which will have the result as a parameter? The logic for the latter is already in the parent class.

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/contracts/execution/AbstractEvmMessageCallProcessor.java
victor-yanev: Because if we call it here, we would modify the behaviour of the processor to call `tracePrecompileCall()` twice in different places instead of once, which we don't want to do.

This code around `tracePrecompileResult()` is copied from the mono code in services in order to replicate 1:1 the flow in which `contractActions` are built and instead of building the contract action, we just increment the `contractActionIndexOfCurrentFrame` in the same exact places to be able to track which contract action we need to get from the list.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/common/ContractCallContext.java
```diff
@@ -135,8 +135,4 @@ public boolean useHistorical() {
     public void incrementContractActionsCounter() {
         this.contractActionIndexOfCurrentFrame++;
```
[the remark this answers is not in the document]
victor-yanev: Simply getting the opcodes size doesn't work because we don't have a 1:1 mapping between opcodes and contract actions

steven-sheehy (approved):
LGTM

bilyana-gospodinova (approved):
LGTM

PR #8684, opened 2024-07-03
Support unlimited max automatic token associations in web3

### LINKED ISSUE #8274: HIP-904 Add support to Web3 for unlimited Max Automatic Token Associations
### Problem

Add support to Web3 for unlimited Max Automatic Token Associations. 

### Solution

Update Web3 `CryptoOpsUsage` and `SyntheticTxnFactory` to handle `-1` values for `maxAutomaticTokenAssociations` for protobuf versions greater than or equal to 0.51. If the HAPI protobuf version is less than that version then the methods should use the existing behavior.

### Alternatives

_No response_

xin-hedera:
**Description**:

This PR adds unlimited max automatic token associations support in web3

- Separate `maxAutoAssociations` and `usedAutoAssociations` in `Account` model to allow -1
- Set `maxAutoAssociations` to -1 for auto created accounts when HAPI >= 0.52.0

**Related issue(s)**:

Fixes #8274 

**Notes for reviewer**:

Confirmed by checking services code that all auto created accounts will have -1 as max automatic token associations, regardless of if it's triggered by a token transfer or hbar transfer.

**Checklist**


hedera-mirror-web3/src/main/java/com/hedera/services/fees/calculation/utils/OpUsageCtxHelper.java
```diff
@@ -105,7 +105,7 @@ public ExtantCryptoContext ctxForCryptoAllowance(TxnAccessor accessor) {
                     .setCurrentExpiry(account.getExpiry())
                     .setCurrentlyHasProxy(account.getProxy() != null)
                     .setCurrentNumTokenRels(account.getNumAssociations())
-                    .setCurrentMaxAutomaticAssociations(account.getMaxAutomaticAssociations())
+                    .setCurrentMaxAutomaticAssociations(account.getMaxAutoAssociations())
```
steven-sheehy, line 108: Why rename it? We should revert to minimize changes and drift from upstream.

xin-hedera: modular code has similar changes since the field in protobuf state type `Account` is named `max_auto_associations`.

Ofc we can also add an additional getter to `Account` model and revert the changes here.

IvanKavaldzhiev (approved):
LGTM!

Great work and nice clean up of the code base! 🥂

hedera-mirror-web3/src/main/java/com/hedera/services/fees/calculation/utils/OpUsageCtxHelper.java
steven-sheehy: Ok, not a big deal either way.

IvanKavaldzhiev: We're in a phase where we don't have an existing upstream to refer to, since we were based on the mono code. Any future changes to the existing web3 will be a mixture of the old mono code and important and necessary changes coming from the mod code.

So until we integrate the `hedera-smart-contract-service` and the `HandleContext` in web3, I think it would be acceptable to make some changes and produce a mixed code base, as far as we have no regressions and allow new futures and enhancements in the web3 module.

steven-sheehy (approved):
LGTM

Issue #8682, opened 2024-07-01
`contract/call` for contract deployment throw 500 error if `from` field is omitted


quiet-node:
### Description

For a `contract/call` request related to contract deployment, omitting the from field results in a 500 - Internal Server Error. But the from field is not required for other types of contract/call requests that are not related to contract deployment.

### Steps to reproduce

```bash
curl -X 'POST' \
  'https://testnet.mirrornode.hedera.com/api/v1/contracts/call' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "data": "0x6080604052348015600f57600080fd5b50609e8061001e6000396000f3fe608060405260043610602a5760003560e01c80635c36b18614603557806383197ef014605557600080fd5b36603057005b600080fd5b348015604057600080fd5b50600160405190815260200160405180910390f35b348015606057600080fd5b50606633ff5b00fea2646970667358221220886a6d6d6c88bcfc0063129ca2391a3d98aee75ad7fe3e870ec6679215456a3964736f6c63430008090033",
  "estimate": false,
  "from": "00000000000000000000000000000000000004e2",
  "to": null
}'
```
This will work fine.

But if the `from` is omitted,
```bash
curl -X 'POST' \
  'https://testnet.mirrornode.hedera.com/api/v1/contracts/call' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "data": "0x6080604052348015600f57600080fd5b50609e8061001e6000396000f3fe608060405260043610602a5760003560e01c80635c36b18614603557806383197ef014605557600080fd5b36603057005b600080fd5b348015604057600080fd5b50600160405190815260200160405180910390f35b348015606057600080fd5b50606633ff5b00fea2646970667358221220886a6d6d6c88bcfc0063129ca2391a3d98aee75ad7fe3e870ec6679215456a3964736f6c63430008090033",
  "estimate": false,
  "to": null
}'
```
It throws 
```bash
{
  "_status": {
    "messages": [
      {
        "message": "Internal Server Error",
        "detail": "",
        "data": ""
      }
    ]
  }
}
```

### Hedera network

mainnet, testnet, previewnet, other

### Version

latest

### Operating system

None

steven-sheehy:
Logs:

```
[...]

PR #8743, opened 2024-07-10
Fix NullPointerException when from field is null on contract create

bilyana-gospodinova:
**Description**

Fix NullPointerException when from field is null on contract create

Fixes https://github.com/hashgraph/hedera-mirror-node/issues/8682

**Checklist**

### DIFF
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/ContractController.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/ContractController.java
@@ -90,7 +90,7 @@ class ContractController {
         // If both fields are missing, we are in a contract create case, but no sender to subtract the fee from, so the
         // request should fail.
         if (fromAddress.isZero() && receiver.isZero()) {
-            throw new InvalidParametersException("to field must not be empty");
+            throw new InvalidParametersException("from field must not be empty");
         }
 
         Bytes data;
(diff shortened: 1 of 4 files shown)


zhpetkov (approved):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/ContractController.java
```diff
@@ -86,6 +86,13 @@ private ContractExecutionParameters constructServiceParameters(ContractCallReque
         } else {
             receiver = Address.fromHexString(request.getTo());
         }
+
+        // If both fields are missing, we are in a contract create case, but no sender to subtract the fee from, so the
+        // request should fail.
+        if (fromAddress.isZero() && receiver.isZero()) {
+            throw new InvalidParametersException("from field must not be empty");
```
IvanKavaldzhiev, line 93: We should allow contract creation even with missing "from" account. Probably we can do the same trick as for the recipient and set the sender to be the zero address?

I think the only case where the "from" account cannot be with zero address is during a transfer.

bilyana-gospodinova: The sender is set to the zero address some lines above in the same method. However, the exception is: 
`java.lang.NullPointerException: Cannot invoke "org.hyperledger.besu.evm.account.MutableAccount.decrementBalance(org.hyperledger.besu.datatypes.Wei)" because "sender" is null. `
The message is kind of misleading, but the point is that the zero address is not usable (null or Address.ZERO or default account) so we can't decrement the balance successfully and we get the exception above.

IvanKavaldzhiev (approved):
LGTM!

kselveliev (approved):

zhpetkov (approved):

xin-hedera (approved):
LGTM

PR #8746, opened 2024-07-10
Optimize transaction_hash to support querying by the 32-byte prefix

### LINKED ISSUE #8701: Research feasibility of query contract logs by transaction hash
### Problem

Synthetic contract results were never enabled as they were a performance concern during ingest. To unblock oracles, they need a way to access synthetic contract logs by hash.

### Solution

* See https://github.com/hashgraph/hedera-mirror-node/pull/8649 for API side changes
* Research feasibility of adding a search by transaction hash to `/api/v1/contracts/results/logs` for both v1 and v2 schema.
* If it's easy enough, add the necessary schema and importer changes.

### Alternatives

_No response_

xin-hedera:
**Description**:

This PR changes `transaction_hash` table to support fast search by the leading 32-byte of a native transaction hash or the 32-byte ethererum transaction hash.

- Add migration to change `transaction_hash` to hash index the leading 32-byte of the hash column
- Add ethereum transaction hash info backfill to `BackfillTransactionHashMigration` with a configurable strategy
- Bump up `BackfillTransactionHashMigration`checksum
- Change rest api transaction by hash query to utilize the new index

**Related issue(s)**:

Fixes #8701 

**Notes for reviewer**:

Time to run the migrations in staging clone db:

```
mirror_node=> drop index if exists transaction_hash_sharded_hash_idx;
DROP INDEX
Time: 284.780 ms
mirror_node=> create index transaction_hash__hash on transaction_hash using hash (substring(hash from 1 for 32));
CREATE INDEX
Time: 528851.245 ms (08:48.851)
mirror_node=> insert into transaction_hash (consensus_timestamp, hash, payer_account_id) select consensus_timestamp, hash, payer_account_id from ethereum_transaction;
INSERT 0 95958
Time: 3812.355 ms (00:03.812)
```

query performance:
```
mirror_node=> select hash from ethereum_transaction limit 3;
                                hash
--------------------------------------------------------------------
 \x960b6915572c7d58d336e0631b81d2f1788daf6209d95b677ae40744a6f6b352
 \xb2a6180fbbd035880e59928d13b076bb5bc67162088fe797ef39ecba8f28461f
 \xb2a6180fbbd035880e59928d13b076bb5bc67162088fe797ef39ecba8f28461f
(3 rows)

Time: 0.853 ms

mirror_node=> select * from transaction_hash_18 where substring(hash from 1 for 32) = '\xb2a6180fbbd035880e59928d13b076bb5bc67162088fe797ef39ecba8f28461f'::bytea;
 consensus_timestamp |                                hash                                | payer_account_id
---------------------+--------------------------------------------------------------------+------------------
 1680797038770270618 | \xb2a6180fbbd035880e59928d13b076bb5bc67162088fe797ef39ecba8f28461f |           995584
 1680797007526573292 | \xb2a6180fbbd035880e59928d13b076bb5bc67162088fe797ef39ecba8f28461f |           995584
(2 rows)

Time: 3.345 ms


mirror_node=> explain select * from transaction_hash_18 where substring(hash from 1 for 32) = '\xb2a6180fbbd035880e59928d13b076bb5bc67162088fe797ef39ecba8f28461f'::bytea;
                                                         QUERY PLAN
-----------------------------------------------------------------------------------------------------------------------------
 Index Scan using transaction_hash_18_substring_idx on transaction_hash_18  (cost=0.00..2.22 rows=1 width=65)
   Index Cond: (SUBSTRING(hash FROM 1 FOR 32) = '\xb2a6180fbbd035880e59928d13b076bb5bc67162088fe797ef39ecba8f28461f'::bytea)
(2 rows)

Time: 2.689 ms

mirror_node=> explain analyze select * from get_transaction_info_by_hash('\xb2a6180fbbd035880e59928d13b076bb5bc67162088fe797ef39ecba8f28461f'::bytea);
                                                           QUERY PLAN
--------------------------------------------------------------------------------------------------------------------------------
 Function Scan on get_transaction_info_by_hash  (cost=0.25..10.25 rows=1000 width=48) (actual time=5.464..5.465 rows=2 loops=1)
 Planning Time: 0.042 ms
 Execution Time: 5.495 ms
(3 rows)

Time: 6.423 ms
```
**Checklist**


hedera-mirror-rest/__tests__/specs/transactions/{id}/no-params.json
```diff
@@ -33,6 +33,34 @@
     ],
     "balances": [],
     "transactions": [
+      {
+        "payerAccountId": "0.0.10",
+        "nodeAccountId": "0.0.3",
+        "consensus_timestamp": "1234567800000000002",
+        "valid_start_timestamp": "1234567800000000001",
+        "name": "CRYPTOTRANSFER",
+        "transaction_hash": "rovr8cn6DzCTVuSAV/YEevfN5jA30FCdFt3Dsg4IUVi/3xTRU0XBsYsZm3L+1N6t",
```
xin-hedera, line 42: the transaction's hash has the same leading 32 bytes as the query target, however the last few bytes are different. It's to test that the extra filtering in js code works as intended to remove such transactions.

hedera-mirror-rest/transactionHash.js
```diff
@@ -0,0 +1,58 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {ETH_HASH_LENGTH} from './constants';
+import {TransactionHash} from './model';
+
+const transactionHashQuery = `
```
xin-hedera, line 20: move the logic into its own module so it can be reused by contract log by hash search

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.98.0__adjust_transaction_hash_index.sql
```diff
@@ -0,0 +1,19 @@
...
+      hash                bytea,
+      payer_account_id    bigint
+    )
+as
+$$
+declare
+shard varchar;
+begin
+  shard := concat('transaction_hash_', to_char(mod(get_byte($1, 0), 32), 'fm00'));
+  return query execute 'select * from ' || shard || ' where substring(hash from 1 for 32) = $1' using $1;
```
xin-hedera, line 17: replace the function so we have `where substring(hash from 1 for 32) = $1` to utilize the new index

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/BackfillEthereumTransactionHashMigration.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    @Override
+    protected void doMigrate() {
+        if (!entityProperties.getPersist().shouldPersistTransactionHash(TransactionType.ETHEREUMTRANSACTION)) {
+            log.info("Skipping migration since transaction hash persistence for ethereum transaction is disabled");
+            return;
+        }
+
+        var stopwatch = Stopwatch.createStarted();
```
xin-hedera, line 55: we can't truncate the table since it may have many existing rows. we can't do check if not exists then insert since it'll be very slow.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/BackfillEthereumTransactionHashMigration.java
```diff
@@ -0,0 +1,64 @@
...
+import com.google.common.base.Stopwatch;
+import com.hedera.mirror.common.domain.transaction.TransactionType;
+import com.hedera.mirror.importer.ImporterProperties;
+import com.hedera.mirror.importer.parser.record.entity.EntityProperties;
+import jakarta.inject.Named;
+import org.springframework.context.annotation.Lazy;
+import org.springframework.jdbc.core.JdbcTemplate;
+
+@Named
+public class BackfillEthereumTransactionHashMigration extends RepeatableMigration {
```
xin-hedera, line 28: need to research if this will deterministically run after `BackfillTransactionHashMigration` since otherwise inserted data will get truncated by the other migration.

jnels124: If this follows the same ordering mechanism as the repeatable sql migrations, this one would actually run first. The sql repeatable migrations are ordered by description.

jnels124 (dismissed):
LGTM

hedera-mirror-importer/src/main/resources/db/migration/v2/V2.3.0__adjust_transaction_hash_index.sql
```diff
@@ -0,0 +1,2 @@
+drop index if exists transaction_hash__hash;
+create index transaction_hash__hash on transaction_hash using hash (substring(hash from 1 for 32));
```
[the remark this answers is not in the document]
xin-hedera: I think the problem stopping us from doing so is we have two types of 32-byte to hash, one is truncated from the 48-byte native hash, the other is by itself 32-byte.

Then we can’t search by the first category because there is no way to expand a 32-byte hash to its original 48-byte value.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/BackfillEthereumTransactionHashMigration.java
xin-hedera: This migration is gone so no longer a concern :)

hedera-mirror-importer/src/main/resources/db/migration/v2/V2.3.0__adjust_transaction_hash_index.sql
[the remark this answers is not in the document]
xin-hedera: Oh I misread it.

It also has problem.. unless we think it’s worth to repopulate mainnet transaction_hash table…

hedera-mirror-rest/transactions.js
```diff
@@ -798,11 +766,13 @@ const extractSqlFromTransactionsByIdOrHashRequest = async (transactionIdOrHash,
       transactionIdOrHash = transactionIdOrHash.substring(2);
     }
 
-    const v1ShardQueryEnabled = await transactionHashShardedQueryEnabled();
-    const usedTransactionHashQuery = v1ShardQueryEnabled ? transactionHashShardedQuery : transactionHashQuery;
+    const query = await getTransactionHashQuery();
```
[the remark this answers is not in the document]
xin-hedera: Yeah I should encapsulate more logic so it also works for a 32-byte hash and thus easier for contract log by hash search.

hedera-mirror-rest/__tests__/integration/transactionHash.test.js
```diff
@@ -0,0 +1,26 @@
+
```
xin-hedera, line 1: will add the license header

steven-sheehy (approved):
LGTM

jnels124 (approved):
LGTM

PR #8760, opened 2024-07-12
Enable remaining services on mainnet-citus

jnels124:
enable chart components


jnels124:
should i enable monitor here too or is that going to cause alerts to go off if ingresses are off?

xin-hedera:
> should i enable monitor here too or is that going to cause alerts to go off if ingresses are off?

we can enable it with its ingress disabled if monitor's needed to silence some alerts. if the ingress is enabled then the google cloud load balancer may get confused because the health check returns UP when it hits the monitor in mainnet and DOWN when it hits mainnet-citus.

jascks (approved):
LGTM

clusters/mainnet-eu/mainnet-citus/helmrelease.yaml
```diff
@@ -72,9 +69,9 @@ spec:
             "stateproof": { "enabled": false }
```
xin-hedera, line 69: the rest monitor config will break all checks since it'll use the server list in the default config, i.e., assume the server listens on localhost

steven-sheehy: Basically remove `rest.monitor`

steven-sheehy (approved):

jascks (approved):
LGTM

xin-hedera (approved):
LGTM

PR #8789, opened 2024-07-16
Add token name query parameter to `/api/v1/tokens`

### LINKED ISSUE #8702: HIP-968: Add token name index
### Problem

To support HIP-968, we need an efficient way to query a token by name.

### Solution

* Research read and write performance of using a pg_trgm index for ILIKE searches of `token.name`
* Add index to v1 and v2 schema

### Alternatives

_No response_

### LINKED ISSUE #8703: HIP-968: Add a `name` query parameter to `/api/v1/tokens`
### Problem

Users would like to search by a token's name like in Etherscan.

### Solution

* Add a `name` query parameter to `/api/v1/tokens`
* Add a k6 test
* Add an acceptance test

### Alternatives

_No response_

jnels124:
**Description**:
* add pg_trgm gin index on `token.name`
* add token name parameter support to rest api
* add new k6 test for token name param on /api/v1/tokens
* update open api spec 

**Related issue(s)**:

Fixes #8702
Fixes #8703

**Notes for reviewer**:

**Checklist**


hedera-mirror-rest/tokens.js
```diff
@@ -161,6 +162,10 @@ const extractSqlFromTokenRequest = (query, params, filters, conditions) => {
       continue;
     }
 
+    if (filter.key === filterKeys.NAME) {
+      conditions.push(`t.name ILIKE $${params.push('%' + filter.value + '%')}`);
```
jnels124, line 166: multiple occurrences of the name parameter do a search for tokens that contain substrings from all occurrences

hedera-mirror-rest/tokens.js
```diff
@@ -314,6 +319,9 @@ const validateTokenQueryFilter = (param, op, val) => {
     case filterKeys.LIMIT:
       ret = utils.isPositiveLong(val);
       break;
+    case filterKeys.NAME:
+      ret = op === queryParamOperators.eq;
```
jnels124, line 323: should we limit the length of name param values?

steven-sheehy: Yes, token [names](https://github.com/hashgraph/hedera-protobufs/blob/54b258dfa27866d2654bb036112298b5a6693505/services/token_create.proto#L78) can't exceed 100 bytes

hedera-mirror-importer/src/main/resources/db/scripts/v2/migration.sh
jnels124, line 1: Small adjustment noticed when doing eu migration

hedera-mirror-common/src/test/resources/init.sql
```diff
@@ -31,6 +31,7 @@ revoke create on schema temporary from public;
 -- Add extensions
 create extension if not exists btree_gist;
 create extension if not exists pg_stat_statements;
+CREATE extension if not exists pg_trgm;
```
xin-hedera, line 34:
```suggestion
create extension if not exists pg_trgm;
```

also add it to https://github.com/hashgraph/hedera-mirror-node/blob/ab647eabbbab117f8f99d94fedac6c4519b922e0/hedera-mirror-rosetta/container/entrypoint.sh#L31

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.99.0__token_name_index.sql
```diff
@@ -0,0 +1 @@
+create index if not exists token__name on token using gin (name gin_trgm_ops);
```
xin-hedera, line 1: `V1.98.1` and it's nice to have a newline at the end of the file.

hedera-mirror-importer/src/main/resources/db/migration/v2/V2.4.0__token_name_index.sql
```diff
@@ -0,0 +1 @@
+create index if not exists token__name on token using gin (name gin_trgm_ops);
```
xin-hedera, line 1: `V2.3.1`

hedera-mirror-rest/utils.js
```diff
@@ -132,6 +132,10 @@ const isHexPositiveInt = (num, allowZero = false) => {
   return false;
 };
 
+const isMaxBytes = (str, maxSize) => {
+  return new Blob([str]).size <= maxSize;
```
xin-hedera, line 136: why can't we use `str.length`?

xin-hedera (changes requested):

hedera-mirror-rest/utils.js
jnels124: I wasn't sure if I can guarantee all chars will be utf-8 and single byte.

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.99.0__token_name_index.sql
mgoelswirlds: Might have to adjust this depending on which PR goes in first.

hedera-mirror-importer/src/main/resources/db/migration/v2/V2.3.1__token_name_index.sql
steven-sheehy, line 1: Please coordinate with Mugdha as I think you're both utilizing the same migration numbers. Whoever gets in last will need to doublecheck.

charts/hedera-mirror/values.yaml
```diff
@@ -306,6 +306,7 @@ stackgres:
       version: "12.1-1"
     - name: btree_gist
       version: stable
+    - name: pg_trgm
```
steven-sheehy, line 309: Is there some specific version or `stable` we should use?

hedera-mirror-rest/api/v1/openapi.yml
```diff
@@ -1074,6 +1074,7 @@ paths:
       operationId: getTokens
       parameters:
         - $ref: "#/components/parameters/accountIdQueryParam"
+        - $ref: "#/components/parameters/tokenNameQueryParam"
```
steven-sheehy, line 1077: Sort after limit so it shows up alphabetically by name

hedera-mirror-rest/tokens.js
```diff
@@ -161,6 +162,10 @@ const extractSqlFromTokenRequest = (query, params, filters, conditions) => {
       continue;
     }
 
+    if (filter.key === filterKeys.NAME) {
```
steven-sheehy, line 165: Should we disallow certain combination of query parameters like token.id+name or publickey+name, etc if they don't make sense or have performance impacts?

hedera-mirror-rest/tokens.js
```diff
@@ -1023,6 +1031,7 @@ const acceptedTokenParameters = new Set([
   filterKeys.ACCOUNT_ID,
   filterKeys.ENTITY_PUBLICKEY,
   filterKeys.LIMIT,
+  filterKeys.NAME,
```
steven-sheehy, line 1034: Please update the docs/rest/README.md with the new param and docs/database/README.md with the new index.

steven-sheehy (changes requested):

charts/hedera-mirror/values.yaml
jnels124: meant to fill this in. Wanted to see what the current default is.

hedera-mirror-rest/tokens.js
jnels124: in the case of public key, there can be many tokens tied to the same public key. We could exclude, but doesn't seem like much of a performance gain. Not sure how much benefit there is now or in the future to specify a public key that has many tokens and then filter by name. 

with name and public key
```
--------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=3750.43..3750.53 rows=40 width=113) (actual time=5.741..5.745 rows=0 loops=1)
   ->  Sort  (cost=3750.43..3750.53 rows=40 width=113) (actual time=5.740..5.743 rows=0 loops=1)
         Sort Key: t.token_id
         Sort Method: quicksort  Memory: 25kB
         ->  Hash Join  (cost=1531.69..3749.37 rows=40 width=113) (actual time=5.735..5.738 rows=0 loops=1)
               Hash Cond: (e.id = (t.token_id)::bigint)
               ->  Bitmap Heap Scan on entity e  (cost=190.32..2405.99 rows=762 width=44) (actual time=0.058..0.351 rows=765 loops=1)
                     Recheck Cond: ((public_key)::text = '421050820e1485acdd59726088e0e4a2130ebbbb70009f640ad95c78dd5a7b38'::text)
                     Heap Blocks: exact=128
                     ->  Bitmap Index Scan on entity__public_key_type  (cost=0.00..190.13 rows=762 width=0) (actual time=0.037..0.038 rows=765 loops=1)
                           Index Cond: ((public_key)::text = '421050820e1485acdd59726088e0e4a2130ebbbb70009f640ad95c78dd5a7b38'::text)
               ->  Hash  (cost=1308.03..1308.03 rows=2668 width=77) (actual time=5.269..5.270 rows=2410 loops=1)
                     Buckets: 4096  Batches: 1  Memory Usage: 293kB
                     ->  Bitmap Heap Scan on token t  (cost=32.68..1308.03 rows=2668 width=77) (actual time=0.420..4.666 rows=2410 loops=1)
                           Recheck Cond: ((name)::text ~~* '%bar%'::text)
                           Heap Blocks: exact=816
                           ->  Bitmap Index Scan on token__name  (cost=0.00..32.01 rows=2668 width=0) (actual time=0.298..0.298 rows=2410 loops=1)
                                 Index Cond: ((name)::text ~~* '%bar%'::text)
Planning Time: 0.495 ms
 Execution Time: 5.800 ms
```

with just public key
```
Limit  (cost=2443.49..3412.44 rows=100 width=113) (actual time=0.538..0.960 rows=100 loops=1)
   ->  Merge Join  (cost=2443.49..8373.45 rows=612 width=113) (actual time=0.537..0.951 rows=100 loops=1)
         Merge Cond: ((t.token_id)::bigint = e.id)
         ->  Index Scan using token_pkey on token t  (cost=0.29..5819.34 rows=40803 width=77) (actual time=0.012..0.364 rows=742 loops=1)
         ->  Sort  (cost=2442.46..2444.37 rows=762 width=44) (actual time=0.519..0.525 rows=100 loops=1)
               Sort Key: e.id
               Sort Method: quicksort  Memory: 85kB
               ->  Bitmap Heap Scan on entity e  (cost=190.32..2405.99 rows=762 width=44) (actual time=0.067..0.395 rows=765 loops=1)
                     Recheck Cond: ((public_key)::text = '421050820e1485acdd59726088e0e4a2130ebbbb70009f640ad95c78dd5a7b38'::text)
                     Heap Blocks: exact=128
                     ->  Bitmap Index Scan on entity__public_key_type  (cost=0.00..190.13 rows=762 width=0) (actual time=0.047..0.047 rows=765 loops=1)
                           Index Cond: ((public_key)::text = '421050820e1485acdd59726088e0e4a2130ebbbb70009f640ad95c78dd5a7b38'::text)
 Planning Time: 0.382 ms
 Execution Time: 1.017 ms
```

In the case of id, we could exclude it in the case of an `eq` op but not really in others. All that is saved in that case though is small filtering on the name as it will use the pk index `token_id`.

jnels124: on second thought, it may become expensive over time using this param and large range gte token.id queries. Maybe we should disallow this param with token.id queries.

hedera-mirror-rest/tokens.js
jnels124: for the rest docs, this param doesn't cause new tables to be accessed so the tables listed under the root endpoint /tokens is correct. Do you want to duplicate the row with the additional param? It appears we only add a separate entry when that param triggers an additional table to be used in the query

steven-sheehy: True, so just the database doc needs updating then.

hedera-mirror-rest/utils.js
```diff
@@ -132,6 +132,10 @@ const isHexPositiveInt = (num, allowZero = false) => {
   return false;
 };
 
+const isMaxBytes = (str, maxSize) => {
```
steven-sheehy, line 135: Should have a test

steven-sheehy (dismissed):
LGTM

jascks (dismissed):
LGTM

xin-hedera (approved):
LGTM

xin-hedera (approved):
LGTM

jascks (approved):
LGTM

jnels124:
sauce -->  271
sauce-quack --> 1
bar --> 2410 

V1
| tokensName | /tokens?limit=100&name=bar | 1500 | 100.00 | 1633.80/s | 1633.80/s | 810.77ms | No | |
| tokensName | /tokens?limit=100&name=sauce | 1500 | 100.00 | 1386.62/s | 1386.62/s | 414.89ms | No | |
| tokensName | /tokens?limit=100&name=QUACK | 1500 | 99.97 | 3293.49/s | 3292.50/s | 229.16ms | No | |

V2
| tokensName | /tokens?limit=100&name=bar | 1500 | 93.23 | 153.19/s | 142.82/s | 8801.91ms | No | |
| tokensName | /tokens?limit=100&name=sauce | 1500 | 99.88 | 337.50/s | 337.10/s | 4393.04ms | No | |
| tokensName | /tokens?limit=100&name=quack | 1500 | 99.88 | 305.27/s | 304.90/s | 4870.99ms | No | |

hedera-mirror-rest/api/v1/openapi.yml
```diff
@@ -4799,6 +4800,14 @@ components:
           value: lte:0.0.700
       schema:
         $ref: "#/components/schemas/EntityIdQuery"
+    tokenNameQueryParam:
+      name: name
+      in: query
+      description: Partial or full token name. Not allowed to be used with token.id or account.id parameter. Pagination is not supported with the use of this parameter and results are ordered by token.id with respect to the order parameter.
+      schema:
+        type: string
+        maxLength: 100
```
steven-sheehy, line 4809: Add a minLength and enforce it

steven-sheehy (changes requested):

hedera-mirror-rest/api/v1/openapi.yml
```diff
@@ -4807,6 +4807,7 @@ components:
       schema:
         type: string
         maxLength: 100
+        minLength: 5
```
steven-sheehy, line 4810: Probably this should be 3

jnels124: I can change that but more characters would be better. Searching for things like bar is not super efficient and will use the token pk index instead. That is why the k6 results are worse for searching with bar.

jnels124: With the other restrictions in place like no pagination and excluding use with other params, 3 should be fine.

steven-sheehy (approved):
LGTM

jascks (approved):
LGTM

xin-hedera (approved):
LGTM

PR #8799, opened 2024-07-18
Port REST specs to Java, phase 1

### LINKED ISSUE #6810: First phase port or REST specs to Java
### Problem

To confidently be able to port over existing JavaScript REST APIs to Java, we need to ensure the existing spec tests continue to work against the newly converted API.

### Solution

Phase 1 implementation of overall framework to deserialize, setup and run a subset of existing JS REST API specs.

* Add a `RestSpecTest` that
  * Injects and uses a `RestClient` similar to `ControllerTest`
  * Can load JSON spec files from `hedera-mirror-rest-java/src/test/resources/specs` with arbitrary depths
  * Model the spec JSON as a `RestSpec` POJO
  * For each spec, use `DomainBuilder` to create default and override with data from `setup` section and persist
  * Call the URLs and verify the response JSON and status codes
  * **Temporarily** add a `hedera-mirror-rest` test container to verify some subset of specs pass against it

### Alternatives

_No response_

jascks:
**Description**:
To confidently be able to port over existing JavaScript REST APIs to Java, we need to ensure the existing spec tests continue to work against the newly converted API.

* Add a `RestSpecTest` that
  * Injects and uses a `RestClient` similar to `ControllerTest`
  * Can load JSON spec files from `hedera-mirror-rest-java/src/test/resources/specs` with arbitrary depths. For this phase 1 PR, a selected few spec files are choosen.
  * Model the spec JSON as a `RestSpec` POJO
  * For each spec, use `DomainBuilder` to create default and override with data from `setup` section and persist
  * Call the URLs and verify the response JSON and status codes
  * **Temporarily** add a `hedera-mirror-rest` test container to verify all specs pass against it

**Related issue(s)**:

Fixes #6810

**Notes for reviewer**:
This branch/PR is a step back from the branch `6810-port-rest-specs-to-java` which contains additional functionality regarding additional DB entity types for setup. That will be the basis for follow on work to build on this phase 1 PR.

**Checklist**


hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/builder/AccountBuilder.java
```diff
@@ -0,0 +1,93 @@
...
+            "alias", BASE32_CONVERTER,
+            "evmAddress", HEX_OR_BASE64_CONVERTER,
+            "id", ENTITY_ID_CONVERTER,
+            "key", HEX_OR_BASE64_CONVERTER
+    );
+
+    AccountBuilder(EntityManager entityManager, TransactionOperations transactionOperations) {
+        super(entityManager, transactionOperations, METHOD_PARAMETER_CONVERTERS);
+    }
+
```
jascks, line 44: The general flow of these entity builders is based on `integrationDomainOps.js` in the JS REST API. In this case, start with the `addAccount()` function. In order to match the API response body defined in a spec, the same defaults, customizations and post-processing need to occur.

JS can take liberties with JSON and the attribute types that Java cannot. Thus the last 'check and finalize' step may invoke `wrapper.get()` to inspect the entity built thus far, and based on that, but invoke further builder methods to adjust things as needed prior to persisting the final entity.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/RestSpecTest.java
```diff
@@ -0,0 +1,131 @@
...
+                    JSONAssert.assertEquals(specResponseJson, response.getBody(), JSONCompareMode.LENIENT);
+                } catch (JSONException e) {
+                    log.warn("Failed to parse expected responseJson within spec file: {}", specFile, e);
+                }
+            }
+        }
+
+    }
+
+    private void setupDatabase(RestSpecNormalized normalizedRestSpec) {
```
jascks, line 127: This is a phase 1 implementation. Something more general to inspect the `SpectSetup` to determine the required database entities is planned.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/builder/TokenAccountBuilder.java
```diff
@@ -0,0 +1,68 @@
...
+import com.hedera.mirror.common.domain.DomainWrapperImpl;
+import com.hedera.mirror.common.domain.token.TokenAccount;
+import jakarta.inject.Named;
+import jakarta.persistence.EntityManager;
+import java.util.Map;
+import java.util.function.Function;
+import org.springframework.transaction.support.TransactionOperations;
+
+@Named
+class TokenAccountBuilder extends AbstractEntityBuilder {
```
jascks, line 29: Patterned after `integrationDomainUtils.js` function `addTokenAccount()`.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/config/SpecTestConfig.java
```diff
@@ -0,0 +1,85 @@
...
+    private boolean v2;
+
+    @Bean
+    Network postgresqlNetwork() {
+        return Network.newNetwork();
+    }
+
+    @Bean("postgresql")
+    @ServiceConnection("postgresql")
+    PostgreSQLContainer<?> postgresqlOverride(Network postgresqlNetwork) {
```
jascks, line 47: For `RestSpecTest`, which sets `spring.main.allow-bean-definition-overriding=true`, this bean overrides that defined in `CommonTestConfiguration`. It is basically the same configuration, except this postgresql container is placed on a network so that the RS REST API container (below) can connect to it.

Doing this directly with the existing container in `CommonTestConfiguration` caused other integration tests to fail.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/model/SpecTestNormalized.java
```diff
@@ -0,0 +1,68 @@
...
+                    : new ArrayList<>(specTest.urls());
+
+            mutableUrls.add(specTest.url());
+            urls = List.copyOf(mutableUrls);
+        }
+        else {
+            urls = specTest.urls() == null ? List.of() : List.copyOf(specTest.urls());
+        }
+
+        return new SpecTestNormalized(urls, specTest.responseStatus(), specTest.responseJson());
```
steven-sheehy, line 58: It's possible, but unlikely we might have both url and urls populated. Either way, this solution reduces conditionals:

```suggestion
        return new SpecTestNormalized(specTest.getNormalizedUrls(), specTest.responseStatus(), specTest.responseJson());
```

Where this helper method is added to `SpecTest`:

```java
    public List<String> getNormalizedUrls() {
        var url = Stream.ofNullable(url);
        var urls = Stream.ofNullable(urls).flatMap(ArrayList::stream);
        return Stream.concat(url, urls).filter(StringUtils::hasText).toList();
    }
```

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/model/SpecSetup.java
```diff
@@ -0,0 +1,35 @@
...
+package com.hedera.mirror.restjava.spec.model;
+
+import java.util.List;
+import java.util.Map;
+
+public record SpecSetup(
+        Map<String, String> features,
+        List<Map<String, Object>> accounts,
+        List<Map<String, Object>> contracts,
+        List<Map<String, Object>> cryptoallowances,
```
steven-sheehy, line 26:
```suggestion
        List<Map<String, Object>> cryptoAllowances,
```

hedera-mirror-common/src/test/java/com/hedera/mirror/common/domain/DomainWrapperImpl.java
steven-sheehy, line 1: revert

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/RestSpecTest.java
```diff
@@ -0,0 +1,131 @@
...
+    private static final List<File> SELECTED_SPECS = List.of(
+            new File("../hedera-mirror-rest/__tests__/specs/accounts/alias-into-evm-address.json"),
+            new File("../hedera-mirror-rest/__tests__/specs/blocks/no-records.json"),
+            new File("../hedera-mirror-rest/__tests__/specs/accounts/specific-id.json")
+    );
+
+    @Resource
+    private SpecDomainBuilder specDomainBuilder;
+
+    @Autowired
```
steven-sheehy, line 68: Mixing resource and autowired annotations. Prefer resource

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/RestSpecTest.java
```diff
@@ -0,0 +1,131 @@
...
+@RunWith(SpringRunner.class)
+@EnableConfigurationProperties(value = RestJavaProperties.class)
+@SpringBootTest(
+        classes = SpecTestConfig.class,
+        webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT,
+        properties = {"spring.main.allow-bean-definition-overriding=true"})
+public class RestSpecTest extends RestJavaIntegrationTest {
+    private static final int JS_REST_API_CONTAINER_PORT = 5551;
+
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
```
steven-sheehy, line 57: There should be an ObjectMapper bean provided by Spring you can use, unless doing so causes issues.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/RestSpecTest.java
```diff
@@ -0,0 +1,131 @@
...
+
+    private static final List<File> SELECTED_SPECS = List.of(
+            new File("../hedera-mirror-rest/__tests__/specs/accounts/alias-into-evm-address.json"),
+            new File("../hedera-mirror-rest/__tests__/specs/blocks/no-records.json"),
+            new File("../hedera-mirror-rest/__tests__/specs/accounts/specific-id.json")
```
steven-sheehy, line 62: Use Path to construct for windows compatibility. Move common prefix to a shared static variable.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/RestSpecTest.java
```diff
@@ -0,0 +1,131 @@
...
+    );
+
+    @Resource
+    private SpecDomainBuilder specDomainBuilder;
+
+    @Autowired
+    GenericContainer<?> jsRestApi;
```
steven-sheehy, line 69: private

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/RestSpecTest.java
```diff
@@ -0,0 +1,131 @@
...
+    private int port;
+
+    private RestClient.Builder restClientBuilder;
+
+    private String baseJsRestApiUrl;
+
+    @BeforeEach
+    final void setup() {
+        baseJsRestApiUrl = "http://%s:%d".formatted(jsRestApi.getHost(), jsRestApi.getMappedPort(JS_REST_API_CONTAINER_PORT));
+        log.info("setup - baseJsRestApiUrl: {}", baseJsRestApiUrl);
```
steven-sheehy, line 84: Should probably remove as it will get verbose

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/RestSpecTest.java
```diff
@@ -0,0 +1,131 @@
...
+        SELECTED_SPECS.forEach(this::runSpecTests);
+    }
+
+    private void runSpecTests(File specFile) {
+        RestSpec restSpec;
+        try {
+            restSpec = OBJECT_MAPPER.readValue(specFile, RestSpec.class);
+        } catch (IOException e) {
+            log.warn("Failed to parse spec file: {}", specFile, e);
+            return;
```
steven-sheehy, line 102: Should either catch and re-throw or use sneaky throws so that we don't stop test without failing.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/RestSpecTest.java
```diff
@@ -0,0 +1,131 @@
...
+                    log.warn("Failed to parse expected responseJson within spec file: {}", specFile, e);
+                }
+            }
+        }
+
+    }
+
+    private void setupDatabase(RestSpecNormalized normalizedRestSpec) {
+        specDomainBuilder.addAccounts(normalizedRestSpec.setup().accounts());
+        specDomainBuilder.addTokenAccounts(normalizedRestSpec.setup().tokenaccounts());
```
steven-sheehy, line 129: nit
```suggestion
        var setup = normalizedRestSpec.setup();
        specDomainBuilder.addAccounts(setup.accounts());
        specDomainBuilder.addTokenAccounts(setup.tokenaccounts());
```

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/RestSpecTest.java
```diff
@@ -0,0 +1,131 @@
...
+    private void runSpecTests(File specFile) {
+        RestSpec restSpec;
+        try {
+            restSpec = OBJECT_MAPPER.readValue(specFile, RestSpec.class);
+        } catch (IOException e) {
+            log.warn("Failed to parse spec file: {}", specFile, e);
+            return;
+        }
+
+        var normalizedRestSpec = RestSpecNormalized.from(restSpec);
```
steven-sheehy, line 105: Consolidate logic related to loading specs to a single method.

```suggestion
        var restSpec = loadSpecs();
```

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/RestSpecTest.java
```diff
@@ -0,0 +1,131 @@
...
+
+        for (var specTest : normalizedRestSpec.tests()) {
+            for (var url : specTest.urls()) {
+                var restClient = restClientBuilder.baseUrl(baseJsRestApiUrl + url).build();
+                var response = restClient.get()
+                        .retrieve()
+                        .toEntity(String.class);
+
+                assertThat(response.getStatusCode().value()).isEqualTo(specTest.responseStatus());
+                var specResponseJson = specTest.responseJson().toString();
```
steven-sheehy, line 116: Why do we both loading `responseJson` as a JsonNode if we then convert it to string anyway? Is it possible to load it directly as a string?

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/RestSpecTest.java
```diff
@@ -0,0 +1,131 @@
...
+                try {
+                    JSONAssert.assertEquals(specResponseJson, response.getBody(), JSONCompareMode.LENIENT);
+                } catch (JSONException e) {
+                    log.warn("Failed to parse expected responseJson within spec file: {}", specFile, e);
```
steven-sheehy, line 120: Should fail the tests

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/builder/SpecDomainBuilder.java
```diff
@@ -0,0 +1,53 @@
...
+import org.springframework.beans.factory.config.ConfigurableBeanFactory;
+import org.springframework.context.annotation.Scope;
+import org.springframework.stereotype.Component;
+
+/**
+ * Used to build domain objects for use in spec testing. This is different from general unit tests utilizing
+ * {@link DomainBuilder}, in that default values and overrides are based on the original REST module
+ * integrationDomainOps.js and attribute names provided in the spec test JSON setup.
+ */
+@Component
```
steven-sheehy, line 33: Use `@Named`

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/builder/SpecDomainBuilder.java
```diff
@@ -0,0 +1,53 @@
...
+@CustomLog
+@RequiredArgsConstructor
+@Scope(ConfigurableBeanFactory.SCOPE_PROTOTYPE)
```
steven-sheehy, line 36: Why prototype? It doesn't seem to have any state.

steven-sheehy (changes requested):

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/model/RestSpecNormalized.java
```diff
@@ -0,0 +1,62 @@
...
+ * @param tests at least one normalized test
+ */
+public record RestSpecNormalized(
+        String description,
+        String extendedDescription,
+        String matrix,
+        SpecSetup setup,
+        List<SpecTestNormalized> tests) {
+
+    public RestSpecNormalized {
```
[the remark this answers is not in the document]
jascks: The canonical constructor must have at least the visibility of the record itself, which must be `public` because it is referenced outside of the `model` package, namely `RestSpecTest`.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/RestSpecTest.java
jascks: Right! Thanks. Done.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/RestSpecTest.java
jascks: I was not able to deserialize the nested JSON to a `String` (Jackson `MistmatchedInputException`), and declaring that field as `Object` results in a number of nested lists and maps which is not easily converted to a string. `JsonNode` on the other hand, handles the job since the JSONAssert library is pretty flexible in parsing and doing comparisons.

Doing some more research, I ended up implementing `JsonAsStringDeserializer` and annotating the models to use that deserializer. In the end, that deserializer produces a `TreeNode` and converts it to a `String`, but at least the models don't have to convey a `JsonNode` going forward.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/RestSpecTest.java
jascks: Fail at what level though? This phase 1 implementation is hardcoded to run three existing specs, only the last of which has more than one URL, and thus API invocation.  When moving forward with a more real implementation that walks some directory hierarchy of spec files, there will be some number of files read and parsed, and one or more test cases per file. All being run as a single unit test (`@Test`).

In this method, if the object mapper fails to deserialize the spec file, then that needs to be reported, with testing moving on to the next file. Upon successful deserialization, each test case needs to be run and verified. A failure at that level needs to be reported, with testing proceeding to the next test case etc.

The code above required the try/catch because `JSONAssert.assertEquals()` will throw the checked `JSONException` if the provided `specResponseJson` fails to parse. That should not happen since the entire spec file did indeed parse to get to this point.

I can see this code needs to better deal with assertion failures (such as `assertThat()` for the response code, and in the future the response headers) and report such failures and continue testing rather than the JUnit level test method simply returning stopping further testing.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/builder/AccountBuilder.java
```diff
@@ -0,0 +1,93 @@
...
+    @Override
+    protected void customizeAndPersistEntity(Map<String, Object> account) {
+        var builder = Entity.builder();
+        // Set defaults
+        builder
+                .alias(null)
+                .autoRenewAccountId(null)
+                .autoRenewAccountId(null)
```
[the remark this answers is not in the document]
jascks: Gone

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/builder/AccountBuilder.java
```diff
@@ -0,0 +1,93 @@
...
+                .balance(null)
+                .balanceTimestamp(null)
+                .createdTimestamp(null)
```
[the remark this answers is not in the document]
jascks: Done

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/builder/SpecDomainBuilder.java
jascks: Done.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/builder/SpecDomainBuilder.java
jascks: Not sure why that's there. Gone.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/model/SpecTestNormalized.java
```diff
@@ -0,0 +1,68 @@
...
+ * @param urls the API URL(s) to be invoked
+ * @param responseStatus expected integer HTTP status code expected upon success
+ * @param responseJson JSON response expected from the API upon success
+ */
+public record SpecTestNormalized(
+        List<String> urls,
+        int responseStatus,
+        JsonNode responseJson) {
+
+    public SpecTestNormalized {
```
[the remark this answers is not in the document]
jascks: A record's canonical constructor must have visibility at least that of the record itself, which is used outside of this package.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/RestSpecTest.java
jascks: Good call. This method is destined to only get longer...  Done.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/model/SpecTestNormalized.java
jascks: Cool. I like it. Done with a slight modification.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/builder/AbstractEntityBuilder.java
```diff
@@ -0,0 +1,155 @@
...
+                            method.invoke(builder, mappedBuilderParameter);
+                        }
+                    } catch (IllegalAccessException | InvocationTargetException e) {
+                        log.warn("Failed to invoke method '{}' for attribute override '{}' for {}",
+                                methodName, customization.getKey(), builderClass.getName(), e);
+                    }
+                }
+                else {
+                    log.warn("Unknown attribute override '{}' for {}", customization.getKey(), builderClass.getName());
+                }
```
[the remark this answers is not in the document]
jascks: I don't see `EntityRegistry`, but I did find `EntityMetadata` and `EntityMetadataRegistry`. They and `ColumnMetadata` are found in `com.hedera.mirror.importer.repository.upsert` which is in `importer`, not `common`.

They aren't referenced in too many places. Do you recommend I relocate the required classes to `common`?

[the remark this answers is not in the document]
jascks: Looking further, the reflection code above is used to access fluent entity builder methods, not setters on the entities themselves. That's the builder's job.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/builder/AccountBuilder.java
```diff
@@ -0,0 +1,93 @@
...
+                .stakedAccountId(null)
+                .stakedNodeId(-1L)
+                .stakePeriodStart(-1L)
+                .submitKey(null)
+                .timestampRange(Range.atLeast(0L))
+                .type(EntityType.ACCOUNT);
+
+        // Customize with spec setup definitions
+        var wrapper = new DomainWrapperImpl<Entity, EntityBuilder<?, ?>>(builder, builder::build, entityManager, transactionOperations);
+        customizeWithSpec(wrapper, account);
```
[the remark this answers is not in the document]
jascks: Okay, I've undone this.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/builder/AbstractEntityBuilder.java
```diff
@@ -0,0 +1,155 @@
...
+abstract class AbstractEntityBuilder {
+
+    /*
+     * Map builder parameter value provided in spec JSON to the type expected by the builder method.
+     */
+    private static final Map<Class<?>, Function<Object, Object>> DEFAULT_PARAMETER_CONVERTERS = Map.of(
+            java.lang.Boolean.class, source -> Boolean.parseBoolean(String.valueOf(source)),
+            java.lang.Integer.class, source -> Integer.parseInt(String.valueOf(source)),
+            java.lang.Long.class, source -> Long.parseLong(String.valueOf(source))
+    );
```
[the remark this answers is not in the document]
jascks: Done, at least for for those default converters above. Not yet sure how to incorporate the builder method specific converters since right at the start there are three different cases where `String -> byte[]` conversion needs to occur.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/model/SpecSetup.java
jascks: Good catch! Done.

hedera-mirror-common/src/test/java/com/hedera/mirror/common/domain/DomainWrapperImpl.java
jascks: Done

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/RestSpecTest.java
jascks: Done.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/RestSpecTest.java
jascks: Done

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/RestSpecTest.java
jascks: Done

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/RestSpecTest.java
jascks: I want to log that there is something funky about the spec file and move on to the next one.

jascks: Given either this phase 1 hardcoded list of three spec files, or a more full featured file system walk with filtering, within a single `@Test` method we are encountering spec JSON files, each of which must be loaded/parsed, DB prepared, and tests/urls run.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/RestSpecTest.java
jascks: Using `@ParameterizedTest`, I have redone this logic.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/builder/AbstractEntityBuilder.java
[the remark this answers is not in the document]
steven-sheehy: Not sure what you mean by `used to access fluent entity builder methods, not setters on the entities themselves`. It is actually setters on the domain model as proven by the test in `EntityMetadataRegistryTest`:

```
                .satisfies(cm -> assertThat(cm.getGetter().apply(entity)).isEqualTo(entity.getAlias()))
                .satisfies(cm -> assertThatCode(() -> cm.getSetter().accept(entity, newValue))
                        .satisfies(d -> assertThat(entity.getAlias()).isEqualTo(newValue)));
```

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/model/RestSpec.java
```diff
@@ -0,0 +1,32 @@
...
+public record RestSpec(
+        String description,
+        String extendedDescription,
+        String matrix,
+        SpecSetup setup,
+        List<SpecTest> tests,
+        String url,
+        List<String> urls,
+        int responseStatus,
+        JsonNode responseJson) {
```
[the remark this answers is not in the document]
steven-sheehy: We should still add it to the spec so we don't forget it. Not everything in the current spec is supported like matrix but it's still present.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/RestSpecTest.java
```diff
@@ -0,0 +1,124 @@
...
+                assertThat(response.getStatusCode().value()).isEqualTo(specTest.responseStatus());
+                JSONAssert.assertEquals(specTest.responseJson(), response.getBody(), JSONCompareMode.LENIENT);
+            }
+        }
+    }
+
+    private void setupDatabase(RestSpecNormalized normalizedRestSpec) {
+        var setup = normalizedRestSpec.setup();
+        specDomainBuilder.addAccounts(setup.accounts());
+        specDomainBuilder.addTokenAccounts(setup.tokenAccounts());
```
steven-sheehy, line 118: As more get added, this method and SpecDomainBuilder will each have to change. It would be better to invert this logic so we can add new builders in isolation. Something like the below (though not sure about name `Persistence`):

```java

    public interface DomainPersistence {
        void persist(SpecSetup specSetup);
    }

    @Named
    @Primary
    @RequiredArgsConstructor
    class CompositeDomainPersistence implements DomainPersistence {
        private final Collection<DomainPersistence> persistences;

        public void persist(SpecSetup specSetup) {
            persistences.forEach(p -> p.persist(specSetup));
        }
    }

    @RequiredArgsConstructor
    abstract class AbstractDomainPersistence implements DomainPersistence {

        private final EntityManager entityManager;
        private final Function<SpecSetup, List<Map<String, Object>>> supplier;

        @Override
        public void persist(SpecSetup specSetup) {
            var entities = supplier.apply(specSetup);
            if (entities != null) {
                for (var entity : entities) {
                    entityManager.persist(build(entity));
                }
            }
        }

        abstract Object build(Map<String, Object> entity);
    }

    @Named
    class EntityPersistence extends AbstractDomainPersistence {
        EntityPersistence() {
            super(SpecSetup::accounts);
        }

        Object build(Map<String, Object> entity) {
        }
    }
```

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/builder/AbstractEntityBuilder.java
```diff
@@ -0,0 +1,137 @@
...
+import org.apache.commons.codec.binary.Base32;
+import org.apache.tuweni.bytes.Bytes;
+import org.springframework.core.convert.ConversionService;
+
+@CustomLog
+abstract class AbstractEntityBuilder<B> {
+
+    private static final Base32 BASE32 = new Base32();
+    private static final Pattern HEX_STRING_PATTERN = Pattern.compile("^(0x)?[0-9A-Fa-f]+$");
+    private static final Object IGNORE_ATTRIBUTE_SIGNAL = new Object();
```
steven-sheehy, line 41: Is this and the IGNORE_CONVERTER used?

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/builder/AbstractEntityBuilder.java
```diff
@@ -0,0 +1,137 @@
...
+
+    /*
+     * Common handy spec attribute value converter functions to be used by subclasses.
+     */
+    protected static final Function<Object, Object> IGNORE_CONVERTER = value -> IGNORE_ATTRIBUTE_SIGNAL;
+
+    protected static final Function<Object, Object> BASE32_CONVERTER = value -> value == null ? null : BASE32.decode(value.toString());
+
+    protected static final Function<Object, Object> ENTITY_ID_CONVERTER = value -> value == null ? null
+            : value instanceof String valueStr ? EntityId.of(valueStr) : EntityId.of((Long)value);
```
steven-sheehy, line 52: We should be able to register a Spring `EntityIdConverter` to handle this and avoid mapping each field to use this.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/builder/AccountBuilder.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.Map;
+import java.util.function.Function;
+
+@Named
+class AccountBuilder extends AbstractEntityBuilder<Entity.EntityBuilder<?, ?>> {
+
+    private static final Map<String, Function<Object, Object>> METHOD_PARAMETER_CONVERTERS = Map.of(
+            "alias", BASE32_CONVERTER,
+            "evmAddress", HEX_OR_BASE64_CONVERTER,
+            "id", ENTITY_ID_CONVERTER,
```
steven-sheehy, line 34: id is a Long, not an EntityId. Think the wrong converter is registered

steven-sheehy (changes requested):

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/builder/AbstractEntityBuilder.java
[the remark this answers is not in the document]
jascks: For example, in `AccountBuilder` I'm. utilizing `Entity.builder()` to set the default values for an account, and then providing that builder to `customizeWithSpec()` to further customize the account based on the spec JSON. This process involves invoking methods on that builder, not an account `Entity` instance.

So, while grok'ing a spec file, given the "num" attribute for an account object in the "accounts" array, I need to invoke `builder.num()`. I'm not currently invoking the Lombok generated `Entity.setNum()`. It looks like you'd prefer I did the latter.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/builder/AbstractEntityBuilder.java
jascks: I admit it is not in this phase 1 PR. I will remove it.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/RestSpecTest.java
```diff
@@ -0,0 +1,124 @@
...
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.boot.test.web.server.LocalServerPort;
+import org.springframework.test.context.junit4.SpringRunner;
+import org.springframework.web.client.RestClient;
+import org.testcontainers.containers.GenericContainer;
+
+@RequiredArgsConstructor
+@RunWith(SpringRunner.class)
+@EnableConfigurationProperties(value = RestJavaProperties.class)
```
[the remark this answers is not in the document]
jascks: Not in phase 1. I'm not even using the injected `RestJavaProperties`. I will remove it.

I don't see it enabled in the parent class, nor its parent class. I do see it enabled also in `ControllerTest`. Perhaps this should be done in the common parent `RestJavaIntegrationTest`, along with the injected field (protected)?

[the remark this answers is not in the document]
steven-sheehy: It shouldn't need to be manually enabled anywhere. Spring Boot will component scan for configuration properties and make them beans thanks to `@ConfigurationPropertiesScan`.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/builder/AbstractEntityBuilder.java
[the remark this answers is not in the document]
steven-sheehy: I understand what you're saying now. So we can change to not use builders and then we can use the registry. I think we can keep it as is for now.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/repository/TokenAccountRepository.java
```diff
@@ -0,0 +1,8 @@
+package com.hedera.mirror.restjava.repository;
```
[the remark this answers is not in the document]
jascks: Fixed.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/repository/TokenAccountRepository.java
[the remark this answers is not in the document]
jascks: Done.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/RestSpecTest.java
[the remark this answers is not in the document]
jascks: Removed, as well as the `properties` field.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/RestSpecTest.java
```diff
@@ -0,0 +1,124 @@
...
+@SpringBootTest(
+        classes = SpecTestConfig.class,
+        webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT,
+        properties = {"spring.main.allow-bean-definition-overriding=true"})
+public class RestSpecTest extends RestJavaIntegrationTest {
+    private static final int JS_REST_API_CONTAINER_PORT = 5551;
+
+
+    private static final Path REST_BASE_PATH = Path.of("..", "hedera-mirror-rest", "__tests__", "specs");
```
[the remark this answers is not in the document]
jascks: Fixed.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/RestSpecTest.java
```diff
@@ -0,0 +1,124 @@
...
+    @LocalServerPort
+    private int port;
+
+    private RestClient.Builder restClientBuilder;
+
+    private String baseJsRestApiUrl;
+
+    @BeforeEach
+    final void setup() {
+        baseJsRestApiUrl = "http://%s:%d".formatted(jsRestApi.getHost(), jsRestApi.getMappedPort(JS_REST_API_CONTAINER_PORT));
```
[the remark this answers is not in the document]
jascks: Done. Explicitly implemented constructor so that I can create immutable `RestClient`.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/builder/SpecDomainBuilder.java
```diff
@@ -0,0 +1,50 @@
...
+import lombok.CustomLog;
+import lombok.RequiredArgsConstructor;
+
+/**
+ * Used to build domain objects for use in spec testing. This is different from general unit tests utilizing
+ * {@link DomainBuilder}, in that default values and overrides are based on the original REST module
+ * integrationDomainOps.js and attribute names provided in the spec test JSON setup.
+ */
+@Named
+@CustomLog
```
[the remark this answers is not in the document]
jascks: Removed.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/model/JsonAsStringDeserializer.java
```diff
@@ -0,0 +1,30 @@
...
+
+package com.hedera.mirror.restjava.spec.model;
+
+import com.fasterxml.jackson.core.JacksonException;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JsonDeserializer;
+import java.io.IOException;
+
+public class JsonAsStringDeserializer extends JsonDeserializer<String> {
```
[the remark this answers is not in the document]
jascks: Done.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/model/RestSpec.java
```diff
@@ -0,0 +1,33 @@
...
+
+public record RestSpec(
+        String description,
+        String extendedDescription,
+        String matrix,
+        SpecSetup setup,
+        List<SpecTest> tests,
+        String url,
+        List<String> urls,
+        int responseStatus,
```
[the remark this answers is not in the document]
jascks: Done.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/RestSpecTest.java
```diff
@@ -0,0 +1,124 @@
...
+
+    @ParameterizedTest
+    @MethodSource("getSpecFilesToRun")
+    void runSpecTests(Path specFilePath) throws IOException, JSONException {
+        var spec = RestSpecNormalized.from(objectMapper.readValue(specFilePath.toFile(), RestSpec.class));
+        log.info("Running spec '{}'", spec.description());
+        setupDatabase(spec);
+
+        for (var specTest : spec.tests()) {
+            for (var url : specTest.urls()) {
```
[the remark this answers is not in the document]
jascks: Looks like `@TestFactory` and `DynamicTest` have great potential. Investigating.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/builder/SpecDomainBuilder.java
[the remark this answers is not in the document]
jascks: Deleted.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/RestSpecTest.java
jascks: I took a bit of a different approach on the implementation.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/repository/TokenAccountRepository.java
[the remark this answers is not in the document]
jascks: Test added.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/model/RestSpec.java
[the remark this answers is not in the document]
jascks: Done.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/RestSpecTest.java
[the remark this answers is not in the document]
jascks: See new implementation utilizing `@TestFactory`, `DynamicTest` and `DynamicContainer`.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/RestSpecTest.java
jascks: Implemented.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/builder/AbstractEntityBuilder.java
jascks: Done. Note that spec JSON will sometimes specify an ID as a string, other times as a numeric value (being treated as a `Long`). Thus, I have created two converters to support both forms which deals with the logic in the above function.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/builder/AccountBuilder.java
jascks: True. The Spring converters introduced above handle both situations.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/builder/AccountBuilder.java
```diff
@@ -0,0 +1,77 @@
...
+
+    private final EntityRepository entityRepository;
+
+    AccountBuilder(EntityRepository entityRepository) {
+        super(METHOD_PARAMETER_CONVERTERS);
+        this.entityRepository = entityRepository;
+    }
+
+    @Override
+    protected void customizeAndPersistEntity(Map<String, Object> account) {
```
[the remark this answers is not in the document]
jascks: I like it. Done!

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/builder/AccountBuilder.java
```diff
@@ -0,0 +1,77 @@
...
+                .timestampRange(Range.atLeast(0L))
+                .type(EntityType.ACCOUNT);
+
+        // Customize with spec setup definitions
+        customizeWithSpec(builder, account);
+
+        // Check and finalize
+        var entity = builder.build();
+        if (entity.getId() == null) {
+            builder.id(EntityId.of(entity.getShard(), entity.getRealm(), entity.getNum()).getId());
```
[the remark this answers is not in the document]
jascks: Done.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/builder/AccountBuilder.java
jascks: Done.

steven-sheehy (changes requested):

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/converter/EntityIdFromStringConverterTest.java
```diff
@@ -0,0 +1,32 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
+
+public class EntityIdFromStringConverterTest {
+
+    @ParameterizedTest(name = "Convert String to EntityId")
+    @CsvSource({"1.2.3, 1.2.3", "0.0.100, 0.0.100"})
```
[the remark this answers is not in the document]
jascks: None that I can see, though there are a lot using "0.0.1001".

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/converter/EntityIdFromLongConverterTest.java
```diff
@@ -0,0 +1,32 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
+
+public class EntityIdFromLongConverterTest {
+
+    @ParameterizedTest(name = "Convert Long to EntityId")
+    @CsvSource({"1, 0.0.1", "100, 0.0.100"})
```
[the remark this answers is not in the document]
jascks: I wondered about that, but how much do I need to repeat `EntityIdTest` vs. make sure code coverage passes? That class also tests for negative numbers,  values exceeding what's supported for encoding shard, realm etc. Testing for such failures requires dealing with `IllegalArgumentException` etc.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/converter/EntityIdFromStringConverterTest.java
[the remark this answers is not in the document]
jascks: Done.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/builder/AbstractEntityBuilder.java
jascks: Persistent, isn't it? 😄 Gone finally!

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/builder/AccountBuilder.java
```diff
@@ -0,0 +1,79 @@
...
+    protected Supplier<List<Map<String, Object>>> getSpecEntitiesSupplier(SpecSetup specSetup) {
+        return specSetup::accounts;
+    }
+
+    @Override
+    protected Entity customizeEntity(Map<String, Object> account) {
+        var builder = Entity.builder();
+        // Set defaults
+        builder
+                .declineReward(Boolean.FALSE)
```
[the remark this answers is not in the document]
jascks: Done.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/builder/TokenAccountBuilder.java
```diff
@@ -0,0 +1,70 @@
...
+    protected Supplier<List<Map<String, Object>>> getSpecEntitiesSupplier(SpecSetup specSetup) {
+        return specSetup::tokenAccounts;
+    }
+
+    @Override
+    protected TokenAccount customizeEntity(Map<String, Object> account) {
+        var builder = TokenAccount.builder();
+        // Set defaults
+        builder
+                .accountId(0L)
```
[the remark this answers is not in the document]
jascks: Done.

steven-sheehy:
Sonar notes some code smells

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/converter/EntityIdFromLongConverterTest.java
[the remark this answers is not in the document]
steven-sheehy: Just null and empty string, as I said. It's not duplicating coverage. EntityId.of will throw an exception but this converter should probably just return null.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/builder/AbstractEntityBuilder.java
```diff
@@ -0,0 +1,154 @@
...
+            Map<String, String> attributeNameMap) {
+        this.methodParameterConverters = methodParameterConverters;
+        this.attributeNameMap = attributeNameMap;
+    }
+
+    @Override
+    public void customizeAndPersistEntities(SpecSetup specSetup) {
+        var specEntities = getSpecEntitiesSupplier(specSetup).get();
+        if (!CollectionUtils.isEmpty(specEntities)) {
+            specEntities.forEach(specEntity -> transactionOperations.executeWithoutResult(t -> entityManager.persist(customizeEntity(specEntity))));
```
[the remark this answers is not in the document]
jascks: This is a nice improvement. I have implemented this, well with another step between customization and persistence to allow for final customization by the sub-class, which `integrationDomainOps.js` does a lot of on the REST side.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/builder/SpecDomainBuilder.java
```diff
@@ -0,0 +1,33 @@
...
+import java.util.List;
+import lombok.RequiredArgsConstructor;
+
+@Named
+@RequiredArgsConstructor
+public class SpecDomainBuilder {
+
+    private final List<SpecEntityBuilder> specEntityBuilders;
+
+    public void addSetupEntities(SpecSetup specSetup) {
```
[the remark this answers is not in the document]
jascks: Indeed. Done!

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/builder/SpecEntityBuilder.java
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hedera.mirror.restjava.spec.builder;
+
+import com.hedera.mirror.restjava.spec.model.SpecSetup;
+
+public interface SpecEntityBuilder {
```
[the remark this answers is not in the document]
jascks: Once I implemented the aforementioned composite correctly, this interface is no longer needed. 😄

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/converter/EntityIdFromLongConverterTest.java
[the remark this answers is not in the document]
jascks: Whoops. Should have refreshed sooner. Working on this still...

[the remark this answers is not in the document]
jascks: Okay, I'm not completely feeling the joy of this regarding the `FromLong` converter, since there is no concept of empty string in this case. I can provide null, but that fails with NPE while unboxing in the converter itself as `EntityId` deals with a primitive `long`.

When such invalid input is encountered in a spec file, I think rather than returning null, an appropriate exception is in order to instruct the developer to fix the spec then and now.

steven-sheehy (dismissed):
LGTM

steven-sheehy (approved):

jnels124 (approved):

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/converter/EntityIdFromStringConverterTest.java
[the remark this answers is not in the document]
jascks: I don't think so. These tests are primarily to get code coverage for the converters and not extensively test `EntityId` itself. Also, here we test the possible  boundary conditions around null and empty (for strings anyway) possibly present within the spec JSON, and how the converters handle that differently from `EntityId`.  Any other circumstances where `EntityId` accepts the input or throws an exception are acceptable.

PR #8825, opened 2024-07-18
Optimize accounts rest api to avoid full entity_stake index scan

### LINKED ISSUE #8804: Accounts REST api unnecessarily full index scan entity_stake table
### Description

With certain combinations of query params, `/api/v1/accounts` does a full index scan of `entity_stake` table to grab a tiny fraction of rows which are needed in the response. This causes periodic high spikes of number of selected rows reported by the database. 

### Steps to reproduce

Check slack thread

### Additional context

_No response_

### Hedera network

other

### Version

v0.108.0

### Operating system

None

xin-hedera:
**Description**:

This PR optimizes the query for accounts rest api to avoid full entity_stake index scan

- Add account id condition to `entity_stake` left join
- Prettify SQL query logging at trace level

**Related issue(s)**:

Fixes #8804 

**Notes for reviewer**:

The SQL query time with `limit 100` is reduced from 58ms to 3ms.

Output of `explain analyze` shows huge difference in actual number of rows.

```
         ->  Index Scan using entity_stake_pkey on entity_stake es  (cost=0.42..7221.88 rows=103484 width=24) (actual time=0.005..41.461 rows=102204 loops=1)
``` 
v.s.

```
         ->  Index Scan using entity_stake_pkey on entity_stake es  (cost=0.42..1350.68 rows=1386 width=24) (actual time=0.015..0.015 rows=1 loops=1)
               Index Cond: (id > 6205393)
```

**Checklist**

### DIFF
--- a/hedera-mirror-rest/accounts.js
+++ b/hedera-mirror-rest/accounts.js
@@ -237,7 +237,7 @@ const getEntityBalanceQuery = (
     from ${entityTable} as e
     left join
       ${getEntityStakeQuery(entityAccountQuery.query, isHistorical)}
-    as es on es.id = e.id
+    as es on es.id = e.id ${getEntityStakeAccountCondition(entityAccountQuery)}
     ${[whereClause, orderClause, limitQuery].filter(Boolean).join('\n')}`);
   const query = queries.join('\n');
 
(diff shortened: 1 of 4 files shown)


hedera-mirror-rest/accounts.js
```diff
@@ -479,8 +489,7 @@ const getOneAccount = async (req, res) => {
     timestampFilters.length > 0
   );
 
-  const pgEntityQuery = utils.convertMySqlStyleQueryToPostgres(entityQuery);
```
steven-sheehy, line 482: There looks to be only two more uses of this function if you want to convert it as well and remove this util?

steven-sheehy (approved):
LGTM

hedera-mirror-rest/accounts.js
xin-hedera: ah no. it's kind of a hack in `getAccountQuery` the pg style placehodlers are generated as a whole. We can decide if it's worth the effort to get rid of all mysql related stuff including many other functions in `utils`, or just change when we feel it's too much churn to make some other changes, or just wait for rest java conversion.

jnels124 (approved):
LGTM

PR #8870, opened 2024-07-25
HIP-540 Update REST API to return correct value for empty KeyLists

### LINKED ISSUE #8845: HIP-540 Show token's sentinel immutable key as null in REST API
### Problem

HIP-540 changes HTS to allow any token key to remove itself by changing to an empty `KeyList` or an unusable key.

Services `TokenGetInfoQuery`  will return null for a key if it matches [IMMUTABILITY_SENTINEL_KEY](https://github.com/hashgraph/hedera-services/blob/13a1bb443779ce5fee45be119561028a857b520c/hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/key/KeyUtils.java#L34), which is just an empty `KeyList`. However, mirrornode REST API behaves differently by showing the raw bytes `0x3200`, e.g.,

```
{
  "fee_schedule_key": {
    "_type": "ProtobufEncoded",
    "key": "3200"
  },
  ...
  "supply_key": {
    "_type": "ProtobufEncoded",
    "key": "3200"
  },
  ...
  "token_id": "0.0.4814059",
  ...
}
```

### Solution

Apply the same logic as HAPI query `TokenGetInfoQuery` to show `null` if a token key is the empty `KeyList`

### Alternatives

_No response_

edwin-greene:
**Description**:
Updates the REST API to display the empty KeyList/Immutable Sentinel value as null instead of the byte representation of the sentinel value (`3200`).

**Related issue(s)**:

Fixes #8845 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-rest/__tests__/specs/tokens/{id}/custom-fees.json
+++ b/hedera-mirror-rest/__tests__/specs/tokens/{id}/custom-fees.json
@@ -104,7 +104,8 @@
         "fee_schedule_key": [1, 2, 3],
         "type": "FUNGIBLE_COMMON",
         "pause_key": "4a5ad514f0957fa170a676210c9bdbddf3bc9519702cf915fa6767a40463b96x",
-        "pause_status": "UNPAUSED"
+        "pause_status": "UNPAUSED",
+        "supply_key": "2\\000"
       },
       {
         "token_id": "0.7.25301",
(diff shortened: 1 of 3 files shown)


hedera-mirror-rest/__tests__/specs/tokens/{id}/custom-fees.json
```diff
@@ -104,7 +104,8 @@
         "fee_schedule_key": [1, 2, 3],
         "type": "FUNGIBLE_COMMON",
         "pause_key": "4a5ad514f0957fa170a676210c9bdbddf3bc9519702cf915fa6767a40463b96x",
-        "pause_status": "UNPAUSED"
+        "pause_status": "UNPAUSED",
+        "supply_key": "2\\000"
```
edwin-greene, line 108: This is the string representation of the row value `x\3200`

steven-sheehy (approved):
LGTM

xin-hedera (approved):
LGTM

Reccetech:
@steven-sheehy will this change also apply to HCS per HIP 1139? Thanks!

steven-sheehy:
It should apply to all keys that appear in any REST API

PR #8907, opened 2024-07-30
Log errors generically

### LINKED ISSUE #8895: Update REST Response Handling
### Problem

Update response handling to be more generic

### Solution

Investigate solutions that allow generic handling of responses

### Alternatives

_No response_

jnels124:
**Description**:
* add handler to log uncaught errors

**Related issue(s)**:

Fixes #8895 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-rest/__tests__/middleware/httpErrorHandler.test.js
+++ b/hedera-mirror-rest/__tests__/middleware/httpErrorHandler.test.js
@@ -1,0 +1,48 @@
+/*
+ * Copyright (C) 2022-2024 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {
+  DbError,
+  FileDecodeError,
+  FileDownloadError,
+  InvalidArgumentError, InvalidClauseError,
+  InvalidConfigError,
+  NotFoundError
+} from '../../errors';
+
+import {handleUncaughtException} from "../../middleware/httpErrorHandler.js";
+
+
+describe('Server error handler', () => {
+
+  test('Throws Error for non rest error', () => {
+    const exception = () => handleUncaughtException(new InvalidConfigError("Bad Config"))
+    expect(exception).toThrow(InvalidConfigError);
+  });
+
+  test('Does not throw error for rest error', () => {
+    const exception = () => {
+      handleUncaughtException(new DbError());
+      handleUncaughtException(new FileDecodeError());
+      handleUncaughtException(new FileDownloadError());
+      handleUncaughtException(new InvalidArgumentError());
+      handleUncaughtException(new InvalidClauseError());
+      handleUncaughtException(new NotFoundError());
+    };
+
+    expect(exception).not.toThrow(Error);
+  });
+});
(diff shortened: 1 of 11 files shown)


hedera-mirror-rest/middleware/httpErrorHandler.js
```diff
@@ -57,6 +57,14 @@ const handleError = async (err, req, res, next) => {
   res.status(statusCode.code).json(errorMessageFormat(errorMessage));
 };
 
+const handleRejection = (reason, promise) => {
+    logger.warn(`Unhandled rejection at:${promise} reason: ${reason}`);
+}
+
+const handleUncaughtException = (err) => {
+    logger.error('Unhandled exception:', err);
```
jnels124, line 65: With current handling this should never be possible to be hit but was added for completeness

hedera-mirror-rest/middleware/httpErrorHandler.js
```diff
@@ -57,6 +57,14 @@ const handleError = async (err, req, res, next) => {
...
+}
+
```
steven-sheehy, line 67: I'm not sure it's safe to replace the default handler for all exception types. By default it will exit the process in these scenarios so that Kubernetes can restart and it recovers. If we avoid restarting for system errors outside our control like out of file descriptors or node memory (not OS) the processes can never recover and we've made the situation much worse than before.

Instead, we should add a common base class to our custom errors so they have a common hierarchy. Then we should only add this logic to our exceptions and fallback to the old behavior for others.

hedera-mirror-rest/__tests__/middleware/httpErrorHandler.test.js
```diff
@@ -0,0 +1,48 @@
...
+import {
+  DbError,
+  FileDecodeError,
+  FileDownloadError,
+  InvalidArgumentError, InvalidClauseError,
+  InvalidConfigError,
+  NotFoundError
+} from '../../errors';
+
+import {handleUncaughtException} from "../../middleware/httpErrorHandler.js";
```
xin-hedera, line 26: nit, and remove `.js` in all imports

```suggestion
import {handleUncaughtException} from "../../middleware/httpErrorHandler";
```

hedera-mirror-rest/__tests__/middleware/httpErrorHandler.test.js
```diff
@@ -0,0 +1,48 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {
+  DbError,
+  FileDecodeError,
+  FileDownloadError,
+  InvalidArgumentError, InvalidClauseError,
```
xin-hedera, line 21: looks like your spotless format is off or just doesn't work

hedera-mirror-rest/errors/restError.js
```diff
@@ -0,0 +1,24 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+class RestError extends Error {
+    constructor(errorMessage) {
+        super();
+        this.message = errorMessage;
+    }
```
xin-hedera, line 21: the constructor is not needed, nodejs `Error`'s constructor `new Error(message[, options])` already has message as the first argument.

Guess it's the same for all our custom error classes.

hedera-mirror-rest/__tests__/middleware/httpErrorHandler.test.js
jnels124: thanks had some intellij issues recently and must have wiped that out. Will add back.

jnels124:
verified exception thrown by server is caught by handler and when rethrown, the app exits as intended.

testing was done by adding below snippet to server.js

```
setTimeout(()=> {
  console.log('gonna throw exception');
  throw new Error('test exception');
}, 10000);
```

hedera-mirror-rest/errors/restError.js
jnels124: in some cases message can be an array so think this is still needed

jnels124: well really it is just the invalid argument error so maybe will remove the constructor from rest error and then just override message in that particular error.

jnels124:
Apologies... hit re-review to early... need to resolve something

steven-sheehy (approved):
LGTM

xin-hedera (approved):
LGTM

PR #8908, opened 2024-07-30
HIP-869 Adding admin_key for the dynamic address book

### LINKED ISSUE #8680: HIP-869 Add admin_key for node transactions
### Problem

In follow-up discussions on [HIP-869](https://hips.hedera.com/hip/hip-869), it was concluded that a new `admin_key` would be added to the Node transactions.
Mirror node needs to maintain the latest admin_key in the database which could change with every NodeUpdate transaction.

### Solution

- Make DB changes to add `admin_key` column to the `node` table or as per the design decision.
- Parse `admin_key` in the NodeTransactionHandlers and update it as needed.
Refer to [this](https://github.com/hashgraph/hedera-protobufs/pull/380/files) services PR .

### Alternatives

_No response_

mgoelswirlds:
**Description**:

This PR adds a new entity table `NODE` to capture Node transactions.

This PR modifies:

* Adds a `Node` and `NodeHistory` table to capture Node transactions in order to implement HIP-869 for mirror node.

**Related issue(s)**:

Fixes #8680

**Notes for reviewer**:

**Checklist**


hedera-mirror-importer/src/main/resources/db/migration/v1/V1.99.0__add_node.sql
```diff
@@ -0,0 +1,35 @@
...
+
+create table if not exists node_history
+(
+    like node including defaults
+);
+
+create index if not exists node_history__node_id_lower_timestamp
+    on node_history (node_id, lower(timestamp_range));
+
+drop table if exists ${tempSchema}.node_temp;
```
mgoelswirlds, line 23: Adding this and the following lines in temporary , till we decide what to do for entire MON.
Both V1 and V2

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/batch/BatchUpserterTest.java
```diff
@@ -541,6 +544,49 @@ void nftAllowance() {
         assertThat(nftAllowanceRepository.findAll()).containsExactlyInAnyOrderElementsOf(nftAllowance);
     }
 
+    @Test
+    void node() {
+        Node node1 = domainBuilder.node().get();
+        Node node2 = domainBuilder.node().get();
+        Node node3 = domainBuilder.node().get();
+        var node = List.of(node1, node2, node3);
```
xin-hedera, line 552: `var` for consistency

and rename the list of node to `nodes` since it's a bit confusing

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/record/entity/EntityRecordItemListenerNodeTest.java
```diff
@@ -114,12 +122,16 @@ void nodeCreate() {
...
                 .returns(recordItem.getTransaction().toByteArray(), Transaction::getTransactionBytes)
                 .returns(recordItem.getTransactionRecord().toByteArray(), Transaction::getTransactionRecordBytes);
+        softly.assertThat(nodeRepository.findAll()).containsExactly(expectedNode);
     }
 
     @Test
     void nodeUpdate() {
         entityProperties.getPersist().setNodes(true);
         var recordItem = recordItemBuilder.nodeUpdate().build();
+        var expectedNode = getExpectedNode(recordItem);
```
xin-hedera, line 132: need another test case for node update without changing admin key

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/transactionhandler/NodeCreateTransactionHandler.java
```diff
@@ -46,5 +53,17 @@ protected void doUpdateTransaction(Transaction transaction, RecordItem recordIte
         }
         transaction.setTransactionBytes(recordItem.getTransaction().toByteArray());
         transaction.setTransactionRecordBytes(recordItem.getTransactionRecord().toByteArray());
+        parseNode(recordItem);
+    }
+
+    private void parseNode(RecordItem recordItem) {
+        var nodeCreate = recordItem.getTransactionBody().getNodeCreate();
```
xin-hedera, line 60: check if transaction is successful

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/transactionhandler/NodeDeleteTransactionHandler.java
```diff
@@ -41,5 +45,16 @@ protected void doUpdateTransaction(Transaction transaction, RecordItem recordIte
         }
         transaction.setTransactionBytes(recordItem.getTransaction().toByteArray());
         transaction.setTransactionRecordBytes(recordItem.getTransactionRecord().toByteArray());
+        parseNode(recordItem);
+    }
+
+    private void parseNode(RecordItem recordItem) {
+        long consensusTimestamp = recordItem.getConsensusTimestamp();
```
xin-hedera, line 52: check if transaction is successful

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/transactionhandler/NodeUpdateTransactionHandler.java
```diff
@@ -46,5 +51,17 @@ protected void doUpdateTransaction(Transaction transaction, RecordItem recordIte
         }
         transaction.setTransactionBytes(recordItem.getTransaction().toByteArray());
         transaction.setTransactionRecordBytes(recordItem.getTransactionRecord().toByteArray());
+        parseNode(recordItem);
+    }
+
+    private void parseNode(RecordItem recordItem) {
+        var nodeUpdate = recordItem.getTransactionBody().getNodeUpdate();
```
xin-hedera, line 58: check if transaction is successful

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/record/transactionhandler/NodeCreateTransactionHandlerTest.java
```diff
@@ -75,5 +81,24 @@ void nodeCreateTransaction(boolean isPersist) {
...
+                                .getNodeCreate()
+                                .getAdminKey()
+                                .toByteArray(),
+                        Node::getAdminKey)
+                .returns(recordItem.getConsensusTimestamp(), Node::getTimestampLower)
+                .returns(false, Node::isDeleted)));
     }
+
+    @SuppressWarnings({"java:S2699", "java:S1186"})
+    @Disabled("Since this handler persists data for unsuccessful transactions & has tests for that")
```
xin-hedera, line 100: where did it test unsuccessful transaction?

xin-hedera (changes requested):

hedera-mirror-common/src/test/java/com/hedera/mirror/common/domain/DomainBuilder.java
```diff
@@ -688,6 +690,30 @@ public DomainWrapper<NftTransfer, NftTransfer.NftTransferBuilder> nftTransfer()
...
+                .timestampRange(Range.atLeast(timestamp));
+
+        return new DomainWrapperImpl<>(builder, builder::build);
+    }
+
+    public DomainWrapper<NodeHistory, NodeHistory.NodeHistoryBuilder<?, ?>> nodeHistory() {
+        long timestamp = timestamp();
+        var builder = NodeHistory.builder()
+                .adminKey(key())
+                .createdTimestamp(timestamp())
```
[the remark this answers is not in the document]
mgoelswirlds: changed it.

hedera-mirror-importer/src/main/resources/db/migration/common/R__01_temp_tables.sql
```diff
@@ -23,10 +23,12 @@ $$ language plpgsql;
 call create_temp_table_safe('contract_state', 'contract_id', 'slot');
 call create_temp_table_safe('crypto_allowance', 'owner', 'spender');
 call create_temp_table_safe('custom_fee', 'token_id');
+call create_temp_table_safe('token_transfer', 'token_id');
```
xin-hedera, line 26: remove it, believe this is a oneoff error when merging changes from `main`

```suggestion
```

hedera-mirror-importer/src/main/resources/db/migration/v2/R__02_temp_table_distribution.sql
```diff
@@ -20,6 +20,7 @@ call create_distributed_table_safe('entity_state_start', 'id', 'entity');
 call create_distributed_table_safe('entity_temp', 'id', 'entity');
 call create_distributed_table_safe('nft_allowance_temp', 'owner', 'nft_allowance');
 call create_distributed_table_safe('nft_temp', 'token_id', 'nft');
+call create_distributed_table_safe('node_temp', 'node_id', 'node');
```
xin-hedera, line 23: remove this too since we shouldn't distribute it

```suggestion
```

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/transactionhandler/NodeDeleteTransactionHandler.java
mgoelswirlds: I need to still store the transaction information, but will add the check for parseNode.

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/record/transactionhandler/NodeCreateTransactionHandlerTest.java
mgoelswirlds: In the parent test class there was a test.

xin-hedera: I'm confused, because this specific test case was disabled by `@Disabled` for no good reason.

hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/entity/AbstractNode.java
```diff
@@ -0,0 +1,48 @@
...
+
+    @ToString.Exclude
+    private byte[] adminKey;
+
+    @Column(updatable = false)
+    private long createdTimestamp;
+
+    private boolean deleted;
+
+    @jakarta.persistence.Id
```
xin-hedera, line 44: nit, since there is no longer name clash with the deleted `Id` class

```suggestion
    @Id
```

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/transactionhandler/NodeCreateTransactionHandler.java
xin-hedera: it looks more integrated with the `isSuccessful()` check inside `parseNode`

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/record/transactionhandler/NodeCreateTransactionHandlerTest.java
mgoelswirlds: updateTransactionUnsuccessful() would fail in the parent class since we were not returning for a failed transaction earlier. That's why I had to override and disable this particular test. Now that I do fail in the transactionHandler in case of a failed transaction , this is no longer relevant.

hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/entity/AbstractNode.java
```diff
@@ -0,0 +1,48 @@
...
+@NoArgsConstructor
+@SuperBuilder(toBuilder = true)
+@Upsertable(history = true, skipPartialUpdate = true)
+public abstract class AbstractNode implements History {
+
+    @ToString.Exclude
+    private byte[] adminKey;
+
+    @Column(updatable = false)
+    private long createdTimestamp;
```
xin-hedera, line 40: this needs to be `Long` to make upsert of a partial update object to work

```suggestion
    private Long createdTimestamp;
```

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/record/entity/EntityRecordItemListenerNodeTest.java
```diff
@@ -114,13 +122,37 @@ void nodeCreate() {
...
+        expectedNode.setAdminKey(
+                recordItem.getTransactionBody().getNodeUpdate().getAdminKey().toByteArray());
+        expectedNode.setCreatedTimestamp(0);
```
xin-hedera, line 135: remove it after changing the type to `Long`, the non-primitive equivalent
```suggestion
```

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/record/entity/EntityRecordItemListenerNodeTest.java
```diff
@@ -114,13 +122,37 @@ void nodeCreate() {
...
+    }
 
+    @Test
+    void nodeUpdateWithoutAdminKeyUpdate() {
+        entityProperties.getPersist().setNodes(true);
+        var recordItem = recordItemBuilder.nodeUpdate().build();
+        var expectedNode = getExpectedNode(recordItem);
+        expectedNode.setCreatedTimestamp(0);
+        expectedNode.setAdminKey(
+                recordItem.getTransactionBody().getNodeUpdate().getAdminKey().toByteArray());
```
xin-hedera, line 155: test function says without admin key update, and here it expects admin key set

There is a common issue for both `nodeUpdate` and `nodeUpdateWithoutAdminKeyUpdate`, that we should first insert a row for a node to the db using `domainiBuilder.node().persist()`, and then ingest a node update tx for the same node to verify that

- for the update with admin key change, the new admin key is persisted to `node` table, and there is a historical row with the old admin key in `node_history`
- for the update without admin key change, that the admin key in `node` table remains the same, and there is a historical row in `node_history`

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/record/transactionhandler/NodeDeleteTransactionHandlerTest.java
```diff
@@ -82,5 +80,9 @@ void nodeDeleteTransaction(boolean isPersist) {
         // then
         assertThat(transaction.getTransactionBytes()).containsExactly(transactionBytes);
         assertThat(transaction.getTransactionRecordBytes()).containsExactly(transactionRecordBytes);
+        verify(entityListener, times(BooleanUtils.toInteger(isPersist))).onNode(assertArg(t -> assertThat(t)
+                .isNotNull()
+                .returns(recordItem.getTransactionRecord().getReceipt().getNodeId(), Node::getNodeId)
+                .returns(isPersist, Node::isDeleted)));
```
xin-hedera, line 86: the logic here is really bizarre. If you don't like the idea of splitting it to two test cases, I'd rather have an explicit if / else to assert the expected result

```java
  if (isPersist) {
    // assert that entityListener.onNode is called once with matching Node object
  } else {
    // assert that entityListener.onNode is not called at all
  }
```

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/record/transactionhandler/NodeUpdateTransactionHandlerTest.java
```diff
@@ -75,5 +79,17 @@ void nodeUpdateTransaction(boolean isPersist) {
         // then
         assertThat(transaction.getTransactionBytes()).containsExactly(transactionBytes);
         assertThat(transaction.getTransactionRecordBytes()).containsExactly(transactionRecordBytes);
+        verify(entityListener, times(BooleanUtils.toInteger(isPersist))).onNode(assertArg(t -> assertThat(t)
```
xin-hedera, line 82: same comment as in the other handler test

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/transactionhandler/NodeCreateTransactionHandler.java
```diff
@@ -46,5 +51,21 @@ protected void doUpdateTransaction(Transaction transaction, RecordItem recordIte
...
+        long consensusTimestamp = recordItem.getConsensusTimestamp();
+        entityListener.onNode(Node.builder()
+                .adminKey(nodeCreate.getAdminKey().toByteArray())
+                .createdTimestamp(consensusTimestamp)
+                .nodeId(recordItem.getTransactionRecord().getReceipt().getNodeId())
```
steven-sheehy, line 67: Even though it's the default, would be clearer to explicitly set deleted to false here.

xin-hedera (changes requested):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/transactionhandler/NodeDeleteTransactionHandler.java
```diff
@@ -41,5 +45,18 @@ protected void doUpdateTransaction(Transaction transaction, RecordItem recordIte
         }
         transaction.setTransactionBytes(recordItem.getTransaction().toByteArray());
         transaction.setTransactionRecordBytes(recordItem.getTransactionRecord().toByteArray());
+        if (!recordItem.isSuccessful()) {
+            return;
+        }
```
steven-sheehy, line 50: It would be clearer if the success check is inside the parseNode methods

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/transactionhandler/NodeDeleteTransactionHandler.java
```diff
@@ -41,5 +45,18 @@ protected void doUpdateTransaction(Transaction transaction, RecordItem recordIte
...
+        parseNode(recordItem);
+    }
+
+    private void parseNode(RecordItem recordItem) {
+        long consensusTimestamp = recordItem.getConsensusTimestamp();
+        entityListener.onNode(Node.builder()
+                .deleted(true)
+                .nodeId(recordItem.getTransactionRecord().getReceipt().getNodeId())
```
steven-sheehy, line 58: Double check they will be populating this in the receipt for update and deletes. To be safe, we can get it from body.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/transactionhandler/NodeUpdateTransactionHandler.java
```diff
@@ -46,5 +51,25 @@ protected void doUpdateTransaction(Transaction transaction, RecordItem recordIte
...
+        long consensusTimestamp = recordItem.getConsensusTimestamp();
+
+        var node = new Node();
+
+        if (nodeUpdate.hasAdminKey()) {
+            node.setAdminKey(nodeUpdate.getAdminKey().toByteArray());
+        }
+
+        node.setNodeId(recordItem.getTransactionRecord().getReceipt().getNodeId());
```
steven-sheehy, line 70: We should set deleted to false

hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/entity/AbstractNode.java
```diff
@@ -0,0 +1,48 @@
...
+import lombok.Data;
+import lombok.NoArgsConstructor;
+import lombok.ToString;
+import lombok.experimental.SuperBuilder;
+
+@Data
+@MappedSuperclass
+@NoArgsConstructor
+@SuperBuilder(toBuilder = true)
+@Upsertable(history = true, skipPartialUpdate = true)
```
steven-sheehy, line 33: I think it would be useful to have partial updates. That way mirror nodes that start up now and miss node create can self correct on next node update. We should remove skipPartialUpdate so it reverts to false. Then change schema to allow null for created_timestamp and use `Long`.

Add a partial update test to SqlEntityListenerTest.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/entity/sql/SqlEntityListener.java
```diff
@@ -578,6 +584,11 @@ private NftAllowance mergeNftAllowance(NftAllowance previous, NftAllowance curre
         return current;
     }
 
+    private Node mergeNode(Node previous, Node current) {
+        previous.setTimestampUpper(current.getTimestampLower());
```
steven-sheehy, line 588: Missing logic here to merge created timestamp and admin key. Also means missing test coverage for updates in single record file.

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/record/entity/sql/SqlEntityListenerTest.java
```diff
@@ -1723,6 +1726,75 @@ void onTransactionWithNftTransferMint(boolean trackBalance) {
...
+    @ParameterizedTest
+    void onNodeHistory(int commitIndex) {
+        // given
+        var nodeCreate1 = domainBuilder.node().get();
+        var nodeUpdate1 = nodeCreate1.toBuilder()
+                .adminKey(domainBuilder.key())
+                .timestampRange(Range.atLeast(domainBuilder.timestamp()))
+                .build();
+
+        var nodeCreate2 = domainBuilder.node().get();
```
steven-sheehy, line 1754: Why do we have a second node create? For this history test we should only test a single node ID. We just need a create, update and a delete for a single node ID where each of those occurs at different commit boundaries. You're welcome to add a separate test for multi-node ID scenario that doesn't use the commitIndex concept.

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/record/entity/sql/SqlEntityListenerTest.java
```diff
@@ -1723,6 +1726,75 @@ void onTransactionWithNftTransferMint(boolean trackBalance) {
...
+        var nodeDelete1 = nodeUpdate1.toBuilder()
+                .deleted(true)
```
steven-sheehy, line 1756: We should not have createdTimestamp populated for updates and deletes, while still verifying they're populated in db.

steven-sheehy (changes requested):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/transactionhandler/NodeDeleteTransactionHandler.java
mgoelswirlds: I have asked Iris about this. Will confirm as soon as I hear back.

steven-sheehy: At least in the proto it says it will be set for delete and update. https://github.com/hashgraph/hedera-protobufs/blob/a644ad4e1bc9c75de92f6d5f9d772100fb8c0ff4/services/transaction_receipt.proto#L175

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/transactionhandler/NodeDeleteTransactionHandler.java
steven-sheehy: Now there's duplicate success checks...

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/entity/sql/SqlEntityListener.java
jnels124: actually I guess a delete coming in shouldn't wipe out admin key nevermind

xin-hedera: We need to merge. e.g.,suppose these happen in the same record file,

- node update set a new admin key
- node update set some field other than admin key

The history row from the first update should be fine with the correct updated admin key.

the row in node table however will end up with the old admin key since the upsert does a coelesce of null and the admin key in the table before ingesting the record file

hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/entity/AbstractNode.java
```diff
@@ -0,0 +1,65 @@
...
+import lombok.NoArgsConstructor;
+import lombok.ToString;
+import lombok.experimental.SuperBuilder;
+
+@Data
+@IdClass(AbstractNode.Id.class)
+@MappedSuperclass
+@NoArgsConstructor
+@SuperBuilder(toBuilder = true)
+@Upsertable(history = true)
```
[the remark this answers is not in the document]
jascks: I see this convo is resolved, but I don't see that suggested change was adopted. Was there an offline discussion or did this get reverted by accident?

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/record/entity/EntityRecordItemListenerNodeTest.java
```diff
@@ -114,37 +122,115 @@ void nodeCreate() {
...
                 .returns(recordItem.getTransactionRecord().toByteArray(), Transaction::getTransactionRecordBytes);
+        softly.assertThat(nodeRepository.findAll()).containsExactly(expectedNode);
+        softly.assertThat(findHistory(Node.class)).containsExactly(node);
+    }
+
+    @Test
+    void nodeUpdateWithoutAdminKeyUpdate() {
+        entityProperties.getPersist().setNodes(true);
+
+        var recordItem = recordItemBuilder.nodeUpdate().build();
```
xin-hedera, line 167: the transaction body should have admin key cleared.

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/record/entity/EntityRecordItemListenerNodeTest.java
```diff
@@ -114,37 +122,115 @@ void nodeCreate() {
...
+        var nodeUpdate = recordItem.getTransactionBody().getNodeUpdate();
+        var timestamp = recordItem.getConsensusTimestamp() - 1;
+        var node = domainBuilder
+                .node()
+                .customize(n -> n.adminKey(nodeUpdate.getAdminKey().toByteArray())
+                        .createdTimestamp(timestamp)
+                        .nodeId(nodeUpdate.getNodeId())
+                        .timestampRange(Range.atLeast(timestamp)))
+                .persist();
```
xin-hedera, line 176: should either do `domainBuilder.node().customize(...).persist()` or use `recordItemBuilder.key()`

xin-hedera (changes requested):

jascks (approved):
LGTM

xin-hedera (approved):
LGTM

steven-sheehy (approved):
LGTM

PR #8912, opened 2024-08-01
Refactor ContractCallDynamicCallsTest

bilyana-gospodinova:
**Description**:
With this PR the `ContractCallDynamicCallsTest` integration test is refactored using the web3j plugin and the big enum is now replaced with multiple junit tests so that it is easier to debug a particular scenario. This way also the number of persisted entities per test is reduced significantly and now we persist only the ones that are needed in the particular test. With this implementation the execution time of the `ContractCallDynamicCallsTest` is reduced 5 times (from about 50 seconds to 9-10 seconds, tested locally).

Fixes https://github.com/hashgraph/hedera-mirror-node/issues/8449

### DIFF
--- a/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/Web3IntegrationTest.java
+++ b/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/Web3IntegrationTest.java
@@ -16,7 +16,7 @@
 
 package com.hedera.mirror.web3;
 
-import static com.hedera.mirror.web3.service.ContractCallTestUtil.TRANSACTION_GAS_LIMIT;
+import static com.hedera.mirror.web3.utils.ContractCallTestUtil.TRANSACTION_GAS_LIMIT;
 import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
 
 import com.hedera.mirror.common.config.CommonIntegrationTest;
--- a/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/Web3IntegrationTest.java
+++ b/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/Web3IntegrationTest.java
@@ -16,6 +16,7 @@
 
 package com.hedera.mirror.web3;
 
+import static com.hedera.mirror.web3.service.ContractCallTestUtil.TRANSACTION_GAS_LIMIT;
 import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
 
 import com.hedera.mirror.common.config.CommonIntegrationTest;
@@ -58,18 +59,16 @@ public abstract class Web3IntegrationTest extends CommonIntegrationTest {
     }
 
     protected ContractExecutionParameters getContractExecutionParameters(
-            final RemoteFunctionCall<TransactionReceipt> functionCall,
-            final Contract contract,
-            final Address senderAddress) {
+            final RemoteFunctionCall<TransactionReceipt> functionCall, final Contract contract) {
         return ContractExecutionParameters.builder()
                 .block(BlockType.LATEST)
                 .callData(Bytes.fromHexString(functionCall.encodeFunctionCall()))
                 .callType(CallType.ETH_CALL)
-                .gas(15_000_000L)
+                .gas(TRANSACTION_GAS_LIMIT)
                 .isEstimate(false)
                 .isStatic(false)
                 .receiver(Address.fromHexString(contract.getContractAddress()))
-                .sender(new HederaEvmAccount(senderAddress))
+                .sender(new HederaEvmAccount(Address.wrap(Bytes.wrap(domainBuilder.evmAddress()))))
                 .value(0L)
                 .build();
     }
--- a/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/Web3IntegrationTest.java
+++ b/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/Web3IntegrationTest.java
@@ -16,25 +16,12 @@
 
 package com.hedera.mirror.web3;
 
-import static com.hedera.mirror.web3.utils.ContractCallTestUtil.TRANSACTION_GAS_LIMIT;
-import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
-
 import com.hedera.mirror.common.config.CommonIntegrationTest;
-import com.hedera.mirror.web3.common.ContractCallContext;
 import com.hedera.mirror.web3.evm.contracts.execution.MirrorEvmTxProcessor;
 import com.hedera.mirror.web3.evm.store.Store;
-import com.hedera.mirror.web3.service.model.CallServiceParameters.CallType;
-import com.hedera.mirror.web3.service.model.ContractExecutionParameters;
-import com.hedera.mirror.web3.viewmodel.BlockType;
-import com.hedera.node.app.service.evm.store.models.HederaEvmAccount;
 import jakarta.annotation.Resource;
-import org.apache.tuweni.bytes.Bytes;
-import org.hyperledger.besu.datatypes.Address;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.springframework.boot.test.mock.mockito.SpyBean;
-import org.web3j.protocol.core.RemoteFunctionCall;
-import org.web3j.protocol.core.methods.response.TransactionReceipt;
-import org.web3j.tx.Contract;
 
 @ExtendWith(ContextExtension.class)
 public abstract class Web3IntegrationTest extends CommonIntegrationTest {
@@ -44,32 +31,4 @@ public abstract class Web3IntegrationTest extends CommonIntegrationTest {
 
     @Resource
     protected Store store;
-
-    @SuppressWarnings("try")
-    protected long gasUsedAfterExecution(final ContractExecutionParameters serviceParameters) {
-        return ContractCallContext.run(ctx -> {
-            ctx.initializeStackFrames(store.getStackedStateFrames());
-            long result = processor
-                    .execute(serviceParameters, serviceParameters.getGas())
-                    .getGasUsed();
-
-            assertThat(store.getStackedStateFrames().height()).isEqualTo(1);
-            return result;
-        });
-    }
-
-    protected ContractExecutionParameters getContractExecutionParameters(
-            final RemoteFunctionCall<TransactionReceipt> functionCall, final Contract contract) {
-        return ContractExecutionParameters.builder()
-                .block(BlockType.LATEST)
-                .callData(Bytes.fromHexString(functionCall.encodeFunctionCall()))
-                .callType(CallType.ETH_CALL)
-                .gas(TRANSACTION_GAS_LIMIT)
-                .isEstimate(false)
-                .isStatic(false)
-                .receiver(Address.fromHexString(contract.getContractAddress()))
-                .sender(new HederaEvmAccount(Address.wrap(Bytes.wrap(domainBuilder.evmAddress()))))
-                .value(0L)
-                .build();
-    }
 }
--- a/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/Web3IntegrationTest.java
+++ b/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/Web3IntegrationTest.java
@@ -16,12 +16,24 @@
 
 package com.hedera.mirror.web3;
 
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
+
 import com.hedera.mirror.common.config.CommonIntegrationTest;
+import com.hedera.mirror.web3.common.ContractCallContext;
 import com.hedera.mirror.web3.evm.contracts.execution.MirrorEvmTxProcessor;
 import com.hedera.mirror.web3.evm.store.Store;
+import com.hedera.mirror.web3.service.model.CallServiceParameters.CallType;
+import com.hedera.mirror.web3.service.model.ContractExecutionParameters;
+import com.hedera.mirror.web3.viewmodel.BlockType;
+import com.hedera.node.app.service.evm.store.models.HederaEvmAccount;
 import jakarta.annotation.Resource;
+import org.apache.tuweni.bytes.Bytes;
+import org.hyperledger.besu.datatypes.Address;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.springframework.boot.test.mock.mockito.SpyBean;
+import org.web3j.protocol.core.RemoteFunctionCall;
+import org.web3j.protocol.core.methods.response.TransactionReceipt;
+import org.web3j.tx.Contract;
 
 @ExtendWith(ContextExtension.class)
 public abstract class Web3IntegrationTest extends CommonIntegrationTest {
@@ -31,4 +43,34 @@ public abstract class Web3IntegrationTest extends CommonIntegrationTest {
 
     @Resource
     protected Store store;
+
+    @SuppressWarnings("try")
+    protected long gasUsedAfterExecution(final ContractExecutionParameters serviceParameters) {
+        return ContractCallContext.run(ctx -> {
+            ctx.initializeStackFrames(store.getStackedStateFrames());
+            long result = processor
+                    .execute(serviceParameters, serviceParameters.getGas())
+                    .getGasUsed();
+
+            assertThat(store.getStackedStateFrames().height()).isEqualTo(1);
+            return result;
+        });
+    }
+
+    protected ContractExecutionParameters getContractExecutionParameters(
+            final RemoteFunctionCall<TransactionReceipt> functionCall,
+            final Contract contract,
+            final Address senderAddress) {
+        return ContractExecutionParameters.builder()
+                .block(BlockType.LATEST)
+                .callData(Bytes.fromHexString(functionCall.encodeFunctionCall()))
+                .callType(CallType.ETH_CALL)
+                .gas(15_000_000L)
+                .isEstimate(false)
+                .isStatic(false)
+                .receiver(Address.fromHexString(contract.getContractAddress()))
+                .sender(new HederaEvmAccount(senderAddress))
+                .value(0L)
+                .build();
+    }
 }
--- a/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
+++ b/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
@@ -31,11 +31,11 @@ import com.hedera.mirror.web3.web3j.TestWeb3jService;
 import com.hedera.mirror.web3.web3j.TestWeb3jService.Web3jTestConfiguration;
 import com.hedera.node.app.service.evm.store.models.HederaEvmAccount;
 import java.util.concurrent.atomic.AtomicLong;
+import lombok.RequiredArgsConstructor;
 import org.apache.tuweni.bytes.Bytes;
 import org.hyperledger.besu.datatypes.Address;
 import org.junit.jupiter.api.AfterEach;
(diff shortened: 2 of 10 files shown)


hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallDynamicCallsTest.java
```diff
@@ -16,266 +16,1701 @@
...
+                        .metadata(NFT_METADATA.getBytes())
+                        .accountId(ownerEntityId)
+                        .tokenId(nftEntity.getId())
+                        .serialNumber(1))
+                .persist();
+        return token;
+    }
+
+    private EntityId treasureEntityPersist() {
+        return accountPersist(TREASURY_ADDRESS);
```
IvanKavaldzhiev, line 1643: Try to adapt tests to work with dynamic data as much as possible, e.g. not relying on hardcoded addresses. An exclusion might be specific aliases, that are expected as return result for example. I believe a lot of these utility methods can be removed in favour of dynamically creating data via `domainBuilder` inside the tests.

Even if we have repetitions and can extract common `domainBuilder` constructs, they should be again dynamic and we shouldn't pass static constants as `Address` or `EntityId`

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallDynamicCallsTest.java
```diff
@@ -16,266 +16,1701 @@
...
+    }
+
+    private EntityId accountPersist(final Address address) {
+        final var accountEntityId = entityIdFromEvmAddress(address);
+        domainBuilder
+                .entity()
+                .customize(e -> e.id(accountEntityId.getId())
+                        .num(accountEntityId.getNum())
+                        .evmAddress(null)
+                        .alias(toEvmAddress(accountEntityId))
```
IvanKavaldzhiev, line 1665: If we don't expect particular `alias` to be used, we can leave the `domainBuilder` to randomly generate one.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallDynamicCallsTest.java
```diff
@@ -16,266 +16,1701 @@
...
+            final EntityId ownerEntityId,
+            final boolean approvedForAll) {
+        domainBuilder
+                .nftAllowance()
+                .customize(a -> a.tokenId(tokenEntityId.getId())
+                        .spender(spenderEntityId.getId())
+                        .owner(ownerEntityId.getId())
+                        .payerAccountId(ownerEntityId)
+                        .approvedForAll(approvedForAll))
+                .persist();
```
IvanKavaldzhiev, line 1714: In none of the tests in this class we have asserts for concrete values as an output. The tests are mainly testing complex nested precompile calls and the correctness of their execution in combination.

Thus, I think we can remove a lot of the utility methods and especially the parameters that set specific static values to be added into the entities built with the domainBuilder. I don't think we need to rely on hardcoded constants but let most of the fields to be auto generated.

In this concrete example if we don't pass:
```
.spender(spenderEntityId.getId())
.owner(ownerEntityId.getId())
.payerAccountId(ownerEntityId),
```
                        
then the `spender`, `owner` and `payerAccountId` will be automatically generated and populated by the `domainBuilder`.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/web3j/TestWeb3jService.java
```diff
@@ -170,16 +185,29 @@ private EthSendTransaction sendEthCall(RawTransaction rawTransaction, String tra
 
     private EthCall ethCall(List<Transaction> reqParams, Request request) {
         var transaction = reqParams.get(0);
+        RawTransaction rawTransaction;
+        if (isAddressAlias(transaction.getTo())) {
+            // If the "to" address is an alias, the transaction could not be decoded,
+            // so we need to override the "to" field in order to generate the correct
+            // transaction hash but preserve the alias and pass it in the request.
+            rawTransaction = copyTransactionWithReceiver(transaction, receiver);
```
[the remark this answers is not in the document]
bilyana-gospodinova: The method is removed. It seems that `toString` did not work as expected so the RLP decoding used to fail. We still need to convert the `Transaction` to `RawTransaction` though.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallDynamicCallsTest.java
bilyana-gospodinova: Comment addressed. Thanks.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallDynamicCallsTest.java
IvanKavaldzhiev: I see, if it's because of the needed relations between entities we should be good.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/web3j/TestWeb3jService.java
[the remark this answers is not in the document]
bilyana-gospodinova: As pointed out by @IvanKavaldzhiev the conversion is not needed at all. I cleared the implementation that is not needed.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractDebugServiceTest.java
```diff
@@ -304,4 +308,180 @@ private void tokenAccountsPersist(EntityId contractId, List<EntityId> tokenIds)
...
+                    new byte[][] {ByteString.copyFromUtf8("firstMeta").toByteArray()},
+                    OWNER_ADDRESS
+                },
+                null),
+        BURN_FUNGIBLE_TOKEN(
+                "burnTokenGetTotalSupplyAndBalanceOfTreasury",
+                new Object[] {NOT_FROZEN_FUNGIBLE_TOKEN_ADDRESS, 12L, new long[0], TREASURY_ADDRESS},
+                null),
+        BURN_NFT(
+                "burnTokenGetTotalSupplyAndBalanceOfTreasury",
```
edwin-greene, line 333: Can you change the name to something like `burnNft`? It has the same name as `BURN_FUNGIBLE_TOKEN`

bilyana-gospodinova: I just moved `DynamicCallsContractFunctions` from `ContractCallDynamicCallsTest` to `ContractDebugServiceTest` as it uses the enum as well. The latter one is not refactored yet, so this enum will probably be removed from there as well when we get to it. The idea behind the same function names is that they will be used for both fungible tokens and NFTs in the same test so that it can be parameterized as the test body is the same.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractDebugServiceTest.java
```diff
@@ -304,4 +308,180 @@ private void tokenAccountsPersist(EntityId contractId, List<EntityId> tokenIds)
...
+
+    @Getter
+    @RequiredArgsConstructor
+    enum DynamicCallsContractFunctions implements ContractFunctionProviderEnum {
+        MINT_FUNGIBLE_TOKEN(
+                "mintTokenGetTotalSupplyAndBalanceOfTreasury",
+                new Object[] {NOT_FROZEN_FUNGIBLE_TOKEN_ADDRESS, 100L, new byte[0][0], TREASURY_ADDRESS},
+                null),
+        MINT_NFT(
+                "mintTokenGetTotalSupplyAndBalanceOfTreasury",
```
edwin-greene, line 320: Can change name to include nft to differentiate with fungible test.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractDebugServiceTest.java
edwin-greene: Nevermind, I see that these are function names.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallDynamicCallsTest.java
```diff
@@ -16,266 +16,1619 @@
...
+@SuppressWarnings("unchecked")
+@RequiredArgsConstructor
+@TestInstance(PER_CLASS)
+class ContractCallDynamicCallsTest extends Web3IntegrationTest {
+
+    private final TestWeb3jService testWeb3jService;
+
+    @BeforeEach
+    void setup() {
+        recordFilePersist();
```
edwin-greene, line 71: Looks like this can just be replaced with `domainBuilder.recordFile().persist();` and then the method can be removed.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallDynamicCallsTest.java
```diff
@@ -16,266 +16,1619 @@
...
+
+        testWeb3jService.setEstimateGas(false);
+        final var actualGasUsed = gasUsedAfterExecution(getContractExecutionParameters(functionCall, contract));
+
+        // Then
+        assertThat(isWithinExpectedGasRange(estimateGasUsedResult, actualGasUsed))
+                .withFailMessage(ESTIMATE_GAS_ERROR_MESSAGE, estimateGasUsedResult, actualGasUsed)
+                .isTrue();
+    }
+
```
edwin-greene, line 149: Many of the tests look to be pairs covering the same ground, except one version includes estimate gas assertions. Can the non-estimate gas tests be removed if the `assertDoesNotThrow(() -> { functionCall.send(); });` is added to the estimate gas tests?

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/web3j/TestWeb3jService.java
```diff
@@ -288,22 +295,26 @@ private EthGetTransactionCount ethGetTransactionCount() {
     }
 
     private EthGetTransactionReceipt getTransactionReceipt(Request request) {
-        final var trxHash = request.getParams().get(0).toString();
-        final var res = new EthGetTransactionReceipt();
+        final var transactionHash = request.getParams().get(0).toString();
```
jnels124, line 298: nit: can be replaced with `request.getParams().getFirst().toString()`

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallTestUtil.java
```diff
@@ -0,0 +1,48 @@
...
+package com.hedera.mirror.web3.service;
+
+import static com.hedera.node.app.service.evm.utils.EthSigsUtils.recoverAddressFromPubKey;
+
+import com.google.protobuf.ByteString;
+import org.apache.tuweni.bytes.Bytes;
+import org.bouncycastle.util.encoders.Hex;
+import org.hyperledger.besu.datatypes.Address;
+
+public class ContractCallTestUtil {
```
IvanKavaldzhiev, line 26: You can add a private constructor, since it's a utility class

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallDynamicCallsTest.java
bilyana-gospodinova: Comment addressed.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallDynamicCallsTest.java
bilyana-gospodinova: I combined the ethCall and estimateGas tests into one. I did not plan to do that initially so that the junit tests can verify only one thing per test but you are right that this way we have a lot of duplication. Thanks.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/web3j/TestWeb3jService.java
bilyana-gospodinova: Comment addressed. Thanks.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallTestUtil.java
bilyana-gospodinova: I added the `@UtilityClass` annotation. Thanks.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractDebugServiceTest.java
```diff
@@ -304,4 +308,180 @@ private void tokenAccountsPersist(EntityId contractId, List<EntityId> tokenIds)
             tokenAccountPersist(contractId, tokenId, TokenFreezeStatusEnum.UNFROZEN);
         }
     }
+
+    @Getter
+    @RequiredArgsConstructor
+    enum DynamicCallsContractFunctions implements ContractFunctionProviderEnum {
```
steven-sheehy, line 314: Mark private to ensure it doesn't get used elsewhere. Should we open a ticket to also refactor `ContractDebugServiceTest`? Or is there another way to share test cases with debug like a shared `@Nested` class?

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/Web3IntegrationTest.java
```diff
@@ -31,4 +44,32 @@ public abstract class Web3IntegrationTest extends CommonIntegrationTest {
 
     @Resource
     protected Store store;
+
+    @SuppressWarnings("try")
+    protected long gasUsedAfterExecution(final ContractExecutionParameters serviceParameters) {
```
steven-sheehy, line 49: Probably we should have an `AbstractContractCallServiceTest` with these methods and any common annotations and setup.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallDynamicCallsTest.java
```diff
@@ -16,266 +16,869 @@
...
+    @BeforeEach
+    void setup() {
+        domainBuilder.recordFile().persist();
+    }
+
+    @AfterEach
+    void cleanup() {
+        testWeb3jService.setEstimateGas(false);
+    }
```
steven-sheehy, line 81: These and the `@Import` can move to a parent `AbstractContractCallServiceTest`.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallDynamicCallsTest.java
```diff
@@ -16,266 +16,869 @@
...
+import org.junit.jupiter.params.provider.CsvSource;
+import org.springframework.context.annotation.Import;
+import org.web3j.protocol.core.RemoteFunctionCall;
+import org.web3j.protocol.core.methods.response.TransactionReceipt;
 
-class ContractCallDynamicCallsTest extends ContractCallTestSetup {
+@Import(Web3jTestConfiguration.class)
+@SuppressWarnings("unchecked")
+@RequiredArgsConstructor
+@TestInstance(PER_CLASS)
```
steven-sheehy, line 68: Why is this needed?

steven-sheehy (changes requested):

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractDebugServiceTest.java
bilyana-gospodinova: Yes, I think we need tickets to refactor `ContractDebugServiceTest` and `OpcodeServiceTest` as well using web3j.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallDynamicCallsTest.java
bilyana-gospodinova: Actually, it is not. We don't need to keep any state.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/Web3IntegrationTest.java
bilyana-gospodinova: `AbstractContractCallServiceTest` is now created.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallDynamicCallsTest.java
bilyana-gospodinova: Moved as requested.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractDebugServiceTest.java
bilyana-gospodinova: The issue for `ContractDebugServiceTest` - https://github.com/hashgraph/hedera-mirror-node/issues/8944.
The issue for `OpcodeServiceTest` - https://github.com/hashgraph/hedera-mirror-node/issues/8945.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
```diff
@@ -0,0 +1,104 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.Import;
+import org.web3j.protocol.core.RemoteFunctionCall;
+import org.web3j.protocol.core.methods.response.TransactionReceipt;
+import org.web3j.tx.Contract;
+
+@Import(Web3jTestConfiguration.class)
+@SuppressWarnings("unchecked")
+public abstract class AbstractContractCallServiceTest extends Web3IntegrationTest {
```
steven-sheehy, line 46: Can be package private

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
```diff
@@ -0,0 +1,104 @@
...
+
+    @Autowired
+    protected TestWeb3jService testWeb3jService;
```
steven-sheehy, line 49: Use constructor injection for parent classes with `@RequiredArgsConstructor`

```suggestion
    protected final TestWeb3jService testWeb3jService;
```

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
bilyana-gospodinova: Done

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
bilyana-gospodinova: Done

steven-sheehy (approved):
LGTM

IvanKavaldzhiev (approved):
LGTM

PR #8955, opened 2024-08-06
Refactor ContractCallNestedCallsTest

bilyana-gospodinova:
**Description**:
With this PR the `ContractCallNestedCallsTest` integration test is refactored using the web3j plugin and the big enum is now replaced with multiple junit tests so that it is easier to debug a particular scenario. This way also the number of persisted entities per test is reduced significantly and now we persist only the ones that are needed in the particular test. With this implementation the execution time of the `ContractCallNestedCallsTest` is reduced 3 times (from about 38 seconds to 13 seconds, tested locally).

The `NestedEthCallContractFunctionsNegativeCases` and `NestedEthCallContractFunctions` enums are moved from `ContractCallNestedCallsTest` to `ContractDebugServiceTest` and `OpcodeServiceTest` respectively as they still need these enums and these test classes are not refactored yet.

Fixes https://github.com/hashgraph/hedera-mirror-node/issues/8455

**Checklist**


hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallNestedCallsTest.java
```diff
@@ -16,935 +16,553 @@
...
-import org.junit.jupiter.params.provider.EnumSource;
+import org.junit.jupiter.params.provider.CsvSource;
 
-class ContractCallNestedCallsTest extends ContractCallTestSetup {
+class ContractCallNestedCallsTest extends AbstractContractCallServiceTest {
+
+    private final long CREATE_TOKEN_VALUE = 3070 * 100_000_000L;
+    private final String EXPECTED_RESULT_NEGATIVE_TESTS = "hardcodedResult";
+
+    public ContractCallNestedCallsTest(TestWeb3jService testWeb3jService) {
```
kselveliev, line 50: we can remove the public modifier here

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallNestedCallsTest.java
```diff
@@ -16,935 +16,553 @@
...
+                    NON_FUNGIBLE_UNIQUE,    ECDSA_SECPK256K1,           WIPE_KEY,
+                    NON_FUNGIBLE_UNIQUE,    ECDSA_SECPK256K1,           SUPPLY_KEY,
+                    NON_FUNGIBLE_UNIQUE,    ECDSA_SECPK256K1,           FEE_SCHEDULE_KEY,
+                    NON_FUNGIBLE_UNIQUE,    ECDSA_SECPK256K1,           PAUSE_KEY,
+                    NON_FUNGIBLE_UNIQUE,    DELEGATABLE_CONTRACT_ID,    ADMIN_KEY,
+                    NON_FUNGIBLE_UNIQUE,    DELEGATABLE_CONTRACT_ID,    KYC_KEY,
+                    NON_FUNGIBLE_UNIQUE,    DELEGATABLE_CONTRACT_ID,    FREEZE_KEY,
+                    NON_FUNGIBLE_UNIQUE,    DELEGATABLE_CONTRACT_ID,    WIPE_KEY,
+                    NON_FUNGIBLE_UNIQUE,    DELEGATABLE_CONTRACT_ID,    SUPPLY_KEY,
+                    NON_FUNGIBLE_UNIQUE,    DELEGATABLE_CONTRACT_ID,    FEE_SCHEDULE_KEY,
```
kselveliev, line 112: One suggestion here is instead of having many lines with NON_FUNGIBLE_UNIQUE or FUNGIBLE_COMMON.
We can split the method into updatedFungibleTokenKeys.. and updateNonFungibleTokenKeys...
And with that we can remove the final TokenTypeEnum tokenType parameter and remove the if statement final var tokenEntityId = tokenType == TokenTypeEnum.FUNGIBLE_COMMON ? fungibleTokenPersist() : nftPersist(); 

And it will just be var tokenEntityId = fungibleTokenPersist() or tokenEntityId = nftPersist(). I think it is a bit cleaner that way

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallNestedCallsTest.java
```diff
@@ -16,935 +16,553 @@
...
+
+        verifyEthCallAndEstimateGas(functionCall, contract);
+    }
+
+    @ParameterizedTest
+    @CsvSource(textBlock = """
+                true, true, true
+                false, false, false
+                """)
+    void createFungibleTokenAndGetIsTokenAndGetDefaultFreezeStatusAndGetDefaultKycStatus(
```
kselveliev, line 285: Can't we just merge the below method into this createFungibleTokenAndGetIsTokenAndGetDefaultFreezeStatusAndGetDefaultKycStatusInheritKey

Probably if we add one more boolean in the csv like: 
true, true, true, false
false, false, false, false
false, false, false, true

The last boolean is the inherit key.

I think with this approach we can delete createFungibleTokenAndGetIsTokenAndGetDefaultFreezeStatusAndGetDefaultKycStatusInheritKey method.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallNestedCallsTest.java
```diff
@@ -16,935 +16,553 @@
...
+
+        final var functionCall = contract.send_createNFTAndGetIsTokenAndGetDefaultFreezeStatusAndGetDefaultKycStatus(
+                tokenInfo, BigInteger.valueOf(CREATE_TOKEN_VALUE));
+
+        verifyEthCallAndEstimateGasWithValue(
+                functionCall, contract, toAddress(treasuryEntity.getId()), CREATE_TOKEN_VALUE);
+    }
+
+    @Test
+    void createNFTAndGetIsTokenAndGetDefaultFreezeStatusAndGetDefaultKycStatusInheritKey() throws Exception {
```
kselveliev, line 377: This is probably same as the above comment.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallNestedCallsTest.java
bilyana-gospodinova: Correct. It seems I forgot to update the modifier since this method is generated. Thanks.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallNestedCallsTest.java
bilyana-gospodinova: Comment addressed.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallNestedCallsTest.java
bilyana-gospodinova: Comment addressed. Thanks.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallNestedCallsTest.java
bilyana-gospodinova: Comment addressed. Thanks.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallNestedCallsTest.java
```diff
@@ -16,935 +16,518 @@
...
-import org.junit.jupiter.params.provider.EnumSource;
+import org.junit.jupiter.params.provider.CsvSource;
 
-class ContractCallNestedCallsTest extends ContractCallTestSetup {
+class ContractCallNestedCallsTest extends AbstractContractCallServiceTest {
+
+    private static final long CREATE_TOKEN_VALUE = 3070 * 100_000_000L;
+    private static final String EXPECTED_RESULT_NEGATIVE_TESTS = "hardcodedResult";
+
+    private ContractCallNestedCallsTest(TestWeb3jService testWeb3jService) {
```
kselveliev, line 47: Making this package private fixes the code smell. I had the same issue in my pr.

bilyana-gospodinova: Yes, but we don't really need to be able to make instances of this class even in the same package, so this code smell does not make sense to me.

kselveliev: I agree :)

kselveliev (dismissed):

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallNestedCallsTest.java
steven-sheehy: We can go back to not using constructor injection for the parent class if it makes this code smell go away. Just use `@Resource` to inject `TestWeb3jService` into parent.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
```diff
@@ -56,6 +60,8 @@ void setup() {
     @AfterEach
     void cleanup() {
         testWeb3jService.setEstimateGas(false);
+        testWeb3jService.setValue(0L);
+        testWeb3jService.setSender(Address.fromHexString(""));
```
steven-sheehy, line 64: Would be better to do in the before each

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallNestedCallsTest.java
```diff
@@ -16,935 +16,518 @@
...
+                        tokenAddress.toHexString(), List.of(tokenKey), keyType.getKeyTypeNumeric())
+                .send();
+
+        // Then
+        assertThat(result).isEqualTo(keyValue);
+
+        final var functionCall = contract.send_updateTokenKeysAndGetUpdatedTokenKey(
+                tokenAddress.toHexString(), List.of(tokenKey), keyType.getKeyTypeNumeric());
+
+        verifyEthCallAndEstimateGas(functionCall, contract);
```
steven-sheehy, line 164: It seems wasteful to call once to verify result, a second time to get actual gas used, and N times to calculate estimate. Maybe in the future (not this PR) we should consider enhancing `ContractCallService.processCall()` to return a wrapper object instead of just a String that we can always attach both the result and gas used to to improve efficiency of the tests.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallNestedCallsTest.java
```diff
@@ -16,935 +16,518 @@
...
+            final boolean defaultKycStatus,
+            final boolean defaultFreezeStatus)
+            throws Exception {
+        // Given
+        final var treasuryEntity = accountPersistWithBalance(CREATE_TOKEN_VALUE);
+        final var contract = testWeb3jService.deploy(NestedCalls::deploy);
+        final var tokenInfo = getHederaToken(
+                TokenTypeEnum.FUNGIBLE_COMMON, withKeys, inheritKey, defaultFreezeStatus, treasuryEntity);
+        tokenInfo.freezeDefault = defaultFreezeStatus;
+        testWeb3jService.setValue(CREATE_TOKEN_VALUE);
```
steven-sheehy, line 327: Believe this value is only used for creates. If so, instead of having a setter for it we should overload `TestWeb3jService.deploy()` to take in an extra value to supply for creates.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/utils/KeyType.java
```diff
@@ -0,0 +1,39 @@
...
+ * limitations under the License.
+ */
+
+package com.hedera.mirror.web3.utils;
+
+import java.math.BigInteger;
+
+public enum KeyType {
+    ADMIN_KEY(BigInteger.valueOf(1)),
+    KYC_KEY(BigInteger.valueOf(2)),
```
steven-sheehy, line 23: As noted in Ivan's PR, can just pass integer and convert to BigInteger in constructor.

steven-sheehy (changes requested):

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallNestedCallsTest.java
bilyana-gospodinova: I agree. I opened issue https://github.com/hashgraph/hedera-mirror-node/issues/9012 for this enhancement.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallNestedCallsTest.java
bilyana-gospodinova: This seems a little misleading to me - I would expected that the passed value will be used during deploy, not in the subsequent requests. It seems more straightforward to me to set the value for the particular request that needs it.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
bilyana-gospodinova: Moved to `BeforeEach`.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/utils/KeyType.java
bilyana-gospodinova: I removed this class as it already exists in `AbstractContractCallServiceTest` on main.

kselveliev (approved):

steven-sheehy (approved):
LGTM

PR #8957, opened 2024-08-06
Calculate acceptance test operator balance from USD

### LINKED ISSUE #8953: Calculate acceptance operator balance from fiat
### Problem

Due to market volatility, the fixed amount given to the acceptance operator might not be enough to complete the acceptance tests successfully.

### Solution

* Calculate how much acceptance tests cost in fiat.
* Change current balance property to fiat
* Use exchange rate to convert fiat to hbars
* Perhaps fallback if exchange rate API is unavailable

### Alternatives

_No response_

steven-sheehy:
**Description**:

* Add a feature flag for rejectToken so mirror node can deploy before services 0.52
* Change `operatorBalance` property to store amount in USD
* Change operator initial balance to convert USD into tinybar using `transactionReceipt.ExchangeRate` or falling back to `/api/v1/network/exchangerate`
* Fix assertion not printing out correct contract call response
* Fix regression caused by null immutable key
* Fix startup probe topic not being deleted
* Fix startup probe taking forever to recover when all nodes are down

**Related issue(s)**:

Fixes #8953

**Notes for reviewer**:

**Checklist**


xin-hedera (approved):
LGTM

jnels124 (approved):
LGTM

PR #8981, opened 2024-08-08
Update non Production Environments for 0.111.0-rc3

jnels124:
* update staging to 0.111.0-rc3
* update performance to 0.111.0-rc3
* configure acceptance tests to not run reject token scenarios


clusters/mainnet-staging-na/mainnet/helmrelease.yaml
```diff
@@ -40,7 +40,6 @@ spec:
     importer:
       env:
         HEDERA_MIRROR_IMPORTER_NETWORK: "MAINNET"
-        HEDERA_MIRROR_IMPORTER_PARSER_RECORD_ENTITY_PERSIST_TRANSACTIONHASH: "true"
```
steven-sheehy, line 43: We should set backfill start timestamp, right?

steven-sheehy (changes requested):

clusters/mainnet-staging-na/mainnet/helmrelease.yaml
jnels124: yes we should.

jascks (approved):
LGTM

steven-sheehy (approved):

PR #8852, opened 2024-07-22
Fix case sensitive query parameters

jascks:
**Description**:

Some rest api endpoints return 500 with `order=ASC` /  `order=DESC`, examples:

- `https://mainnet-public.mirrornode.hedera.com/api/v1/contracts/results/logs?order=ASC`
- `https://mainnet-public.mirrornode.hedera.com/api/v1/accounts/0.0.1274288/nfts?limit=100&order=ASC`

The transactions REST API misinterprets query filter `result=SUCCESS` to look for transactions with response code not equal to 22. The issue is only all lower case `success` is interpreted as response code equal to 22. On top of the inverted interpretation, this may also cause query timeout, e.g., when combined with `account.id` for which there's very few failed transactions.

This PR, in the `requestHander.js` middleware Express query parser, ensures the query parameter values to selected parameters (currently just `order` and `result`) are forced to lowercase before the request is routed to the request handler code.

**Related issue(s)**:

Fixes #8790
Fixes #8801

**Notes for reviewer**:

**Checklist**

- [x] Documented (Code comments, README, etc.)
- [x] Tested (unit, integration, etc.)

PR #8988, opened 2024-08-09
Fix case sensitive array query parameters

### LINKED ISSUE #8790: Some rest api endpoints return 500 with order=ASC / order=DESC
### Description

Some rest api endpoints return 500 with `order=ASC` /  `order=DESC`, examples:

- `https://mainnet-public.mirrornode.hedera.com/api/v1/contracts/results/logs?order=ASC`
- `https://mainnet-public.mirrornode.hedera.com/api/v1/accounts/0.0.1274288/nfts?limit=100&order=ASC`

Stack trace:

```
TypeError: Cannot read properties of undefined (reading 'test')
    at updateReqQuery (file:///home/node/app/utils.js:815:32)
    at getNextParamQueries (file:///home/node/app/utils.js:848:5)
    at Module.getPaginationLink (file:///home/node/app/utils.js:782:12)
    at AccountController.getPaginationLink (file:///home/node/app/controllers/baseController.js:231:18)
    at getNftsByAccountId (file:///home/node/app/controllers/accountController.js:113:20)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async wrappedMiddleware (/home/node/app/node_modules/@awaitjs/express/index.js:64:9)
```

problem is the order is not normalized to lower case and `operatorPatterns` in `utils.js` can only map lower case `order` to its regex pattern. 

### Steps to reproduce

check the description

### Additional context

_No response_

### Hedera network

other

### Version

v0.108.0

### Operating system

None

jascks:
**Description**:
My first attempt (#8852) at fixing this issue did not accommodate the scenario where a query parameter is provided more than once. When that happens, the Express query parser aggregates the values for the key into an array. Thus, when the `order=ASC` parameter is specified a single time,  the value is simply a string. However `order=ASC&order=ASC` results in the single `order` key, but this time with a value of `[ ASC, ASC]`.

The solution is to accommodate the possibility of the array, and in that case to canonicalize each element individually.

**Related issue(s)**:

Fixes #8790

**Notes for reviewer**:
Beyond the fact this scenario was not tested for previously, my assumption was that the process of merging multiple values within `requestQueryParser()` was handling this possibility. And, it was in the sense that the original code could handle whether the request parameter was a single value or an array, and that it was also merging to support case-insensitive keys.

**Checklist**

### DIFF
--- a/hedera-mirror-rest/__tests__/requestHandler.test.js
+++ b/hedera-mirror-rest/__tests__/requestHandler.test.js
@@ -51,4 +51,15 @@ describe('qs tests', () => {
       'token.id': ['2', '8', 'gt:4', 'gte:6', '10'],
     });
   });
+
+  test('requestQueryParser for single lowercased query param values', () => {
+    const val = requestQueryParser('order=ASC&result=SUCCESS');
+    expect(val.order).toStrictEqual('asc');
+    expect(val.result).toStrictEqual('success');
+  });
+
+  test('requestQueryParser for multiple lowercased query param keys and values', () => {
+    const val = requestQueryParser('order=ASC&ORder=ASC');
+    expect(val.order).toStrictEqual(['asc', 'asc']);
+  });
 });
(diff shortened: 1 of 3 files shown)


xin-hedera (approved):
LGTM

steven-sheehy (approved):

PR #9009, opened 2024-08-12
Handling numeric values for EntityId query parameters

### LINKED ISSUE #8357: RestJava nft allowance controller doesn't accept numeric entity id in URL params
### Description

The nft allowances endpoint returns 400 if the entity id values in the query param is numeric instead of `shard.realm.num`, e.g.,

`/api/v1/accounts/0.0.1001/allowances/nfts?account.id=gt:5000`

### Steps to reproduce

check the description

### Additional context

_No response_

### Hedera network

other

### Version

V0.106.0-SNAPSHOT

### Operating system

None

mgoelswirlds:
**Description**:

This PR fixes rest-java nft allowance controller to accept numeric entity id in URL params.

This PR modifies ...
* EntityId to handle numeric entity not in the `shard.realm.num` format

**Related issue(s)**:

Fixes #8357

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/entity/EntityId.java
+++ b/hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/entity/EntityId.java
@@ -131,10 +131,6 @@ public final class EntityId implements Serializable, Comparable<EntityId> {
                 .filter(n -> n >= 0)
                 .toList();
 
-        if (parts.size() == 1) {
-            return of(0, 0, parts.get(0));
-        }
-
         if (parts.size() != 3) {
             throw new IllegalArgumentException("Invalid entity ID: " + entityId);
         }
--- a/hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/entity/EntityId.java
+++ b/hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/entity/EntityId.java
@@ -131,6 +131,10 @@ public final class EntityId implements Serializable, Comparable<EntityId> {
                 .filter(n -> n >= 0)
                 .toList();
 
+        if (parts.size() == 1) {
+            return of(0, 0, parts.get(0));
+        }
+
         if (parts.size() != 3) {
             throw new IllegalArgumentException("Invalid entity ID: " + entityId);
         }
--- a/hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/EntityIdRangeParameter.java
+++ b/hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/EntityIdRangeParameter.java
@@ -45,8 +45,6 @@ public record EntityIdRangeParameter(RangeOperator operator, EntityId value) imp
     private static EntityId getEntityId(String entityId) {
 
         List<Long> parts = Splitter.on('.')
-                .omitEmptyStrings()
-                .trimResults()
                 .splitToStream(Objects.requireNonNullElse(entityId, ""))
                 .map(Long::valueOf)
                 .filter(n -> n >= 0)
--- a/hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/EntityIdRangeParameter.java
+++ b/hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/EntityIdRangeParameter.java
@@ -16,7 +16,12 @@
 
 package com.hedera.mirror.restjava.common;
 
+import static com.hedera.mirror.restjava.common.EntityIdParameter.DEFAULT_SHARD;
+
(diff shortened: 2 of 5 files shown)


hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/controller/AllowancesControllerTest.java
```diff
@@ -221,6 +221,39 @@ void allRangeConditions() {
             assertThat(result).isEqualTo(getExpectedResponse(List.of(allowance2), next));
         }
 
+        @Test
+        void allRangeConditionsNumericQueryParameters() {
```
mgoelswirlds, line 225: For the reviewers, the upper bounds operators have the `shard.realm.num` format inspite of the original url have only the numeric code. This is to keep consistency with /token/allowances API

jascks (dismissed):
LGTM

hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/entity/EntityId.java
```diff
@@ -131,6 +131,10 @@ public static EntityId of(String entityId) {
                 .filter(n -> n >= 0)
                 .toList();
 
+        if (parts.size() == 1) {
+            return of(0, 0, parts.get(0));
```
steven-sheehy, line 135: This happens to work now, but will have issues later whenever we support non-zero shards. Instead, we should handle this at the rest-java layer and feed in the shard property as the default if it's missing.

hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/entity/EntityId.java
```diff
@@ -131,6 +131,10 @@ public static EntityId of(String entityId) {
                 .filter(n -> n >= 0)
                 .toList();
 
+        if (parts.size() == 1) {
```
steven-sheehy, line 134: We also support `realm.num` syntax on REST API

steven-sheehy (changes requested):

hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/entity/EntityId.java
mgoelswirlds: The problem is this error occurs in the controller ParameterResolution for `EntityIdRangeParameter` itself. So I would need to handle it here. Also, I cannot access the shard property here.
But I agree when we support non-zero shards this will be a problem.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/common/EntityIdRangeParameterTest.java
```diff
@@ -49,7 +49,6 @@ void testConversion() {
                 "-1.-1.-1",
                 "-1",
                 "0.0.-1",
-                "100000.000000001",
```
steven-sheehy, line 52: We should have positive tests for `realm.num` and `num`. Add negative tests with whitespace and negative numbers and 1 and 2 parts.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/EntityIdRangeParameter.java
```diff
@@ -30,13 +35,35 @@ public static EntityIdRangeParameter valueOf(String entityIdRangeParam) {
...
     }
 
+    private static EntityId getEntityId(String entityId) {
+
+        List<Long> parts = Splitter.on('.')
+                .omitEmptyStrings()
+                .trimResults()
+                .splitToStream(Objects.requireNonNullElse(entityId, ""))
+                .map(Long::valueOf)
+                .filter(n -> n >= 0)
```
steven-sheehy, line 52: This parsing is too permissive for user input. It would allow input like `" 0 .  0.1 "`

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/controller/AllowancesControllerTest.java
steven-sheehy, line 225: Instead of a new test, we should make an existing test parameterized with a test enum:

```java
  private enum EntityIdType {
    NUM(e -> "" + e.getNum()),
    REALM_NUM(e -> "%d.%d".format(e.getRealm(), e.getNum()),
    SHARD_REALM_NUM(e -> e.toString());
...
 }
```

steven-sheehy (changes requested):

steven-sheehy (approved):
LGTM

jascks (approved):
LGTM

PR #9030, opened 2024-08-15
Rosetta return generic unknown operation type for future transaction types

### LINKED ISSUE #8320: Rosetta - return generic `UNKNOWN` operation type for future HAPI transaction types
### Problem

It's a maintenance burden to add support of resolving a new transaction type (an integer) to its transaction name (in text, and in rosetta world, it's used as operation type). In the mean time, if the customer needs support of a specific transaction for both rosetta data API and construction API, a request shall be made and worked ahead of going to production.

Thus, in mirrornode rosetta module, for any transaction type out of the already supported ones, `UNKNOWN` should be provided as the operation type.

### Solution

- make `types.TransactionTypes` private, i.e., change it to `types.transactionTypes`
- add a public function `func GetTransactionType(type int32) string`
- replace references of `types.TransactionTypes` to `types.GetTransactionType`
- Check if we can enumerate transaction results or types from the sdk instead of explicitly defining them

### Alternatives

_No response_

xin-hedera:
**Description**:

This PR improves Rosetta to return `UNKNOWN` as transaction type for new HAPI transactions

- Map any new transaction type to `UNKNOWN`
- Update TransactionRepository to show `UNKNOWN` operation type
- Bump `go-ethereum` to v1.14.8 and remove `btcd` version override
- Remove `tiff` version override

**Related issue(s)**:

Fixes #8320 

**Notes for reviewer**:

**Checklist**


steven-sheehy (approved):
LGTM

jnels124 (approved):
LGTM

edwin-greene (dismissed):
Looks great!

mgoelswirlds (dismissed):
LGTM

hedera-mirror-rosetta/app/domain/types/constants.go
```diff
@@ -16,11 +16,18 @@
...
+	"github.com/hashgraph/hedera-protobufs-go/services"
+	"strings"
+)
 
 const (
-	generalErrorStatus = "GENERAL_ERROR"
-	MetadataKeyMemo    = "memo"
+	generalErrorStatus     = "GENERAL_ERROR"
+	MaxTransactionResult   = int32(337)
+	MaxTransactionType     = int32(57)
```
xin-hedera, line 28: walks back the max transaction type from 60 to 57 since the protobuf go package paired with the go sdk doesn't have the three token airdrop transaction types

hedera-mirror-rosetta/app/domain/types/constants.go
```diff
@@ -443,3 +96,45 @@ func GetTransactionResult(code int32) string {
...
+	for i := 0; i < dataFields.Len(); i++ {
+		dataField := dataFields.Get(i)
+		protoId := int32(dataField.Number())
+		if protoId > MaxTransactionType {
+			continue
+		}
+
+		name := strings.ToUpper(string(dataField.Name()))
+		transactionTypes[protoId] = strings.ReplaceAll(name, "_", "")
+	}
```
xin-hedera, line 139: these are apparently not blazing fast, so store them in the map for reuse

hedera-mirror-rosetta/app/domain/types/constants.go
```diff
@@ -443,3 +96,45 @@ func GetTransactionResult(code int32) string {
...
+	return transactionTypes
+}
+
+func init() {
+	// Initialize transaction results
+	transactionResults = make(map[int32]string)
+	transactionResults[-1] = generalErrorStatus
+	for code, name := range services.ResponseCodeEnum_name {
+		if code > MaxTransactionResult {
+			continue
```
xin-hedera, line 119: we need to `continue` instead of `break` since both the response code and the transaction type data structure are not order by code / protoId when being iterated over

hedera-mirror-rosetta/app/domain/types/constants_test.go
```diff
@@ -17,57 +17,386 @@
...
-	"github.com/hashgraph/hedera-protobufs-go/services"
 	"github.com/stretchr/testify/assert"
 )
 
-const maxCode = int32(337)
+var expectedTransactionResults = map[int32]string{
+	-1:  generalErrorStatus,
+	0:   "OK",
+	1:   "INVALID_TRANSACTION",
+	2:   "PAYER_ACCOUNT_NOT_FOUND",
```
steven-sheehy, line 29: I don't think we should remove the hardcoded values in constants.go just to add them to the tests and have to maintain them there.

Instead, I think for both results and types we should just verify ordinal range of ordinals (e.g. `[-1,337]`) and also that string matches `/[A-Z_]+/`, but not specific values.

xin-hedera: We don’t have to maintain them since for anything beyond the max it’s going to be default. IE the two maps in the test case should stay as is forever

steven-sheehy: In theory but people will hunt up references to response codes and types and add them to the test maps in practice.

xin-hedera: That should be fine, as those two maps will be static moving forward. And if in any case the values from the proto package change, it must be a bad proto package release, then we'd scrape the dependency upgrade, instead of propagating the bugs.

xin-hedera: And it's more changes if someone wants to update the map in the tests cases to make tests pass v.s. without the map and just verify a generic regex match, right?

I mean, if that ever happens, isn't it more noticeable for reviewers than there are more changes in code than just increasing `MaxResponsdeCode` and `MaxTransactionType` in constants.go?

steven-sheehy (approved):

edwin-greene (approved):

PR #9038, opened 2024-08-15
Implement additional single-entity builders and run more spec tests

### LINKED ISSUE #8829: Implement additional single-entity builders and run more spec tests
### Problem

Advance from the first phase implementation, which largely laid in the groundwork for deserializing existing spec tests, being able to customize and create a small subset of database entities defined in test setup and executes a fixed list of curated spec test files against the JS REST API container.

Support the setup of additional database entities so that at least some of their relevant REST API spec tests can be successfully executed. These are still only entity builders that create a single entity based on the spec.

Note that special features such as matrix the sinon time functionality will come later.

### Solution

Features
  * Implement additional single-entity builders to create, customize and persist additional test DB entities.
  * Replace `RestSpecTest.setupDatabase()` with something more general to process all `SpecSetup` definitions.

### Alternatives

_No response_

jascks:
**Description**:
Advance from the first phase implementation, which largely laid in the groundwork for deserializing existing spec tests, being able to customize and create a small subset of database entities defined in test setup and executes a fixed list of curated spec test files against the JS REST API container.

Support the setup of the additional database entities so that many more of the JS REST API spec tests can be successfully executed. This PR is a progression toward getting all entity types support.

**Related issue(s)**:

Fixes #8829

**Notes for reviewer**:
Now that I got spotless working properly, a number of changes in these files are due to spotless formatting.

This PR adds additional single-entity spec entity builders and adds (hardcodes) additional spec tests to demonstrate that they do indeed function. This is just a step toward getting all entity types referenced in the existing specs to work as expected. This PR does not close #8829. A future PR (or two) will do that.

**Checklist**


IvanKavaldzhiev (dismissed):
LGTM

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/converter/ByteArrayFromStringConverter.java
```diff
@@ -0,0 +1,32 @@
...
+import org.springframework.boot.context.properties.ConfigurationPropertiesBinding;
+import org.springframework.core.convert.converter.Converter;
+import org.springframework.util.StringUtils;
+
+@Named
+@ConfigurationPropertiesBinding
+public class ByteArrayFromStringConverter implements Converter<String, byte[]> {
+    @Override
+    public byte[] convert(String source) {
+        return StringUtils.hasText(source) ? source.getBytes(StandardCharsets.UTF_8) : null;
```
edwin-greene, line 30: I was looking at this and wondering if we want the String value of `" "` to return null vs the byte array for white space `[32]`?

jascks: `hasText()` returns false in that case and null is returned.

jascks: I can instead utilize `hasLength()` and return null only when `source` is null or empty, and then whitespace characters can be reflected in the byte array.

steven-sheehy (approved):
LGTM

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/converter/ByteArrayFromStringConverter.java
jascks: Okay, done. Only empty string results in null.

steven-sheehy (approved):

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/converter/ByteArrayFromStringConverterTest.java
```diff
@@ -0,0 +1,53 @@
...
+        var converter = new ByteArrayFromStringConverter();
+        assertThat(converter.convert(source)).isNull();
+    }
+
+    private static Stream<Arguments> provideTestCases() {
+        return Stream.of(
+                Arguments.of("0", new byte[] {0x30}),
+                Arguments.of("01", new byte[] {0x30, 0x31}),
+                Arguments.of("hashgraph", "hashgraph".getBytes(StandardCharsets.UTF_8)),
+                Arguments.of("", null),
```
jascks, line 49: Okay, this scenario is already covered above with `@NullAndEmptySource`. I'll leave it in this PR and remove it in my follow on work (#9047).

edwin-greene (approved):
Looks good

PR #9042, opened 2024-08-16
8456 Refactor ContractCallServiceERCTokenTest readonly redirect functions

nickeynikolovv:
**Description**:
With this PR the ContractCallServiceERCTokenTest integration test file is partially refactored using the web3j plugin and the big enum is now replaced with multiple junit tests so that it is easier to debug a particular scenario. This way also the number of persisted entities per test is reduced significantly and now we persist only the ones that are needed in the particular test.

This PR refactors the read only redirect functions tests 

**Related issue(s)**:
Related and partially fixes: #8456 
Fixes #

**Checklist**


steven-sheehy (approved):
LGTM

kselveliev (approved):

PR #9076, opened 2024-08-20
Extract the internal calls tests from `ContractCallServiceTest`

bilyana-gospodinova:
**Description**:
This PR extracts some of the tests from the `ContractCallServiceTest` that are related to the `InternalCaller` contract and that test different internal calls scenarios. Some cleanup in the `ContractCallTestSetup` was made as well.

**Related issue(s)**: https://github.com/hashgraph/hedera-mirror-node/issues/8452


IvanKavaldzhiev (approved):
LGTM

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/InternalCallsTest.java
```diff
@@ -0,0 +1,94 @@
...
+import org.junit.jupiter.api.Test;
+
+public class InternalCallsTest extends AbstractContractCallServiceTest {
+
+    private static final String NON_EXISTING_ADDRESS = toAddress(123456789L).toHexString();
+
+    @Test
+    void callToNonExistingContract() throws Exception {
+        final var contract = testWeb3jService.deploy(InternalCaller::deploy);
+        meterRegistry.clear();
```
steven-sheehy, line 37: We already clear the meter registry in the parent `CommonIntegrationTest`. This extra clear is because deploy() also invokes a top level create. 

Isn't it inefficient and unnecessary to invoke create before every call when create doesn't modify state? Probably we should in the future intercept these unneeded creates and fake their execution?

steven-sheehy (approved):
LGTM

PR #9095, opened 2024-08-21
Create Mainnet Citus in NA Cluster

jnels124:
create mainnet citus in na cluster


clusters/mainnet-na/mainnet-citus/sealed-secret.yaml
jnels124, line 1: Some double check this by creating the namespace in na and applying just this yaml please

steven-sheehy: Looks good

steven-sheehy (changes requested):
We also need to add mainnet-citus to `alert-slack.yaml`. Which it's apparently also missing from mainnet-eu

steven-sheehy (approved):

jascks (approved):
LGTM

PR #9106, opened 2024-08-22
Enable Stackgres Mainnet NA 

jnels124:
* enable stackgres operator


jascks (approved):
LGTM

clusters/mainnet-na/common/helmrelease.yaml
```diff
@@ -61,6 +61,8 @@ spec:
                   regex: (api_.+_bucket|executor.*|lettuce_.*_bucket|traefik_.+_bucket)
                   sourceLabels: [ __name__ ]
           replicaExternalLabelName: __replica__
+    stackgres:
```
xin-hedera, line 64: should we also increate prometheus memory limits?

jnels124: yes probably a good time to do that as well. Done

jascks (approved):
LGTM

steven-sheehy (approved):

PR #9137, opened 2024-08-23
Refactor `ContractCallServiceERCTokenTest` negative scenarios

nickeynikolovv:
**Description**:

This PR refactors the negative tests so they can be more readable. 

**Related issue(s)**:

Related and partially addresses: #9108

**Checklist**


bilyana-gospodinova:
LGTM, but:
1. It would be nice to add comments why a test is expected to fail or rename the test so that you would know immediately why it should fail. I know that in the original tests it is not specified, but I think it can be useful.
2. When either of this, or the another PR for the ERC modification functions is merged, we need to delete the Redirect contract's resources (abi, json, sol) and its constants and methods in `ContractCallTestSetup`.

steven-sheehy (dismissed):

kselveliev (dismissed):

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServiceERCTokenReadOnlyFunctionsTest.java
```diff
@@ -938,6 +940,61 @@ void ethCallTokenURIRedirect() throws Exception {
         verifyEthCallAndEstimateGas(functionCall, contract);
     }
 
+    // NEGATIVE TESTS BELOW
+    @Test
+    void decimalsNegative() {
+        final var tokenEntity = nftPersist();
```
IvanKavaldzhiev, line 946: Add Given, When, Then clauses for all tests.

nickeynikolovv: Done

PR #9144, opened 2024-08-24
Disable postgres SSL to fix random query timeout

### LINKED ISSUE #8748: Expose pg-util container config and enable it by default
### Problem

stackgres provides an embedded pg-util container with `psql` and many other tools to make checking database status easy. It's especially useful when we want to check pgbouncer stats, since otherwise it's hard to do so

- there is psql in the `patroni` container, however because many other tools like a pager (more) are missing, even psql can connect to pgbouncer via unix socket and port 6432 without authentication, many output will just fail to show up due to the pager issue
- port forwarding 6432 will not work since by default pgbouncer only allows no password auth with user `pgbouncer` via unix socket

### Solution

expose `disablePostgreUtil` config  and default it to `true` in `hedera-mirror` charts values.yaml

### Alternatives

_No response_

### LINKED ISSUE #9143: Some readonly queries time out in citus with pgbouncer pooling
### Description

Some readonly queries time out in citus, when coordinator to worker db connections have pgbouner as the connection pool in the middle.

At least two readonly queries have shown the issue

- fileDataRepository.findAddressBooksBetween
- topic messages by id query when the where clause has two timestamp range conditions to include two `topic_message` time partitions

Most of the time these queries will timeout, and very rarely they can succeed. When successful, the queries take less than 5ms to complete. 

### Steps to reproduce

as the description

### Additional context

_No response_

### Hedera network

other

### Version

v0.113.0-SNAPSHOT

### Operating system

None

xin-hedera:
**Description**:

This PR fixes the random readonly query timeout issue in citus

- Disable SSL
- Enable postgres-util container
- Reduce postgresql work_mem to 24MB to lower memory pressue
- Add `TEST_REPORTS_DIR` env variable to all testkube tests

**Related issue(s)**:

Fixes #8748 
Fixes #9143 

**Notes for reviewer**:

After much effort to diagnose and experiment, found disabling SSL solves the issue. Though I'm not clear about why SSL connection (via unix socket) can cause such issue, the changes here are much needed so we can move on to other tasks.

If needed, we can open a ticket upstream, however, the configuration itself (coordinator -> worker connection pgbouncer pooling) isn't supported by stackgres.

The changes which enable postgres-util container is a must to gain pgbouncer info. Without it, it's impossible to view a lot of pgbouner stats. The problem is pgbouner can only be connected via unix socket with psql command, `psql -p 6432 -U pgbouncer pgbouncer`. We can do so in the `patroni` container.  However a lot of commands will just fail to show results, e.g., `show clients`, complaining about `more` is not available as it's not installed in the patroni container.

Verified for the upgrade path from <= 0.112.0-rc3 helm charts with SSL defaults to enabled, to charts in this PR, no manual steps are needed.
 
**Checklist**

### DIFF
--- a/charts/hedera-mirror-common/templates/test-rest-java.yaml
+++ b/charts/hedera-mirror-common/templates/test-rest-java.yaml
@@ -31,6 +31,10 @@ spec:
         name: BASE_URL
         type: basic
         value: http://{{ $target.release }}-restjava.{{ $target.namespace }}.svc.cluster.local
+      TEST_REPORTS_DIR:
+        name: TEST_REPORTS_DIR
+        type: basic
+        value: /share
   type: k6-custom/script
 {{- end}}
 {{- end }}
--- a/charts/hedera-mirror/values.yaml
+++ b/charts/hedera-mirror/values.yaml
@@ -280,6 +280,7 @@ stackgres:
       citus.executor_slow_start_interval: "20ms"
       citus.max_cached_conns_per_worker: "6"
       citus.max_shared_pool_size: "850"
+      citus.node_conninfo: "sslmode=disable"  # Disable SSL to work around https://github.com/hashgraph/hedera-mirror-node/issues/9143
       full_page_writes: "true"  # Required for replication to work in recovery scenarios
       log_checkpoints: "true"
       log_timezone: "Etc/UTC"
@@ -293,6 +294,7 @@ stackgres:
       wal_recycle: "off"  # Not needed with ZFS
       work_mem: "32MB"
     enableMetricsExporter: false
+    enablePostgresUtil: true
     instances: 1
     replication:
       mode: sync-all
@@ -331,6 +333,7 @@ stackgres:
     config:
       autovacuum_max_workers: "2"
       checkpoint_timeout: "1800"
+      citus.node_conninfo: "sslmode=disable" # Disable SSL to work around https://github.com/hashgraph/hedera-mirror-node/issues/9143
       full_page_writes: "false"  # Not required for ZFS and Enabling this causes hash indexes to have spikes in write performance.
       log_checkpoints: "true"
       log_timezone: "Etc/UTC"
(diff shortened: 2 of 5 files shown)


jascks (dismissed):
LGTM

xin-hedera:
> In previous pr to disable postgres node exporter container, we left configuration that allocates I beleive 100m cpu. Can we remove that and add in specific resource constraints for postgres-util?

My recollection of when I enabled it manually, the container had no resources section at all. Perhaps we should double check and give a reasonable cpu limit. I’d prefer getting it in a separate PR still.

edwin-greene (dismissed):
Looks good

jnels124 (dismissed):
LGTM

steven-sheehy (dismissed):

mgoelswirlds (approved):

xin-hedera:
> > In previous pr to disable postgres node exporter container, we left configuration that allocates I beleive 100m cpu. Can we remove that and add in specific resource constraints for postgres-util?
> 
> My recollection of when I enabled it manually, the container had no resources section at all. Perhaps we should double check and give a reasonable cpu limit. I’d prefer getting it in a separate PR still.

Ticket here: https://github.com/hashgraph/hedera-mirror-node/issues/9166

charts/hedera-mirror/values.yaml
```diff
@@ -277,9 +277,10 @@ stackgres:
     config:
       autovacuum_max_workers: "2"
       checkpoint_timeout: "1800"
-      citus.executor_slow_start_interval: "20ms"
+      citus.executor_slow_start_interval: "10ms"
```
xin-hedera, line 280: Since the connections are through pgbouncer pool, creating connections too fast are no longer a concern, especially with transaction pool_mode and when there are already server side connections in the pool to reuse.

charts/hedera-mirror-common/templates/test-rest-java.yaml
```diff
@@ -31,6 +31,10 @@ spec:
         name: BASE_URL
         type: basic
         value: http://{{ $target.release }}-restjava.{{ $target.namespace }}.svc.cluster.local
+      TEST_REPORTS_DIR:
```
xin-hedera, line 34: This is for convenience when we run a single testkube test, that we'll always get reports scraped.

jascks (approved):
LGTM

steven-sheehy (approved):
LGTM

PR #9163, opened 2024-08-26
Refactor `ContractCallServiceERCTokenTest` delegate transfer

nickeynikolovv:
**Description**:

This PR refactors delegate transfer test, so it can be more readable. 

**Related issue(s)**:

Related and partially addresses: #9162 

**Checklist**


IvanKavaldzhiev (approved):
LGTM

steven-sheehy (approved):

PR #9173, opened 2024-08-27
Add an application cache for REST API

### LINKED ISSUE #9111: Add an application cache for REST API
### Problem

There are many times that the same request from different sessions all hit our REST server even though cloud CDN is enabled and the responses have proper cache control header set.

There are also periodic high rate of response code 304 from our REST server because some clients will issue a HEAD request with cache id to probe if the cache is still valid. Due to lack of head handler in our js REST module, it ends up to go though the get handler and query data, then expressjs learns the cache is still valid, zeros out the response body and sends a 304 response.

We should add a application layer cache to handle such scenarios so the rest server doesn't unnecessarily go to the database for the data.

### Solution

- Add logic in response handler middleware to cache responses in redis
  - we can use the hash of the request as the key
  - we should set the cache ttl the same as how it's configured by path in the response cache-control header
- Add a middleware before any of the handlers to check if the request url has a cached response, and if so, respond with the cached response and terminate the request handling

### Alternatives

_No response_

jascks:
**Description**:

Very preliminary draft to get feedback on general approach to the solution. There is still more implementation and testing to be done.

There are many times that the same request from different sessions all hit our REST server even though cloud CDN is enabled and the responses have proper cache control header set.

There are also periodic high rate of response code 304 from our REST server because some clients will issue a HEAD request with cache id to probe if the cache is still valid. Due to lack of head handler in our js REST module, it ends up to go though the get handler and query data, then expressjs learns the cache is still valid, zeros out the response body and sends a 304 response.

We should add a application layer cache to handle such scenarios so the rest server doesn't unnecessarily go to the database for the data.

**Related issue(s)**:

Fixes #9111

**Notes for reviewer**:
* Still working out Snyk code security check failure.
* Working on getting performance testing results.
* Determine Redis `maxMemory`.

**Checklist**


hedera-mirror-rest/server.js
```diff
@@ -125,6 +127,9 @@ if (config.metrics.enabled) {
   app.use(metricsHandler());
 }
 
+// Check for cached response
+app.use(responseCacheCheckHandler);
```
jascks, line 131: Should this be `use()` or `useAsync()`?

hedera-mirror-rest/config/application.yml
```diff
@@ -6,6 +6,8 @@ hedera:
         entityId:
           maxAge: 1800
           maxSize: 100000
+        response:
+          enabled: false
```
jascks, line 10: I figured we'd start with the response cache disabled by default. Then we can selectively turn it on where it makes sense.

Issue #9074, opened 2024-08-20
Update the web3j-cli download task to use the official release


bilyana-gospodinova:
### Problem

There is a custom gradle task (`downloadWeb3j`) that downloads the `web3j-cli` from a custom repo. There is an official release of the cli now - https://github.com/hyperledger/web3j-cli/releases/tag/v1.6.1.

### Solution

We need to update the task to use the official release. If possible, we shouldn't use a task at all. Research if the `web3j-cli` can be used as a dependency without any scripts.

PR #9182, opened 2024-08-29
Update the web3j-cli download task to use the official release

bilyana-gospodinova:
**Description**:
There is an official release of the `web3j-cli`, so this PR removes the custom download logic from the custom repo and updates the gradle task to use the official release. 

It would be nice if we can add the cli as a gradle dependency and not use gradle tasks to download it at all.

**Related issue(s)**: 

Fixes: #9074

### DIFF
--- a/hedera-mirror-web3/build.gradle.kts
+++ b/hedera-mirror-web3/build.gradle.kts
@@ -123,6 +123,7 @@ val compileHistoricalSolidityContracts =
         description = "Compiles the historical solidity contracts to java files using web3j-cli"
         group = "historical"
         mustRunAfter(tasks.named("generateTestContractWrappers"))
+        dependsOn(downloadWeb3j)
         dependsOn(extractOpenZeppelin)
         dependsOn(tasks.named("compileTestSolidity"))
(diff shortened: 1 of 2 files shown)


hedera-mirror-web3/build.gradle.kts
```diff
@@ -70,35 +70,12 @@ tasks.compileJava { options.compilerArgs.add("--enable-preview") }
 
 tasks.test { jvmArgs = listOf("--enable-preview") }
 
-val web3jDir = rootDir.resolve(".gradle").resolve("web3j")
-val web3jVersion = "1.6.0"
```
steven-sheehy, line 74: The script doesn't seem to provide a means to download a specific version. Hopefully they don't introduce breaking changes.

steven-sheehy (approved):
LGTM

hedera-mirror-web3/build.gradle.kts
bilyana-gospodinova: Correct - the script always downloads the latest version. There are rarely any changes in the repo at all.

PR #9189, opened 2024-08-29
Return 501 NOT_SUPPORTED error response for unsupported precompiles

### LINKED ISSUE #9136: Add temporary logic for `501 NOT_SUPPORTED` status in web3
### Problem

Currently the `hedera-mirror-web3` business logic is based on a copied code of the services mono code base. Services, however, has switched to modularized code base, which will add some new Hedera precompiles, which we can't directly copy in web3. This opens a need to add a logic for throwing exception for unsupported precompiles, so that the json-rpc-relay distinguish between contract revert scenario and a call to unsupported precompile.

### Solution

- Throw custom exception in HTSPrecompiledContract for the cases of calling unsupported precompile
- Add interceptor in the exception handler advice component for throwing NOT_SUPPORTED error

### Alternatives

_No response_

IvanKavaldzhiev:
**Description**:
This PR aims to protect the web3 calls to reach unimplemented precompiles due to missing functionality in the copied mono code base. When the `json-rpc-relay` tries to invoke such precompile it will receive the informative `501` error and perform the needed fallback logic. Currently they receive `500` `INTERNAL_ERROR` with `CONTRACT_REVERT_EXECUTED` status, which might be misleading.

The PR adds:
- logic to return NOT_SUPPORTED status for any new dynamic/modification precompile
- logic to return NOT_SUPPORTED status for HRC_IS_ASSOCIATED view precompile. That's the only read only precompile added in services after switching to modularized code and is not supported in `hedera-mirror-web3`. Unfortunately, we don't have a way to add a universal check for each not supported view precompile, since there is not distinction between not found precompile and failed input validation for the view precompiles, so we will have to add a check for each not supported precompile explicitly

**Related issue(s)**:

Fixes #9136 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/GenericControllerAdvice.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/GenericControllerAdvice.java
@@ -19,10 +19,12 @@ package com.hedera.mirror.web3.controller;
 import static org.springframework.http.HttpStatus.BAD_REQUEST;
 import static org.springframework.http.HttpStatus.INTERNAL_SERVER_ERROR;
 import static org.springframework.http.HttpStatus.NOT_FOUND;
+import static org.springframework.http.HttpStatus.NOT_IMPLEMENTED;
 import static org.springframework.http.HttpStatus.SERVICE_UNAVAILABLE;
 import static org.springframework.http.HttpStatus.TOO_MANY_REQUESTS;
 import static org.springframework.web.context.request.RequestAttributes.SCOPE_REQUEST;
 
+import com.hedera.mirror.web3.evm.exception.PrecompileNotSupportedException;
 import com.hedera.mirror.web3.exception.EntityNotFoundException;
 import com.hedera.mirror.web3.exception.InvalidInputException;
 import com.hedera.mirror.web3.exception.MirrorEvmTransactionException;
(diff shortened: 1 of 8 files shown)


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/GenericControllerAdvice.java
```diff
@@ -101,6 +103,17 @@ private ResponseEntity<?> queryTimeoutException(final QueryTimeoutException e, W
...
     }
 
+    /**
+     * Temporary handler, intended for dealing with forthcoming features that are not yet available, such as the absence
+     * of a precompile
+     **/
+    @ExceptionHandler
+    private ResponseEntity<?> precompileNotSupportedException(
+            final PrecompileNotSupportedException e, WebRequest request) {
+        //        return new ResponseEntity<>(new GenericErrorResponse(e.getMessage()), NOT_IMPLEMENTED);
```
kselveliev, line 113: I think we can delete this commented part.

hedera-mirror-web3/src/main/java/com/hedera/services/store/contracts/precompile/HTSPrecompiledContract.java
```diff
@@ -466,6 +495,13 @@ private Pair<Long, Bytes> handleReadsFromDynamicContext(Bytes input, @NonNull fi
     }
 
     private Pair<Long, Bytes> computeUsingRedirectExecutor(final Bytes input, final MessageFrame frame) {
+        final var target = getRedirectTarget(input);
+        if (AbiConstants.ABI_HRC_IS_ASSOCIATED == target.descriptor()) {
+            // We can't support this precompile since the copied code from mono services does not have the necessary
+            // logic
+            throw new PrecompileNotSupportedException(StringUtils.EMPTY);
```
steven-sheehy, line 502: We should put the name of the precompile in the exception message. This message while not delivered in the response will be see in the access logs.

kselveliev (approved):
LGTM

kselveliev (approved):

steven-sheehy (approved):

PR #9190, opened 2024-08-29
Add REST API query normalization

### LINKED ISSUE #9113: Add query normalization to support REST API Caching
### Problem

Caching on the REST API can be enhanced by adding logic that handles absent query parameters indentically to default query parameters. For example: 
`/api/v1/accounts` and  `/api/v1/accounts?limit=25`

### Solution

Add query parameter normalization functionality. It should be easily usable by the caching middleware from #9111

### Alternatives

_No response_

edwin-greene:
**Description**:
Adds a means to normalize REST API queries.

**Related issue(s)**:

Fixes #9113 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-rest/__tests__/middleware/requestNormalizer.test.js
+++ b/hedera-mirror-rest/__tests__/middleware/requestNormalizer.test.js
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2022-2024 Hedera Hashgraph, LLC
+ * Copyright (C) 2024 Hedera Hashgraph, LLC
  *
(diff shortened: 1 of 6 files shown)


jascks (approved):
I really like this. I can import `normalizeRequestQueryParams()` into the new response caching middleware and use it to form most of the cache key, along with `Accept-Encoding` header value.

As far as this PR goes, it can be merged independently as no code is invoked as configured Express middleware or as a query processor.

steven-sheehy (changes requested):

PR #9224, opened 2024-09-02
Add check for `web3j-cli` existence before installation

bilyana-gospodinova:
**Description**:
As investigated by @kselveliev, it seems that the `web3j-cli` download script has a rate limit. This PR adds a check if it is already installed. A similar check exists in the download script but this PR aims to fix the rate limit issue by performing the check earlier - in our gradle task.

### DIFF
--- a/hedera-mirror-web3/build.gradle.kts
+++ b/hedera-mirror-web3/build.gradle.kts
@@ -70,12 +70,16 @@ tasks.compileJava { options.compilerArgs.add("--enable-preview") }
 
 tasks.test { jvmArgs = listOf("--enable-preview") }
 
+val homeDir = System.getenv("HOME")
+val web3jDirectory = file("$homeDir/.web3j")
+
 val downloadWeb3j =
     tasks.register<Exec>("downloadWeb3j") {
         description = "Download and install Web3j CLI"
         group = "historical"
 
         commandLine("bash", "-c", "curl -L get.web3j.io | sh")
+        onlyIf { !web3jDirectory.exists() }
     }
 
 // Tasks to download OpenZeppelin contracts


hedera-mirror-web3/build.gradle.kts
```diff
@@ -70,12 +70,16 @@ tasks.compileJava { options.compilerArgs.add("--enable-preview") }
...
+val homeDir = System.getenv("HOME")
+val web3jDirectory = file("$homeDir/.web3j")
+
 val downloadWeb3j =
     tasks.register<Exec>("downloadWeb3j") {
         description = "Download and install Web3j CLI"
         group = "historical"
 
         commandLine("bash", "-c", "curl -L get.web3j.io | sh")
+        onlyIf { !web3jDirectory.exists() }
```
steven-sheehy, line 82: If the web3j script is ever updated, this won't download the newer version until the .web3j folder is manually deleted. This is something we have to keep in mind.

Do we have a ticket to work on the upstream enhancement to support multiple solidity versions? Is there a ticket created upstream to ask for such an enhancement? We should create both, if not.

steven-sheehy (approved):
LGTM

hedera-mirror-web3/build.gradle.kts
bilyana-gospodinova: As a result from this change we will use the same web3j-cli version until we delete the folder and this way if there are some breaking changes in the latest release we won't be affected.

There is no ticket created upstream to ask for an enhancement in web3j cli/plugin. I will open one.

bilyana-gospodinova: I opened an upstream enhancement issue - https://github.com/hyperledger/web3j-gradle-plugin/issues/90.

PR #9237, opened 2024-09-02
Prepare db resources for index creation in mainnet-citus na

jnels124:
prepare mainnet-citus na for transaction index creation


steven-sheehy (dismissed):

clusters/mainnet-na/mainnet-citus/helmrelease.yaml
```diff
@@ -91,6 +104,12 @@ spec:
             pods:
               persistentVolume:
                 size: 3200Gi
+        resources:
+          cpu: 44000m
+          memory: 176Gi
+          requests:
+            patroni: 44000m
```
xin-hedera, line 111: This should be requests.containers.patrons.cpu

Same problem in testnet-eu

steven-sheehy (approved):

PR #9262, opened 2024-09-03
Refactor ContractCallServicePrecompileHistoricalTest

### LINKED ISSUE #9072: Refactor `ContractCallServicePrecompileHistoricalTest`
### Problem

The tests needs to be refactored to use the web3j generated java test files.

### Solution

The tests needs to be refactored to use the web3j generated java test files.

### Alternatives

_No response_

kselveliev:
**Description**:
This Pr refactors `ContractCallServicePrecompileHistoricalTest` to use the web3j plugin.

This PR modifies:
`ContractCallServicePrecompileHistoricalTest` - refactor all methods to use the web3j plugin
`AbstractContractCallServiceTest` - extracted some common methods in it.
Added the solidity contract `PrecompileTestContractHistorical`.sol to solidity_historical folder

**Related issue(s)**:

Fixes #9072 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServicePrecompileHistoricalTest.java
+++ b/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServicePrecompileHistoricalTest.java
@@ -1019,24 +1019,24 @@ class ContractCallServicePrecompileHistoricalTest extends AbstractContractCallSe
                 .type(tokenType));
 
         switch (keyType) {
-            case AbstractContractCallServiceTest.KeyType.ADMIN_KEY:
+            case ADMIN_KEY:
                 break;
-            case AbstractContractCallServiceTest.KeyType.KYC_KEY:
+            case KYC_KEY:
                 tokenBuilder.customize(t -> t.kycKey(key.toByteArray()));
                 break;
-            case AbstractContractCallServiceTest.KeyType.FREEZE_KEY:
+            case FREEZE_KEY:
                 tokenBuilder.customize(t -> t.freezeKey(key.toByteArray()));
                 break;
-            case AbstractContractCallServiceTest.KeyType.WIPE_KEY:
+            case WIPE_KEY:
                 tokenBuilder.customize(t -> t.wipeKey(key.toByteArray()));
                 break;
-            case AbstractContractCallServiceTest.KeyType.SUPPLY_KEY:
+            case SUPPLY_KEY:
                 tokenBuilder.customize(t -> t.supplyKey(key.toByteArray()));
                 break;
-            case AbstractContractCallServiceTest.KeyType.FEE_SCHEDULE_KEY:
+            case FEE_SCHEDULE_KEY:
                 tokenBuilder.customize(t -> t.feeScheduleKey(key.toByteArray()));
                 break;
-            case AbstractContractCallServiceTest.KeyType.PAUSE_KEY:
+            case PAUSE_KEY:
                 tokenBuilder.customize(t -> t.pauseKey(key.toByteArray()));
                 break;
             default:
--- a/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServicePrecompileHistoricalTest.java
+++ b/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServicePrecompileHistoricalTest.java
@@ -64,6 +64,7 @@ import java.math.BigInteger;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import org.apache.commons.lang3.tuple.Pair;
 import org.apache.tuweni.bytes.Bytes;
 import org.hyperledger.besu.datatypes.Address;
 import org.junit.jupiter.params.ParameterizedTest;
@@ -78,24 +79,13 @@ class ContractCallServicePrecompileHistoricalTest extends AbstractContractCallSe
     void isTokenFrozen(long blockNumber) throws Exception {
         // Given
         final var historicalRange = setUpHistoricalContext(blockNumber);
-        final var account = persistAccountEntityHistorical(historicalRange);
-        final var tokenEntity = persistTokenEntityHistorical(historicalRange);
-        persistFungibleTokenHistorical(tokenEntity, historicalRange);
-        domainBuilder
-                .tokenAccount()
-                .customize(ta -> ta.tokenId(tokenEntity.getId())
-                        .accountId(account.getId())
-                        .kycStatus(TokenKycStatusEnum.GRANTED)
-                        .freezeStatus(TokenFreezeStatusEnum.FROZEN)
-                        .associated(true)
-                        .timestampRange(historicalRange))
-                .persist();
+        final var accountAndToken = persistAccountTokenAndFrozenRelationshipHistorical(historicalRange);
 
         final var contract = testWeb3jService.deploy(PrecompileTestContractHistorical::deploy);
 
         // When
-        final var functionCall =
-                contract.call_isTokenFrozen(getAddressFromEntity(tokenEntity), getAddressFromEntity(account));
+        final var functionCall = contract.call_isTokenFrozen(
+                getAddressFromEntity(accountAndToken.getRight()), getAddressFromEntity(accountAndToken.getLeft()));
 
         // Then
         assertThat(functionCall.send()).isTrue();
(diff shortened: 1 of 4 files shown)


steven-sheehy (dismissed):

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServicePrecompileHistoricalTest.java
```diff
@@ -16,352 +16,1322 @@
...
+
+        domainBuilder
+                .tokenTransfer()
+                .customize(ab -> ab.id(new TokenTransfer.Id(entity.getCreatedTimestamp(), entity.toEntityId(), token))
+                        .amount(100))
+                .persist();
+        return entity;
+    }
+
+    private Range<Long> setUpHistoricalContext(final long blockNumber) {
```
bilyana-gospodinova, line 1290: Isn't that logic common for all historical tests? It can moved to a common place and called in @BeforeEach or wherever else needed.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServicePrecompileHistoricalTest.java
```diff
@@ -16,352 +16,1322 @@
...
-                .map(blockNumber -> Arguments.of(htsFunction, blockNumber)));
+    @ParameterizedTest
+    @CsvSource({"50", "49"})
+    void isTokenFrozen(long blockNumber) throws Exception {
+        // Given
+        final var historicalRange = setUpHistoricalContext(blockNumber);
+        final var account = persistAccountEntityHistorical(historicalRange);
+        final var tokenEntity = persistTokenEntityHistorical(historicalRange);
+        persistFungibleTokenHistorical(tokenEntity, historicalRange);
+        domainBuilder
```
bilyana-gospodinova, line 78: These lines are common in multiple methods - they can be extracted in a separate method.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServicePrecompileHistoricalTest.java
```diff
@@ -16,352 +16,1322 @@
...
-            new Object[] {
-                NFT_HBAR_TOKEN_AND_KEYS_HISTORICAL,
-                2L,
+    private List<PrecompileTestContractHistorical.TokenKey> getExpectedTokenKeys(
+            final Entity tokenEntity, final Token token) {
+        final var expectedTokenKeys = new ArrayList<PrecompileTestContractHistorical.TokenKey>();
+        expectedTokenKeys.add(
+                new PrecompileTestContractHistorical.TokenKey(BigInteger.ONE, getKeyValue(tokenEntity.getKey())));
+        expectedTokenKeys.add(
+                new PrecompileTestContractHistorical.TokenKey(BigInteger.valueOf(2), getKeyValue(token.getKycKey())));
```
bilyana-gospodinova, line 1189: There are constants for these values in `AbstractContractCallServiceTest.KeyType`.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServicePrecompileHistoricalTest.java
```diff
@@ -16,352 +16,1322 @@
...
+        // When
+        final var functionCall =
+                contract.call_isTokenFrozen(getAddressFromEntity(tokenEntity), getAliasFromEntity(account));
+
+        // Then
+        assertThat(functionCall.send()).isTrue();
+    }
+
+    @ParameterizedTest
+    @CsvSource({"50", "49"})
```
bilyana-gospodinova, line 126: The logic behind the testing with 2 different blocks has changed here. Before the refactoring the test was:
1. Pass a historical block before the point where the data was persisted -> the test fails since the data is not persisted yet. Assert for an exception.
2. Persist the data.
3. Pass a historical block after the point where the data was persisted -> the test succeeds since the data is available at that moment.

Now after the refactoring the setup is the same for the two passed blocks - a record file is persisted, the entities are persisted and we make the contract call and it succeeds. It makes more sense to me to keep the previous logic. Otherwise, there is no point in performing the same test for 2 different blocks.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServicePrecompileHistoricalTest.java
kselveliev: It is common for some of the historical tests. But this PR is already big and any more changes to this and other classes make it harder to review. Prefer to do such refactoring in the next tasks we have for cleanup and optimizations of the tests

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServicePrecompileHistoricalTest.java
```diff
@@ -16,352 +16,1322 @@
...
 
-class ContractCallServicePrecompileHistoricalTest extends ContractCallTestSetup {
-    private static Stream<Arguments> htsContractFunctionArgumentsProviderHistoricalReadOnly() {
-        List<String> blockNumbers = List.of(String.valueOf(EVM_V_34_BLOCK - 1), String.valueOf(EVM_V_34_BLOCK));
+class ContractCallServicePrecompileHistoricalTest extends AbstractContractCallServiceTest {
 
-        return Arrays.stream(ContractReadFunctionsHistorical.values()).flatMap(htsFunction -> blockNumbers.stream()
-                .map(blockNumber -> Arguments.of(htsFunction, blockNumber)));
+    @ParameterizedTest
+    @CsvSource({"50", "49"})
```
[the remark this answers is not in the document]
kselveliev: Comment addressed

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServicePrecompileHistoricalTest.java
kselveliev: Hmm, I understand now. 
The tests in this class are aiming to execute the tests against block 49 - which loads evm 30 and block 50 which load evm - 34.
and verify everything is passing successfully with proper setup.

While with the previous version it’s pretty hard to understand since the setup is hidden somewhere in contractCallTestSetup and it’s almost impossible to understand.

I also don’t think we should be having a negative test for every single test in this class. Generally most of the tests will be using the same db queries. (query for historical nft, token, account)

So I’d rather we create a different test class only with negative cases which has like around 10 methods that will be using the historical db queries for nft, token, account, tokenRelationship, entity etc. and just verify there that exception will be thrown with no data in the db.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServicePrecompileHistoricalTest.java
bilyana-gospodinova: Ok, agreed. Can you please write this down as a comment in the cleanup task so that we won't forget?

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServicePrecompileHistoricalTest.java
bilyana-gospodinova: I am not sure if we are on the same page. I meant that `BigInteger.valueOf(2)` can be replaced with: `KeyType.KYC_KEY.getKeyTypeNumeric()` and so on.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServicePrecompileHistoricalTest.java
bilyana-gospodinova: I see. In this case can you please create a separate task for the new test class with the negative tests?

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServicePrecompileHistoricalTest.java
kselveliev: Correct. I extracted the tokenAccount persist code in a common method. It's hard to put account and token persist in one method since we'll need both persisted entities references when we call the contract method.

bilyana-gospodinova: You can return them as a `Pair`. Reference: `ContractCallServicePrecompileTest#persistTokenWithAutoRenewAndTreasuryAccounts`.

steven-sheehy (approved):

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServicePrecompileHistoricalTest.java
kselveliev: Done

steven-sheehy (approved):

PR #9270, opened 2024-09-04
Add Run Book for Creating and Restoring Disk Snapshots

### LINKED ISSUE #7763: Create Run Book To Restore Citus Cluster In Disaster Recovery
### Problem

We need a working plan in place to handle a citus cluster failure. 

### Solution

Create a run book that documents:

- [ ] Restoring a single node in the cluster
- [ ] Restoring all nodes within the cluster
- [ ] Moving to a new cluster entirely

### Alternatives

_No response_

jnels124:
**Description**:
* add run book for creating disk snapshots for citus cluster
* add run book for restoring disk snapshots for citus cluster

**Related issue(s)**:

Fixes #7763 

**Notes for reviewer**:

**Checklist**


docs/runbook/restore-citus-from-disk-snapshot.md
```diff
@@ -0,0 +1,314 @@
...
+
+```bash
+gcloud compute disks describe {diskName} --project {gcpProject} --zone {diskZone}
+gcloud compute instances detach-disk {instanceName}  --disk {diskName} --zone {diskZone} --project {gcpProject}
+gcloud compute disks delete {diskName} --project {gcpProject} --zone {diskZone}
+```
+
+## Steps
+
+1. Configure kubectl to point to the `target cluster`
```
xin-hedera, line 47: Is it not possible to do the following?

1. Use our charts to deploy and wait till the citus `SGShardedCluster` to come online. Doing so the disks will get created, attached to the worker and coordinator google cloud compute nodes, related zfs volume, pvc, and etc resources created and mounted to the citus coordinator / worker pods
2. Annotate `SGShardedCluster` with `stackgres.io/reconciliation-pause="true"` and  delete `SGClusters` (1mirror-citus-coordinator1, 1mirror-citus-shard{0,1,2}1) so the volumes are no longer mounted
4. Detach the disks
5. Restore the snapshots to the disks
6. Attach the disks and `zfs rename` so the match what's in the auto created PVC
7. Remove annotation `stackgres.io/reconciliation-pause="true"` from `SGShardedCluster` so stackgres operator will recreate `SGClusters`

If this works, we don't have to create pvc / pvc / zfsVolume

docs/runbook/restore-citus-from-disk-snapshot.md
```diff
@@ -0,0 +1,314 @@
...
+   `kubectl get sgcluster -n {targetNamespace} -o json | jq '.items[].metadata.name'`
+   <br>
+   `kubectl edit sgcluster -n {targetNamespace} {clusterName}`
+10. [update passwords](#update-passwords)
+11. Repeat steps 2-10 for each namespace where `hedera-mirror` chart is installed and contained in the restored
+    snapshots
+
+### Update Passwords
+
+1. `Stackgres` doesn't allow for the `superuser` or `replication` password to be set by configuration so you will need
```
xin-hedera, line 103: For step 1, as an alternative, we can update `postgres` password to the new one generated by stackgres. Ofc this needs to be done on the coordinator and all workers. However doing so doesn't need the old password.

docs/runbook/restore-citus-from-disk-snapshot.md
```diff
@@ -0,0 +1,314 @@
...
+alter user mirror_rosetta with password '';
+$cmd$);
+SELECT run_command_on_workers($cmd$
+alter user mirror_web3 with password '';
+$cmd$);
+SELECT run_command_on_workers($cmd$
+alter user replicator with password '';
+$cmd$);
+
+alter user replicator with password '';
```
xin-hedera, line 171: missing `authenticator`? Believe pgbouncer will fail if `authenticator`'s password isn't changed to match what's in the secrets.

docs/runbook/restore-citus-from-disk-snapshot.md
jnels124: This doesn't work :( tried that. Can't run `alter user postgres` on the shards. And altering the postgres user directly when using the postgres-util fails because can't communicate with the nodes.

docs/runbook/restore-citus-from-disk-snapshot.md
jnels124: Hmm i was wondering that myself but things seem to function correctly. Maybe will just add it for completeness though.

docs/runbook/restore-citus-from-disk-snapshot.md
jnels124: I can try your approach because on second thought would probably work because the error is on the patroni side and deleting and recreating the shards would probably work since it will go through initialization again. However, being able to access the nodes requires additional steps of configuring the firewall for ssh access or setting up up a pod to mount the volumes in a way that allows access to the nodes so zfs commands can be ran

docs/runbook/restore-citus-from-disk-snapshot.md
xin-hedera: what I can recall is it's not allowed when connected to `mirror_node` database, however this should work inside `patroni` / `postgres-util` container

`$ psql -c "alter role postgresl with password ''"`

jnels124: will try that

docs/runbook/restore-citus-from-disk-snapshot.md
xin-hedera: What I can confirm is when

- `stackgres.io/reconciliation-pause="true"` on `SGClusters`
- delete the sts `mirror-citus-coordinator`, `mirror-citus-shard0`, `mirror-citus-shard1`, and `mirror-citus-shard2`

The PVCs will stay

docs/runbook/restore-citus-from-disk-snapshot.md
jnels124: This works will update the docs to do that instead

docs/runbook/restore-citus-from-disk-snapshot.md
jnels124: Ok this should work then but not sure it is any better than just creating the resources due to having to configure firewall etc for ssh access to the nodes.

jnels124: the alternative to firewall changes is this something like this:
```
apiVersion: apps/v1
kind: DaemonSet
metadata:
  name: busybox-pod-c
spec:
  selector:
    matchLabels:
      mydaemon: busybox-pod-c
  template:
    metadata:
      labels:
        mydaemon: busybox-pod-c
    spec:
      hostNetwork: true
      hostPID: true
      hostIPC: true
      containers:
        - name: busybox
          image: ubuntu
          securityContext:
            privileged: true
          command:
            - tail
            - "-f"
            - /dev/null
          volumeMounts:
            - name: nodefs
              mountPath: /node-fs
            - name: devfs
              mountPath: /dev
      volumes:
        - name: nodefs
          hostPath:
            path: / 
        - name: devfs
          hostPath:
            path: /dev
      nodeSelector:
        kubernetes.io/hostname: gke-citus-worker-b07ff60d-0hdd
      tolerations:
        - effect: NoSchedule
          key: zfs
          operator: Equal
          value: "true"
```

PR #9272, opened 2024-09-04
Refactor `ContractCallServiceERCTokenTest` historical tests

nickeynikolovv:
**Description**:

This is the final PR for refactoring ContractCallServiceERCTokenTest 

**Related issue(s)**:

Fixes: #9168

**Checklist**


hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServiceERCTokenHistoricalTest.java
```diff
@@ -0,0 +1,1283 @@
...
+                        .autoRenewAccountId(autoRenewEntityId.getId())
+                        .num(tokenEntityId.getNum())
+                        .evmAddress(tokenAddress.toArrayUnsafe())
+                        .type(TOKEN)
+                        .balance(1500L)
+                        .key(KEY_PROTO)
+                        .expirationTimestamp(9999999999999L)
+                        .memo("TestMemo")
+                        .timestampRange(Range.atLeast(lowerTimestamp))
+                        .deleted(false))
```
IvanKavaldzhiev, line 1168: I believe we can set the num, key, expirationTimestamp and memo dynamically. This setup seems to mimic the way tokens were set in ContractCallTestSetup. Ideally, we would want as less hardcoded setup as possible.

The tokens shouldn't have balance, since it's not an account or contract.

If you need to assert specific fields like key, expirationTimestamp, memo etc. you can return the entity from the method and use getters to see what values the `DomainBuilder` has assigned.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServiceERCTokenHistoricalTest.java
```diff
@@ -0,0 +1,1283 @@
...
+                        .name("Hbars")
+                        .totalSupply(12345L)
+                        .decimals(12)
+                        .wipeKey(KEY_PROTO)
+                        .freezeKey(KEY_PROTO)
+                        .pauseStatus(TokenPauseStatusEnum.PAUSED)
+                        .pauseKey(KEY_PROTO)
+                        .supplyKey(KEY_PROTO)
+                        .symbol("HBAR")
+                        .timestampRange(Range.openClosed(lowerTimestamp, historicalBlock.upperEndpoint() + 1)))
```
IvanKavaldzhiev, line 1191: The same comment for the historical token. If you need to return both you can use `Pair<Entity, Entity>` as return type.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServiceERCTokenHistoricalTest.java
```diff
@@ -0,0 +1,1283 @@
...
+
+    @ParameterizedTest
+    @ValueSource(booleans = {true, false})
+    void emptyOwnerOfAfterEvmV34(final boolean isStatic) throws Exception {
+        // Given
+        recordFileAfterEvmV34Persist();
+        testWeb3jService.setBlockType(BlockType.of(String.valueOf(EVM_V_34_BLOCK)));
+        testWeb3jService.setHistoricalRange(
+                Range.closedOpen(recordFileAfterEvm34.getConsensusStart(), recordFileAfterEvm34.getConsensusEnd()));
+        final var tokenAddress = toAddress(1063);
```
bilyana-gospodinova, line 896: Why do we need to hardcode the addresses? Can't we use the dynamic addresses from the persisted entities?

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServiceERCTokenHistoricalTest.java
```diff
@@ -0,0 +1,1283 @@
...
+                .nft()
+                .customize(n -> n.accountId(spenderEntityId)
+                        .createdTimestamp(1475067194949034022L)
+                        .serialNumber(3L)
+                        .metadata("NFT_METADATA_URI".getBytes())
+                        .accountId(ownerEntity)
+                        .tokenId(nftEntityId.getId())
+                        .deleted(false)
+                        .timestampRange(Range.atLeast(historicalBlock.upperEndpoint() + 1)))
+                .persist();
```
IvanKavaldzhiev, line 1127: As per comments below, a lot of the data could be left to be dynamically set by `DomainBuilder` for all these entity types and the setup will decrease in size and complexity.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServiceERCTokenHistoricalTest.java
```diff
@@ -0,0 +1,1283 @@
...
+                : contract.call_nameNonStatic(tokenAddress.toHexString()).send();
+        // Then
+        assertThat(result).isEqualTo("Hbars");
+    }
+
+    @ParameterizedTest
+    @ValueSource(booleans = {true, false})
+    void ownerOfBeforeEvmV34(final boolean isStatic) {
+        // Given
+        recordFileBeforeEvmV34Persist();
```
bilyana-gospodinova, line 798: I think we can combine the "beforeEvm34" tests in a nested class and move the first 4 lines of the tests to a `@BeforeEach` and setup the record files and the testWeb3jService's state there. This would make the code a bit cleaner. The same is valid for the "afterEvm34" tests.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServiceERCTokenHistoricalTest.java
```diff
@@ -0,0 +1,1283 @@
...
+import com.hedera.mirror.web3.web3j.generated.ERCTestContractHistorical;
+import java.math.BigInteger;
+import org.hyperledger.besu.datatypes.Address;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+class ContractCallServiceERCTokenHistoricalTest extends AbstractContractCallServiceTest {
+
+    private RecordFile recordFileBeforeEvm34;
+    protected static RecordFile recordFileAfterEvm34;
```
kselveliev, line 56: Could be private and non static

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServiceERCTokenHistoricalTest.java
nickeynikolovv: Fixed - I refactored the structure and persisting of the data

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServiceERCTokenHistoricalTest.java
nickeynikolovv: Fixed - I refactored the structure and persisting of the data, now the dynamic addresses are used

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServiceERCTokenHistoricalTest.java
nickeynikolovv: Fixed - I refactored the structure and persisting of the data, now the dynamic addresses are used

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServiceERCTokenHistoricalTest.java
nickeynikolovv: Done

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServiceERCTokenHistoricalTest.java
```diff
@@ -0,0 +1,1283 @@
...
+                        .alias(SENDER_PUBLIC_KEY.toByteArray())
+                        .balance(10000 * 100_000_000L)
+                        .createdTimestamp(recordFileAfterEvm34.getConsensusStart())
+                        .timestampRange(Range.closedOpen(
+                                recordFileAfterEvm34.getConsensusStart(), recordFileAfterEvm34.getConsensusEnd())))
+                .persist();
+        return senderEntityId;
+    }
+
+    private EntityId spenderEntityPersistHistorical(Address spenderAddress) {
```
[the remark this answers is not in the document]
nickeynikolovv: After the refactoring this method is not used anymore

steven-sheehy (approved):
LGTM

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServiceERCTokenHistoricalTest.java
```diff
@@ -0,0 +1,1016 @@
...
+import org.junit.jupiter.params.provider.ValueSource;
+
+class ContractCallServiceERCTokenHistoricalTest extends AbstractContractCallServiceTest {
+    private Range<Long> historicalRangeAfterEvm34;
+
+    @Nested
+    class beforeEvm34Tests {
+        @BeforeEach
+        void beforeEach() {
+            historicalRangeAfterEvm34 = setUpHistoricalContextBeforeEvm34();
```
bilyana-gospodinova, line 57: The historical range "after" evm 34 is set from a method "before" evm 34. Is that expected as it looks confusing? Perhaps this field can be moved in the nested classes and renamed.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServiceERCTokenHistoricalTest.java
```diff
@@ -0,0 +1,1016 @@
...
+        }
+
+        @ParameterizedTest
+        @ValueSource(booleans = {true, false})
+        void getApprovedEmptySpenderBeforeEvmV34(final boolean isStatic) {
```
bilyana-gospodinova, line 62: Since the nested classes contain in their name the before/after EVM version, I don't think we need the EVM version in the test names as well.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServiceERCTokenHistoricalTest.java
```diff
@@ -0,0 +1,1016 @@
...
+            // Given
+            final var tokenEntity = persistTokenEntityHistorical(historicalRangeAfterEvm34);
+            domainBuilder
+                    .token()
```
bilyana-gospodinova, line 66: The `token`, `nft`, etc. tables contain the data only for the latest block (an exception is the `entity` table). The entries in the mentioned tables have only the lower timestamp. The data that is older than the latest block is kept in the `*history` tables, e.g. `tokenHistory`, `nftHistory`, etc. The tests use mainly the non-history tables. Having in mind that they test against EVM 30 and 34 the data would be in the history tables. The tests pass but the setup does not fully simulate the real-world data.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServiceERCTokenHistoricalTest.java
```diff
@@ -0,0 +1,1016 @@
...
+        @ParameterizedTest
+        @ValueSource(booleans = {true, false})
+        void allowanceBeforeEvmV34(final boolean isStatic) {
+            // Given
+            final var owner = persistAccountEntityHistorical(historicalRangeAfterEvm34);
+            final var spender = persistAccountEntityHistorical(historicalRangeAfterEvm34);
+            final var token = persistFungibleTokenHistorical(historicalRangeAfterEvm34);
+            final var amountGranted = 10L;
+
+            domainBuilder
```
bilyana-gospodinova, line 167: All of the domainBuilder-constructed entities can be separated in a common method. This would make the task for the final cleanup easier and the overall test size smaller and easier to read.

nickeynikolovv: done

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServiceERCTokenHistoricalTest.java
nickeynikolovv: done

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServiceERCTokenHistoricalTest.java
nickeynikolovv: fixed

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServiceERCTokenHistoricalTest.java
nickeynikolovv: it was really confusing so i renamed it to just historicalRange

kselveliev (dismissed):

bilyana-gospodinova (approved):
LGTM

steven-sheehy (approved):

Issue #9222, opened 2024-09-02
Combine duplicated helper methods in the web3 integration tests


bilyana-gospodinova:
### Problem

Most of the test smart contracts in the web3 integration tests inherit the `HederaTokenService` smart contract which uses `IHederaTokenService`. There are some inner structs in `IHederaTokenService` which web3j generates multiple times for each child contract. As a result there are multiple java classes of the same type that do not have a common parent. For example: `Expiry` in `NestedCalls` and  `Expiry` in `PrecompileTestContract`. They are both constructed the same way, such as:
```
private Expiry getTokenExpiry(final Entity autoRenewAccountEntity) {
        return new Expiry(
                BigInteger.valueOf(4_000_000_000L),
                toAddress(autoRenewAccountEntity.toEntityId()).toHexString(),
                BigInteger.valueOf(8_000_000L));
    }
```

### Solution

It would be nice if we don't need to duplicate the same methods only to return the different class type. This way we can reduce the code length and improve the maintainability.

bilyana-gospodinova:
This task was implemented in https://github.com/hashgraph/hedera-mirror-node/pull/9278. It seems that it did not add more value, hence closing the issue.

PR #9278, opened 2024-09-05
Remove duplicated methods by introducing factories

bilyana-gospodinova:
**Description**:
The `web3j` library generates a lot of duplicated inner classes due to the same imports in multiple contracts. Example for these classes are: `TokenKey`, `HederaToken`, `Expiry`, `KeyValue`. With the current implementation there are a lot of helper methods that construct object for these classes and they are duplicated because of the return type, e.g. `ModificationPrecompileTestContract.Expiry`, `NestedCalls.Expiry`, etc. This PR aims to introduce factories that return the correct type depending of the caller class. Now the duplicated methods are removed and there are new common generic methods in the parent abstract class. The execution time of the integration tests remains the same after the refactoring in this PR.

Fixes https://github.com/hashgraph/hedera-mirror-node/issues/9222

**Checklist**

### DIFF
--- a/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
+++ b/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
@@ -281,10 +281,10 @@ public abstract class AbstractContractCallServiceTest extends Web3IntegrationTes
             final Boolean inheritAccountKey,
             final String contractId,
             final byte[] ed25519,
-            final byte[] ECDSA_secp256k1,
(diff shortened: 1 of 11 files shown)


IvanKavaldzhiev (approved):
I like the approach for dynamically returning the correct type that is needed! 

LGTM

bilyana-gospodinova:
Synced with main

steven-sheehy (changes requested):
I have to agree with @kselveliev. I don't see much value in this change. I prefer the reptition as it's easier to understand and maintain then reflection-based code. Also, if the point was to reduce code we actually added more lines than removed:

Screenshot 2024-09-11 at 19 42 28

PR #9303, opened 2024-09-09
Remove `web3j-cli`

### LINKED ISSUE #9271: Remove the `web3j-cli` and use only the `web3j-gradle-plugin`
The current download script is unstable, so it would be much better if we don't need to use the `web3j-cli` at all. This would need either defining multiple gradle tasks, or some changes in the implementation upstream.

bilyana-gospodinova:
**Description**
This PR removes the custom scripts that download the `web3j-cli` and compile the historical contracts with it. The need for that is that the scripts are not fully stable. Also, the download script always downloads the latest `web3j-cli` version (if the cached local folder is removed), so this way the latest version could possibly contain some breaking changes that might break our code as well. Now the `solidity-gradle-plugin` and the `web3j-gradle-plugin` are used as gradle dependencies. This fixes the issues that are described above.

Fixes #9271


hedera-mirror-web3/build.gradle.kts
```diff
@@ -119,31 +126,114 @@ val extractOpenZeppelin =
...
+// The generated java files are moved from "testHistorical" to "test" folder in build so that
+// they can be resolved successfully in imports.
+tasks.register<Copy>("moveTestHistoricalFiles") {
+    description = "Move files from testHistorical to test in the generated web3j sources directory."
+    group = "historical"
+
+    // Define source and destination directories
+    val srcDir =
+        file(
+            "$rootDir/hedera-mirror-web3/build/generated/sources/web3j/testHistorical/java/com/hedera/mirror/web3/web3j/generated"
```
steven-sheehy, line 186: Use `layout.buildDirectory`

hedera-mirror-web3/build.gradle.kts
```diff
@@ -119,31 +126,114 @@ val extractOpenZeppelin =
...
+    }
+    dependsOn(tasks.named("generateTestContractWrappers"))
+    dependsOn(tasks.named("generateTestHistoricalContractWrappers"))
+}
+
+tasks.register<Delete>("deleteTestHistoricalDir") {
+    description = "Delete the testHistorical directory after moving files."
+    group = "historical"
+
+    delete(file("$rootDir/hedera-mirror-web3/build/generated/sources/web3j/testHistorical"))
```
steven-sheehy, line 211: Can we just manually delete this in a `doLast` inside `moveTestHistoricalFiles`? Then we can avoid the extra tasks `deleteTestHistoricalDir` and `moveAndCleanTestHistorical`

hedera-mirror-web3/build.gradle.kts
```diff
@@ -119,31 +126,114 @@ val extractOpenZeppelin =
...
+    // Copy only files that match the pattern "*Historical.java"
+    from(srcDir) { include("**/*Historical.java") }
+    into(destDir)
+
+    // Only move if there are files
+    doFirst {
+        if (!srcDir.exists()) {
+            throw GradleException("Source directory $srcDir does not exist!")
+        }
+    }
```
steven-sheehy, line 202: The comment makes it seems like it's an `onlyIf` but it's halting the build if no source. Should we use  `onlyIf` instead or can we just remove this logic since a prior task would've already failed for it to be missing?

steven-sheehy (changes requested):
Do we still the ticket upstream if we found a way to accomplish it with the plugin?

hedera-mirror-web3/build.gradle.kts
```diff
@@ -60,41 +62,46 @@ web3j {
...
 
 val downloadOpenZeppelin =
     tasks.register("downloadOpenZeppelin") {
         description = "Download OpenZeppelin contracts"
         group = "historical"
         doLast {
+            val buildDir = layout.buildDirectory.asFile.get()
+            if (!buildDir.exists()) {
+                mkdir(buildDir)
+            }
```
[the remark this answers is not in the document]
bilyana-gospodinova: Comment addressed

hedera-mirror-web3/build.gradle.kts
```diff
@@ -119,31 +126,114 @@ val extractOpenZeppelin =
...
 
-// Task to compile Solidity contracts and generate Java files
-val compileHistoricalSolidityContracts =
-    tasks.register<Exec>("compileHistoricalSolidityContracts") {
-        description = "Compiles the historical solidity contracts to java files using web3j-cli"
+tasks.openApiGenerate { mustRunAfter(tasks.named("resolveSolidity")) }
+
+tasks.processTestResources {
+    dependsOn(tasks.named("generateTestContractWrappers"))
+    dependsOn(tasks.named("generateTestHistoricalContractWrappers"))
```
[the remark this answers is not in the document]
bilyana-gospodinova: This task is named automatically by the web3j plugin. The convention is: "generate" + <source_set> + "ContractWrappers" and the source set is "testHistorical".

hedera-mirror-web3/build.gradle.kts
bilyana-gospodinova: Comment addressed

hedera-mirror-web3/build.gradle.kts
bilyana-gospodinova: Comment addressed.

hedera-mirror-web3/build.gradle.kts
bilyana-gospodinova: This check seems redundant now. I am removing it.

hedera-mirror-web3/build.gradle.kts
```diff
@@ -119,31 +126,114 @@ val extractOpenZeppelin =
...
-        doFirst { file(scriptPath).setExecutable(true) }
-        commandLine("bash", scriptPath)
     }
+}
 
-tasks.assemble { dependsOn(tasks.processTestResources) }
+afterEvaluate {
+    tasks.named("generateTestHistoricalContractWrappers") {
+        dependsOn(tasks.named("generateTestContractWrappers"))
+        dependsOn(extractOpenZeppelin)
```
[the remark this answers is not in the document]
bilyana-gospodinova: Actually, it seems that the plugin resolves the `openzeppelin` files by itself and we don't need the task to download the contracts anymore.

kselveliev (dismissed):

hedera-mirror-web3/build.gradle.kts
```diff
@@ -60,90 +60,115 @@ web3j {
...
+
+    // Copy only files that match the pattern "*Historical.java"
+    from(srcDir) { include("**/*Historical.java") }
+    into(destDir)
+
+    doLast {
+        val testHistoricalDir =
+            layout.buildDirectory.dir("generated/sources/web3j/testHistorical").get().asFile
+        testHistoricalDir.deleteRecursively()
+        println("Deleted testHistorical directory: $testHistoricalDir")
```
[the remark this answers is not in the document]
bilyana-gospodinova: Removed

PR #9318, opened 2024-09-10
Enhance contract deployment in integration tests

### LINKED ISSUE #9109: Refactor `TestWeb3jService` to not actually deploy the contracts wherever possible
### Problem

The idea for this came from the PR comment in [here](https://github.com/hashgraph/hedera-mirror-node/pull/9076#discussion_r1723794388). 

### Solution

Basically, we can enhance the integration tests refactoring further if we intercept the actual contract deploy and just create the contract as a java object and set its address. The address can be taken from the previously persisted entity in the DomainBuilder. This way the overall execution time of the tests will be decreased and we will avoid some redundant initialisations.

### Alternatives

_No response_

kselveliev:
**Description**:

This pr improves contract deployment during tests.

Previously on deploy we `sendTopLevelCreate`
and used the mirrorResult from 
`contractExecutionService.processCall(serviceParameters);` which returns the contract runtime code.

With this new change: instead of calling `contractExecutionService.processCall` we are using a new utility method which quickly extracts 
the runtime bytecode from the `rawTransaction.getData() `which is the init bytecode.

Using this approach we skip the need for calling `contractExecutionService.processCall` in 99% of the test cases. For some corner cases we still use the old approach with the `contractExecutionService.processCall` .

Couple of possible scenarios explored but **not applied in this PR**.

 - Find a way to even skip persisting the contract to db on every call. Did not find an straightorward way to do that.

 - Add a before each method in every test suite which with the help of a flag will deploy the contract only in the first before 
each.
and the rest of the before each executions will do nothing. This will mean we only deploy the contract only once during a full test class execution. Unfortunately this approach led to 0 peformance gain.

This PR modifies:
Add a utility class RuntimeBytecodeExtractor - which has a method that returns the runtime bytecode and takes init bytecode as a parameter.

Adds a test class RuntimeBytecodeExtractorTest - which tests the utility class method with different scenarios.

Modified TestWeb3jService - adds a new flag - which when enabled uses `contractExecutionService.processCall` for contract deployment. By default it will use the approach with using the new utility class.

Modified couple of corner test cases to preserve contract deployment with `contractExecutionService.processCall`

**Related issue(s)**:

Fixes #9109 

**Checklist**


hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/utils/RuntimeBytecodeExtractorTest.java
```diff
@@ -0,0 +1,76 @@
...
...6103da8161037e565b600182811680156103f2576001811461040757610431565b60ff1984168752821515830287019450610431565b895f5260205f205f5b8481101561042957815489820152908301908701610410565b505082870194505b50929998505050505050505050565b5f60208284031215610450575f80fd5b815167ffffffffffffffff811115610466575f80fd5b8201601f81018413610476575f80fd5b80516104846102f482610289565b818152856020838501011115610498575f80fd5b6104a982602083016020860161032a565b95945050505050565b601f8211156104f957805f5260205f20601f840160051c810160208510156104d75750805b601f840160051c820191505b818110156104f6575f81556001016104e3565b50505b505050565b815167ffffffffffffffff81111561051857610518610244565b61052c81610526845461037e565b846104b2565b602080601f83116001811461055f575f84156105485750858301515b5f19600386901b1c1916600185901b1785556105b6565b5f85815260208120601f198616915b8281101561058d5788860151825594840194600190910190840161056e565b50858210156105aa57878501515f19600388901b60f8161c191681555b505060018460011b0185555b50505050505056fea2646970667358221220f9d4ec29ea6039c4ca833a4f380cc942538bdf4f493376c24d4853357941676664736f6c63430008180033a2646970667358221220992b617f20432b99cb9340eed5a817a1ba6bd3d2256265ae4753da69bc2f6da464736f6c63430008180033";
```
steven-sheehy, line 34: Instead of this large binary can we use a generated contract static `BINARY` field as input and a call to `contractExecutionService.processCall()` to get the runtime bytecode result?

steven-sheehy (dismissed):
LGTM

nickeynikolovv (approved):
LGTM

bilyana-gospodinova (dismissed):
LGTM, but agree with https://github.com/hashgraph/hedera-mirror-node/pull/9318#discussion_r1752958873

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/utils/RuntimeBytecodeExtractorTest.java
kselveliev: Addressed

steven-sheehy (approved):
LGTM

mgoelswirlds (approved):

PR #9327, opened 2024-09-11
Reducing cache-control for scheduled transactions.

### LINKED ISSUE #9174: Reduce cache for the transactions endpoint with scheduled transactions
### Problem

When we make scheduled transactions and right after that we call the endpoint `/transactions/:transactionId`, we get that the `ScheduleCreate` was done, but the outcome of the scheduled transaction is not received for 10 minutes. This is due to the cache time of the `/transactions/:transactionId` endpoint.

### Solution

- Reduce the cache period for a schedule creates without a corresponding `schedule=true`.

### Alternatives

_No response_

mgoelswirlds:
**Description**:
Reduce cache for the transactions endpoint with scheduled transactions.

This PR modifies 
* Reduces the value of max-age for `ScheduleCreate` transactions
* Changes the way `content-type` is set for the ` /network/supply` endpoint

**Related issue(s)**:

Fixes #9174

**Notes for reviewer**:
**Checklist**


hedera-mirror-rest/__tests__/specs/transactions/{id}/no-params.json
```diff
@@ -177,118 +177,6 @@
     ]
   },
   "tests": [
-    {
```
mgoelswirlds, line 180: This test case has been removed in order for the builds to pass. I have a question here , if it makes sense to separate the tests out with scheduled transactions from the others.

hedera-mirror-rest/middleware/responseHandler.js
```diff
@@ -37,8 +37,10 @@ const responseHandler = async (req, res, next) => {
     throw new NotFoundError();
   } else {
     const path = res.locals[requestPathLabel] ?? req.route.path;
-    res.set(headers.default);
-    res.set(headers.path[path]);
+    if (!res.get('cache-control')) {
```
steven-sheehy, line 40: Use a constant from express

hedera-mirror-rest/middleware/responseHandler.js
```diff
@@ -37,8 +37,10 @@ const responseHandler = async (req, res, next) => {
...
+      res.set(headers.default);
+      res.set(headers.path[path]);
+    }
```
steven-sheehy, line 43: We should not omit all headers from config just because of one header being present. Even though it only has cache-control now, `response.headers` property is intended as a generic way to add extra headers not just cache-control.

hedera-mirror-rest/transactions.js
```diff
@@ -847,6 +850,18 @@ const getTransactionsByIdOrHash = async (req, res) => {
...
   const transactions = await formatTransactionRows(rows);
 
+  if (transactions.length > 1) {
+    for (const transaction of transactions) {
+      const cacheControlMaxAge = 5;
+      if (
+        transaction.name === scheduleCreate &&
+        !filters.some((f) => f.key === constants.filterKeys.SCHEDULED && f.value === true)
+      ) {
+        res.set(CACHE_CONTROL_HEADER_NAME, `public, max-age=${cacheControlMaxAge}`);
```
steven-sheehy, line 860: Use a constants for value

steven-sheehy, line 860: Add a break after first instance to optimize.

hedera-mirror-rest/transactions.js
```diff
@@ -48,6 +48,9 @@ const {
 
 const cache = new Cache();
 
+const scheduleCreate = TransactionType.getName(42);
```
steven-sheehy, line 51: Instead of converting a hardcoded number to a string, just hardcode the string directly here.

hedera-mirror-rest/transactions.js
```diff
@@ -847,6 +850,18 @@ const getTransactionsByIdOrHash = async (req, res) => {
 
   const transactions = await formatTransactionRows(rows);
 
+  if (transactions.length > 1) {
+    for (const transaction of transactions) {
+      const cacheControlMaxAge = 5;
+      if (
+        transaction.name === scheduleCreate &&
+        !filters.some((f) => f.key === constants.filterKeys.SCHEDULED && f.value === true)
```
steven-sheehy, line 858: This logic is not correct. Regardless of filters, the transaction rows can contain a scheduled transaction that should not be cached as long. Specifically we should look for a schedule create that does not have a corresponding scheduled=true for the same transaction ID row in the same list.

hedera-mirror-rest/__tests__/specs/transactions/{id}/scheduled-transactions/scheduled-false.json
steven-sheehy, line 1: This violates the folder hierarchy. We should not create folders that don't correspond to url paths.

hedera-mirror-rest/__tests__/specs/transactions/{id}/no-params.json
xin-hedera: The spec schema supports a `responseHeaders` to set test specific response headers to validate.

Perhaps the logic in `template.js` needs to be first add folder specific response headers or the default, then merge `spec.responseHeaders`

```javascript
const getResponseHeaders = (spec, specPath) => {
  spec.responseHeaders = {
    ...(spec.responseHeaders ?? {}),
    ...getResponseHeadersFromFileOrDefault(specPath),
  };
};
```

xin-hedera (changes requested):

hedera-mirror-rest/transactions.js
```diff
@@ -847,6 +850,18 @@ const getTransactionsByIdOrHash = async (req, res) => {
...
+      ) {
+        res.set(CACHE_CONTROL_HEADER_NAME, `public, max-age=${cacheControlMaxAge}`);
+      }
+    }
+  }
```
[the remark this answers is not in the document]
xin-hedera: we may want to and a configurable duration to go back to the default longer max-age for the situation that the scheduled transaction never executes (e.g. when not enough signatures collected in 30 minutes)

hedera-mirror-rest/__tests__/integration/template.js
```diff
@@ -91,8 +91,8 @@ const getResponseHeadersFromFileOrDefault = (specPath) => {
 
 const getResponseHeaders = (spec, specPath) => {
   spec.responseHeaders = {
-    ...(spec.responseHeaders ?? {}),
     ...getResponseHeadersFromFileOrDefault(specPath),
```
mgoelswirlds, line 94: The order of this has been changed in the template so that the headers from the spec file can take precedence.

hedera-mirror-rest/transactions.js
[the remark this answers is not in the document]
mgoelswirlds: Still need to address the configurable duration part.

hedera-mirror-rest/middleware/responseHandler.js
mgoelswirlds: Removed this with current changes.

hedera-mirror-rest/transactions.js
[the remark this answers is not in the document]
xin-hedera: should use the existing `maxTransactionConsensusTimestampRange` config, the default value is based on how long a schedule is valid for getting ready to execute, i.e., 30 minutes

there are more corner cases to consider

- if the `SCHEDULECREATE` tx fails (note there can be duplicate `SCHEDULECREATE` txs all reaching consensus, so here fail means all `SCHEDULECREATE` with the same tx id fail), don't override the max-age
- if there is `scheduled=?` in the request URL, don't override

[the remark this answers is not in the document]
xin-hedera: one more corner case: if it's a query by transaction hash, don't override, since a transaction hash should always map to exact one transaction, if it exists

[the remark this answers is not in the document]
mgoelswirlds: Ok Thanks. Looking into the corner cases now.

hedera-mirror-rest/transactions.js
```diff
@@ -830,6 +837,48 @@ const extractSqlFromTransactionsByIdOrHashRequest = async (transactionIdOrHash,
...
+        continue;
+      }
+
+      timestamp = math.subtract(currSecs, math.bignumber(transaction.consensus_timestamp).toNumber());
+      hasScheduleCreate = true;
+    } else if (transaction.scheduled === true) {
+      scheduleExecuted = true;
+    } else if (
+      timestamp > Number(config.query.maxTransactionConsensusTimestampRangeNs / 1_000_000_000n) &&
+      !transaction.scheduled
```
steven-sheehy, line 870: Can remove this since `transaction.scheduled` has to equal false here (the db column is not null and has a default).

hedera-mirror-rest/transactions.js
```diff
@@ -830,6 +837,48 @@ const extractSqlFromTransactionsByIdOrHashRequest = async (transactionIdOrHash,
...
+  }
+
+  for (const transaction of transactions) {
+    if (transaction.name === scheduleCreate) {
+      if (!isTransactionSuccessful(TransactionResult.getProtoId(transaction.result))) {
+        // If all duplicate SCHEDULECREATE transactions_t fail, the cache max-age will not be shortened.
+        continue;
+      }
+
+      timestamp = math.subtract(currSecs, math.bignumber(transaction.consensus_timestamp).toNumber());
```
steven-sheehy, line 864: `transaction.consensus_timestamp` should already be a bigint. And it seems to be doing arithmetic with mixed units. `currSecs` is in seconds but `math.bignumber(transaction.consensus_timestamp).toNumber()` will be in nanoseconds.

hedera-mirror-rest/transactions.js
```diff
@@ -830,6 +837,48 @@ const extractSqlFromTransactionsByIdOrHashRequest = async (transactionIdOrHash,
...
+}
+
+function getTransactionsByIdOrHashCacheControlHeader(transactions, transactionIdOrHash, filters) {
+  let hasScheduleCreate = false;
+  let scheduleExecuted = false;
+  const currSecs = Math.floor(new Date().getTime() / 1000);
+  let timestamp = 0;
+  if (filters.some((f) => f.key === constants.filterKeys.SCHEDULED) || isValidTransactionHash(transactionIdOrHash)) {
+    return {}; // no override
+  }
```
steven-sheehy, line 855: If transaction hash or scheduled=true then there will be only one result in the list so we can simplify.

```suggestion
  if (transactions.length < 2 || filters.some((f) => f.key === constants.filterKeys.SCHEDULED)) {
    return {}; // no override
  }
```

hedera-mirror-rest/transactions.js
steven-sheehy, line 870: The timestamp logic is incorrect. Instead of calculating it inline, you only need to check it inside `if (hasScheduleCreate && !scheduleExecuted)` at the end. 

But it's tricky if there's multiple schedule creates in the list. Not sure it's really worth trying to micro-optimize for this corner case. Be simpler to just use the shorter cache.

steven-sheehy (changes requested):

mgoelswirlds:
I still need to address the new comments on the PR @steven-sheehy

hedera-mirror-rest/transactions.js
```diff
@@ -830,6 +837,48 @@ const extractSqlFromTransactionsByIdOrHashRequest = async (transactionIdOrHash,
...
+    result === successMissingOperationTransactionResult
+  );
+}
+
+function getTransactionsByIdOrHashCacheControlHeader(transactions, transactionIdOrHash, filters) {
+  let hasScheduleCreate = false;
+  let scheduleExecuted = false;
+  const currSecs = Math.floor(new Date().getTime() / 1000);
+  let timestamp = 0;
+  if (filters.some((f) => f.key === constants.filterKeys.SCHEDULED) || isValidTransactionHash(transactionIdOrHash)) {
```
xin-hedera, line 853: should return a query context object from `extractSqlFromTransactionsByIdOrHashRequest` with fields to indicate if there is `scheduled` query param and transaction hash path variable, instead of reparsing.

hedera-mirror-rest/transactions.js
```diff
@@ -830,6 +837,48 @@ const extractSqlFromTransactionsByIdOrHashRequest = async (transactionIdOrHash,
...
+    result === successTransactionResult ||
+    result === successFeeSchedulePartUploadedTransactionResult ||
+    result === successMissingOperationTransactionResult
+  );
+}
+
+function getTransactionsByIdOrHashCacheControlHeader(transactions, transactionIdOrHash, filters) {
+  let hasScheduleCreate = false;
+  let scheduleExecuted = false;
+  const currSecs = Math.floor(new Date().getTime() / 1000);
```
xin-hedera, line 851: `utils.nowInNs()`

hedera-mirror-rest/transactions.js
```diff
@@ -830,6 +837,48 @@ const extractSqlFromTransactionsByIdOrHashRequest = async (transactionIdOrHash,
...
+      scheduleExecuted = true;
+    } else if (
+      timestamp > Number(config.query.maxTransactionConsensusTimestampRangeNs / 1_000_000_000n) &&
+      !transaction.scheduled
+    ) {
+      //If the scheduled transaction never executes in the given timeframe , the cache max-age will not be shortened
+      return {};
+    }
+  }
+  if (hasScheduleCreate && !scheduleExecuted) {
```
xin-hedera, line 876: the logic starting from the for loop with the corner cases are way too complicated and buggy.

```
let successScheduleCreateTimestamp;

foreach transaction

  if transaction is SCHEDULECREATE
    if transaction is successful
      successScheduleCreateTimestamp = transaction's consensus timestamp;
    else if transaction.scheduled
      return SHORTER_CACHE_CONTROL_HEADER;

if successScheduleCreateTimestamp !== undefined && nowInNs() - successScheduleCreateTimestamp < maxTransactionConsensusTimestampRangeNs
  return SHORTER_CACHE_CONTROL_HEADER;

return {};
```

hedera-mirror-rest/__tests__/specs/transactions/{id}/scheduled-create-non-scheduled-transaction.json
```diff
@@ -0,0 +1,135 @@
...
+            "is_approval": true
+          }
+        ],
+        "entity_id": "0.0.1000"
+      }
+    ],
+    "cryptotransfers": [
+      {
+        "consensus_timestamp": "1896279200000000003",
+        "valid_start_timestamp": "1896279200000000001",
```
xin-hedera, line 45: this is impossible: two transactions with the same transaction id, both are successful, none of them has `scheduled = true`.

hedera-mirror-rest/transactions.js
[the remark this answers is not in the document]
mgoelswirlds: Removing these.

hedera-mirror-rest/transactions.js
mgoelswirlds: Removing the condition due to the changes suggested by Xin.

hedera-mirror-rest/transactions.js
xin-hedera: @mgoelswirlds I updated the pseudo code to fix the `else if` indentation

hedera-mirror-rest/transactions.js
mgoelswirlds: Actually, `transaction.consensus_timestamp ` is not a bigint so it would need explicit conversion. But will fix the equation.

xin-hedera: Then you should go back use rows instead of the formatted transactions array

hedera-mirror-rest/transactions.js
xin-hedera: the changes made don't match what I asked for: `if there is scheduled query param` is not `if scheduled flag is true`

hedera-mirror-rest/transactions.js
```diff
@@ -766,17 +771,51 @@ const extractSqlFromTransactionsByIdOrHashRequest = async (transactionIdOrHash,
...
+
+  let successScheduleCreateTimestamp;
+
+  for (const transaction of transactions) {
+    if (transaction.name === scheduleCreate) {
+      if (TransactionResult.isSuccessful(transaction.result)) {
+        successScheduleCreateTimestamp = successScheduleCreateTimestamp =
+          BigInt(math.bignumber(transaction.consensus_timestamp).toNumber()) * constants.NANOSECONDS_PER_SECOND;
+      } else if (transaction.scheduled) {
+        return SHORTER_CACHE_CONTROL_HEADER;
```
xin-hedera, line 795: I updated my pseudo code and notified you, this is wrong logic.

xin-hedera (changes requested):

hedera-mirror-rest/middleware/responseHandler.js
```diff
@@ -15,7 +15,13 @@
  */
 
 import config from '../config';
-import {requestPathLabel, requestStartTime, responseContentType, responseDataLabel} from '../constants';
+import {
+  requestPathLabel,
+  requestStartTime,
+  responseContentType,
```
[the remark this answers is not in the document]
mgoelswirlds: I did try changing this. Works for network supply but many other tests fail complaining about the `Invalid media type`. Looking further into what needs to be resolved here.

hedera-mirror-rest/middleware/responseHandler.js
```diff
@@ -37,11 +43,17 @@ const responseHandler = async (req, res, next) => {
...
+      ...headers.default,
+      ...(headers.path[path] ?? {}),
+      ...(res.locals[responseHeadersLabel] ?? {}),
+    };
+    res.set(mergedHeaders);
 
     const code = res.locals.statusCode;
-    const contentType = res.locals[responseContentType] || APPLICATION_JSON;
+    const responseHeader = res.locals.responseHeadersLabel;
+    const contentType = (responseHeader === undefined || typeof responseHeader === 'object') ? APPLICATION_JSON : responseHeader ;
```
xin-hedera, line 55: why is such logic needed?

Will this not work?

- default to application json content type, i.e., put it into the `headers.default` config (in the default application.yml file`
- override if needed in individual handlers

hedera-mirror-rest/transactions.js
```diff
@@ -766,17 +773,50 @@ const extractSqlFromTransactionsByIdOrHashRequest = async (transactionIdOrHash,
...
 
+function getTransactionsByIdOrHashCacheControlHeader(transactionsRows, isTransactionHash, scheduled) {
+  if (scheduled || isTransactionHash) {
+    return {}; // no override
+  }
+
+  let successScheduleCreateTimestamp;
+
+  for (const transaction of transactionsRows) {
+    if (TransactionType.getName(transaction.type) === scheduleCreate) {
```
xin-hedera, line 792: nit, just compare the transaction type proto id

hedera-mirror-rest/transactions.js
```diff
@@ -766,17 +773,50 @@ const extractSqlFromTransactionsByIdOrHashRequest = async (transactionIdOrHash,
...
+    }
+  }
+  if (
+    successScheduleCreateTimestamp !== undefined &&
+    utils.nowInNs() - successScheduleCreateTimestamp < maxTransactionConsensusTimestampRangeNs
+  ) {
+    return SHORTER_CACHE_CONTROL_HEADER;
+  }
+  return {}; // no override
+}
```
xin-hedera, line 807: spacing please

hedera-mirror-rest/middleware/responseHandler.js
mgoelswirlds: This logic was needed because network controller was setting a different content type for supply.
So you are suggesting I move all of the custom logic to the applciation.yaml right? Will do that.

xin-hedera: That is NOT my suggestion. What I said is move the default to where it’s supposed to be. And leave whatever customization in individual handlers…

hedera-mirror-rest/__tests__/middleware/responseHandler.test.js
```diff
@@ -59,29 +60,25 @@ describe('Response middleware', () => {
...
   });
 
   test('Default headers', async () => {
     mockRequest.route.path = '/api/v1/accounts';
     await responseHandler(mockRequest, mockResponse, null);
     expect(mockResponse.send).toBeCalledWith(JSONStringify(responseData));
-    expect(mockResponse.set).toHaveBeenNthCalledWith(1, headers.default);
-    expect(mockResponse.set).toHaveBeenNthCalledWith(2, headers.path[mockRequest.route.path]);
-    expect(mockResponse.set).toHaveBeenNthCalledWith(3, 'Content-Type', 'application/json; charset=utf-8');
+    expect(mockResponse.set).toHaveBeenCalledWith({'Content-type': 'application/json; charset=utf-8'},headers.path[mockRequest.route.path]);
```
mgoelswirlds, line 71: I have spent a lot of time trying to fix these 2 tests with this new approach. But have not been successful.
Would like to get the other parta approved and continue working on these on the side.

hedera-mirror-rest/transactions.js
```diff
@@ -766,17 +773,51 @@ const extractSqlFromTransactionsByIdOrHashRequest = async (transactionIdOrHash,
...
+  return {
+    query: getTransactionQuery(mainConditions.join(' and '), commonConditions.join(' and ')),
+    params,
+    scheduledParamExists: scheduledParamExists,
+    isTransactionHash: isTransactionHash,
+  };
 };
 
+function getTransactionsByIdOrHashCacheControlHeader(transactionsRows, isTransactionHash, scheduled) {
+  if (scheduled || isTransactionHash) {
```
steven-sheehy, line 785: As I mentioned [previously](https://github.com/hashgraph/hedera-mirror-node/pull/9327#discussion_r1760252662).
```suggestion
function getTransactionsByIdOrHashCacheControlHeader(transactionsRows) {
  if (transactionsRows.length < 2) {
```

hedera-mirror-rest/config/application.yml
```diff
@@ -87,7 +87,8 @@ hedera:
       response:
         compression: true
         headers:
-          default: { "cache-control": "public, max-age=1" }
+          default: { "cache-control": "public, max-age=1",
+          "Content-type": "application/json; charset=utf-8"}
```
steven-sheehy, line 91: Use consistent casing.
```suggestion
          "content-type": "application/json; charset=utf-8"}
```

hedera-mirror-rest/transactions.js
```diff
@@ -766,17 +773,51 @@ const extractSqlFromTransactionsByIdOrHashRequest = async (transactionIdOrHash,
...
+    return {}; // no override
+  }
+
+  let successScheduleCreateTimestamp;
+
+  for (const transaction of transactionsRows) {
+
+    if (transaction.type === scheduleCreateProtoId) {
+      if (TransactionResult.isSuccessful(transaction.result)) {
```
steven-sheehy, line 794: We can combine these two if statements since it's not like a schedule create can have scheduled=true

hedera-mirror-rest/transactions.js
```diff
@@ -766,17 +773,51 @@ const extractSqlFromTransactionsByIdOrHashRequest = async (transactionIdOrHash,
...
+        successScheduleCreateTimestamp = transaction.consensus_timestamp;
+      }
+    } else if (transaction.scheduled) {
+      return {};
+    }
+  }
+
+  if (successScheduleCreateTimestamp !== undefined
+      && (utils.nowInNs() - successScheduleCreateTimestamp) < maxTransactionConsensusTimestampRangeNs ) {
```
steven-sheehy, line 803: nit: Readability

```suggestion
  const elapsed = utils.nowInNs() - successScheduleCreateTimestamp;
  if (successScheduleCreateTimestamp && elapsed < maxTransactionConsensusTimestampRangeNs ) {
```

hedera-mirror-rest/transactions.js
```diff
@@ -698,8 +701,10 @@ const extractSqlFromTransactionsByIdOrHashRequest = async (transactionIdOrHash,
   const mainConditions = [];
   const commonConditions = [];
   const params = [];
+  const isTransactionHash = isValidTransactionHash(transactionIdOrHash);
+  let scheduledParamExists = false;
```
steven-sheehy, line 705: Think updates to this method can be undone with my other comment.

steven-sheehy (changes requested):

hedera-mirror-rest/transactions.js
mgoelswirlds: ok.

hedera-mirror-rest/config/application.yml
mgoelswirlds: Actually other places in the code so you `"Content-type"` but will change it here.

hedera-mirror-rest/transactions.js
xin-hedera: not entirely true. it's possible to have `schedule=false` query param and return multiple transactions for the same transaction id. Then we can have a successful schedulecreate in the response with all others failed, ideally we should use the longer max-age because the existence of the `schedule=?` query param.

However given that such scenario happens rarely, and using a shorter max-age doesn't hurt function, agree we can go with the simpler approach.

steven-sheehy: I missed that this scheduled variable represents the query param exists and not that it exists and is set to true, hence my comment. We can keep the scheduled check but omit isTransactionHash and just use list size for that.

hedera-mirror-rest/transactions.js
```diff
@@ -777,17 +784,51 @@ const extractSqlFromTransactionsByIdOrHashRequest = async (transactionIdOrHash,
...
-  return {query: getTransactionQuery(mainConditions.join(' and '), commonConditions.join(' and ')), params};
+  return {
+    query: getTransactionQuery(mainConditions.join(' and '), commonConditions.join(' and ')),
+    params,
+    scheduledParamExists: scheduledParamExists,
+    isTransactionHash: isTransactionHash,
+  };
 };
 
+function getTransactionsByIdOrHashCacheControlHeader(transactionsRows, isTransactionHash, scheduled) {
```
xin-hedera, line 795: coding style consistency, please use arrow function

```suggestion
const getTransactionsByIdOrHashCacheControlHeader = (transactionsRows, isTransactionHash, scheduled) =>  {
```

hedera-mirror-rest/transactions.js
```diff
@@ -777,17 +784,51 @@ const extractSqlFromTransactionsByIdOrHashRequest = async (transactionIdOrHash,
...
+
 /**
  * Handler function for /transactions/:transactionIdOrHash API.
  * @param {Request} req HTTP request object
  * @return {Promise<None>}
  */
 const getTransactionsByIdOrHash = async (req, res) => {
   const filters = utils.buildAndValidateFilters(req.query, acceptedSingleTransactionParameters);
-  const {query, params} = await extractSqlFromTransactionsByIdOrHashRequest(req.params.transactionIdOrHash, filters);
+  const {query, params, scheduled, isTransactionHash} = await extractSqlFromTransactionsByIdOrHashRequest(
```
xin-hedera, line 828: obviously missing test coverage,  the returned object no longer has `scheduled` key, the extracted value for `scheduled` will be always `undefined`

xin-hedera (changes requested):

hedera-mirror-rest/middleware/responseHandler.js
```diff
@@ -37,11 +43,16 @@ const responseHandler = async (req, res, next) => {
...
   } else {
     const path = res.locals[requestPathLabel] ?? req.route.path;
-    res.set(headers.default);
-    res.set(headers.path[path]);
+    const mergedHeaders = {
+      ...headers.default,
+      ...(headers.path[path] ?? {}),
+      ...(res.locals[responseHeadersLabel] ?? {}),
+    };
+    res.set(mergedHeaders);
```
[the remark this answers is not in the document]
mgoelswirlds: I tried adding a unit test, but that would need to work with mocking the result and then there seemed no point to hardcode what set and get would do.
```
  test('should merge headers in a case insesitive manner', async () => {
    const path = '/api/v1/transactions/:transactionId';
    headers.path[path] = { "Cache-control": "public, max-age=1",
      "content-Type": "application/json; charset=utf-8" };
    mockResponse.locals[responseHeadersLabel] = { "cache-Control": "public, max-age=1",
      "Content-type": 'text/plain; charset=utf-8' };
    await responseHandler(mockRequest, mockResponse, null);
    expect(mockResponse.getHeader('content-type')).toHaveLastReturnedWith(mockResponse.locals[responseHeadersLabel]['Content-type']);
  });
```
The network/supply spec tests do test the case insensitiveness to some extend atleast for` headers.default` and `res.locals[responseHeadersLabel] `
If that is acceptable , I can work on a future PR to add a more comprehensive test for this .

hedera-mirror-rest/model/transactionResult.js
```diff
@@ -363,8 +363,13 @@ const getSuccessProtoIds = () => {
   ];
 };
 
+const isSuccessful = (result) => {
+  return getSuccessProtoIds().includes(result);
```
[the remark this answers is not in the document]
mgoelswirlds: I have added a constant which calls `getSuccessProtoIds()` once.

hedera-mirror-rest/transactions.js
```diff
@@ -777,17 +786,51 @@ const extractSqlFromTransactionsByIdOrHashRequest = async (transactionIdOrHash,
...
-  return {query: getTransactionQuery(mainConditions.join(' and '), commonConditions.join(' and ')), params};
+  return {
+    query: getTransactionQuery(mainConditions.join(' and '), commonConditions.join(' and ')),
+    params,
+    scheduledParamExists: scheduledParamExists,
+  };
 };
 
+const getTransactionsByIdOrHashCacheControlHeader = (transactionsRows, scheduledParamExists) => {
+  if (scheduledParamExists || transactionsRows.length < 2) {
```
xin-hedera, line 797: Well, we have to remove `transactionRows.length < 2` since this is exactly the most common scenario that we don't want to use the longer max-age for

- query by just transaction id, no `scheduled=?` flag
- the response is a single successful schedulecreate transaction, the scheduled transaction has yet to reach consensus

hedera-mirror-rest/transactions.js
```diff
@@ -39,6 +39,8 @@ import {
 
 import {AssessedCustomFeeViewModel, NftTransferViewModel} from './viewmodel';
 
+const _SUCCESS_PROTO_IDS = TransactionResult.getSuccessProtoIds();
```
xin-hedera, line 42: why prefix it with `_`? we never do that per our coding style.

xin-hedera (changes requested):

hedera-mirror-rest/transactions.js
xin-hedera: a follow-up comment is, the test cases are either wrong, or there's missing coverage

xin-hedera: I don't know what stops you from creating a spec based integration test.

we have all building blocks. We can for sure set up data with just one successful schedule create transaction, or one successful schedule create transaction + many other failed transactions with the same tx id, we can use `fakeTime` feature in a spec test to mimic the scenario that the time elapsed from the schedule tx's consensus time to the `fakeTime` is less than `maxTransactionConsensusTimestampRangeNs`, then we should expect the shorter max-age.

When we don't use `fakeTime`, and set the schedulecreate consensus timestamp sufficiently old in the past, we should expect the standard longer max-age.

mgoelswirlds: I have added them with fake-time and the ones without were already added. Thanks.

hedera-mirror-rest/controllers/networkController.js
```diff
@@ -42,7 +41,7 @@ const networkNodesDefaultSize = 10;
 const networkNodesMaxSize = 25;
 
 class NetworkController extends BaseController {
-  static contentTypeTextPlain = 'text/plain; charset=utf-8';
+  static contentTypeTextPlain = {'Content-type' : 'text/plain; charset=utf-8'};
```
xin-hedera, line 44: shouldn't this be all lower case and even better use a constant for the key instead?

```suggestion
  static contentTypeTextPlain = {'content-type' : 'text/plain; charset=utf-8'};
```

xin-hedera (changes requested):

hedera-mirror-rest/middleware/responseHandler.js
```diff
@@ -15,15 +15,15 @@
...
+import {requestPathLabel, requestStartTime, responseDataLabel, responseHeadersLabel} from '../constants';
 import {NotFoundError} from '../errors';
 import {JSONStringify} from '../utils';
 
 const {
   response: {headers},
 } = config;
 
-const CONTENT_TYPE_HEADER = 'Content-Type';
+const CONTENT_TYPE_HEADER = 'content-type';
```
xin-hedera, line 26: why have the same constant defined in two places?

xin-hedera (approved):

xin-hedera:
I approved it since I'm off on Monday so don't unnecessarily block the PR. There is still one minor change I asked for.

steven-sheehy (dismissed):
LGTM

jascks (approved):
LGTM

hedera-mirror-rest/middleware/responseHandler.js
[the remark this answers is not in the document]
mgoelswirlds: I have updated this so marking this conversation resolved.

hedera-mirror-rest/middleware/responseHandler.js
mgoelswirlds: Removed this and hence resolving it to unblock.

jnels124 (approved):
LGTM

PR #9334, opened 2024-09-11
Disable drift detection in performance

steven-sheehy:
**Description**:

* Bump performance to v0.113.1
* Disable drift detection in performance so reset works properly
* Remove grpc resource overrides that are now in 0.113
* Reduce initial disk size and pv size for citus

**Related issue(s)**:

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/clusters/preprod/common/helmrelease.yaml
+++ b/clusters/preprod/common/helmrelease.yaml
@@ -92,3 +92,5 @@ spec:
           cloud.google.com/neg: '{ "exposed_ports":{ "443":{"name": "mirrornode-preprod-new"} } }'
     zfs:
       enabled: true
+      worker:
+        initialDiskSize: 1024GB
--- a/clusters/preprod/common/helmrelease.yaml
+++ b/clusters/preprod/common/helmrelease.yaml
@@ -10,7 +10,7 @@ spec:
       sourceRef:
         kind: HelmRepository
         name: hedera-mirror-node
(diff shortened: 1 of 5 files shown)


xin-hedera:
We may want to make the following changes now

- remove the additional 2 citus-coordinator nodes, and the additional3 citus-worker nodes
- rebuild the citus nodes with smaller initial disks, especially worker node disk, 512GB should be enough v.s. 3200GB as the default

clusters/preprod/common/helmrelease.yaml
```diff
@@ -92,3 +92,7 @@ spec:
           cloud.google.com/neg: '{ "exposed_ports":{ "443":{"name": "mirrornode-preprod-new"} } }'
     zfs:
       enabled: true
+      coordinator:
+        initialDiskSize: 100GB
+      worker:
+        initialDiskSize: 1024GB
```
xin-hedera, line 98: this is more than what we need size wise.

however we have to bloat it up to get the IOPS we need for 10K tps tests in performance network

for balanced pd, the equation is

`iops = base + iops per GB * size in GB`

base is 3000, iops per GB for balanced PD is 6. We need somewhere around 9000 IOPS, so ended up to be 1024GB.

xin-hedera (approved):
LGTM

jnels124 (approved):
LGTM

Issue #9344, opened 2024-09-13
Docker build of web3 failing


kselveliev:
### Description

After adding the gradle tasks for compiling historical contracts with older version of solidity - 0.8.7. After assemble task the build/libs directory is polluted with a new 0.8.7-web3j.jar and then running dockerBuild command fails encountering this 0.8.7 jar. 
We need a new task that removes this 0.8.7 jar.
Currently in order to run dockerBuild of web3 you need to manually go to build/libs and delete the 0.8.7.jar and then dockerBuild will succeed.

### Steps to reproduce

in web3 module
do assemble
do dockerBuild and observe the issue.

### Version

114

### Operating system

macOS

PR #9348, opened 2024-09-13
Delete redundant web3j jars

bilyana-gospodinova:
**Description**:
Recently we added some gradle tasks for compiling historical contracts with older version of solidity - 0.8.7. After assemble task the build/libs directory is polluted with a couple of new 0.8.7-web3j*.jar jars and then running dockerBuild command fails encountering these jars.

This PR adds a new gradle task that deletes the problematic jars.

Fixes https://github.com/hashgraph/hedera-mirror-node/issues/9344

**Checklist**

### DIFF
--- a/hedera-mirror-web3/build.gradle.kts
+++ b/hedera-mirror-web3/build.gradle.kts
@@ -207,6 +207,17 @@ tasks.register<Copy>("moveAndCleanTestHistoricalFiles") {
     dependsOn(tasks.named("generateTestHistoricalContractWrappers"))
 }
 
+val deleteWeb3jJars =
+    tasks.register<Delete>("deleteWeb3jJars") {
+        description = "Deletes the web3 jars after they are no longer needed"
+        group = "historical"
+
+        val web3Jars = fileTree("build/libs") { include("web3*-${historicalSolidityVersion}*.jar") }
+        delete(web3Jars)
+
+        mustRunAfter(processTestHistoricalResources)
+    }
+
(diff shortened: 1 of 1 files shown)


kselveliev (approved):

hedera-mirror-web3/build.gradle.kts
```diff
@@ -207,6 +207,17 @@ tasks.register<Copy>("moveAndCleanTestHistoricalFiles") {
...
+    tasks.register<Delete>("deleteWeb3jJars") {
+        description = "Deletes the web3 jars after they are no longer needed"
+        group = "historical"
+
+        val web3Jars = fileTree("build/libs") { include("web3*-${historicalSolidityVersion}*.jar") }
+        delete(web3Jars)
+
+        mustRunAfter(processTestHistoricalResources)
+    }
+
```
steven-sheehy, line 220: This is the incorrect fix. It's not like there's two jars being produced, it's just one jar with the wrong version. The contents are identical to the one with the correct version in dockerBuild. 

The correct fix is to remove this line in `resolveSolidityHistorical` which accidentally overrides the project.version. It's not a solidity version property.

```java
version = historicalSolidityVersion
```

steven-sheehy (changes requested):

hedera-mirror-web3/build.gradle.kts
bilyana-gospodinova: Oh, nice catch! Fixed.

steven-sheehy (approved):

PR #9399, opened 2024-09-18
Add hedera app dependency to hedera-mirror-web3

### LINKED ISSUE #9249: Add hedera.app dependency into `hedera-mirror-node`
### Problem

In order to integrate the reusable services into `hedera-mirror-web3`, we should first add `hedera.app` as a dependency there.

### Solution

Add a new dependency to `hedera-mirror-web3` - `api("com.hedera.hashgraph:app:0.53.5")` and adapt the codebase, so that the project compiles successfully.

### Alternatives

_No response_

IvanKavaldzhiev:
**Description**:
Add `com.hedera.hashgraph:app` as a dependency in web3, so that the work for integrating reusable services is unblocked.

In addition, this PR stabilizes some inconsistent failing integration tests.

Fixes #9249 

**Notes for reviewer**:

**Checklist**


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
```diff
@@ -199,7 +199,7 @@ public boolean isCreate2Enabled() {
 
     @Override
     public boolean allowCallsToNonContractAccounts() {
-        return getSemanticEvmVersion().compareTo(EVM_VERSION_0_46) >= 0;
+        return getSemanticEvmVersion().minor() > 46;
```
steven-sheehy, line 202: This is a breaking change whenever a major version bump occurs. Please use https://github.com/hashgraph/hedera-services/blob/develop/platform-sdk/swirlds-state-api/src/main/java/com/swirlds/state/spi/HapiUtils.java#L38

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/convert/SemanticVersionConvertor.java
```diff
@@ -26,6 +26,7 @@
 public class SemanticVersionConvertor implements Converter<String, SemanticVersion> {
     @Override
     public SemanticVersion convert(String source) {
-        return SemanticVersion.parse(source);
+        final var versions = source.split("\\.");
+        return new SemanticVersion(0, Integer.parseInt(versions[1]), 0, "", "");
```
steven-sheehy, line 30: This breaks with non-zero major. Can we use https://github.com/hashgraph/hedera-services/blob/b2ea7600799d5df3921151e38be498a14d3d9af9/hedera-node/hedera-config/src/main/java/com/hedera/node/config/converter/SemanticVersionConverter.java instead of this class?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
```diff
@@ -209,7 +209,7 @@ public Set<Address> grandfatherContracts() {
 
     @Override
     public boolean callsToNonExistingEntitiesEnabled(Address target) {
-        return !(getSemanticEvmVersion().compareTo(EVM_VERSION_0_46) < 0
+        return !(getSemanticEvmVersion().minor() < 46
```
steven-sheehy, line 212: Same

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/convert/SemanticVersionConvertor.java
```diff
@@ -16,16 +16,47 @@
...
-import com.swirlds.common.utility.SemanticVersion;
+import com.hedera.hapi.node.base.SemanticVersion;
 import jakarta.inject.Named;
+import java.util.regex.Pattern;
 import org.springframework.boot.context.properties.ConfigurationPropertiesBinding;
 import org.springframework.core.convert.converter.Converter;
 
 @Named
 @ConfigurationPropertiesBinding
 public class SemanticVersionConvertor implements Converter<String, SemanticVersion> {
```
steven-sheehy, line 27: Can we not directly use the upstream [SemanticVersionConverter.java](https://github.com/hashgraph/hedera-services/blob/b2ea7600799d5df3921151e38be498a14d3d9af9/hedera-node/hedera-config/src/main/java/com/hedera/node/config/converter/SemanticVersionConverter.java)? Is it not available in the library or just that we need to use it with Spring Converter interface? If the latter can just use this class to delegate to upstream.

IvanKavaldzhiev: We have access directly to the upstream `SemanticVersionConverter`, however it implements `ConfigConverter<SemanticVersion>`. I couldn't configure Spring to automatically use the upstream class without manually providing implementation for `Converter<String, SemanticVersion>`.

steven-sheehy: That's why I said delegate to upstream. 

```java
public class SemanticVersionConvertor implements Converter<String, SemanticVersion> {

    private final com.hedera.node.config.converter.SemanticVersionConverter delegate = new com.hedera.node.config.converter.SemanticVersionConverter();

    @Override
    public SemanticVersion convert(String source) {
        return delegate.convert(source);
    }
}

IvanKavaldzhiev: Oh, I thought you were referring to a way of completely removing the SemanticVersionConvertor class implementation. I will adapt this delegate approach.

steven-sheehy (approved):
LGTM

Issue #9378, opened 2024-09-16
HIP-904 Add NFTs to Airdrop REST API response


edwin-greene:
### Problem

The Outstanding and Pending airdrops endpoints currently only return fungible tokens.

### Solution

-Add nfts to the response and paginate based on sender/receiver, token.id and serialnumber.

PR #9424, opened 2024-09-20
Update jooq repository to improve reusability

### LINKED ISSUE #9379: Update Jooq Repository ConditionalFieldBounds to improve reusability
### Problem

Make the ConditionalFieldBounds more reusable to support future Rest-Java endpoints.

### Solution

Combine Bound and FieldBound by moving field into Bound, and update ConditionalFieldBounds to support n-numbered fields. Plus maybe we can put an interface on the TokenAirdropRequest to use as this.

See comments from the [Outstanding Token Airdrops PR](https://github.com/hashgraph/hedera-mirror-node/pull/9286)

### Alternatives

_No response_

edwin-greene:
**Description**:
Updates the Bound class to include fields. 
Simplifies the JooqRepository to only accept Bounds as parameters.
Updates the JooqRepository to support N-numbered conditions. 

**Related issue(s)**:

Fixes #9379 

**Notes for reviewer**:
This should not be merged until the work for pagination for airdrop nfts is completed in #9378 

**Checklist**


hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/Bound.java
```diff
@@ -16,33 +16,39 @@
...
+    @Getter
+    @Setter
+    private RangeParameter<Long> upper;
 
     private final String parameterName;
 
     private final EnumMap<RangeOperator, Integer> cardinality = new EnumMap<>(RangeOperator.class);
 
-    public Bound(List<EntityIdRangeParameter> params, boolean primarySortField, String parameterName) {
+    public Bound(RangeParameter<Long>[] params, boolean primarySortField, String parameterName, Field<Long> field) {
```
edwin-greene, line 47: Bounds can now be `NumberRangeParameter` bounds (because of the serial number bound), instead of just `EntityIdRangeParameter`, so I updated this from a list to an array to accept the interface of both `RangeParameter`.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/dto/TokenAirdropRequest.java
```diff
@@ -64,4 +67,11 @@ public enum AirdropRequestType {
         // and Sender Id for Pending Airdrops
         private final Field<Long> primaryField;
     }
+
+    public List<Bound> getBounds() {
+        var tertiaryBound = serialNumbers == null ? Bound.EMPTY : serialNumbers;
+        var secondaryBound = tokenIds == null ? tertiaryBound : tokenIds;
+        var primaryBound = entityIds == null ? secondaryBound : entityIds;
+        return List.of(primaryBound, secondaryBound, tertiaryBound);
+    }
```
edwin-greene, line 76: This class already contained the specifics concerning the ordering of the jooq Fields in `AirdropRequestType`, so it seemed to fit to place the specifics for the ordering of the bounds here. But we could place this in one of the other layers as well.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/Bound.java
steven-sheehy: How does list vs array help? Can't you do `List<RangeParameter<Long>>`?

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/repository/JooqRepository.java
```diff
@@ -22,112 +22,125 @@
...
+        }
+
+        if (!primary.isEmpty() && primary.hasEqualBounds()) {
             // If the primary param has a range with a single value, rewrite it to EQ
-            primaryLower = new EntityIdRangeParameter(EQ, EntityId.of(primaryBound.adjustLowerBound()));
-            primaryUpper = null;
+            var primaryLower = new NumberRangeParameter(
+                    EQ, EntityId.of(primary.adjustLowerBound()).getId());
+            primary.setLower(primaryLower);
+            primary.setUpper(null);
```
steven-sheehy, line 73: Can't we just do this in the constructor of the Bound? Or possibly in the getters for lower/upper? This would keep the class more immutable. 

We also have to take into consideration 2nd of 3 params can probably also have this behavior.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/repository/JooqRepository.java
```diff
@@ -22,112 +22,125 @@
...
-        var secondaryUpper = secondaryBound.getUpper();
-        if (!secondaryBound.isEmpty() && (secondaryLower != null && secondaryLower.operator() == EQ)) {
-            // If the secondary param operator is EQ, set the secondary upper bound to the same
-            secondaryUpper = secondaryLower;
+        for (var bound : secondaryBounds) {
+            if (!bound.isEmpty()) {
+                var secondaryLower = bound.getLower();
+                if (secondaryLower != null && secondaryLower.operator() == EQ) {
+                    // If the secondary param operator is EQ, set the secondary upper bound to the same
+                    bound.setUpper(secondaryLower);
```
steven-sheehy, line 81: Can we move this logic into bound?

steven-sheehy (changes requested):

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/Bound.java
edwin-greene: I tried a few different ways but couldn't get it to compile as such. If the signature of Bounds is changed to `List<RangedParameter>` or `List<RangedParameter<Long>>` or `List<RangedParameter<?>>`  then
passing a `List<EntityIdRangeParameter>` errors with: 

```
Required type: List<RangeParameter>
Provided: List<EntityIdRangeParameter>>```

File: hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/RangeParameter.java [hunk 5]
```diff
@@ -30,4 +30,12 @@ default boolean hasLowerBound() {
     default boolean hasUpperBound() {
         return operator() == RangeOperator.LT || operator() == RangeOperator.LTE;
     }
+
+    default boolean hasLTorGT() {
+        return operator() == RangeOperator.LT || operator() == RangeOperator.GT;
```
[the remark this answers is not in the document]
edwin-greene: Thanks I have updated it.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/EntityIdRangeParameter.java
```diff
@@ -42,7 +46,7 @@ public static EntityIdRangeParameter valueOf(String entityIdRangeParam) {
         };
     }
 
-    private static EntityId getEntityId(String entityId) {
+    private static Long getEntityId(String entityId) {
```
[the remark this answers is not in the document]
edwin-greene: Good catch, I have reverted that part.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/controller/TokenAirdropsController.java
```diff
@@ -71,37 +73,41 @@ TokenAirdropsResponse getOutstandingAirdrops(
...
             AirdropRequestType type,
             String primarySortField) {
         var request = TokenAirdropRequest.builder()
                 .accountId(id)
                 .entityIds(entityIdsBound)
                 .limit(limit)
                 .order(order)
-                .tokenIds(new Bound(tokenIds, false, TOKEN_ID))
+                .serialNumbers(new Bound(serialNumbers, false, SERIAL_NUMBER, TOKEN_AIRDROP.SERIAL_NUMBER))
+                .tokenIds(new Bound(tokenIds, false, TOKEN_ID, TOKEN_AIRDROP.TOKEN_ID))
```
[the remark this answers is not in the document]
edwin-greene: If it is set to true, I found that certain next links became invalid, such as: 
```
?limit=1&receiver.id=lt:0.0.4&receiver.id=gte:0.0.3&token.id=lte:0.0.5&token.id=gt:0.0.5
```

That link is from the `allParameters()` test in `TokenAirdropsControllerTest`

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/dto/NftAllowanceRequest.java
```diff
@@ -40,4 +41,10 @@ public class NftAllowanceRequest {
     private Bound ownerOrSpenderIds;
 
     private Bound tokenIds;
+
+    public List<Bound> getBounds() {
+        var secondaryBound = tokenIds == null ? Bound.EMPTY : tokenIds;
```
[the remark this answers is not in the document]
edwin-greene: Sounds good, I have updated it.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/dto/NftAllowanceRequest.java
```diff
@@ -40,4 +41,10 @@ public class NftAllowanceRequest {
...
+        var primaryBound = ownerOrSpenderIds == null ? secondaryBound : ownerOrSpenderIds;
```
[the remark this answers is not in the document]
edwin-greene: Thanks. I have updated this section, the variable matching the conditional is now the first item.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/dto/NftAllowanceRequest.java
```diff
@@ -40,4 +41,10 @@ public class NftAllowanceRequest {
...
+        return List.of(primaryBound, secondaryBound);
```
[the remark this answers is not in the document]
edwin-greene: Sounds good, I have reworked this for both request types.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/dto/TokenAirdropRequest.java
```diff
@@ -64,4 +67,11 @@ public enum AirdropRequestType {
         // and Sender Id for Pending Airdrops
         private final Field<Long> primaryField;
     }
+
+    public List<Bound> getBounds() {
+        var tertiaryBound = serialNumbers == null ? Bound.EMPTY : serialNumbers;
+        var secondaryBound = tokenIds == null ? tertiaryBound : tokenIds;
+        var primaryBound = entityIds == null ? secondaryBound : entityIds;
+        return List.of(primaryBound, secondaryBound, tertiaryBound);
```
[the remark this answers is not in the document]
edwin-greene: Got it, this has been reworked.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/RangeParameter.java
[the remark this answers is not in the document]
steven-sheehy: You updated the name, but didn't invert the logic. So now the logic is wrong.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/RangeOperator.java
```diff
@@ -36,6 +36,10 @@ public enum RangeOperator {
     private final String operator;
     private final BiFunction<Field, Object, Condition> function;
 
+    public static boolean isInclusive(RangeOperator operator) {
+        return operator == RangeOperator.LT || operator == RangeOperator.GT;
```
steven-sheehy, line 40: It doesn't need to be static and the logic is inverted for being inclusive.

```suggestion
    public boolean isInclusive() {
        return this != RangeOperator.LT && operator != RangeOperator.GT;
```

edwin-greene: Thanks I have fixed this.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/common/RangeParameter.java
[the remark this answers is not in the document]
edwin-greene: I see! Thanks I have fixed it.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/dto/NftAllowanceRequest.java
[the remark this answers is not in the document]
steven-sheehy: Alternatively, it'd be simpler to rationalize if for 3 column pagination we always had 3 items in the list. And for first item it would always be ownerOrSpender, second always tokenId and third always serial. Then the framework would be smart enough to handle empty items and we wouldn't need conditional logic in this getBounds()

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/repository/JooqRepository.java
```diff
@@ -22,112 +22,126 @@
...
+        var condition = noCondition();
+        for (var bound : bounds) {
+            var param = upper ? bound.getUpper() : bound.getLower();
+            condition = condition.and(getCondition(param, bound.getField()));
+        }
+
+        return condition;
+    }
+
+    default Condition getBoundCondition(List<Bound> bounds) {
```
xin-hedera, line 57: I don't think the changes here can solve the N pagination column problem.

It's difficult to explain the issues thoroughly in symbolic representation or pseudo code. However, it should be quite straightforward by looking at a complicated request and the generated SQL.

Given the following request,

```bash
$ curl 'localhost:8084/api/v1/accounts/0.0.1500/airdrops/outstanding?receiver.id=gte:0.0.2000&receiver.id=lte:0.0.3000&token.id=gte:0.0.4000&token.id=lte:0.0.5000&serialnumber=gte:5&serialnumber=lte:100&limit=50'
```

we get the generated SQL and params

```
select "token_airdrop"."amount", "token_airdrop"."receiver_account_id", "token_airdrop"."sender_account_id", "token_airdrop"."serial_number", "token_airdrop"."state", "token_airdrop"."timestamp_range", "token_airdrop"."token_id"
from "token_airdrop"
where ("token_airdrop"."sender_account_id" = $1 and
  (
   ("token_airdrop"."receiver_account_id" = $2 and "token_airdrop"."token_id" >= $3 and "token_airdrop"."serial_number" >= $4) or
   ("token_airdrop"."receiver_account_id" >= $5 and "token_airdrop"."receiver_account_id" <= $6) or 
   ("token_airdrop"."receiver_account_id" = $7 and "token_airdrop"."token_id" <= $8 and "token_airdrop"."serial_number" <= $9)
  ) and
  "token_airdrop"."state" = cast($10 as "airdrop_state"))
order by "token_airdrop"."receiver_account_id" asc, "token_airdrop"."token_id" asc, "token_airdrop"."serial_number" asc
fetch next $11 rows only

parameters: $1 = '1500', $2 = '2000', $3 = '4000', $4 = '5', $5 = '2001', $6 = '2999', $7 = '3000', $8 = '5000', $9 = '100', $10 = 'PENDING', $11 = '50'
```

There are problems in these two conditions

- `("token_airdrop"."receiver_account_id" = $2 and "token_airdrop"."token_id" >= $3 and "token_airdrop"."serial_number" >= $4)`
-  `("token_airdrop"."receiver_account_id" = $7 and "token_airdrop"."token_id" <= $8 and "token_airdrop"."serial_number" <= $9)`

A concrete example of a token airdrop which should get included (let's assume there is no `limit`), however due to the bug in the first of the two conditions above, will never show up is as follows:

`(sender=0.0.1500, receiver=0.0.2000, token=0.0.4001,serialnumber=1)`

I agree the two-column pagination support itself is quite complicated and corner case optimizations make it even harder to understand and maintain, so expanding it to N-column is a much bigger challenge.

We should be able to generalize the problem and solve it recursively with the algorithm in the following psuedo code


```
function getConditions(Bound[] bounds) {
  primaryBound = bounds[0]
  if bounds.length == 1
    return condition created from primaryBound

  remaining = bounds[1...N]
  lowerCondition = ((primary column = inclusive lower bound value) or no condition if primaryBound doesn't have lower bound) and
    getConditions(lower bounds from remaining)
  middleCondition = getMiddleCondition(primaryBound)
  upperCondition = ((primary column = inclusive upper bound value) or no condition if primaryBound doesn't have upper bound) and
    getConditions(upper bounds from remaining)

  return lowerCondition or middleCondition or upperCondition
}

function getMiddleCondition(Bound bound) {
	lowerCondition = column >= lower value + 1 or no condition if there's no lower bound
	upperCondition = column <= upper value - 1 or no condition if there's no upper bound
	return lowerCondition and upperCondition
}
```

xin-hedera (changes requested):

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/repository/JooqRepository.java
edwin-greene: I have updated this to be a new adjustment method in Bound.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/repository/JooqRepository.java
edwin-greene: Got it, this has been moved to Bound

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/controller/TokenAirdropsController.java
```diff
@@ -71,37 +73,41 @@ TokenAirdropsResponse getOutstandingAirdrops(
...
             @RequestParam(defaultValue = "asc") Sort.Direction order,
-            @RequestParam(name = SENDER_ID, required = false) @Size(max = 2) List<EntityIdRangeParameter> senderIds,
-            @RequestParam(name = TOKEN_ID, required = false) @Size(max = 2) List<EntityIdRangeParameter> tokenIds) {
-        var entityIdsBound = new Bound(senderIds, true, ACCOUNT_ID);
-        return processRequest(id, entityIdsBound, limit, order, tokenIds, PENDING, SENDER_ID);
+            @RequestParam(name = SENDER_ID, required = false) @Size(max = 2) EntityIdRangeParameter[] senderIds,
+            @RequestParam(name = SERIAL_NUMBER, required = false) @Size(max = 2) NumberRangeParameter[] serialNumbers,
+            @RequestParam(name = TOKEN_ID, required = false) @Size(max = 2) EntityIdRangeParameter[] tokenIds) {
+        var entityIdsBound = new Bound(senderIds, true, ACCOUNT_ID, TOKEN_AIRDROP.SENDER_ACCOUNT_ID);
+        return processRequest(id, entityIdsBound, limit, order, serialNumbers, tokenIds, PENDING, SENDER_ID);
```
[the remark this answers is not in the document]
edwin-greene: Thanks I have updated both of those.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/repository/JooqRepository.java
edwin-greene: Got it, I have reworked this section to implement a recursive solution. I have added this example as a test case in `TokenAirdropsControllerTest.allParametersNft`

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/repository/JooqRepository.java
```diff
@@ -21,113 +21,133 @@
...
+            }
         }
 
-        return getCondition(primaryField, EQ, value).and(getCondition(secondaryField, secondaryParam));
+        // Lower conditions need to be discovered before upper conditions because the methods involved update the
+        // primary bound
+        var lowerCondition = getOuterCondition(primary, secondaryBounds, false, lowerProcessed);
+        var middleCondition = getMiddleCondition(primary, secondaryBounds);
+        var upperCondition = getOuterCondition(primary, secondaryBounds, true, upperProcessed);
+
```
edwin-greene, line 82: Todo, I still need to add Javadocs as to how these conditions are generated with bounds and sql examples.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/controller/TokenAirdropsController.java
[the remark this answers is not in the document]
xin-hedera: Don't think it's addressed per my change request.

What I meant is, for example, for pending request,  we shouldn't pass both type `PENDING` and `SENDER_ID`, since if it's PENDING, it has to be `SENDER_ID`, and the enum `AirdropRequestType` already has the info encapsulated.

For `entityIdsBound`, we should remove code duplication to moving it into function `processRequest`, we can do so because of the same reason, that for `PENDING`, it's always `SENDER_ACCOUNT_ID` as the primary sort column, and for `OUTSTANDING`, it's always `RECEIVER_ACCOUNT_ID` as the primary sort column.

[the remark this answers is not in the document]
xin-hedera: umm, my bad, was looking at the stale code...

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/repository/JooqRepository.java
```diff
@@ -21,113 +21,205 @@
...
+import java.util.Arrays;
+import java.util.List;
 import org.jooq.Condition;
 import org.jooq.Field;
 
 interface JooqRepository {
 
-    default Condition getCondition(Field<Long> field, EntityIdRangeParameter param) {
-        if (param == null || param == EntityIdRangeParameter.EMPTY) {
+    default Condition getCondition(RangeParameter<Long> param, Field<Long> field) {
```
xin-hedera, line 34: there are many new methods which should be private instead of public with default implementation.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/repository/JooqRepository.java
```diff
@@ -21,113 +21,205 @@
...
 
-        // If the secondary param operator is EQ, there should only have the middle condition
-        if (secondaryParam != null && secondaryParam.operator() == EQ) {
+        var primary = bounds.getFirst();
+        if (bounds.size() == 1) {
+            return getCondition(primary);
+        }
+
+        var secondaryBounds = bounds.subList(1, bounds.size());
+        if (primary.isEmpty() && secondaryBounds.isEmpty()) {
```
xin-hedera, line 92: this is dead code, since when the flow reaches here, `secondaryBounds.isEmpty()` is always `false`. There's probably no coverage either however needs to recheck after the condition is fixed

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/repository/JooqRepository.java
```diff
@@ -21,113 +21,205 @@
...
+     *  [Upper Outer Condition]
+     *  or ("receiver_account_id" = 3000 and
+     *       (
+     *           "token_id" < 5000 or
+     *           ("token_id" = 5000 and "serial_number" <= 100)
+     *       )
+     *  )
+     */
+    default Condition getBoundConditions(List<Bound> bounds, boolean lowerProcessed, boolean upperProcessed) {
+        if (bounds.isEmpty()) {
```
xin-hedera, line 82: no test coverage here.

xin-hedera (changes requested):

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/Bound.java
```diff
@@ -55,39 +66,84 @@ public Bound(List<EntityIdRangeParameter> params, boolean primarySortField, Stri
...
-        long adjustedLower = adjustLowerBound();
+        long adjustedLower = getAdjustedLowerRangeValue();
         long adjustedUpper = adjustUpperBound();
         if (primarySortField && adjustedLower > adjustedUpper) {
             throw new IllegalArgumentException("Invalid range provided for %s".formatted(parameterName));
         }
     }
 
+    // A bound with an upper or lower rangeParameter, but not both
+    public Bound(Bound bound, boolean isUpper) {
```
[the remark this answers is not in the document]
edwin-greene: Thanks I have changed this to those methods.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/Bound.java
```diff
@@ -55,39 +66,84 @@ public Bound(List<EntityIdRangeParameter> params, boolean primarySortField, Stri
...
         return upperBound;
     }
 
-    public long adjustLowerBound() {
+    public RangeParameter<Long> adjustLowerRange() {
+        if (!this.isEmpty() && this.hasEqualBounds()) {
+            // If the primary param has a range with a single value, rewrite it to EQ
+            this.setLower(new NumberRangeParameter(EQ, this.getAdjustedLowerRangeValue()));
+            this.setUpper(null);
+        }
```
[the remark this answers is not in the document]
edwin-greene: It looks like it cannot be done to secondary bounds before the primary bound is analyzed, there are Nft repository tests that fail if this is done in the constructor.  

I have updated it and removed the Setters.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/repository/JooqRepository.java
edwin-greene: Thanks I have added a test case for this.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/repository/JooqRepository.java
edwin-greene: Good catch, I have removed this.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/repository/JooqRepository.java
edwin-greene: Thanks I have updated them to private.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/Bound.java
```diff
@@ -16,33 +16,44 @@
...
 
     private final EnumMap<RangeOperator, Integer> cardinality = new EnumMap<>(RangeOperator.class);
 
-    public Bound(List<EntityIdRangeParameter> params, boolean primarySortField, String parameterName) {
+    public Bound(RangeParameter<Long>[] params, boolean primarySortField, String parameterName, Field<Long> field) {
+        this.field = field;
         this.parameterName = parameterName;
 
-        if (CollectionUtils.isEmpty(params)) {
+        if (params == null || params.length == 0) {
```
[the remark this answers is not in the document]
edwin-greene: Got it, I have updated this.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/Bound.java
```diff
@@ -55,39 +66,84 @@ public Bound(List<EntityIdRangeParameter> params, boolean primarySortField, Stri
...
         if (this.upper.operator() == RangeOperator.LT) {
             upperBound--;
         }
 
         return upperBound;
     }
 
-    public long adjustLowerBound() {
+    public RangeParameter<Long> adjustLowerRange() {
+        if (!this.isEmpty() && this.hasEqualBounds()) {
```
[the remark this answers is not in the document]
edwin-greene: Thanks I have fixed this.

hedera-mirror-rest-java/src/main/java/com/hedera/mirror/restjava/service/Bound.java
```diff
@@ -103,7 +143,17 @@ public boolean hasLowerAndUpper() {
...
+    }
+
+    // Returns a new bound with only a lower rangeParameter
+    public Bound toLower(Bound bound) {
+        return createBound(bound.getLower());
+    }
+
+    // Returns a new bound with only an upper rangeParameter
+    public Bound toUpper(Bound bound) {
+        return createBound(bound.getUpper());
```
steven-sheehy, line 156: This was not what I suggested. It doesn't make sense to pass in the same bound that you're invoking it on.

```suggestion
    public Bound toLower() {
        return createBound(getLower());
    }

    // Returns a new bound with only an upper rangeParameter
    public Bound toUpper() {
        return createBound(getUpper());
```

edwin-greene: Doh, thanks, much better. I have fixed it now.

steven-sheehy (approved):

xin-hedera (approved):

PR #9448, opened 2024-09-24
Increase RAM For Prometheus and PVC Size For Loki in Mainnet NA

jnels124:
* increase loki pvc to 400gb
* increase ram for prometheus to have requests and limits of 10752Mi


jascks (approved):
LGTM

steven-sheehy (approved):

Issue #9255, opened 2024-09-03
Add ContractBytecodeDatabaseAccessor for reading contract runtime bytecode


IvanKavaldzhiev:
### Problem

We need to add a new type of database accessor to read contract runtime bytecode, so that we have specific ReadableKVState instance for this type of data.

### Solution

Add ContractBytecodeDatabaseAccessor class that reads contract runtime bytecode data and converts the bytecode to PBJ type.

PR #9452, opened 2024-09-25
Implement `ContractBytecodeReadableKVState`

bilyana-gospodinova:
**Description**:
As part of the implementation for the reusable services we need to implement `ContractBytecodeReadableKVState`. It reads the contract's runtime bytecode and converts it to `Bytecode` PBJ type.

Fixes https://github.com/hashgraph/hedera-mirror-node/issues/9255

**Checklist**

### DIFF
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/ContractBytecodeReadableKVState.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/ContractBytecodeReadableKVState.java
@@ -21,6 +21,7 @@ import static com.hedera.services.utils.EntityIdUtils.entityIdFromContractId;
 
 import com.hedera.hapi.node.base.ContractID;
 import com.hedera.hapi.node.state.contract.Bytecode;
+import com.hedera.mirror.common.domain.entity.Entity;
 import com.hedera.mirror.common.domain.entity.EntityId;
 import com.hedera.mirror.common.util.DomainUtils;
 import com.hedera.mirror.web3.repository.ContractRepository;
@@ -79,8 +80,8 @@ public class ContractBytecodeReadableKVState extends ReadableKVStateBase<Contrac
                 return commonEntityAccessor
                         .getEntityByEvmAddressAndTimestamp(
                                 contractID.evmAddress().toByteArray(), Optional.empty())
-                        .get()
-                        .toEntityId();
+                        .map(Entity::toEntityId)
+                        .orElse(EntityId.EMPTY);
             }
         }
         return EntityId.EMPTY;
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/ContractBytecodeReadableKVState.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/ContractBytecodeReadableKVState.java
@@ -16,34 +16,31 @@
 
 package com.hedera.mirror.web3.state;
 
-import static com.hedera.mirror.web3.state.Utils.convertContractIDToEntityId;
+import static com.hedera.services.utils.EntityIdUtils.toEntityId;
 
 import com.hedera.hapi.node.base.ContractID;
 import com.hedera.hapi.node.state.contract.Bytecode;
 import com.hedera.mirror.web3.repository.ContractRepository;
 import com.hedera.pbj.runtime.io.buffer.Bytes;
 import com.swirlds.state.spi.ReadableKVStateBase;
+import jakarta.annotation.Nonnull;
+import jakarta.inject.Named;
 import java.util.Collections;
 import java.util.Iterator;
-import org.jetbrains.annotations.NotNull;
 
+@Named
 public class ContractBytecodeReadableKVState extends ReadableKVStateBase<ContractID, Bytecode> {
-
-    private static final String KEY = "BYTECODE";
 
     private final ContractRepository contractRepository;
 
     protected ContractBytecodeReadableKVState(final ContractRepository contractRepository) {
-        super(KEY);
+        super("BYTECODE");
         this.contractRepository = contractRepository;
     }
 
     @Override
-    protected Bytecode readFromDataSource(@NotNull ContractID contractID) {
-        final var entityId = convertContractIDToEntityId(contractID);
-        if (entityId == null) {
-            return null;
-        }
+    protected Bytecode readFromDataSource(@Nonnull ContractID contractID) {
+        final var entityId = toEntityId(contractID);
 
         return contractRepository
                 .findRuntimeBytecode(entityId.getId())
@@ -52,7 +49,7 @@ public class ContractBytecodeReadableKVState extends ReadableKVStateBase<Contrac
                 .orElse(null);
     }
 
-    @NotNull
+    @Nonnull
     @Override
     protected Iterator<ContractID> iterateFromDataSource() {
         return Collections.emptyIterator();
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/ContractBytecodeReadableKVState.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/ContractBytecodeReadableKVState.java
@@ -16,39 +16,43 @@
 
 package com.hedera.mirror.web3.state;
 
+import static com.hedera.mirror.web3.evm.utils.EvmTokenUtils.entityIdNumFromEvmAddress;
+import static com.hedera.mirror.web3.state.Utils.convertPbjBytesToBesuAddress;
+
 import com.hedera.hapi.node.base.ContractID;
 import com.hedera.hapi.node.state.contract.Bytecode;
+import com.hedera.mirror.common.domain.entity.EntityId;
 import com.hedera.mirror.web3.repository.ContractRepository;
 import com.hedera.pbj.runtime.io.buffer.Bytes;
 import com.swirlds.state.spi.ReadableKVStateBase;
 import java.util.Collections;
 import java.util.Iterator;
-import java.util.Optional;
 import org.jetbrains.annotations.NotNull;
 
 public class ContractBytecodeReadableKVState extends ReadableKVStateBase<ContractID, Bytecode> {
 
     private static final String KEY = "BYTECODE";
 
-    private final CommonEntityAccessor commonEntityAccessor;
     private final ContractRepository contractRepository;
 
-    protected ContractBytecodeReadableKVState(
-            final CommonEntityAccessor commonEntityAccessor, final ContractRepository contractRepository) {
+    protected ContractBytecodeReadableKVState(final ContractRepository contractRepository) {
         super(KEY);
-        this.commonEntityAccessor = commonEntityAccessor;
         this.contractRepository = contractRepository;
     }
 
     @Override
     protected Bytecode readFromDataSource(@NotNull ContractID contractID) {
-        final var contractEntity = commonEntityAccessor.get(contractID, Optional.empty());
-        if (contractEntity.isEmpty()) {
+        EntityId entityId;
+        if (contractID.hasContractNum()) {
+            entityId = EntityId.of(contractID.shardNum(), contractID.realmNum(), contractID.contractNum());
+        } else if (contractID.hasEvmAddress()) {
+            final var evmAddress = convertPbjBytesToBesuAddress(contractID.evmAddress());
+            entityId = EntityId.of(entityIdNumFromEvmAddress(evmAddress));
+        } else {
             return null;
         }
 
-        final var runtimeCode =
-                contractRepository.findRuntimeBytecode(contractEntity.get().getId());
+        final var runtimeCode = contractRepository.findRuntimeBytecode(entityId.getId());
         final var runtimeBytes = runtimeCode.map(Bytes::wrap).orElse(null);
         return runtimeBytes == null ? null : new Bytecode(runtimeBytes);
     }
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/ContractBytecodeReadableKVState.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/ContractBytecodeReadableKVState.java
@@ -45,9 +45,11 @@ public class ContractBytecodeReadableKVState extends ReadableKVStateBase<Contrac
             return null;
         }
 
-        final var runtimeCode = contractRepository.findRuntimeBytecode(entityId.getId());
-        final var runtimeBytes = runtimeCode.map(Bytes::wrap).orElse(null);
-        return runtimeBytes == null ? null : new Bytecode(runtimeBytes);
+        return contractRepository
+                .findRuntimeBytecode(entityId.getId())
+                .map(Bytes::wrap)
+                .map(Bytecode::new)
+                .orElse(null);
     }
 
     @NotNull
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/ContractBytecodeReadableKVState.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/ContractBytecodeReadableKVState.java
@@ -16,13 +16,13 @@
 
 package com.hedera.mirror.web3.state;
 
-import static com.hedera.mirror.web3.evm.utils.EvmTokenUtils.entityIdNumFromEvmAddress;
 import static com.hedera.mirror.web3.state.Utils.isMirror;
-import static com.hedera.services.utils.EntityIdUtils.toAddress;
+import static com.hedera.services.utils.EntityIdUtils.entityIdFromContractId;
 
 import com.hedera.hapi.node.base.ContractID;
 import com.hedera.hapi.node.state.contract.Bytecode;
 import com.hedera.mirror.common.domain.entity.EntityId;
+import com.hedera.mirror.common.util.DomainUtils;
 import com.hedera.mirror.web3.repository.ContractRepository;
 import com.hedera.pbj.runtime.io.buffer.Bytes;
 import com.swirlds.state.spi.ReadableKVStateBase;
@@ -70,11 +70,11 @@ public class ContractBytecodeReadableKVState extends ReadableKVStateBase<Contrac
 
     private EntityId toEntityId(@Nonnull final com.hedera.hapi.node.base.ContractID contractID) {
         if (contractID.hasContractNum()) {
-            return EntityId.of(contractID.shardNum(), contractID.realmNum(), contractID.contractNum());
+            return entityIdFromContractId(contractID);
         } else if (contractID.hasEvmAddress()) {
-            final var evmAddress = toAddress(contractID.evmAddress());
+            final var evmAddress = contractID.evmAddress().toByteArray();
             if (isMirror(evmAddress)) {
-                return EntityId.of(entityIdNumFromEvmAddress(evmAddress));
+                return DomainUtils.fromEvmAddress(evmAddress);
             } else {
                 return commonEntityAccessor
                         .getEntityByEvmAddressAndTimestamp(
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/ContractBytecodeReadableKVState.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/ContractBytecodeReadableKVState.java
@@ -78,8 +78,7 @@ public class ContractBytecodeReadableKVState extends ReadableKVStateBase<Contrac
                 return DomainUtils.fromEvmAddress(evmAddress);
             } else {
                 return commonEntityAccessor
-                        .getEntityByEvmAddressAndTimestamp(
-                                contractID.evmAddress().toByteArray(), Optional.empty())
+                        .getEntityByEvmAddressAndTimestamp(evmAddress, Optional.empty())
                         .map(Entity::toEntityId)
                         .orElse(EntityId.EMPTY);
             }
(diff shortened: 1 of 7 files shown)


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/ContractBytecodeReadableKVState.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    @Override
+    protected Bytecode readFromDataSource(@NotNull ContractID contractID) {
+        final var contractEntity = commonEntityAccessor.get(contractID, Optional.empty());
+        if (contractEntity.isEmpty()) {
+            return null;
+        }
+
+        final var runtimeCode =
```
IvanKavaldzhiev, line 50: We can skip `entity` check and directly read from `contract` table by `EntityId.of(contractID.getShard(), contractID.getRealm(), contractID.getNum()).getId()`

bilyana-gospodinova: The current implementation in `MirrorEntityAccess` calls `fetchEntityId` first. There we read the address from the `Entity` table and try to find it as an account. If we fail, we try to read it as a token. If it is not found again, we return `null` directly without any queries to the contract-related repositories. This is valid for both the runtime bytecode and the storage slot. I guess this was added because in there we don't have the id directly but only the address as a parameter and I wanted to double check.

bilyana-gospodinova: Comment addressed. Added also a case when we have an evm address set.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/ContractBytecodeReadableKVState.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    @Override
+    protected Bytecode readFromDataSource(@NotNull ContractID contractID) {
+        final var entityId = convertContractIDToEntityId(contractID);
+        if (entityId == null) {
+            return null;
+        }
+
+        final var runtimeCode = contractRepository.findRuntimeBytecode(entityId.getId());
```
kselveliev, line 48: One suggestion for this code section to be cleaner.
```
return contractRepository.findRuntimeBytecode(entityId.getId())
                .map(Bytes::wrap)
                .map(Bytecode::new)
                .orElse(null);
```

bilyana-gospodinova: You are right. It is better this way. Thanks, comment addressed.

kselveliev (approved):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/Utils.java
```diff
@@ -0,0 +1,48 @@
...
+@UtilityClass
+public class Utils {
+
+    public static Address convertPbjBytesToBesuAddress(final Bytes bytes) {
+        final var evmAddressBytes = bytes.toByteArray();
+        return Address.wrap(org.apache.tuweni.bytes.Bytes.wrap(evmAddressBytes));
+    }
+
+    @Nullable
+    public static EntityId convertContractIDToEntityId(final ContractID contractID) {
```
steven-sheehy, line 39: Use a simpler name so we can have many overloaded variants for each pbj type.

```suggestion
    public static EntityId toEntityId(final ContractID contractID) {
```

steven-sheehy, line 39: These need a test.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/ContractBytecodeReadableKVState.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @NotNull
+    @Override
+    protected Iterator<ContractID> iterateFromDataSource() {
+        return Collections.emptyIterator();
+    }
+
+    @Override
+    public long size() {
+        return 0;
```
steven-sheehy, line 63: Add tests to increase coverage.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/Utils.java
steven-sheehy, line 39: Should handle null and empty ContractID input

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/Utils.java
```diff
@@ -0,0 +1,48 @@
...
+import jakarta.annotation.Nullable;
+import lombok.CustomLog;
+import lombok.experimental.UtilityClass;
+import org.hyperledger.besu.datatypes.Address;
+
+@CustomLog
+@UtilityClass
+public class Utils {
+
+    public static Address convertPbjBytesToBesuAddress(final Bytes bytes) {
```
steven-sheehy, line 33:
```suggestion
    public static Address toBesuAddress(final Bytes bytes) {
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/ContractBytecodeReadableKVState.java
```diff
@@ -0,0 +1,65 @@
...
+        super(KEY);
+        this.contractRepository = contractRepository;
+    }
+
+    @Override
+    protected Bytecode readFromDataSource(@NotNull ContractID contractID) {
+        final var entityId = convertContractIDToEntityId(contractID);
+        if (entityId == null) {
+            return null;
+        }
```
steven-sheehy, line 46: If contractID can't be null, in what situations would this ever be null?

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/ContractBytecodeReadableKVState.java
bilyana-gospodinova: I added it in order to keep the existing check in `MirrorEntityAccess#fetchCodeIfPresent`. We probably don't need it here. I am removing the check.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/ContractBytecodeReadableKVState.java
```diff
@@ -0,0 +1,65 @@
...
+import com.hedera.hapi.node.base.ContractID;
+import com.hedera.hapi.node.state.contract.Bytecode;
+import com.hedera.mirror.web3.repository.ContractRepository;
+import com.hedera.pbj.runtime.io.buffer.Bytes;
+import com.swirlds.state.spi.ReadableKVStateBase;
+import java.util.Collections;
+import java.util.Iterator;
+import org.jetbrains.annotations.NotNull;
+
+public class ContractBytecodeReadableKVState extends ReadableKVStateBase<ContractID, Bytecode> {
```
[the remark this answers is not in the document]
bilyana-gospodinova: Done

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/Utils.java
```diff
@@ -0,0 +1,48 @@
...
+import com.hedera.mirror.common.domain.entity.EntityId;
+import com.hedera.pbj.runtime.io.buffer.Bytes;
+import jakarta.annotation.Nullable;
+import lombok.CustomLog;
+import lombok.experimental.UtilityClass;
+import org.hyperledger.besu.datatypes.Address;
+
+@CustomLog
+@UtilityClass
+public class Utils {
```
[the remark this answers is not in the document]
bilyana-gospodinova: Moved

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/Utils.java
bilyana-gospodinova: Renamed

bilyana-gospodinova: Added tests.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/ContractBytecodeReadableKVState.java
bilyana-gospodinova: Added tests

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/Utils.java
bilyana-gospodinova: Done

kselveliev (approved):

hedera-mirror-web3/src/test/java/com/hedera/services/utils/EntityIdUtilsTest.java
```diff
@@ -329,4 +330,50 @@ void toTokenIdFromEntityId() {
...
+                .realmNum(2)
+                .contractNum(3)
+                .build();
+
+        assertEquals(EntityId.of(1, 2, 3), EntityIdUtils.toEntityId(contractId));
+    }
+
+    @Test
+    void toEntityIdFromContractIdWithEvmAddress() {
+        final var address = Address.fromHexString("0x0000000000000000000000000000000000000001");
```
IvanKavaldzhiev, line 347: `evmAddress` usually will be not a long zero string, but have letters and numbers like `388C818CA8B9251b393131C08a736A67ccB19297`. We can use more realistic evmAddress value.

hedera-mirror-web3/src/main/java/com/hedera/services/utils/EntityIdUtils.java
```diff
@@ -194,6 +195,24 @@ public static com.hedera.hapi.node.base.TokenID toTokenId(final EntityId entityI
                 .build();
     }
 
+    public static Address toAddress(final com.hedera.pbj.runtime.io.buffer.Bytes bytes) {
+        final var evmAddressBytes = bytes.toByteArray();
+        return Address.wrap(org.apache.tuweni.bytes.Bytes.wrap(evmAddressBytes));
+    }
+
+    public static EntityId toEntityId(final com.hedera.hapi.node.base.ContractID contractID) {
```
IvanKavaldzhiev, line 203: After we merge this PR we should apply the same logic for AccountID. It seems the case for `evmAddress` is not covered there.

hedera-mirror-web3/src/test/java/com/hedera/services/utils/EntityIdUtilsTest.java
bilyana-gospodinova: I made the needed changes. It seems that we need to read the entity from the repository after all in order to get the entity id from an evm address.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/ContractBytecodeReadableKVState.java
IvanKavaldzhiev: We don't have to strongly comply with the legacy mono code, as now we have different business rules and checks.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/ContractBytecodeReadableKVState.java
```diff
@@ -0,0 +1,88 @@
...
+
+    private EntityId toEntityId(@Nonnull final com.hedera.hapi.node.base.ContractID contractID) {
+        if (contractID.hasContractNum()) {
+            return EntityId.of(contractID.shardNum(), contractID.realmNum(), contractID.contractNum());
+        } else if (contractID.hasEvmAddress()) {
+            final var evmAddress = toAddress(contractID.evmAddress());
+            if (isMirror(evmAddress)) {
+                return EntityId.of(entityIdNumFromEvmAddress(evmAddress));
+            } else {
+                return commonEntityAccessor
```
IvanKavaldzhiev, line 79: We can have a new method that returns only the EntityId instead of the whole Entity DTO, so that we reduce the throughput in DB and have a lighter query.

IvanKavaldzhiev: Actually `entityIdNumFromEvmAddress` will return the EntityId of the evmAddress. It was already converted when we convert a `ContractID` to `Address` with `toAddress`. Even if it have a byte evm address it will be converted to a Long numeric value.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/ContractBytecodeReadableKVState.java
```diff
@@ -0,0 +1,88 @@
...
+        return 0;
+    }
+
+    private EntityId toEntityId(@Nonnull final com.hedera.hapi.node.base.ContractID contractID) {
+        if (contractID.hasContractNum()) {
+            return EntityId.of(contractID.shardNum(), contractID.realmNum(), contractID.contractNum());
+        } else if (contractID.hasEvmAddress()) {
+            final var evmAddress = toAddress(contractID.evmAddress());
+            if (isMirror(evmAddress)) {
+                return EntityId.of(entityIdNumFromEvmAddress(evmAddress));
```
steven-sheehy, line 77: Can simplify util execution chain by just keeping it at bytes.

```suggestion
                return DomainUtils.fromEvmAddress(evmAddress);
```

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/ContractBytecodeReadableKVState.java
```diff
@@ -0,0 +1,88 @@
...
+    public long size() {
+        return 0;
+    }
+
+    private EntityId toEntityId(@Nonnull final com.hedera.hapi.node.base.ContractID contractID) {
+        if (contractID.hasContractNum()) {
+            return EntityId.of(contractID.shardNum(), contractID.realmNum(), contractID.contractNum());
+        } else if (contractID.hasEvmAddress()) {
+            final var evmAddress = toAddress(contractID.evmAddress());
+            if (isMirror(evmAddress)) {
```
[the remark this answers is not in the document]
bilyana-gospodinova: The conversion was not just for the `isMirror` check. `entityIdNumFromEvmAddress(address)` requires an `Address`.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/ContractBytecodeReadableKVState.java
bilyana-gospodinova: Yes, but if we have the address "b794f5ea0ba39494ce839613fffba74279579268" `entityIdNumFromEvmAddress` tries to convert it to 3285643.84543857934.203384024080 for example and an exception is thrown that the entity is not valid.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/Utils.java
```diff
@@ -39,4 +44,16 @@ public static Key parseKey(final byte[] keyBytes) {
...
         return null;
     }
+
+    public boolean isMirror(final Address address) {
+        return isMirror(address.toArrayUnsafe());
+    }
+
+    public static boolean isMirror(final byte[] address) {
+        if (address.length != EVM_ADDRESS_LEN) {
+            return false;
```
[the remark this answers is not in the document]
bilyana-gospodinova: Added tests.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/ContractBytecodeReadableKVState.java
bilyana-gospodinova: Done, thanks.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/ContractBytecodeReadableKVState.java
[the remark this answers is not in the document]
bilyana-gospodinova: Done after the suggestion in https://github.com/hashgraph/hedera-mirror-node/pull/9452#discussion_r1786482220

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/ContractBytecodeReadableKVState.java
bilyana-gospodinova: Added a new DB query.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/repository/EntityRepositoryTest.java
```diff
@@ -80,6 +80,25 @@ void findByEvmAddressAndDeletedTrueCall() {
                 .isEmpty();
     }
 
+    @Test
+    void findByEvmAddressSuccess() {
+        final var evmAddress = domainBuilder.evmAddress();
+        domainBuilder.entity().customize(e -> e.evmAddress(evmAddress)).persist();
+        assertThat(entityRepository.findIdByEvmAddressAndDeletedIsFalse(evmAddress))
+                .isNotNull();
```
IvanKavaldzhiev, line 88: We can assert that the entity returned by the query has the exact same EntityId as the persisted one

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/ContractBytecodeReadableKVState.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    @Override
+    public long size() {
+        return 0;
+    }
+
+    private EntityId toEntityId(@Nonnull final com.hedera.hapi.node.base.ContractID contractID) {
+        if (contractID.hasContractNum()) {
+            return EntityId.of(contractID.shardNum(), contractID.realmNum(), contractID.contractNum());
```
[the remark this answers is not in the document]
bilyana-gospodinova: Done

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/repository/EntityRepositoryTest.java
bilyana-gospodinova: Done

IvanKavaldzhiev (dismissed):
LGTM

kselveliev (dismissed):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/ContractBytecodeReadableKVState.java
```diff
@@ -0,0 +1,84 @@
...
+    private Long getEntityId(@Nonnull final com.hedera.hapi.node.base.ContractID contractID) {
+        if (contractID.hasContractNum()) {
+            return entityIdFromContractId(contractID).getId();
+        } else if (contractID.hasEvmAddress()) {
+            final var evmAddress = contractID.evmAddress().toByteArray();
+            if (isMirror(evmAddress)) {
+                return DomainUtils.fromEvmAddress(evmAddress).getId();
+            } else {
+                return commonEntityAccessor
+                        .getEntityIdByEvmAddress(contractID.evmAddress().toByteArray())
```
steven-sheehy, line 78: Double conversion

```suggestion
                        .getEntityIdByEvmAddress(evmAddress)
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/ContractBytecodeReadableKVState.java
steven-sheehy: The old query was cached while the new one is not. So the use of the new query might actually result in reduced throughput if contract IDs are repeated in multiple requests (as they are in k6 tests). We can add the cache to the new query as well, but then we have two ways to cache the same evm address so we might be 50% less likely to actually hit a warm cache. Also, PostgreSQL reads rows in 8KiB pages so reducing selected columns in single row results doesn't necessarily help the database load. I think we should revert this change.

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/ContractBytecodeReadableKVState.java
bilyana-gospodinova: Thank you for the detailed explanation. I reverted the changes.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/ContractBytecodeReadableKVState.java
steven-sheehy: Still a problem

bilyana-gospodinova: Fixed. Thank you!

IvanKavaldzhiev (approved):
LGTM

kselveliev (approved):

steven-sheehy (approved):

PR #9495, opened 2024-10-02
Enable Monitor In mainnet-citus and Decrease Worker Memory - Staging

jnels124:
enable monitor and decrease worker memory


jnels124:
handled by another PR

PR #9496, opened 2024-10-03
Reduce Worker Memory For Citus and Delete Mainnet Deployment in Mainnet NA

jnels124:
reduce worker memory for citus in mainnet na


xin-hedera (approved):
LGTM

steven-sheehy (approved):

PR #9497, opened 2024-10-03
Reduce Worker Memory For Citus and Delete Mainnet Deployment in Mainnet EU

jnels124:
reduce worker memory for citus in mainnet eu


steven-sheehy (approved):

xin-hedera (approved):
LGTM

PR #9538, opened 2024-10-09
Add `CustomCallOperation`

bilyana-gospodinova:
**Description**:

Fixes https://github.com/hashgraph/hedera-services/issues/15598

**Checklist**


IvanKavaldzhiev (approved):
LGTM!

Even though, we shouldn't mix mono and modularized code, I think with this small change, we will unblock some users scenarios and improve user experience and users won't have to wait for the full reusable services integration to get correct estimates for those corner cases where warm addresses are used.

kselveliev (approved):

steven-sheehy (approved):

Issue #9253, opened 2024-09-03
Implement `TokenRelationshipDatabaseReadableKVState`


IvanKavaldzhiev:
### Problem

Add implementation for `TokenRelationshipDatabaseReadableKVState`, so that we can return token relationships in PBJ format from DB.

### Solution

Add `TokenRelationshipDatabaseReadableKVState` in the following manner:

- Copy `TokenRelationshipDatabaseAccessor` logic as a starting point
- Add repository logic for reading token relationships from DB
- Add a new converter between DTO types to PBJ, so that the returned object is compliant with the hedera.app state

PR #7430, opened 2023-12-20
HIP-584 Historical: Fix historical NFT/Fungible token balance

0xivanov:
**Description**:

This PR includes the following fixes:
- get historical fungible token balance from `TokenBalanceRepository`
- get historical NFT balance from `NftRepository`. Historical NFT balance should be obtained from `account.getOwnedNfts()`, as explained in the code comments
- fix `findHistoricalTokenBalanceUpToTimestamp` in edge case where `balance_snapshot` and/or `base` queries do not match anything in the db. Currently reproducible in acceptance tests:
1. create account
2. make fungible token transfer +50
3. make fungible token transfer +50
4. make historical call with timestamp between the token transfers, expected balance is 50, but returns 0

I have observed that the state in db after these operations is the following:
`account_balance` - no entry for this accountId
`token_balance` - no entry for this accountId
`token_transfers` - 2 entries for this accountId, one for first transfer (50) and one for the second transfer (50)

`balance_snapshot` and `base` match nothing because there is no entry for this accountId before the given timestamp.
This causes the statement `tt.consensus_timestamp > s.consensus_timestamp` to be false and cannot match any token transfers, always returning 0 due to the coalesce at the end of the query.
NB. `token_balance` and `account_balance` entries get persisted in ~8 minute intervals so it is not certain that there is entry in the db after the first transfer.

Current solution:
If `balance_snapshot` and/or `base` do not match anything, find token transfers that occured after the account creation and before the given timestamp:
```sql
tt.consensus_timestamp >= coalesce((select consensus_timestamp from base), accountCreatedTimestamp)
```
Changed `>` to `>=` ensures the inclusion of the token transfer that occurred when the account was created.

**Optimisation proposals:**
- remove the `balance_snapshot` part of the query as it is not relevant to the token balances

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

- [x] Documented (Code comments, README, etc.)
- [x] Tested (unit, integration, etc.)

PR #9582, opened 2024-10-15
Implement `TokenRelationshipReadableKVState`

bilyana-gospodinova:
**Description**:
As part of the implementation for the reusable services we need to implement `TokenRelationshipReadableKVState`. It reads a token account association and converts it to `TokenRelation` PBJ type. In order to achieve this we read the token type, the account entity and the token balances. The token type is read by introducing a new DB query that returns only the required column instead of using the existing queries that return the full Token object.

Fixes https://github.com/hashgraph/hedera-mirror-node/issues/9253

**Checklist**


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenRepository.java
```diff
@@ -31,6 +33,28 @@ public interface TokenRepository extends CrudRepository<Token, Long> {
...
+                            select type
+                            from token_history
+                            where token_id = ?1
+                            order by lower(timestamp_range) desc
+                            limit 1
+                        )
+                    )
+                    """,
+            nativeQuery = true)
+    Optional<TokenTypeEnum> findTokenTypeById(Long tokenId);
```
IvanKavaldzhiev, line 56: Shouldn't we have a separate query for the historical cases that looks like this but has a timestamp filter and another only for latest state that searches only in `token` table?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/TokenRelationshipReadableKVState.java
```diff
@@ -0,0 +1,190 @@
...
+import java.util.function.Supplier;
+
+@Named
+public class TokenRelationshipReadableKVState extends ReadableKVStateBase<EntityIDPair, TokenRelation> {
+
+    private final TokenAccountRepository tokenAccountRepository;
+    private final TokenBalanceRepository tokenBalanceRepository;
+    private final TokenRepository tokenRepository;
+    private final NftRepository nftRepository;
+    private final CommonEntityAccessor commonEntityAccessor;
```
IvanKavaldzhiev, line 53: You san sort alphabetically the repositories here and in the constructor.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenRepository.java
bilyana-gospodinova: I think that the token type doesn't change so:
1. Either we search for the latest block and we perform the first part of the query against the "token" table.
2. Or the token was deleted at some point and exists only in the "token_history" table, hence the second part of the query.
So no matter the timestamp the type of the token is the same as well as the token id.

hedera-mirror-web3/src/test/java/com/hedera/hapi/node/state/token/TokenRelationTest.java
```diff
@@ -0,0 +1,423 @@
...
+ * limitations under the License.
+ */
+
+package com.hedera.hapi.node.state.token;
+
+import static com.hedera.pbj.runtime.ProtoTestTools.BOOLEAN_TESTS_LIST;
+import static com.hedera.pbj.runtime.ProtoTestTools.LONG_TESTS_LIST;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.junit.jupiter.api.Assertions.assertEquals;
```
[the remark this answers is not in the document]
bilyana-gospodinova: Done

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/TokenRelationshipReadableKVState.java
bilyana-gospodinova: Done. Sorted.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/TokenRelationshipReadableKVState.java
```diff
@@ -0,0 +1,190 @@
...
+            final AccountID accountID,
+            final TokenAccount tokenAccount,
+            final Optional<Long> timestamp) {
+        return TokenRelation.newBuilder()
+                .tokenId(tokenID)
+                .accountId(accountID)
+                .balanceSupplier(getBalance(tokenType, tokenAccount, timestamp))
+                .frozen(tokenAccount.getFreezeStatus() == TokenFreezeStatusEnum.FROZEN)
+                .kycGranted(tokenAccount.getKycStatus() != TokenKycStatusEnum.REVOKED)
+                .automaticAssociation(tokenAccount.getAutomaticAssociation())
```
IvanKavaldzhiev, line 136: Sort alphabetically all fields

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/TokenRelationshipReadableKVState.java
```diff
@@ -0,0 +1,190 @@
...
+                .build();
+    }
+
+    /**
+     * Determines fungible or NFT balanceSupplier based on block context.
+     */
+    private Supplier<Long> getBalance(
+            final TokenTypeEnum tokenType, final TokenAccount tokenAccount, final Optional<Long> timestamp) {
+        if (tokenType.equals(TokenTypeEnum.NON_FUNGIBLE_UNIQUE)) {
```
IvanKavaldzhiev, line 145: You can use single Suppliers.memoize() invocation, wrapping the if statement inside.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/TokenRelationshipReadableKVState.java
```diff
@@ -0,0 +1,190 @@
...
+        if (account.isEmpty()) {
+            return null;
+        }
+        // If the AccountID is defined by alias, we need to define a new AccountID with num, so that we can execute the
+        // next search queries.
+        final var searchableAccountId = accountId.hasAccountNum()
+                ? accountId
+                : toAccountId(account.get().toEntityId());
+
+        final var tokenType = findTokenType(tokenId);
```
IvanKavaldzhiev, line 88: We should also take into account the `timestamp`.

bilyana-gospodinova: I left a comment related to this here - https://github.com/hashgraph/hedera-mirror-node/pull/9582#discussion_r1800881298. What do you think about that?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/TokenRelationshipReadableKVState.java
bilyana-gospodinova: But why? This is the order of the fields in the model (`TokenRelation`). So looking at the model you can easily set consecutive fields using the builder without missing a field. Also, the most important fields are at the top of the model and first in the builder method chain. Also, if one of the fields gets renamed, we will need to refactor such builder invocations as well which will be hard to maintain.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/TokenRelationshipReadableKVState.java
bilyana-gospodinova: Thanks, done.

kselveliev (approved):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenRepository.java
IvanKavaldzhiev: You're right. We should just find the token in any of the tables and get the type. That's one of the fields which cannot change during different timestamps, so we can use this simplified query.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/TokenRelationshipReadableKVState.java
IvanKavaldzhiev: It makes it easier to navigate in the code base when we have more fields like >10, but here we have only a few, so it might be not so beneficial to sort here.

IvanKavaldzhiev (dismissed):
LGTM

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/config/EvmConfiguration.java
```diff
@@ -157,6 +158,7 @@ CacheManager cacheManagerToken() {
                 CACHE_NAME_NFT,
                 CACHE_NAME_NFT_ALLOWANCE,
                 CACHE_NAME_TOKEN,
+                CACHE_NAME_TOKEN_TYPE,
```
steven-sheehy, line 161: This means for token type we are caching using `expireAfterWrite=1s,maximumSize=10000,recordStats` which would be very sub-optimal. We can cache this data for much longer. Something like `expireAfterAccess=24h,maximumSize=100000,recordStats`

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/Utils.java
```diff
@@ -31,6 +31,7 @@
 public class Utils {
 
     public static final long DEFAULT_AUTO_RENEW_PERIOD = 7776000L;
+    public static final Long ZERO_BALANCE = 0L;
```
steven-sheehy, line 34: Seems like it was moved here but not used by new code? Also not sure this really qualifies as a useful constant since 0L is not really a special value and self explanatory.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/TokenRelationshipReadableKVState.java
```diff
@@ -0,0 +1,189 @@
...
+     * Fungible Token Balance Explanation:
+     * Non-historical Call:
+     * The balanceSupplier is obtained from `tokenAccount.getBalance()`.
+     * Historical Call:
+     * In historical block queries, since the `token_account` table lacks historical state for fungible balances,
+     * the fungible balance is determined from the `token_balance` table using the `findHistoricalTokenBalanceUpToTimestamp` query.
+     */
+    private Long getFungibleBalance(final TokenAccount tokenAccount, final Optional<Long> timestamp) {
+        return timestamp
+                .map(t -> tokenBalanceRepository.findHistoricalTokenBalanceUpToTimestamp(
```
steven-sheehy, line 176: Are you sure we don't need the logic to check account created timestamp first like the old accessor had? Did you test this change against real data?

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/TokenRelationshipReadableKVState.java
```diff
@@ -0,0 +1,189 @@
...
+        final var timestamp = ContractCallContext.get().getTimestamp();
+        final var account = findAccount(accountId, timestamp);
+        if (account.isEmpty()) {
+            return null;
+        }
+        // If the AccountID is defined by alias, we need to define a new AccountID with num, so that we can execute the
+        // next search queries.
+        final var searchableAccountId = accountId.hasAccountNum()
+                ? accountId
+                : toAccountId(account.get().toEntityId());
```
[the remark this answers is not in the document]
steven-sheehy: Checked with Michael T and he confirmed that the EntityIDPair will always be in `shard.realm.num` format.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenRepository.java
```diff
@@ -31,6 +33,28 @@ public interface TokenRepository extends CrudRepository<Token, Long> {
...
+                            from token
+                            where token_id = ?1
+                        ),
+                        (
+                            select type
+                            from token_history
+                            where token_id = ?1
+                            order by lower(timestamp_range) desc
+                            limit 1
+                        )
```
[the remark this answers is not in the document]
bilyana-gospodinova: What if the token is deleted at some point? Won't we have an entry just in the history table in this case after the next block?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/TokenRelationshipReadableKVState.java
bilyana-gospodinova: I removed this check because some lines above there is:
```
final var account = findAccount(accountId, timestamp)
if (account.isEmpty()) {
            return null;
}
```
So, if the account does not exist we would not get to `getBalance()` anyway.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/TokenRelationshipReadableKVState.java
```diff
@@ -0,0 +1,189 @@
...
+        if (tokenType.isEmpty()) {
+            return null;
+        }
+
+        final var tokenAccount = findTokenAccount(tokenId, searchableAccountId, timestamp);
+        if (tokenAccount.isEmpty()) {
+            return null;
+        }
+
+        return tokenAccount
```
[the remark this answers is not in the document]
bilyana-gospodinova: Done

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/TokenRelationshipReadableKVState.java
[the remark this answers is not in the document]
bilyana-gospodinova: That's good. Thanks. I removed the alias lookup.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/Utils.java
bilyana-gospodinova: It was used in the `TokenRelationshipReadableKVState` as well in previous commits but it seems that it is no longer needed. I removed it, thanks.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/TokenRelationshipReadableKVState.java
```diff
@@ -0,0 +1,189 @@
...
+        // If the AccountID is defined by alias, we need to define a new AccountID with num, so that we can execute the
+        // next search queries.
+        final var searchableAccountId = accountId.hasAccountNum()
+                ? accountId
+                : toAccountId(account.get().toEntityId());
+
+        final var tokenType = findTokenType(tokenId);
+        if (tokenType.isEmpty()) {
+            return null;
+        }
```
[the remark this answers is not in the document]
bilyana-gospodinova: Correct, I moved the check.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/TokenRelationshipReadableKVState.java
```diff
@@ -0,0 +1,189 @@
...
+        this.tokenRepository = tokenRepository;
+    }
+
+    @Override
+    protected TokenRelation readFromDataSource(@Nonnull EntityIDPair key) {
+        final var tokenId = key.tokenId();
+        final var accountId = key.accountId();
+        if (tokenId == null || accountId == null) {
+            return null;
+        }
```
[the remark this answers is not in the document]
bilyana-gospodinova: I added a check for DEFAULT as well.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/config/EvmConfiguration.java
bilyana-gospodinova: I updated the cache settings, thanks.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TokenRepository.java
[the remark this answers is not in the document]
steven-sheehy: Deleted tokens stay in the `token` table. In general, the main table contains the current state (including deletions) and the history table contains past state. Any ID in the history table should also exist in the main table.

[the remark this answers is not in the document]
bilyana-gospodinova: I see. Thank you for the clarification! I updated the query to make a select only against the token table.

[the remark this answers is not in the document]
steven-sheehy: Does it not work without the `@Query`? `findTypeByTokenId` should auto generate the query, I think.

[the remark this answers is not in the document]
bilyana-gospodinova: That was the first thing that I tried. However, `findTypeByTokenId` tries to return a full `Token` object so we get a runtime exception because it cannot be parsed to a `TokenTypeEnum`.

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/TokenRelationshipReadableKVState.java
bilyana-gospodinova: I will copy the check from the old accessor. It seems that the check was introduced as part of fixing a bug before. Reference: https://github.com/hashgraph/hedera-mirror-node/pull/7430

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/TokenRelationshipReadableKVState.java
```diff
@@ -0,0 +1,166 @@
...
+                || AccountID.DEFAULT.equals(accountId)
+                || TokenID.DEFAULT.equals(tokenId)) {
+            return null;
+        }
+
+        final var timestamp = ContractCallContext.get().getTimestamp();
+        final var account = findAccount(accountId, timestamp);
+        if (account.isEmpty()) {
+            return null;
+        }
```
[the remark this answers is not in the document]
bilyana-gospodinova: It seems that there are cases that the account's created_timestamp is needed. Added the old logic from the accessor to this state.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/TokenRelationshipReadableKVState.java
steven-sheehy: I asked that this change be tested before being added back. I believed it didn't need to be there but wanted real tests to validate that instead of assuming some old code or PR added it on purpose.

I've created a test that proves we don't need it. Please remove this logic and add this test to `TokenBalanceRepositoryTest`:

```java

    @Test
    void findHistoricalTokenBalanceUpToTimestampMissingTokenBalance() {
        var accountId = domainBuilder.entityId();
        var tokenId = domainBuilder.entityId();
        long snapshotTimestamp = domainBuilder.timestamp();
        domainBuilder.accountBalance().customize(ab -> ab.id(new Id(snapshotTimestamp, TREASURY_ENTITY_ID))).persist();
        var tokenTransfer1 = domainBuilder.tokenTransfer()
                .customize(tt -> tt.id(new TokenTransfer.Id(snapshotTimestamp + 2, tokenId, accountId))).persist();
        var tokenTransfer2 = domainBuilder.tokenTransfer()
                .customize(tt -> tt.id(new TokenTransfer.Id(snapshotTimestamp + 3, tokenId, accountId))).persist();

        assertThat(tokenBalanceRepository.findHistoricalTokenBalanceUpToTimestamp(tokenId.getId(), accountId.getId(),
                snapshotTimestamp))
                .get()
                .isEqualTo(0L);
        assertThat(tokenBalanceRepository.findHistoricalTokenBalanceUpToTimestamp(tokenId.getId(), accountId.getId(),
                snapshotTimestamp + 1)) // Assumed account creation
                .get()
                .isEqualTo(0L);
        assertThat(tokenBalanceRepository.findHistoricalTokenBalanceUpToTimestamp(tokenId.getId(), accountId.getId(),
                snapshotTimestamp + 2))
                .get()
                .isEqualTo(tokenTransfer1.getAmount());
        assertThat(tokenBalanceRepository.findHistoricalTokenBalanceUpToTimestamp(tokenId.getId(), accountId.getId(),
                snapshotTimestamp + 3))
                .get()
                .isEqualTo(tokenTransfer2.getAmount() + tokenTransfer2.getAmount());
    }
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/TokenRelationshipReadableKVState.java
```diff
@@ -0,0 +1,175 @@
...
+
+    /**
+     * For the latest block we have the balance directly as a field in the TokenAccount object.
+     * For the historical block we need to execute a query to calculate the historical balance, but
+     * we first need to find the account created timestamp and the token type in order to use the correct repository.
+     * If the account creation is after the passed timestamp - return 0L (the entity was not created).
+     */
+    private Supplier<Long> getBalance(final TokenAccount tokenAccount, final Optional<Long> timestamp) {
+        return Suppliers.memoize(() -> timestamp
+                .map(ts -> findAccount(toAccountId(tokenAccount.getAccountId()), Optional.of(ts))
```
steven-sheehy, line 131: Believe you can just pass the existing one.
```suggestion
                .map(ts -> findAccount(toAccountId(tokenAccount.getAccountId()), timestamp)
```

bilyana-gospodinova: After removing the check for account creation timestamp, we don't need to fetch the account anymore, so this is no more applicable.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/TokenRelationshipReadableKVState.java
bilyana-gospodinova: Comment addressed.

kselveliev (approved):

steven-sheehy (approved):
LGTM

PR #9591, opened 2024-10-16
Add REST API query normalization

### LINKED ISSUE #9113: Add query normalization to support REST API Caching

edwin-greene:
**Description**:
Adds a means to normalize REST API queries.

**Related issue(s)**:

Fixes #9113 

**Notes for reviewer**:

**Checklist**


hedera-mirror-rest/middleware/requestNormalizer.js
```diff
@@ -0,0 +1,78 @@
...
+
+import _ from 'lodash';
+import qs from 'qs';
+
+import {getOpenApiMap} from './openapiHandler.js';
+
+const openApiMap = getOpenApiMap();
+
+// Do not sort these query parameters as the results of the sql query changes based on their order
+const NON_SORTED_PARAMS = ['balance', 'block.hash', 'block.number', 'nonce', 'scheduled', 'timestamp'];
```
steven-sheehy, line 25: I don't understand how the sql query can change based upon the param order. That would make results inconsistent for users with the same params. Can you clarify maybe with an example?

hedera-mirror-rest/middleware/requestNormalizer.js
```diff
@@ -0,0 +1,78 @@
...
+  return _.isEmpty(query) ? path : path + '?' + sortQueryProperties(query);
+};
+
+const sortQueryProperties = (query) => {
+  // Sort the order of the parameters in the query
+  return qs.stringify(query, stringifyOptions);
+};
+
+// alphabeticalSort is used to sort the properties of the Object
+const stringifyOptions = {encode: false, indices: false, sort: alphabeticalSort};
```
steven-sheehy, line 76: nit: Move constant object to the top of file

hedera-mirror-rest/__tests__/middleware/requestNormalizer.test.js
```diff
@@ -0,0 +1,179 @@
...
+    {
+      description: 'Balance is not sorted',
+      input: '/api/v1/accounts?balance=true&balance=false',
+      expected: '/api/v1/accounts?balance=true&balance=false&limit=25&order=asc',
+    },
+    {
+      description: 'Timestamp values are not sorted',
+      input: '/api/v1/transactions?timestamp=1639010141.000000001&timestamp=1639010141.000000000',
+      expected:
+        '/api/v1/transactions?limit=25&order=desc&timestamp=1639010141.000000001&timestamp=1639010141.000000000',
```
steven-sheehy, line 156: Shouldn't we sort by values for same param to provide consistent results and reduce cache misses?

hedera-mirror-rest/middleware/openapiHandler.js
```diff
@@ -80,6 +82,58 @@ const getV1OpenApiObject = () => {
...
 
+/**
+ * Get the path to parameter properties map for the OpenApi Spec
+ *
+ * @returns {Map<string, Array<{parameterName, defaultValue}>>}
+ */
+const getOpenApiMap = () => {
+  if (_.isUndefined(openApiMap)) {
+    const openApiObject = getV1OpenApiObject();
+    const map = new Map();
```
steven-sheehy, line 93: We should sort the map so that we iterate it in the order we want the normalized output.

hedera-mirror-rest/middleware/requestNormalizer.js
```diff
@@ -0,0 +1,78 @@
...
+      !NON_SORTED_PARAMS.includes(param?.parameterName) &&
+      Array.isArray(normalizedQuery[param?.parameterName])
+    ) {
+      // Sort the order of the parameters within the array
+      normalizedQuery[param.parameterName].sort();
+    }
+  }
+
+  return formatPathQuery(path, normalizedQuery);
+};
```
steven-sheehy, line 60: This is inefficient since it loops over N params once to generate a new list, then again to stringify and sort that list. Instead, we should have the OpenAPI params already sorted by name on startup and loop once to directly construct final sorted string. Then we also don't need to clone the params.

hedera-mirror-rest/middleware/openapiHandler.js
```diff
@@ -80,6 +82,58 @@ const getV1OpenApiObject = () => {
...
+  const pathObject = openApiObject.paths[path];
+  const parameters = pathObject?.get?.parameters;
+  if (parameters === undefined) {
+    return {};
+  }
+
+  return (
+    parameters
+      // Each open api parameter is prefixed by #/components/parameters/, which is 24 characters long
+      .map((p) => p.$ref?.substring(24))
```
steven-sheehy, line 120: Should put `#/components/parameters/` in a constant and first filter the parameters by prefix and then use the string size to get substring.

hedera-mirror-rest/__tests__/middleware/requestNormalizer.test.js
```diff
@@ -0,0 +1,179 @@
...
+    },
+    {
+      description: 'Block query parameters are added',
+      input: '/api/v1/blocks',
+      expected: '/api/v1/blocks?limit=25&order=desc',
+    },
+    {
+      description: 'Known and unknown query parameters are sorted',
+      input: '/api/v1/blocks?order=asc&unknown=3',
+      expected: '/api/v1/blocks?limit=25&order=asc&unknown=3',
```
steven-sheehy, line 60: We shouldn't need to handle unknown query parameters since we throw an error if any unknown.

hedera-mirror-rest/middleware/openapiHandler.js
edwin-greene: Sounds good I have updated this.

hedera-mirror-rest/__tests__/middleware/requestNormalizer.test.js
edwin-greene: I have updated the comments with examples of when the array parameters should not be sorted, as in some cases the sorting of the array parameters changes the result.

hedera-mirror-rest/middleware/requestNormalizer.js
edwin-greene: Thanks I added some examples that I found from running the query normalizer on the spec tests.

hedera-mirror-rest/middleware/requestNormalizer.js
edwin-greene: Got it, I have updated this to generate the query string with a single loop.

hedera-mirror-rest/middleware/requestNormalizer.js
edwin-greene: This should now be good, this method was removed.

hedera-mirror-rest/__tests__/middleware/requestNormalizer.test.js
edwin-greene: At this current revision I have simplified the looping to generate the query string to such an extent that it ignores unknown parameters. This may not be what we want, considering that the unknown parameter will later throw an error, I will work on fixing this next.

hedera-mirror-rest/middleware/requestNormalizer.js
steven-sheehy: I see. So it's for APIs that honor the last such parameter. But the way we implemented it it will occur for all APIs that have such param even if they don't have that behavior. 

Can we double check if any of these params are used in multiple APIs where one only uses last while other uses all? If they're always last one in wins then we can just change the behavior to collapse them in the normalizer as well.

steven-sheehy: Either way, it might be nice to add some sentence to each parameter's description that indicates if multiple values are passed to this parameter, only the last one will be honored. Then we can search for that sentence when normalizing and use the presence of that wording to control normalizing logic.

hedera-mirror-rest/__tests__/middleware/requestNormalizer.test.js
steven-sheehy: In the future, we could use this openapi map to implement the unknown parameter handling and remove all the custom logic we added throughout.

hedera-mirror-rest/__tests__/middleware/requestNormalizer.test.js
```diff
@@ -0,0 +1,149 @@
...
+
+import {normalizeRequestQueryParams} from '../../middleware/requestNormalizer.js';
+import request from 'supertest';
+import {openApiValidator} from '../../middleware/index.js';
+import express from 'express';
+
+const app = express();
+openApiValidator(app);
+
+describe('request normalizer', () => {
```
xin-hedera, line 25: it's worth to register a test-only middleware inside openapiHandler `openApiValidator` function, after the OpenAPI 
validator middleware, to normalize the the request URL, and set it in the http response as done in this integration test.

That way, we can

- add expected normalized URLs to all existing json spec files, and
- update the integration template.js to validate it

That'll get us full coverage and less error-prone to miss it in the future.

It's a lot of work though, just to update all existing specs...

hedera-mirror-rest/middleware/requestNormalizer.js
```diff
@@ -0,0 +1,86 @@
...
+    const value = query[name];
+    if (value !== undefined) {
+      queryString = Array.isArray(value)
+        ? appendArrayToQuery(queryString, name, value)
+        : appendToQuery(queryString, name, value);
+    } else if (param?.defaultValue !== undefined) {
+      // Add the default value to the query parameter
+      queryString = appendToQuery(queryString, name, param.defaultValue);
+    }
+  }
```
xin-hedera, line 64: why can't we build an array of `k=v` fragments and do a `join('&')` at the end, instead of appending one by one?

xin-hedera (changes requested):

hedera-mirror-rest/__tests__/middleware/requestNormalizer.test.js
edwin-greene: I am working on this now, I will add it as a separate comment to addressing the other feedback as it is a larger commit.

steven-sheehy: Can't we do this as a follow up?

xin-hedera: fine with that.

xin-hedera (dismissed):
LGTM

hedera-mirror-rest/middleware/requestNormalizer.js
```diff
@@ -0,0 +1,81 @@
...
+  return queryArray.length !== 0 ? path + '?' + queryArray.join('&') : path;
+};
+
+const getArrayValues = (name, valuesArray) => {
+  if (!NON_SORTED_PARAMS.includes(name)) {
+    // Sort the order of the parameters within the array
+    valuesArray.sort();
+  } else if (COLLAPSABLE_PARAMS.includes(name)) {
+    // Only add the last item in the array to the query parameter
+    return name + '=' + valuesArray[valuesArray.length - 1];
```
steven-sheehy, line 75: Is it possible valuesArray can be empty in which case this would throw an error? Should add a check and a test just in case

hedera-mirror-rest/middleware/requestNormalizer.js
```diff
@@ -0,0 +1,81 @@
...
+ *
+ * It is expected that this is called after any error handling for the request
+ *
+ * @param openApiRoute {string}
+ * @param path {string}
+ * @param query request query object
+ * @returns {string}
+ */
+const normalizeRequestQueryParams = (openApiRoute, path, query) => {
+  const openApiParameters = openApiMap.get(openApiRoute);
```
steven-sheehy, line 52: Is it possible the route is ever not found? Maybe add a check and a test just in case.
```suggestion
  const openApiParameters = openApiMap.get(openApiRoute) || {};
```

hedera-mirror-rest/middleware/requestNormalizer.js
```diff
@@ -0,0 +1,81 @@
...
+    if (value !== undefined) {
+      const queryValue = Array.isArray(value) ? getArrayValues(name, value) : name + '=' + value;
+      queryArray.push(queryValue);
+    } else if (param?.defaultValue !== undefined) {
+      // Add the default value to the query parameter
+      queryArray.push(name + '=' + param.defaultValue);
+    }
+  }
+
+  return queryArray.length !== 0 ? path + '?' + queryArray.join('&') : path;
```
steven-sheehy, line 66: Why do we need an array just to generate a string? We can just concatenate strings inside the loop and avoid the array.

hedera-mirror-rest/middleware/requestNormalizer.js
```diff
@@ -0,0 +1,81 @@
...
+  const queryArray = [];
+  for (const param of openApiParameters) {
+    const name = param.parameterName;
+    const value = query[name];
+    if (value !== undefined) {
+      const queryValue = Array.isArray(value) ? getArrayValues(name, value) : name + '=' + value;
+      queryArray.push(queryValue);
+    } else if (param?.defaultValue !== undefined) {
+      // Add the default value to the query parameter
+      queryArray.push(name + '=' + param.defaultValue);
```
steven-sheehy, line 62: `name + '='` is repeated in multiple places. Pull up outside the conditionals so it occurs once per loop.

steven-sheehy (changes requested):

hedera-mirror-rest/middleware/requestNormalizer.js
steven-sheehy: I see now you did have it that way then switched it based upon Xin's suggestion. Not sure why that wasn't discussed since I had suggested it be String earlier.

hedera-mirror-rest/middleware/requestNormalizer.js
edwin-greene: Thanks, there is some trust in the parameters from OpenApi, where it seems that the object passed from it, if an array, is always an array with 2 elements. But if that changes at some point, I have added a check to the beginning of the method to check to see if at least 1 element is in the array, if not then the method doesn't need to go any further.

hedera-mirror-rest/middleware/requestNormalizer.js
edwin-greene: I tried this, or something similar: ` const openApiParameters = openApiMap.get(openApiRoute) || []` and added a test case for a request that has no matching openApi parameters. I found that if that request had query parameters that the current iteration of the method would remove them. So instead I added a check for the open api parameters, and if missing, then the normalized query will be generated with querystring.

hedera-mirror-rest/middleware/requestNormalizer.js
edwin-greene: Thanks I see, I had thought that the second variation, which reintroduced the array, had kept the best loop performance enhancements of the first revision (the string concatenation revision), but I neglected to consider that the .join at the end was another loop through the same parameters. I have reverted it to the string concatenation version with some additions per feedback.

hedera-mirror-rest/middleware/requestNormalizer.js
edwin-greene: Thanks, I updated this to consolidate those concatenations. There is still one exception, when the query is an array of values then they are concatenated in that method to save the need from adding the .isArray conditional to the main loop a second time.

hedera-mirror-rest/middleware/requestNormalizer.js
```diff
@@ -0,0 +1,99 @@
...
+
+import {getOpenApiMap} from './openapiHandler.js';
+import {filterKeys} from '../constants.js';
+import _ from 'lodash';
+import * as querystring from 'node:querystring';
+
+const openApiMap = getOpenApiMap();
+
+// Multiple values of these params can be collapsed to the last value
+const COLLAPSABLE_PARAMS = [filterKeys.BALANCE, filterKeys.BLOCK_HASH, filterKeys.NONCE, filterKeys.SCHEDULED];
```
edwin-greene, line 25: I've updated this list, adding block.hash

hedera-mirror-rest/middleware/requestNormalizer.js
```diff
@@ -0,0 +1,99 @@
...
+};
+
+const appendToQuery = (queryString, queryValue) => {
+  return queryString + getQueryPrefix(queryString) + queryValue;
+};
+
+const getNormalizedArrayValue = (name, valueArray) => {
+  if (_.isEmpty(valueArray)) {
+    return;
+  }
```
edwin-greene, line 82: This is not covered in tests, as the OpenApi object doesn't seem to generate empty arrays.

steven-sheehy (approved):

hedera-mirror-rest/middleware/requestNormalizer.js
steven-sheehy, line 25: Probably should updated description to note behavior of block.hash in openapi.yml

jnels124 (approved):
LGTM

hedera-mirror-rest/middleware/requestNormalizer.js
edwin-greene: Thanks I have updated it

xin-hedera (approved):
LGTM

steven-sheehy (approved):

Issue #9257, opened 2024-09-03
Implement `AirdropsDatabaseReadableKVState`


IvanKavaldzhiev:
### Problem

We need to add a new type of database accessor to read airdrop related data, so that we have specific ReadableKVState instance for this type of data.

### Solution

Add AirdropsDatabaseReadableKVState class that reads airdrop data and converts it to AccountPendingAirdrop PBJ type.

Add unit tests to cover all cases including historical reads and correct mapping for all fields.

PR #9598, opened 2024-10-17
Implement `AirdropsReadableKVState`

bilyana-gospodinova:
**Description**:
As part of the implementation for the reusable (modularized) services we need to implement `AirdropsReadableKVState`. It reads a pending airdrop id and converts it to `AccountPendingAirdrop` PBJ type.

Fixes https://github.com/hashgraph/hedera-mirror-node/issues/9257

**Checklist**

### DIFF
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/AirdropsReadableKVState.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/AirdropsReadableKVState.java
@@ -1,0 +1,80 @@
+/*
+ * Copyright (C) 2024 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hedera.mirror.web3.state;
+
+import static com.hedera.services.utils.EntityIdUtils.toEntityId;
+
+import com.hedera.hapi.node.base.PendingAirdropId;
+import com.hedera.hapi.node.base.PendingAirdropValue;
+import com.hedera.hapi.node.state.token.AccountPendingAirdrop;
+import com.hedera.mirror.web3.common.ContractCallContext;
+import com.hedera.mirror.web3.repository.TokenAirdropRepository;
+import com.swirlds.state.spi.ReadableKVStateBase;
+import jakarta.annotation.Nonnull;
+import jakarta.inject.Named;
+import java.util.Collections;
+import java.util.Iterator;
+
+@Named
+public class AirdropsReadableKVState extends ReadableKVStateBase<PendingAirdropId, AccountPendingAirdrop> {
+
+    private final TokenAirdropRepository tokenAirdropRepository;
+
+    protected AirdropsReadableKVState(final TokenAirdropRepository tokenAirdropRepository) {
+        super("PENDING_AIRDROPS");
+        this.tokenAirdropRepository = tokenAirdropRepository;
+    }
+
+    @Override
+    protected AccountPendingAirdrop readFromDataSource(@Nonnull PendingAirdropId key) {
+        if (key.hasNonFungibleToken()) {
+            return AccountPendingAirdrop.DEFAULT;
+        }
+
+        final var senderId = toEntityId(key.senderId()).getId();
+        final var receiverId = toEntityId(key.receiverId()).getId();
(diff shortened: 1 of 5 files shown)


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/AirdropsReadableKVState.java
```diff
@@ -0,0 +1,80 @@
...
+    private final TokenAirdropRepository tokenAirdropRepository;
+
+    protected AirdropsReadableKVState(final TokenAirdropRepository tokenAirdropRepository) {
+        super("PENDING_AIRDROPS");
+        this.tokenAirdropRepository = tokenAirdropRepository;
+    }
+
+    @Override
+    protected AccountPendingAirdrop readFromDataSource(@Nonnull PendingAirdropId key) {
+        if (key.hasNonFungibleToken()) {
```
kselveliev, line 44: Why do we need this statement?
If it is an nft should we just return AccountPendingAirdrop with an amount of 1. 
Since the default value is 0 in the model.
I'm not really sure what is the correct business logic here.

kselveliev: I just saw the model docs -> 
  *               If the token is a non-fungible/unique token, this message SHOULD NOT be set, and if
  *               set, this field MUST be `0`.
 So this should be correct

kselveliev (dismissed):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/AirdropsReadableKVState.java
bilyana-gospodinova: Correct - the model documentation requires this return value.

steven-sheehy: I don't think this is true. That means nft airdrops are not stored in state and always return empty? Instead, we should lookup nft airdrop in the db and just return AccountPendingAirdrop with unset value.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/AirdropsReadableKVState.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    @Override
+    public long size() {
+        return 0;
+    }
+
+    private AccountPendingAirdrop mapToAccountPendingAirdrop(final long amount) {
+        return AccountPendingAirdrop.newBuilder()
+                .pendingAirdropValue(mapToPendingAirdropValue(amount))
```
steven-sheehy, line 73: Do we not need to populate the next and previous links?

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/AirdropsReadableKVState.java
IvanKavaldzhiev: I think they are used in services for iterating for example trying to delete an account and fetching all their sent pending airdrops. We won't have such cases in our web3 simulations, so for the moment it seems redundant to populate these links.

IvanKavaldzhiev (dismissed):
LGTM

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/AirdropsReadableKVState.java
bilyana-gospodinova: `AccountPendingAirdrop` has 3 fields:
```
@Nullable PendingAirdropValue pendingAirdropValue,
@Nullable PendingAirdropId previousAirdrop,
@Nullable PendingAirdropId nextAirdrop
```
The last two we set to `null` as per the comment here: https://github.com/hashgraph/hedera-mirror-node/pull/9598#discussion_r1808867018

The first field (`PendingAirdropValue`) has only: `long amount` as a field. So, why should we read the nft airdrop from the DB when we know that we are going to return `AccountPendingAirdrop` with all null values? @steven-sheehy, can you please clarify?

steven-sheehy: The return value of `AccountPendingAirdrop.DEFAULT` is correct when the ID exists, but you have not verified existence of the `PendingAirdropId` in the database before returning it. If it does not exist you should return null.

You can see [upstream](https://github.com/hashgraph/hedera-services/blob/6d4c18868da1f4fb985615608b74e1ba18dcc6fd/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenAirdropHandler.java#L352) that an AccountPendingAirdrop with a null value is put into the airdrop store for NFTs.

bilyana-gospodinova: Got it. I updated the nft case implementation and added more tests.

kselveliev (approved):

steven-sheehy (approved):
LGTM

PR #9605, opened 2024-10-18
Added gas correction to failing acceptance tests

nickeynikolovv:
**Description**:
There were some failing acceptance tests for read functions because of [this](https://github.com/hashgraph/hedera-services/issues/10828) change in services, so a correction to the gas is added temporarily to the failing tests

**Related issue(s)**:

Fixes #

**Notes for reviewer**:
This correction is temporary and will not be needed when reusable services implementation in mirror node is done

**Checklist**


bilyana-gospodinova:
I think that we can leave the tests commented out for now (until we merge the reusable/modularized services). The point of the tests is to verify that the estimation is correct and right now - it isn't. Just the tests pass which is kind of a false positive.

nickeynikolovv:
Sure, we can also leave them commented as they are now. We can close this pr if will proceed that way

steven-sheehy:
Agree with Bilyana, so will close

PR #9622, opened 2024-10-21
Convert EVM properties to a map of configuration properties

### LINKED ISSUE #9584: Construct map of properties for initializing TransactionExecutor
### Problem

In order to construct the `TransactionExecutor` component, used to run a transaction in the modularized EVM components, we need to pass a Map of properties that will be used upstream in the added libraries. They will override some properties, that will be set, so that they suit the mirror node needs.

### Solution

We should find a way to extract the properties from the `MirrorNodeEvmProperties` bean class we currently have. They should be put in a `Map<String, String>` and passed as an argument, when we run a transaction via the `TransactionExecutor`.

### Alternatives

_No response_

kselveliev:
**Description**:

This PR converts MirrorNodeEvmProperties to a map of configuration properties for transaction executors.

In order to use the modularized services code we need to build a TransactionExecutor object which has a 
`final Map<String, String> properties` parameter - those are configuration properties such as

```
contracts.chainId=295
contracts.evm.version=v0.51
```

`public TransactionExecutor newExecutor(@NonNull final State state, @NonNull final Map<String, String> properties){`

These kinds of properties in web3 are stored in `MirrorNodeEvmProperties`.

We can also pass an empty map in the transactionExecutor creator method if we want to have the default values in modularized code.

**Related issue(s)**:

Fixes #9584 

**Checklist**


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/ModularizedProperties.java
```diff
@@ -0,0 +1,138 @@
...
+     * keys and values that will be passed as properties to hedera-app transaction executors.
+     *
+     * @return a map containing the configuration keys as keys and their corresponding values as strings.
+     */
+    public Map<String, String> getPropertiesMap() {
+        Map<String, String> propertiesMap = new HashMap<>();
+        propertiesMap.put(
+                AUTO_RENEW_TARGET_TYPES,
+                mirrorNodeEvmProperties.getAutoRenewTargetTypes().toString());
+        propertiesMap.put(
```
steven-sheehy, line 94: The problem here is now we have two levels of properties each with their own validation and defaults: Spring boot properties and consensus node properties. If we have a default to match consensus nodes and later they change their default upstream, we won't be able to catch these types of problems.

I think since upstream has validation and defaults, we should just use those. That way we don't need to do explicit mapping between Spring properties and node properties. Only properties that we need to explicitly override to work with mirror node should be in that map. Everything else should be omitted to use the upstream default. Of course this results in extra properties for a little while until later we can delete the relevant hardcoded properties in MirrorNodeEvmProperties.


```java
@ConfigurationProperties(prefix = "hedera.mirror.web3.evm")
public class MirrorNodeEvmProperties {
    ...
    private final Map<String, String> properties = Map.of("contracts.chainId", chainIdBytes32().toBigInteger().toString());
}
```

Then we can override using application.yml:

```yaml
hedera:
  mirror:
    web3:
      evm:
        properties:
          contracts.allowCreate2: "false"
```


If any properties are dynamic per request, we just create a copy of `properties` and `put` the dynamic properties into that for reach request.

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/ModularizedProperties.java
```diff
@@ -0,0 +1,138 @@
...
+
+    private final MirrorNodeEvmProperties mirrorNodeEvmProperties;
+
+    private static final String EVM_PREFIX = "v";
+
+    private static final String AUTO_RENEW_TARGET_TYPES = "autoRenew.targetTypes";
+
+    private static final String CHAIN_ID_BYTES32 = "contracts.chainId";
+
+    private static final String EVM_VERSION = "contracts.evm.version";
```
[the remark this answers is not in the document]
kselveliev: I’m not sure if we can read the keys directly from the upstream code
For example upstream reads the config from 
```
@ConfigData("contracts")
public record ContractsConfig(
        @ConfigProperty(defaultValue = "true") @NetworkProperty boolean itemizeStorageFees,
        @ConfigProperty(defaultValue = "1062787,1461860") Set<Long> permittedDelegateCallers,
        @ConfigProperty(defaultValue = "31536000") @NetworkProperty long referenceSlotLifetime,
        @ConfigProperty(defaultValue = "100") @NetworkProperty int freeStorageTierLimit,
        @ConfigProperty(defaultValue = "0til100M,2000til450M") @NetworkProperty String storageSlotPriceTiers,
        @ConfigProperty(defaultValue = "7890000") @NetworkProperty long defaultLifetime,

```

which as a string key will map to "contracts.itemizeStorageFees"

I think with reflection we might obtain the keys dynamically from the upstream class but it’s probablly not worth it.

Maybe here we can just use the default values as you said use a minimal set of ovverides in our map.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/ModularizedProperties.java
kselveliev: If this is okay then we can delete the ModularizedProperties class

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
```diff
@@ -167,10 +169,23 @@ public class MirrorNodeEvmProperties implements EvmProperties {
     @NotNull
     private HederaNetwork network = HederaNetwork.TESTNET;
 
+    private final Map<String, String> defaultOverrides = Map.of(
```
kselveliev, line 172: Contains properties which we definitely need to override like chainId.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/ConfigKeyExtractor.java
```diff
@@ -0,0 +1,49 @@
...
+import java.lang.reflect.Field;
+
+/**
+ * The ConfigKeyExtractor class reads configuration keys from the modularized services code and returns them.
+ */
+public class ConfigKeyExtractor {
+
+    public static final String DOT_SEPARATOR = ".";
+    public static final String CHAIN_ID_KEY = "chainId";
+    public static final String MAX_GAS_REFUND_PERCENTAGE_KEY = "maxRefundPercentOfGasLimit";
```
steven-sheehy, line 30: This class doesn't really serve any purpose since we have to hardcode strings to pass via reflection just to get back similar hardcoded strings. We should just directly use the hardcoded strings like `contracts.chainId` in `MirrorNodeEvmProperties`.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
```diff
@@ -167,10 +169,23 @@ public class MirrorNodeEvmProperties implements EvmProperties {
...
+
     @Getter
     @Min(1)
     private int feesTokenTransferUsageMultiplier = 380;
 
+    public Map<String, String> getProperties() {
+        Map<String, String> props = new HashMap<>();
+        props.putAll(properties);
+        props.putAll(defaultOverrides);
+        return Collections.unmodifiableMap(props);
```
steven-sheehy, line 186: The order is wrong. Defaults are used first and users can override them via properties config.

But why do we even need 4 maps to get the properties (including unmodifiable)? Why can't we just do what I suggested in my original comment?

```java
private Map<String, String> properties = Map.of("contracts.chainId", chainIdBytes32().toBigInteger().toString());
```

steven-sheehy (changes requested):

bilyana-gospodinova:
I think this PR is a good place to add a utility method that returns all of the properties (that need override) that we need to pass to the `TransactionExecutors` as a map.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
kselveliev: My concern here was if we have this yaml 
```
hedera:
  mirror:
    web3:
      evm:
        properties:
          contracts.allowCreate2: "false"
```

And we have default 
private Map<String, String> properties = Map.of("contracts.chainId", chainIdBytes32().toBigInteger().toString());


Then (i think) the end result will be [contracts.allowCreate2: "false"]  and we lose the contracts.chainId property in the map.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/ConfigKeyExtractor.java
```diff
@@ -0,0 +1,49 @@
...
+        return getConfigKey(ContractsConfig.class, CHAIN_ID_KEY);
+    }
+
+    public static String getMaxGasRefundPercentageKey() {
+        return getConfigKey(ContractsConfig.class, MAX_GAS_REFUND_PERCENTAGE_KEY);
+    }
+
+    public static String getConfigKey(Class<?> configClass, String fieldName) {
+        try {
+            Field field = configClass.getDeclaredField(fieldName);
```
bilyana-gospodinova, line 42: Perhaps this method can be used just in the tests instead in order to verify that a property in services has been renamed/deleted.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/ConfigKeyExtractor.java
kselveliev: Comment addressed. Deleted the class.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
```diff
@@ -167,10 +165,24 @@ public class MirrorNodeEvmProperties implements EvmProperties {
...
+    private Map<String, String> properties = new HashMap<>();
+
     @Getter
     @Min(1)
     private int feesTokenTransferUsageMultiplier = 380;
 
+    public Map<String, String> getProperties() {
+        Map<String, String> props = new HashMap<>();
+        props.putAll(defaultOverrides);
+        props.putAll(properties);
```
bilyana-gospodinova, line 182: `properties` is just an empty `HashMap`, so why is it added at all? Do we need it?

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/properties/ModularizedPropertiesTest.java
```diff
@@ -0,0 +1,75 @@
...
+import com.hedera.mirror.web3.Web3IntegrationTest;
+import com.hedera.node.config.data.ContractsConfig;
+import com.swirlds.config.api.ConfigData;
+import java.lang.reflect.Field;
+import java.util.Map;
+import lombok.RequiredArgsConstructor;
+import org.junit.jupiter.api.Test;
+
+@RequiredArgsConstructor
+class ModularizedPropertiesTest extends Web3IntegrationTest {
```
bilyana-gospodinova, line 31: There is a `MirrorNodeEvmPropertiesTest` class - can't we add the tests there? Is there a specific reason for a new class?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
kselveliev: Added integration tests covering the scenarios here.
It seems that we need 2 maps since yaml will overwrite any default values that we have in the initial map and we will lose them.
Switched the order as per the initial comment.
First we put the default values then we put the ones from the yaml config in properties.
Properties existing in both maps will be taken from the config and override initial values.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/ConfigKeyExtractor.java
kselveliev: Comment addressed

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/properties/ModularizedPropertiesTest.java
kselveliev: The MirrorNodeEvmProperties currently only contains unit tests.
If we make it extend integration test and inject MirrorNodeEvmProperties we will have some failing tests.
I added the new tests in a new class since they are only integration tests

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
kselveliev: I don't want to return the original properties since we'll later pass it into TransactionExecutors and then we can't be sure what will happen to it. So I prefer to copy the values in a new map an return it.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/properties/ModularizedPropertiesTest.java
bilyana-gospodinova: Ok, that's fine then.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
bilyana-gospodinova: Got it - `properties` gets filled in from yaml.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/properties/ModularizedPropertiesTest.java
```diff
@@ -0,0 +1,75 @@
...
+            Field field = configClass.getDeclaredField(fieldName);
+            ConfigData configProperty = configClass.getAnnotation(ConfigData.class);
+            return configProperty.value() + DOT_SEPARATOR + field.getName();
+        } catch (NoSuchFieldException e) {
+            throw new IllegalArgumentException("Field " + fieldName + " not found in " + configClass.getSimpleName());
+        }
+    }
+
+    @Test
+    void getModularizedProperties() {
```
bilyana-gospodinova, line 57: Perhaps we need one more test for the case when we have the same property in defaultOverrides and in the yaml. Then we need to verify that the expected value is set.

kselveliev: Done

kselveliev: Added this scenario in the same test

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
kselveliev: Just worked on my machine as well. Comment addressed.

bilyana-gospodinova (dismissed):
LGTM

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/properties/ModularizedPropertiesTest.java
```diff
@@ -0,0 +1,76 @@
...
+import com.hedera.mirror.web3.Web3IntegrationTest;
+import com.hedera.node.config.data.ContractsConfig;
+import com.swirlds.config.api.ConfigData;
+import java.lang.reflect.Field;
+import java.util.Map;
+import lombok.RequiredArgsConstructor;
+import org.junit.jupiter.api.Test;
+
+@RequiredArgsConstructor
+class ModularizedPropertiesTest extends Web3IntegrationTest {
```
steven-sheehy, line 31: Agree with Bilyana. The standard is if there's both unit and integration to name it with `IntegrationTest` suffix.

```suggestion
class MirrorNodeEvmPropertiesIntegrationTest extends Web3IntegrationTest {
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
```diff
@@ -167,10 +164,20 @@ public class MirrorNodeEvmProperties implements EvmProperties {
     @NotNull
     private HederaNetwork network = HederaNetwork.TESTNET;
 
+    private Map<String, String> properties = Map.of(
```
steven-sheehy, line 167: Add `@NotNull`

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
```diff
@@ -167,10 +164,20 @@ public class MirrorNodeEvmProperties implements EvmProperties {
...
+            "contracts.maxRefundPercentOfGasLimit",
+            String.valueOf(maxGasRefundPercentage()));
+
     @Getter
     @Min(1)
     private int feesTokenTransferUsageMultiplier = 380;
 
+    public Map<String, String> getProperties() {
+        return Map.copyOf(properties);
+    }
```
steven-sheehy, line 179: Why do we need to copy this every time it's requested? This would be extra memory overhead for no gain.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
```diff
@@ -72,7 +72,6 @@ public class MirrorNodeEvmProperties implements EvmProperties {
     private double estimateGasIterationThresholdPercent = 0.10d;
 
     private boolean directTokenCall = true;
-
```
steven-sheehy, line 75: We should revert the 3 new line deletions in this class since most variables have spacing in between and this would make it inconsistent.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/properties/ModularizedPropertiesTest.java
```diff
@@ -0,0 +1,76 @@
...
+                .doesNotContainEntry(CHAIN_ID_KEY_CONFIG, "296")
+                .containsEntry(MAX_GAS_REFUND_PERCENTAGE_KEY_CONFIG, "100")
+                .containsAllEntriesOf(YAML_PROPERTIES);
+    }
+
+    @Test
+    void verifyUpstreamPropertiesExist() {
+        assertThat(getConfigKey(ContractsConfig.class, CHAIN_ID)).isEqualTo(CHAIN_ID_KEY_CONFIG);
+        assertThat(getConfigKey(ContractsConfig.class, MAX_GAS_REFUND_PERCENTAGE))
+                .isEqualTo(MAX_GAS_REFUND_PERCENTAGE_KEY_CONFIG);
```
steven-sheehy, line 69: Instead of hardcoding the two, we should loop over properties.getProperties().keySet() and verify getConfigKey for each.

hedera-mirror-web3/src/test/resources/config/application.yml
```diff
@@ -2,6 +2,10 @@ hedera:
   mirror:
     web3:
       evm:
+        properties:
+          contracts.allowCreate2: "false"
+          contracts.maxGasPerSec: "10000000"
+          contracts.chainId: "297"
```
steven-sheehy, line 8: nit: sort

steven-sheehy (changes requested):

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/properties/ModularizedPropertiesTest.java
kselveliev: Comment addressed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
kselveliev: Comment addressed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
kselveliev: I was scared to pass the real reference to the modularized code. Then I check what it does with it, initially just loops over it and passes the values in a config builder and does not use it again.
So I removed this and now just use Getter annotation for getProperties.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/properties/ModularizedPropertiesTest.java
kselveliev: Comment addressed

hedera-mirror-web3/src/test/resources/config/application.yml
kselveliev: Comment addressed

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmPropertiesIntegrationTest.java
```diff
@@ -0,0 +1,83 @@
...
+                // override from yaml
+                .doesNotContainEntry(CHAIN_ID_KEY_CONFIG, "296")
+                .containsEntry(MAX_GAS_REFUND_PERCENTAGE_KEY_CONFIG, "100")
+                .containsAllEntriesOf(YAML_PROPERTIES);
+    }
+
+    @Test
+    void verifyUpstreamPropertiesExist() {
+        Set<String> propertyKeys = properties.getProperties().keySet();
+        for (String configKey : propertyKeys) {
```
nickeynikolovv, line 74: You can use java streams here

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmPropertiesIntegrationTest.java
```diff
@@ -0,0 +1,83 @@
...
+            Field field = configClass.getDeclaredField(fieldName);
+            ConfigData configProperty = configClass.getAnnotation(ConfigData.class);
+            return configProperty.value() + DOT_SEPARATOR + field.getName();
+        } catch (NoSuchFieldException e) {
+            throw new IllegalArgumentException("Field " + fieldName + " not found in " + configClass.getSimpleName());
+        }
+    }
+
+    private static String getContractsConfigKey(final String configKey) {
+        String fieldName = configKey.replace(CONTRACTS_CONFIG + DOT_SEPARATOR, "");
```
nickeynikolovv, line 56: can use var

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmPropertiesIntegrationTest.java
```diff
@@ -0,0 +1,83 @@
...
+        return getConfigKey(ContractsConfig.class, fieldName);
+    }
+
+    @Test
+    void getModularizedProperties() {
+        Map<String, String> modularizedProperties = properties.getProperties();
+        assertThat(modularizedProperties)
+                .isNotEmpty()
+                // override from yaml
+                .doesNotContainEntry(CHAIN_ID_KEY_CONFIG, "296")
```
nickeynikolovv, line 66: Why is testnet chainId hardcoded - Isn't the test going to fail if run on testnet?

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmPropertiesIntegrationTest.java
kselveliev: Comment addressed

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmPropertiesIntegrationTest.java
kselveliev: Comment addressed

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmPropertiesIntegrationTest.java
kselveliev: For integration tests the chainId is currently explicitly set in test yaml file to 297. This test yaml of the web3 is probably the same in every environment, so I don't think this could be an issue,.

nickeynikolovv (approved):
LGTM

steven-sheehy (approved):
LGTM

PR #9626, opened 2024-10-23
Add an application cache for REST API 

### LINKED ISSUE #9111: Add an application cache for REST API

jnels124:
**Description**:
* implement application cache

**Related issue(s)**:

Fixes #9111 

**Notes for reviewer**:

**Checklist**


hedera-mirror-rest/middleware/responseCacheHandler.js
```diff
@@ -0,0 +1,126 @@
...
+ * header, and therefore that request header value should be used as part of the cache key, the cache
+ * implementation stores the response body as the original JSON object without any encoding applied. Thus it
+ * is the same regardless of the accept encoding specified, and chosen by the compression middleware.
+ *
+ * Current key format:
+ *
+ *   path?query - In the future, this will utilize Edwin's request normalizer (9113).
+ */
+const cacheKeyGenerator = (req) => {
+  return req.originalUrl;
```
xin-hedera, line 108: we should use the hash of the url as the key. We can have very long URLs, and it'll be slower for redis. to lookup long keys.

Suggestion is try to utilize xxhash3 128bit hashing algorithm provided by https://github.com/ktrongnhan/xxhash-addon. The key will be 32 bytes, the hex encoding of the 128-bit hash.

hedera-mirror-rest/middleware/responseCacheHandler.js
```diff
@@ -0,0 +1,126 @@
...
+
+// Response middleware that caches the completed response.
+const responseCacheUpdateHandler = async (req, res, next) => {
+  const compressionEnabled = config.cache.response.compress;
+  const responseCacheKey = res.locals[responseCacheKeyLabel];
+  const responseBody = res.locals[responseBodyLabel];
+
+  // 304 cached here as well due to etag handling
+  if (responseBody && responseCacheKey && (res.statusCode === 200 || res.statusCode === 304)) {
+    const cacheControlHeaderExpiry = getCacheControlExpiry(res.getHeaders()[CACHE_CONTROL_HEADER]);
```
xin-hedera, line 76: nit, encapsulate logic of the fallback into `getCacheControlExpiry`, perhaps rename the function to `getCacheControlExpiryOrDefault`

xin-hedera (changes requested):

hedera-mirror-rest/cache.js
```diff
@@ -73,6 +73,41 @@ export class Cache {
     return this.redis.flushall();
   }
 
+  async getSingleWithTtl(key) {
+    if (!this.ready) {
+      return undefined;
+    }
+
+    let valueWithTtl = undefined; // Cache miss to caller
+    await this.redis
```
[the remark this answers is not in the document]
jnels124: didn't see this will take a look

hedera-mirror-rest/middleware/responseCacheHandler.js
```diff
@@ -0,0 +1,126 @@
...
+const VARY_HEADER = 'vary';
+const DEFAULT_REDIS_EXPIRY = 1;
+
+let cache = new Cache();
+
+// Response middleware that checks for and returns cached response.
+const responseCacheCheckHandler = async (req, res, next) => {
+  const startTime = res.locals[requestStartTime] || Date.now();
+  const responseCacheKey = cacheKeyGenerator(req);
+  const cachedTtlAndValue = await cache.getSingleWithTtl(responseCacheKey);
```
xin-hedera, line 36: the handling of http HEAD requests should be different. We only need to know if the key exists and it's TTL, we should not need the body or compressed body. If the we have a cached response, we set response code to 304 and complete the processing with an empty body.

hedera-mirror-rest/middleware/responseCacheHandler.js
```diff
@@ -0,0 +1,126 @@
...
+};
+
+// Response middleware that caches the completed response.
+const responseCacheUpdateHandler = async (req, res, next) => {
+  const compressionEnabled = config.cache.response.compress;
+  const responseCacheKey = res.locals[responseCacheKeyLabel];
+  const responseBody = res.locals[responseBodyLabel];
+
+  // 304 cached here as well due to etag handling
+  if (responseBody && responseCacheKey && (res.statusCode === 200 || res.statusCode === 304)) {
```
xin-hedera, line 75: we should handle all 2xx code and 304. we do return 206 in one endpoint.

It's also very likely when expressjs determines to return 304 for a HEAD request, the body is erased. So we'll be completely wrong here to cache an empty body.

The easiest way to handle 304 is just not handle it here. We can try to get a better understanding and improve the logic in the future.

xin-hedera (changes requested):

hedera-mirror-rest/middleware/responseCacheHandler.js
jnels124: Should it be 304 though? I thought that code indicates to the requester that they have the value in client cache and therefore would not actually need the response body to be sent on the following GET?

hedera-mirror-rest/middleware/responseCacheHandler.js
jnels124: This will only cache the 304 if the body is present. Often times, the computed etag will match the `if-none-match` header on the request (meaning this client already has the data). We should still cache the non empty response here to be available for other requesters.

hedera-mirror-rest/middleware/responseCacheHandler.js
jnels124: this seems like the more appropriate logic.

```javascript
const startTime = res.locals[requestStartTime] || Date.now();
  const responseCacheKey = cacheKeyGenerator(req);
  const cachedTtlAndValue = await cache.getSingleWithTtl(responseCacheKey);

  if (cachedTtlAndValue) {
    const {ttl: redisTtl, value: redisValue} = cachedTtlAndValue;
    const cachedResponse = Object.assign(new CachedApiResponse(), redisValue);
    const compressedBody = cachedResponse.compressedBody;
    const clientCached = req.get(CONDITIONAL_HEADER) === cachedResponse.headers[ETAG_HEADER]; // 304
    const statusCode = clientCached ? 304 : 200;
    const isHead = req.method === 'HEAD';

    cachedResponse.headers[CACHE_CONTROL_HEADER] = `public, max-age=${redisTtl}`;
    
    if (isHead || clientCached) {
        res.set(cachedResponse.headers);
        res.status(statusCode);
        res.end();
        return;
    }

    if (compressedBody) {
      cachedResponse.body = Buffer.from(cachedResponse.body);
      const acceptsGzip = req.get('accept-encoding')?.includes('gzip');
      if (!acceptsGzip) {
        cachedResponse.body = unzipSync(cachedResponse.body).toString();
      } else {
        cachedResponse.headers[CONTENT_ENCODING_HEADER] = 'gzip';
      }
    }
    
    res.set(cachedResponse.headers);
    res.status(statusCode);
    res.send(cachedResponse.body);
```

hedera-mirror-rest/middleware/responseCacheHandler.js
xin-hedera: missed that the body is from the `locals` object. sounds good.

hedera-mirror-rest/middleware/responseCacheHandler.js
xin-hedera: yup this looks correct.

hedera-mirror-rest/middleware/responseCacheHandler.js
```diff
@@ -0,0 +1,126 @@
...
+
+    const elapsed = Date.now() - startTime;
+    logger.info(
+      `${req.ip} ${req.method} ${req.originalUrl} from cache (ttl: ${redisTtl}) in ${elapsed} ms: 200`
+    );
+    return;
+  }
+
+  res.locals[responseCacheKeyLabel] = responseCacheKey;
+  next();
```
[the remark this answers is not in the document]
jnels124: It wasn't going to the next middleware function because it was setup with `app.use` instead `app.useAsync`. I have changed that and moved the `ipmetrics` middleware to be invoked before the cache check handler so that it is not skipped when enabled.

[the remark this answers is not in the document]
jnels124: It wasn't working because the middleware was enabled with `app.use` instead of `app.useAsync`.  Looking at things though, need to make some additional changes if we want the `ipMetrics` middleware to function correctly.

hedera-mirror-rest/middleware/responseCacheHandler.js
jnels124: still working on this one. Needs to be recompiled when doing npm install for this package and having some issues

hedera-mirror-rest/middleware/metricsHandler.js
```diff
@@ -47,6 +47,7 @@ const recordIpAndEndpoint = async (req, res, next) => {
   if (req.route !== undefined) {
     ipEndpointHistogram.labels(req.route.path, ipMask(req.ip)).inc();
   }
+  next();
```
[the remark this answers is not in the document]
jnels124: nope

xin-hedera (dismissed):
LGTM

hedera-mirror-rest/middleware/responseCacheHandler.js
```diff
@@ -0,0 +1,130 @@
...
+ *
+ *   path?query - In the future, this will utilize Edwin's request normalizer (9113).
+ */
+const cacheKeyGenerator = (req) => {
+  return crypto.createHash('md5').update(req.originalUrl).digest('hex');
+};
+
+const getCacheControlExpiryOrDefault = (headerValue) => {
+  if (headerValue) {
+    const maxAge = headerValue.match(/^.*max-age=(\d+)/);
```
steven-sheehy, line 116: Move regex to a constant so it doesn't have to be recompiled.

hedera-mirror-rest/middleware/responseCacheHandler.js
```diff
@@ -0,0 +1,130 @@
...
+          cachedResponse.headers[CONTENT_ENCODING_HEADER] = 'gzip';
+        }
+      }
+      res.send(cachedResponse.body);
+    }
+
+    const elapsed = Date.now() - startTime;
+    logger.info(
+      `${req.ip} ${req.method} ${req.originalUrl} from cache (ttl: ${redisTtl}) in ${elapsed} ms: ${statusCode}`
+    );
```
steven-sheehy, line 71: We should not have another info level log for every request. That would double our log storage. Move to debug with if debug level check.

steven-sheehy (changes requested):

hedera-mirror-rest/middleware/responseCacheHandler.js
jnels124: this will be the only line that logs. The response handler middleware will not run

docs/configuration.md
```diff
@@ -503,10 +503,12 @@ The following table lists the available properties along with their default valu
...
 | `hedera.mirror.rest.cache.entityId.maxAge`                         | 1800                    | The number of seconds until the entityId cache entry expires                                                                                                                                  |
 | `hedera.mirror.rest.cache.entityId.maxSize`                        | 100000                  | The maximum number of entries in the entityId cache                                                                                                                                           |
 | `hedera.mirror.rest.cache.token.maxSize`                           | 100000                  | The maximum number of entries in the token cache                                                                                                                                              |
+| `hedera.mirror.rest.cache.response.enabled`                        | false                   | Whether or not the Redis based REST API response cache is enabled. If so, Redis itself must be enabled and properly configured.                                                               |
```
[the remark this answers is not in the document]
jnels124: Correct

hedera-mirror-rest/middleware/responseCacheHandler.js
steven-sheehy: Ah, sounds good

hedera-mirror-rest/config/application.yml
```diff
@@ -74,7 +77,7 @@ hedera:
         connectTimeout: 10000
         enabled: true
         maxBackoff: 128000
-        maxMemory: 250Mb
+        maxMemory: 750Mb
```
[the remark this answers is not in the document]
jnels124: will be ready for that. And make adjustments if needed.

steven-sheehy (approved):

xin-hedera (approved):
LGTM

PR #9634, opened 2024-10-24
Track parser's recoverable errors at warning level

Neurone:
**Description**:

Track recoverable errors of the parser at warning level.

Considering that other parts of the code seem to consider this as a warning, and considering that the user has the ability to set whether to consider this type of error as a real error to worry about with the use of the `HEDERA_MIRROR_IMPORTER_PARSER_HALTONERROR` env parameter, it would be useful to set these messages to warning level so that we can exclude them from the logs, while retaining the ability to track all other parser's errors.

**Related issue(s)**:

N/A

**Notes for reviewer**:

N/A

**Checklist**

### DIFF
--- a/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/util/Utility.java
+++ b/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/util/Utility.java
@@ -195,7 +195,7 @@ public class Utility {
     /**
      * Handle a parser recoverable error. Depending on the value of the system property
      * HEDERA_MIRROR_IMPORTER_PARSER_HALTONERROR, when false (default), the provided message and arguments are logged at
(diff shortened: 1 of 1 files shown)


steven-sheehy:
We currently have alerts that trigger based upon recoverable errors that this would break. For full mirror nodes, these are errors not warnings that have helped us catch serious issues with the record stream. Plus, setting them to warn doesn't help since by default we log warnings.

Neurone:
I see. I care about setting this as a warning because in my case, for example, I only want to log errors that require human intervention, so I set the logs that way, overriding the default logging levels via application.yml.

My understanding of a recoverable error is that this is something that the mirror node will take care of and I don't actually need to do anything on the DB.

Is this the case, or do I need to do something if I see these errors in the logs?

PR #9668, opened 2024-10-28
fix: Fix contract deployment with value

### LINKED ISSUE #9636: Gas estimation on contract deployment with payable constructor reverts
### Description

If we try to estimate gas for a contract deployment with a payable constructor, the web3 module reverts instead of returning a proper gas value.

### Steps to reproduce

We have a simple contract:

```
pragma solidity ^0.8.24;

contract PayableConstructor {
    constructor() payable {}
}
```

The bytecode is: 
```
0x6080604052603e80600f5f395ff3fe60806040525f80fdfea2646970667358221220a38ca427a091f1d38af455b9ec034fe31be0290310c3e2888017a1fce83553d464736f6c634300081a0033
```

---

- if we execute a gas estimation against the mirror node web 3 module via POST to https://testnet.mirrornode.hedera.com/api/v1/contracts/call with that body:
```
{
    "value":1000000,
    "estimate": true,
    "from":"0x6e5d3858f53fc66727188690946631bde0466b1a",
    "data":"0x6080604052603e80600f5f395ff3fe60806040525f80fdfea2646970667358221220a38ca427a091f1d38af455b9ec034fe31be0290310c3e2888017a1fce83553d464736f6c634300081a0033"
}
```

it reverts with a:
```
{
    "_status": {
        "messages": [
            {
                "message": "Bad Request",
                "detail": "to field must not be empty",
                "data": ""
            }
        ]
    }
}
```

- if we remove the value from the body:
```
{
    "estimate": true,
    "from":"0x6e5d3858f53fc66727188690946631bde0466b1a",
    "data":"0x6080604052603e80600f5f395ff3fe60806040525f80fdfea2646970667358221220a38ca427a091f1d38af455b9ec034fe31be0290310c3e2888017a1fce83553d464736f6c634300081a0033"
}
```
then we get a expected result:
```
{
    "result": "0x000000000000e228"
}
```

I also tried passing "to" as a null or empty string but it failed again.

### Additional context

_No response_

### Hedera network

_No response_

### Version

testnet

### Operating system

None

kselveliev:
**Description**:

This PR fixes contract deployment when the contract deployment is made with value.

This PR modifies :
Modified `ContractCallRequest` - change the 'to' validation to check if the call is for contract deployment

`BytecodeUtils` - Added a new method `isInitBytecode` which checks if a given data is an init bytecode - needed in `ContractCallRequest` validation

`ContractControllerTest` - Added new test and modified a few existing ones based on the new changes to the validation of the to field.

Modified `ContractCallAddressThisTest` - added tests for deployment of payable contract with and without value
Modified `ContractCallServiceERCTokenModificationFunctionsTest` - added tests for deployment of non payable contract with and without value.

**Related issue(s)**:

Fixes #9636 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/utils/BytecodeUtils.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/utils/BytecodeUtils.java
@@ -33,7 +33,7 @@ import lombok.experimental.UtilityClass;
 @UtilityClass
 public class BytecodeUtils {
 
-    public static final String SKIP_INIT_CODE_CHECK = "HEDERA_MIRROR_WEB3_EVM_SKIP_CHECKINITCODE";
+    public static final String SKIP_INIT_CODE_CHECK = "HEDERA_MIRROR_WEB3_EVM_SKIPINITCODECHECK";
     private static final String CODECOPY = "39";
     private static final String RETURN = "f3";
     private static final long MINIMUM_INIT_CODE_SIZE = 14L;
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/utils/BytecodeUtils.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/utils/BytecodeUtils.java
@@ -33,7 +33,7 @@ import lombok.experimental.UtilityClass;
 @UtilityClass
 public class BytecodeUtils {
 
-    public static final String SKIP_INIT_CODE_CHECK = "HEDERA_MIRROR_WEB3_EVM_CHECKINITCODE";
+    public static final String SKIP_INIT_CODE_CHECK = "HEDERA_MIRROR_WEB3_EVM_SKIP_CHECKINITCODE";
     private static final String CODECOPY = "39";
     private static final String RETURN = "f3";
     private static final long MINIMUM_INIT_CODE_SIZE = 14L;
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/utils/BytecodeUtils.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/utils/BytecodeUtils.java
@@ -30,7 +30,7 @@ import lombok.experimental.UtilityClass;
  * </p>
  */
 @UtilityClass
-public class RuntimeBytecodeExtractor {
+public class BytecodeUtils {
 
     private static final String CODECOPY = "39";
     private static final String RETURN = "f3";
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/utils/BytecodeUtils.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/utils/BytecodeUtils.java
@@ -19,6 +19,7 @@ package com.hedera.mirror.web3.utils;
 import static com.hedera.mirror.web3.validation.HexValidator.HEX_PREFIX;
 
 import jakarta.annotation.Nonnull;
+import java.util.regex.Pattern;
 import lombok.experimental.UtilityClass;
 
 /**
@@ -39,6 +40,23 @@ public class BytecodeUtils {
     private static final String FREE_MEMORY_POINTER_2 = "60606040";
     private static final String RUNTIME_CODE_PREFIX =
             "6080"; // The pattern to find the start of the runtime code in the init bytecode
+
+    /**
+     * Compiled regex pattern to match the init bytecode sequence. The pattern checks for a sequence of a free memory
+     * pointer setup, a CODECOPY operation, and a RETURN operation, in that order. The sequence is matched
+     * case-insensitively to account for hexadecimal representations. Pattern explanation: - (?i) enables
+     * case-insensitive matching for the entire pattern. - (%s|%s) matches either FREE_MEMORY_POINTER or
+     * FREE_MEMORY_POINTER_2, which represent setup instructions for the free memory pointer, required for
+     * initialization bytecode. - .* matches any sequence of characters (including none) after the free memory pointer
+     * setup. - %s matches the CODECOPY opcode, which copies code to memory and typically follows the memory pointer
+     * setup in init bytecode. - .* matches any sequence of characters (including none) between CODECOPY and RETURN. %s
+     * matches the RETURN opcode, signaling the end of the initialization bytecode.
+     * <p>
+     * Example pattern: (?i)(60806040|60606040).*39.*f3 This example would match any sequence where either "60806040" or
+     * "60606040" appears, followed by "39" (CODECOPY) and then "f3" (RETURN), with any characters in between.
+     */
+    private static final Pattern INIT_BYTECODE_PATTERN = Pattern.compile(
+            String.format("(?i)(%s|%s).*%s.*%s", FREE_MEMORY_POINTER, FREE_MEMORY_POINTER_2, CODECOPY, RETURN));
 
     public static String extractRuntimeBytecode(String initBytecode) {
         // Check if the bytecode starts with "0x" and remove it if necessary
@@ -78,36 +96,10 @@ public class BytecodeUtils {
      * @return true if it is init bytecode, false otherwise.
      */
     public static boolean isInitBytecode(String data) {
-        if (data == null || data.isEmpty()) {
+        if (data == null || data.length() < MINIMUM_INIT_CODE_SIZE) {
             return false;
         }
 
-        String lowerCaseData = data.toLowerCase();
-
-        // Check if bytecode meets minimum length requirement
-        if (lowerCaseData.length() < MINIMUM_INIT_CODE_SIZE) {
-            return false;
-        }
-
-        // Check if (free memory pointer setup) exists
-        int freeMemoryPointerIndex = lowerCaseData.indexOf(FREE_MEMORY_POINTER);
-        if (freeMemoryPointerIndex == -1) {
-            freeMemoryPointerIndex = lowerCaseData.indexOf(FREE_MEMORY_POINTER_2);
-        }
-        if (freeMemoryPointerIndex == -1) {
-            return false;
-        }
-
-        // Verify CODECOPY occurs after free memory pointer setup
-        int codeCopyIndex = lowerCaseData.indexOf(CODECOPY, freeMemoryPointerIndex + FREE_MEMORY_POINTER.length());
-        if (codeCopyIndex == -1) {
-            return false;
-        }
-
-        // Check if RETURN occurs after CODECOPY
-        int returnIndex = lowerCaseData.indexOf(RETURN, codeCopyIndex + CODECOPY.length());
-
-        // If all conditions are met, this is likely init bytecode
-        return returnIndex != -1;
+        return INIT_BYTECODE_PATTERN.matcher(data).find();
     }
 }
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/utils/BytecodeUtils.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/utils/BytecodeUtils.java
@@ -33,6 +33,7 @@ import lombok.experimental.UtilityClass;
 @UtilityClass
 public class BytecodeUtils {
 
+    public static final String SKIP_INIT_CODE_CHECK = "HEDERA_MIRROR_WEB3_EVM_CHECKINITCODE";
     private static final String CODECOPY = "39";
     private static final String RETURN = "f3";
     private static final long MINIMUM_INIT_CODE_SIZE = 14L;
@@ -44,19 +45,24 @@ public class BytecodeUtils {
     /**
      * Compiled regex pattern to match the init bytecode sequence. The pattern checks for a sequence of a free memory
      * pointer setup, a CODECOPY operation, and a RETURN operation, in that order. The sequence is matched
-     * case-insensitively to account for hexadecimal representations. Pattern explanation: - (?i) enables
-     * case-insensitive matching for the entire pattern. - (%s|%s) matches either FREE_MEMORY_POINTER or
-     * FREE_MEMORY_POINTER_2, which represent setup instructions for the free memory pointer, required for
-     * initialization bytecode. - .* matches any sequence of characters (including none) after the free memory pointer
-     * setup. - %s matches the CODECOPY opcode, which copies code to memory and typically follows the memory pointer
-     * setup in init bytecode. - .* matches any sequence of characters (including none) between CODECOPY and RETURN. %s
-     * matches the RETURN opcode, signaling the end of the initialization bytecode.
+     * case-insensitively to account for hexadecimal representations.
      * <p>
-     * Example pattern: (?i)(60806040|60606040).*39.*f3 This example would match any sequence where either "60806040" or
-     * "60606040" appears, followed by "39" (CODECOPY) and then "f3" (RETURN), with any characters in between.
+     * Pattern explanation: - (%s|%s) matches either FREE_MEMORY_POINTER or FREE_MEMORY_POINTER_2, which represent setup
+     * instructions for the free memory pointer, required for initialization bytecode. - [0-9a-z]+ matches one or more
+     * valid hexadecimal characters (0-9, a-f) after the free memory pointer setup. - %s matches the CODECOPY opcode,
+     * which copies code to memory and typically follows the memory pointer setup in init bytecode. - [0-9a-z]+ matches
+     * one or more valid hexadecimal characters (0-9, a-f) between CODECOPY and RETURN. - %s matches the RETURN opcode,
+     * signaling the end of the initialization bytecode.
+     *
+     * <p>
+     * Example pattern: (60806040|60606040)[0-9a-z]+39[0-9a-z]+f3 This example would match any sequence where either
+     * "60806040" or "60606040" appears, followed by "39" (CODECOPY) and then "f3" (RETURN), with valid hexadecimal
+     * characters in between.
      */
     private static final Pattern INIT_BYTECODE_PATTERN = Pattern.compile(
-            String.format("(?i)(%s|%s).*%s.*%s", FREE_MEMORY_POINTER, FREE_MEMORY_POINTER_2, CODECOPY, RETURN));
+            String.format(
+                    "(%s|%s)[0-9a-z]+%s[0-9a-z]+%s", FREE_MEMORY_POINTER, FREE_MEMORY_POINTER_2, CODECOPY, RETURN),
+            Pattern.CASE_INSENSITIVE);
 
     public static String extractRuntimeBytecode(String initBytecode) {
         // Check if the bytecode starts with "0x" and remove it if necessary
@@ -95,11 +101,19 @@ public class BytecodeUtils {
      * @param data the data string to check.
      * @return true if it is init bytecode, false otherwise.
      */
-    public static boolean isInitBytecode(String data) {
+    public static boolean isInitBytecode(final String data) {
         if (data == null || data.length() < MINIMUM_INIT_CODE_SIZE) {
             return false;
         }
 
         return INIT_BYTECODE_PATTERN.matcher(data).find();
+    }
+
+    public static boolean isValidInitBytecode(final String data) {
+        return shouldSkipBytecodeCheck() || BytecodeUtils.isInitBytecode(data);
+    }
+
+    private static boolean shouldSkipBytecodeCheck() {
+        return Boolean.parseBoolean(System.getenv(SKIP_INIT_CODE_CHECK));
     }
 }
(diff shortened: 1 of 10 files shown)


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/viewmodel/ContractCallRequest.java
```diff
@@ -60,9 +59,4 @@ public class ContractCallRequest {
     private boolean hasFrom() {
         return value <= 0 || from != null;
     }
-
-    @AssertTrue(message = "must not be empty")
```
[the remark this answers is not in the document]
bilyana-gospodinova: I think it is safer if we move the old check in the `ContractController` and add additional check before that if the `data` contains an init bytecode (and if it does - skip the other `to` checks, if not - make the old checks for `value`, `from`, empty `to`). This way we will cover the scenario in the issue and cut the failing requests as soon as possible and we will avoid doing any unnecessary computations while preserving the old validations.
If we just remove all validations of the `to` field, the response will change a bit in some situations. For example: a contract call with `to == null`. The controller will not cut the request, the users would not get a message that the "to" field is missing, the receiver will be assigned to `Address.ZERO`. the `ContractExecutionService` will be called, some address lookups of the sender and receiver will be made. Since the receiver is empty, the call will be assumed to be a contract create case instead of a contract call and so on. As a result the returned error might be misleading.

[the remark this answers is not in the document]
kselveliev: I just added a new validation to check if data contains init bytecode. Can you check it

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/utils/RuntimeBytecodeExtractor.java
```diff
@@ -66,4 +67,22 @@ private static String getRuntimeBytecode(final String initBytecode) {
...
+        if (data == null || data.isEmpty()) {
+            return false;
+        }
+
+        // CODECOPY (0x39) and RETURN (0xf3)
+        boolean hasCodeCopy = data.contains(CODECOPY);
+        boolean hasReturn = data.contains(RETURN);
+
+        // Check if both CODECOPY and RETURN opcodes are present
+        return hasCodeCopy && hasReturn;
```
bilyana-gospodinova, line 86: I am concerned about this check. If we have a normal contract call with function selector such as: "0xf30039" this check will say that this is a contract create and clearly it is not. I think that the position of these opcodes is important. We need to research this further.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/viewmodel/ContractCallRequest.java
```diff
@@ -63,6 +64,6 @@ private boolean hasFrom() {
 
     @AssertTrue(message = "must not be empty")
     private boolean hasTo() {
-        return value <= 0 || from == null || StringUtils.isNotEmpty(to);
+        return RuntimeBytecodeExtractor.isInitBytecode(data) || StringUtils.isNotEmpty(to);
```
bilyana-gospodinova, line 67: Why do we remove the old check for `value <= 0 || from == null`? If there is a direct transfer, there needs to have `from`, `to` and `value` set. This was validated in the `ContractControllerTest#callInvalidToDueToTransfer` which is now removed but I think it shouldn't be.

bilyana-gospodinova: Also, this way we verify twice if the `to` field is empty - once in `RuntimeBytecodeExtractor.isInitBytecode` and a second time in `StringUtils.isNotEmpty(to)` but I think this validation needs to be changed anyway as explained in my other comments.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/controller/ContractControllerTest.java
```diff
@@ -118,14 +119,13 @@ private ResultActions contractCall(ContractCallRequest request) {
...
-    @ValueSource(strings = {"0x00000000000000000000000000000000000007e7"})
     @ParameterizedTest
-    void estimateGas(String to) throws Exception {
+    void estimateGasWithEmptyTo(String to) throws Exception {
         final var request = request();
         request.setEstimate(true);
         request.setValue(0);
         request.setTo(to);
-        contractCall(request).andExpect(status().isOk());
+        contractCall(request).andExpect(status().isBadRequest());
```
bilyana-gospodinova, line 128: Changing this test indicates there is a modification in the previous logic that was unexpected.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/utils/RuntimeBytecodeExtractor.java
kselveliev: Yes this is correct. I see that the minimum size of the init bytecode for a valid contract deployment is around 14 bytes so I will add this in the check.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/controller/ContractControllerTest.java
kselveliev: I don't think so. Previous version of the test did:

data: randomFunctionSelector(not init code)
to: empty
from: address

and expect success - which is not correct in my opinion.
As far as I understand the only time we can allow the "to" to be empty is when we do contract deployment.

bilyana-gospodinova: Correct - the only time when "to" can be empty is on contract deploy. So, we need 2 parameterized tests:
1.  One in which the data is set as a correct init bytecode and the old check for success or we wouldn't have a test to verify the positive path for estimate gas.
2. Second parameterized test without "to" and non init bytecode data (but contract call) which is expected to return bad request response code.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/viewmodel/ContractCallRequest.java
kselveliev: If we are looking only at the ‘to’ validation

As far as I know we can only allow empty ‘to’ when we do contract deployment
meaning: 
```
to: empty 
data: init code for contract deployment
```


In the case where we have direct transfer as you said we want to have 

```
from: address (not empty)
to: address (not empty)
value: some value set
```

from is already handled by its own validation 
to - it will verify that the address is not init bytecode and then it will just do the not empty check for 'to'
value - already has a positiveOrZero validation 

so this scenario should be covered. So this old check only confuses the reader `value <= 0 || from == null` - thats why I removed it.

Maybe I’m missing a corner case.

bilyana-gospodinova: Maybe you are right.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/controller/ContractControllerTest.java
kselveliev: Comment addressed I added a couple more estimate gas tests.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/utils/RuntimeBytecodeExtractor.java
kselveliev: Added one more condition to check for min size of the data

bilyana-gospodinova: I still don't think this check is enough. Example: "getCustomFeesForToken(address)" -> function selector is: "44**f3**8bc8". And if you pass the address of token "00..0**39**", the data would look like: "0x44f38bc8000...039" and this method will identify it as an init bytecode.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/viewmodel/ContractCallRequest.java
IvanKavaldzhiev, line 67: I think removing the previous check will require `to` to be set even if `from` is `null` and value is <=0. I think, we should add additional validation on top of the current one.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/utils/RuntimeBytecodeExtractor.java
```diff
@@ -24,15 +24,17 @@
...
+ * Smart contracts have init bytecode (constructor bytecode) and runtime bytecode (the code executed when the contract
+ * is called). This class helps in extracting the runtime bytecode from the given init bytecode by searching for
+ * specific patterns.
  * </p>
  */
 @UtilityClass
 public class RuntimeBytecodeExtractor {
 
     private static final String CODECOPY = "39";
+    private static final String RETURN = "f3";
```
bilyana-gospodinova, line 36: What if the return code is in uppercase ("F3")? The check will be bypassed.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/utils/RuntimeBytecodeExtractor.java
```diff
@@ -53,17 +55,37 @@ private static String getRuntimeBytecode(final String initBytecode) {
...
+    /**
+     * Checks if a given data string is likely init bytecode.
+     *
+     * @param data the data string to check.
+     * @return true if it is init bytecode, false otherwise.
+     */
+    public static boolean isInitBytecode(final String data) {
+        if (data == null || data.isEmpty()) {
+            return false;
+        }
```
IvanKavaldzhiev, line 81: We can add additional logic checking if the data starts with `606060` or `608060` sequence, as this is the current format, where PUSH1 operation is inserting a free memory pointer. In this way we can fail fast, if the `data` doesn't even begin with this sequence. I believe in a lot of cases this check might be very efficient and enough to distinguish between a `bytecode` and `calldata`.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/viewmodel/ContractCallRequest.java
kselveliev: Comment addressed. I added the new logic on top of the current check.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/utils/RuntimeBytecodeExtractor.java
kselveliev: Correct, I fixed that

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/utils/RuntimeBytecodeExtractor.java
kselveliev: This is a great suggestion -> however doing that there was another corner case here you can observe one of the autogenerated solidity classes EthCall.sol init code starts with 6104d25. I'm not sure what this case is but perhaps we should handle it as well. I'm now thinking if there can be other cases like that which we might miss.

bilyana-gospodinova: I found [a great article](https://medium.com/coinmonks/evm-part-ii-the-journey-of-smart-contracts-from-solidity-code-to-bytecode-a21300b23dee) that explains the structure of the init bytecode. It seems that in each init bytecode there is the sequence "60806040" which is responsible for assigning a free memory pointer. Before that there could be some other instructions for constructing the contract. In your example for `EthCall.sol` the bytecode starts with "6104d2" and this means: "PUSH2 0x4D2", which is "push 1234" which is one of the fields in the contract. And also in each contract there is a `codecopy` instruction afterwards followed by the runtime bytecode. I think this method can be enhanced with these findings and also the position of the opcodes is important. For example - to contain "f3" for the `return` opcode after the `codecopy` and not anywhere in the string. The `codecopy` should be after the free memory pointer "60806040" and so on.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/controller/ContractControllerTest.java
```diff
@@ -79,6 +88,7 @@ class ContractControllerTest {
     private static final String CALL_URI = "/api/v1/contracts/call";
     private static final String ONE_BYTE_HEX = "80";
     private static final long THROTTLE_GAS_LIMIT = 10_000_000L;
+    private static final String INIT_CODE = "0x60006000556000526000526000f3";
```
bilyana-gospodinova, line 91: Perhaps it will be better to add a valid example of an init bytecode that is in compliance with the latest changes to avoid confusion.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/utils/RuntimeBytecodeExtractor.java
```diff
@@ -24,15 +24,19 @@
...
- * Smart contracts have init bytecode (constructor bytecode) and runtime bytecode (the code
- * executed when the contract is called). This class helps in extracting the runtime bytecode from the
- * given init bytecode by searching for specific patterns.
+ * Smart contracts have init bytecode (constructor bytecode) and runtime bytecode (the code executed when the contract
+ * is called). This class helps in extracting the runtime bytecode from the given init bytecode by searching for
+ * specific patterns.
  * </p>
  */
 @UtilityClass
 public class RuntimeBytecodeExtractor {
```
bilyana-gospodinova, line 33: This class should be renamed as now it does not only extract runtime bytecode.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/utils/RuntimeBytecodeExtractor.java
bilyana-gospodinova: Thanks. Looks good now.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/utils/RuntimeBytecodeExtractor.java
kselveliev: Renamed to BytecodeUtils

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/controller/ContractControllerTest.java
kselveliev: Comment addressed I added a valid init code of and empty contract

bilyana-gospodinova (dismissed):
LGTM

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/utils/BytecodeUtils.java
```diff
@@ -53,17 +57,57 @@ private static String getRuntimeBytecode(final String initBytecode) {
...
+     */
+    public static boolean isInitBytecode(String data) {
+        if (data == null || data.isEmpty()) {
+            return false;
+        }
+
+        String lowerCaseData = data.toLowerCase();
+
+        // Check if bytecode meets minimum length requirement
+        if (lowerCaseData.length() < MINIMUM_INIT_CODE_SIZE) {
```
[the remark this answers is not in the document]
kselveliev: With the new pattern I added this lowercase checking is removed and added in the pattern.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/utils/BytecodeUtils.java
```diff
@@ -53,17 +57,57 @@ private static String getRuntimeBytecode(final String initBytecode) {
...
+        int codeCopyIndex = lowerCaseData.indexOf(CODECOPY, freeMemoryPointerIndex + FREE_MEMORY_POINTER.length());
+        if (codeCopyIndex == -1) {
+            return false;
+        }
+
+        // Check if RETURN occurs after CODECOPY
+        int returnIndex = lowerCaseData.indexOf(RETURN, codeCopyIndex + CODECOPY.length());
+
+        // If all conditions are met, this is likely init bytecode
+        return returnIndex != -1;
```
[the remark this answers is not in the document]
kselveliev: Comment addressed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/utils/BytecodeUtils.java
```diff
@@ -53,17 +57,57 @@ private static String getRuntimeBytecode(final String initBytecode) {
...
         return initBytecode.substring(runtimeCodePrefixIndex);
     }
+
+    /**
+     * Checks if a given data string is likely init bytecode.
+     *
+     * @param data the data string to check.
+     * @return true if it is init bytecode, false otherwise.
+     */
+    public static boolean isInitBytecode(String data) {
```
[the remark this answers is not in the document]
kselveliev: If not in properties you mean like having this in our code right? ->  
```
private static final boolean CHECK_INIT_CODE_ENABLED = 
        Boolean.parseBoolean(System.getenv("HEDERA_MIRROR_WEB3_EVM_CHECKINITCODE"));
```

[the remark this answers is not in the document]
steven-sheehy: Correct

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/utils/BytecodeUtils.java
```diff
@@ -0,0 +1,105 @@
...
+     * initialization bytecode. - .* matches any sequence of characters (including none) after the free memory pointer
+     * setup. - %s matches the CODECOPY opcode, which copies code to memory and typically follows the memory pointer
+     * setup in init bytecode. - .* matches any sequence of characters (including none) between CODECOPY and RETURN. %s
+     * matches the RETURN opcode, signaling the end of the initialization bytecode.
+     * <p>
+     * Example pattern: (?i)(60806040|60606040).*39.*f3 This example would match any sequence where either "60806040" or
+     * "60606040" appears, followed by "39" (CODECOPY) and then "f3" (RETURN), with any characters in between.
+     */
+    private static final Pattern INIT_BYTECODE_PATTERN = Pattern.compile(
+            String.format("(?i)(%s|%s).*%s.*%s", FREE_MEMORY_POINTER, FREE_MEMORY_POINTER_2, CODECOPY, RETURN));
```
steven-sheehy, line 59: Can pass Pattern.CASE_INSENSITIVE as a parameter to compile.

steven-sheehy, line 59: Believe we should have anchoring `^...$` to ensure strings start and end at beginning. Add test.

steven-sheehy, line 59: Instead of .* we can use \w* since it should be valid hex. BTW, are you sure it shouldn't be + instead of *?

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/utils/BytecodeUtils.java
kselveliev: Don’t think we can achor we have some scenarios like EthCall.java binary 

where we have some characters before the begin 60806040 pattern → 6104d25…. 60806040

And for the ending achor we pretty much in all cases have f3… and then followed by some extra characters

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/utils/BytecodeUtils.java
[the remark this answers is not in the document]
kselveliev: Added the env variable similarly to -> HEDERA_MIRROR_IMPORTER_PARSER_HALTONERROR.
Env variable - HEDERA_MIRROR_WEB3_EVM_SKIPINITCODECHECK
Currently it defaults to false. If false it does the init bytecode check.
If set to true it skips the init bytecode check and the 'to" check works similarly as before

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/utils/BytecodeUtils.java
kselveliev: Comment addressed

kselveliev: Changed to [0-9a-z]*. It think + is better than * so changed it to +.

I experimented with getting the minimum initialization bytecode for an empty contract but could never achieve it with zero characters between the pattern variables.

steven-sheehy (approved):
LGTM

PR #9673, opened 2024-10-29
Add network info subcomponent needed for the state

### LINKED ISSUE #9260: Add needed subcomponents for State initialization
### Problem

In order to properly initialize the State in `hedera-mirror-web3`, we should define several components, so that we can migrate the schema definitions and initialize the State with the needed structure.

### Solution

Add implementation for the following components:
- NetworkInfo
- SchemaRegistry
- ServiceMigrator

Add unit tests for these components.

### Alternatives

_No response_

kselveliev:
**Description**:
This PR add NetworkInfo component needed for the state of the reusable services code

Add NetworkInfoImpl class similar to services
Add NetworkInfoImplTest

**Related issue(s)**:

Partially closes #9260 
Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/components/NetworkInfoImpl.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/components/NetworkInfoImpl.java
@@ -26,6 +26,7 @@ import com.swirlds.state.spi.info.SelfNodeInfo;
 import jakarta.annotation.Nonnull;
 import jakarta.annotation.Nullable;
 import jakarta.inject.Named;
+import java.util.Collections;
 import java.util.List;
 import lombok.RequiredArgsConstructor;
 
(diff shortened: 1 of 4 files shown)


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/components/NetworkInfoImpl.java
```diff
@@ -0,0 +1,155 @@
...
+import lombok.RequiredArgsConstructor;
+
+@Named
+@RequiredArgsConstructor
+public class NetworkInfoImpl implements NetworkInfo {
+
+    private final MirrorNodeEvmProperties mirrorNodeEvmProperties;
+
+    private static final Bytes DEV_LEDGER_ID = Bytes.wrap(new byte[] {0x03});
+    private static final List<NodeInfo> FAKE_NODE_INFOS = List.of(
```
IvanKavaldzhiev, line 39: We don't need so much fake elements. Won't an empty list do the job?

kselveliev: I can fix to return an empty list for now. I'm still not really sure how this node ids info will be used later.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/components/NetworkInfoImpl.java
```diff
@@ -0,0 +1,89 @@
...
+
+@Named
+@RequiredArgsConstructor
+public class NetworkInfoImpl implements NetworkInfo {
+
+    private final MirrorNodeEvmProperties mirrorNodeEvmProperties;
+
+    @Nonnull
+    @Override
+    public Bytes ledgerId() {
```
steven-sheehy, line 41: Are we sure that ledgerId and addressBook() aren't used by EVM calls? We have the ledgerId in the properties we could return it.

If they're not used, maybe we should throw exceptions instead of return empty objects so we can catch problems earlier in testing?

kselveliev: We can't really be sure if they will be used or not until we start testing executing transaction.
Throwing exceptions for now for these methods seems pretty reasonable so I applied it. 

And once we start testing if we hit those exceptions then we'll implement them.

steven-sheehy (approved):

Issue #9259, opened 2024-09-03
Create `State` implementation with registering all of the specific `ReadableKVStates`


IvanKavaldzhiev:
### Problem

One of the main efforts to integrate the reusable services with the `hedera-mirror-web3` code base will be having a State implementation. We should register different `ReadableKVState` instances for each specific entity type that will be kept in state. This will be the linking point between the internal State logic inside reusable services and the Postgre DB.

### Solution

Implement the State interface from `hedera.app` with light-weight logic, registering a specific `ReadableKVState` type for each register entity type in the Schemas that come from `hedera.app`. This registering will be inside `getReadableStates()` method.

E.g. 

For `FILES_KEY` schema type register a `FileReadableKVState`

Provide tests for the State reading all type of entities, check corner cases and historical scenarios.

PR #9700, opened 2024-11-04
Create `State` implementation with registering all of the specific `ReadableKVStates`

bilyana-gospodinova:
**Description**:
This PR adds a `MirrorNodeState` class that implements the `State` interface. The class registers all of needed readable and writable states. Each service (`TokenService`, `ContractService` or `FileService`) needs a different set of states in order to function properly. The readable states are initialised only once and are singleton instances and will be used by all of the threads. They are the implementations of all of the `ReadableKVState`s. For the writable states we need only 1 facade implementation since no actual changes will be persisted from the memory. The writable state uses a readable state for the read operations and the modifications are kept in memory and discarded at the end of the transaction execution.

Fixes https://github.com/hashgraph/hedera-mirror-node/issues/9259

### DIFF
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/MirrorNodeState.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/MirrorNodeState.java
@@ -16,91 +16,98 @@
 
 package com.hedera.mirror.web3.state;
 
+import static java.util.Objects.requireNonNull;
+
+import com.hedera.mirror.web3.state.utils.MapReadableKVState;
 import com.hedera.mirror.web3.state.utils.MapReadableStates;
 import com.hedera.mirror.web3.state.utils.MapWritableKVState;
 import com.hedera.mirror.web3.state.utils.MapWritableStates;
-import com.hedera.node.app.service.contract.ContractService;
-import com.hedera.node.app.service.file.FileService;
-import com.hedera.node.app.service.token.TokenService;
 import com.swirlds.state.State;
-import com.swirlds.state.spi.ReadableKVState;
+import com.swirlds.state.spi.EmptyReadableStates;
+import com.swirlds.state.spi.EmptyWritableStates;
 import com.swirlds.state.spi.ReadableStates;
 import com.swirlds.state.spi.WritableStates;
+import edu.umd.cs.findbugs.annotations.NonNull;
 import jakarta.annotation.Nonnull;
 import jakarta.inject.Named;
-import java.util.Collections;
-import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
+@SuppressWarnings({"rawtypes", "unchecked"})
 @Named
 public class MirrorNodeState implements State {
-    private final Map<String, ReadableKVState<?, ?>> tokenReadableServiceStates = new HashMap<>();
-    private final Map<String, ReadableKVState<?, ?>> contractReadableServiceStates = new HashMap<>();
-    private final Map<String, ReadableKVState<?, ?>> fileReadableServiceStates = new HashMap<>();
 
     private final Map<String, ReadableStates> readableStates = new ConcurrentHashMap<>();
+    // Key is Service, value is Map of state name to state datasource
+    private final Map<String, Map<String, Object>> states = new ConcurrentHashMap<>();
 
-    public MirrorNodeState(
-            final AccountReadableKVState accountReadableKVState,
-            final AirdropsReadableKVState airdropsReadableKVState,
-            final AliasesReadableKVState aliasesReadableKVState,
-            final ContractBytecodeReadableKVState contractBytecodeReadableKVState,
-            final ContractStorageReadableKVState contractStorageReadableKVState,
-            final FileReadableKVState fileReadableKVState,
-            final NftReadableKVState nftReadableKVState,
-            final TokenReadableKVState tokenReadableKVState,
-            final TokenRelationshipReadableKVState tokenRelationshipReadableKVState) {
+    public MirrorNodeState addService(@NonNull final String serviceName, @NonNull final Map<String, ?> dataSources) {
+        final var serviceStates = this.states.computeIfAbsent(serviceName, k -> new ConcurrentHashMap<>());
+        dataSources.forEach((k, b) -> {
+            if (!serviceStates.containsKey(k)) {
+                serviceStates.put(k, b);
+            }
+        });
 
-        tokenReadableServiceStates.put("ACCOUNTS", accountReadableKVState);
-        tokenReadableServiceStates.put("PENDING_AIRDROPS", airdropsReadableKVState);
-        tokenReadableServiceStates.put("ALIASES", aliasesReadableKVState);
-        tokenReadableServiceStates.put("NFTS", nftReadableKVState);
-        tokenReadableServiceStates.put("TOKENS", tokenReadableKVState);
-        tokenReadableServiceStates.put("TOKEN_RELS", tokenRelationshipReadableKVState);
+        // Purge any readable states whose state definitions are now stale,
+        // since they don't include the new data sources we just added
+        readableStates.remove(serviceName);
+        return this;
+    }
 
-        contractReadableServiceStates.put("BYTECODE", contractBytecodeReadableKVState);
-        contractReadableServiceStates.put("STORAGE", contractStorageReadableKVState);
-
-        fileReadableServiceStates.put("FILES", fileReadableKVState);
+    /**
+     * Removes the state with the given key for the service with the given name.
+     *
+     * @param serviceName the name of the service
+     * @param stateKey the key of the state
+     */
+    public void removeServiceState(@NonNull final String serviceName, @NonNull final String stateKey) {
+        requireNonNull(serviceName);
+        requireNonNull(stateKey);
+        this.states.computeIfPresent(serviceName, (k, v) -> {
+            v.remove(stateKey);
+            readableStates.remove(serviceName); // Remove the service so that its states will be repopulated.
+            return v;
+        });
     }
 
     @Nonnull
     @Override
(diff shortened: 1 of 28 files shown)


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/utils/MapWritableKVState.java
```diff
@@ -0,0 +1,86 @@
...
+        return Collections.emptyIterator();
+    }
+
+    @Override
+    public void commit() {
+        reset();
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
kselveliev, line 74: Would we need to compare different writable and readable states in our use case. Because I don't see them implemented in services classes

bilyana-gospodinova: These methods were needed in the junit tests.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/utils/ListReadableQueueState.java
```diff
@@ -0,0 +1,56 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hedera.mirror.web3.state.utils;
+
+import com.swirlds.state.spi.ReadableQueueStateBase;
+import edu.umd.cs.findbugs.annotations.NonNull;
```
kselveliev, line 20: Needs jakarta annotations

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/MirrorNodeState.java
```diff
@@ -0,0 +1,268 @@
...
+public class MirrorNodeState implements State {
+
+    private final Map<String, ReadableStates> readableStates = new ConcurrentHashMap<>();
+    private final Map<String, WritableStates> writableStates = new ConcurrentHashMap<>();
+
+    // Key is Service, value is Map of state name to state datasource
+    private final Map<String, Map<String, Object>> states = new ConcurrentHashMap<>();
+    private final List<StateChangeListener> listeners = new ArrayList<>();
+
+    public MirrorNodeState addService(@NonNull final String serviceName, @NonNull final Map<String, ?> dataSources) {
```
kselveliev, line 65: Since now everything is dynamic and from what we've seen in testing maybe we can add a method that populates our custom kv states similar to the initial commit in pr -> 

```
        tokenReadableServiceStates.put("ACCOUNTS", accountReadableKVState);
        tokenReadableServiceStates.put("ALIASES", aliasesReadableKVState);
        tokenReadableServiceStates.put("NFTS", nftReadableKVState);
        tokenReadableServiceStates.put("TOKENS", tokenReadableKVState);
```
etc.

bilyana-gospodinova: I think this needs to be done when we build the state and all its subcomponents. This will happen only once after all. This suggestion might provide more benefits if we needed to build the state multiple times.

IvanKavaldzhiev (dismissed):
At first glance the PR looks reasonable and have all of the needed state core components. The state should be utilized via integration tests (added in separate PR/PRs) to validate the correctness and consistency of the state. A lot of pitfalls or little issues can be missed without running actual transactions against this state. So I think this PR can be merged in this form and if it needs some touch ups - they can be made in separate PRs.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/utils/AbstractMapReadableState.java
```diff
@@ -0,0 +1,44 @@
...
+package com.hedera.mirror.web3.state.utils;
+
+import com.swirlds.state.spi.ReadableStates;
+import jakarta.annotation.Nonnull;
+import java.util.Collections;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+
+abstract class AbstractMapReadableState implements ReadableStates {
```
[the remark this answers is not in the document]
bilyana-gospodinova: Moved to "core" package as suggested.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/utils/ListWritableQueueState.java
```diff
@@ -0,0 +1,57 @@
...
+
+package com.hedera.mirror.web3.state.utils;
+
+import com.swirlds.state.spi.WritableQueueStateBase;
+import jakarta.annotation.Nonnull;
+import java.util.Iterator;
+import java.util.Objects;
+import java.util.Queue;
+
+public class ListWritableQueueState<E> extends WritableQueueStateBase<E> {
```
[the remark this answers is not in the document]
bilyana-gospodinova: Moved to "core" package as suggested.

kselveliev (approved):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/core/MapWritableStates.java
```diff
@@ -0,0 +1,125 @@
...
+        return Objects.hash(states);
+    }
+
+    @Override
+    public void commit() {
+        states.values().forEach(state -> {
+            switch (state) {
+                case WritableKVStateBase kv -> kv.commit();
+                case WritableSingletonStateBase singleton -> singleton.commit();
+                case WritableQueueStateBase queue -> queue.commit();
```
steven-sheehy, line 116: We should ask upstream to add a common interface that contains commit() to these three so we don't break when they add a fourth type. Please open a ticket and discuss with them.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/MirrorNodeState.java
```diff
@@ -0,0 +1,268 @@
...
+        return Objects.equals(readableStates, that.readableStates)
+                && Objects.equals(writableStates, that.writableStates)
+                && Objects.equals(states, that.states)
+                && Objects.equals(listeners, that.listeners);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(readableStates, writableStates, states, listeners);
+    }
```
steven-sheehy, line 262: Why do we need to implement equals and hash code? Is it put into a map as a key?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/MirrorNodeState.java
```diff
@@ -0,0 +1,268 @@
...
+                            withAnyRegisteredListeners(serviceName, new ListWritableQueueState<>(stateName, queue)));
+                } else if (state instanceof Map<?, ?> map) {
+                    data.put(
+                            stateName,
+                            withAnyRegisteredListeners(serviceName, new MapWritableKVState<>(stateName, map)));
+                } else if (state instanceof AtomicReference<?> ref) {
+                    data.put(stateName, withAnyRegisteredListeners(serviceName, stateName, ref));
+                }
+            }
+            return new MapWritableStates(data, () -> readableStates.remove(serviceName));
```
steven-sheehy, line 147: Will this cause a commit to flush the readable state and cause it to get re-created for every request?

steven-sheehy (commented):
Mostly just questions

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/core/MapWritableStates.java
bilyana-gospodinova: I asked in the slack channel. I will update when I have more information.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/MirrorNodeState.java
bilyana-gospodinova: No, it is not but it is easier this way to compare states in the tests.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/core/MapReadableKVState.java
```diff
@@ -0,0 +1,81 @@
...
+import com.swirlds.state.spi.ReadableKVStateBase;
+import jakarta.annotation.Nonnull;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * A simple implementation of {@link ReadableKVState} backed by a
+ * {@link Map}. Test code has the option of creating an instance disregarding the backing map, or by
+ * supplying the backing map to use. This latter option is useful if you want to use Mockito to spy
```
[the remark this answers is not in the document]
bilyana-gospodinova: It seems that code was updated but the comment was not. I will update it, thanks.

steven-sheehy (approved):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/core/MapWritableStates.java
bilyana-gospodinova: It seems that it was intentionally written this way. I will forward you the full explanation in Slack.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/MirrorNodeState.java
bilyana-gospodinova: Yes, it will. I tested it without the runnable and it seems to work fine for 1 request. This was added in the example code probably for a reason. I will remove it for now and we will investigate the behaviour on parallel requests later.

PR #9708, opened 2024-11-04
Use Alias When Available and No sender_id On Contract Results Response

### LINKED ISSUE #9667: Returned `from` address on `/api/v1/contracts/results?timestamp...` is null for `INVALID_ACCOUNT_ID` reverted transactions
### Description

Each Ethereum transaction should have a valid `from` value, no matter if it's regarding specific a hedera validation logic or evm reversion. Currently, there is a specific case in which we have a reverted transaction due to hedera validations and missing `from` address in the `/contracts/results?timestamp...` endpoint.

### Steps to reproduce

Send a transaction to a relay that has an invalid operator account. This is a rare case but it happened on mainnet so we should take care of it. I reproduced it on testnet:
- take a look at the nullable `from` - https://testnet.mirrornode.hedera.com/api/v1/contracts/results?timestamp=gte:1714043932.450373578&timestamp=lte:1714043933.966640830&limit=2&order=asc
- also transaction appears in hashscan https://hashscan.io/testnet/transaction/1714043933.763109003

The way to reproduce it is quite long but you should step in any step and do unit testing in each hedera module:
- run a local relay with an invalid operator id
- use hardhat (or any other tool) to send a transaction (simple hbar transfer is enough) to the local relay pointed to any network (e.g. testnet)
- the transaction is expected to be reverted
- investigate the transaction on hashscan and `api/v1/contracts/results?timestamp...` mirror node REST endpoint - on both places `from` address is null but it shouldn't be
- once the transaction is sent to the services, IMO it hits this hedera-specific check https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/TransactionProcessor.java#L233
- at this point, the services exports record file (because it's not a precheck) - here is the 1st place we should check if the exported record file contains valid `from`
- the record file is being imported by the mirror node importer - here is the 2nd place we should check if the record file for that type of reversion is being imported successfully and `from` value is available in the database
- the `api/v1/contracts/results?timestamp...` endpoint returns nullable from - here is the 3rd place we should check if the actual signer's address exists somewhere in the mirror node db
- at the end, we should be able to get the proper `from` value (derived from the signer's signature) on the `api/v1/contracts/results?timestamp...` endpoint

### Additional context

_No response_

### Hedera network

_No response_

### Version

testnet

### Operating system

None

jnels124:
**Description**:

**Related issue(s)**:

Fixes #9667

**Notes for reviewer**:

**Checklist**


hedera-mirror-rest/viewmodel/contractResultViewModel.js
```diff
@@ -47,6 +48,19 @@ class ContractResultViewModel {
     this.to = contractId.toEvmAddress();
     this.hash = toHexString(contractResult.transactionHash, true);
   }
+
+  #extractSenderFromFunctionResult(contractResult) {
+    if (!contractResult.sender_id && contractResult.functionResult) {
+      try {
+        const functionResult = proto.ContractFunctionResult.decode(contractResult.functionResult);
+        return functionResult?.senderId?.alias?.length ? toHexString(Array.from(functionResult.senderId.alias), true) : null;
```
jnels124, line 56: Please double check the UintArray conversion here...

steven-sheehy (dismissed):
LGTM

hedera-mirror-rest/viewmodel/contractResultViewModel.js
xin-hedera, line 56: the conversion sounds unnecessary, `functionResult.senderId.alias` is a nodejs `Buffer` which extends `Uint8Array`

xin-hedera (approved):
LGTM

steven-sheehy (approved):

PR #9711, opened 2024-11-05
Set ZFS ARC max/min before importing zpool

### LINKED ISSUE #9629: ZFS ARC max isn't honored
### Description

In our production environments with existing large disks and large zpools, the set value of 2GB ZFS ARC max isn't honored, instead, the effective max was likely auto set to around 50% of the available system memory, e.g.,

```
ARC status:                                                      HEALTHY
        Memory throttle count:                                         0

ARC size (current):                                    60.5 %   28.5 GiB
        Target size (adaptive):                        61.2 %   28.9 GiB
        Min size (hard limit):                          6.2 %    2.9 GiB
        Max size (high water):                           16:1   47.1 GiB
```

This would then lead to much higher ARC memory usage than expected, e.g., 28.5GB as shown above.

The theory is in our init script, we do the two things in order

- zpool import existing pool
- set ZFS ARC max to 2GB

In case of a very large existing pool, ZFS will auto calculate and set ARC min/max. When the effective ARC min (here it's 2.9GB) is higher than the ARC max in subsequent attempted changes, the change is just ignored.

We also need to test that with large pool, we don't run into the issues [here](https://github.com/openzfs/zfs/issues/16322) and [here](https://github.com/openzfs/zfs/issues/16325). 

### Steps to reproduce

check the description

### Additional context

_No response_

### Hedera network

_No response_

### Version

v0.0.116-rc3

### Operating system

None

xin-hedera:
**Description**:

This PR fixes the ZFS ARC max issue

- Set ZFS ARC max/min before importing zpool

**Related issue(s)**:

Fixes #9629

**Notes for reviewer**:

Tested with mainnet disk / pool size, arc_summary with the fix

```
ARC status:                                                      HEALTHY
        Memory throttle count:                                         0

ARC size (current):                                     1.2 %   23.6 MiB
        Target size (adaptive):                       100.0 %    2.0 GiB
        Min size (hard limit):                         25.0 %  512.0 MiB
        Max size (high water):                            4:1    2.0 GiB
        Most Frequently Used (MFU) cache size:          2.9 %  639.0 KiB
        Most Recently Used (MRU) cache size:           97.1 %   21.0 MiB
        Metadata cache size (hard limit):              75.0 %    1.5 GiB
        Metadata cache size (current):                  1.5 %   23.6 MiB
        Dnode cache size (hard limit):                 10.0 %  153.6 MiB
        Dnode cache size (current):                     0.6 %  981.3 KiB
```

Note the value of ARC min isn't important, 1/4 is kind of a randomly picked ratio. However we want to set it before setting max, since the new max value will get ignored if it's smaller than the current min value.

**Checklist**

### DIFF
--- a/charts/hedera-mirror-common/templates/zfs/configmap-init.yaml
+++ b/charts/hedera-mirror-common/templates/zfs/configmap-init.yaml
@@ -194,6 +194,13 @@ data:
       echo "Configuring zpool ${POOL}"
       partprobe
 
+      ARC_SIZE_GB="${ARC_SIZE_GB:-2}"
+      echo "Configuring arc max to ${ARC_SIZE_GB}GB and arc min to 1/4 of max"
+      ARC_MAX="$((ARC_SIZE_GB*1073741824))"
+      ARC_MIN="$((ARC_MAX/4))"
+      echo $ARC_MIN >> /sys/module/zfs/parameters/zfs_arc_min
+      echo $ARC_MAX >> /sys/module/zfs/parameters/zfs_arc_max
+
       if (zfs list | grep -q "${POOL}"); then
           echo "Found existing pool. Skipping creation"
       elif (zpool create -o autoexpand=on "${POOL}" /dev/sdb); then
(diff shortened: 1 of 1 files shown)


steven-sheehy (approved):
LGTM

jnels124 (approved):
LGTM

PR #9716, opened 2024-11-07
HIP-1056 Add block streams design

### LINKED ISSUE #9573: HIP-1056: Design block stream support
### Problem

We need to design how we can integrate block streams.

### Solution

* Add a design doc for block stream support (not block nodes)
* Create sub-tasks

### Alternatives

_No response_

edwin-greene:
**Description**:
Adds the design for phase one, support for block stream files.

**Related issue(s)**:

Fixes #9573 

**Notes for reviewer**:

**Checklist**


docs/design/block-streams.md
```diff
@@ -0,0 +1,142 @@
...
+
+- Add a new downloader, `BlockStreamDownloader extends Downloader<BlockFile, BlockItem>`, that will download block stream files from an S3/GCP bucket. This downloader will poll the bucket for the next numbered block file and download it. It should not use any list operations.
+- Add a new interface, `BlockFileReader extends StreamFileReader<BlockFile, BlockItem>` that will read block files.
+- Add a `ProtoBlockFileReader implements BlockFileReader` that will generate a `BlockFile` from a `StreamFileData`.
+- Add a `BlockStreamTranslator` that will translate a block file into a record file according to the mapping below.
+
+### Block protobuf to mirror node database mapping
+
+#### Contract Create Transaction
+
```
edwin-greene, line 37: More details on the missing entries still to come.

docs/design/block-streams.md
```diff
@@ -0,0 +1,142 @@
...
+### Importer
+
+- Add a new downloader, `BlockStreamDownloader extends Downloader<BlockFile, BlockItem>`, that will download block stream files from an S3/GCP bucket. This downloader will poll the bucket for the next numbered block file and download it. It should not use any list operations.
+- Add a new interface, `BlockFileReader extends StreamFileReader<BlockFile, BlockItem>` that will read block files.
+- Add a `ProtoBlockFileReader implements BlockFileReader` that will generate a `BlockFile` from a `StreamFileData`.
+- Add a `BlockStreamTranslator` that will translate a block file into a record file according to the mapping below.
+
+### Block protobuf to mirror node database mapping
+
```
edwin-greene, line 35: I've created a prototype of a block to record file translator for the contract create transaction. If we'd like to include it I can push it up as well.

docs/design/block-streams.md
```diff
@@ -0,0 +1,327 @@
...
+     */
+    private void verifyHashChain(String expected, String actual);
+
+    // Verifies that the number of the block file contained in its file name matches the block number within the block file
+    private void verifyBlockNumber(String expected, String actual);
+}
+```
+
+### Database
+
```
edwin-greene, line 165: Specific sql schema still to come

docs/design/block-streams.md
```diff
@@ -0,0 +1,381 @@
...
```
edwin-greene, line 169: Currently it is expected that there is a single `round` per `block`. The `round` has a `round_number` value that currently equals the block number. In the future multiple rounds could be included in a single block. I am not sure if we want to capture this somewhere.

docs/design/block-streams.md
```diff
@@ -0,0 +1,381 @@
...
+Add `software_version` to the `record_file` table. This is the consensus node version that generated the block.
+Add `congestionPricingMultiplier` to the `record_file` table.
+Rename `record_file` to `block_file`.
```
steven-sheehy, line 172: Show the sql for these so we can agree on their data types

docs/design/block-streams.md
steven-sheehy: Rounds are sequential right? If so, can future proof by adding round_start, round_end columns.

docs/design/block-streams.md
```diff
@@ -0,0 +1,381 @@
...
+#### StreamFileNotifier
+
+-Rename `verified` to `notify` as blocks will not be verifiable until each state change has been processed by the BlockStreamVerifier.
+
+#### BlockStreamVerifier
+
+```java
+package com.hedera.mirror.importer.downloader.block;
+
+public class BlockStreamVerifier implements StreamFileNotifier, Closable {
```
steven-sheehy, line 147: Closable is misspelled. But why is it needed? This should be a singleton bean.

docs/design/block-streams.md
```diff
@@ -0,0 +1,381 @@
...
+| load_start         | System.currentTimeMillis() at beginning of parsing                                                                                  |
+| load_end           | System.currentTimeMillis() at end of parsing                                                                                        |
+| logs_bloom         | Aggregate calculated by mirror node                                                                                                 |
+| name               | Name of the blk file                                                                                                                |
+| node_id            |                                                                                                                                     |
+| prev_hash          | block_header.previous_block_hash                                                                                                    |
+| sidecar_count      | Count of all transaction*output.\_transaction type*.sidecars in the blk file                                                        |
```
steven-sheehy, line 204: Should be zero. Sidecars should go away.

docs/design/block-streams.md
```diff
@@ -0,0 +1,381 @@
...
+| size               | byte size of the blk file                                                                                                           |
+| version            | Record stream version - Perhaps a new version should be added to indicate "Translated from Block"                                   |
```
steven-sheehy, line 206: Just assume it's 7.

docs/design/block-streams.md
```diff
@@ -0,0 +1,381 @@
...
+| topic_message.sequence_number       | state_changes[i].state_change.map_update.value.sequenceNumber |
+| topic_message.topic_id              | state_changes[i].state_change.map_update.value.topicId        |
+| topic_message.valid_start_timestamp |                                                               |
+
+### Topic Delete Transaction
+
+| Database       | Block Item                                             |
+| -------------- | ------------------------------------------------------ |
+| entity.id      | state_changes[i].state_change.map_update.value.topicId |
+| entity.deleted | state_changes[i].state_change.map_update.value.deleted |
```
steven-sheehy, line 359: This value doesn't exist. This should be true i map_delete present or false for any other update.

steven-sheehy (changes requested):

docs/design/block-streams.md
```diff
@@ -0,0 +1,259 @@
...
+
+## Non-Goals
+
+- Support for Block Nodes will be covered in a separate design document.
+
+## Architecture
+
+### Data Flow
+
+![Data Flow](images/blockstream.png)
```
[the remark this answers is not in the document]
edwin-greene: Thanks I have added it

docs/design/block-streams.md
```diff
@@ -0,0 +1,381 @@
...
+## Non-Goals
+
+- Refactoring the importer to support block streams natively and the removal of record stream parsing.
+- Support for Block Nodes will be covered in a separate design document.
+
+## Architecture
+
+### Data Flow
+
+![Data Flow](images/blockstream.png)
```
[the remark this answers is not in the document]
edwin-greene: Got it, it has been added

docs/design/block-streams.md
edwin-greene: Sounds good, I have added them.

docs/design/block-streams.md
```diff
@@ -0,0 +1,259 @@
...
+| entity.staked_account_id                      | state_changes[i].state_change.map_update.value.account_value.staked_account_id                          |
+| entity.staked_node_id                         | state_changes[i].state_change.map_update.value.account_value.staked_node_id                             |
+| entity.stake_period_start                     | state_changes[i].state_change.map_update.value.account_value.stake_period_start                         |
+| entity.submit_key                             | state_changes[i].state_change.map_update.value.account_value.submit_key                                 |
+| entity.timestamp_range                        | Calculated by Mirror Node Upsert                                                                        |
+| entity.type                                   | state_changes[i].state_change.map_update.value: hasAccountValue/hasTokenValue                           |
+
+## REST API
+
+-Deprecate the REST API's State Proof Alpha. The record files and signature files will no longer be provided in the cloud buckets.
```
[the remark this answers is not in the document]
edwin-greene: Fixed

docs/design/block-streams.md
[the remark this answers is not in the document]
edwin-greene: Got it, I have updated that part.

docs/design/block-streams.md
```diff
@@ -0,0 +1,381 @@
...
+}
+```
+
+#### StreamPoller
+
+```java
+package com.hedera.mirror.importer.downloader;
+
+public abstract class StreamPoller<StreamFile> {
+    public abstract void poll();
```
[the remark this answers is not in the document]
edwin-greene: Sounds good, I have updated it to an interface.

docs/design/block-streams.md
```diff
@@ -0,0 +1,381 @@
...
+```
+
+#### StreamFileTranslator
+
+```java
+package com.hedera.mirror.importer.downloader;
+
+public interface StreamFileTranslator<T extends StreamFile<?>, S extends StreamFile<?>> {
+    // Used for translating a block file into a record file in phase one and for translating a record file into a block file in phase two.
+    T translate(S s);
```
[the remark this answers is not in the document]
edwin-greene: Good idea, I have changed it to transformer.

docs/design/block-streams.md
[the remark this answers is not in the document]
edwin-greene: Thanks I have updated it and added a note to do it as a lower priority task at the end.

edwin-greene: Thanks I have added the sql.

docs/design/block-streams.md
edwin-greene: Thanks I have removed that part, it will not be needed if StreamFileNotifier is used rather than implemented.

docs/design/block-streams.md
edwin-greene: Got it I have updated it with a note.

docs/design/block-streams.md
edwin-greene: Sounds good I have updated this to 7

docs/design/block-streams.md
edwin-greene: Thanks for spotting, I have updated it to map_delete/key/value

docs/design/block-streams.md
```diff
@@ -0,0 +1,381 @@
...
+## Domain
+
+### Interfaces and Classes
+
+#### BlockItem
+
+```java
+package com.hedera.mirror.common.domain.transaction;
+
+public class BlockItem implements StreamItem {
```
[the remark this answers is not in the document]
steven-sheehy: We should show the fields in BlockItem. We should filter out proto BlockItems that don't apply to mirror node like EventTransaction in the reader. Thus, our BlockItem doesn't correspond to proto BlockItems but actually aggregates them so that there's one of our BlockItem per Transaction that includes Transaction, result, output, and state changes.

docs/design/block-streams.md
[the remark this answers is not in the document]
steven-sheehy: It's still not quite right. StreamFileProvider is now there but the direction is wrong and the BlockStreamPoller also goes directly to S3. It should just be S3 <- provider <- poller.

docs/design/block-streams.md
```diff
@@ -0,0 +1,400 @@
...
+
+- Add `software_version` to the `record_file` table. This is the consensus node version that generated the block.
+- Add `congestionPricingMultiplier`, `round_start`, `round_end` to the `record_file` table, these come from the block.
+- Rename `record_file` to `block`. This will be a low priority task near the end of implementing phase one, as it requires a large number of changes.
+
+```sql
+alter table if exists record_file
+    add column if not exists software_version_major        int          null,
+    add column if not exists software_version_minor        int          null,
+    add column if not exists software_version_patch        int          null,
```
steven-sheehy, line 187: We should add a note that we should migrate this to be populated from hapi_version* since they're always identical before block streams. And also set it in existing record file readers

docs/design/block-streams.md
```diff
@@ -0,0 +1,400 @@
...
+## Block protobuf to mirror node database mapping
+
+### Record File
+
+| Database           | Block Item                                                                                                                          |
+| ------------------ | ----------------------------------------------------------------------------------------------------------------------------------- |
+| bytes              | raw bytes that comprise the blk file                                                                                                |
+| consensus_start    | block_header.first_transaction_consensus_time                                                                                       |
+| consensus_end      | consensus_timestamp of last transaction_output or transaction_result in the blk file                                                |
+| count              | Sum of the transaction_output and transaction_result typed block items in the blk file                                              |
```
steven-sheehy, line 209:
```suggestion
| count              | Sum of the EventTransaction typed block items in the blk file                                              |
```

docs/design/block-streams.md
```diff
@@ -0,0 +1,400 @@
...
+
+| Database            | Block Item                                 |
+| ------------------- | ------------------------------------------ |
+| consensus_timestamp | transaction_result.consensus_timestamp     |
+| prng_bytes          | transaction_output.util_prng.entropy.value |
+| prng_number         | transaction_output.util_prng.entropy.value |
+
+## REST API
+
+- Deprecate the REST API's State Proof Alpha. The record files and signature files will no longer be provided in the cloud buckets.
```
steven-sheehy, line 400: We should have a section that discusses the non-functional requirements including performance testing. We should also enumerate the system test cases.

docs/design/block-streams.md
```diff
@@ -0,0 +1,400 @@
...
+| prev_hash          | block_header.previous_block_hash                                                                                                    |
+| sidecar_count      | Set to 0 as sidecar data is being integrated into the block                                                                         |
+| size               | byte size of the blk file                                                                                                           |
+| version            | Set to 7                                                                                                                            |
+
+### Contract Create Transaction
+
+| Database                                      | Block Item                                                                                              |
+| --------------------------------------------- | ------------------------------------------------------------------------------------------------------- |
+| contract.id                                   | transaction_output.contract_create.sidecars[i].bytecode.contract_id                                     |
```
steven-sheehy, line 231: sidecars has gone away in current HIP. please update this table

docs/design/block-streams.md
```diff
@@ -0,0 +1,400 @@
...
+| contract.file_id                              | state_changes[i].state_change.map_update.value.file_value.file_id                                       |
```
steven-sheehy, line 232: We can get this from the transaction body. In which case we can omit this mapping.

docs/design/block-streams.md
```diff
@@ -0,0 +1,400 @@
...
+| contract.runtime_bytecode                     | transaction_output.contract_create.sidecars[i].bytecode.runtime_bytecode                                |
+| contract_action.call_depth                    | transaction_output.contract_create.sidecars[i].actions.contract_actions.call_depth                      |
+| contract_action.call_operation_type           | transaction_output.contract_create.sidecars[i].actions.contract_actions.call_operation_type             |
+| contract_action.call_type                     | transaction_output.contract_create.sidecars[i].actions.contract_actions.call_type                       |
+| contract_action.caller                        | transaction_output.contract_create.sidecars[i].actions.contract_actions.caller                          |
+| contract_action.caller_type                   | transaction_output.contract_create.sidecars[i].actions.contract_actions.caller_case                     |
+| contract_action.consensus_timestamp           | transaction_output.contract_create.sidecars[i].consensus_timestamp                                      |
```
steven-sheehy, line 240: Believe this timestamp is going away. Can get timestamp from result

docs/design/block-streams.md
```diff
@@ -0,0 +1,400 @@
...
+| contract_action.gas                           | transaction_output.contract_create.sidecars[i].actions.contract_actions[j].gas                          |
+| contract_action.gas_used                      | transaction_output.contract_create.sidecars[i].actions.contract_actions[j].gas_used                     |
+| contract_action.index                         | Index j of transaction_output.contract_create.sidecars[i].actions.contract_actions[j]                   |
+| contract_action.input                         | transaction_output.contract_create.sidecars[i].actions.contract_actions.input                           |
+| contract_action.payer_account_id              | state_changes[i].state_change.map_update.value.account_value.account_id                                 |
```
steven-sheehy, line 245: Don't think this is accurate. Can get it from TransactionID

docs/design/block-streams.md
```diff
@@ -0,0 +1,400 @@
...
+| contract_action.recipient_contract            | transaction_output.contract_create.sidecars[i].actions.contract_actions[j].recipient_contract           |
+| contract_action.result_data                   | transaction_output.contract_create.sidecars[i].actions.contract_actions[j].result_data                  |
+| contract_action.result_data_type              | transaction_output.contract_create.sidecars[i].actions.contract_actions[j].result_data_case             |
+| contract_action.value                         | transaction_output.contract_create.sidecars[i].actions.contract_actions[j].value                        |
+| contract_log.bloom                            | transaction_output.contract_create.contract_create_result.log_info[i].bloom                             |
+| contract_log.consensus_timestamp              | transaction_result.consensus_timestamp                                                                  |
+| contract_log.contract_id                      | transaction_output.contract_create.contract_create_result.log_info[i].contract_ID                       |
```
steven-sheehy, line 254: contract_ID casing

docs/design/block-streams.md
```diff
@@ -0,0 +1,400 @@
...
+| contract_log.data                             | transaction_output.contract_create.contract_create_result.log_info[i].data                              |
+| contract_log.index                            | Index i of transaction_output.contract_create.contract_create_result.log_info[i]                        |
+| contract_log.root_contract_id                 | transaction_output.contract_create.contract_create_result.contractID                                    |
```
steven-sheehy, line 257: The contract_create_result oneof has gone away in latest HIP.

docs/design/block-streams.md
```diff
@@ -0,0 +1,400 @@
...
+
+### CryptoTransfer Transaction
+
+| Database                                        | Block Item                                                       |
+| ----------------------------------------------- | ---------------------------------------------------------------- |
+| assessed_custom_fee.amount                      | transaction_output.assessedCustomFees[i].amount                  |
+| assessed_custom_fee.collector_account_id        | transaction_output.assessedCustomFees[i].feeCollectorAccountId   |
+| assessed_custom_fee.consensus_timestamp         | transaction_result.consensus_timestamp                           |
+| assessed_custom_fee.effective_payer_account_ids | transaction_output.assessedCustomFees[i].effectivePayerAccountId |
+| assessed_custom_fee.token_id                    | transaction_output.assessedCustomFees[i].tokenId                 |
```
steven-sheehy, line 351: These are in transaction_output.(crypto_transfer|token_airdrop). Please go over latest changes in HIP and update table

docs/design/block-streams.md
```diff
@@ -0,0 +1,400 @@
...
+
+### Token Transfer
+
+| Database                           | Block Item                                                                                       |
+| ---------------------------------- | ------------------------------------------------------------------------------------------------ |
+| token_transfer.account_id          | transaction_result.tokenTransferLists.tokenTransferList[i].transfers[j].accountAmount.accountID  |
+| token_transfer.amount              | transaction_result.tokenTransferLists.tokenTransferList[i].transfers[j].accountAmount.amount     |
+| token_transfer.consensus_timestamp | transaction_result.consensus_timestamp                                                           |
+| token_transfer.is_approval         | transaction_result.tokenTransferLists.tokenTransferList[i].transfers[j].accountAmount.isApproval |
+| token_transfer.token_id            | transaction_result.tokenTransferLists.tokenTransferList[i].token.tokenID                         |
```
steven-sheehy, line 361: Anything in `transaction_result` will be nested on level deeper under a transaction specific path.

docs/design/block-streams.md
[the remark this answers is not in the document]
edwin-greene: Sounds good I have added them and noted that it is an aggregate of protobuf block items.

docs/design/block-streams.md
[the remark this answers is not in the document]
edwin-greene: Thanks, I have updated it.

docs/design/block-streams.md
edwin-greene: Good idea, I have added a note about the migration and the change to the readers.

docs/design/block-streams.md
```diff
@@ -0,0 +1,400 @@
...
+
+## Block protobuf to mirror node database mapping
+
+### Record File
+
+| Database           | Block Item                                                                                                                          |
+| ------------------ | ----------------------------------------------------------------------------------------------------------------------------------- |
+| bytes              | raw bytes that comprise the blk file                                                                                                |
+| consensus_start    | block_header.first_transaction_consensus_time                                                                                       |
+| consensus_end      | consensus_timestamp of last transaction_output or transaction_result in the blk file                                                |
```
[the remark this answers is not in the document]
edwin-greene: I've added a note to that line, and added some text around empty blocks to the reader.

docs/design/block-streams.md
edwin-greene: Fixed.

docs/design/block-streams.md
edwin-greene: Great idea, I have added a section on testing with some ideas.

docs/design/block-streams.md
edwin-greene: Got it, those are removed.

docs/design/block-streams.md
edwin-greene: Sounds good I have removed that line.

docs/design/block-streams.md
edwin-greene: I have update it.

docs/design/block-streams.md
edwin-greene: I see, I have fixed that line.

docs/design/block-streams.md
edwin-greene: Got it, fixed

docs/design/block-streams.md
edwin-greene: Got it, I have updated it.

docs/design/block-streams.md
edwin-greene: Thanks, I think I have caught them all.

docs/design/block-streams.md
edwin-greene: Got it, I believe I have caught them all.

docs/design/block-streams.md
steven-sheehy: Well, I guess it's only EventTransactions with application transaction field populated. So maybe better to just use transaction_result. Main thing was we shouldn't count both output and result since it would be double the count.

steven-sheehy:
We should discuss somewhere how we're going to handle filtered block items

docs/design/block-streams.md
```diff
@@ -0,0 +1,501 @@
...
+
+#### BlockItem
+
+```java
+package com.hedera.mirror.common.domain.transaction;
+
+// Multiple protobuf BlockItems will be combined into a single BlockItem
+public record BlockItem(Transaction transaction,
+                        TransactionResult transactionResult,
+                        List<TransactionOutput> transactionOutput, // Note: List may be empty
```
steven-sheehy, line 39: There will be a 1:1 mapping between Transaction and TransactionOutput. I'm not sure if for transaction types that don't have an item in the output oneof whether they will include an empty TransactionOutput or no output. You should check.

```suggestion
                       TransactionOutput transactionOutput,
```

jnels124 (dismissed):
LGTM

docs/design/block-streams.md
edwin-greene: In testing I've found an `EventTransaction` that is a topic update transaction, and in that case the block contained just a `TransactionResult` and had no `TransactionOutput`. 

It may need to be a `List<TransactionOutput>` as an `EventTransaction` that is a token transfer looks like it can be followed by more than one `TransactionOutput`, a `CallContractOutput` and a `CryptoTransferOutput`.

steven-sheehy: You're right. Looks like EVM invoking transactions can contain an extra CallContractOutput.

docs/design/block-streams.md
```diff
@@ -0,0 +1,503 @@
...
+| entity.id      | state_changes[i].state_change.map_update.value.topicId |
+| entity.deleted | state_changes[i].state_change.map_delete.key.topicId   |
+
+### Topic Submit Message
+
+| Database                           | Block Item                                                                                            |
+| ---------------------------------- | ----------------------------------------------------------------------------------------------------- |
+| topic_message.consensus_timestamp  | state_changes.consensus_timestamp                                                                     |
+| topic_message.running_hash         | state_changes[i].state_change.map_update.value.runningHash                                            |
+| topic_message.running_hash_version | No longer submitted by the block stream: This is a fixed value that is only updated with a HIP change |
```
steven-sheehy, line 470: We should make this field nullable and stop populating it then. This can reduce the database size for such a large table.

docs/design/block-streams.md
```diff
@@ -0,0 +1,503 @@
...
+| topic_message.sequence_number      | state_changes[i].state_change.map_update.value.sequenceNumber                                         |
+| topic_message.topic_id             | state_changes[i].state_change.map_update.value.topicId                                                |
```
steven-sheehy, line 472: They don't need to update the topic id for a message submit. This will come from the body so can omit from mapping.

docs/design/block-streams.md
```diff
@@ -0,0 +1,503 @@
...
+### Token Airdrop
+
+| Database                           | Block Item                                                                                     |
+| ---------------------------------- | ---------------------------------------------------------------------------------------------- |
+| token_airdrop.amount               | state_changes[i].state_change.map_update.value.accountPendingAirdropValue.pendingAirdropValue  |
+| token_airdrop.receiver_account_id  | state_changes[i].state_change.map_update.key.receiverId                                        |
+| token_airdrop.sender_account_id    | state_changes[i].state_change.map_update.key.senderId                                          |
+| token_airdrop.serial_number        | state_changes[i].state_change.map_update.key.tokenReference.serial                             |
+| token_airdrop.token_id             | state_changes[i].state_change.map_update.key.tokenReference.tokenId                            |
+| Updates accessed_custom_fee        | Similar to crypto_transfer, accessed at transaction_output.token_airdrop                       |
```
steven-sheehy, line 414:
```suggestion
| assessed_custom_fee        | Similar to crypto_transfer, accessed at transaction_output.token_airdrop                       |
```

docs/design/block-streams.md
```diff
@@ -0,0 +1,503 @@
...
+
+| Database    | Block Item                                                  |
+| ----------- | ----------------------------------------------------------- |
+| schedule.id | transaction_output.schedule_create.scheduled_transaction_id |
+
+### Schedule Delete Transaction
+
+| Database    | Block Item                                                               |
+| ----------- | ------------------------------------------------------------------------ |
+| schedule.id | state_changes[i].state_change.map_update.value.schedule_value.scheduleID |
```
steven-sheehy, line 397: Will delete not use map_delete or will it do map_update and set a deleted=true flag in state? Either way, we can just map via schedule_id in body.

docs/design/block-streams.md
```diff
@@ -0,0 +1,503 @@
...
+
+| Database     | Block Item                                                     |
+| ------------ | -------------------------------------------------------------- |
+| node.node_id | state_changes[i].state_change.map_update.key.entity_number_key |
+
+### Schedule Create Transaction
+
+| Database    | Block Item                                                  |
+| ----------- | ----------------------------------------------------------- |
+| schedule.id | transaction_output.schedule_create.scheduled_transaction_id |
```
steven-sheehy, line 391: a transaction ID is not the same as an entity ID

docs/design/block-streams.md
```diff
@@ -0,0 +1,503 @@
...
+| Database                      | Block Item                                             |
+| ----------------------------- | ------------------------------------------------------ |
+| file_data.consensus_timestamp | transaction_result.consensus_timestamp                 |
+| file_data.entity_id           | state_changes[i].state_change.map_update.key.fileIdKey |
+
+### Node Create Transaction
+
+| Database     | Block Item                                                     |
+| ------------ | -------------------------------------------------------------- |
+| node.node_id | state_changes[i].state_change.map_update.key.entity_number_key |
```
steven-sheehy, line 385: Are you sure it's entity_number_key? node_id is not an Entity ID as it has a separate counter. Maybe it's proto_long. Should double check.

edwin-greene:
> We should discuss somewhere how we're going to handle filtered block items

Thanks I have added a note about that to the ProtoBlockReader.

docs/design/block-streams.md
edwin-greene: I have updated the design with sql to make it nullable and then added a section for the importer to change the handling of the runningHashVersion.

docs/design/block-streams.md
edwin-greene: Thanks I have removed it.

docs/design/block-streams.md
edwin-greene: From my testing so far it will be in a map_update with delete flag. Since it will be in the body I have removed its mapping here.

docs/design/block-streams.md
edwin-greene: Thanks, I double checked and it is now fixed, it should be a value in state changes.

docs/design/block-streams.md
```diff
@@ -0,0 +1,507 @@
...
+     * Transforms the block file into a record file, verifies the hash chain and then parses it
+     * Does not parse Block N until Block N+1 has been downloaded and used to verify Block N
+     */
+    public void notify(@Nonnull StreamFile<?> streamFile);
+
+    /**
+     * The hash chain cannot be verified for Block N until Block N+1 has been downloaded.
+     * For Block N the hash must be verified to match the previousBlockHash protobuf value provided by Block N+1
+     */
+    private void verifyHashChain(String expected, String actual);
```
xin-hedera, line 183: I don't think the claim is true. We should be able to calculate the current block's root hash given the data in `BlockProof`, there is previous block's root hash in `BlockHeader`. With the info, we can still verify the chain of hash. For example, suppose we are at block `N`, and the calculated root hash of `N` is `H(N)`; when the next block `N+1` is downloaded, we check `BlockHeader. previous_block_hash` against `H(N)`, a mismatch means we should not further process block `N+1`

What's unclear is the detailed algorithm to calculate `H(N)`, given the previous block root hash, event transactions, start state root hash, and state changes.

docs/design/block-streams.md
```diff
@@ -0,0 +1,507 @@
...
+
+    // Verifies that the number of the block file contained in its file name matches the block number within the block file
+    private void verifyBlockNumber(String expected, String actual);
+}
+```
+
+### Database
+
+- Add `software_version` to the `record_file` table. This is the consensus node version that generated the block.
+- Add `congestionPricingMultiplier`, `round_start`, `round_end` to the `record_file` table, these come from the block.
```
xin-hedera, line 193: congestionPricingMultiplier is in `TransactionResult`, is there anywhere in the HIP stating it's the same across a whole block? I strongly doubt so, at least we have different throttles for different transactions, so it's normal for just one type of transaction having congestion pricing applied since only that type of transaction is hitting throttle.

docs/design/block-streams.md
```diff
@@ -0,0 +1,507 @@
...
+    add column if not exists round_end                     bigint       null;
+
+alter table if exists topic_message
+    alter column if exists running_hash_version drop not null;
+```
+
+### Block to Record File Transformation
+
+Blocks are composed of block items. A record item may be transformed from a set of multiple block items.
+Beginning from an `EventTransaction` block item, a record item is composed of one `TransactionResult` block item, zero to N `TransactionOutput` block items and zero to N `StateChange` block items.
```
xin-hedera, line 214: there should be 0 to 1 `StateChanges` block item for an `EventTransaction`, since the `StateChanges` proto message itself is a container of multiple state changes

```protobuf
message StateChanges {
    /**
     * The consensus timestamp of this set of changes.
     * See the full specification above for how this value is determined.
     * <p>
     * This value SHALL be deterministic.
     */
    proto.Timestamp consensus_timestamp = 1;

    /**
     * An ordered list of individual changes.
     * <p>
     * These changes MUST be applied in the order listed to produce
     * a correct modified state.
     */
    repeated StateChange state_changes = 2;
}
```

docs/design/block-streams.md
```diff
@@ -0,0 +1,507 @@
...
+
+### Token Airdrop
+
+| Database                           | Block Item                                                                                     |
+| ---------------------------------- | ---------------------------------------------------------------------------------------------- |
+| token_airdrop.amount               | state_changes[i].state_change.map_update.value.accountPendingAirdropValue.pendingAirdropValue  |
+| token_airdrop.receiver_account_id  | state_changes[i].state_change.map_update.key.receiverId                                        |
+| token_airdrop.sender_account_id    | state_changes[i].state_change.map_update.key.senderId                                          |
+| token_airdrop.serial_number        | state_changes[i].state_change.map_update.key.tokenReference.serial                             |
+| token_airdrop.token_id             | state_changes[i].state_change.map_update.key.tokenReference.tokenId                            |
```
xin-hedera, line 418: TokenReference is a oneof, so token id can be either fungible_token_type or non_fungible_token.token_ID. similar comment for serial.

```
oneof token_reference {
 /**
  * A token ID.<br/>
  * This is the type of token for a fungible/common token airdrop.<br/>
  * This field is REQUIRED for a fungible/common token and MUST NOT be used for a
  * non-fungible/unique token.
  */
 TokenID fungible_token_type = 3;

 /**
  * The id of a single NFT, consisting of a Token ID and serial number.<br/>
  * This is the type of token for a non-fungible/unique token airdrop.<br/>
  * This field is REQUIRED for a non-fungible/unique token and MUST NOT be used for a
  * fungible/common token.
  */
 NftID non_fungible_token = 4;
}
```

docs/design/block-streams.md
```diff
@@ -0,0 +1,507 @@
...
+| Updates accessed_custom_fee        | Similar to crypto_transfer, accessed at transaction_output.token_airdrop                       |
+| token_account.account_id           | transaction_output.token_airdrop.automatic_token_associations[i].accountId                     |
+| token_account.token_id             | transaction_output.token_airdrop.automatic_token_associations[i].tokenId                       |
+| token_transfer.account_id          | transaction_output.token_airdrop.token_transfer_lists[i].transfers[j].accountAmount.accountID  |
+| token_transfer.amount              | transaction_output.token_airdrop.token_transfer_lists[i].transfers[j].accountAmount.amount     |
+| token_transfer.consensus_timestamp | transaction_result.consensus_timestamp                                                         |
+| token_transfer.is_approval         | transaction_output.token_airdrop.token_transfer_lists[i].transfers[j].accountAmount.isApproval |
+| token_transfer.token_id            | transaction_output.token_airdrop.token_transfer_lists[i].token.tokenID                         |
+
```
xin-hedera, line 427: There is no `TokenClaimAirdropOutput` with token transfer list for claimed airdrops. I asked about it in the HIP, however was told that we can get the token transfers from state changes. What I don't understand is why we'd have the token transfer lists in `CryptoTransferOutput`, `TokenAirdrop`.

Btw there's still `token_tranfser_lists` in the global `TransactionResult`, my recollection is it should be removed.

The impact to our design is we may need a mapping from state change to token transfers somehwere.

docs/design/block-streams.md
```diff
@@ -0,0 +1,507 @@
...
+| ---------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| topic_message.consensus_timestamp  | state_changes.consensus_timestamp                                                                                                                             |
+| topic_message.running_hash         | state_changes[i].state_change.map_update.value.runningHash                                                                                                    |
+| topic_message.running_hash_version | Map this value to the current value of 3. This value is no longer submitted by the block stream: This is a fixed value that is only updated with a HIP change |
+| topic_message.sequence_number      | state_changes[i].state_change.map_update.value.sequenceNumber                                                                                                 |
+
+### Transaction
+
+| Database                               | Block Item                                              |
```
xin-hedera, line 480: there is `TransactionRecord.transactionID`, however it's not in block stream `TransactionResult`. The impact to us is if the payer account is in the alias form, now we have to lookup its numerical account id from database, for partial mirrornode, we may not be able to get it. Perhaps, we should figure out how to get the info from state changes (the account pays the fee).

docs/design/block-streams.md
```diff
@@ -0,0 +1,507 @@
...
+     *   Block items are only iterated once in the transform method
+     *   State changes are accumulated and used for calculating the block hash
+     *
+     *   If the Block File contains a Wrapped Record File, then convert the Wrapped Record File to a Record File
+     */
+    @Override
+    public RecordFile transform(BlockFile block);
+
+    // Block hashes are not included in the block. They must be calculated from the state changes within the block.
+    private byte[] calculateBlockHash(byte[] previousBlockHash, List<StateChanges> stateChanges);
```
xin-hedera, line 116: Per the [diagram](https://raw.githubusercontent.com/hashgraph/hedera-improvement-proposal/refs/heads/block-streams/assets/hip-1056/block-stream-merkle.svg) in the HIP, the following is needed to calculate the block hash

- previous block root hash
- root of input merkle tree (formed by event header and even transaction pairs)
- root of output merkle tree (formed by block header, transaction output, and state changes)
- root of state merkle tree at beginning of the block (this is included in BlockProof)

docs/design/block-streams.md
edwin-greene: One thought, when block N+1 is downloaded and the check the previousBlockhash indicates that we have a mismatch... if we have already parsed block N, couldn't we have introduced invalid data into the database?

xin-hedera: That’s what TSS signature is for. It replaces today’s multi node signature files. So when processing a block file we should also very tss signature in the block proof, though the details of the algorithm is left out from the HIP

docs/design/block-streams.md
edwin-greene: Thanks Xin, this needs to be fixed. Looking into it more, the `root of input merkle tree` may be included in the last StateChange in a block, the value coming from `SingletonUpdateChange.BlockStreamInfo.inputTreeRootHash`. 

To get the `root of the output merkle tree` it looks like we will need to pass a List of BlockItems (only the specific ones you mentioned as shown in the diagram). 

At the moment this design calls for protobuf block items being filtered and consolidated into mirror node blockItems in the `BlockFileReader`. So the protobuf block items won't be present here in the transformer. One way to addess this may be for the reader to calculate the output hash for the consolidated mirror node block item and have that as another field on the block item domain object.

edwin-greene: I have updated the design with a section for methods for the blockhash; depending on what we would like to do with the Reader, I will update that section to be fields from the domain objects or fields from the protobuf BlockItems.

docs/design/block-streams.md
edwin-greene: I have updated this to remove the "cannot be verified" text. From the current schedule, it looks like the TSS Signatures will be mocked in Blockstreams phase 0; and included in Blockstreams phase 1. We may have to revise this section once more information has been nailed down.

docs/design/block-streams.md
edwin-greene: Thanks, I have removed it from the record file; do you think it should be captured somewhere else? The transaction table?

docs/design/block-streams.md
edwin-greene: Thanks I have updated this.

docs/design/block-streams.md
edwin-greene: Thanks I have updated this section, with different rows for fungible/nft.

docs/design/block-streams.md
xin-hedera: not sure how important the info is. if we want to persist it for future use cases, it should go to transaction table. Assume the default is `1` when no congestion pricing applied, we can make that as the column's default value to reduce table storage.

docs/design/block-streams.md
xin-hedera: alias form payer isn't supported in services yet, so can disregard it for now though it remains a future possiblity.

docs/design/block-streams.md
```diff
@@ -0,0 +1,525 @@
...
+    */
+    private byte[] calculateBlockHash(
+            byte[] previousBlockHash,
+            byte[] inputTreeRootHash,
+            byte[] outputTreeRootHash,
+            byte[] startOfBlockStateRootHash
+    );
+
+    // The transaction hash will not be included in the block stream output so we will need to calculate it
+    private byte[] calculateTransactionHash(EventTransaction transaction);
```
IvanKavaldzhiev, line 139: Note that there are ongoing activities to get rid of the `EventTransaction` wrapper. The block stream will include the Transaction proto type directly as a BlockItem and EventTransaction is going to be deprecated.

More information can be found [here](https://github.com/hashgraph/hedera-services/issues/16665).

edwin-greene: Thanks for pointing that out. I have added a bullet to the open issues to capture this feedback

docs/design/block-streams.md
edwin-greene: Thanks, I have added a bullet point to the open issues for this. Once the HIP is finalized we may have to go back and update our design

docs/design/block-streams.md
steven-sheehy: It still shows as accessed instead of assessed

docs/design/block-streams.md
```diff
@@ -0,0 +1,598 @@
...
+package com.hedera.mirror.importer.reader.block;
+
+public class ProtoBlockFileReader implements BlockFileReader {
+    // Generates a BlockFile from a StreamFileData.
+    // Converts the protobuf BlockItems into mirror node BlockItems.
+    // Protobuf BlockItems that do not represent a transaction will be filtered out here.
+    // Note that a protobuf BlockFile with no transactions will still produce a mirror node BlockFile and persist that block to the database.
+    public BlockFile read(StreamFileData streamFileData);
+}
+```
```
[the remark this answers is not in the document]
steven-sheehy: Can't you just store the necessary info in the `BlockFile` which also contains the list of block items?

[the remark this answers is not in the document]
edwin-greene: Great idea, sounds good.

docs/design/block-streams.md
edwin-greene: Got it! Fixed now

steven-sheehy (approved):

jnels124 (approved):
LGTM

xin-hedera (approved):
LGTM

PR #9719, opened 2024-11-10
Updated the query to return all relevant transactions

### LINKED ISSUE #9606: Transactions API Should Return All Relevant Transactions
### Problem

`parent_consensus_timestamp` is only set on following children. #4672 made changes to exclude synthetic transactions and inadvertently excludes transactions that precede another transaction. For instance, a `CRYPTOTRANSFER` that creates a  `CRYPTOCREATEACCOUNT` transaction. These `CRYPTOCREATEACCOUNT` transactions will have a non-zero `nonce` and will not have a `parent_consensus_timestamp` so the query adjustment in the mentioned pr will exclude these transactions in the response.

### Solution

- [ ] Determine other transaction types that may be effected
- [ ] Adjust query so that all relevant transactions are returned

### Alternatives

_No response_

HarshSawarkar:
**Description**:

Modified the query to return all relevant transaction
**Related issue(s)**:

Fixes #9606 

**Notes for reviewer**:

**Checklist**


jnels124 (changes requested):

hedera-mirror-rest/transactions.js
```diff
@@ -343,8 +343,12 @@ const getTransferDistinctTimestampsQuery = (
 };
 
 // the condition to exclude synthetic transactions attached to a user submitted transaction
+const transactionTypesToInclude = ['CRYPTOAPPROVEALLOWANCE', 'CRYPTOTRANSFER', 'CRYPTODELETEALLOWANCE'];
```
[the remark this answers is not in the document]
HarshSawarkar: Updated.

hedera-mirror-rest/transactions.js
```diff
@@ -344,7 +345,8 @@ const getTransferDistinctTimestampsQuery = (
 
 // the condition to exclude synthetic transactions attached to a user submitted transaction
 const transactionByPayerExcludeSyntheticCondition = `${Transaction.getFullName(Transaction.NONCE)} = 0 or
-  ${Transaction.getFullName(Transaction.PARENT_CONSENSUS_TIMESTAMP)} is not null`;
+  ${Transaction.getFullName(Transaction.PARENT_CONSENSUS_TIMESTAMP)} is not null 
+  or ${Transaction.getFullName(Transaction.TYPE)} in (${typesToInclude.map((type) => `'${type}'`).join(', ')})`;
```
jnels124, line 349: Once other comments are addressed, `.map` can be removed. Also, we should only add condition if array is not empty.

hedera-mirror-rest/config/application.yml
```diff
@@ -41,6 +41,10 @@ hedera:
           username: metrics
           uriPath: "/swagger"
         ipMetrics: false
+      transactions:
```
jnels124, line 44: This should be the protoId of these types. See `TransactionType.java`.

jnels124: Also, this property should be added to `configuration.md`

hedera-mirror-rest/transactions.js
jnels124: Should update the `hedera-mirror-rest/__tests__/specs/transactions/account-id.json` spec test to include these types of transactions.

jnels124 (changes requested):

hedera-mirror-rest/config/application.yml
HarshSawarkar: Updated

hedera-mirror-rest/transactions.js
HarshSawarkar: Updated

hedera-mirror-rest/config.js
```diff
@@ -188,6 +188,13 @@ const parseNetworkConfig = () => {
   }
 };
 
+const parseTransactionConfig = () => {
```
jnels124, line 191: can parse this in `parseQueryConfg`

PR #9777, opened 2024-11-18
Fix file readable kv state for latest file

kselveliev:
**Description**:
Seems like there is an issue with the `FileReadableKVState` when we try to find latest file by id.

`FileDataRepository.findById` does not work with entity id since the primary key of the table is the consensus timestamp column and currently we pass as a parameter the entity id from the `FileID` object which is incorrect.

To fix it we can use the existing method `getFileAtTimestamp` which uses both the entity id and the consensus timestamp.
We will just pass current timestamp as a parameter and the entity id.

 

FileReadableKVState - For latest data we now use `getFileAtTimestamp` method passing current time.
Added FileReadableKVStateIntegrationTest which adds integration test for the latest scenario

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/FileReadableKVState.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/FileReadableKVState.java
@@ -23,6 +23,7 @@ import com.hedera.hapi.node.state.file.File;
 import com.hedera.mirror.common.domain.entity.AbstractEntity;
 import com.hedera.mirror.common.domain.entity.EntityId;
 import com.hedera.mirror.common.domain.file.FileData;
+import com.hedera.mirror.common.util.DomainUtils;
 import com.hedera.mirror.web3.common.ContractCallContext;
 import com.hedera.mirror.web3.repository.EntityRepository;
 import com.hedera.mirror.web3.repository.FileDataRepository;
@@ -31,6 +32,7 @@ import com.hedera.pbj.runtime.io.buffer.Bytes;
 import com.swirlds.state.spi.ReadableKVStateBase;
 import jakarta.annotation.Nonnull;
 import jakarta.inject.Named;
+import java.time.Instant;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.Optional;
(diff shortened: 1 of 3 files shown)


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/FileReadableKVState.java
```diff
@@ -57,10 +59,9 @@ public FileReadableKVState(final FileDataRepository fileDataRepository, final En
     protected File readFromDataSource(@Nonnull FileID key) {
         final var timestamp = ContractCallContext.get().getTimestamp();
         final var fileId = toEntityId(key).getId();
-
         return timestamp
                 .map(t -> fileDataRepository.getFileAtTimestamp(fileId, t))
-                .orElseGet(() -> fileDataRepository.findById(fileId))
+                .orElseGet(() -> fileDataRepository.getFileAtTimestamp(fileId, getCurrentTimestamp()))
```
steven-sheehy, line 64: This shouldn't be necessary since `timestamp` for LATEST block will be filled in with the latest block timestamp from the db via StaticBlockMetaSource. So we don't even need the `orElseGet()`.

```suggestion
```

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/FileReadableKVState.java
kselveliev: Hmm, from what I've been testing and looking at the ContractCallService seems like we are only setting record file when we are passing a block which is not latest. 
```
// if we have historical call, then set the corresponding record file in the context
        if (params.getBlock() != BlockType.LATEST) {
                    .findByBlockType(params.getBlock())
                    .orElseThrow(BlockNumberNotFoundException::new));
        }
```
And then if we are using latest block then timestamp will be an empty optional since we are fetching it from ContractCallContext. I'm not sure if we can remove the line.

IvanKavaldzhiev: If we override `ContractCallContext.get().getTimestamp()` method to return current timestamp when Optional<Long> timestamp is empty, then we won't need to have `orElseGet()` statement.

kselveliev: Just debugged with contract create flow from feature-flag branch.
If we omit the orElseGet statement we are going to get -> Possibly CATASTROPHIC failure - exception thrown while handling dispatch java.lang.NullPointerException: Field currentRate is null -> since our code will return null for File.

So we should either use it how it is just getting the timestamp in some other way. Or another way to do it is to do 
`.orElseGet(() -> fileDataRepository.findByEntityId(fileEntityId))`

What do you think?

steven-sheehy: We can keep the current code.

steven-sheehy (approved):
LGTM

Issue #9791, opened 2024-11-21
Enhance `MirrorNodeState` to get initialized on startup


bilyana-gospodinova:
### Problem

We added a `MirrorNodeState` and its subcomponents but they are not called yet.

### Solution

Get `MirrorNodeState` initialized on startup.

Issue #9264, opened 2024-09-03
Verify correctness of state in the reusable services integration


IvanKavaldzhiev:
### Problem

Make sure we don't have cross-contamination of State between different concurrent transactions.

We also should make sure we create a single instance bean of the other state related components - NetworkInfo, ServiceMigrator, ServicesRegistry.

Make sure the new database accessors have cache on the repository layer and the settings defined work correctly.

### Solution

Validate the mentioned ideas for improvement.

If possible, opt to delete the ContractCallContext and make sure each transaction has its own set of components for proper execution.

bilyana-gospodinova:
https://github.com/hashgraph/hedera-mirror-node/pull/9527#discussion_r1795085661 - the comment here suggest some additional optimizations related to the way accounts are being searched by

steven-sheehy:
Another optimization suggestion: https://github.com/hashgraph/hedera-mirror-node/pull/9502#discussion_r1794316070

bilyana-gospodinova:
A suggestion about the schemas - https://github.com/hashgraph/hedera-mirror-node/pull/9793#discussion_r1852028900

PR #9793, opened 2024-11-21
Add `MirrorNodeState` init with subcomponents

bilyana-gospodinova:
**Description**:
This PR adds the `MirrorNodeState` initialization and subcomponents invocation. Some additional changes were made in order for the init method to finish successfully.
It will stay in draft until the PR with the flag introduction is merged.

Fixes https://github.com/hashgraph/hedera-mirror-node/issues/9791


hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/state/MirrorNodeStateTest.java
```diff
@@ -52,7 +48,6 @@
 import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
 
-@SuppressWarnings("unchecked")
 @ExtendWith(MockitoExtension.class)
 class MirrorNodeStateTest {
```
kselveliev, line 52: I thing we could also add a MirrorNodeStateIntegrationTest where we inject the MirrorNodeState and add a test 
that verifies that the MirrorNodeState is not null, has not empty states and contains as expected our custom kv states. 

This will test the init method in the MirrorNodeState and verify that it is populated in application startup

kselveliev: Just added the following test locally and it seems to be working as expected. 
Just needs more assertions -> 
```
class MirrorNodeStateIntegrationTest extends Web3IntegrationTest {
    Resource
    private MirrorNodeState mirrorNodeState;

    Test
    void checkMirrorNodeState() {
        assertThat(mirrorNodeState).isNotNull();
    }


}
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/common/ContractCallContext.java
```diff
@@ -79,6 +79,10 @@ public static ContractCallContext get() {
         return SCOPED_VALUE.get();
     }
 
+    public static Optional<Long> getTimestamp() {
+        return SCOPED_VALUE.isBound() ? SCOPED_VALUE.get().timestamp : Optional.empty();
```
IvanKavaldzhiev, line 83: In this [PR](https://github.com/hashgraph/hedera-mirror-node/pull/9777), it seems it might be beneficial to return Optional.of current/instant time. Do we have checks that are connected tightly to have Optional.empty() as return value that won't work with having the instant timestamp?

bilyana-gospodinova: In a lot of tests we mock that the returned timestamp is `Optional.empty()`, so this will break a lot of tests.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/state/MirrorNodeStateTest.java
bilyana-gospodinova: Yes, I debugged the initialization and it seems fine. I will try to add some integration tests.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/MirrorNodeState.java
```diff
@@ -265,4 +360,52 @@ public int hashCode() {
...
+                        new BlockRecordService(),
+                        new FeeService(),
+                        new CongestionThrottleService(),
+                        new RecordCacheService())
+                .forEach(servicesRegistry::register);
+    }
+
+    private Map<Long, Function<Configuration, Bytes>> genesisContentProviders(
+            final NetworkInfo networkInfo, final com.swirlds.config.api.Configuration config) {
+        final var genesisSchema = new V0490FileSchema();
```
IvanKavaldzhiev, line 381: Do we plan to use dynamically the most recent schema? Having a hardcoded schema will mean we will have to support it and change it from time to time.

bilyana-gospodinova: This was the example in `TransactionExecutorsTest`. I haven't investigated yet if we need to pass here the latest schema. This can be done in a separate PR after we clear up most of the issues.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/MirrorNodeState.java
```diff
@@ -265,4 +360,52 @@ public int hashCode() {
...
+        final var filesConfig = config.getConfigData(FilesConfig.class);
+        return Map.of(
+                filesConfig.addressBook(), ignore -> genesisSchema.genesisAddressBook(networkInfo),
+                filesConfig.nodeDetails(), ignore -> genesisSchema.genesisNodeDetails(networkInfo),
+                filesConfig.feeSchedules(), genesisSchema::genesisFeeSchedules,
+                filesConfig.exchangeRates(), genesisSchema::genesisExchangeRates,
+                filesConfig.networkProperties(), genesisSchema::genesisNetworkProperties,
+                filesConfig.hapiPermissions(), genesisSchema::genesisHapiPermissions,
+                filesConfig.throttleDefinitions(), genesisSchema::genesisThrottleDefinitions);
```
IvanKavaldzhiev, line 390: Would execution fail if we leave only `filesConfig.feeSchedules()` and `filesConfig.exchangeRates()` to be set?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/MirrorNodeState.java
IvanKavaldzhiev: Agree, that we can check the schema population and adapt it if needed, in a future PR.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/state/MirrorNodeStateTest.java
bilyana-gospodinova: Added some tests. They are not very extensive but will verify that the init method is called. Further verifications will be covered in all of the other existing integration tests.

kselveliev (approved):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/MirrorNodeState.java
```diff
@@ -265,4 +362,58 @@ public int hashCode() {
...
+    }
+
+    private void registerServices(ServicesRegistry servicesRegistry) {
+        // Register all service schema RuntimeConstructable factories before platform init
+        final var appContext =
+                new AppContextImpl(InstantSource.system(), signatureVerifier(), Gossip.UNAVAILABLE_GOSSIP);
+        Set.of(
+                        new EntityIdService(),
+                        new TokenServiceImpl(),
+                        new FileServiceImpl(),
```
steven-sheehy, line 378: Do we need the other services like consensus, schedule, util, etc? I would think util service would be used by prng precompile. Shouldn't we just register all of the ones services does so in case new precompiles come in then we don't have to change?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/components/MetricsImpl.java
```diff
@@ -0,0 +1,81 @@
...
+package com.hedera.mirror.web3.state.components;
+
+import com.swirlds.metrics.api.Metric;
+import com.swirlds.metrics.api.MetricConfig;
+import com.swirlds.metrics.api.Metrics;
+import jakarta.annotation.Nonnull;
+import jakarta.annotation.Nullable;
+import java.util.Collection;
+
+public class MetricsImpl implements Metrics {
```
steven-sheehy, line 26: Can we use UnavailableMetrics?

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/MirrorNodeState.java
bilyana-gospodinova: I added this in the task for optimizations - https://github.com/hashgraph/hedera-mirror-node/issues/9264

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/common/ContractCallContext.java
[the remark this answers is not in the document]
bilyana-gospodinova: This change was necessary because in the `MirrorNodeState` class we added an `init` method that runs all of the migrations and some of them require access to the DB and they use the `*ReadableKVState` classes. In them there is always logic when depending on the timestamp in the context we make a query against latest block or a historical block. This `init` method is run by the Spring container after the beans are initialized. At this point the scoped value is not always bound and this is why I added this check.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/MirrorNodeState.java
bilyana-gospodinova: It seems that the number of the failing tests does not change if I leave the code as you suggested.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/MirrorNodeState.java
bilyana-gospodinova: I have added only the minimal amount of services that we need for now. Note that when we start a test, all of the state initialization takes place including all of the services, migrations, subcomponents and this is quite slow even now. I am concerned that if we add more services this might slow down the development and the test execution even more. Also, the consumed memory will increase unnecessarily. We can't predict all future changes - there might be even a new services some day. In my opinion we should leave them like this for now.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/components/MetricsImpl.java
bilyana-gospodinova: We can, actually. I will remove the new class. Thanks.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/MirrorNodeState.java
```diff
@@ -62,6 +101,49 @@ public class MirrorNodeState implements State {
...
 
+    private final AccountReadableKVState accountReadableKVState;
+    private final AirdropsReadableKVState airdropsReadableKVState;
+    private final AliasesReadableKVState aliasesReadableKVState;
+    private final ContractBytecodeReadableKVState contractBytecodeReadableKVState;
+    private final ContractStorageReadableKVState contractStorageReadableKVState;
+    private final FileReadableKVState fileReadableKVState;
+    private final NftReadableKVState nftReadableKVState;
+    private final TokenReadableKVState tokenReadableKVState;
+    private final TokenRelationshipReadableKVState tokenRelationshipReadableKVState;
```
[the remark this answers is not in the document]
bilyana-gospodinova: Done, thanks.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/MirrorNodeState.java
IvanKavaldzhiev: I doubt we would ever need consensus service, unless we support precompile functionality related to topics.

Schedule service, might be needed if there is a support for scheduled precompiles.

I agree with Bilyana to start with the minimal amount of services and schemas and add components on demand.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/state/MirrorNodeStateIntegrationTest.java
```diff
@@ -0,0 +1,132 @@
...
+    @Test
+    void verifyMirrorNodeStateHasRegisteredServices() {
+        Set<Class<? extends Service>> expectedServices = new HashSet<>(List.of(
+                EntityIdService.class,
+                TokenServiceImpl.class,
+                FileServiceImpl.class,
+                ContractServiceImpl.class,
+                BlockRecordService.class,
+                FeeService.class,
+                CongestionThrottleService.class,
```
IvanKavaldzhiev, line 59: Does the state initiliazation fail if we remove `CongestionThrottleService`?

IvanKavaldzhiev (dismissed):
LGTM

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/state/MirrorNodeStateIntegrationTest.java
bilyana-gospodinova: Yes, I just tested it.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/common/ContractCallContext.java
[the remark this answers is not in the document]
steven-sheehy: Then we should fix the init method to wrap it in `ContractCallContext.run()`. I tested locally and this works. Please revert getTimestamp() change.

Also, we should probably disable all these migrations if they're touching the database on startup. This will greatly slow down startup with probably no benefit since it's read only state.

kselveliev (approved):

steven-sheehy (approved):

Issue #9798, opened 2024-11-22
Add option in the `ContractCallContext` to hold the contract that is currently being deployed


bilyana-gospodinova:
### Problem

During the implementation of the reusable services integration it turns out that now the contracts are deployed in 2 steps:
1. The initcode is uploaded and saved as a file using a `FileCreateTransactionBody`.
2. The returned file id from step 1 is then passed to a `ContractCreateTransactionBody`.
Each step performs a separate transaction.
For step 2 even if we pass the correct file id, since the mirror node data is readonly, the `FileReadableKVState` is not able to populate the contract's bytecode from the DB since it was never explicitly persisted in the DB.

### Solution

Add 2 new fields in the `ContractCallContext` that preserve the file id and the contract's bytecode. If they are available when we call `FileReadableKVState#readFromDataSource`, return them with priority.

PR #9799, opened 2024-11-22
Add file id and init bytecode in `ContractCallContext`

bilyana-gospodinova:
**Description**:
During the implementation of the reusable services integration it turns out that now the contracts are deployed in 2 steps:

1. The initcode is uploaded and saved as a file using a `FileCreateTransactionBody`.
2. The returned file id from step 1 is then passed to a `ContractCreateTransactionBody`.
Each step performs a separate transaction.
For step 2 even if we pass the correct file id, since the mirror node data is readonly, the `FileReadableKVState` is not able to populate the contract's bytecode from the DB since it was never explicitly persisted in the DB.

This PR add a new field in the `ContractCallContext` that preserve the file id and the contract's bytecode. The new field is called "file". If it is available when we call `FileReadableKVState#readFromDataSource` then we return the file from the context with priority.

This field might need to be cleared up at some point but this will probably added later when the correct place is found during testing.

Fixes https://github.com/hashgraph/hedera-mirror-node/issues/9798


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/common/ContractCallContext.java
```diff
@@ -73,6 +78,25 @@ public class ContractCallContext {
...
+     * since it was never explicitly persisted in the DB.
+     *
+     * This is the function of the fields "fileID" and "initBytecode" - to hold temporary these values
+     * during contract deploy.
+     */
+    @Setter
+    private Optional<FileID> fileID = Optional.empty();
+
+    @Setter
+    private Optional<Bytes> initBytecode = Optional.empty();
```
steven-sheehy, line 98: Instead of storing two separate fields then combining them into a File, can't we just store a single File in the context? Or are they both not set at the same time?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/FileReadableKVState.java
```diff
@@ -58,7 +58,20 @@ public FileReadableKVState(final FileDataRepository fileDataRepository, final En
     @Override
     protected File readFromDataSource(@Nonnull FileID key) {
         final var timestamp = ContractCallContext.getTimestamp();
-        final var fileId = toEntityId(key).getId();
+        final var fileEntityId = toEntityId(key);
+        final var fileId = fileEntityId.getId();
+
+        final var contextFileId = ContractCallContext.get().getFileID();
+        // If we are in a contract create case, the fileID and the init bytecode are in the ContractCallContext.
+        if (contextFileId.isPresent()
```
steven-sheehy, line 66: While this might be fine for now, it seems a bit hacky. It would be better if we could use a cache in the KV state that spans transactions such that when the file create occurs it's persisted in the cache for the next contract create. Some sort of special cache that doesn't purge between transactions originating from the same request.

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/FileReadableKVState.java
bilyana-gospodinova: Yeah, we are going to need this caching mechanism in all of out states. I will add this in the optimizations task.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/common/ContractCallContext.java
bilyana-gospodinova: I think we can. I combined the two fields into one.

steven-sheehy (approved):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/common/ContractCallContext.java
```diff
@@ -73,6 +77,24 @@ public class ContractCallContext {
...
+     * the {@link FileReadableKVState} is not able to populate the contract's bytecode from the DB
+     * since it was never explicitly persisted in the DB.
+     *
+     * This is the function of the field "file" to hold temporary the bytecode and the fileId
+     * during contract deploy.
+     */
+    @Setter
+    private Optional<File> file = Optional.empty();
+
+    ;
```
kselveliev, line 96: remove extra ;

bilyana-gospodinova: Done, thanks.

kselveliev (dismissed):

steven-sheehy (dismissed):
LGTM

steven-sheehy (approved):

PR #9801, opened 2024-11-22
AccountReadableKVState returns account with AccountNum accountId

### LINKED ISSUE #9800: AccountReadableKVState must return pbj account with AccountNum accountId
### Problem

When running tests with transaction executors flow. 
Reusable services can be checking

1. for Contract id -> 
```
        final var contract = accountStore.getContractById(body.contractIDOrThrow());
        if (contract != null) {
		        // Throws exception if accountId is of returned with type alias 
            final var contractNum = contract.accountIdOrThrow().accountNumOrThrow();
            final var mayNotExist = featureFlags.isAllowCallsToNonContractAccountsEnabled(contractsConfig, contractNum);
            validateTrue(mayNotExist || !contract.deleted(), CONTRACT_DELETED);
        }
```
Throws exception when contractId has its accountId with type AccountOneOfType.Alias
2. Checks for account id

   ```
protected Account getAliasedAccountLeaf(@NonNull final AccountID id) {
        // The Account ID may be aliased, in which case we need to convert it to a number-based account ID first.
        requireNonNull(id);
        final var accountId = lookupAliasedAccountId(id);
        return accountId == null ? null : accountState.get(accountId);
    }
```
Tries to find the aliased account first and then returns it in a form of AccountOneOfType.AccountNum (expects the accountState to return it in account num -> this is our AccountReadableKVState)

Currently our AccountReadableKVState uses toAccountId method that returns it with type ALIAS
if the entity has evm address or alias.

While we observed from the above cases that we want to return the account with AccountOneOfType.AccountNum always instead of alias.

Otherwise we will be getting around 200 integration tests failing with exception thrown while handling dispatch `java.lang.IllegalArgumentException: No account has id AccountID[Alias]`

### Solution

In AccountReadableKVState when returning the account model. 
We should set it AccountId to an AccountNum type.

### Alternatives

_No response_

### PR DESCRIPTION
**Pull request by kselveliev**
**Description**:
`AccountReadableKVState` must return the accountId of the account with a type `AccountNum`
Full issue and reason for solution described in the issue that this pr fixes

This PR modifies:
`AccountReadableKVState` - use to `AccountNumId` method for getting the accountId
`EntityIdUtils` - add `toAccountNumId` method that takes and entity parameter and return `AccountID` with `accountNum` type

**Related issue(s)**:

Fixes #9800

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-web3/src/main/java/com/hedera/services/utils/EntityIdUtils.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/services/utils/EntityIdUtils.java
@@ -188,14 +188,6 @@ public final class EntityIdUtils {
         return accountIdWithAlias != null ? accountIdWithAlias : toAccountId(entity.toEntityId());
     }
 
-    public static com.hedera.hapi.node.base.AccountID toAccountNumId(final Entity entity) {
-        if (entity == null) {
-            return com.hedera.hapi.node.base.AccountID.DEFAULT;
-        }
-
-        return toAccountId(entity.toEntityId());
-    }
-
     public static com.hedera.hapi.node.base.AccountID toAccountId(final EntityId entityId) {
         return com.hedera.hapi.node.base.AccountID.newBuilder()
                 .shardNum(entityId.getShard())
--- a/hedera-mirror-web3/src/main/java/com/hedera/services/utils/EntityIdUtils.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/services/utils/EntityIdUtils.java
@@ -188,6 +188,14 @@ public final class EntityIdUtils {
         return accountIdWithAlias != null ? accountIdWithAlias : toAccountId(entity.toEntityId());
     }
 
+    public static com.hedera.hapi.node.base.AccountID toAccountNumId(final Entity entity) {
+        if (entity == null) {
+            return com.hedera.hapi.node.base.AccountID.DEFAULT;
+        }
+
+        return toAccountId(entity.toEntityId());
+    }
+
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: hedera-mirror-web3/src/main/java/com/hedera/services/utils/EntityIdUtils.java [hunk 1]
```diff
@@ -188,6 +188,14 @@ public static com.hedera.hapi.node.base.AccountID toAccountId(final Entity entit
         return accountIdWithAlias != null ? accountIdWithAlias : toAccountId(entity.toEntityId());
     }
 
+    public static com.hedera.hapi.node.base.AccountID toAccountNumId(final Entity entity) {
```
bilyana-gospodinova, line 191: Do we really need a separate method? There are already:
1. `toAccountId(final EntityId entityId)`
2. `toAccountId(final long shard, final long realm, final long num)` and
3. `toAccountId(final Long id)`

The only place where this method is called is inside `AccountReadableKVState` -> `readFromDataSource()` -> `accountFromEntity()`. If the account entity is not found in the repository, an empty optional will be returned, so `accountFromEntity` will not be called. Is there really a case where the entity might be null and we need to return `AccountID.DEFAULT`?

kselveliev: You are actually right. And in the context of the usage entity cannot be null. So adjusted the code to now use `toAccountId(final EntityId entityId)`.

bilyana-gospodinova (approved):
LGTM

IvanKavaldzhiev (approved):
LGTM

steven-sheehy (approved):

PR #9841, opened 2024-11-27
HIP 904 add token associate on claim airdrop

### LINKED ISSUE #9865: Mirrornode fails to track airdrop token balance changes
### Description

Mirrornode fails to track account token balance changes when an account does `TokenClaimAirdrop` without an existing token association

### Steps to reproduce

- Claim a pending airdrop without an existing token association
- Query `/api/v1/accounts/{accountId}/tokens`, there will be no token association with balance info for the airdropped token

### Additional context

_No response_

### Hedera network

other

### Version

v0.118.0

### Operating system

None

edwin-greene:
**Description**:

* Associates a token when a claim airdrop occurs
* Change balance generation transaction timeout from 5m to 10m
* Handle scenario where user manually associates token before claiming

**Related issue(s)**:

Fixes #9865

**Notes for reviewer**:

Tested with claim airdrop transaction on Testnet. With this change, the token airdrop correctly appears in the `token_account` table after the claim.

**Checklist**


hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/entity/sql/SqlEntityListener.java
```diff
@@ -293,9 +296,19 @@ public void onToken(Token token) throws ImporterException {
...
+        if (tokenAccount.isClaim() && hasExistingTokenAccount(tokenAccount.getId())) {
+            return;
+        }
+
         context.merge(tokenAccount.getId(), tokenAccount, this::mergeTokenAccount);
     }
 
+    private boolean hasExistingTokenAccount(Id id) {
+        return context.get(TokenAccount.class, id) != null
+                || tokenAccountRepository.findById(id).isPresent();
```
xin-hedera, line 309: it's possible for the following transactions to happen in the same record file, assume there is no existing token association for `(A, T)`

1. `A` claims airdrop to accept some amount of token `T`
2. `A` sends all token `T` to someone else
3. `A` dissociate from token `T`
4. Another airdrop (pending) of token `T` for `A`
5. `A` claims airdrop of token `T`

We should create token association in step 1 and step 5. The logic here would only create it in step 1.

So not only we check the `TokenAccount` object existence in the context, we have to look at the associated flag too, i.e., when the associated flag is not null and is `false`, we need to merge and insert the new association.

Same applies to `TokenAccount` from the database, we need to check `associated` flag, and add such an association if it's `false`.

xin-hedera (changes requested):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/entity/sql/SqlEntityListener.java
xin-hedera: as a follow up, should rename the function to `isTokenAccountAlreadyAssociated` for better clarify.

steven-sheehy: Hopefully addressed

xin-hedera (approved):
LGTM

edwin-greene (commented):
Looks good to me

jnels124 (approved):
LGTM

PR #9863, opened 2024-12-03
Add entity type filtering in Token and Account kv states

### LINKED ISSUE #9862: Calls to HTS contracts are not properly propagated in reusable services
### Description

Specifically, contract calls that interact with HTS ERC functions like name, decimals, or totalSupply (which take a token address as a parameter) fail with CONTRACT_REVERT_EXECUTED.

For example, in the following method:

```
function name(address token) public view returns (string memory) {
    return IERC20Metadata(token).name();
}
```

When executing such calls, services tries to find the token in the method ->
`DispatchingEvmFrameState.getMutableAccount(Address address)`. 
This method searches for the token address in two stages:

Account Lookup:
Calls nativeOperations.getAccount(accountID) which checks AccountReadableKVState.
If the account is found, it is returned.

Token Lookup (if the account isn’t found):
Calls nativeOperations.getToken(number) which checks TokenReadableKVState.
If the token is found, it is returned.

The issue arises because AccountReadableKVState does not filter by entity type. If the token address is mistakenly treated as an account and returned from the account state, execution fails.

### Steps to reproduce

Explained in the description of the issue

### Additional context

_No response_

### Hedera network

other

### Version

v0.118.0

### Operating system

macOS

kselveliev:
**Description**:
This PR adds filtering for entity type in the account and token readable kv states.

TokenReadableKVState - return token only if the entity type return from entity table is of type token
AccountReadableKVState - return the account when the entity type is not token
Unit test for the new cases

**Related issue(s)**:

Fixes #9862 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/AccountReadableKVState.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/AccountReadableKVState.java
@@ -17,6 +17,7 @@
 package com.hedera.mirror.web3.state;
 
 import static com.hedera.mirror.common.domain.entity.EntityType.CONTRACT;
+import static com.hedera.mirror.common.domain.entity.EntityType.TOKEN;
 import static com.hedera.mirror.web3.state.Utils.DEFAULT_AUTO_RENEW_PERIOD;
 import static com.hedera.mirror.web3.state.Utils.parseKey;
 import static com.hedera.services.utils.EntityIdUtils.toAccountId;
@@ -92,6 +93,7 @@ public class AccountReadableKVState extends ReadableKVStateBase<AccountID, Accou
         final var timestamp = ContractCallContext.get().getTimestamp();
         return commonEntityAccessor
                 .get(key, timestamp)
+                .filter(entity -> entity.getType() != TOKEN)
                 .map(entity -> accountFromEntity(entity, timestamp))
                 .orElse(null);
     }
--- a/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/state/AccountReadableKVStateTest.java
+++ b/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/state/AccountReadableKVStateTest.java
@@ -50,6 +50,7 @@ import com.hedera.mirror.web3.repository.TokenAllowanceRepository;
 import com.hedera.mirror.web3.repository.projections.TokenAccountAssociationsCount;
 import com.hedera.pbj.runtime.OneOf;
 import com.hedera.pbj.runtime.io.buffer.Bytes;
+import com.hedera.services.utils.EntityIdUtils;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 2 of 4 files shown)


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/AccountReadableKVState.java
```diff
@@ -92,6 +93,7 @@ protected Account readFromDataSource(@Nonnull AccountID key) {
         final var timestamp = ContractCallContext.get().getTimestamp();
         return commonEntityAccessor
                 .get(key, timestamp)
+                .filter(entity -> entity.getType() != TOKEN)
```
kselveliev, line 96: Here if we use this filter `.filter(entity -> entity.getType() == ACCOUNT)` a couple of hundred tests start failing,
This is covering the case described in the issue that this pr fixes.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/state/AccountReadableKVStateTest.java
```diff
@@ -163,6 +167,14 @@ void setup() {
         entity.setMaxAutomaticTokenAssociations(MAX_AUTOMATIC_TOKEN_ASSOCIATIONS);
         entity.setType(EntityType.ACCOUNT);
 
+        token = new Entity();
+        token.setId(TOKEN_NUM);
```
IvanKavaldzhiev, line 171: To be precise we should use EntityIdUtils to construct the ID based on SHARD, REALM, TOKEN_NUM and aggregate their values. The result will equal TOKEN_NUM but once we have shards and realms the values will differ.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/state/TokenReadableKVStateTest.java
```diff
@@ -70,6 +70,8 @@ class TokenReadableKVStateTest {
 
     private static final TokenID TOKEN_ID =
             TokenID.newBuilder().shardNum(0L).realmNum(0L).tokenNum(1252L).build();
+    private static final TokenID TOKEN_ID_ACCOUNT =
```
IvanKavaldzhiev, line 73: We can just name it like TOKEN_ID_FOR_NEGATIVE_TEST or TOKEN_ID_SECOND or something like this.

steven-sheehy (dismissed):

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/state/AccountReadableKVStateTest.java
kselveliev: Comment addressed

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/state/TokenReadableKVStateTest.java
kselveliev: Comment addressed

steven-sheehy (approved):

PR #9867, opened 2024-12-04
Fix sonar issues

### LINKED ISSUE #9010: Fix code smells
### Problem

The number of code smells has spiked recently, most likely due to some rule changes on sonar side.

### Solution

* Fix as many code smells as possible:
  * Exclude web3 copied services code
  * Break into multiple PRs perhaps by fixing all smells for the largest rule first, then the next largest, etc.
  * Don't suppress any warnings locally or in the UI unless approved with the team beforehand

### Alternatives

_No response_

sdimitrov9:
**Description**:

Number of sonar issues before changes of this PR being merged:
Screenshot 2024-12-04 at 14 48 06

Replacing String.length() == 0 with String.isEmpty() to communicate the code’s intention clearly, which is to test if the string is empty. String.length() == 0 is less direct and makes the code less readable.

If all the keys in a Map are values from a single enum use EnumMap instead of HashMap. An EnumMap, which has the advantage of knowing all possible keys in advance, is more efficient compared to other implementations, as it can use a simple array as its underlying data structure.

**Related issue(s)**:

Partially fixes issue with id 9010 

**Checklist**


edwin-greene (approved):
Looks good to me

steven-sheehy (approved):
LGTM

PR #9868, opened 2024-12-04
Add additional coverage for token airdrops tests

### LINKED ISSUE #9861: Add additional coverage for token airdrop acceptance tests

nickeynikolovv:
**Description**:
This PR adds additional coverage for fungible and non-fungible token airdrops in cases when the account is not associated to the token and claims the airdrop. The association should happen upon claim.  Also token airdrops tests are extracted in separate test scenarios for cleaner overview

**Related issue(s)**:

Fixes #9861 

**Notes for reviewer**:
Currently the tests for Claim airdrops in cases when the account is not associated before are failing because of an issue in mirror node where in that case, token relationship is not created properly

**Checklist**


hedera-mirror-test/src/test/resources/features/token/token.feature
```diff
@@ -153,3 +115,79 @@ Feature: HTS Base Coverage Feature
...
+    Then the mirror node REST API should return the transaction
+    And I verify "cancelled" airdrop of <amount> tokens to CAROL
+#    CAROL is not associated to the token and is associated before claim
+    Then I airdrop <amount> tokens to CAROL
+    Then the mirror node REST API should return the transaction
+    And I associate CAROL with token
+    Then the mirror node REST API should return the transaction
+    Then CAROL claims the airdrop
+    Then the mirror node REST API should return the transaction
+    And I verify "successful" airdrop of <amount> tokens to CAROL
```
edwin-greene, line 138: Did we want to check for the TokenAccount created_timestamp? Here it would be the timestamp of the manual associate.

hedera-mirror-test/src/test/resources/features/token/token.feature
```diff
@@ -153,3 +115,79 @@ Feature: HTS Base Coverage Feature
...
+    Then I airdrop <amount> tokens to CAROL
+    Then the mirror node REST API should return the transaction
+    And I verify "successful" airdrop of <amount> tokens to CAROL
+#    DAVE is not associated to the token and the association happens upon claim
+    Then I airdrop <amount> tokens to DAVE
+    Then the mirror node REST API should return the transaction
+    And I verify "pending" airdrop of <amount> tokens to DAVE
+    Then DAVE claims the airdrop
+    Then the mirror node REST API should return the transaction
+    And I verify "successful" airdrop of <amount> tokens to DAVE
```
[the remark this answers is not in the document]
nickeynikolovv: yes -  added validation for the timestamp

hedera-mirror-test/src/test/resources/features/token/token.feature
nickeynikolovv: added validation for the timestamp

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/TokenFeature.java
```diff
@@ -621,6 +631,12 @@ public void verifyMirrorAPIResponses() {
         verifyTransactions();
     }
 
+    @Then("the mirror node REST API should return the transaction and get consensus timestamp")
+    @RetryAsserts
+    public void verifyMirrorAPIResponsesAndGetConsensusTimestamp() {
+        consensusTimestamp = verifyTransactions().getConsensusTimestamp();
```
steven-sheehy, line 637: If we make verifyTransactions() save the TransactionDetail then we can reuse `the mirror node REST API should return the transaction` instead of needing this step and consensusTimestamp. This would be more reusable.

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/TokenFeature.java
```diff
@@ -758,7 +774,7 @@ public void verifyFungibleTokenAirdrop(String status, int amount, AccountNameEnu
         var sender = accountClient.getAccount(AccountNameEnum.OPERATOR);
 
         switch (status) {
-            case "successful" -> verifySuccessfulAirdrop(tokenId, sender, receiver, amount);
+            case "successful" -> verifySuccessfulAirdrop(tokenId, sender, receiver, amount, consensusTimestamp);
```
steven-sheehy, line 777: Why do we need to pass consensusTimestamp around to some methods when it is a class variable they can access directly?

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/TokenFeature.java
```diff
@@ -1220,12 +1241,17 @@ private void verifySuccessfulAirdrop(
         assertThat(tokenRelationshipReceiver.getTokens())
                 .hasSize(1)
                 .first()
-                .returns(tokenId.toString(), TokenRelationship::getTokenId);
+                .returns(tokenId.toString(), TokenRelationship::getTokenId)
```
xin-hedera, line 1244: can we also assert `TokenRelationShip.balance` matches `amount`, same for `verifySuccessfulNftAirdrop`

We don't have a scenario where the airdrops accumulate, so it's fine to just use the airdrop's amount as the expected balance here.

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/TokenFeature.java
nickeynikolovv: Fixed it to save the `TransactioinDetail` instead of `consenusTimestamp`,  but i think we will still need separate method for saving the transaction detail, because if we have associate prior to the claim we would verify both transactions and would want to assert the consensus timestamp from the associate call.  If we save transaction detail on each `verifyTransactions` we will have only the latest `consensusTimestamp` which will be from the Claim in that case(which is not what we want to check). 
e.g.
1. associate 
2. the mirror node REST API should return the transaction
3. Claim 
4. the mirror node REST API should return the transaction
5. verify successful airdrop -> this step will fail because we will have the latest consensus timestamp from the claim but we want the one from associate

It will work only for the cases that we have implicit association upon claim

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/TokenFeature.java
nickeynikolovv: Yes we don't have to pass it. Fixed

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/TokenFeature.java
nickeynikolovv: Added

steven-sheehy (approved):
LGTM

xin-hedera (approved):
LGTM

edwin-greene (approved):

PR #9875, opened 2024-12-05
Add feature flag for the admin_key in the get network nodes response

### LINKED ISSUE #9873: Feature flag admin_key
### Problem

Admin key functionality from HIP-869 might not go out right away so we should feature flag the field in the response.

### Solution

* Add a temporary `hedera.mirror.response.includeAdminKey=false` property (no need to document)
* Adjust NetworkNodeViewModel to not return admin_key if it's false
* Adjust test application.yml to set it to true by default so specs don't need to change
* Add one new spec for when false

### Alternatives

_No response_

sdimitrov9:
**Description**:
Admin key functionality from HIP-869 might not go out right away so we should feature flag the field in the response.

**Related issue(s)**:

Fixes #9873 

**Checklist**

### DIFF
--- a/hedera-mirror-rest/viewmodel/networkNodeViewModel.js
+++ b/hedera-mirror-rest/viewmodel/networkNodeViewModel.js
@@ -17,7 +17,9 @@
 import AddressBookServiceEndpointViewModel from './addressBookServiceEndpointViewModel';
 import EntityId from '../entityId';
 import * as utils from '../utils';
+import {getIncludeAdminKey} from '../config';
 
+const includeAdminKey = getIncludeAdminKey();
 /**
  * Network node view model
  */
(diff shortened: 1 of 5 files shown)


hedera-mirror-rest/viewmodel/networkNodeViewModel.js
```diff
@@ -17,7 +17,9 @@
 import AddressBookServiceEndpointViewModel from './addressBookServiceEndpointViewModel';
 import EntityId from '../entityId';
 import * as utils from '../utils';
+import {getIncludeAdminKey} from '../config';
 
+const includeAdminKey = getIncludeAdminKey();
```
steven-sheehy, line 22: I don't think we need one function per property.

```suggestion
import config from '../config';

const includeAdminKey = config.response.includeAdminKey;
```

steven-sheehy (approved):
LGTM

PR #9882, opened 2024-12-06
Add feature flag for the admin_key in the get network nodes response …

sdimitrov9:
…(#9875)

 Admin key functionality from HIP-869 might not go out right away so we should feature flag the field in the response.


xin-hedera (approved):
LGTM

Issue #9261, opened 2024-09-03
Add feature flag for reusable services integration


IvanKavaldzhiev:
### Problem

The integration of reusable services will span through multiple releases and will be delivered in chunks. That's why we need a feature flag to disable the new code before it's fully implemented and properly tested.

### Solution

We can define a feature flag that will control which flow to be used (old mono code base or new reusable services code
base). The deviation point will be inside `ContractCallService.doProcessCall`, where:

- if the feature flag is enabled, we will call the new `TransactionExecutor.execute` method
- if the feature flag is disabled, we will call the `MirrorEvmTxProcessor.execute` method

After the new code base is fully integrated and tested, we will remove the old mono code base and the feature flag.

PR #9927, opened 2024-12-10
Add modularized property

bilyana-gospodinova:
**Description**:
The integration of reusable services will span through multiple releases and will be delivered in chunks. That's why we need a feature flag to disable the new code before it's fully implemented and properly tested.

Fixes https://github.com/hashgraph/hedera-mirror-node/issues/9261


hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
```diff
@@ -190,7 +194,15 @@ protected Entity tokenEntityPersist() {
     protected Entity accountEntityPersist() {
         return domainBuilder
                 .entity()
-                .customize(e -> e.type(EntityType.ACCOUNT).deleted(false).balance(1_000_000_000_000L))
+                .customize(e ->
+                        e.type(EntityType.ACCOUNT).evmAddress(null).alias(null).balance(1_000_000_000_000L))
```
kselveliev, line 198: Тhis will need some adjustments since we observed that is causing some test failures in some tests

bilyana-gospodinova: Comment addressed.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
```diff
@@ -106,7 +141,12 @@ protected HederaEvmTransactionProcessingResult doProcessCall(
             CallServiceParameters params, long estimatedGas, boolean restoreGasToThrottleBucket)
             throws MirrorEvmTransactionException {
         try {
-            var result = mirrorEvmTxProcessor.execute(params, estimatedGas);
+            HederaEvmTransactionProcessingResult result;
+            if (!mirrorNodeEvmProperties.isModularizedServices()) {
+                result = mirrorEvmTxProcessor.execute(params, estimatedGas);
+            } else {
+                result = processModularizedCall(params, estimatedGas);
```
steven-sheehy, line 148: Can we have at least one test in ContractCallServiceTest where modularized is enabled so we get code coverage? Whatever the simplest test is that can pass.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
```diff
@@ -118,6 +158,145 @@ protected HederaEvmTransactionProcessingResult doProcessCall(
         }
     }
 
+    private Map<String, String> buildTransactionExecutorProperties() {
```
steven-sheehy, line 161: We can create this once and reuse between calls to improve performance.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
```diff
@@ -118,6 +158,145 @@ protected HederaEvmTransactionProcessingResult doProcessCall(
...
+        return mirrorNodeProperties;
+    }
+
+    private HederaEvmTransactionProcessingResult buildSuccessResult(
+            final boolean isContractCreate,
+            final List<SingleTransactionRecord> receipt,
+            final CallServiceParameters params) {
+        var result = isContractCreate
+                ? receipt.getFirst().transactionRecord().contractCreateResult()
+                : receipt.getFirst().transactionRecord().contractCallResult();
```
steven-sheehy, line 180: This repeats. Can move to a separate function. Or at least move up to calling function.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
```diff
@@ -118,6 +158,145 @@ protected HederaEvmTransactionProcessingResult doProcessCall(
...
+                                params.getCallData().toArray()))
+                        .gas(estimatedGas)
+                        .build())
+                .nodeAccountID(TREASURY_ACCOUNT_ID)
+                .transactionID(TransactionID.newBuilder()
+                        .transactionValidStart(TRANSACTION_START)
+                        .accountID(TREASURY_ACCOUNT_ID)
+                        .build())
+                .transactionValidDuration(TRANSACTION_DURATION)
+                .build();
```
steven-sheehy, line 256: We can statically create the builder with all defaults then use to set custom fields per request. Same for other transaction bodies.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
```diff
@@ -118,6 +158,145 @@ protected HederaEvmTransactionProcessingResult doProcessCall(
...
+                                        params.getReceiver().toArrayUnsafe()))
+                                .build())
+                        .functionParameters(com.hedera.pbj.runtime.io.buffer.Bytes.wrap(
+                                params.getCallData().toArray()))
+                        .gas(estimatedGas)
+                        .build())
+                .nodeAccountID(TREASURY_ACCOUNT_ID)
+                .transactionID(TransactionID.newBuilder()
+                        .transactionValidStart(TRANSACTION_START)
+                        .accountID(TREASURY_ACCOUNT_ID)
```
steven-sheehy, line 253: Shouldn't this be the sender?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
```diff
@@ -118,6 +158,145 @@ protected HederaEvmTransactionProcessingResult doProcessCall(
...
+                .contractCall(ContractCallTransactionBody.newBuilder()
+                        .contractID(ContractID.newBuilder()
+                                .evmAddress(com.hedera.pbj.runtime.io.buffer.Bytes.wrap(
+                                        params.getReceiver().toArrayUnsafe()))
+                                .build())
+                        .functionParameters(com.hedera.pbj.runtime.io.buffer.Bytes.wrap(
+                                params.getCallData().toArray()))
+                        .gas(estimatedGas)
+                        .build())
+                .nodeAccountID(TREASURY_ACCOUNT_ID)
```
steven-sheehy, line 250: It probably doesn't matter, but we should use an actual node account like 0.0.3

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
```diff
@@ -118,6 +158,145 @@ protected HederaEvmTransactionProcessingResult doProcessCall(
...
+        final var maxLifetime =
+                DEFAULT_CONFIG.getConfigData(EntitiesConfig.class).maxLifetime();
+        var executor = TransactionExecutors.TRANSACTION_EXECUTORS.newExecutor(
+                mirrorNodeState, buildTransactionExecutorProperties(), null);
+
+        TransactionBody transactionBody;
+        HederaEvmTransactionProcessingResult result;
+        if (isContractCreate) {
+            // Upload the init bytecode
+            transactionBody = buildFileCreateTransactionBody(params, maxLifetime);
```
steven-sheehy, line 271: We should use `ContractCreateTransactionBody.initcode` to create as it will be quicker than having to process multiple transactions. This will also avoid need for `File` in context. Will `initcode` fail if exceeding 6 KiB max transaction size? Even if it does, we should use it for the cases when it's below max size.

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
bilyana-gospodinova: I will cache this value in a field. So far the properties are always the same but in the future when we get to the historical tests we will need to pass different collections with different properties depending on the EVM version, so this method might have to change again.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
bilyana-gospodinova: Is there any specific setup for a node account? I might need to change the integration tests setup, if so.

steven-sheehy: Does this incur an extra account load for every transaction? That is, if we change this to 0.0.3 and have payer as 0.0.2 does it load both accounts? If so, we might not actually want to have them separate. We should use the minimal number of accounts and even try to use an account from the request itself.

hedera-mirror-web3/build.gradle.kts
```diff
@@ -32,7 +32,7 @@ dependencies {
     implementation(platform("org.springframework.cloud:spring-cloud-dependencies"))
     implementation(project(":common"))
     implementation("com.bucket4j:bucket4j-core")
-    implementation("com.esaulpaugh:headlong")
+    implementation("com.esaulpaugh:headlong:6.1.1")
```
kselveliev, line 35: Why do we need to specify 6.1.1 version. I see that latest version is 12.

bilyana-gospodinova: This is the version that is used in the dependency. If we don't specify it here, we cannot decode successfully the response hex tuples.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
```diff
@@ -180,6 +180,9 @@ public class MirrorNodeEvmProperties implements EvmProperties {
     @Min(1)
     private int feesTokenTransferUsageMultiplier = 380;
 
+    @Getter
+    private boolean modularizedServices;
```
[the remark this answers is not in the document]
bilyana-gospodinova: Documented.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
bilyana-gospodinova: Added a method with lazy loading.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
bilyana-gospodinova: Extracted in a separate method.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
bilyana-gospodinova: Correct. We don't always have a sender, since the field is optional when the `value` in the request is 0. I changed the code to set the sender if available and use the default treasury account, if optional and not set.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
bilyana-gospodinova: Comment addressed. Added a check for the initcode length and added a new logic that uploads the initcode directly without using a file.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/TransactionExecutionService.java
```diff
@@ -0,0 +1,281 @@
...
+                ContractCallContext.get().setFile(Optional.of(file));
+
+                // Create the contract with the init bytecode
+                transactionBody =
+                        buildContractCreateTransactionBodyWithFileID(params, fileID, estimatedGas, maxLifetime);
+            }
+        } else {
+            transactionBody = buildContractCallTransactionBody(params, estimatedGas);
+        }
+        List<OperationTracer> operationTracers = new ArrayList<>();
```
steven-sheehy, line 120: Avoid allocating the list unless there is a non-null tracer.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/TransactionExecutionService.java
```diff
@@ -0,0 +1,281 @@
...
+        if (ContractCallContext.get().getOpcodeTracerOptions() != null) {
+            operationTracers.add(opcodeTracer);
+        }
+
+        var receipt =
+                executor.execute(transactionBody, Instant.EPOCH, operationTracers.toArray(OperationTracer[]::new));
+        if (receipt.getFirst().transactionRecord().receiptOrThrow().status() == ResponseCodeEnum.SUCCESS) {
```
steven-sheehy, line 127: We call `receipt.getFirst().transactionRecord()` 4 times and don't call anything else on `receipt`. Store the first transactionRecord into a var once and pass everywhere.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
```diff
@@ -82,8 +108,24 @@ public static Key getKeyWithContractId(final Contract contract) {
...
+                .entity()
+                .customize(e -> e.id(3L).num(3L).balance(5000000000000000000L))
+                .persist();
+        domainBuilder.entity().customize(e -> e.id(800L).num(800L)).persist();
+        domainBuilder.entity().customize(e -> e.id(801L).num(801L)).persist();
+        domainBuilder.entity().customize(e -> e.id(98L).num(98L)).persist();
+        domainBuilder
+                .fileData()
+                .customize(f -> f.entityId(EntityId.of(112L)).fileData(EXCHANGE_RATES_SET))
+                .persist();
```
steven-sheehy, line 128: This is a lot of setup work for every test that can slow down tests overall. Do we really need all these?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
bilyana-gospodinova: I thought about that but I have some concerns. If we set:
`private static final ContractCreateTransactionBody.Builder CONTRACT_CREATE = ...`
as a field and then in each transaction:
`CONTRACT_CREATE.initcode(params.getCallData()...).build()`
each call will override the initcode value in the builder since they share the same object. This means that we will need to use synchronization which will also slow the performance down.

The other option is to set:
`private static final ContractCreateTransactionBody CONTRACT_CREATE = ...`
and then in each transaction we can call `CONTRACT_CREATE.copyBuilder().initcode(params.getCallData()...).build()` but this copyBuilder will create a new object as well. So we don't gain anything if we compare this approach to the current one.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/TransactionExecutionService.java
kselveliev: Comment addressed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
bilyana-gospodinova: I tested it and it seems that the max transaction size is 6144 bytes. For contracts over this size, the contract create transaction fails on pre-check. The byte value is hardcoded and not configurable. So for larger contracts it seems that we still need to have the file upload logic.

steven-sheehy: Is the 6144 a setting? If so, can just increase it.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/TransactionExecutionService.java
```diff
@@ -0,0 +1,278 @@
...
+            result = buildSuccessResult(isContractCreate, receipt, params);
+        } else {
+            result = buildFailedResult(receipt, isContractCreate);
+        }
+        return result;
+    }
+
+    private Map<String, String> getTransactionProperties() {
+        if (transactionProperties == null) {
+            final var mirrorNodeProperties = mirrorNodeEvmProperties.getProperties();
```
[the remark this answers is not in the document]
bilyana-gospodinova: Comment addressed.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/AccountReadableKVState.java
```diff
@@ -88,6 +90,12 @@ public AccountReadableKVState(
         this.tokenAllowanceRepository = tokenAllowanceRepository;
     }
 
+    @Nullable
+    @Override
+    public Account get(@NonNull AccountID key) {
+        return readFromDataSource(key);
+    }
```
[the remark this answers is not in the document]
bilyana-gospodinova: Comment addressed.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/MirrorNodeState.java
```diff
@@ -132,19 +130,13 @@ private void init() {
...
-                        fileId,
-                        File.newBuilder()
-                                .fileId(fileId)
-                                .keys(KeyList.DEFAULT)
-                                .contents(provider.apply(bootstrapConfig))
-                                .build());
-            });
-            ((CommittableWritableStates) fileServiceStates).commit();
+            final var accountReadableKVState = (AccountReadableKVState) readableKVStates.stream()
+                    .filter(r -> r.getStateKey().equals("ACCOUNTS"))
```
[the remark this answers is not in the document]
bilyana-gospodinova: Comment addressed.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/TransactionExecutionService.java
bilyana-gospodinova: Comment addressed.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/TransactionExecutionService.java
```diff
@@ -0,0 +1,264 @@
...
+        return (AccountID) mirrorNodeState
+                .getReadableStates(TokenService.NAME)
+                .get(AliasesReadableKVState.KEY)
+                .get(convertAddressToProtoBytes(senderAddress));
+    }
+
+    private OperationTracer[] getOperationTracers() {
+        return ContractCallContext.get().getOpcodeTracerOptions() != null
+                ? new OperationTracer[] {opcodeTracer}
+                : new OperationTracer[0];
```
steven-sheehy, line 249: Can store empty array as a static constant.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/MirrorNodeState.java
```diff
@@ -131,20 +130,7 @@ private void init() {
...
-                files.put(
-                        fileId,
-                        File.newBuilder()
-                                .fileId(fileId)
-                                .keys(KeyList.DEFAULT)
-                                .contents(provider.apply(bootstrapConfig))
-                                .build());
-            });
-            ((CommittableWritableStates) fileServiceStates).commit();
+            // point.
```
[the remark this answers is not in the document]
bilyana-gospodinova: Removed, thanks.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
bilyana-gospodinova: I removed some of them. The others can be added just for the tests/suits that use them.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
bilyana-gospodinova: I opened an issue in the services repo to request to make it configurable - https://github.com/hashgraph/hedera-services/issues/17099.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
bilyana-gospodinova: Sure. Comment addressed.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
bilyana-gospodinova: Comment addressed.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/TransactionExecutionService.java
kselveliev: Comment addressed

steven-sheehy (approved):
LGTM

PR #9929, opened 2024-12-10
Fix `ContractCallServicePrecompileReadonlyTest` tests

bilyana-gospodinova:
**Description**:
Fix the failing tests in `ContractCallServicePrecompileReadonlyTest` when the modularized flag is set to `true`.

1. Some methods were added to `Token` and `TokenRelation` because of the `NoSuchMethodException`.
2. Some tests needed setup adjustments - persist the entities with aliases where we needed to pass an alias to a function. Also, the HRC precompile is now supported, so adjust the assert with the modularized flag.


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/ContractBytecodeReadableKVState.java
```diff
@@ -27,7 +27,9 @@
 import com.hedera.mirror.web3.repository.ContractRepository;
 import com.hedera.pbj.runtime.io.buffer.Bytes;
 import com.swirlds.state.spi.ReadableKVStateBase;
+import edu.umd.cs.findbugs.annotations.NonNull;
```
kselveliev, line 30: Change to jakarta annotation

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/ContractStorageReadableKVState.java
```diff
@@ -23,7 +23,9 @@
 import com.hedera.pbj.runtime.io.buffer.Bytes;
 import com.hedera.services.utils.EntityIdUtils;
 import com.swirlds.state.spi.ReadableKVStateBase;
+import edu.umd.cs.findbugs.annotations.NonNull;
```
kselveliev, line 26: same as above comment

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/ContractBytecodeReadableKVState.java
bilyana-gospodinova: Changed, thanks.

kselveliev (dismissed):

kselveliev (approved):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/Utils.java
```diff
@@ -48,13 +48,15 @@ public static Key parseKey(final byte[] keyBytes) {
...
      *
-     * @param timestamp The timestamp in milliseconds.
+     * @param timestamp The timestamp in nanoseconds.
      * @return The PBJ Timestamp object.
      */
     public static Timestamp convertToTimestamp(final long timestamp) {
-        var instant = Instant.ofEpochMilli(timestamp);
+        final long seconds = timestamp / 1_000_000_000;
+        final int remainingNanos = (int) (timestamp % 1_000_000_000);
+        var instant = Instant.ofEpochSecond(seconds, remainingNanos);
```
steven-sheehy, line 59:
```suggestion
        var instant = Instant.ofEpochSecond(0, timestamp);
```

bilyana-gospodinova: We can't really do that, since the only constructor for `Timestamp` is `Timestamp(long seconds, int nanos)` and the `timestamp` parameter is a `long`.

steven-sheehy: Not sure I follow. Your current code uses an Instant to parse seconds and nanos to pass to Timestamp. My suggestion continues this but uses the built-in parsing of nanoseconds in Instant that we use everywhere else.

steven-sheehy: Or alternatively we can skip the Instant and just use the division and mod logic. But we should probably use Math.floorDiv and floorMod if we do that.

bilyana-gospodinova: I misunderstood initially but I got it now. Comment addressed. Thanks.

steven-sheehy (approved):

PR #9943, opened 2024-12-12
feat: Wrap modularized code in transactionExecutionService

### LINKED ISSUE #9830: Adapt OpcodeService for reusable services
### Problem

Currently when running opcode service tests with the reusable services TransactionExecutor tests are failing with the following exception -> `ContractCallContext.getOpcodes()" because "this.contextCaptor" is null`
We need to adapt the OpcodeService and the OpcodeServiceTests to be able to work with the reusable services.

### Solution

Refine OpcodeService flow so that it works with the TransactionExecutor

### Alternatives

_No response_

kselveliev:
**Description**:

This PR wraps modularized execution and logic into it's own standalone `TransactionExecutionService`.
Doing this the adaptation of the `OpcodeServiceTest` arguments and result captors becomes incredibly easy.
Also added opcodeTracer when the call is coming from an opcode call.

Main changes:
Added `TransactionExecutionService` class
Which contains the new modularized execution logic in itself. The code is moved from `ContractCallService` into `TransactionExecutionService` with the only addition of the opcodeTracer if the opcode tracer options are set.

`AbstractContractCallServiceOpcodeTracerTest` - now check based on the modularized feature flag to capture arguments and result either from `mirrorEvmTxProcessor.execute` or `transactionExecutionService.execute`

`TransactionExecutionServiceTest` - adds coverage for the newly created class.

Fixes #9830 

**Notes for reviewer**:

**Checklist**


hedera-mirror-web3/src/test/resources/config/application.yml
```diff
@@ -2,6 +2,7 @@ hedera:
   mirror:
     web3:
       evm:
+        modularizedServices: false
```
bilyana-gospodinova, line 5: This is not needed as by the default the value is false.

kselveliev: Correct, comment addressed

bilyana-gospodinova (approved):
LGTM

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/TransactionExecutionService.java
```diff
@@ -0,0 +1,228 @@
...
+        final var maxLifetime =
+                DEFAULT_CONFIG.getConfigData(EntitiesConfig.class).maxLifetime();
+        var executor = ExecutorFactory.newExecutor(mirrorNodeState, buildTransactionExecutorProperties(), null);
+
+        TransactionBody transactionBody;
+        HederaEvmTransactionProcessingResult result;
+        if (isContractCreate) {
+            // Upload the init bytecode
+            transactionBody = buildFileCreateTransactionBody(params, maxLifetime);
+            var uploadReceipt = executor.execute(transactionBody, Instant.EPOCH);
```
IvanKavaldzhiev, line 86: Do we need this calls? What if we use the next available FileID in the mirror node, as a value?

kselveliev: This can be addressed in the base [PR](https://github.com/hashgraph/hedera-mirror-node/pull/9927) since this pr is just copy paste the code from the modularized property pr

steven-sheehy (approved):

kselveliev:
Merging any other changes will be addressed in the https://github.com/hashgraph/hedera-mirror-node/pull/9927 PR

PR #9944, opened 2024-12-12
Fix sonar issues

sdimitrov9:
**Description**:
Fix issues detected by the sonar scanner.

**Related issue(s)**:

Partially fixes 9010

**Checklist**

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -281,8 +281,9 @@ fun replaceVersion(files: String, match: String) {
 
 tasks.nodeSetup { onlyIf { !this.nodeDir.get().asFile.exists() } }
 
-// Replace release version in files
 tasks.register("release") {
+    description = "Replaces release version in files."
+    group = "Versioning"
     doLast {
         replaceVersion("charts/**/Chart.yaml", "(?<=^(appVersion|version): ).+")
         replaceVersion("docker-compose.yml", "(?<=gcr.io/mirrornode/hedera-mirror-.+:).+")
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -82,6 +82,7 @@ dependencies {
         api("io.vertx:vertx-codegen:$vertxVersion")
         api("io.vertx:vertx-core:$vertxVersion")
         api("jakarta.inject:jakarta.inject-api:2.0.1")
+        api("javax.inject:javax.inject:1")
         api("net.devh:grpc-spring-boot-starter:3.1.0.RELEASE")
         api("net.java.dev.jna:jna:5.15.0")
         api("org.apache.commons:commons-collections4:4.4")
@@ -283,7 +284,7 @@ tasks.nodeSetup { onlyIf { !this.nodeDir.get().asFile.exists() } }
 
 tasks.register("release") {
     description = "Replaces release version in files."
-    group = "Versioning"
+    group = "release"
     doLast {
         replaceVersion("charts/**/Chart.yaml", "(?<=^(appVersion|version): ).+")
         replaceVersion("docker-compose.yml", "(?<=gcr.io/mirrornode/hedera-mirror-.+:).+")
--- a/hedera-mirror-test/build.gradle.kts
+++ b/hedera-mirror-test/build.gradle.kts
@@ -70,6 +70,8 @@ dependencies {
 tasks.named("test") { enabled = false }
 
 tasks.register<Test>("acceptance") {
+    description = "Acceptance tests configuration"
+    group = "Testing"
     val maxParallelism = project.property("maxParallelism") as String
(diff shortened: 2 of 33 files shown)


hedera-mirror-web3/build.gradle.kts
```diff
@@ -28,6 +28,8 @@ plugins {
     id("spring-conventions")
 }
 
+val javaxInjectVersion = "1"
```
steven-sheehy, line 31: The correct fix is to move to root build.gradle.kts as `api("javax.inject:javax.inject:1")` then drop the version here.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/NftAllowanceRepository.java
```diff
@@ -73,4 +73,47 @@ and lower(timestamp_range) <= :blockTimestamp
...
+                            ) as nft_allowance_history
+                        ) as row_numbered_data
+                        where row_number = 1
+                    )
+                    select *
+                    from nft_allowances
+                    order by timestamp_range desc
+                    """,
+            nativeQuery = true)
+    List<NftAllowance> findByOwnerAndTimestampAndApprovedForAllIsFalse(long owner, long blockTimestamp);
```
steven-sheehy, line 118: Why are we adding this new repository method?

hedera-mirror-test/build.gradle.kts
```diff
@@ -70,6 +70,8 @@ dependencies {
 tasks.named("test") { enabled = false }
 
 tasks.register<Test>("acceptance") {
+    description = "Acceptance tests configuration"
+    group = "Testing"
```
steven-sheehy, line 74: Since test task is under verification group, it makes sense to also have acceptance there.

```suggestion
    group = "verification"
```

build.gradle.kts
```diff
@@ -281,8 +281,9 @@ fun replaceVersion(files: String, match: String) {
 
 tasks.nodeSetup { onlyIf { !this.nodeDir.get().asFile.exists() } }
 
-// Replace release version in files
 tasks.register("release") {
+    description = "Replaces release version in files."
+    group = "Versioning"
```
steven-sheehy, line 286:
```suggestion
    group = "release"
```

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/NftAllowanceRepository.java
sdimitrov9: Removed. Now an additional parameter is added to the existing repository method so we can set the approved_for_all property. In order not to modify the existing method calls across the classes a default method with the original signature is added as well:

`  
...         
nativeQuery = true)
    List<NftAllowance> findByOwnerAndTimestamp(long owner, long blockTimestamp, boolean approvedForAll);

    default List<NftAllowance> findByOwnerAndTimestampAndApprovedForAllIsTrue(long owner, long blockTimestamp) {
        return findByOwnerAndTimestamp(owner, blockTimestamp, true);
    }`

hedera-mirror-web3/build.gradle.kts
sdimitrov9: Fixed.

hedera-mirror-test/build.gradle.kts
sdimitrov9: Fixed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/NftAllowanceRepository.java
```diff
@@ -61,7 +61,7 @@ and lower(timestamp_range) <= :blockTimestamp
                                 select *
                                 from nft_allowance_history
                                 where owner = :owner
-                                    and approved_for_all = true
+                                    and approved_for_all <= :approvedForAll
```
steven-sheehy, line 64: This now introduces a bug in production code where we're not filtering the nft_allowance table as well. Suggestion is to remove all changes to repository class. We shouldn't add new repository methods that are only used by tests. Instead just use findAll/findById to lookup and assert the approveForAll field in code.

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/NftAllowanceRepository.java
sdimitrov9: Fixed

steven-sheehy (approved):
LGTM

Issue #9977, opened 2024-12-17
Web3 does not support a gas throttle limit higher than the default (1B)


Neurone:
### Description

MN does not allow you to set a gas throttling limit (`hedera.mirror.web3.throttle.gasPerSecond`) higher than the default (1B).

For example, if you set `hedera.mirror.web3.throttle.gasPerSecond` to 10B, the web3 startup error is the following:

```zsh
web3-1   | Caused by: org.springframework.beans.BeanInstantiationException: Failed to instantiate [io.github.bucket4j.Bucket]: Factory method 'gasLimitBucket' threw exception with message: 10 token/nanosecond is not permitted refill rate, because highest supported rate is 1 token/nanosecond
```

Using just a single unit of GAS in addition to the default 1_000_000_001 (1B + 1), the error is:

```zsh
web3-1   | Caused by: org.springframework.beans.BeanInstantiationException: Failed to instantiate [io.github.bucket4j.Bucket]: Factory method 'gasLimitBucket' threw exception with message: 1 token/nanosecond is not permitted refill rate, because highest supported rate is 1 token/nanosecond
```

The problem appears during the [bucket4j initialization](https://github.com/hashgraph/hedera-mirror-node/blob/26878b734d978420c5ecb6cf0cfc08f96109b783/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/config/ThrottleConfiguration.java#L52) and in particular it seems to be related to how bucket4j handles buckets and how the refill operation works. It seems that the refill operation [by design cannot work with more than 1 token/nanosecond](https://github.com/bucket4j/bucket4j/blob/d7aa6744afdf39dcb1b83abcbc06a26631adff50/bucket4j-core/src/main/java/io/github/bucket4j/BucketExceptions.java#L171) (whatever that means ^^).

### Steps to reproduce

1. Set `hedera.mirror.web3.throttle.gasPerSecond` to 10_000_000_000
2. Verify the error appears during web3 startup

### Hedera network

mainnet, testnet, previewnet

### Version

v.0.119.0

### Operating system

None

PR #9994, opened 2024-12-20
Introduce `gasUnit` property to allow greater gas per second throttle

bilyana-gospodinova:
**Description**:
The current bucket4j implementation does not allow a greater consumption than 1 token per nanosecond. Because of this the current max gas per second is capped at 1_000_000_000 gas.

This PR adds a new property called `hedera.mirror.web3.throttle.gasUnit`. This is a multiplier that allows a mapping between the token and the gas that will be consumed in the bucket. For example, if the value of the property is set to 10, this means that 1 token in the bucket will correspond to 10 gas. This will make the max supported gas limit per second to be equal to 10_000_000_000. Please note that there might be some small precision loss due to number rounding.

The implementation change is that when a request is made, the passed gas limit is divided by the configured `gasUnit` property value (by default it is set to 1). After the transaction processing is finished, the remaining (unused) gas is then returned back to the bucket and is available to be used by another transaction.
Example: a request is made with gasLimit set to 1_000_000. The `gasUnit` is set to 10. This means that 100_000 tokens will be reserved in the gas throttle bucket. The transaction finishes and the actual gas used is 800_000. This means that there are 200_000 gas remaining. Using the `gasUnit` multiplier, 200_000 / `gasUnit` = 20_000 gas will be returned back to the bucket and can be used by another transaction.

Fixes https://github.com/hashgraph/hedera-mirror-node/issues/9977


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/throttle/ThrottleProperties.java
```diff
@@ -34,8 +34,13 @@ public class ThrottleProperties {
 
     @Getter
     @Min(21_000)
+    @Max(1_000_000_000)
     private long gasPerSecond = 1_000_000_000L;
 
+    @Getter
+    @Min(1)
```
kselveliev, line 41: Probably we can also add a max for this as well like 10 or bigger.

bilyana-gospodinova: The max can be as much as the environment could handle. I am not sure at how much we should cap and if we need to cap it at all.

kselveliev (approved):

steven-sheehy (approved):

xin-hedera (approved):
LGTM

PR #10028, opened 2024-12-31
Return long term scheduled transaction in get transactions by id REST API

### LINKED ISSUE #10027: `/api/v1/transactions/{id}` will not show long term scheduled transaction executed more than 35 minutes later
### Description

[HIP-423](https://github.com/hashgraph/hedera-improvement-proposal/blob/master/HIP/hip-423.md) extends scheduled transaction's lifetime to beyond 30 minutes.

This means mirrornode REST API `/api/v1/transactions/{transactionId}`'s heuristic hint that consensus timestamp should be no later than 35 minutes after the transaction valid start timestamp will fail to find such long term scheduled transactions.

### Steps to reproduce

1. ScheduleCreate, https://testnet.mirrornode.hedera.com/api/v1/transactions/0.0.4528042-1734873411-044799761
2. Scheculed crypto transfer transaction, https://testnet.mirrornode.hedera.com/api/v1/transactions?timestamp=1735576935.976035001
3. Note, this returns not found, https://testnet.mirrornode.hedera.com/api/v1/transactions/0.0.4528042-1734873411-044799761?scheduled=true

### Additional context

_No response_

### Hedera network

testnet

### Version

0.120.0

### Operating system

None

xin-hedera:
**Description**:

This PR fixes the bug that get transactions by id REST API doesn't return scheduled transactions executed more than 35 mins after transaction valid start timestamp

- Query again with max long term scheduled transaction timestamp range if the result may miss such transactions
- Fix passing incorrect value to `getTransactionsByIdOrHashCacheControlHeader`
- Fix `getTransactionsByIdOrHashCacheControlHeader` unit tests

**Related issue(s)**:

Fixes #10027 

**Notes for reviewer**:

**Checklist**


hedera-mirror-rest/config/application.yml
```diff
@@ -67,6 +67,7 @@ hedera:
       port: 5551
       query:
         bindTimestampRange: false
+        maxLongTermScheduledTransactionConsensusTimestampRange: 89285m
```
xin-hedera, line 70: services default max expiration seconds for long term scheduled transaction is `5356800`, i.e., 62 days, we add an extra 5 minutes because

- the range is on top of the transaction's valid start timestamp
- long term scheduled transactions if configured so, are guaranteed to run after the given expiration timestamp, not  before

hedera-mirror-rest/transactions.js
```diff
@@ -840,8 +839,16 @@ const getTransactionsByIdOrHash = async (req, res) => {
     filters
   );
 
-  // Execute query
-  const {rows} = await pool.queryQuietly(query, params);
+  const {rows} = await pool.queryQuietly(query, params).then((result) => {
+    if (!mayMissLongTermScheduledTransaction(isTransactionHash, scheduled, result.rows)) {
```
xin-hedera, line 843: only when there's possibility that the result has missed a long term scheduled transaction, we run the query with much larger consensus timestamp range. Doing so avoids the performance penalty (scanning more timestamp partitions) in most cases.

hedera-mirror-rest/transactions.js
```diff
@@ -793,39 +795,36 @@ const extractSqlFromTransactionsByIdOrHashRequest = async (transactionIdOrHash,
     mainConditions.push(`${Transaction.SCHEDULED} = $${params.length}`);
   }
   mainConditions.unshift(...commonConditions);
+
   return {
+    isTransactionHash,
     query: getTransactionQuery(mainConditions.join(' and '), commonConditions.join(' and ')),
     params,
-    scheduledParamExists: scheduledParamExists,
```
xin-hedera, line 799: the key was `scheduledParamExists`, however in the caller, it was extracted as `scheduled`, so the value for `scheduled` in the caller is always `undefined` regardless of the value of `scheduledParamExists`

hedera-mirror-rest/__tests__/specs/transactions/{id}/scheduled-no-filter.json
```diff
@@ -52,10 +38,7 @@
       }
     ]
   },
-  "urls": [
-    "/api/v1/transactions/0.0.10-1234567890-000000001",
-    "/api/v1/transactions/0x0102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f202122232425262728292a2b2c2d2e2f30"
```
xin-hedera, line 57: incorrect logic, scheculecreate and the corresponding scheduled transaction should always have different transaction hashes, so it makes no sense to test with get transaction by transaction hash url.

edwin-greene (dismissed):
Looks good to me

jnels124 (approved):
LGTM.

docs/configuration.md
```diff
@@ -504,82 +504,83 @@ recursively merged into) the current configuration:
...
+| `hedera.mirror.rest.port`                                                         | 5551                    | The REST API port                                                                                                                                                                             |
+| `hedera.mirror.rest.query.bindTimestampRange`                                     | false                   | Whether to bind the timestamp range to maxTimestampRange                                                                                                                                      |
+| `hedera.mirror.rest.query.maxLongTermScheduledTransactionConsensusTimestampRange` | 89285m                  | The maximum amount of time of a long term scheduled transaction's consensus timestamp from its valid start timestamp.                                                                         |
```
steven-sheehy, line 541: This is quite a long name. Maybe we can shorten to `maxScheduledTransactionTimestampRange`?

xin-hedera: sounds good

steven-sheehy (approved):
LGTM

PR #10059, opened 2025-01-06
chore: Add docker buildkit configuration for docker buildx support

### LINKED ISSUE #10054: Add CI Configuration for the front end caching registry
Add buildkit configuration steps to prevent rate limiting

rbarkerSL:
**Description**:

Adds docker buildx build-kit configuration for front end caching.

**Related issue(s)**:

Fixes #10054 

**Checklist**

### DIFF
--- a/.github/workflows/release-integration.yml
+++ b/.github/workflows/release-integration.yml
@@ -79,6 +79,12 @@ jobs:
 
       - name: Setup Docker Buildx
         uses: docker/setup-buildx-action@6524bf65af31da8d45b59e8c27de4bd072b392f5 # v3.8.0
+        with:
+          version: v0.16.2
+          driver-opts: network=host
+          buildkitd-config-inline: |
+            [registry."docker.io"]
+              mirrors = ["https://hub.mirror.docker.lat.ope.eng.hashgraph.io"]
 
       - name: Build and push images
         uses: docker/build-push-action@48aba3b46d1b1fec4febb7c5d0c644b249a11355 # v6.10.0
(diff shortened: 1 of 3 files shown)


nathanklick (approved):

.github/workflows/release-integration.yml
```diff
@@ -79,6 +79,12 @@ jobs:
 
       - name: Setup Docker Buildx
         uses: docker/setup-buildx-action@6524bf65af31da8d45b59e8c27de4bd072b392f5 # v3.8.0
+        with:
+          version: v0.16.2
```
steven-sheehy, line 83: Why do we need a hardcoded older version of buildx? This means whenever dependabot automatically bumps setup-buildx-action to a newer version we won't get an updated buildx.

.github/workflows/release-integration.yml
```diff
@@ -79,6 +79,12 @@ jobs:
...
+          driver-opts: network=host
+          buildkitd-config-inline: |
+            [registry."docker.io"]
+              mirrors = ["https://hub.mirror.docker.lat.ope.eng.hashgraph.io"]
```
steven-sheehy, line 87: This is a really weird, long DNS name. Are we sure it's correct?

rbarkerSL: Per @nathanklick this is where we should be mirroring to.

.github/workflows/release-integration.yml
rbarkerSL: That's a good point. We should be able to go without the version specificity. Will update.

nathanklick (approved):

.github/workflows/release-integration.yml
rbarkerSL: For the future: We will need to add a hard-coded version if/when mirror node adopts deterministic builds.

Issue #10052, opened 2025-01-06
Fix `RevertTest` tests


bilyana-gospodinova:
### Problem

As part of the integration of the hedera app dependency, the `RevertTest` tests are now failing as the revert message reading is not adapted.

### Solution

Fix `RevertTest` tests.

PR #10063, opened 2025-01-07
Fix `RevertTest` tests

bilyana-gospodinova:
**Description**:
Fixed some failing tests in the `RevertTest` suite. Added an implementation to read the error messages from the returned transaction records.

There are 2 types of messages:
- Encoded messages such as: "0x08c379a000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000013536f6d6520726576657274206d65737361676500000000000000000000000000" - function selector + reason
- error statuses such as: `INVALID_TOKEN_ID`
For the second case the `HederaEvmTransactionProcessingResult` class needed to be extended to handle messages that are not in byte format as the additional byte encoding and decoding was not needed.

Fixes [#10052](https://github.com/hashgraph/hedera-mirror-node/issues/10052)


hedera-mirror-web3/src/main/java/com/hedera/node/app/service/evm/contracts/execution/HederaEvmTransactionProcessingResult.java
```diff
@@ -0,0 +1,185 @@
...
+import org.apache.tuweni.bytes.Bytes;
+import org.hyperledger.besu.datatypes.Address;
+import org.hyperledger.besu.evm.frame.ExceptionalHaltReason;
+import org.hyperledger.besu.evm.log.Log;
+
+/**
+ * Model object holding all the necessary data to build and externalise the result of a single EVM
+ * transaction
+ */
+public class HederaEvmTransactionProcessingResult {
```
kselveliev, line 32: Here can't we just extend the original HederaEvmTransactionProcessingResult and in the new class mainly add the logic with the decodedRevertReason?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/TransactionExecutionService.java
```diff
@@ -216,6 +230,7 @@ private TransactionBody buildContractCallTransactionBody(
                         .functionParameters(com.hedera.pbj.runtime.io.buffer.Bytes.wrap(
                                 params.getCallData().toArrayUnsafe()))
                         .gas(estimatedGas)
+                        .amount(params.getValue())
```
kselveliev, line 233: Do we need this change for these specific tests fixes. I removed this line and they still pass. Maybe it is a change needed for other tests.

hedera-mirror-web3/src/main/java/com/hedera/node/app/service/evm/contracts/execution/HederaEvmTransactionProcessingResult.java
bilyana-gospodinova: Comment addressed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/TransactionExecutionService.java
bilyana-gospodinova: Correct - it seems that this change is not needed but not for this PR.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/TransactionExecutionService.java
```diff
@@ -149,17 +150,30 @@ private HederaEvmTransactionProcessingResult buildSuccessResult(
...
+    private Optional<Bytes> getErrorMessage(final ContractFunctionResult result) {
+        return result.errorMessage().startsWith(HEX_PREFIX)
+                ? Optional.of(Bytes.fromHexString(result.errorMessage()))
+                : Optional.empty();
+    }
+
+    private Optional<ResponseCodeEnum> getDecodedErrorMessage(final ContractFunctionResult result) {
+        try {
+            return Optional.of(ResponseCodeEnum.valueOf(result.errorMessage()));
+        } catch (IllegalArgumentException iae) {
```
steven-sheehy, line 172: Using valueOf and catching IAEs might be a bit slow. Is there any way to detect if the error message is a function selector first and avoid the enum conversion?

steven-sheehy (changes requested):

kselveliev (dismissed):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/TransactionExecutionService.java
```diff
@@ -149,17 +150,30 @@ private HederaEvmTransactionProcessingResult buildSuccessResult(
...
+            // This is expected in some cases. No additional action required.
+            log.debug("Error message could not be decoded.", iae);
```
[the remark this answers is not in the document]
bilyana-gospodinova: Log message removed.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/TransactionExecutionService.java
bilyana-gospodinova: It seems that after the latest changes we don't need this method at all.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/TransactionExecutionService.java
```diff
@@ -149,17 +150,30 @@ private HederaEvmTransactionProcessingResult buildSuccessResult(
...
+
+    private Optional<Bytes> getErrorMessage(final ContractFunctionResult result) {
+        return result.errorMessage().startsWith(HEX_PREFIX)
+                ? Optional.of(Bytes.fromHexString(result.errorMessage()))
+                : Optional.empty();
+    }
+
+    private Optional<ResponseCodeEnum> getDecodedErrorMessage(final ContractFunctionResult result) {
+        try {
+            return Optional.of(ResponseCodeEnum.valueOf(result.errorMessage()));
```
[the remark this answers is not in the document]
bilyana-gospodinova: We had that logic in ContractCallService. This is why it was not added here. WIth the exception throwing approach, now we have that here as well, so this comment is covered.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/TransactionExecutionService.java
```diff
@@ -134,13 +135,13 @@ private ContractFunctionResult getTransactionResult(
...
 
-    private HederaEvmTransactionProcessingResult buildSuccessResult(
+    private MirrorEvmTransactionProcessingResult buildSuccessResult(
             final boolean isContractCreate,
             final TransactionRecord transactionRecord,
             final CallServiceParameters params) {
         var result = getTransactionResult(transactionRecord, isContractCreate);
 
-        return HederaEvmTransactionProcessingResult.successful(
+        return MirrorEvmTransactionProcessingResult.successful(
```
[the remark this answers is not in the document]
bilyana-gospodinova: Comment addressed. The error result returning is replaced by exception throwing.

steven-sheehy (dismissed):

kselveliev (approved):

steven-sheehy (approved):

PR #10064, opened 2025-01-07
Add github actions step for executing web3 tests using transaction executor

### LINKED ISSUE #9995: Add workflow step to execute web3 tests with transaction executor enabled

nickeynikolovv:
**Description**:
This PR adds new github actions step that executes the web3 tests using the new implementation for transaction executor. 

**Related issue(s)**:

Fixes #9995 

**Notes for reviewer**:
This step will be currently disabled until the full implementation for the reusable services is completed

**Checklist**

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -94,7 +94,7 @@ jobs:
         run: ./gradlew :${{matrix.project}}:build --scan ${{ secrets.GRADLE_ARGS }}
 
       - name: Execute Gradle using transaction executor
-        if: ${{ matrix.project == 'web3'}}
+        if: ${{ matrix.project == 'web3' && matrix.schema == 'v1'}}
         continue-on-error: true
         env:
           MIRROR_NODE_SCHEMA: ${{ matrix.schema}}
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -21,7 +21,7 @@ permissions:
 env:
   LC_ALL: C.UTF-8
   CGO_ENABLED: 1
-  ENABLE_TRANSACTION_EXECUTOR_STEP: "false" #Enable when the full implementation of reusable services is completed
+  ENABLE_TRANSACTION_EXECUTOR_STEP: "true" #Enable when the full implementation of reusable services is completed
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


.github/workflows/gradle.yml
```diff
@@ -93,6 +94,14 @@ jobs:
           SPRING_PROFILES_ACTIVE: ${{ matrix.schema}}
         run: ./gradlew :${{matrix.project}}:build --scan ${{ secrets.GRADLE_ARGS }}
 
+      - name: Execute Gradle using transaction executor
+        if: ${{ matrix.project == 'web3' && env.ENABLE_TRANSACTION_EXECUTOR_STEP == 'true'}}
+        env:
+          MIRROR_NODE_SCHEMA: ${{ matrix.schema}}
+          SPRING_PROFILES_ACTIVE: ${{ matrix.schema}}
+          HEDERA_MIRROR_WEB3_EVM_MODULARIZED_SERVICES: "true"
```
steven-sheehy, line 102: I believe the correct flag is `HEDERA_MIRROR_WEB3_EVM_MODULARIZEDSERVICES: "true"`. Spring Boot doesn't use underscore except to replace `.` in property paths.

.github/workflows/gradle.yml
```diff
@@ -21,6 +21,7 @@ permissions:
 env:
   LC_ALL: C.UTF-8
   CGO_ENABLED: 1
+  ENABLE_TRANSACTION_EXECUTOR_STEP: "false" #Enable when the full implementation of reusable services is completed
```
steven-sheehy, line 24: Shouldn't we just enable it and let it fail so we can see the progress being made? We can just not mark the check as required. Perhaps we only do it for v1 as well for now. We can set [continue-on-error](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error) to not fail the job.

.github/workflows/gradle.yml
nickeynikolovv: Actually both ways are working fine, but i renamed it to `HEDERA_MIRROR_WEB3_EVM_MODULARIZEDSERVICES ` to keep the correct way from the spring docs

.github/workflows/gradle.yml
nickeynikolovv: done

.github/workflows/gradle.yml
```diff
@@ -93,6 +93,15 @@ jobs:
           SPRING_PROFILES_ACTIVE: ${{ matrix.schema}}
         run: ./gradlew :${{matrix.project}}:build --scan ${{ secrets.GRADLE_ARGS }}
 
+      - name: Execute Gradle using transaction executor
+        if: ${{ matrix.project == 'web3'}}
```
steven-sheehy, line 97: As mentioned previously, we can just test v1 for now to save execution time.

```suggestion
        if: ${{ matrix.project == 'web3' && matrix.schema == 'v1'}}
```

nickeynikolovv: added

steven-sheehy (approved):

PR #10072, opened 2025-01-09
HIP-1056 Block File Reader

### LINKED ISSUE #9894: HIP-1056 Read block files
### Problem

Add the functionality to read blk files and populate mirror node domain classes.

### Solution

From the HIP-1056 design:
-Add the `BlockFileReader` and `ProtoBlockFileReader`
-Add a test that reads a blk file and verifies the correctness of the domain objects produced by the reader

### Alternatives

_No response_

xin-hedera:
**Description**:
This PR adds block file reader to read block stream files as defined by HIP-1056

- Add `ProtoBlockFileReader`
- Add `BlockRootHashDigest` to calculate block file's root hash

**Related issue(s)**:

Fixes #9894 

**Notes for reviewer**:

**Checklist**


docs/design/block-streams.md
```diff
@@ -37,7 +37,8 @@ package com.hedera.mirror.common.domain.transaction;
 public record BlockItem(Transaction transaction,
                         TransactionResult transactionResult,
                         List<TransactionOutput> transactionOutput, // Note: List may be empty
-                        Optional<StateChanges> stateChanges) implements StreamItem {}
+                        List<StateChanges> stateChanges // Note: List may be empty
```
xin-hedera, line 40: two reasons to make the change

- it's unclear from the doc if there should be strictly at most 1 statechanges for a transaction or not
- there are always non-transactional state changes before block proof, in case the last event in the last round has an event transaction, there will be more than 1 statechanges; it's unnecessary for the reader to sort out which statechanges belong to a transaction, and which belong to the event transaction.

hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/transaction/BlockFile.java
```diff
@@ -43,9 +44,6 @@ public class BlockFile implements StreamFile<BlockItem> {
     // Used to generate block hash
     private BlockProof blockProof;
 
-    // Contained within the last StateChange of the block, contains hashes needed to generate the block hash
```
xin-hedera, line 46: we have the following two hash leaves

- previous hash from `BlockHeader`
- start of block state hash from `BlockProof`

The other two are calculated from input block items merkle tree and output block item merkle tree, so we don't need `BlockStreamInfo`

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/reader/block/BlockRootHashDigest.java
```diff
@@ -0,0 +1,140 @@
...
+        validateHash(previousHash, "previousHash");
+        this.previousHash = previousHash;
+    }
+
+    public void setStartOfBlockStateHash(byte[] startOfBlockStateHash) {
+        validateHash(startOfBlockStateHash, "startOfBlockStateHash");
+        this.startOfBlockStateHash = startOfBlockStateHash;
+    }
+
+    @SneakyThrows
```
xin-hedera, line 95: left here by accident, will remove

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/reader/block/ProtoBlockFileReader.java
```diff
@@ -0,0 +1,253 @@
...
+
+    private void readEventTransactions(ReaderContext context) {
+        BlockItem protoBlockItem;
+        while (context.hasNextBlockItem() && (protoBlockItem = context.readBlockItemFor(EVENT_TRANSACTION)) != null) {
+            try {
+                var eventTransaction = protoBlockItem.getEventTransaction();
+                var transaction = eventTransaction.hasApplicationTransaction()
+                        ? Transaction.parseFrom(eventTransaction.getApplicationTransaction())
+                        : null;
+
```
edwin-greene, line 147: If `hasApplicationTransaction()` is false here, would it be ok for the method to return at this point?

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/reader/block/ProtoBlockFileReader.java
```diff
@@ -0,0 +1,253 @@
...
+     * transactional statechanges.
+     *
+     * @param context - The reader context
+     */
+    private void readTrailingStateChanges(ReaderContext context) {
+        while (context.hasNextBlockItem()) {
+            if (context.readBlockItemFor(STATE_CHANGES) == null) {
+                break;
+            }
+        }
```
edwin-greene, line 204: Are the trailing state changes mandatory for gathering the output hashes and calculating the block hash? If so, maybe this method should throw an exception similar to the others if no trailing state changes are found.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/reader/block/ProtoBlockFileReader.java
xin-hedera: we still need to read relevant block items, e.g., transaction result / output / statechanges for two reasons

- read all of them so we can advance to the next group of block items
- add them to block root hash digest

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/reader/block/ProtoBlockFileReader.java
xin-hedera: no, not needed for us

xin-hedera: We used to be very strict to fail on all incorrect data as much as we can detect in the stream files, however I believe we decided to move away from that approach. So that in the block file reader, unless the data is critical for mirrornode to properly parse and ingest the stream files, we should not fail / halt when missing / incorrect.

hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/transaction/BlockFile.java
```diff
@@ -60,6 +58,8 @@ public class BlockFile implements StreamFile<BlockItem> {
     @ToString.Exclude
     private String hash;
 
+    private Long index;
```
xin-hedera, line 61: add back the `index` field so we have a setter which makes testing easier. Not really benefit test cases in this PR, however it makes the block stream verifier PR tests easier so as to just set the index, instead of crafting a proper `BlockHeader` with the correct block number.

edwin-greene (approved):
Looks good to me!

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/reader/block/ProtoBlockFileReader.java
```diff
@@ -0,0 +1,253 @@
...
+                    .loadStart(streamFileData.getStreamFilename().getTimestamp())
+                    .name(filename)
+                    .size(bytes.length)
+                    .version(VERSION);
+
+            var blockItem = context.readBlockItemFor(RECORD_FILE);
+            if (blockItem != null) {
+                return blockFileBuilder
+                        .recordFileItem(blockItem.getRecordFile())
+                        .build();
```
steven-sheehy, line 75: What's the future flow here? Will we have to detect this upstream and send back through the reader chain?

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/reader/block/ProtoBlockFileReader.java
```diff
@@ -0,0 +1,253 @@
...
+            var blockFileBuilder = context.getBlockFile()
+                    .bytes(bytes)
+                    .loadStart(streamFileData.getStreamFilename().getTimestamp())
+                    .name(filename)
+                    .size(bytes.length)
+                    .version(VERSION);
+
+            var blockItem = context.readBlockItemFor(RECORD_FILE);
+            if (blockItem != null) {
+                return blockFileBuilder
```
steven-sheehy, line 73: For record file, we can optimize by not doing heavyweight `streamFileData.getBytes()` just to throw it away.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/reader/block/ProtoBlockFileReader.java
```diff
@@ -0,0 +1,253 @@
...
+            readTrailingStateChanges(context);
+            readBlockProof(context);
+
+            return blockFileBuilder
+                    .hash(context.getBlockRootHashDigest().digest())
+                    .build();
+        } catch (Exception e) {
+            if (e instanceof InvalidStreamFileException invalidStreamFileException) {
+                throw invalidStreamFileException;
+            }
```
steven-sheehy, line 89:
```suggestion
        } catch (InvalidStreamFileException e) {
            throw e;
        } catch (Exception e) {
```

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/reader/block/ProtoBlockFileReader.java
```diff
@@ -0,0 +1,253 @@
...
+            if (blockItem != null) {
+                return blockFileBuilder
+                        .recordFileItem(blockItem.getRecordFile())
+                        .build();
+            }
+
+            readBlockHeader(context);
+            readRounds(context);
+            readTrailingStateChanges(context);
+            readBlockProof(context);
```
steven-sheehy, line 81: Did we consider centralizing the block item reading in a loop with switch? Is there a risk that nodes might change the order of different types of items in the future but we have it hardcoded?

```java
for (var blockItem : blockItems) {
    switch (blockItem.getCase()) {
        case BLOCK_HEADER -> readBlockHeader(context, blockItem);
        ...
    }
}
```

steven-sheehy (changes requested):

hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/transaction/BlockFile.java
```diff
@@ -94,13 +96,55 @@ public String getFileHash() {
...
+            }
+
+            consensusEnd = consensusTimestamp;
+            return this;
+        }
+
+        void prebuild() {
+            if (count == null) {
+                count = items$value != null ? (long) items$value.size() : 0;
+            }
```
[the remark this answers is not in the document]
xin-hedera: this is more inherent. can't say the other way is uglier however at least it's more explicit, since we'll have to add a count state variable into the context and increment it every time an application transaction is read.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/reader/block/ProtoBlockFileReader.java
```diff
@@ -0,0 +1,253 @@
...
+
+            var blockItem = blockItems.get(index);
+            if (blockItem.getItemCase() != itemCase) {
+                return null;
+            }
+
+            index++;
+            switch (itemCase) {
+                case EVENT_HEADER, EVENT_TRANSACTION -> blockRootHashDigest.addInputBlockItem(blockItem);
+                case STATE_CHANGES, TRANSACTION_OUTPUT, TRANSACTION_RESULT -> blockRootHashDigest.addOutputBlockItem(
```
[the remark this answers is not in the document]
xin-hedera: don't think HIP said that. but that's a valid point since round header is also an output block item. If HIP is amended, we'd ask services to make the change, and we'll follow suit.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/reader/block/ProtoBlockFileReader.java
xin-hedera: plan is block stream poller would delegate it to record file reader and signature file reader to read the files in record file item.

originally wanted to do that here, however there is a class dependency issue that we can't reference `StreamFileSignature` (a class in `importer` module, returned by `SignatureFileReader.read`) in `BlockFile` in common module.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/reader/block/ProtoBlockFileReader.java
xin-hedera: I think the for loop above is overly-simplified so doesn't really show its inherent complexity.

by design, most block items have rules about the location they should appear and how they are grouped together as a unit. The most obvious is event transaction has to be followed by a transaction result, and then optionally by transaction output and state changes. Then at least we'd need to

- loop with control of how many items to advance
- have a context to validate things like if the current block item matches expectation, e.g., if the previous item is event transaction, the current block item must be transaction result

steven-sheehy, line 81: Where do we handle filtered items?

steven-sheehy: My main concern is around resiliency against change in consensus node order. Like if suddenly they change states changes to appear before outputs. Though I guess in most cases that would mean the leaves of the hash would change as well.

xin-hedera: yeah. I'd bet once it goes production, we should not make such changes as they are breaking.

xin-hedera: nope, it's not handled yet. I should have created a ticket as follow-up. Reason is I don't see a complete picture of how it actually works step by step, given there's already many things I need to dig out from services code w.r.t. block root hash calcualtion.

xin-hedera: #10132

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/reader/block/BlockRootHashDigest.java
```diff
@@ -0,0 +1,138 @@
...
+    private List<byte[]> outputHashes = new ArrayList<>();
+
+    @NonFinal
+    private byte[] previousHash;
+
+    @NonFinal
+    private byte[] startOfBlockStateHash;
+
+    public void addInputBlockItem(BlockItem blockItem) {
+        inputHashes.add(createMessageDigest().digest(blockItem.toByteArray()));
```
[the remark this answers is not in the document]
xin-hedera: turns out can just use the same digest object over and over, since by spec, `MessageDigest.digest()` is required to reset the internal state.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/reader/block/BlockRootHashDigest.java
```diff
@@ -0,0 +1,129 @@
...
+        validateHash(previousHash, "previousHash");
+        this.previousHash = previousHash;
+    }
+
+    public void setStartOfBlockStateHash(byte[] startOfBlockStateHash) {
+        validateHash(startOfBlockStateHash, "startOfBlockStateHash");
+        this.startOfBlockStateHash = startOfBlockStateHash;
+    }
+
+    private byte[] getRootHash(List<byte[]> leaves) {
```
xin-hedera, line 86: changed it from a static method to instance method to reuse the instance variable `digest`

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/reader/block/BlockRootHashDigest.java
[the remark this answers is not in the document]
xin-hedera: changed to the shared instance variable `digest` instead

steven-sheehy (approved):
LGTM

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/reader/block/ProtoBlockFileReader.java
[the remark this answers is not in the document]
steven-sheehy: I'm referring to this comment:

`All items in the block other than the final BlockProof, a RecordFileItem, or a FilteredItemHash are either Input or Output items.`

I'll check with Jasper.

PR #10116, opened 2025-01-13
Fix failing tests in ContractCallServicePrecompileModificationTest

### LINKED ISSUE #9984: Fix `ContractCallServicePrecompileModificationTest`
### Problem

As part of the integration of modularized services there are some failing tests in `ContractCallServicePrecompileModificationTest`. We need to investigate why the tests fail and fix them.

### Solution

Fix the failing tests - there might be some setup issue or something might be missed in the implementation.

### Alternatives

_No response_

sdimitrov9:
**Description**:
Fix multiple tests within ContractCallServicePrecompileModificationTest class

**Related issue(s)**:
Fixes #9984

**Checklist**


hedera-mirror-common/src/test/java/com/hedera/mirror/common/domain/DomainBuilder.java
```diff
@@ -920,6 +920,22 @@ public DomainWrapper<SidecarFile, SidecarFile.SidecarFileBuilder> sidecarFile()
         return new DomainWrapperImpl<>(builder, builder::build);
     }
 
+    public DomainWrapper<TokenAccount, TokenAccount.TokenAccountBuilder<?, ?>> tokenAccount(Long balance) {
+        long timestamp = timestamp();
+        var builder = TokenAccount.builder()
+                .accountId(id())
+                .automaticAssociation(false)
+                .associated(true)
+                .balance(balance)
```
bilyana-gospodinova, line 929: Is this new method really needed? Can't we write directly:
```
protected void tokenAccountPersist(final Entity token, final Long accountId, final Long balance) {
        domainBuilder
                .tokenAccount(balance)
                .customize(ta -> ta.tokenId(token.getId())
                        .accountId(accountId)
                        .freezeStatus(TokenFreezeStatusEnum.UNFROZEN)
                        .kycStatus(TokenKycStatusEnum.GRANTED)
                        .balance(balance)
                        .associated(true))
                .persist();
    }
```

sdimitrov9: Since it is used only at one place as of now we can go with your proposal as well. Done.

bilyana-gospodinova: My proposal is a little different - use the `balance` method inside `customize` and remove the new method in the DomainBuilder:
```
protected void tokenAccountPersist(final Entity token, final Long accountId, final Long balance) {
        domainBuilder
                .tokenAccount() // remove the parameter here and use the default method
                .customize(ta -> ta.tokenId(token.getId())
                        .accountId(accountId)
                        .freezeStatus(TokenFreezeStatusEnum.UNFROZEN)
                        .kycStatus(TokenKycStatusEnum.GRANTED)
                        .balance(balance) // use the balance method in customize
                        .associated(true))
                .persist();
    }
```

hedera-mirror-web3/src/test/resources/config/application.yml
```diff
@@ -5,13 +5,16 @@ hedera:
...
+          contracts.maxGasPerSec: "15000000"
+          contracts.maxRefundPercentOfGasLimit: "100"
         evm_versions:
           0: 0.30.0
           50: 0.34.0
           100: 0.38.0
           150: 0.46.0
           200: 0.50.0
         network: OTHER
+        modularizedServices: false
```
bilyana-gospodinova, line 17: If it is set to false, it is the same as if the property is not added at all here.

sdimitrov9: I would prefer to have the property there and edit the value when needed. Wouldn't it be easier this way rather than writing the whole property every time when having to set the value to true?

hedera-mirror-common/src/test/java/com/hedera/mirror/common/domain/DomainBuilder.java
sdimitrov9: Fixed.

hedera-mirror-web3/src/test/resources/config/application.yml
steven-sheehy: You mean for manually testing modularized services? Why can you just set `HEDERA_MIRROR_WEB3_EVM_MODULARIZEDSERVICES=true` in the env or modify `MirrorNodeEvmProperties` class instead?

sdimitrov9: I actually can, thanks! The row is now removed.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/TransactionExecutionService.java
```diff
@@ -165,7 +164,7 @@ private HederaEvmTransactionProcessingResult buildFailedResult(
     private TransactionBody.Builder defaultTransactionBodyBuilder(final CallServiceParameters params) {
         return TransactionBody.newBuilder()
                 .transactionID(TransactionID.newBuilder()
-                        .transactionValidStart(TRANSACTION_START)
+                        .transactionValidStart(new Timestamp(Instant.now().getEpochSecond(), 0))
```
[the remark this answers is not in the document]
sdimitrov9: After the changes from the other branch were merged I pulled from main and now we should be good to go.

bilyana-gospodinova (approved):
LGTM

steven-sheehy (dismissed):
LGTM but looks like merge conflicts after merging some others

steven-sheehy (approved):

PR #10117, opened 2025-01-13
Fix consensus time and transaction valid start time

kselveliev:
**Description**:
A few tests like update token expiry are failing since currently in transaction executor we are passing the EPOCH instant. In a test like update token expiry info consensus time is 0 due to that, consequently the tests fails.
In this pr we are now passing the now instant as transaction valid start and  consensus time. 
This will resolve issue like the ones in update token expiry test.

This PR modifies 
TransactionExecutionService - now instead of EPOCH time we are passing Instant.now()

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


bilyana-gospodinova (approved):
LGTM

steven-sheehy (approved):

sdimitrov9 (approved):

PR #10131, opened 2025-01-14
Revert "Add feature flag for the admin_key in the get network nodes response"

steven-sheehy:
Reverts hashgraph/hedera-mirror-node#9875 since it's no longer necessary to feature flag it since consensus nodes have support in all environments.


xin-hedera (approved):
LGTM

PR #10134, opened 2025-01-15
Fix NodeUpdateTransaction always updating node zero

### LINKED ISSUE #10133: NodeUpdateTransaction always updates same node
### Description

The NodeUpdateTransaction always updates the same node due to an assumption that the node_id in the receipt would be populated.

### Steps to reproduce

NodeUpdateTransaction with different node IDs

### Additional context

_No response_

### Hedera network

testnet

### Version

v0.121.0

### Operating system

None

steven-sheehy:
**Description**:

* Change node update to populate `createdTimestamp` if it's missing and an inner transaction. This is because nodes erroneously sent us updates instead of creates when migrating state.
* Fix `NodeUpdateTransaction` updating the wrong node
* Fix `NodeDeleteTransaction` deleting the wrong node

**Related issue(s)**:

Fixes #10133

**Notes for reviewer**:

The protobuf spec silently changed in https://github.com/hashgraph/hedera-services/pull/16622/files after we implemented the logic.

**Checklist**


xin-hedera (approved):
LGTM

Issue #10079, opened 2025-01-10
Fix `ContractCallEvmCodesTest` test suite


nickeynikolovv:
Fix failing tests for `ContractCallEvmCodesTest` using the modularized EVM

PR #10148, opened 2025-01-16
Support block hash operation in modularized web3

### LINKED ISSUE #9983: Historical block hash support
### Problem

As part of the implementation of modularized services we came into the issue that only the last 256 block hashes can be read (this comes from the hedera.app dependency). The mirror node needs to return the block hash of any block.

### Solution

We need to adjust the state to return any block hash. Might need to replace the singleton logic with KV.

### Alternatives

_No response_

### LINKED ISSUE #10149: Fix gas estimate when `modularizedServices` flag is enabled
### Problem

In `AbstractContractCallServiceTest#gasUsedAfterExecution` we make an eth_call and get the actual gas used for the transaction. This method is used in the estimate gas scenarios were the estimated value is compared to the value returned from the method. The issue now is that the `modularizedServices` flag is not used in the method, so by default the actual gas used call is always made against the mono code.

### Solution

The `AbstractContractCallServiceTest#gasUsedAfterExecution` method needs to be adapted to use the `modularizedServices` flag and compare estimate gas values against the modularized code to actual gas used values against the modularized code as well.

### Alternatives

_No response_

steven-sheehy:
**Description**:

* Add a `BlockInfo` singleton to return the last 256 block hashes
* Add a `RunningHashes` singleton to return the running hash
* Add `MirrorOperationTracer` to modularized execution
* Change modularized code to not initialize stacked state frame
* Change to always load requested block in modularized code for use in `BlockInfo` later
* Change to initialize `VersionedConfiguration` once and use it everywhere
* Change to reuse stateless `TransactionExecutor` between requests
* Disable `contracts.sidecars` to avoid generating unnecessary sidecar data
* Fix EVM version being cached for the first request
* Fix some tests by always inserting exchange rate

**Related issue(s)**:

Fixes #9983
Fixes #10149
Part of #10079

**Notes for reviewer**:

Before
```
3342 tests completed, 397 failed, 6 skipped
```

After
```
3350 tests completed, 302 failed, 6 skipped
```

**Checklist**

### DIFF
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/common/ContractCallContext.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/common/ContractCallContext.java
@@ -152,8 +152,10 @@ public class ContractCallContext {
     }
 
     public boolean useHistorical() {
-        return (callServiceParameters != null && callServiceParameters.getBlock() != BlockType.LATEST)
-                || recordFile != null; // Remove recordFile comparison after mono code deletion
+        if (callServiceParameters != null) {
+            return callServiceParameters.getBlock() != BlockType.LATEST;
+        }
+        return recordFile != null; // Remove recordFile comparison after mono code deletion
     }
 
     public void incrementContractActionsCounter() {
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/common/ContractCallContext.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/common/ContractCallContext.java
@@ -153,7 +153,7 @@ public class ContractCallContext {
 
     public boolean useHistorical() {
         return (callServiceParameters != null && callServiceParameters.getBlock() != BlockType.LATEST)
-                || recordFile != null; // TODO: Remove recordFile comparison after mono code deletion
+                || recordFile != null; // Remove recordFile comparison after mono code deletion
     }
 
     public void incrementContractActionsCounter() {
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/common/ContractCallContext.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/common/ContractCallContext.java
@@ -26,7 +26,9 @@ import com.hedera.mirror.web3.evm.contracts.execution.traceability.OpcodeTracer;
 import com.hedera.mirror.web3.evm.contracts.execution.traceability.OpcodeTracerOptions;
 import com.hedera.mirror.web3.evm.store.CachingStateFrame;
 import com.hedera.mirror.web3.evm.store.StackedStateFrames;
+import com.hedera.mirror.web3.service.model.CallServiceParameters;
 import com.hedera.mirror.web3.state.FileReadableKVState;
+import com.hedera.mirror.web3.viewmodel.BlockType;
 import java.util.ArrayList;
 import java.util.EmptyStackException;
 import java.util.List;
@@ -57,6 +59,10 @@ public class ContractCallContext {
 
     @Setter
     private List<Opcode> opcodes = new ArrayList<>();
+
+    @Setter
+    private CallServiceParameters callServiceParameters;
+
     /**
      * Record file which stores the block timestamp and other historical block details used for filtering of historical
      * data.
@@ -101,7 +107,6 @@ public class ContractCallContext {
     }
 
     public void reset() {
-        recordFile = null;
         stack = stackBase;
         file = Optional.empty();
     }
@@ -147,7 +152,8 @@ public class ContractCallContext {
     }
 
     public boolean useHistorical() {
-        return recordFile != null;
+        return (callServiceParameters != null && callServiceParameters.getBlock() != BlockType.LATEST)
+                || recordFile != null; // TODO: Remove recordFile comparison after mono code deletion
     }
 
     public void incrementContractActionsCounter() {
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
@@ -198,7 +198,7 @@ public class MirrorNodeEvmProperties implements EvmProperties {
     private int feesTokenTransferUsageMultiplier = 380;
 
     @Getter
-    private boolean modularizedServices = true;
+    private boolean modularizedServices;
 
     public boolean shouldAutoRenewAccounts() {
         return autoRenewTargetTypes.contains(EntityType.ACCOUNT);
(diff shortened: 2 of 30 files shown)


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
```diff
@@ -176,17 +182,23 @@ public class MirrorNodeEvmProperties implements EvmProperties {
...
+    private final VersionedConfiguration versionedConfiguration =
+            new ConfigProviderImpl(false, null, getTransactionProperties()).getConfiguration();
+
     @Getter
     @Min(1)
     private int feesTokenTransferUsageMultiplier = 380;
 
     @Getter
-    private boolean modularizedServices;
+    private boolean modularizedServices = true;
```
bilyana-gospodinova, line 201: Isn't it too early to set the flag to true by default?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/singleton/BlockInfoSingleton.java
```diff
@@ -0,0 +1,93 @@
...
+     * supports custom implementations for operations that don't need to use the BlockInfo singleton.
+     */
+    @SneakyThrows
+    private byte[] getBlockHashes(RecordFile recordFile) {
+        if (recordFile.getIndex() == 0) {
+            return Hex.decodeHex(recordFile.getHash());
+        }
+
+        var config = properties.getVersionedConfiguration().getConfigData(BlockRecordStreamConfig.class);
+        int blockHashCount = config.numOfBlockHashesInState();
```
bilyana-gospodinova, line 78: What if we configure the block hashes in the state to be 1? I can't think of a case when we will need more than 1 block hash per transaction. This way we will avoid the calculations below.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/common/ContractCallContext.java
```diff
@@ -147,7 +152,8 @@ public void initializeStackFrames(final StackedStateFrames stackedStateFrames) {
     }
 
     public boolean useHistorical() {
-        return recordFile != null;
+        return (callServiceParameters != null && callServiceParameters.getBlock() != BlockType.LATEST)
```
bilyana-gospodinova, line 155: `callServiceParameters` can be a boolean in the context - we don't use any other property from the object.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/singleton/BlockInfoSingleton.java
IvanKavaldzhiev: A given user can call a smart contract method that makes multiple `BLOCKHASH` opcode invocations, with `assembly` block for example. So I think there will be cases when multiple hashes are needed.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
```diff
@@ -99,14 +100,19 @@ protected ContractCallService(
      */
     protected HederaEvmTransactionProcessingResult callContract(CallServiceParameters params, ContractCallContext ctx)
             throws MirrorEvmTransactionException {
-        // if we have historical call, then set the corresponding record file in the context
-        if (params.getBlock() != BlockType.LATEST) {
+        ctx.setCallServiceParameters(params);
+
+        if (mirrorNodeEvmProperties.isModularizedServices() || params.getBlock() != BlockType.LATEST) {
             ctx.setRecordFile(recordFileService
```
kselveliev, line 106: ’ve debugged this but it’s currently causing some regressions in the tests and most modularized test that are latest block
 are running against `EVM_VERSION`.  0.30 (instead of 0.50)
. 
With this change we are always setting the record file in the `contractCallContext` for modularized execution.
And we are getting in the following case -> `recordFileRepository.findLatest();`

recordFileService
        .findByBlockType(params.getBlock()) -> params.block() is latest

which internally does ->` recordFileRepository.findLatest();`
In test db we only have the genesis block setup so  latest block has index 0.

The problem with that is that later on in `TransactionExecutorFactory`
we do -> `var version = mirrorNodeEvmProperties.getSemanticEvmVersion();`

which internally does -> 
```
if (context.useHistorical()) { // we use historical if we have set record file here
    return getEvmVersionForBlock(context.getRecordFile().getIndex());
}
```
but the record file we have is with 0 index.

And test yaml mapping for evm versions is:
0: 0.30.0
50: 0.34.0
100: 0.38.0
150: 0.46.0
200: 0.50.0

We somehow need to figure out how not to get in historical context during latest call tests.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/singleton/BlockInfoSingleton.java
steven-sheehy: This method is just temporary as the comment explains. With https://github.com/hashgraph/hedera-services/pull/17354 we can write a custom operator that will directly lookup the block num arg on an as needed basis

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/common/ContractCallContext.java
steven-sheehy: Prefer to keep it as CallServiceParameters for future needs where we might need other fields from the params. It doesn't reduce memory by changing to a boolean since CallServiceParameters is already on the stack.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
steven-sheehy: Thanks for pointing this out. Hard to see regressions when so many tests are failing. I fixed this through a number of changes:

* Gas estimation failing for same problem as https://github.com/hashgraph/hedera-mirror-node/pull/10150. Fixed differently than that PR be exposing `HederaEvmTransactionProcessingResult` in `ContractCallService`.
* Changed logic in `ContractCallContext.isHistorical()` to only run if `callServiceParameters` is populated.
* Change to use epoch for genesis block so bug https://github.com/hashgraph/hedera-services/issues/17423 in config override is not encountered

kselveliev:
Looks great to me. From running the tests from main an here the only diff I see in the test result is ContractCallNativePrecompileTest here is getting THROTTLED_AT_CONSENSUS since on main are passing.
Currently trying to figure out what causes it.

steven-sheehy:
THROTTLED_AT_CONSENSUS is usually due to lack of throttle definition. Let's get this then I can add the system file defaults as a follow up.

kselveliev (approved):

PR #10150, opened 2025-01-16
Adapt gas estimation with modularized services code

bilyana-gospodinova:
**Description**:
The `AbstractContractCallServiceTest#gasUsedAfterExecution` method is adapted to use the `modularizedServices` flag and compare estimate gas values against the modularized code to actual gas used values against the modularized code as well.

Fixes https://github.com/hashgraph/hedera-mirror-node/issues/10149

### DIFF
--- a/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
+++ b/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
@@ -23,6 +23,9 @@ import static com.hedera.mirror.web3.validation.HexValidator.HEX_PREFIX;
 import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
 import static org.assertj.core.api.AssertionsForClassTypes.assertThatThrownBy;
 import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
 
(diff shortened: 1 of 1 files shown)


hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
```diff
@@ -139,12 +151,21 @@ void cleanup() {
...
+                        .execute(serviceParameters, serviceParameters.getGas())
+                        .getGasUsed();
+
+                assertThat(store.getStackedStateFrames().height()).isEqualTo(1);
+            } else {
+                when(gasUsedCounter.withTags(anyString(), anyString(), anyString(), anyString()))
+                        .thenReturn(mock(Counter.class));
+                result = transactionExecutionService
+                        .execute(serviceParameters, serviceParameters.getGas(), gasUsedCounter)
+                        .getGasUsed();
```
steven-sheehy, line 167: I don't think this is the right approach to take as it duplicates logic and it uses mocking in an integration test. I'm changing this logic in my PR so I'd then have to remember to change it here again depending on who merges first.

Instead, we should expose the processing result from the existing `ContractCallService`. You'll have to move validateResult into callContract as well:

```java
    @VisibleForTesting
    public HederaEvmTransactionProcessingResult callContract(CallServiceParameters params)
        return ContractCallContext.run(context -> callContract(params, context));
    }
```

I have to make a similar change to my PR since my other changes also caused this issue to surface and regress a bunch of tests. So we might not need this PR if so.

bilyana-gospodinova:
Closing this PR as the issue is fixed in https://github.com/hashgraph/hedera-mirror-node/pull/10148.

PR #10152, opened 2025-01-16
Fix hbar and token transfer tests

kselveliev:
**Description**:
This Pr fixes `cryptoTransferHbars`, `cryptoTransferToken` and `cryptoTransferHbarsAndToken` test in ContractCallServicePrecompileModificationTest.
Those test were failing in the opcode assertion calls.
In the opcode flow we are always having a timestamp in the request.

So even if the timestamp is pretty similar to the latest case token balance and account balance are calculated with the historical queries. This PR adds the token and account balance setup for opcode flow.

Having the blueprint of the token and account balance, now we can fix other opcode test failure which are failing with insufficient payer or token balance.

`ContractCallServicePrecompileModificationTest` - add token and account balance setup for transfer tests

**Related issue(s)**:

Fixes #

**Notes for reviewer**:
3342 tests completed, 295 failed, 6 skipped

**Checklist**

### DIFF
--- a/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServicePrecompileModificationTest.java
+++ b/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServicePrecompileModificationTest.java
@@ -1193,7 +1193,6 @@ class ContractCallServicePrecompileModificationTest extends AbstractContractCall
         long timestampForBalances = payer.getCreatedTimestamp();
         persistAccountBalance(payer, payer.getBalance());
         persistAccountBalance(sender, sender.getBalance(), timestampForBalances);
-        treasuryEntity = domainBuilder.entity().customize(e -> e.id(2L).num(2L)).get();
         persistAccountBalance(treasuryEntity, treasuryEntity.getBalance(), timestampForBalances);
 
         // When
@@ -1228,7 +1227,6 @@ class ContractCallServicePrecompileModificationTest extends AbstractContractCall
         persistTokenBalance(sender, tokenEntity, timestampForBalances);
 
         persistTokenBalance(receiver, tokenEntity, timestampForBalances);
-        treasuryEntity = domainBuilder.entity().customize(e -> e.id(2L).num(2L)).get();
         persistAccountBalance(treasuryEntity, treasuryEntity.getBalance(), timestampForBalances);
 
         tokenAccountPersist(tokenEntity, sender);
@@ -1271,7 +1269,6 @@ class ContractCallServicePrecompileModificationTest extends AbstractContractCall
 
         persistTokenBalance(receiver, tokenEntity, timestampForBalances);
 
-        treasuryEntity = domainBuilder.entity().customize(e -> e.id(2L).num(2L)).get();
         persistAccountBalance(treasuryEntity, treasuryEntity.getBalance(), timestampForBalances);
 
         tokenAccountPersist(tokenEntity, sender);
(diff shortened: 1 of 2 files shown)


hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServicePrecompileModificationTest.java
```diff
@@ -1190,6 +1192,12 @@ void cryptoTransferHbars() throws Exception {
         final var receiver = accountEntityWithEvmAddressPersist();
         final var payer = accountEntityWithEvmAddressPersist();
 
+        long timestampForBalances = payer.getCreatedTimestamp();
+        persistAccountBalance(payer, payer.getBalance());
+        persistAccountBalance(sender, sender.getBalance(), timestampForBalances);
+        treasuryEntity = domainBuilder.entity().customize(e -> e.id(2L).num(2L)).get();
```
bilyana-gospodinova, line 1198: We probably don't need to do this. In the next line we can use the sender's/payer's timestamp and it should work. We did this in other tests.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServicePrecompileModificationTest.java
```diff
@@ -1542,4 +1572,28 @@ private HederaToken convertTokenEntityToHederaToken(final Token token, final Ent
                         EntityIdUtils.asHexedEvmAddress(new Id(0, 0, entityRenewAccountId)),
                         BigInteger.valueOf(entity.getEffectiveExpiration())));
     }
+
+    private void persistAccountBalance(Entity account, long balance, long timestamp) {
```
bilyana-gospodinova, line 1576: This and the next 2 methods can be put in a parent class since they will be used in other classes too.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServicePrecompileModificationTest.java
kselveliev: Hmm how so? looking at the historical account balance query we cannot skip adding an account_balance row for account 2 and should be >= of the account_balance time of the other test accounts. Otherwise we return 0 balance for the account and the test starts failing.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServicePrecompileModificationTest.java
kselveliev: Comment addressed

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServicePrecompileModificationTest.java
bilyana-gospodinova: Correct. However, on the mentioned line you don't create an account balance entry for account 2, but generate a new entry for account 2 as an entity, which is different.

kselveliev: Right, comment addressed

bilyana-gospodinova (approved):
LGTM

steven-sheehy (approved):

steven-sheehy (approved):

PR #10170, opened 2025-01-18
Remove automatic publishing of builds data

Neurone:
**Description:**

To be more privacy-friendly by default, this PR removes the automatic publishing of build data to Gradle servers. The report is still published automatically when the build is run in a CI environment.

If developers want to publish build scans, they can use the `--scan` option:

```sh
./gradlew clean build --scan
````

or set the `publishing.onlyIf` parameter to `true` to always publish the report like this:

```
develocity {
    buildScan {
        publishing.onlyIf { true }
        termsOfUseUrl = "https://gradle.com/terms-of-service"
        termsOfUseAgree = "yes
        tag("CI")
    }
}
```

In this case, developers can use `--no-scan` to avoid selectively publishing the report:

```sh
./gradlew clean build --no-scan
````

**Related issue(s)**:

N/A

**Notes for reviewer**:

N/A

**Checklist**

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -56,6 +56,7 @@ fun shortenProjectName(project: ProjectDescriptor) {
 
 develocity {
     buildScan {
+        publishing.onlyIf { false }
         termsOfUseUrl = "https://gradle.com/terms-of-service"
         termsOfUseAgree = "yes"
         tag("CI")
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -56,7 +56,7 @@ fun shortenProjectName(project: ProjectDescriptor) {
 
 develocity {
     buildScan {
-        publishing.onlyIf { false }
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: settings.gradle.kts [hunk 1]
```diff
@@ -56,6 +56,7 @@ fun shortenProjectName(project: ProjectDescriptor) {
 
 develocity {
     buildScan {
+        publishing.onlyIf { false }
```
steven-sheehy, line 59: How about a middle ground where it's only published from CI? The logs are lacking in gradle so it has helped us at various times to have the build scan in CI.

```suggestion
        publishing.onlyIf { System.getenv().containsKey("CI") }
```

Neurone: It sounds good!

steven-sheehy (approved):
LGTM. Thanks!

Issue #9975, opened 2024-12-17
RestJava RestSpecTest will likely fail when the selected specs are changed


xin-hedera:
### Description

RestJava `RestSpecTest` runs a selected subset of spec tests against `hedera-mirror-rest:latest` image running in a container. Problem arises when any PR changes the spec due to either enhancement or bug fix, `RestSpecTest` would then fail since the response from the latest production image will cause mismatch.

### Steps to reproduce

Check CI workflow run log [here](https://github.com/hashgraph/hedera-mirror-node/actions/runs/12377057368/job/34546066347?pr=9973)

### Hedera network

other

### Version

v0.121.0-SNAPSHOT

### Operating system

None

steven-sheehy:
Closing as no longer applicable.

PR #10215, opened 2025-01-23
HIP-1056: Update topic message running hash version to be nullable

### LINKED ISSUE #9933: HIP-1056 Update TopicMessage runningHashVersion for blockstreams
### Problem

The runningHashVersion will not be included in the blockstream. 

### Solution

Update the topic_message table to make the runningHashVersion nullable.

In the Importer:

Change `TopicMessage.runningHashVersion` from int to Integer to allow for null values.
Update `ConsensusSubmitMessageTransactionHandler` to account for block streams no longer sending the runningHashVersion value:
If the TransactionReceipt runningHashVersion is `3` (the current value) set the runningHashVersion to null.
If the TransactionReceipt runningHashVersion is not `3` set the runningHashVersion to that value.

### Alternatives

_No response_

edwin-greene:
**Description**:

- Adds a migration to make topic_message.running_hash_version nullable.
- REST and GRPC APIs, when fetching a null value, will return the current default running hash version value. 

**Related issue(s)**:

Fixes #9933 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/topic/TopicMessage.java
+++ b/hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/topic/TopicMessage.java
@@ -61,7 +61,7 @@ public class TopicMessage implements Comparable<TopicMessage>, Persistable<Long>
     @ToString.Exclude
     private byte[] runningHash;
 
-    private int runningHashVersion;
+    private Integer runningHashVersion;
 
     private long sequenceNumber;
 
--- a/hedera-mirror-grpc/src/main/java/com/hedera/mirror/grpc/controller/ConsensusController.java
+++ b/hedera-mirror-grpc/src/main/java/com/hedera/mirror/grpc/controller/ConsensusController.java
@@ -29,6 +29,7 @@ import com.hedera.mirror.grpc.util.ProtoUtil;
 import com.hederahashgraph.api.proto.java.ConsensusMessageChunkInfo;
 import com.hederahashgraph.api.proto.java.Timestamp;
 import com.hederahashgraph.api.proto.java.TransactionID;
+import java.util.Objects;
 import lombok.CustomLog;
 import lombok.RequiredArgsConstructor;
 import net.devh.boot.grpc.server.service.GrpcService;
@@ -89,14 +90,12 @@ public class ConsensusController extends ReactorConsensusServiceGrpc.ConsensusSe
 
     // Consider caching this conversion for multiple subscribers to the same topic if the need arises.
     private ConsensusTopicResponse toResponse(TopicMessage t) {
-        int runningHashVersion =
-                t.getRunningHashVersion() == null ? DEFAULT_RUNNING_HASH_VERSION : t.getRunningHashVersion();
-
         var consensusTopicResponseBuilder = ConsensusTopicResponse.newBuilder()
                 .setConsensusTimestamp(ProtoUtil.toTimestamp(t.getConsensusTimestamp()))
                 .setMessage(ProtoUtil.toByteString(t.getMessage()))
                 .setRunningHash(ProtoUtil.toByteString(t.getRunningHash()))
-                .setRunningHashVersion(runningHashVersion)
+                .setRunningHashVersion(
+                        Objects.requireNonNullElse(t.getRunningHashVersion(), DEFAULT_RUNNING_HASH_VERSION))
                 .setSequenceNumber(t.getSequenceNumber());
 
         if (t.getChunkNum() != null) {
--- a/hedera-mirror-grpc/src/main/java/com/hedera/mirror/grpc/controller/ConsensusController.java
+++ b/hedera-mirror-grpc/src/main/java/com/hedera/mirror/grpc/controller/ConsensusController.java
@@ -46,6 +46,9 @@ import reactor.core.publisher.Mono;
 @RequiredArgsConstructor
 public class ConsensusController extends ReactorConsensusServiceGrpc.ConsensusServiceImplBase {
 
+    // Blockstreams no longer contain runningHashVersion, default to the latest version
+    static final int DEFAULT_RUNNING_HASH_VERSION = 3;
+
     private final TopicMessageService topicMessageService;
 
     @Override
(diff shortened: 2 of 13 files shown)


edwin-greene:
Note: Rest-Java test failing due to outdated Docker Image for REST.

hedera-mirror-grpc/src/main/java/com/hedera/mirror/grpc/controller/ConsensusController.java
```diff
@@ -86,11 +89,14 @@ private long convertTimestamp(Timestamp timestamp) {
...
     private ConsensusTopicResponse toResponse(TopicMessage t) {
+        int runningHashVersion =
+                t.getRunningHashVersion() == null ? DEFAULT_RUNNING_HASH_VERSION : t.getRunningHashVersion();
+
         var consensusTopicResponseBuilder = ConsensusTopicResponse.newBuilder()
                 .setConsensusTimestamp(ProtoUtil.toTimestamp(t.getConsensusTimestamp()))
                 .setMessage(ProtoUtil.toByteString(t.getMessage()))
                 .setRunningHash(ProtoUtil.toByteString(t.getRunningHash()))
-                .setRunningHashVersion(t.getRunningHashVersion())
+                .setRunningHashVersion(runningHashVersion)
```
steven-sheehy, line 99:
```suggestion
                .setRunningHashVersion(Objects.requireNonNullElse(t.getRunningHashVersion(), DEFAULT_RUNNING_HASH_VERSION))
```

hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/topic/TopicMessage.java
```diff
@@ -61,7 +61,7 @@ public class TopicMessage implements Comparable<TopicMessage>, Persistable<Long>
     @ToString.Exclude
     private byte[] runningHash;
 
-    private int runningHashVersion;
+    private Integer runningHashVersion;
```
xin-hedera, line 64: why cant' we keep the column not null and in block item transformer set receipt.topicRunningHashVersion to 3? Then we can scrape this PR...

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/transactionhandler/ConsensusSubmitMessageTransactionHandler.java
```diff
@@ -69,10 +72,15 @@ protected void doUpdateTransaction(Transaction transaction, RecordItem recordIte
...
-        int runningHashVersion =
-                receipt.getTopicRunningHashVersion() == 0 ? 1 : (int) receipt.getTopicRunningHashVersion();
         var topicMessage = new TopicMessage();
 
+        // Only persist the value if it is not the default
+        if (receipt.getTopicRunningHashVersion() != DEFAULT_RUNNING_HASH_VERSION) {
+            var runningHashVersion =
+                    receipt.getTopicRunningHashVersion() == 0 ? 1 : (int) receipt.getTopicRunningHashVersion();
+            topicMessage.setRunningHashVersion(runningHashVersion);
+        }
```
xin-hedera, line 82: this is wrong without setting a sentinel value in block item transformer. `receipt.topicRunningHashVersion` is a primitive, means it defaults to 0.

hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/topic/TopicMessage.java
steven-sheehy: Believe we said it's worth the reduction in storage for such a large table

steven-sheehy (changes requested):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/transactionhandler/ConsensusSubmitMessageTransactionHandler.java
xin-hedera: We should update the design doc to highlight that the consensus submit message block item transformer should set receipt.runningHashVersion to 3.

hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/topic/TopicMessage.java
edwin-greene: @steven-sheehy what do you think about setting the value to 3? I believe the original intent in the block streams design was to not persist this value to save on space, since it will only change if a new HIP is approved.

xin-hedera: we have to set it to some value in block item transformer (since if not set, it defaults to 0), likely 3. however this doesn't mean we need to get rid of this PR.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/transactionhandler/ConsensusSubmitMessageTransactionHandler.java
edwin-greene: Thanks for catching this, I have added that note to the design doc. Is that sufficient to cover the potential problem with the value defaulting to 0?

hedera-mirror-grpc/src/main/java/com/hedera/mirror/grpc/controller/ConsensusController.java
edwin-greene: Great suggestion, fixed.

hedera-mirror-rest/viewmodel/topicMessageViewModel.js
```diff
@@ -26,19 +26,26 @@ import {encodeBase64, encodeBinary, nsToSecNs} from '../utils';
...
+  static DEFAULT_RUNNING_HASH_VERSION = 3;
+
   /**
    * Constructs topicMessage view model
    *
    * @param {TopicMessage} topicMessage
    * @param {String} messageEncoding the encoding to display the message in
    */
   constructor(topicMessage, messageEncoding) {
+    const runningHashVersion = _.isNil(topicMessage.runningHashVersion)
```
xin-hedera, line 39: nit, can simplify, can also make it a one-liner to where it's assigned to `this.running_hash_version`

```suggestion
    const runningHashVersion = topicMessage.runningHashVersion ?? TopicMessageViewModel.DEFAULT_RUNNING_HASH_VERSION;
```

edwin-greene: Excellent, thanks I have updated it.

xin-hedera:
rest-java tests are failing due to issue #9975. you'll have to exclude the failing topic spec files.

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/RestSpecTest.java
```diff
@@ -64,7 +64,7 @@
         properties = {"spring.main.allow-bean-definition-overriding=true"})
 public class RestSpecTest extends RestJavaIntegrationTest {
     private static final Pattern INCLUDED_SPEC_DIRS = Pattern.compile(
-            "^(accounts|accounts/\\{id}/allowances.*|accounts/\\{id}/rewards.*|blocks.*|contracts|network/exchangerate.*|network/fees.*|network/stake.*|topics/\\{id}/messages)$");
+            "^(accounts|accounts/\\{id}/allowances.*|accounts/\\{id}/rewards.*|blocks.*|contracts|network/exchangerate.*|network/fees.*|network/stake.*)$");
```
steven-sheehy, line 67: Let's make sure we remember to re-enable this. Or create a ticket to track.

steven-sheehy (approved):

Issue #10031, opened 2025-01-02
Bump hedera app dependency to 0.58 in order to configure `maxSignedTxnSize`


bilyana-gospodinova:
### Problem

As described in the issue here - https://github.com/hashgraph/hedera-services/issues/17099 PR https://github.com/hashgraph/hedera-services/pull/17142 was merged in services. We need to wait until it is released and adopt the change in our code.

### Solution

Update the hedera app dependency to version 0.58. After that we need to increase the `maxSignedTxnSize` in the map of the properties that is passed to the TransactionExecutor, e.g. `Map.of("executor.maxSignedTxnSize", "262144")`. After increasing the size we can remove the logic of uploading the contracts as a file and creating them in a separate transaction and we can only use the second case where the init bytecode is passed directly in a single transaction to create a contract. This will improve the performance and will fix some other failing tests that depend on the transaction size.

PR #10218, opened 2025-01-24
Remove file upload on contract create and use `maxSignedTxnSize` property

bilyana-gospodinova:
**Description**:
After the hedera app dependency bump to 0.58.x we now have available the `maxSignedTxnSize` property. This allows us to set it to a value higher than 6KB that was the default one before. The configured callData size in the requests is currently 128 KB. The size of the call data + the additional transaction fields and signing increase the overall transaction size a little (about 50-200 bytes depending on the request), so value of `maxSignedTxnSize` is now set to 129KB to ensure that there is enough buffer and that the transaction will not be rejected in hedera app.

Setting this property allows us to:
1. Remove the checks in `TransactionExecutionService` for transaction size.
2. Remove the file upload logic for contract create when the transaction is bigger that 6KB.
3. Remove the `file` field in the `ContractCallContext` that saved the content of the uploaded file between the separate transaction executions (file uplaod & contract create).
4. Clean the `FileReadableKVState` from using the `file` field from the `ContractCallContext`.

This makes the overall code a lot cleaner and easier to understand.

Fixes https://github.com/hashgraph/hedera-mirror-node/issues/10031

### DIFF
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/common/ContractCallContext.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/common/ContractCallContext.java
@@ -16,9 +16,6 @@
 
 package com.hedera.mirror.web3.common;
 
-import com.hedera.hapi.node.contract.ContractCreateTransactionBody;
-import com.hedera.hapi.node.file.FileCreateTransactionBody;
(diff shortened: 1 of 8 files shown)


steven-sheehy (dismissed):
LGTM

PR #10220, opened 2025-01-24
Fix rest java specs failing when they're changed

### LINKED ISSUE #9975: RestJava RestSpecTest will likely fail when the selected specs are changed

steven-sheehy:
**Description**:

* Fix REST java specs failing due to out of date container by always pulling latest

**Related issue(s)**:

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/config/SpecTestConfig.java
+++ b/hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/config/SpecTestConfig.java
@@ -31,6 +31,7 @@ import org.testcontainers.containers.GenericContainer;
 import org.testcontainers.containers.Network;
 import org.testcontainers.containers.PostgreSQLContainer;
 import org.testcontainers.containers.output.Slf4jLogConsumer;
+import org.testcontainers.images.PullPolicy;
 import org.testcontainers.utility.DockerImageName;
 
 @TestConfiguration(proxyBeanMethods = false)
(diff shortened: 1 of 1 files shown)


edwin-greene (approved):
Looks good

hedera-mirror-rest-java/src/test/java/com/hedera/mirror/restjava/spec/config/SpecTestConfig.java
```diff
@@ -81,8 +82,9 @@ GenericContainer<?> jsRestApi(PostgreSQLContainer<?> postgresql, Network prostgr
 
         return new GenericContainer<>(DockerImageName.parse("gcr.io/mirrornode/hedera-mirror-rest:latest"))
                 .dependsOn(postgresql)
-                .withNetwork(prostgresqlNetwork)
                 .withExposedPorts(5551)
-                .withEnv(envBuilder.build());
+                .withEnv(envBuilder.build())
+                .withImagePullPolicy(PullPolicy.alwaysPull())
```
xin-hedera, line 87: since the `latest` image tag points to the image tagged and pushed by release-production github workflow jobs, the spec tests can still fail when fixes in the js rest module are only in SNAPSHOT.

We should instead use the `main` image when it's a SNAPSHOT build and use the `latest` image otherwise.

It can still fail when we tag in release branches as the `latest` image will again not match what the test class expects. That is transient failure and should work once the images of the new tag are published.

xin-hedera: umm. actually I'm wrong. anytime we make changes to js REST API module that the spec tests need the matching changes to pass, the spec tests in rest java will still fail.

steven-sheehy: So this PR fixes a real issue I experienced locally and also that Edwin has seen in his PR. We definitely need this change, but maybe it doesn't fully fix the ticket. I can find a way to feed the gradle `version` property to the test config and use that version to pull. Will that not fix most of the problems except during the period where that image is being replaced which should auto resolve on retry after publishing?

xin-hedera: nope, using SNAPSHOT images won't fully fix the issue. think we can leave it as is. what we really need is a js rest image built right from the PR branch or a way to programmatically run the js rest api service from source.

xin-hedera (approved):
LGTM

PR #10217, opened 2025-01-24
Modularized fix ContractCallNestedCallsTest

kselveliev:
**Description**:
This Pr fixes all failing tests in ContractCallNestedCallsTest against modularized. 20+ test fixes. 
Including an added check in HederaTokenStore for checking if the treasury for nft update is the same then skip checks.

This PR modifies ... in order to support ...
ContractCallServicePrecompileModification - Extract token info setup and ethCall and gas assertions into an abstract class to be used from the NestedCallsTests. Since NestedCalls setup was incomplete.
ContractCallNestedCallsTest - use exctracted setup from ContractCallServicePrecompileModification for token update operations.

HederaTokenStore - added a check for nft update - 
```
!changes.getTreasury()
                .equals(token.getTreasury().getId()
                        .asGrpcAccount())

``` 
otherwise we get TRANSACTION_REQUIRES_ZERO_TOKEN_BALANCES if we try to update the token nft passing the same treasury. This was previouslly bypassed by not persisting token association with the treasury in the test setup which seems incorrect. Modularized code has similar check ->

` if (!newTreasury.equals(existingTreasury)) {`

**Related issue(s)**:

Fixes #10082 

**Notes for reviewer**:
3358 tests completed, 247 failed, 6 skipped

**Checklist**

- [ ] Documented (Code comments, README, etc.)
- [x] Tested (unit, integration, etc.)

Issue #10090, opened 2025-01-10
Fix `OpcodeServiceTest` test suite


nickeynikolovv:
Fix failing tests for `OpcodeServiceTest` using the modularized EVM

PR #10247, opened 2025-01-27
Fix OpcodeServiceTest for modularized

kselveliev:
**Description**:
This PR fixes 80+ tests from OpcodeServiceTest 
Fixed ContractCallDynamicCallsTest- mintTokenGetTotalSupplyAndBalanceOfTreasury which is also used in OpcodeServiceTest.
PR is checkout from https://github.com/hashgraph/hedera-mirror-node/pull/10217 since it contains same setup in as in NestedCalls tests.
Must wait for https://github.com/hashgraph/hedera-mirror-node/pull/10217 to be merged first.
The only test that is not passing is `callWithDifferentCombinationsOfTracerOptions` with the exception of the test that set storage to true in OpcodeTracer().  
```
   public Map<UInt256, UInt256> getUpdatedStorage() {
        throw new UnsupportedOperationException("Storage changes are managed by the EvmFrameState");
    }
``` 
fails in upstream code as not supported.

 A few tests in updateTokenKeysAndGetUpdatedTokenKeyForNFT with CONSENSUS_GAS_EXHAUSTED but this will be fixed in another PR.

OpcodeServiceTest - add proper test setup as in nested calls tests

**Related issue(s)**:
Part of https://github.com/hashgraph/hedera-mirror-node/issues/10090

**Notes for reviewer**:

3358 tests completed, 150 failed, 6 skipped

**Checklist**


hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/OpcodeServiceTest.java
```diff
@@ -364,21 +385,28 @@ void deleteTokenAndGetTokenInfoIsDeleted(final TokenTypeEnum tokenType) {
...
     void createFungibleTokenAndGetIsTokenAndGetDefaultFreezeStatusAndGetDefaultKycStatus(
             final boolean withKeys,
             final boolean inheritKey,
-            final boolean defaultKycStatus,
+            boolean defaultKycStatus,
             final boolean defaultFreezeStatus) {
         // Given
-        final var treasuryEntity = accountPersistWithBalance(CREATE_TOKEN_VALUE);
+        if (mirrorNodeEvmProperties.isModularizedServices()) {
+            defaultKycStatus = !defaultKycStatus;
```
bilyana-gospodinova, line 399: Why do we need this?

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/OpcodeServiceTest.java
```diff
@@ -861,45 +858,40 @@ private void persistContractTransactionHash(
...
                 .persist();
     }
 
-    private Token nftPersist(final EntityId treasuryEntityId) {
-        final var tokenEntity = domainBuilder
-                .entity()
-                .customize(e -> e.type(TOKEN).autoRenewAccountId(treasuryEntityId.getId()))
-                .persist();
+    private Token nftPersist(final Entity treasuryEntity) {
+        final var nftEntity =
```
bilyana-gospodinova, line 877: We have existing `tokenEntityPersist()` method for this line.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/OpcodeServiceTest.java
```diff
@@ -911,7 +903,7 @@ private Entity accountPersist() {
     private Entity accountPersistWithBalance(final long balance) {
         final var entity = domainBuilder
                 .entity()
-                .customize(e -> e.evmAddress(null).balance(balance))
+                .customize(e -> e.evmAddress(null).alias(null).balance(balance))
```
bilyana-gospodinova, line 906: Can't we reuse the `accountEntityPersist()` method from the parent?

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/OpcodeServiceTest.java
kselveliev: A bit tricky to explain. I spoke with Stani from smart contracts and seems our mono code works the opposite way. 
There are 3 possible kyc statuses but 2 booleans.

`KycNotApplicable`: This status is returned when the token does not have a KYC key set. In this case, the KYC boolean will return TRUE.

`Granted`: This status indicates that KYC has been granted for the account. The KYC boolean will return true.
`Revoked`: This status means that KYC has been revoked for the account. The KYC boolean will return false.

If we create the token without kyc the status returned is TRUE. KycNotApplicable

if we create the token with kyc key the initial status will be Revoked -> returns false

Other case when we need to return TRUE status -> 
when you create a token with KYC key:
The initial KYC status for accounts will be effectively "Revoked". -> returns FALSE
Accounts will not be able to transact with the token initially.
You will need to explicitly grant KYC. then it will return TRUE

bilyana-gospodinova: Thank you for the detailed explanation. I think we shouldn't reassign the input method parameter. If we need some kind of flag flip, I think it would be better if we do this somewhere around: `getExpectedResultFromBooleans(defaultKycStatus,...)` with a comment why we flip the flag similar to your comment here.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/OpcodeServiceTest.java
kselveliev: We can `accountEntityPersist()` in the` accountPersisteWithBalance(..)` and it's okay

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/OpcodeServiceTest.java
kselveliev: Comment addressed

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/OpcodeServiceTest.java
kselveliev: Comment addressed

bilyana-gospodinova (approved):
LGTM

steven-sheehy (approved):

PR #10248, opened 2025-01-27
HIP-1056 Add File BlockItem to RecordItem transformers

### LINKED ISSUE #10166: HIP-1056 File BlockItem to RecordItem transformers
### Problem

Add block item to record item transformers.

### Solution

Add transformers for the following transaction types:
FileCreate
FileDelete
FileUpdate
FileAppend

### Alternatives

_No response_

filev94:
**Description**:

Adding transformers for FileCreate, FileAppend, FileDelete and FileUpdate

**Related issue(s)**:

Fixes #10166

**Notes for reviewer**:

Created the above said changes

**Checklist**

### DIFF
--- a/hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/downloader/block/BlockFileTransformerTest.java
+++ b/hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/downloader/block/BlockFileTransformerTest.java
@@ -268,7 +268,7 @@ class BlockFileTransformerTest extends ImporterIntegrationTest {
     }
 
     @Test
-    void fileCreateTransform_whenStatusIsNotSuccess() {
+    void fileCreateTransformWhenStatusIsNotSuccess() {
         // given
         var expectedRecordItem = recordItemBuilder
                 .fileCreate()


hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/FileCreateTransformer.java
```diff
@@ -0,0 +1,50 @@
...
+    protected void updateTransactionRecord(BlockItem blockItem, TransactionRecord.Builder transactionRecordBuilder) {
+        for (StateChanges stateChange : blockItem.stateChanges()) {
+            for (StateChange change : stateChange.getStateChangesList()) {
+                if (Objects.nonNull(change)
+                        && change.hasMapUpdate()
+                        && change.getMapUpdate().hasKey()
+                        && change.getMapUpdate().getKey().hasFileIdKey()) {
+
+                    var fileIdKey = change.getMapUpdate().getKey().getFileIdKey();
+                    transactionRecordBuilder.getReceiptBuilder().setFileID(fileIdKey);
```
edwin-greene, line 40: It may be that you don't need to check for null; that `getStateChangesList()` is guaranteed to not have any null elements.

I'm unsure about the performance increase from combining all of the predicates into a single if statement, vs getting the values a single time and then having a nested if statement... In this case `change.getMapUpdate()` is called 3 times, and `change.getMapUpdate().getKey()` is called twice. It may be that there isn't any difference in performance one way vs the other?

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/FileUpdateTransformer.java
```diff
@@ -0,0 +1,45 @@
...
+@Named
+public class FileUpdateTransformer extends AbstractBlockItemTransformer {
+
+    @Override
+    protected void updateTransactionRecord(BlockItem blockItem, TransactionRecord.Builder transactionRecordBuilder) {
+        for (StateChanges stateChange : blockItem.stateChanges()) {
+            for (StateChange change : stateChange.getStateChangesList()) {
+                if (change.hasSingletonUpdate() && change.getSingletonUpdate().hasExchangeRateSetValue()) {
+                    var exchangeRate = change.getSingletonUpdate().getExchangeRateSetValue();
+                    transactionRecordBuilder.getReceiptBuilder().setExchangeRate(exchangeRate);
```
edwin-greene, line 35: Can you break from both loops once an exchange rate has been found?

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/downloader/block/BlockFileTransformerTest.java
```diff
@@ -173,6 +173,139 @@ void unknownTransform() {
...
+                                transactionRecord.getReceipt().getFileID().getFileNum()));
+    }
+
+    @Test
+    void fileUpdateTransform() {
+        // given
+        var expectedRecordItem = recordItemBuilder
+                .fileUpdate()
+                .recordItem(r -> r.hapiVersion(HAPI_VERSION))
+                .build();
```
edwin-greene, line 268: Rather than excluding `transactionRecord.receipt_.exchangeRate_` from `assertRecordFile`, you should be able set the receipt:

```
var expectedRecordItem = recordItemBuilder
                .fileUpdate()
                .recordItem(r -> r.hapiVersion(HAPI_VERSION))
                .receipt(r -> r.setExchangeRate(ExchangeRateSet.newBuilder().setCurrentRate(ExchangeRate.newBuilder().setCentEquiv(1).build()).build()))
                .build();
```

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/downloader/block/BlockFileTransformerTest.java
```diff
@@ -227,6 +360,7 @@ private void assertRecordItem(RecordItem recordItem, RecordItem expectedRecordIt
                         "transactionRecord.assessedCustomFees_",
                         "transactionRecord.scheduleRef_",
                         "transactionRecord.parentConsensusTimestamp_",
+                        "transactionRecord.receipt_.exchangeRate_",
```
edwin-greene, line 363: Should not need to ignore the exchange rate.

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/domain/BlockItemBuilder.java
```diff
@@ -16,18 +16,11 @@
 
 package com.hedera.mirror.importer.parser.domain;
 
-import com.hedera.hapi.block.stream.output.protoc.CallContractOutput;
-import com.hedera.hapi.block.stream.output.protoc.CryptoTransferOutput;
-import com.hedera.hapi.block.stream.output.protoc.StateChanges;
-import com.hedera.hapi.block.stream.output.protoc.TransactionOutput;
-import com.hedera.hapi.block.stream.output.protoc.TransactionResult;
+import com.hedera.hapi.block.stream.output.protoc.*;
```
edwin-greene, line 19: Mirror Node code standard is not to use wildcards for imports.

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/domain/BlockItemBuilder.java
```diff
@@ -88,6 +81,71 @@ public BlockItemBuilder.Builder unknown(RecordItem recordItem) {
...
+
+        var change =
+                StateChange.newBuilder().setSingletonUpdate(singletonUpdate).build();
+        var stateChanges = StateChanges.newBuilder().addStateChanges(change).build();
+
+        return new BlockItemBuilder.Builder(
+                recordItem.getTransaction(), transactionResult, List.of(), List.of(stateChanges));
+    }
+
+    private static StateChanges buildStateChanges(RecordItem recordItem) {
```
edwin-greene, line 139: Can you change the name to something like buildFileIdStateChanges? Other tests will likely have state changes and the name of this method looks generic too all of them.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/FileAppendTransformer.java
```diff
@@ -0,0 +1,29 @@
...
+ * limitations under the License.
+ */
+
+package com.hedera.mirror.importer.downloader.block.transformer;
+
+import com.hedera.mirror.common.domain.transaction.TransactionType;
+import jakarta.inject.Named;
+
+@Named
+public class FileAppendTransformer extends AbstractBlockItemTransformer {
```
steven-sheehy, line 23: nit: None of the transformers need to be public and can be final
```suggestion
final class FileAppendTransformer extends AbstractBlockItemTransformer {
```

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/FileUpdateTransformer.java
steven-sheehy, line 35: We don't use the exchange rate in the receipt in the mirror node. We can omit mapping this. If we did want to map it then I believe it would be done for every transaction generically and not just for file update.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/FileDeleteTransformer.java
```diff
@@ -0,0 +1,50 @@
...
+                if (Objects.nonNull(change)
+                        && change.hasMapUpdate()
+                        && change.getMapUpdate().hasKey()
+                        && change.getMapUpdate().getKey().hasFileIdKey()) {
+
+                    var fileIdKey = change.getMapUpdate().getKey().getFileIdKey();
+                    transactionRecordBuilder.getReceiptBuilder().setFileID(fileIdKey);
+                }
+            }
+        }
```
steven-sheehy, line 43: We can omit this since I don't believe the receipt.fileId is set except for create and we get the ID from the body anyway.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/FileCreateTransformer.java
```diff
@@ -0,0 +1,50 @@
...
+import jakarta.inject.Named;
+import java.util.Objects;
+
+@Named
+public class FileCreateTransformer extends AbstractBlockItemTransformer {
+
+    @Override
+    protected void updateTransactionRecord(BlockItem blockItem, TransactionRecord.Builder transactionRecordBuilder) {
+        for (StateChanges stateChange : blockItem.stateChanges()) {
+            for (StateChange change : stateChange.getStateChangesList()) {
```
steven-sheehy, line 33: Use var

steven-sheehy (changes requested):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/FileCreateTransformer.java
filev94: Fixed

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/FileUpdateTransformer.java
filev94: Removed the whole method since it is not needed anymore

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/FileAppendTransformer.java
filev94: Fixed

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/FileDeleteTransformer.java
filev94: In the tests the fileDelete is built with the fileId set in the receipt. So i removed this code and added the transactionBody to the updateTransactionRecord so i can set the fileId in the receipt here.

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/domain/BlockItemBuilder.java
filev94: Changed to buildFileIdStateChanges

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/domain/BlockItemBuilder.java
```diff
@@ -88,6 +81,71 @@ public BlockItemBuilder.Builder unknown(RecordItem recordItem) {
...
+
+    public Builder fileUpdate(RecordItem recordItem) {
+        var instant = Instant.ofEpochSecond(0, recordItem.getConsensusTimestamp());
+        var timestamp = Utility.instantToTimestamp(instant);
+        var transactionRecord = recordItem.getTransactionRecord();
+        var transactionResult = transactionResult(transactionRecord, timestamp).build();
+
+        var exchangeRate = ExchangeRateSet.newBuilder()
+                .setCurrentRate(ExchangeRate.newBuilder().setCentEquiv(1).build())
+                .build();
```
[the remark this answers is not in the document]
filev94: Since Steven said the exchange rate is not used currently, i've removed it.

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/domain/BlockItemBuilder.java
filev94: Fixed

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/downloader/block/BlockFileTransformerTest.java
filev94: As per Steven's comment, exchange rate is not used currently, so i removed it.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/FileCreateTransformer.java
```diff
@@ -0,0 +1,47 @@
...
+@Named
+final class FileCreateTransformer extends AbstractBlockItemTransformer {
+
+    @Override
+    protected void updateTransactionRecord(
+            BlockItem blockItem, TransactionRecord.Builder transactionRecordBuilder, TransactionBody transactionBody) {
+        outerLoop:
+        for (var stateChange : blockItem.stateChanges()) {
+            for (var change : stateChange.getStateChangesList()) {
+                if (change != null) {
```
xin-hedera, line 34: several issues to address

- change won't be null so the null check isn't needed
- the state changes are structured in a way that to check for a file entity id, we should do
  - check if `change.getStateId == StateIdentifier.StateIdentifier.STATE_ID_FILES`
  - check if the the type of state change is map update, i.e., `change.hasMapUpdate()`
  - check if mapupdate change key is file, i.e., `MapChangeValue.hasFileIdKey()`
  - get the file id from the key


We have to go through all checks since even the simplest transaction still cause account balance changes so the statechanges will contain mapupdate changes for many accounts.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/FileDeleteTransformer.java
```diff
@@ -0,0 +1,40 @@
...
+@Named
+final class FileDeleteTransformer extends AbstractBlockItemTransformer {
+
+    @Override
+    protected void updateTransactionRecord(
+            BlockItem blockItem, TransactionRecord.Builder transactionRecordBuilder, TransactionBody transactionBody) {
+        transactionRecordBuilder
+                .getReceiptBuilder()
+                .setFileID(transactionBody.getFileDelete().getFileID());
+    }
```
xin-hedera, line 34: protobuf doc says receipt.file_id is only set for `FileCreate`

```suggestion
```

xin-hedera (changes requested):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/FileCreateTransformer.java
```diff
@@ -0,0 +1,47 @@
...
+                    var fileId = change.getMapUpdate().getKey().getFileIdKey();
+                    transactionRecordBuilder.getReceiptBuilder().setFileID(fileId);
+                    break outerLoop;
```
xin-hedera, line 37: nit, can just `return` instead of `break`

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/FileCreateTransformer.java
filev94: Added the additional checks

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/FileDeleteTransformer.java
filev94: Removed the method altogether + removed the fileId being set in the RecordItemBuilder

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/FileCreateTransformer.java
filev94: Fixed

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/FileCreateTransformer.java
```diff
@@ -0,0 +1,55 @@
...
+                        if (mapUpdate.hasKey()) {
+                            var key = mapUpdate.getKey();
+                            if (key.hasFileIdKey()) {
+                                var fileId = key.getFileIdKey();
+                                transactionRecordBuilder.getReceiptBuilder().setFileID(fileId);
+                                return;
+                            }
+                        }
+                    }
+                }
```
xin-hedera, line 46: prefer less nested levels. for `MapUpdate.key`, `hasKey()` is unnecessary, because `getKey()` returns the default `MapChangeKey` instance, and `key.hasFileIdKey()` will be false.

```suggestion
                if (change.getStateId() == StateIdentifier.STATE_ID_FILES.getNumber() && change.hasMapUpdate()) {
                    var key = change.getMapUpdate().getKey();
                    if (key.hasFileIdKey()) {
                        transactionRecordBuilder.getReceiptBuilder().setFileID(key.getFileIdKey);
                        return;
                    }
                }
```

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/FileCreateTransformer.java
```diff
@@ -0,0 +1,55 @@
...
+import com.hederahashgraph.api.proto.java.TransactionRecord;
+import jakarta.inject.Named;
+
+@Named
+final class FileCreateTransformer extends AbstractBlockItemTransformer {
+
+    @Override
+    protected void updateTransactionRecord(
+            BlockItem blockItem, TransactionRecord.Builder transactionRecordBuilder, TransactionBody transactionBody) {
+        for (var stateChange : blockItem.stateChanges()) {
```
xin-hedera, line 32: shortcut and return if transaction result is not `SUCCESS`.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/FileCreateTransformer.java
filev94: Fixed

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/FileCreateTransformer.java
filev94: Added the check to be first thing in the method

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/domain/RecordItemBuilder.java
```diff
@@ -657,10 +657,16 @@ public Builder<FileCreateTransactionBody.Builder> fileCreate() {
         return new Builder<>(TransactionType.FILECREATE, builder).receipt(b -> b.setFileID(fileId()));
     }
 
+    public Builder<FileCreateTransactionBody.Builder> fileCreateAuthorizationFailed() {
```
xin-hedera, line 660: should move this to the test case. instead of repeating some code, should to

```
   recordItemBuilder.fileCreate()
     .receipt(r -> r.clearFileID().setStatus(ResponseCodeEnum.AUTHORIZATION_FAILED))
     .build();
```

filev94: Agreed, much simpler that way.

xin-hedera (approved):
LGTM

edwin-greene (approved):
Looks good to me

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/AbstractBlockItemTransformer.java
```diff
@@ -48,11 +48,12 @@ public TransactionRecord getTransactionRecord(BlockItem blockItem, TransactionBo
...
                 .setTransferList(transactionResult.getTransferList());
 
-        updateTransactionRecord(blockItem, transactionRecordBuilder);
+        updateTransactionRecord(blockItem, transactionRecordBuilder, transactionBody);
         return transactionRecordBuilder.build();
     }
 
-    protected void updateTransactionRecord(BlockItem blockItem, TransactionRecord.Builder transactionRecordBuilder) {
+    protected void updateTransactionRecord(
+            BlockItem blockItem, TransactionRecord.Builder transactionRecordBuilder, TransactionBody transactionBody) {
```
steven-sheehy, line 56: This transactionBody was added but is now unused. So we can revert.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/FileAppendTransformer.java
```diff
@@ -0,0 +1,29 @@
...
+ * limitations under the License.
+ */
+
+package com.hedera.mirror.importer.downloader.block.transformer;
+
+import com.hedera.mirror.common.domain.transaction.TransactionType;
+import jakarta.inject.Named;
+
+@Named
+final class FileAppendTransformer extends AbstractBlockItemTransformer {
```
steven-sheehy, line 23: We can avoid adding these no-op transformers since BlockItemTransformerFactory will default to a no-op one already.

steven-sheehy (changes requested):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/AbstractBlockItemTransformer.java
filev94: Reverted

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/FileAppendTransformer.java
filev94: Deleted the no-op transformers

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/FileCreateTransformer.java
```diff
@@ -0,0 +1,53 @@
...
+import com.hederahashgraph.api.proto.java.TransactionRecord;
+import jakarta.inject.Named;
+
+@Named
+final class FileCreateTransformer extends AbstractBlockItemTransformer {
+
+    @Override
+    protected void updateTransactionRecord(BlockItem blockItem, TransactionRecord.Builder transactionRecordBuilder) {
+
+        if (blockItem.transactionResult().getStatus() != ResponseCodeEnum.SUCCESS) {
```
steven-sheehy, line 32: This isn't the only success response code. We need to replicate the logic in RecordItem.parseSuccess()

steven-sheehy, line 32: This isn't the only success response code. We need to replicate the logic in RecordItem.parseSuccess()

steven-sheehy, line 32: This isn't the only success response code. We need to replicate the logic in RecordItem.parseSuccess()

filev94: Added two more success codes as per parseSuccess() method

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/FileCreateTransformer.java
```diff
@@ -28,8 +28,10 @@ final class FileCreateTransformer extends AbstractBlockItemTransformer {
 
     @Override
     protected void updateTransactionRecord(BlockItem blockItem, TransactionRecord.Builder transactionRecordBuilder) {
-
-        if (blockItem.transactionResult().getStatus() != ResponseCodeEnum.SUCCESS) {
+        var status = blockItem.transactionResult().getStatus();
+        if (status != ResponseCodeEnum.SUCCESS
+                && status != ResponseCodeEnum.FEE_SCHEDULE_FILE_PART_UPLOADED
+                && status != ResponseCodeEnum.SUCCESS_BUT_MISSING_EXPECTED_OPERATION) {
```
steven-sheehy, line 34: Move to `BlockItem.isSuccessful()`. Also add `if (parent != null && !parent.isSuccessful()) return false` logic

filev94: Added isSuccessful logic as per parseSuccess method + parseParent

hedera-mirror-common/src/test/java/com/hedera/mirror/common/domain/transaction/BlockItemTest.java
```diff
@@ -0,0 +1,287 @@
...
+
+    static Stream<ResponseCodeEnum> validStatuses() {
+        return Stream.of(
+                ResponseCodeEnum.FEE_SCHEDULE_FILE_PART_UPLOADED,
+                ResponseCodeEnum.SUCCESS,
+                ResponseCodeEnum.SUCCESS_BUT_MISSING_EXPECTED_OPERATION);
+    }
+
+    @ParameterizedTest
+    @MethodSource("validStatuses")
```
steven-sheehy, line 42: `@EnumSource` is more appropriate

hedera-mirror-common/src/test/java/com/hedera/mirror/common/domain/transaction/BlockItemTest.java
```diff
@@ -0,0 +1,287 @@
...
+    void parseSuccess_whenNoParentPresentAndCorrectStatus_returnTrue(ResponseCodeEnum status) {
```
steven-sheehy, line 43: We don't use this underscore syntax. You should use `@DisplayName` if you want more verbose test names with variable replacement.
```suggestion
    @DisplayName("isSuccessful() returns true")
    @EnumSource(value = ResponseCodeEnum.class, mode = INCLUDE, names = "SUCCESS,...")
    @ParameterizedTest("with status {0}")
    void isSuccessfulWithSuccessEnums(ResponseCodeEnum status) {
```

hedera-mirror-common/src/test/java/com/hedera/mirror/common/domain/transaction/BlockItemTest.java
```diff
@@ -0,0 +1,287 @@
...
+import com.hederahashgraph.api.proto.java.ResponseCodeEnum;
+import com.hederahashgraph.api.proto.java.Timestamp;
+import com.hederahashgraph.api.proto.java.Transaction;
+import java.util.List;
+import java.util.stream.Stream;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+
+public class BlockItemTest {
```
steven-sheehy, line 32: Test classes should always be package private

```suggestion
class BlockItemTest {
```

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/domain/RecordItemBuilder.java
```diff
@@ -1302,6 +1302,27 @@ public enum TransferType {
         TOKEN
     }
 
+    @Getter
+    public enum SuccessfulStatusCreateFile {
```
steven-sheehy, line 1306: a) This enum should not be in RecordItemBuilder.
b) I don't believe it needs to exist at all. Should be able to just use `@EnumSource`

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/downloader/block/BlockFileTransformerTest.java
```diff
@@ -173,6 +176,148 @@ void unknownTransform() {
...
+                .extracting(RecordItem::getTransactionRecord)
+                .returns(expectedTransactionHash, TransactionRecord::getTransactionHash)
+                .returns(
+                        expectedFileId,
+                        transactionRecord ->
+                                transactionRecord.getReceipt().getFileID().getFileNum()));
+    }
+
+    @Test
+    void fileCreateTransform_whenStatusIsNotSuccess() {
```
steven-sheehy, line 268: Same comment in all tests. Don't use underscores in method names.

```suggestion
    void fileCreateTransformWhenStatusIsNotSuccess() {
```

hedera-mirror-common/src/test/java/com/hedera/mirror/common/domain/transaction/BlockItemTest.java
```diff
@@ -0,0 +1,287 @@
...
+
+        // Then: The parent should match the previous block item
+        assertEquals(
+                previousBlockItem, parent, "Parent should match the previous block item based on consensus timestamp");
+    }
+
+    @Test
+    void parseParent_whenConsensusTimestampDoNotMatch_NoParent() {
+        // Given: Create a previous block item with a non-matching parent consensus timestamp
+        TransactionResult transactionResult = TransactionResult.newBuilder()
```
steven-sheehy, line 202: Use var for all non-primitives. Same comment for all tests in this file.

hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/transaction/BlockItem.java
```diff
@@ -20,14 +20,54 @@
...
+        boolean isSuccessful)
+        implements StreamItem {
+
+    // Custom Builder
+    public static class BlockItemBuilder {
+        public BlockItem build() {
+            parent = parseParent();
+            isSuccessful = parseSuccess();
+            return builderInternal();
+        }
```
steven-sheehy, line 45: Why can't we use the record custom constructor to do this logic? Then it doesn't force us to only create via the builder.
```suggestion
    public BlockItem {
        parent = parseParent();
        isSuccessful = parseSuccess();
    }
```

hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/transaction/BlockItem.java
```diff
@@ -20,14 +20,54 @@
...
 public record BlockItem(
         Transaction transaction,
         TransactionResult transactionResult,
         List<TransactionOutput> transactionOutput,
-        List<StateChanges> stateChanges)
-        implements StreamItem {}
+        List<StateChanges> stateChanges,
+        BlockItem parent,
+        BlockItem previous,
+        boolean isSuccessful)
```
steven-sheehy, line 36:
```suggestion
        boolean successful)
```

steven-sheehy (changes requested):

hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/transaction/BlockItem.java
edwin-greene: Just to note, I've updated my PR with a similar change in this overlapping area. Not sure if it is exactly what we want, but this is how I've implemented it thus far:
 https://github.com/hashgraph/hedera-mirror-node/pull/10254

hedera-mirror-common/src/test/java/com/hedera/mirror/common/domain/transaction/BlockItemTest.java
filev94: Fixed

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/domain/RecordItemBuilder.java
filev94: Removed the enum.

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/downloader/block/BlockFileTransformerTest.java
filev94: Fixed

hedera-mirror-common/src/test/java/com/hedera/mirror/common/domain/transaction/BlockItemTest.java
filev94: Fixed

hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/transaction/BlockItem.java
filev94: I deleted the BlockItemBuilder and used the default constructor. The builder pattern uses the default constructor as well so there shouldn't be an issue.

hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/transaction/BlockItem.java
filev94: I changed this but now the getter is just successful(). I didn't want to override it to become isSuccessful because in my opinion it is clear what we check in the if.

filev94:
The codecov fails for FileCreateTransformer (72%, 92 needed) because 2 closing brackets in the end of the logic are not tested. I feel like if i need to test them i have to cheat somehow (we need to have no stateChanges, but then there is issue with the FileId being set in the RecordItemBuilder)

hedera-mirror-common/src/test/java/com/hedera/mirror/common/domain/transaction/BlockItemTest.java
filev94: Swapped to EnumSource. My only concern is that it is used 3 times and if a change is needed someone has to change it in the 3 places. That's why i initially went with methodSource

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/downloader/block/BlockFileTransformerTest.java
steven-sheehy: Still has underscore

steven-sheehy (dismissed):
LGTM

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/downloader/block/BlockFileTransformerTest.java
filev94: Sorry, totally missed this one lone test.

steven-sheehy (approved):

edwin-greene (approved):
Looks good

PR #10251, opened 2025-01-28
Fix the ERC historical tests

### LINKED ISSUE #10083: Fix `ContractCallServiceERCTokenHistoricalTest` test suite
Fix failing tests for `ContractCallServiceERCTokenHistoricalTest` using the modularized EVM

bilyana-gospodinova:
**Description**:
This PR fixes the setup for the ERC historical tests. Also, the timestamp from the context is now fetched differently:
- If there is timestamp set directly -> return this timestamp.
- If there isn't -> return the timestamp from the consensus end timestamp of the set record file in the context.
The above points are valid only if we are in a historical context. Otherwise -> return empty optional.

Note: Changing the timestamp fetch logic now enables us to read historical records correctly. However, there is a cache mechanism in `ReadableKVStateBase` and when multiple test suites are run it gets polluted and some tests sometimes fail and pass when run separately. This is a known issue but the change from this PR seems to make it appear more frequently. This is not a regression.

Fixes #10083

### DIFF
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/common/ContractCallContext.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/common/ContractCallContext.java
@@ -125,9 +125,7 @@ public class ContractCallContext {
      */
     public void initializeStackFrames(final StackedStateFrames stackedStateFrames) {
         if (stackedStateFrames != null) {
-            final var stateTimestamp = timestamp.isPresent()
-                    ? timestamp
-                    : Optional.ofNullable(recordFile).map(RecordFile::getConsensusEnd);
+            final var stateTimestamp = getTimestampOrDefaultFromRecordFile();
             stackBase = stack = stackedStateFrames.getInitializedStackBase(stateTimestamp);
         }
     }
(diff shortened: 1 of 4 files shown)


hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
```diff
@@ -119,7 +119,6 @@ protected void setup() {
                                 treasuryEntity.getCreatedTimestamp(), treasuryEntity.toEntityId()))
                         .balance(treasuryEntity.getBalance()))
                 .persist();
-        testWeb3jService.reset();
```
bilyana-gospodinova, line 122: This line is called in the `cleanup()` method, so there is no need for this duplication.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/common/ContractCallContext.java
```diff
@@ -142,4 +140,20 @@ public boolean useHistorical() {
...
+        if (useHistorical()) {
+            return getTimestampOrDefaultFromRecordFile();
+        }
+        return Optional.empty();
+    }
+
+    private Optional<Long> getTimestampOrDefaultFromRecordFile() {
+        return timestamp.isPresent()
+                ? timestamp
+                : Optional.ofNullable(recordFile).map(RecordFile::getConsensusEnd);
```
steven-sheehy, line 157: nit

```suggestion
        return timestamp.or(() ->
                Optional.ofNullable(recordFile).map(RecordFile::getConsensusEnd));
```

bilyana-gospodinova: Comment addressed, thanks.

steven-sheehy (approved):

sdimitrov9 (approved):

PR #10252, opened 2025-01-29
Fix broken tests and document methods

sdimitrov9:
**Description**:
Fix several ContractCallServicePrecompileModificationTest broken tests and document some code.

Related to https://github.com/hashgraph/hedera-mirror-node/issues/9984

**Checklist**

### DIFF
--- a/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServicePrecompileModificationTest.java
+++ b/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServicePrecompileModificationTest.java
@@ -1082,18 +1082,29 @@ class ContractCallServicePrecompileModificationTest extends AbstractContractCall
     @CsvSource({"single", "multiple"})
     void transferToken(final String type) throws Exception {
         // Given
-        final var contract = testWeb3jService.deploy(ModificationPrecompileTestContract::deploy);
-        final var tokenEntity = persistFungibleToken();
+        final var tokenEntity = tokenEntityPersist();
+        final var treasuryAccount = accountEntityPersist();
+        fungibleTokenPersist(tokenEntity, treasuryAccount);
+
         final var sender = accountEntityWithEvmAddressPersist();
         final var receiver = accountEntityWithEvmAddressPersist();
-        final var payer = accountEntityWithEvmAddressPersist();
 
-        tokenAccountPersist(tokenEntity, payer);
+        // Create token-account associations so sender and receiver can operate with the token
         tokenAccountPersist(tokenEntity, sender);
         tokenAccountPersist(tokenEntity, receiver);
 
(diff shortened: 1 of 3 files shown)


steven-sheehy (approved):

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServicePrecompileModificationTest.java
```diff
@@ -1082,18 +1087,29 @@ void createNftWithInvalidAutoRenewAccountFails() {
...
         final var sender = accountEntityWithEvmAddressPersist();
         final var receiver = accountEntityWithEvmAddressPersist();
-        final var payer = accountEntityWithEvmAddressPersist();
 
-        tokenAccountPersist(tokenEntity, payer);
+        // Create token-account associations so sender and receiver can operate with the token
         tokenAccountPersist(tokenEntity, sender);
         tokenAccountPersist(tokenEntity, receiver);
 
+        accountBalanceRecordsPersist(sender.toEntityId(), sender.getCreatedTimestamp(), sender.getBalance());
```
nickeynikolovv, line 1101: Can't we use directly treasury account as a sender? That way we will probably not need to associate and avoid using additional sender account

sdimitrov9: I guess we could, but I would prefer to use different entities for different operations, in the sense of single responsibility. In my opinion this makes the test easier to read and we will avoid the moment of someone thinking "why the sender is used for something different than transferring tokens(as a treasury account)". The performance gain for using the same entity would be negligible and given this is a test and not a production code that is highly important to run as fast as possible, I would prefer to not apply the proposed change.

nickeynikolovv (dismissed):

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServicePrecompileModificationTest.java
```diff
@@ -440,19 +440,24 @@ void burnNFT() throws Exception {
...
     }
 
+    /**
+     * Wiping tokens is the process of removing tokens from an account's balance. The total supply is not affected.
+     */
     @Test
     void wipeFungibleToken() throws Exception {
         // Given
         final var owner = accountEntityWithEvmAddressPersist();
+        final var treasuryAccount = accountEntityPersist();
```
bilyana-gospodinova, line 450: Do we need a specific treasury account or we can use the generated in the domain builder?

sdimitrov9: We can use the already existing one, fixed.

bilyana-gospodinova (approved):
LGTM

steven-sheehy (approved):

Issue #10087, opened 2025-01-10
Fix `ContractCallServiceTest` test suite


nickeynikolovv:
Fix failing tests for `ContractCallServiceTest` using the modularized EVM

PR #10253, opened 2025-01-29
Fix `INSUFFICIENT_GAS` issue in ContractCallServiceTests

nickeynikolovv:
**Description**:
This PR fixes 7 tests that were failing with INSUFFICIENT_GAS in ContractCallServiceTests by Increasing the gasLimit value in the method source for some of the tests from 30_000 to 34_000

**Related issue(s)**:
Related to  #10087 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/utils/BinaryGasEstimator.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/utils/BinaryGasEstimator.java
@@ -51,17 +51,10 @@ public class BinaryGasEstimator {
             contractCallContext.reset();
 
             long mid = (hi + lo) / 2;
-            HederaEvmTransactionProcessingResult transactionResult = null;
-
-            try {
-                transactionResult = call.apply(mid);
-            } catch (Exception ignore) {
-            }
+            HederaEvmTransactionProcessingResult transactionResult = call.apply(mid);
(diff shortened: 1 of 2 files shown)


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/utils/BinaryGasEstimator.java
```diff
@@ -51,10 +51,17 @@ public long search(
             contractCallContext.reset();
 
             long mid = (hi + lo) / 2;
-            HederaEvmTransactionProcessingResult transactionResult = call.apply(mid);
+            HederaEvmTransactionProcessingResult transactionResult = null;
+
+            try {
+                transactionResult = call.apply(mid);
+            } catch (Exception ignore) {
```
sdimitrov9, line 58: Sonar is probably going to complain for the empty catch clause. Anyway, it would probably be better to have at least a comment within the block with some reasoning behind the empty block.

nickeynikolovv: Reverted this change, because will think of a prettier way to fix that

steven-sheehy (approved):

sdimitrov9 (approved):

kselveliev (approved):

PR #10258, opened 2025-01-31
Handle exception throwing in binary gas estimator when performing estimate gas

nickeynikolovv:
**Description**:
This PR adds an exception handling during estimate gas using transaction executor. Using the modularized services when performing a contract call in binary gas estimator with the mid value and if the call fails it throws an exception instead of result with status failed. That results in a failing in the test. 

- Added a method that performs a safe contract call with try-catch mechanism
- Added additional check for 'err' if the contract result is null
- added logic to differentiate between the two flows for modularizedServices

This PR also fixes 1 test in contractCallSErviceTest

**Related issue(s)**:

Related to #10087 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/utils/BinaryGasEstimator.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/utils/BinaryGasEstimator.java
@@ -51,17 +51,10 @@ public class BinaryGasEstimator {
             contractCallContext.reset();
 
             long mid = (hi + lo) / 2;
-            HederaEvmTransactionProcessingResult transactionResult = null;
-
-            try {
-                transactionResult = call.apply(mid);
-            } catch (Exception ignore) {
-            }
+            HederaEvmTransactionProcessingResult transactionResult = call.apply(mid);
             iterationsMade++;
 
-            boolean err = transactionResult == null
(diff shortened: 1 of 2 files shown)


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/utils/BinaryGasEstimator.java
```diff
@@ -71,4 +79,17 @@ public long search(
         metricUpdater.accept(totalGasUsed, iterationsMade);
         return hi;
     }
+
+    // This method is needed because within the modularized services if the contract call fails an exceptio is thrown
```
kselveliev, line 83: exceptio -> **exception**

bilyana-gospodinova (commented):
Looks good, but the coverage needs to be increased in order for the checks to pass, similar to `ContractCallServiceTest#pureCallModularizedServices`.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/utils/BinaryGasEstimator.java
nickeynikolovv: fixed

bilyana-gospodinova (approved):
LGTM

steven-sheehy (approved):

PR #10260, opened 2025-02-01
Fix node data migration

### LINKED ISSUE #10158: Fix node data
### Description

https://github.com/hashgraph/hedera-mirror-node/pull/10134 fixed the node update transaction handling but didn't correct any existing data. While Hedera mirror nodes have been updated on mainnet before the bug could take place, we should provide a migration so the data is correct in testnet and for our partners who couldn't upgrade in time.

### Steps to reproduce

Run importer and process genesis NodeUpdate transactions

### Additional context

_No response_

### Hedera network

testnet

### Version

v0.121.0

### Operating system

None

jnels124:
**Description**:
* create versioned java migration to fix data from node transactions

**Related issue(s)**:

Fixes #10158 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/FixNodeTransactionsMigration.java
+++ b/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/FixNodeTransactionsMigration.java
@@ -185,7 +185,7 @@ public class FixNodeTransactionsMigration extends ConfigurableJavaMigration {
 
     @Override
     public MigrationVersion getVersion() {
-        return v2 ? MigrationVersion.fromVersion("2.8.0") : MigrationVersion.fromVersion("1.103.0");
+        return v2 ? MigrationVersion.fromVersion("2.8.1") : MigrationVersion.fromVersion("1.103.1");
     }
 
     @Override
--- a/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/FixNodeTransactionsMigration.java
+++ b/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/FixNodeTransactionsMigration.java
@@ -45,7 +45,7 @@ import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
 @Named
 public class FixNodeTransactionsMigration extends ConfigurableJavaMigration {
     // Earliest consensus timestamp to consider
-    private static final long RECORD_FILE_LOWER_BOUND = 1733961600000000000L;
+    private static final long LOWER_TIMESTAMP = 1733961600000000000L;
 
     private static final String DROP_DATA_SQL =
             """
@@ -167,7 +167,7 @@ public class FixNodeTransactionsMigration extends ConfigurableJavaMigration {
         return jdbcTemplate
                 .query(
                         NODE_TRANSACTIONS_SQL,
-                        Map.of("startTimestamp", RECORD_FILE_LOWER_BOUND),
+                        Map.of("startTimestamp", LOWER_TIMESTAMP),
                         new DataClassRowMapper<>(Transaction.class))
                 .stream()
                 .map(this::toRecordItem)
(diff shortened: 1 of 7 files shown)


hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/FixNodeTransactionsMigration.java
```diff
@@ -0,0 +1,221 @@
...
+    private static final String DROP_DATA_SQL =
+            """
+            truncate node;
+            truncate node_history;
+            """;
+
+    private static final String BOUNDS_SQL =
+            """
+            select consensus_start
+            from record_file
```
jnels124, line 67: Couldn't we just do `select lower(timestamp_range) from node union all select lower(timestamp_range) node_history` ?

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/parser/record/transactionhandler/NodeCreateTransactionHandler.java
```diff
@@ -44,27 +42,19 @@ public TransactionType getType() {
...
-        if (!entityProperties.getPersist().isNodes()) {
-            return;
-        }
-        transaction.setTransactionBytes(recordItem.getTransaction().toByteArray());
-        transaction.setTransactionRecordBytes(recordItem.getTransactionRecord().toByteArray());
-        parseNode(recordItem);
-    }
-
-    private void parseNode(RecordItem recordItem) {
+    public Node parseNode(RecordItem recordItem) {
```
xin-hedera, line 45: The other approach to fix the data without utilizing the transaction handlers thus can revert these changes are

- get all node rows from `node` and `node_history` tables
- grab all transactions using lower timestamp range
- iterate them in timestamp order
  - for node create, get node id from record, and update the `node` object
  - for node update / delete, get correct node id from transaction body
  - maintain node state and historical nodes in the meantime

xin-hedera: guess the issue is the above alternative assumed the only thing wrong is the node id, and that is a very strong assumption which may not hold. so scrape that :)

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/FixNodeTransactionsMigration.java
steven-sheehy: Definitely don't need to query record_file. But I don't think we need to even union node/node_history. Can simply `select * from transaction where type in (54, 55, 56) order by consensus_timestamp asc`.

jnels124: Xin, is this still a big concern on your end? At least for our schema this executes quickly and it should execute quickly  fairly quickly in v1 as well but don't really have a good way to test that.

xin-hedera: sounds good to me

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/migration/FixNodeTransactionsMigration.java
```diff
@@ -0,0 +1,189 @@
...
+import org.springframework.core.env.Environment;
+import org.springframework.core.env.Profiles;
+import org.springframework.jdbc.core.DataClassRowMapper;
+import org.springframework.jdbc.core.ParameterizedPreparedStatementSetter;
+import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
+
+@Named
+public class FixNodeTransactionsMigration extends ConfigurableJavaMigration {
+    // Earliest consensus timestamp to consider
+    private static final long RECORD_FILE_LOWER_BOUND = 1733961600000000000L;
```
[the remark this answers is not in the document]
jnels124: Yes. This is a few days before the earliest transaction in any environment.

steven-sheehy (approved):
LGTM

xin-hedera (approved):
LGTM

xin-hedera (approved):
LGTM

PR #10293, opened 2025-02-04
Refactor web3 tests code

### LINKED ISSUE #10255: Refactor code in web3 tests
### Problem

A lot of tests have similar individual setup that can be extracted in a parent class and thus making the code easier to read and maintain. Methods like persistFungibleToken and fungibleTokenPersist exist in child and parent classes that are duplicated in different suits and mistake in one needs to be corrected in all. This will remove the need to fix broken tests individually

### Solution

- Add general "setup" methods in AbstractContractCallServiceTest that child classes can use to setup the test (accountPersist, accountPersistWithAlias, fungibleToken, etc)
-  Add customizable methods (that accept Supplier) in AbstractContractCallServiceTest that child classes can use when there is a small difference in setup of a test
If this can be done with 1 method that serves both purposes - best.

### Alternatives

_No response_

filev94:
**Description**:

Removed similar code from several tests and put it in parent class so that if a change is needed it should be done in one place only. This is the first PR, more might be needed if i find more code that needs to be extracted in a similar way.

**Related issue(s)**: 

Fixes #10255

**Notes for reviewer**:

This pr fixes setup of ContractCallDynamicCallsTest - as a result there are 20+ test fixes in this class

**Checklist**


hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallDynamicCallsTest.java
```diff
@@ -822,27 +806,16 @@ private Token nftPersist(
     }
 
     private EntityId accountPersist() {
-        return domainBuilder
-                .entity()
-                .customize(e -> e.evmAddress(null))
-                .persist()
-                .toEntityId();
+        return accountEntityPersistCustomizable(e -> e.evmAddress(null)).toEntityId();
```
bilyana-gospodinova, line 809: This test is an example why we needed this refactoring. Usually when we want both alias and evmAddress to be null or non-null or with the reusable services integration this will probably fail. We can use directly `accountEntityPersist` method from the parent.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallDynamicCallsTest.java
```diff
@@ -822,27 +806,16 @@ private Token nftPersist(
...
-        return domainBuilder
-                .entity()
-                .customize(e -> e.evmAddress(alias.toArray()).alias(publicKey.toByteArray()))
-                .persist()
-                .toEntityId();
-    }
 
     private EntityId senderEntityPersistWithAlias() {
-        return accountPersistWithAlias(SENDER_ALIAS, SENDER_PUBLIC_KEY);
+        return accountPersistWithAlias(SENDER_ALIAS, SENDER_PUBLIC_KEY).toEntityId();
```
bilyana-gospodinova, line 814: In this or a future PR these `SENDER_ALIAS` and `SENDER_PUBLIC_KEY` need to be removed as well and we need to use generated values for them.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallNestedCallsTest.java
```diff
@@ -84,7 +84,7 @@ class ContractCallNestedCallsTest extends AbstractContractCallServiceOpcodeTrace
     void updateTokenKeysAndGetUpdatedTokenKeyForFungibleToken(final KeyValueType keyValueType, final KeyType keyType)
             throws Exception {
         // Given
-        final var tokenEntityId = fungibleTokenPersist();
+        final var tokenEntityId = fungibleTokenPersistWithTreasuryAccount(domainBuilder.entity().persist());
```
bilyana-gospodinova, line 87: If the treasury account is generated, can't we use the base method were we don't pass any treasury account?

filev94: Yes, you are correct. Fixing it

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallDynamicCallsTest.java
filev94: Changed from customizable method to accountEntityPersist

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallDynamicCallsTest.java
filev94: Thanks for the feedback. I think this can be fixed in a future PR. Mentioned it in the [issue](https://github.com/hashgraph/hedera-mirror-node/issues/10255#issuecomment-2634276686)

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
```diff
@@ -227,6 +228,62 @@ protected Entity tokenEntityPersist() {
...
+    protected Token fungibleTokenPersist() {
+        return fungibleTokenCustomizable(t -> {});
+    }
+
+    /**
+     *
+     * @param treasuryEntity - the treasuryEntity that has to be set in the token
+     * @return Token object that is persisted in db
+     */
+    protected Token fungibleTokenPersistWithTreasuryAccount(final Entity treasuryEntity) {
```
bilyana-gospodinova, line 244: Do we really need both `fungibleTokenPersistWithTreasuryAccount(final Entity treasuryEntity)` and `fungibleTokenPersistWithTreasuryAccount(final EntityId treasuryEntityId)`? Can't we change the code to use only 1 of them?

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
```diff
@@ -307,19 +364,47 @@ protected NftAllowance nftAllowancePersist(Token token, Entity owner, Entity spe
...
+     * @return Entity object that is persisted in the db
+     */
     protected Entity accountEntityPersist() {
-        return domainBuilder
-                .entity()
-                .customize(e ->
-                        e.type(EntityType.ACCOUNT).evmAddress(null).alias(null).balance(100_000_000_000_000_000L))
-                .persist();
+        return accountEntityPersistCustomizable(
+                e -> e.type(EntityType.ACCOUNT).evmAddress(null).alias(null).balance(100_000_000_000_000_000L));
```
bilyana-gospodinova, line 373: Is there a reason the balance in this method and in the other one (`accountEntityWithEvmAddressPersist()`) to differ?  If not, we can choose one of the values and extract it into a constant.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServiceERCTokenModificationFunctionsTest.java
```diff
@@ -607,9 +612,10 @@ void transferFromNFTRedirect() {
     @Test
     void transferWithAliasRedirect() {
         // Given
-        final var recipient = accountPersistWithAlias(SPENDER_ALIAS, SPENDER_PUBLIC_KEY);
+        final var recipient =
+                accountPersistWithAlias(SPENDER_ALIAS, SPENDER_PUBLIC_KEY).toEntityId();
         final var treasury = accountPersistWithAlias(SENDER_ALIAS, SENDER_PUBLIC_KEY);
-        final var token = fungibleTokenPersist(treasury);
+        final var token = fungibleTokenPersistWithTreasuryAccount(treasury);
```
bilyana-gospodinova, line 618: Do we really need to create and set a specific account everywhere? If we don't use it in the test later, we can just leave it to the generated one.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/OpcodeServiceTest.java
```diff
@@ -887,7 +870,7 @@ private Token nftPersist(final Entity treasuryEntity) {
     }
 
     private Entity accountPersist() {
-        return domainBuilder.entity().customize(a -> a.evmAddress(null)).persist();
+        return accountEntityPersistCustomizable(e -> e.evmAddress(null));
```
bilyana-gospodinova, line 873: The same is valid as in https://github.com/hashgraph/hedera-mirror-node/pull/10293#discussion_r1941294169

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
filev94: Yes, we can change the code. My idea was to use the Entity as a parameter for the method so that you can extend the functionality in the future, if needed. However, there are tests that use the EntityId and to refactor it back to Entity would be a big change, so that's why i left it with both for now. In the end, i can remove the Entity parameter and just leave the EntityId, that way i should only refactor tests in one class (TokenModificationFunctionsTest)

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
filev94: Good catch. I tried using one balance for both, the smaller value - the tests passed so I will extract it in a constant.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServiceERCTokenModificationFunctionsTest.java
filev94: You are correct. When i removed the treasuryAccount and used the fungibleTokenPersist(), the tests passed. And this the need for the fungibleTokenPersistWithTreasuryAccount(Entity ...) method was removed. Two birds with one stone i guess.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
filev94: This is removed as per this comment's resolution[#10293 comment](https://github.com/hashgraph/hedera-mirror-node/pull/10293#discussion_r1942541069)

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/OpcodeServiceTest.java
filev94: Fixed

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
```diff
@@ -72,6 +73,7 @@
 public abstract class AbstractContractCallServiceTest extends Web3IntegrationTest {
 
     protected static final String TREASURY_ADDRESS = EvmTokenUtils.toAddress(2).toHexString();
+    public static final long BALANCE = 1_000_000_000_000_000L;
```
bilyana-gospodinova, line 76: This is a generic abstract class, so we probably need a more specific name for this constant, such as DEFAULT_ACCOUNT_BALANCE or something like that.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/OpcodeServiceTest.java
```diff
@@ -887,7 +870,7 @@ private Token nftPersist(final Entity treasuryEntity) {
     }
 
     private Entity accountPersist() {
-        return domainBuilder.entity().customize(a -> a.evmAddress(null)).persist();
+        return accountEntityPersist();
```
bilyana-gospodinova, line 873: This method just calls and returns another method. Can be removed. Just use the parent method.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServiceERCTokenModificationFunctionsTest.java
filev94: Unfortunately tests with modularizedServices flag = true do not pass when i remove the WithTreasuryAccount, so i've reverted back that change.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
filev94: DEFAULT_ACCOUNT_BALANCE sounds better. Fixed. Also changed to the greater value since one of the tests didn't pass with modularizedFlag = true

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/OpcodeServiceTest.java
filev94: Removed accountPersist() method and used directly accountEntityPersist()

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServiceERCTokenReadOnlyFunctionsTest.java
```diff
@@ -1091,54 +1090,15 @@ void tokenURINegativeRedirect() {
...
-
-    private Token fungibleTokenPersist() {
-        final var tokenEntity =
-                domainBuilder.entity().customize(e -> e.type(TOKEN)).persist();
-
-        return domainBuilder
-                .token()
-                .customize(t -> t.tokenId(tokenEntity.getId()).type(TokenTypeEnum.FUNGIBLE_COMMON))
-                .persist();
+        return accountEntityPersistWithBalance(12L).toEntityId();
```
bilyana-gospodinova, line 1101: I wonder if we can make the tests work with the default balance or a randomly generated balance (from the domain builder) to avoid such harcoded values.

filev94: Yes, the tests passed with default balance as well. Changed to that.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
```diff
@@ -73,6 +74,7 @@
 public abstract class AbstractContractCallServiceTest extends Web3IntegrationTest {
 
     protected static final String TREASURY_ADDRESS = EvmTokenUtils.toAddress(2).toHexString();
+    public static final long DEFAULT_ACCOUNT_BALANCE = 100_000_000_000_000_000L;
```
bilyana-gospodinova, line 77: Doesn't need to be public - we want to use it only in the tests that inherit it.

filev94: Changed to protected

bilyana-gospodinova:
There is a `persistFungibleToken()` method in the `ContractCallServicePrecompileModificationTest` as well. Please address this an another PR, as well as the rest of the methods that need refactoring, such as: `tokenAccountPersist()`, `persistTokenBalance()`, etc.

filev94:
persistFungibleToken() refactoring along with the methods used by the return object will be addressed in a future PR. I've mentioned it in the [issue](https://github.com/hashgraph/hedera-mirror-node/issues/10255#issuecomment-2639745988)

bilyana-gospodinova (approved):

steven-sheehy (approved):

Issue #10374, opened 2025-02-13
Adapt acceptance tests for modularized services


kselveliev:
### Problem

Currently running acceptance tests against modularized web3 causes call to web3 module two fail.
For example `I call estimateGas with associate function for fungible token`.
This task may be split in more issue after further investigation.

### Solution

Find the related issues and adapt web3 calls in the acceptance tests for modularized services

PR #10375, opened 2025-02-13
Fix unboxing issue causing errors in acceptance tests

kselveliev:
**Description**:
After running acceptance tests against modularized web3 most of the web3 calls were failing. 
This was caused by a subtle bug caused by unboxing of null Long in `toAccountId` method called by `AccountReadableKVState`. `EntityId.of(id)` fails with nullPointer when id is null. 
Now just returning null if the id is null.
Reproduced with` Then I call estimateGas with associate function for fungible token` acceptance test.

* `ContractCallServicePrecompileModificationTest` - adds test to verify calls are passing null auto renew account now.
* `EntityIdUtils` - `toAccountId(final Long)` adds null check

**Related issue(s)**:

Partially completes #10374

**Notes for reviewer**:
Acceptance tests are now more stable but still need additional adjustments to make all pass.

**Checklist**

### DIFF
--- a/hedera-mirror-web3/src/main/java/com/hedera/services/utils/EntityIdUtils.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/services/utils/EntityIdUtils.java
@@ -162,6 +162,9 @@ public final class EntityIdUtils {
     }
 
     public static com.hedera.hapi.node.base.AccountID toAccountId(final Long id) {
+        if (id == null) {
+            return null;
+        }
         final var decodedEntityId = EntityId.of(id);
 
         return toAccountId(decodedEntityId);
--- a/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServicePrecompileModificationTest.java
+++ b/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServicePrecompileModificationTest.java
@@ -201,21 +201,17 @@ class ContractCallServicePrecompileModificationTest extends AbstractContractCall
         // Given
         final var notAssociatedAccount = accountEntityPersist();
 
-        final var tokenEntity = tokenEntityPersist();
-
-        domainBuilder
-                .token()
-                .customize(t -> t.tokenId(tokenEntity.getId()).type(TokenTypeEnum.FUNGIBLE_COMMON))
-                .persist();
-
+        final var tokenEntity = fungibleTokenPersist();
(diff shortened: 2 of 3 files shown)


hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServicePrecompileModificationTest.java
```diff
@@ -222,6 +222,37 @@ void associateToken(final Boolean single) throws Exception {
...
+        // Given
+        final var notAssociatedAccount = accountEntityPersistCustomizable(e -> e.type(EntityType.ACCOUNT)
+                .balance(DEFAULT_ACCOUNT_BALANCE)
+                .autoRenewAccountId(null)
+                .alias(null)
+                .evmAddress(null));
+
+        final var tokenEntity = tokenEntityPersist();
+
+        domainBuilder
```
bilyana-gospodinova, line 237: You can use the `fungibleTokenPersist()` method directly.

hedera-mirror-web3/src/main/java/com/hedera/services/utils/EntityIdUtils.java
```diff
@@ -162,6 +162,9 @@ public static EntityId toEntityId(final com.hedera.hapi.node.base.FileID fileID)
     }
 
     public static com.hedera.hapi.node.base.AccountID toAccountId(final Long id) {
+        if (id == null) {
```
bilyana-gospodinova, line 165: A junit test can be added for this check.

kselveliev: Already added

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServicePrecompileModificationTest.java
bilyana-gospodinova: Just one more small thing - the return type of this method is not an entity anymore. Please rename the variable as currently it is a little misleading.

bilyana-gospodinova: We have an unmerged PR that renames all such occurrences to `token`.

kselveliev: Renamed to token

Issue #10077, opened 2025-01-10
Fix `ContractCallDynamicCallsTest` test suite


nickeynikolovv:
Fix failing tests for `ContractCallDynamicCallsTest` using the modularized EVM

sdimitrov9:
Draft PR opened with 4 test fixes: https://github.com/hashgraph/hedera-mirror-node/pull/10284 ,soon to be added more.

PR #10381, opened 2025-02-14
Fix the rest `ContractCallDynamicCallsTest` tests

bilyana-gospodinova:
**Description**:
This PR fixes the setup for the `ContractCallDynamicCallsTest` tests when the `modularizedServices` flag is set to `true`. Also, the `TransactionExecutionService` is modified in case of a failure in an opcode trace scenario to return a failed result instead of throwing an exception directly. The reason for this is that we need to be able to return the opcode list from the context before throwing an exception.

Fixes https://github.com/hashgraph/hedera-mirror-node/issues/10077


steven-sheehy (approved):
LGTM

filev94 (approved):

PR #10382, opened 2025-02-14
Fix more tests in ContractCallServiceTests

nickeynikolovv:
**Description**:
This PR fixes some tests in `ContractCallServicesTests`
- made changes to capture the result from handleFailedResult but still throw an exception
- Lowered the initial balance of 0.0.2 to avoid FAILED_INVALID because it was set to maximum and wasn't able to receive reward
- Fix INSUFFICIENT_TX_FEE

Tests left to be fixed - 4
**Related issue(s)**:

Related to #10087 

**Checklist**


hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
```diff
@@ -133,6 +134,20 @@ protected void setup() {
                                 treasuryEntity.getCreatedTimestamp(), treasuryEntity.toEntityId()))
                         .balance(treasuryEntity.getBalance()))
                 .persist();
+        domainBuilder
```
bilyana-gospodinova, line 137: This and the next account is needed only only in a couple of tests in `ContractCallServiceTest`. We don't need to initialize them in every test. We can add them only in the needed tests.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
```diff
@@ -128,22 +128,33 @@ protected HederaEvmTransactionProcessingResult callContract(CallServiceParameter
...
+            result = e.getResult();
+            // In order to be able to restoreGasToBucket we have to capture the exception, then call
+            // restoreGasToBucket and finally throw the exception because there are some tests that depend on it
+            exceptionToThrow = e;
+        } finally {
+            if (restoreGasToThrottleBucket) {
+                restoreGasToBucket(result, params.getGas());
+            }
         }
+        if (exceptionToThrow != null) {
```
bilyana-gospodinova, line 153: You don't need to do this separately. A finally block is guaranteed to execute no matter if an exception is thrown.

nickeynikolovv: Fixed

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
nickeynikolovv: Fixed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/TransactionExecutionService.java
```diff
@@ -82,7 +82,8 @@ public HederaEvmTransactionProcessingResult execute(
         TransactionBody transactionBody;
         HederaEvmTransactionProcessingResult result = null;
         if (isContractCreate) {
-            transactionBody = buildContractCreateTransactionBody(params, estimatedGas, maxLifetime);
+            transactionBody = buildContractCreateTransactionBody(
+                    params, estimatedGas, maxLifetime, mirrorNodeEvmProperties.getContractCreateTxFee());
```
bilyana-gospodinova, line 86: The 4th parameter is always the same value, so it can be used in `buildContractCreateTransactionBody()` directly.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
```diff
@@ -142,15 +142,21 @@ void cleanup() {
 
     protected long gasUsedAfterExecution(final ContractExecutionParameters serviceParameters) {
         try {
-            return contractExecutionService.callContract(serviceParameters).getGasUsed();
+            if (mirrorNodeEvmProperties.isModularizedServices()) {
```
bilyana-gospodinova, line 145: The if and the else are the same statement. Can be simplified.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
```diff
@@ -140,6 +140,9 @@ public class MirrorNodeEvmProperties implements EvmProperties {
     @Min(21_000L)
     private long maxGasLimit = 15_000_000L;
 
+    @Getter
+    private long contractCreateTxFee = 100_000_000L;
```
bilyana-gospodinova, line 144: Does this really need to be a property? I am not sure if this needs to be configurable. I think a constant in the `TransactionExecutionService` should be enough.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/TransactionExecutionService.java
```diff
@@ -145,7 +146,7 @@ private HederaEvmTransactionProcessingResult handleFailedResult(
             var detail = maybeDecodeSolidityErrorStringToReadableMessage(errorMessage);
             updateErrorGasUsedMetric(gasUsedCounter, result.gasUsed(), 1);
             if (ContractCallContext.get().getOpcodeTracerOptions() == null) {
-                throw new MirrorEvmTransactionException(status.protoName(), detail, errorMessage.toHexString());
+                throw new MirrorEvmTransactionException(status.protoName(), detail, errorMessage.toHexString(), HederaEvmTransactionProcessingResult.failed(result.gasUsed(), 0L, 0L, Optional.empty(), Optional.empty()));
```
bilyana-gospodinova, line 149: We already have the `Optional.of(errorMessage)` - might as well pass it in the constructed result. Can become useful at some point.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
```diff
@@ -116,7 +116,7 @@ protected void setup() {
                 .entity()
                 .customize(e -> e.id(2L)
                         .num(2L)
-                        .balance(5000000000000000000L)
+                        .balance(1000000000000000000L)
```
bilyana-gospodinova, line 119: We can add a comment here - to say that if a balance is set near the max value, some unexpected errors occur as the balance overflows the max value if we use the same account for the node operator. This might save us some headaches in the future to avoid the same issue.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/TransactionExecutionService.java
nickeynikolovv: fixed

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
nickeynikolovv: fixed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
nickeynikolovv: we don't need it, removed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
```diff
@@ -152,7 +156,9 @@ private void restoreGasToBucket(HederaEvmTransactionProcessingResult result, lon
         // of the gasLimit value back in the bucket.
         final var gasLimitToRestoreBaseline =
                 (long) (Math.floorDiv(gasLimit, gasUnit) * throttleProperties.getGasLimitRefundPercent() / 100f);
-        if (!result.isSuccessful() && gasLimit == result.getGasUsed()) {
+        if (result == null) {
```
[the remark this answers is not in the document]
nickeynikolovv: fixed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/TransactionExecutionService.java
nickeynikolovv: added

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
nickeynikolovv: added

bilyana-gospodinova (approved):
LGTM

steven-sheehy (approved):

PR #10383, opened 2025-02-14
Added the logic to increase the response cache maxAge for /schedule/id

### LINKED ISSUE #10321: Increase response cache max-age for `/schedules/{id}`
### Problem

`/schedules/{id}` is in among the most queried endpoints, the response cache header is set with the default 1 second max-age. This can cause a lot of unnecessary queries sent to the backend when the schedule has already finalized (either executed or expired).

In order to reduce the load, a much longer max-age (e.g., 1 hour) should be used instead.

### Solution

Set a longer max-age for `/schedules/{id}` response if any of the following is true

- `deleted` is true
- `executed_timestamp` is not null
- `executed_timestamp` is null, and one of the following is true
  - `expiration_time` is null and the wall clock time is at least 31 minutes after `consensus_timestamp`
  - `expiration_time` is not null and the wall clock time is at least 1 minute after `expiration_time`

### Alternatives

_No response_

HarshSawarkar:
**Description**:

Increase cache max-age to 1 hour for finalized schedules to reduce unnecessary backend queries.

**Related issue(s)**:

Fixes #10321 

**Notes for reviewer**:

**Checklist**


hedera-mirror-rest/schedules.js
```diff
@@ -134,7 +134,27 @@ const getScheduleById = async (req, res) => {
...
+  const schedule = rows[0];
+  res.locals[constants.responseDataLabel] = formatScheduleRow(schedule);
+
+  let maxAge = 1;
+  const now = Date.now() / 1000;
+  const executedTimestamp = utils.nsToSecNs(schedule.executed_timestamp);
+  const expirationTime = utils.nsToSecNs(schedule.expiration_time);
+  const consensusTimestamp = utils.nsToSecNs(schedule.consensus_timestamp);
+
+  if (schedule.deleted || (executedTimestamp && executedTimestamp !== '')) {
```
jnels124, line 146: `executedTimestamp` is sufficient. `''` is falsey

jnels124: This is a little easier to understand.

```bash
const hasExecuted = !!executedTimestamp;
const hasAutoExpired = !schedule.expiration_time && now >= consensusTimestamp + 1860;
const hasExpired = expirationTime && now >= expirationTime + 60;
const maxAge = hasExecuted || hasAutoExpired || hasExpired ? 3600 : 1;
```

hedera-mirror-rest/__tests__/specs/schedules/{id}/responseHeaders.json
jnels124, line 1: I believe this will break other tests as not all tests should result in the longer max age. Tests that rely on the different header value will need to be grouped into the same folder with each folder having the correct `responseHeaders.json`.

hedera-mirror-rest/__tests__/specs/schedules/{id}/executed.json
jnels124, line 1: We should add a test case for explicitly expired and one for implicitly expired.

jnels124: can look in the specs for examples of.

```
"features": {
      "fakeTime": "TIME_VALUE"
 },
```

jnels124 (changes requested):

hedera-mirror-rest/schedules.js
```diff
@@ -134,7 +134,21 @@ const getScheduleById = async (req, res) => {
...
+
+  const now = Date.now() / 1000;
+  const executedTimestamp = utils.nsToSecNs(schedule.executed_timestamp);
+  const expirationTime = utils.nsToSecNs(schedule.expiration_time);
+  const consensusTimestamp = utils.nsToSecNs(schedule.consensus_timestamp);
+
+  const hasExecuted = !!executedTimestamp || schedule.deleted;
+  const hasAutoExpired = !expirationTime && now >= consensusTimestamp + 1860;
+  const hasExpired = expirationTime && now >= expirationTime + 60;
+  const maxAge = hasExecuted || hasAutoExpired || hasExpired ? 3600 : 1;
```
xin-hedera, line 148: - pull the logic to its own function so we can unit test all possible cases
- not a fan of converting a  timestamp of type `BigInt` to a string in the format of `ssss.nnnn`, and in fact there is bug in `consensusTimestamp + 1860` since it'll end up appending string `1860` to `consensusTimestamp`. Instead, use `utils.nowInNs()`, and use the nanosecond timestamp values in `schedule`, and finally add 31 minutes in nanosecond / 60 seconds in nanosecond for the relevant arithmetic operations
- we should not explicitly set `maxAge` if the conditions do not warrant a 3600 secs max-age. logic in responseHander will then set a default max-age from configuration file. Check the logic in transactions.js, `getTransactionsByIdOrHash`, around when `getTransactionsByIdOrHashCacheControlHeader` is called

hedera-mirror-rest/schedules.js
```diff
@@ -134,7 +134,21 @@ const getScheduleById = async (req, res) => {
     throw new NotFoundError();
   }
 
+  const schedule = rows[0];
+  res.locals[constants.responseDataLabel] = formatScheduleRow(schedule);
```
xin-hedera, line 138: this is duplicate of line 150

xin-hedera (changes requested):

hedera-mirror-rest/schedules.js
HarshSawarkar: Updated. I initially overlooked that adding an integer to a string would result in concatenation. Are there any other edge cases I should cover in unit tests?

hedera-mirror-rest/schedules.js
```diff
@@ -134,7 +134,29 @@ const getScheduleById = async (req, res) => {
     throw new NotFoundError();
   }
 
-  res.locals[constants.responseDataLabel] = formatScheduleRow(rows[0]);
+  const schedule = rows[0];
+  const maxAge = calculateScheduleMaxAge(schedule);
+  if (maxAge !== undefined) {
+    res.set('Cache-Control', `public, max-age=${maxAge}`);
```
xin-hedera, line 140: cache-control header set here most likely will get overridden by the default value, in the logic in `responseHandler` since it runs at the end.

Instead, should do it the same way as in `transactions.js`, i.e.

```javascript
  res.locals[constants.responseHeadersLabel] = getTransactionsByIdOrHashCacheControlHeader(
    isTransactionHash,
    scheduled !== undefined,
    rows
  );
```

the function returns an object with customized headers and it's set to the `responseHeadersLabel` in res.locals. Then `responseHandler` will merge the customized headers (as the last one) from endpoint handler with the default headers.

hedera-mirror-rest/schedules.js
```diff
@@ -134,7 +134,29 @@ const getScheduleById = async (req, res) => {
...
+  }
+  res.locals[constants.responseDataLabel] = formatScheduleRow(schedule);
+};
+
+const calculateScheduleMaxAge = (schedule) => {
+  const ONE_BILLION = BigInt(1_000_000_000);
```
xin-hedera: the simpler form of a BigInt is `1_000_000_000n`. prefer making this a constant in `constants.js`, similar to `NANOSECONDS_PER_MILLISECOND`.

hedera-mirror-rest/schedules.js
```diff
@@ -134,7 +134,29 @@ const getScheduleById = async (req, res) => {
...
+  const nowNs = utils.nowInNs();
+  const executedTimestamp = schedule.executed_timestamp;
+  const expirationTime = schedule.expiration_time;
+  const consensusTimestamp = schedule.consensus_timestamp;
+
+  const hasExecuted = executedTimestamp !== undefined || schedule.deleted;
+  const hasAutoExpired = expirationTime === undefined && nowNs >= consensusTimestamp +  THIRTY_ONE_MINUTES * ONE_BILLION;
+  const hasExpired = expirationTime !== undefined && nowNs >= expirationTime + SIXTY_SECONDS * ONE_BILLION;
+
+  return hasExecuted || hasAutoExpired || hasExpired ? 3600 : undefined;
```
xin-hedera, line 159: make the header a constant in this module, e.g., `const LONGER_SCHEDULE_CACHE_CONTROL_HEADER = {'cache-control': 'public, max-age=3600'}`. Note we use call lower cases.

hedera-mirror-rest/schedules.js
```diff
@@ -134,7 +134,29 @@ const getScheduleById = async (req, res) => {
...
-  res.locals[constants.responseDataLabel] = formatScheduleRow(rows[0]);
+  const schedule = rows[0];
+  const maxAge = calculateScheduleMaxAge(schedule);
+  if (maxAge !== undefined) {
+    res.set('Cache-Control', `public, max-age=${maxAge}`);
+  }
+  res.locals[constants.responseDataLabel] = formatScheduleRow(schedule);
+};
+
+const calculateScheduleMaxAge = (schedule) => {
```
xin-hedera, line 145: since the logic is now in its own function, we should add unit tests to `schedules.test.js`

xin-hedera (changes requested):

hedera-mirror-rest/schedules.js
HarshSawarkar: Updated

hedera-mirror-rest/schedules.js
HarshSawarkar: Added the test.

hedera-mirror-rest/__tests__/specs/schedules/{id}/executed.json
```diff
@@ -0,0 +1,105 @@
...
+      }
+    ]
+  },
+  "url": "/api/v1/schedules/2000",
+  "responseStatus": 200,
+  "responseHeaders": {
+    "cache-control": "public, max-age=3600"
+  },
+  "responseJson": {
+    "schedules": [
```
jnels124, line 48: this shouldn't be a list response

hedera-mirror-rest/__tests__/specs/schedules/{id}/explicitly-expired.json
```diff
@@ -0,0 +1,48 @@
...
+  },
+  "responseJson": {
+    "admin_key": {
+      "_type": "ProtobufEncoded",
+      "key": "AQEB"
+    },
+    "consensus_timestamp": "1234567890.000000001",
+    "creator_account_id": "0.0.1024",
+    "deleted": false,
+    "expiration_time": "1234567890.000000001",
```
jnels124, line 39: should be `1234567890.600000001`

hedera-mirror-rest/__tests__/specs/schedules/{id}/explicitly-expired.json
```diff
@@ -0,0 +1,48 @@
...
+  },
+  "url": "/api/v1/schedules/0.0.2001",
+  "responseStatus": 200,
+  "responseHeaders": {
+    "cache-control": "public, max-age=3600"
+  },
+  "responseJson": {
+    "admin_key": {
+      "_type": "ProtobufEncoded",
+      "key": "AQEB"
```
jnels124, line 34: should be `010101`

hedera-mirror-rest/__tests__/specs/schedules/{id}/implicitly-expired.json
```diff
@@ -0,0 +1,47 @@
...
+  },
+  "url": "/api/v1/schedules/0.0.2002",
+  "responseStatus": 200,
+  "responseHeaders": {
+    "cache-control": "public, max-age=3600"
+  },
+  "responseJson": {
+    "admin_key": {
+      "_type": "ProtobufEncoded",
+      "key": "AQEB"
```
jnels124, line 33: should be `010101`

jnels124 (changes requested):
spec 'not-executed-yet.json' will need to be updated to use fake time. Also can probably delete the no-params.json spec as it doesn't add value with the new test changes.

hedera-mirror-rest/__tests__/specs/schedules/{id}/implicitly-expired.json
HarshSawarkar: Updated.

hedera-mirror-rest/schedules.js
```diff
@@ -134,7 +135,26 @@ const getScheduleById = async (req, res) => {
...
-  res.locals[constants.responseDataLabel] = formatScheduleRow(rows[0]);
+  const schedule = rows[0];
+  res.locals[constants.responseHeadersLabel] = getScheduleCacheControlHeader(schedule);
+  res.locals[constants.responseDataLabel] = formatScheduleRow(schedule);
+};
+
+const getScheduleCacheControlHeader = (schedule) => {
+  const nowNs = BigInt(utils.nowInNs());
+  const executedTimestamp = schedule.executed_timestamp;
+  const expirationTime = schedule.expiration_time !== undefined ? BigInt(schedule.expiration_time) : 0n;
```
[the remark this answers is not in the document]
HarshSawarkar: Updated.

jnels124:
@HarshSawarkar can you address the test failures shown [here](https://github.com/hiero-ledger/hiero-mirror-node/actions/runs/13457490304/job/37651771513?pr=10383)?

hedera-mirror-rest/schedules.js
```diff
@@ -134,7 +135,26 @@ const getScheduleById = async (req, res) => {
...
+
+  const hasExecuted = executedTimestamp || schedule.deleted;
+  const hasAutoExpired =
+    expirationTime === 0n && nowNs >= consensusTimestamp + constants.THIRTY_ONE_MINUTES * constants.ONE_BILLION;
+  const hasExpired = expirationTime !== 0n && nowNs >= expirationTime + constants.SIXTY_SECONDS * constants.ONE_BILLION;
+
+  if (hasExecuted || hasAutoExpired || hasExpired) {
+    return LONGER_SCHEDULE_CACHE_CONTROL_HEADER;
+  }
+  return {};
```
xin-hedera, line 157: remove all `BigInt(...)` in the function

- `utils.NowInNs()` already returns bigint
- all timestamp values returned from our sql query for the schedule should be also bigints
- database `NULL` will be javascript `null`, so setting `expirationTime` / `consensusTimestamp` to default value `0n` is unnecessary and make code more complicated, and please make boolean conditions explicit against `null`, e.g., `executedTimestamp !== null`, `expirationTime === null && nowNs >= consensus_timestamp + constants...`

hedera-mirror-rest/__tests__/schedules.test.js
```diff
@@ -189,6 +189,87 @@ const verifyExtractSqlFromScheduleFilters = (filters, expectedQuery, expectedPar
...
 
+describe('getScheduleCacheControlHeader', () => {
+  const LONGER_SCHEDULE_CACHE_CONTROL_HEADER = {'cache-control': 'public, max-age=3600'};
+  const testSpecs = [
+    {
+      name: 'Schedule has executed',
+      input: {
+        executed_timestamp: utils.nowInNs() - 1_000_000_000n,
+        expiration_time: undefined,
+        consensus_timestamp: undefined,
```
xin-hedera, line 200:
```suggestion
        expiration_time: null,
        consensus_timestamp: null,
```

hedera-mirror-rest/__tests__/schedules.test.js
```diff
@@ -189,6 +189,87 @@ const verifyExtractSqlFromScheduleFilters = (filters, expectedQuery, expectedPar
...
+        deleted: false,
+      },
+      expected: LONGER_SCHEDULE_CACHE_CONTROL_HEADER,
+    },
+    {
+      name: 'Schedule is deleted',
+      input: {
+        executed_timestamp: undefined,
+        expiration_time: undefined,
+        consensus_timestamp: undefined,
```
xin-hedera, line 210: consensus_timestamp should never be null / undefined

xin-hedera (changes requested):

hedera-mirror-rest/__tests__/schedules.test.js
HarshSawarkar: Updated

hedera-mirror-rest/__tests__/schedules.test.js
```diff
@@ -175,6 +175,87 @@ const verifyExtractSqlFromScheduleFilters = (filters, expectedQuery, expectedPar
   expect(limit).toStrictEqual(expectedLimit);
 };
 
+describe('getScheduleCacheControlHeader', () => {
+  const LONGER_SCHEDULE_CACHE_CONTROL_HEADER = {'cache-control': 'public, max-age=3600'};
+  const testSpecs = [
+    {
+      name: 'Schedule has executed',
+      input: {
+        executed_timestamp: '1234567890000000002',
```
xin-hedera, line 184: this is still wrong. it should be a BigInt. some columns can be null so `null` is also appropriate depending on the test case. when it's a string, the arithmetic operations in the function like `+` will again append or prepend a number to the string.
```suggestion
        executed_timestamp: 1234567890000000002n,
```

xin-hedera (changes requested):

hedera-mirror-rest/__tests__/schedules.test.js
HarshSawarkar: Updated

jnels124:
@HarshSawarkar there are some test failures [here](https://github.com/hiero-ledger/hiero-mirror-node/actions/runs/13559376702/job/37957580308?pr=10383). Can you address these?

HarshSawarkar:
> @HarshSawarkar there are some test failures [here](https://github.com/hiero-ledger/hiero-mirror-node/actions/runs/13559376702/job/37957580308?pr=10383). Can you address these?

Updated.

jnels124 (dismissed):
LGTM

hedera-mirror-rest/constants.js
```diff
@@ -1,6 +1,9 @@
 // SPDX-License-Identifier: Apache-2.0
 
 const NANOSECONDS_PER_MILLISECOND = 10n ** 6n;
+const ONE_BILLION = 1_000_000_000n;
```
steven-sheehy, line 4:
```suggestion
const NANOS_PER_SECOND = 10n ** 9n;
```

steven-sheehy (approved):
LGTM just a field could be better named

xin-hedera (dismissed):
LGTM

jnels124 (approved):

steven-sheehy (approved):

PR #10435, opened 2025-02-19
Update Mainnet-na to 0.123.0

edwin-greene:
Updates Mainnet-na to 0.123.0


edwin-greene:
DCO failed, will re open.

clusters/mainnet-na/mainnet-citus/helmrelease.yaml
```diff
@@ -14,7 +14,7 @@ spec:
         - values.yaml
         - values-prod.yaml
         - values-v2.yml
-      version: '0.122.0'
```
xin-hedera, line 17: can you remove the following for both coordinator and worker? I forgot when deploying `0.122.0`. Same for mainnet-eu

Issue #10407, opened 2025-02-17
Refactor System Account method to remove the hardcoded value of EntityId of 700


filev94:
### Problem

The hardcoded value in systemAccountEntityWithEvmAddressPersist might produce unexpected behaviour so it needs to be refactored.

### Solution

Remove the hardcoded value and generate it automatically via domainBuilder object

steven-sheehy:
Not needed

PR #10448, opened 2025-02-21
Refactor hardcoded id of system account in web3 tests

filev94:
**Description**:
This PR refactors the hardcoded id of 700 for system account and instead generates one between 1 to 750 inclusive with the exception of 2 and 98 which i found to be special ids. 

**Related issue(s)**:

Fixes [#10407](https://github.com/hiero-ledger/hiero-mirror-node/issues/10407)

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/contracts/operations/SelfDestructOperationTest.java
+++ b/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/contracts/operations/SelfDestructOperationTest.java
@@ -54,7 +54,8 @@ class SelfDestructOperationTest extends AbstractContractCallServiceTest {
 
     @Test
     void testExecuteWithInvalidOwner() {
-        final var systemAccountAddress = toAddress(700);
+        final var systemAccountId = generateSystemAccountId();
+        final var systemAccountAddress = toAddress(systemAccountId);
         final var contract = testWeb3jService.deployWithValue(SelfDestructContract::deploy, BigInteger.valueOf(1000));
         final var functionCall = contract.send_destructContract(systemAccountAddress.toUnprefixedHexString());
 
--- a/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
+++ b/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
@@ -16,6 +16,7 @@
 
 package com.hedera.mirror.web3.service;
 
+import static com.hedera.mirror.common.util.DomainUtils.toEvmAddress;
 import static com.hedera.mirror.web3.utils.ContractCallTestUtil.ESTIMATE_GAS_ERROR_MESSAGE;
 import static com.hedera.mirror.web3.utils.ContractCallTestUtil.TRANSACTION_GAS_LIMIT;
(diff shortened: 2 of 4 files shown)


hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
```diff
@@ -75,6 +78,10 @@ public abstract class AbstractContractCallServiceTest extends Web3IntegrationTes
     protected static final String TREASURY_ADDRESS = EvmTokenUtils.toAddress(2).toHexString();
     protected static final long DEFAULT_ACCOUNT_BALANCE = 100_000_000_000_000_000L;
     protected static final int DEFAULT_TOKEN_BALANCE = 100;
+    protected static final int SYSTEM_ACCOUNT_ID_LOWER_BOUND_INCLUSIVE = 1;
+    protected static final int SYSTEM_ACCOUNT_ID_HIGHER_BOUND_EXCLUSIVE = 751;
+    protected static final List<Long> EXCLUDED_SPECIAL_IDS = List.of(2L, 98L);
+    protected static final long DEFAULT_SYSTEM_ACCOUNT_BALANCE = 20000L;
```
kselveliev, line 84: We probably don't even to have default system balance, just let domain builder set its default value and see if the test passes

filev94: Yes, you are correct. The tests are passing without the balance as well. Removed the constant.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/evm/contracts/operations/SelfDestructOperationTest.java
```diff
@@ -54,7 +54,8 @@ void testSuccessfulExecute() throws Exception {
 
     @Test
     void testExecuteWithInvalidOwner() {
-        final var systemAccountAddress = toAddress(700);
+        final var systemAccountId = generateSystemAccountId();
```
kselveliev, line 57: I just checked how the domainBuilder.id() method works and it is the following. 
```
 /**
     * Generates an encoded entity id above the max reserved id 1000. Use {@link #number()} instead for a number
     * starting from 1.
     *
     * @return The generated encoded entity id.
     */
    public long id() {
        return id.incrementAndGet() + LAST_RESERVED_ID;
    }
```

Looks like the domainBuilder auto generates ids but from 1000 -> upwards, so we shouldn't really be concerned about collisions if our custom id is below 1000. I'm pretty confident thats the case. So we might even close the PR without making any changes to this since it probably isn't a problem

filev94: @bilyana-gospodinova, please share your opinion on the case - if we should continue with the change or close the PR.

bilyana-gospodinova: I double checked and I think Kris is right - the automatic ids that the domain builder generates start from 1000, so there shouldn't be a case where a collision might occur (which was the issue that we were trying to fix).

filev94:
Closing this PR as per [comment](https://github.com/hiero-ledger/hiero-mirror-node/pull/10448#discussion_r1965605305) regarding the domainBuilder ids. Change is not needed.

Issue #10430, opened 2025-02-19
Fix call.feature acceptance test against modularized


kselveliev:
Current failing tests against modularized:

```
I call function with HederaTokenService getTokenDefaultKycStatus token FUNGIBLE - get kyc status works differently with modularized see what to do with this

I call function with update and I expect return of the updated value - 400 Bad Request: "{"_status":{"messages":[{"message":"FAIL_INVALID","detail":"","data":""}]}}"

I call function that makes N times state update
I call function with nested deploy using create function - Internal server error

Then I burn FUNGIBLE token and get the total supply and balance - Contract revert executed
Then I burn NFT and get the total supply and balance - Contract revert executed
Then I pause "FUNGIBLE" token, unpause and get the status of the token - Contract revert executed
Then I pause "NFT" token, unpause and get the status of the token - Contract revert executed
Then I mint FUNGIBLE token and get the total supply and balance - Contract revert executed
I mint NFT token and get the total supply and balance - Contract revert executed
I wipe FUNGIBLE token and get the total supply and balance - Contract revert executed
I wipe NFT and get the total supply and balance - Contract revert executed
I freeze "FUNGIBLE" token, unfreeze and get status - Contract revert executed
I freeze "NFT" token, unfreeze and get status - Contract revert executed
Then I dissociate a NFT and fail transfer  - Contract revert executed
I dissociate a FUNGIBLE token and fail transfer - Contract revert executed
I approve a NFT token and transfer it - Contract revert executed
I grant and revoke KYC - Contract revert executed
I approve a FUNGIBLE token and transfer it - Contract revert executed
```

Most of the contract reverts may be caused the same issue.

Fix failing tests.

kselveliev:
After fixing local node issues the real list of failing tests in call feature shrinked to: 

I call function with HederaTokenService getTokenDefaultKycStatus token FUNGIBLE
I call function with update and I expect return of the updated value
I call function that makes N times state update
I call function with nested deploy using create function
I call function with nested deploy using create2 function
I burn NFT and get the total supply and balance
I mint FUNGIBLE token and get the total supply and balance
I approve a NFT token and transfer it
I approve a FUNGIBLE token and transfer it

PR #10480, opened 2025-02-24
Fix contract storage readable kv state slot values

kselveliev:
**Description**:
This PR fixes slot values in `ContractStorageReadableKVState` and expand the slot values to 32 bytes.

Partially completes #10430 

**Notes for reviewer**:

Tests that this pr fixes
```
Then I call function with update and I expect return of the updated value
Then I call function that makes N times state update
```

**Checklist**

### DIFF
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/Utils.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/Utils.java
@@ -24,6 +24,7 @@ import java.time.Instant;
 import java.util.Arrays;
 import lombok.CustomLog;
 import lombok.experimental.UtilityClass;
+import org.apache.tuweni.bytes.Bytes32;
 import org.hyperledger.besu.datatypes.Address;
 
 @CustomLog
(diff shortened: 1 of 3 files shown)


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/Utils.java
```diff
@@ -69,4 +70,14 @@ public static boolean isMirror(final byte[] address) {
 
         return Arrays.equals(MIRROR_PREFIX, 0, 12, address, 0, 12);
     }
+
+    public static byte[] padToBytes32(byte[] input) {
```
bilyana-gospodinova, line 74: In `DomainUtils` there is a method - `leftPadBytes(byte[] bytes, int length)`- can we use it instead?

kselveliev: Correct, didn't see we have such a method. Addressed.

bilyana-gospodinova (approved):
LGTM

steven-sheehy (approved):

PR #10481, opened 2025-02-24
Add batch transaction design

### LINKED ISSUE #10161: HIP-551: Design batch transactions
### Problem

Need to design batch transaction support in mirrornode

### Solution

Create a design doc to implement [HIP - 551](https://github.com/hashgraph/hedera-improvement-proposal/pull/1090)

### Alternatives

_No response_

jnels124:
**Description**:
This pr will add the design document for the implementation of `HIP-551`

**Related issue(s)**:

Fixes #10161

**Notes for reviewer**:

**Checklist**


docs/design/batch-transactions.md
```diff
@@ -0,0 +1,179 @@
...
+- Submit a batch transaction with inner transaction producing hollow create AND child transactions 
+
+## Non-Functional Requirements
+
+- The mirror node should be able to handle batch transactions with minimal performance impact on ingest
+- Retrieving transactions by transaction id should have no performance impact for non batch transactions and minimal performance impact for batch transactions
+
+## Open Questions
+
+- Can block streams remove duplication of inner transaction bytes?
```
jnels124, line 178: I have design details I can add if we would like to assume that they will not create an EventTransaction for each inner transaction.

docs/design/batch-transactions.md
```diff
@@ -0,0 +1,187 @@
...
+## Architecture
+
+### Database
+
+#### Transaction
+
+```postgresql
+alter table if exists transaction
+    add column if not exists batch_key          bytea default null,
+    add column if not exists inner_transactions jsonb default null;
```
steven-sheehy, line 26: Why are we storing it as a jsonb with payer + time instead of using a bigint array of consensus timestamps? Is it for sharded queries so it has the distribution column?

docs/design/batch-transactions.md
```diff
@@ -0,0 +1,187 @@
...
+    }
+}
+```
+
+Update
+
+```java
+public class EntityRecordItemListener implements RecordItemListener {
+    private Transaction buildTransaction(EntityId entityId, RecordItem recordItem) {
+        transaction.setBatchKey(body.getBatchKey());
```
steven-sheehy, line 78: Might be implementation detail, but need to check hasBatchKey() first.

docs/design/batch-transactions.md
```diff
@@ -0,0 +1,187 @@
...
+    public void addInnerTransaction(Transaction innerTransaction) {
+        if (innerTransaction.getNonce() != 0) {
+            return;
+        }
+
+        if (innerTransactions == null) {
+            innerTransactions = new ArrayList<>();
+        }
+
+        innerTransactions.add(innerTransaction);
```
steven-sheehy, line 68: The two types don't match.

docs/design/batch-transactions.md
```diff
@@ -0,0 +1,187 @@
...
+```java
+public class Transaction {
+    private byte[] batchKey;
+
+    @JsonSerialize(using = ObjectToStringSerializer.class)
+    @JdbcTypeCode(SqlTypes.JSON)
+    private List<InnerTransaction> innerTransactions;
+
+    public void addInnerTransaction(Transaction innerTransaction) {
+        if (innerTransaction.getNonce() != 0) {
```
steven-sheehy, line 60: This check looks to already be done elsewhere.

docs/design/batch-transactions.md
```diff
@@ -0,0 +1,187 @@
...
+
+Update
+
+```java
+public class SqlEntityListener implements EntityListener, RecordStreamFileListener {
+    @Override
+    public void onTransaction(Transaction transaction) throws ImporterException {
+        context.add(transaction);
+
+        if (transaction.getBatchKey() != null && transaction.getNonce() == 0) {
```
steven-sheehy, line 91: Will synthetic child transactions have their batchKey populated? Or only the inners?

docs/design/batch-transactions.md
```diff
@@ -0,0 +1,187 @@
...
+
+            while (batchParent != null && batchParent.getType() != TransactionType.BATCH) {
+                batchParent = context.get(Transaction.class, batchParent.getParentConsensus());
+            }
+
+            if (batchParent == null) {
+                throw new ImporterException("Batch parent not found for transaction: " + transaction.getConsensusTimestamp());
+            }
+
+            batchParent.addInnerTransaction(new InnerTransaction(transaction.getPayerAccountId(), transaction.getValidStartNs()));
```
steven-sheehy, line 102: Could add a `Transaction.toInnerTransaction()` in case you have to invoke this constructor in other places.

docs/design/batch-transactions.md
```diff
@@ -0,0 +1,187 @@
...
+    @Override
+    public void onTransaction(Transaction transaction) throws ImporterException {
+        context.add(transaction);
+
+        if (transaction.getBatchKey() != null && transaction.getNonce() == 0) {
+            Transaction batchParent = context.get(Transaction.class, transaction.getParentConsensus());
+
+            while (batchParent != null && batchParent.getType() != TransactionType.BATCH) {
+                batchParent = context.get(Transaction.class, batchParent.getParentConsensus());
+            }
```
steven-sheehy, line 96: Is this necessary considering the inners will always have the root as their parent and inners can't also be batch transactions?

steven-sheehy (changes requested):

docs/design/batch-transactions.md
jnels124: It is not necessary. I missed removing this when I added the nonce check

docs/design/batch-transactions.md
jnels124: It is done elsewhere but left here as well. Will remove here as the other check prevents having to traverse the hierarchy

docs/design/batch-transactions.md
jnels124: Yes because it is sharded and we also want to retrieve all transactions matching an inner transactions transaction id.

jnels124: If preferable for row sizing purposes we can do something like:

```sql
CREATE TABLE batch_transaction(batch_payer, valid_start_ns, payer_account_id) 
create_distributed_table('batch_transaction', 'batch_payer', colocate_with => 'entity')
```

Then we can join this table in the transaction by id query.

steven-sheehy (approved):
LGTM

docs/design/batch-transactions.md
```diff
@@ -0,0 +1,186 @@
...
+
+```psudo
+const getTransactionsByIdOrHash = async (req, res) => {
+  If request by transactionId
+    Get Transactions by transactionId using current logic
+    If scheduled query param is NOT true
+      For each transaction in result
+        Collect innerTransactions
+      If innerTransactions is not empty
+        Using initial requestFilters, query for all inner transactions
```
edwin-greene, line 141: Can this recurse? Such that an inner transaction itself has an inner transaction, etc?

docs/design/batch-transactions.md
```diff
@@ -0,0 +1,186 @@
...
+          "key": "84e9575e7c9c3f5c553fb4c54e5dc408e41a1d58c12388bea37d7d7365320f6f"
+        }
+      }
+    ]
+  }
+  ```
+- Update `Transactions.getTransactionsByIdOrHash` to query all `payer_account_id` and `valid_start_ns` combinations
+  contained in `transaction.inner_transactions` for transaction(s) matching the `transactionId`
+
+```psudo
```
edwin-greene, line 133: Nit, typo for pseudo, looks like the tag isn't needed for the markdown though.

docs/design/batch-transactions.md
jnels124: We currently do not support nested batches but in the event we do, we should only return the inner batch root and not follow.

edwin-greene (dismissed):
Looks good

steven-sheehy (dismissed):

docs/design/batch-transactions.md
```diff
@@ -0,0 +1,186 @@
...
+## Architecture
+
+### Database
+
+#### Transaction
+
+```postgresql
+alter table if exists transaction
+    add column if not exists batch_key          bytea default null,
+    add column if not exists inner_transactions jsonb default null;
```
xin-hedera, line 26: per the discussion during the call, the size efficient way to store the inner transaction's payer and valid start timestamp is using a two-dimension array, `bigint[][]`

a sidenote, posts on SO suggest postgresql JSONB may take more space than JSON which is text copied directly from the input...

steven-sheehy (approved):

xin-hedera (approved):
LGTM

PR #10490, opened 2025-02-25
bootstrap.sh improvements

nirbosl:
**Description**:

- Simplified process tracking by removing the `processing_files` associative array and relying on the pids array for job tracking
- Fixed an issue where all child processes would print out the same PID as theirs, which was in fact the script parent-group-ID; each process now properly prints its own job PID, also resulting in PGAPPNAME reflecting the correct PID in the DB COPY jobs.
- Extended the `validate_special_files` function with additional validation logic
  - Fixes an issue of an error thrown for special files validation when re-running the script
- Improved robustness of file validation checks
- Enhanced the `import_file` function with additional error handling and verification steps
- Improved import reliability
- Improved code organization
- Maintained the same overall script structure and core functionality
- Converted DECOMPRESS_FLAGS into an array and added proper array handling for it; This allows to resolve a shellcheck/codacy smell regarding the var not being double-quoted to prevent globbing/splitting of the value.

**Notes for reviewer**:
The changes to the script's code are due to an issue a customer ran into. I was not able to reproduce the issue they ran into (valid file being imported but post-import row-count returns incorrectly then counts again and increments by the amount of rows it counts over and over).

I have focused on potential weaknesses I've identified that could potentially result in such edge-cases, and have refactored and improved their code to be more robust and safer in order to avoid such issues in the future.

I have repeat-tested the script using a dataset that included two large table file parts of each large table, and all small tables in the following two scnearios:
- Single run import with no interruptions (good path)
- Started an import, let it complete several files successfully then stop the script and re-run it (to validate resumption logic and proper skip of already-imported files)

All repeat tests in both scenarios finished successfully.

I have then started a local mirror-importer against the bootstrapped DB and it started cleanly and resumed sync from September 18th timestamps, which is the date this data export was taken on (0.113.2).

**Checklist**

- [x] Documented (Code comments)
- [x] Tested (Actual runs of the script, described above in the Notes section.)

### DIFF
--- a/hedera-mirror-importer/src/main/resources/db/scripts/bootstrap.sh
+++ b/hedera-mirror-importer/src/main/resources/db/scripts/bootstrap.sh
@@ -634,12 +634,44 @@ import_file() {
     fi
   done
 
+  # Fallback boundary check
+  if [ $psql_status -ne 0 ]; then
+    log "Count query failed after ${retries} retries, trying fallback boundary check" "WARN"
+
+    # Only perform boundary check for large table parts
+    if [[ "$is_small_table" != "true" ]]; then
+      boundary_count=$(psql -v ON_ERROR_STOP=1 -q -Atc "SELECT COUNT(*) FROM ${table} WHERE consensus_timestamp IN ('$start_ts', '$end_ts');")
+      boundary_status=$?
+
+      if [ $boundary_status -eq 0 ]; then
+        # Check if both boundary timestamps exist (count should be 2)
+        if [ "$boundary_count" -eq 2 ]; then
+          log "Boundary check successful for ${file}, both start and end timestamps found"
+          # Set psql_status to success and use expected_count for actual_count
+          # This assumes all rows in between exist
+          psql_status=0
+          actual_count=$expected_count
+          log "Using manifest row count ${expected_count} based on boundary verification"
+        else
+          log "Boundary check found ${boundary_count}/2 boundary timestamps for ${file}" "ERROR"
+        fi
+      else
+        log "Fallback boundary check failed for ${file}" "ERROR"
+      fi
+    fi
+  fi
+
   # Check the exit status of the count query
   if [ $psql_status -ne 0 ]; then
-    log "Final count query failure for ${file} after ${retries} retries" "ERROR"
-    write_discrepancy "${file}" "COUNT_QUERY_FAILURE" "${expected_count}"
-    write_tracking_file "$filename" "FAILED_TO_IMPORT"
-    return 1
+    log "Final count query failure for ${file} after ${retries} retries, but continuing" "WARN"
+    write_discrepancy "${file}" "ROW_COUNT_QUERY_FAILURE" "${expected_count}"
+
+    # Mark the row count as unverified but don't fail the import
+    write_tracking_file "$filename" "IMPORTED" "ROW_COUNT_UNVERIFIED"
+
+    # Log an error and proceed with the import
+    log "Import process continuing despite row count verification failure for $filename" "ERROR"
+    return 0
   fi
 
   # Verify the count matches expected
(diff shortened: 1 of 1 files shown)


hedera-mirror-importer/src/main/resources/db/scripts/bootstrap.sh
```diff
@@ -588,67 +588,81 @@ import_file() {
...
+      else
+        actual_count=$(psql -v ON_ERROR_STOP=1 -q -Atc "SELECT COUNT(*) FROM ${table} WHERE consensus_timestamp BETWEEN '$start_ts' AND '$end_ts';")
+        psql_status=$?
       fi
-      write_tracking_file "$file" "FAILED_TO_IMPORT"
-      return 1
     fi
+  done
+
+  # Check the exit status of the count query
```
steven-sheehy, line 641: If the count query fails after 3 times we should fallback to it returning true after logging an error. In theory, the import should've returned failure if any rows were missed and this is just a safety check.

hedera-mirror-importer/src/main/resources/db/scripts/bootstrap.sh
```diff
@@ -588,67 +588,81 @@ import_file() {
...
+    fi
+    if [ $i -lt $retries ]; then
+      log "Count query attempt $((i+1)) failed for ${file}, retrying in ${delay}s" "WARN"
+      sleep $delay
+      delay=$((delay * 2))
+      if [[ "$is_small_table" == "true" ]]; then
+        actual_count=$(psql -v ON_ERROR_STOP=1 -q -Atc "SELECT COUNT(*) FROM ${table};")
+        psql_status=$?
+      else
+        actual_count=$(psql -v ON_ERROR_STOP=1 -q -Atc "SELECT COUNT(*) FROM ${table} WHERE consensus_timestamp BETWEEN '$start_ts' AND '$end_ts';")
```
steven-sheehy, line 635: We could also fallback to a quicker query like `where consensus_timestamp in ($start_ts, $end_ts)` and verify the first and last are present.

jnels124 (commented):
LGTM

hedera-mirror-importer/src/main/resources/db/scripts/bootstrap.sh
```diff
@@ -394,7 +398,7 @@ process_manifest() {
 
     if [[ -f "$file_path" ]]; then
       # Skip validation if file is already imported successfully
-      if grep -q "^$file_path IMPORTED" "$TRACKING_FILE" 2>/dev/null; then
+      if [[ "$(read_tracking_status "$(basename "$file_path")")" == "IMPORTED" ]]; then
```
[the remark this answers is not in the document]
nirbosl: Correct, fixed.

hedera-mirror-importer/src/main/resources/db/scripts/bootstrap.sh
```diff
@@ -477,16 +481,23 @@ validate_special_files() {
...
       validation_failed=true
-      write_tracking_file "$file" "FAILED_VALIDATION"
+      write_tracking_file "$filename" "FAILED_VALIDATION"
     else
       log "Successfully validated special file: $filename"
-      write_tracking_file "$file" "IMPORTED"
+      current_status=$(read_tracking_status "$file")
+      if [[ "$current_status" != "IMPORTED" ]]; then
+        write_tracking_file "$file" "IMPORTED"
+      fi
```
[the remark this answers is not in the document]
nirbosl: That check would have saved writes if the status is already "IMPORTED", however the cost of just writing the line over is super small.
I've removed the check for simplicity's sake.

I've also noticed an inconsistency, that function has been using $file rather than $filename.
Both var names would have worked, however $filename is more consistent with the rest of this function's calls.

hedera-mirror-importer/src/main/resources/db/scripts/bootstrap.sh
nirbosl: Added a new conditional block to handle count query failures and fallback to a boundary check for large table parts.

steven-sheehy (approved):

PR #10501, opened 2025-02-27
Add a common realm property

### LINKED ISSUE #10411: Add a realm property
### Problem

To support non-zero realms, a mirror node needs to be configured with some base realm ID to get system files, etc.

### Solution

We don't currently have the concept of common properties across modules. We should take the time to add it for this effort.
 
* Add a `CommonProperties` to `hedera-mirror-common` and ensure it's component scanned and available in all Java modules that use common (or alternatively if that doesn't work, create it in common but register as bean in each module manually)
* Add a `hedera.mirror.common.realm=0` property
* Add the same property to REST API
* Verify in record stream that the node account ID in the bucket path matches the realm
* Verify in block stream that the realm matches (TBD)

### Alternatives

_No response_

steven-sheehy:
**Description**:

Add a common realm property to all modules except REST (will be in a follow up)

* Add a `hedera.mirror.common.realm` property
* Add a `hedera.mirror.common.shard` property
* Add a `CommonConfiguration` class
* Move `spring-boot-starter-validation` dependency to common
* Remove now redundant properties:
  * `hedera.mirror.importer.shard`
  * `hedera.mirror.restJava.shard`
  * `hedera.mirror.rosetta.realm`
  * `hedera.mirror.rosetta.shard`
* Remove `SpringApplicationContext` in favor of an `AtomicReference`
* Validate realm during address book processing

**Related issue(s)**:

Fixes #10411

**Notes for reviewer**:

**Checklist**


jnels124 (approved):
LGTM

PR #10503, opened 2025-02-27
Add EntityIdSingleton for modularized

### LINKED ISSUE #10502: Implement EntityIdService singleton
### Problem

Some acceptance tests are failing and this internally happens due to missing singleton implementation of EntityIdService singleton. Wrong ids are assigned to contract created withing the function call due to not checking db.

### Solution

Implement EntityIdService.ENTITY_ID singleton

### Alternatives

_No response_

kselveliev:
**Description**:
This PR add an `EntityIdSingleton` implementation for getting a peek of the latest entity id. 

This is needed because currently without custom implementation and db reads if we make a contract call that internally has a contract create - the contract will be assigned the first id after the last system account clashing with already existing accounts.

`EntityIdSingleton` - implements get method - to retrieve the first available entityId from db
`EntityRepository` - adds query to retrieve the first available entityId after the last system account

**Related issue(s)**:
Partially completes #10430 
Fixes #10502

**Notes for reviewer**:
Fixes acceptance tests in call.feature
```
Then I call function with nested deploy using create function
Then I call function with nested deploy using create2 function
Then I approve a NFT token and transfer it 
Then I approve a FUNGIBLE token and transfer it
```
**Checklist**


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/EntityRepository.java
```diff
@@ -33,125 +33,127 @@ public interface EntityRepository extends CrudRepository<Entity, Long> {
...
+                                order by lower(timestamp_range) desc
+                                limit 1
+                            )
+                            order by timestamp_range desc
+                            limit 1
+                            """,
             nativeQuery = true)
     Optional<Entity> findActiveByIdAndTimestamp(long id, long blockTimestamp);
+
+    @Query(value = "select coalesce(max(e.id), 1001) from entity e where e.id > 1000", nativeQuery = true)
```
bilyana-gospodinova, line 157: The value 1000 would probably work in most cases but it is configurable and I think it would be safer to use the configured value - `HederaConfig#firstUserEntity`.

kselveliev: Comment addressed but had to refactor it a bit to make this work

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/singleton/EntityIdSingleton.java
```diff
@@ -0,0 +1,31 @@
...
+
+import com.hedera.hapi.node.state.common.EntityNumber;
+import com.hedera.mirror.web3.repository.EntityRepository;
+import jakarta.inject.Named;
+import lombok.RequiredArgsConstructor;
+
+@Named
+@RequiredArgsConstructor
+@SuppressWarnings("deprecation")
+public class EntityIdSingleton implements SingletonState<EntityNumber> {
```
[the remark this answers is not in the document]
kselveliev: Yes we should worry but unfortunately we cannot use long in this case due to the following reason.
At some point upstream library calls `WritableEntityIdStore.peekAtNextNumber()`
which internally does:
```
 final var oldEntityNum = entityIdState.get();
 return oldEntityNum == null ? 1 : oldEntityNum.number() + 1;
```
entityIdState is defined like this: `private final WritableSingletonState<EntityNumber> entityIdState;`

If we make it long we are getting into class cast exception and transactions using this will always fail.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/repository/EntityRepositoryTest.java
```diff
@@ -12,8 +12,9 @@
 import org.junit.jupiter.api.Test;
 
 @RequiredArgsConstructor
-class EntityRepositoryTest extends Web3IntegrationTest {
+public class EntityRepositoryTest extends Web3IntegrationTest {
```
bilyana-gospodinova, line 15: The class should remain package-private as the others. The constant can be defined in a utility test class such as `ContractCallTestUtil`.

kselveliev: Addressed

bilyana-gospodinova (approved):
LGTM

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/singleton/EntityIdSingleton.java
```diff
@@ -0,0 +1,36 @@
...
+import com.hedera.mirror.web3.evm.properties.MirrorNodeEvmProperties;
+import com.hedera.mirror.web3.repository.EntityRepository;
+import com.hedera.node.config.data.HederaConfig;
+import jakarta.inject.Named;
+import lombok.RequiredArgsConstructor;
+
+@Named
+@RequiredArgsConstructor
+@SuppressWarnings("deprecation")
+public class EntityIdSingleton implements SingletonState<EntityNumber> {
```
steven-sheehy, line 17: We should have an integration test that reproduces the inconsistent created ID problem from the acceptance tests.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/singleton/EntityIdSingleton.java
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    @Override
+    public EntityNumber get() {
+        final var firstUserEntity = mirrorNodeEvmProperties
+                .getVersionedConfiguration()
+                .getConfigData(HederaConfig.class)
+                .firstUserEntity();
+        final Long maxId = entityRepository.findMaxId(firstUserEntity);
+        final long nextId = (maxId != null && maxId >= firstUserEntity) ? maxId + 1 : firstUserEntity;
```
steven-sheehy, line 33: There's an off by one error when entity table contains less than 1000. `firstUserEntity=1001` indicates that the first entity ID should be 1001 but this would assign it as 1002. So we should only increment the max in table but not `firstUserEntity`.

filev94 (approved):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/singleton/EntityIdSingleton.java
steven-sheehy, line 33: The logic is mixed between query and singleton. Should consolidate it to one or the other. I think it's easiest to move logic to singleton since with query you have to encode `firstUserEntity` as EntityId including with realm and shard property:

```suggestion
        final var maxId = EntityId.of(entityRepository.findMaxId()); // coalesce in query with zero to avoid NPEs
        final var nextId = Math.max(maxId.getNum() + 1, firstUserEntity);
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/singleton/EntityIdSingleton.java
steven-sheehy, line 17: If two entities are created in a single contract call, will this singleton be invoked twice or does the library invoke it once and increment that cached value? This is important to check for correctness sake. Otherwise we need to cache the value within a request. Probably should have a test for this.

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/singleton/EntityIdSingleton.java
kselveliev: Just checked testing with this ->
 ```
 function deployNestedContracts() public returns (address, address) {
        MockContract newContract1 = new MockContract();
        MockContract newContract2 = new MockContract();

        return (address(newContract1), address(newContract2));
    }
```
The singleton is invoked twice and the same value is returned from both invocations which is a problem. :(

Comment (steven-sheehy) [reply hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/singleton/EntityIdSingleton.java:17]: Should check how consensus nodes handle this since they have to cache it and be able to revert.

File: hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/singleton/EntityIdSingleton.java (hunk 5 above)
Comment (steven-sheehy) [reply hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/singleton/EntityIdSingleton.java:33]: Should probably have a test where we start at 998 in db and call singleton to get next 4 times. After each next, we persist that id to the db. Then we verify we get `999, 1000, 1001, 1002` in order. This will rule out any corner cases.

File: hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/EntityRepository.java [hunk 6]
```diff
@@ -33,125 +33,127 @@ public interface EntityRepository extends CrudRepository<Entity, Long> {
...
+                                limit 1
+                            )
+                            order by timestamp_range desc
+                            limit 1
+                            """,
             nativeQuery = true)
     Optional<Entity> findActiveByIdAndTimestamp(long id, long blockTimestamp);
+
+    @Query(value = "select coalesce(max(e.id), ?1) from entity e where e.id >= ?1", nativeQuery = true)
+    Long findMaxId(Long firstUserEntity);
```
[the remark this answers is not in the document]
kselveliev: I don’t think we can use  EntityId directly in the repository layer since bot id and num are longs.
We can manually generate:
```
EntityId firstUserEntityId = EntityId.of(config.shard(), config.realm(), config.firstUserEntity());
        long firstUserEntityLongId = firstUserEntityId.getId();
```
And then query the whole Entity by the generated above id and retrieve the entity object do toEntityId from it and get the num is the maybe the best approach?.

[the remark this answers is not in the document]
steven-sheehy: I think the approach I laid out [here](https://github.com/hiero-ledger/hiero-mirror-node/pull/10503#discussion_r1973966664) is best.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/singleton/EntityIdSingleton.java
kselveliev: I've been debugging this extensively today using both integration and acceptance tests. It looks like this behavior is managed upstream, where the first response gets cached.

For example, if the latest entity ID is 1067, the first time an account creation is triggered via Address setupPendingCreation(@NonNull final MessageFrame frame), it internally calls our method to determine the next available ID. This method is invoked multiple times during the initial account creation and returns 1068 for the new account.

However, on the next invocation of setupPendingCreation(), our custom database query logic is bypassed, and instead, a cached response is used, returning 1069 as the next ID.

I think we should't apply any new caching.

I have and integration test covering this which I will push later.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/EntityRepository.java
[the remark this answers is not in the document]
kselveliev: Implemented following https://github.com/hiero-ledger/hiero-mirror-node/pull/10503#discussion_r1973966664

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/state/singleton/EntityIdSingletonTest.java
```diff
@@ -0,0 +1,69 @@
...
+    @Mock
+    private HederaConfig hederaConfig;
+
+    @BeforeEach
+    void setup() {
+        entityIdSingleton = new EntityIdSingleton(entityRepository, mirrorNodeEvmProperties);
+    }
+
+    @Test
+    void shouldReturnFirstUserEntityIdWhenMaxIdIsLessThanLastSystemAccount() {
```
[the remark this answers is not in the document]
kselveliev: Added such a test in `EntityIdSingletonIntegrationTest`

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/state/singleton/EntityIdSingletonTest.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    @Test
+    void shouldReturnNextIdWhenMaxIdIsGreaterThanLastSystemAccount() {
+        when(mirrorNodeEvmProperties.getVersionedConfiguration()).thenReturn(versionedConfiguration);
+        when(versionedConfiguration.getConfigData(HederaConfig.class)).thenReturn(hederaConfig);
+        when(hederaConfig.firstUserEntity()).thenReturn(FIRST_USER_ENTITY_ID);
+        long maxId = 2000;
+        when(entityRepository.findMaxId(FIRST_USER_ENTITY_ID)).thenReturn(maxId);
+        assertEquals(maxId + 1, entityIdSingleton.get().number());
```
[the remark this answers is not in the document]
kselveliev: Addressed

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/state/singleton/EntityIdSingletonTest.java
```diff
@@ -0,0 +1,69 @@
...
+    @BeforeEach
+    void setup() {
+        entityIdSingleton = new EntityIdSingleton(entityRepository, mirrorNodeEvmProperties);
+    }
+
+    @Test
+    void shouldReturnFirstUserEntityIdWhenMaxIdIsLessThanLastSystemAccount() {
+        when(mirrorNodeEvmProperties.getVersionedConfiguration()).thenReturn(versionedConfiguration);
+        when(versionedConfiguration.getConfigData(HederaConfig.class)).thenReturn(hederaConfig);
+        when(hederaConfig.firstUserEntity()).thenReturn(FIRST_USER_ENTITY_ID);
```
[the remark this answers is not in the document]
kselveliev: Addressed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/singleton/EntityIdSingleton.java
```diff
@@ -0,0 +1,36 @@
...
+    private final MirrorNodeEvmProperties mirrorNodeEvmProperties;
+
+    @Override
+    public String getKey() {
+        return ENTITY_ID_STATE_KEY;
+    }
+
+    @Override
+    public EntityNumber get() {
+        final var firstUserEntity = mirrorNodeEvmProperties
```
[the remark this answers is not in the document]
kselveliev: Addressed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/singleton/EntityIdSingleton.java
kselveliev: Added such test in `ContractCallNestedCallsTest`

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/singleton/EntityIdSingleton.java
kselveliev: Addressed

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/repository/EntityRepositoryTest.java
```diff
@@ -463,4 +464,16 @@ void findHistoricalEntityByEvmAddressOrAliasAndTimestampRangeEqualToBlockTimesta
                 .usingRecursiveComparison()
                 .isEqualTo(entityHistory);
     }
+
+    @Test
+    void findMaxIdOnlySystemAccountsPresent() {
+        assertThat(entityRepository.findMaxId(FIRST_USER_ENTITY_ID)).isEqualTo(FIRST_USER_ENTITY_ID);
```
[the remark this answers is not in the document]
kselveliev: Removed FIRST_USER_ENTITY_ID here after adapting code to other comments.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/singleton/EntityIdSingleton.java
kselveliev: I've been trying to write this test, but it's difficult because domainBuilder.id() is called multiple times during entity creation, and I can't seem to bypass it. Even when using entity.persist() with custom values (proxy account, obtainer, etc.), the assigned IDs aren't consecutive. For example, one entity gets ID 1003, but the next persist assigns 1008, making it hard to verify sequential ID cases.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/utils/ContractCallTestUtil.java
```diff
@@ -12,6 +12,7 @@
 @UtilityClass
 public class ContractCallTestUtil {
 
+    public static final long FIRST_USER_ENTITY_ID = 1001;
```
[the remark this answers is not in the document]
kselveliev: Seems to still have to use cases in EntityIdSingletonTest, I can extract is a constant there?

steven-sheehy (dismissed):
LGTM

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/utils/ContractCallTestUtil.java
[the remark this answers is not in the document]
steven-sheehy: It's not needed there. You can just use properties to get the default and that is more future proof anyway

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/singleton/EntityIdSingleton.java
```diff
@@ -0,0 +1,37 @@
...
+        return ENTITY_ID_STATE_KEY;
+    }
+
+    @Override
+    public EntityNumber get() {
+        final long firstUserEntity = mirrorNodeEvmProperties
+                .getVersionedConfiguration()
+                .getConfigData(HederaConfig.class)
+                .firstUserEntity();
+        final var maxId = EntityId.of(entityRepository.findMaxId());
```
steven-sheehy, line 33: I think for non-zero realm project we should filter by shard and realm as well.
```suggestion
        final var maxId = EntityId.of(entityRepository.findMaxId(commonProperties.getShard(), commonProperties.getRealm()));
```

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/state/singleton/EntityIdSingletonIntegrationTest.java
```diff
@@ -0,0 +1,30 @@
...
+import lombok.RequiredArgsConstructor;
+import org.junit.jupiter.api.Test;
+
+@RequiredArgsConstructor
+public class EntityIdSingletonIntegrationTest extends Web3IntegrationTest {
+
+    private final EntityIdSingleton entityIdSingleton;
+
+    @Test
+    void shouldReturnNextIdWithIncrementAndRealmAndShard() {
```
steven-sheehy, line 17: I don't think this has to be an integration test. It could be done in the unit test.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/state/singleton/EntityIdSingletonIntegrationTest.java
```diff
@@ -0,0 +1,30 @@
...
+        final var entity1 = domainBuilder.entity().persist();
+        final var entityNumber1 = entityIdSingleton.get();
+        final var entity2 = domainBuilder.entity().persist();
+        final var entityNumber2 = entityIdSingleton.get();
+        final var entityShardAndRealm =
+                domainBuilder.entity().customize(e -> e.shard(1L).realm(1L)).persist();
```
steven-sheehy, line 23: Move this to be created first so it verifies it's not returned for max id in zero realm. Then set commonProperties.realm=shard=1 and verify it is returned.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/singleton/EntityIdSingleton.java
kselveliev: Using common properties now, modified the query to work with shard and realm

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/state/singleton/EntityIdSingletonIntegrationTest.java
kselveliev: Added similar unit test

bilyana-gospodinova:
Looks good but we need to see why the checks are faling.

bilyana-gospodinova (approved):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/EntityRepository.java
```diff
@@ -154,4 +154,9 @@ order by lower(timestamp_range) desc
                     """,
             nativeQuery = true)
     Optional<Entity> findActiveByIdAndTimestamp(long id, long blockTimestamp);
+
+    @Query(
+            value = "select coalesce(max(e.id), 0) from entity e where e.shard = ?1 and e.realm = ?2",
+            nativeQuery = true)
```
kselveliev, line 160: Should probably be changed to:
```
select id 
from entity 
where shard = ?1 and realm = ?2 
order by id desc 
limit 1;
```

kselveliev: Adjusted the query with the above version, small code adaptations were needed in `EntityIdSingleton` after this change

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/state/singleton/EntityIdSingletonTest.java
```diff
@@ -0,0 +1,79 @@
...
+
+    @Test
+    void shouldReturnFirstUserEntityIdWhenMaxIdIsLessThanLastSystemAccount() {
+        when(entityRepository.findMaxId(0, 0)).thenReturn(900L);
+        assertThat(entityIdSingleton.get().number()).isEqualTo(FIRST_USER_ENTITY_ID);
+    }
+
+    @Test
+    void shouldReturnFirstUserEntityIdWhenMaxIdIsNull() {
+        when(entityRepository.findMaxId(0, 0)).thenReturn(0L);
```
steven-sheehy, line 42: Change this to return null like both method name implies and repository does on empty db. You'll find there's an off by one error in singleton.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/state/singleton/EntityIdSingletonIntegrationTest.java
```diff
@@ -0,0 +1,52 @@
...
+        commonProperties.setShard(0L);
+
+        final var entity2 = domainBuilder.entity().persist();
+        final var entityNumber2 = entityIdSingleton.get();
+
+        final var entity3 = domainBuilder.entity().persist();
+        final var entityNumber3 = entityIdSingleton.get();
+
+        assertThat(entityNumberBeforeConfig.number())
+                .isNotEqualTo(entityWithShardAndRealm.toEntityId().getNum() + 1);
```
steven-sheehy, line 43: Entity is special in that it also has shard, realm and num separated out as columns that do not have any encoding. So no need to convert to EntityId first. Same comment elsewhere.

```suggestion
                .isNotEqualTo(entityWithShardAndRealm.getNum() + 1);
```

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/state/singleton/EntityIdSingletonTest.java
kselveliev: Changed and adapted singleton to handle corner case.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/state/singleton/EntityIdSingletonIntegrationTest.java
kselveliev: Addressed

kselveliev:
Seems like this singleton implementation uncovered some wrong test setup causing two non fungible token tests to fails against modularized where treasury is persisted with evm address but no alias which is incorrect. Fixed those in the PR as well.

PR #10504, opened 2025-02-27
Fix `estimatePrecompile` acceptance test with modularized services

### LINKED ISSUE #10431: Fix estimatePrecompile.feature acceptance tests against modularized
Some tests in estimatePrecompile.feature are failing agains modularized code.

Here a list of the first failing test from the suite - 
```
I call estimate gas that associates FUNGIBLE token dissociates and fails token transfer
I call estimate gas that associates NFT token dissociates and fails token transfer
I call estimateGas with transferNFT function
I call estimateGas with ERC transferFrom function
I call estimateGas with ERC transfer function
I call estimateGas with associateTokens function for fungible tokens
I call estimateGas with associateTokens function for NFTs
.
.
.
```

Fix failing tests most of them might be caused by the same underlying issue.

bilyana-gospodinova:
**Description**:
This PR fixes the failing acceptance tests in the `estimatePrecompile.feature`. The original flag `hedera.mirror.web3.evm.modularizedServices` cannot be used directly in the test module, so it was temporarily duplicated in the test module as well under `hedera.mirror.test.acceptance.web3.modularizedServices`. There is currently no common place for Spring configurations for all modules, so this is why the duplication was needed. What is more this property will be needed for a couple of more sprints and then it will be deleted from both of the modules. I also didn't see the need to define a common place for Spring properties for just 1 property that will be used in 2 modules temporarily.

A new enum was defined - `ContractMethodsModularizedServices`. It has the updated actual gas used values for the currently failing tests. In the problematic tests the `modularizedServices` flag value is compared and the correct enum value is returned.

The base `HederaTokenService` contract was updated to reflect the correct return value for `redirectForToken(...) function` and the affected contracts were recompiled and replaced as bytecode.

Fixes #10431


nickeynikolovv (dismissed):
LGTM

hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/config/Web3Properties.java
```diff
@@ -21,6 +21,8 @@ public class Web3Properties {
 
     private boolean enabled = false;
 
+    private boolean modularizedServices;
+
```
kselveliev, line 25: Having this will be able to fix one of the failing tests in call.feature :) - `Then I call function with HederaTokenService getTokenDefaultKycStatus token FUNGIBLE`

kselveliev (dismissed):
Looks good to me.

steven-sheehy (approved):

steven-sheehy:
Has a spotless failure

bilyana-gospodinova:
I applied spotless again. The check now passes.

steven-sheehy (approved):

PR #10506, opened 2025-02-27
HIP-1056 Contract BlockItem Transformer

### LINKED ISSUE #10165: HIP-1056 Contract BlockItem to RecordItem transformers
### Problem

Add block item to record item transformers.

### Solution

Add transforms for the following transaction types:
ContractCall
ContractCreate
ContractDelete
ContractUpdate
EthereumTransaction

### Alternatives

_No response_

xin-hedera:
**Description**:

This PR add contract BlockItem transformers

- Add transformers for ContractCall, ContractCreate, ContractDelete, ContractUpdate, and EthereumTransaction
- Change BlockItem transformer interface to enable setting sidecar transaction records
- Change `BlockItem` to better support transaction outputs processing
- Add `StateChangeContext` and logic to parsing and populating info from state changes
- Process blockitems in a block file in reverse order to infer changes for child transactions
- Handle response code `IDENTICAL_SCHEDULE_ALREADY_CREATED` in ScheduleCreate transformer
- Sort NFT serial numbers in TokenMint transformer
- Refactor `TokenAirdrop` transformer to infer fungible token airdrop amount with best effort
- Refactor transformer test cases
- Update design doc

**Related issue(s)**:

Fixes #10165

**Notes for reviewer**:

**Checklist**


hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/StreamFile.java
```diff
@@ -45,9 +44,9 @@ default Long getIndex() {
 
     default void setIndex(Long index) {}
 
-    Collection<T> getItems();
```
xin-hedera, line 48: only with `List`, we can iterate the block items in reverse order.

hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/transaction/RecordItem.java
```diff
@@ -67,6 +68,7 @@ public class RecordItem implements StreamItem {
     private final TransactionRecord transactionRecord;
     private final int transactionType;
 
+    @EqualsAndHashCode.Exclude
```
xin-hedera, line 71: this causes `Objects.equals` to fail when two `RecordItem` objects have everything else literally the same

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/AbstractBlockItemTransformer.java
```diff
@@ -5,21 +5,27 @@
...
 import com.hederahashgraph.api.proto.java.TransactionBody;
 import com.hederahashgraph.api.proto.java.TransactionReceipt;
-import com.hederahashgraph.api.proto.java.TransactionRecord;
 import java.security.MessageDigest;
 
 abstract class AbstractBlockItemTransformer implements BlockItemTransformer {
 
     private static final MessageDigest DIGEST = createSha384Digest();
 
-    public TransactionRecord getTransactionRecord(BlockItem blockItem, TransactionBody transactionBody) {
```
xin-hedera, line 19: with contract transactions, we have to change the method to allow setting `SidecarTransactionRecord`s

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/BlockItemTransformerFactory.java
```diff
@@ -28,18 +31,32 @@ public class BlockItemTransformerFactory {
...
-        return blockItemTransformer.getTransactionRecord(blockItem, transactionBody);
+    public void transform(BlockItem blockItem, RecordItem.RecordItemBuilder builder) {
+        try {
+            var context = getContext(blockItem);
+            var transactionBody = parse(blockItem.transaction().getSignedTransactionBytes());
+            var blockItemTransformer = get(transactionBody);
+            // pass transactionBody for performance
+            blockItemTransformer.transform(blockItem, builder, context, transactionBody);
+        } finally {
+            removeContext(blockItem);
```
xin-hedera, line 42: Remove the context in a finally block so that don't accidentally keep the stateChangeContext in the map forever.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/BlockItemTransformerFactory.java
```diff
@@ -48,4 +65,8 @@ private TransactionBody parse(ByteString signedTransactionBytes) {
             throw new ProtobufException("Error parsing transaction body from signed transaction bytes", e);
         }
     }
+
+    private void removeContext(BlockItem blockItem) {
+        stateChangeContexts.remove(blockItem.consensusTimestamp());
```
xin-hedera, line 70: only parent (or itself when there's no parent) block item should have statechanges. since block items are processed in reverse order, when processing the last child by timestamp (reverse order, so it's processed first), the parent's state changes are parsed into a context, and the context is removed when processing the parent.

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/downloader/block/transformer/AbstractTransformerTest.java
```diff
@@ -0,0 +1,95 @@
...
+        expectedPreviousItems.addFirst(null);
+        assertThat(actual)
+                .usingRecursiveFieldByFieldElementComparator(RECORD_ITEMS_COMPARISON_CONFIG)
+                .containsExactlyElementsOf(expected)
+                .map(RecordItem::getPrevious)
+                .containsExactlyElementsOf(expectedPreviousItems);
+    }
+
+    protected void finalize(RecordItemBuilder.Builder<?> builder) {
+        builder.contractTransactionPredicate(null)
```
xin-hedera, line 91: clear the predicates since block item transformer should not and will not set them. Otherwise they cause `Objects.equals` to fail.

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/domain/RecordItemBuilder.java
```diff
@@ -1408,6 +1449,10 @@ public RecordItem build() {
 
             incrementer.accept(transactionBodyWrapper, transactionRecord);
             var transaction = transaction().build();
+            if (transactionRecord.getTransactionHash() == ByteString.EMPTY) {
+                transactionRecord.setTransactionHash(
```
xin-hedera, line 1453: calculate the transaction hash, instead of using a random 48-byte byte array. this removes numerous explicit transaction hash assert in the transformer test cases, and enables `assertThat(actualRecordItem).isEqualTo(expectedRecordItem)`

hedera-mirror-importer/src/test/java/com/hedera/mirror/importer/parser/record/transactionhandler/TokenMintTransactionHandlerTest.java
```diff
@@ -124,7 +125,7 @@ void updateTransactionNonFungibleMismatch() {
         assertThat(nft.getAllValues())
                 .hasSize(expectedNfts)
                 .extracting(Nft::getSerialNumber)
-                .containsExactly(1L, 2L);
+                .containsExactly(2L, 3L);
```
xin-hedera, line 128: have to adapt to the changes in `RecordItemBuilder`

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/BlockFileTransformer.java
```diff
@@ -60,25 +59,48 @@ public RecordFile transform(BlockFile blockFile) {
...
+        // - token total supply. The state changes has the final total supply of applying all changes, reverse
+        //   processing allows applying the delta of an applicable transaction to get the correct token total supply
+        //   for the preceding
+        // - newly created entities. For example, the child transactions can create many topics, and also delete many.
+        //   It's hard to figure out which transaction creates which topic since a consensus delete topic transaction
+        //   can delete either a pre-existing topic or a new topic. With the invariance that a consensus create topic
+        //   transaction reaching consensus later should always get a larger topic id, when processing child consensus
+        //   create topic transactions in reverse order, the topic created by such a transaction is always the largest
+        //   unclaimed one
+        var builders = new ArrayList<RecordItem.RecordItemBuilder>(blockItems.size());
```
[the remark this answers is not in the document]
xin-hedera: yeah. I think I can keep the array list, however first reverse process the block items, then reverse build the record items to get the correct order.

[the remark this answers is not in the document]
xin-hedera: updated

jnels124 (approved):
LGTM

edwin-greene (dismissed):
Looks good to me

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/ContractUpdateTransformer.java
```diff
@@ -0,0 +1,33 @@
...
+@Named
+final class ContractUpdateTransformer extends AbstractContractTransformer {
+
+    @Override
+    protected void doTransform(
+            BlockItem blockItem,
+            RecordItemBuilder recordItemBuilder,
+            StateChangeContext stateChangeContext,
+            TransactionBody transactionBody) {
+        if (!blockItem.successful()) {
```
steven-sheehy, line 20: Are we sure unsuccessful updates/deletes won't have contractID resolved?

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/StateChangeContext.java
```diff
@@ -0,0 +1,225 @@
...
+    private void processTopicStateChange(MapUpdateChange mapUpdate) {
+        var topic = mapUpdate.getValue().getTopicValue();
+        topicIds.add(topic.getTopicId());
+        if (topic.getSequenceNumber() != 0) {
+            topicState.put(
+                    topic.getTopicId(),
+                    TopicMessage.builder()
+                            .runningHash(DomainUtils.toBytes(topic.getRunningHash()))
+                            .sequenceNumber(topic.getSequenceNumber())
+                            .build());
```
steven-sheehy, line 222: Why waste memory converting it to our domain TopicMessage when it's not needed? Can we keep it as Topic state?

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/StateChangeContext.java
```diff
@@ -0,0 +1,225 @@
...
+
+        return Optional.of(tokenIds.removeFirst());
+    }
+
+    public Optional<TopicID> getNewTopicId() {
+        if (topicIds.isEmpty()) {
+            return Optional.empty();
+        }
+
+        return Optional.of(topicIds.removeFirst());
```
steven-sheehy, line 116: What happens if there's a topic create 2, topic submit 2, topic create 1, topic submit 1? The first create will remove 2 but the second create will get 2 again, I believe. Probably it should be a set or we should only insert on creates.

In general for all transaction types, if we always insert IDs for create, updates and deletes how do we skip the duplicate updates and deleted IDs in the getNew methods?

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/StateChangeContext.java
```diff
@@ -0,0 +1,225 @@
...
+                        // do nothing
+                    }
+                }
+            }
+        }
+
+        fileIds.sort(FILE_ID_COMPARATOR);
+        nodeIds.sort(NODE_ID_COMPARATOR);
+        tokenIds.sort(TOKEN_ID_COMPARATOR);
+        topicIds.sort(TOPIC_ID_COMPARATOR);
```
steven-sheehy, line 80: Will the logic in this class work fine in the presence of non-zero realms and shards? It seems brittle and hacky to sort instead of depend upon encounter order.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/StateChangeContext.java
```diff
@@ -0,0 +1,225 @@
...
+        }
+
+        return Optional.of(topicIds.removeFirst());
+    }
+
+    public Optional<TopicMessage> getTopicMessage(@Nonnull TopicID topicId) {
+        return Optional.ofNullable(topicState.remove(topicId)).map(topicMessage -> {
+            if (topicMessage.getSequenceNumber() > 1) {
+                // running hash is lost for any earlier message, set it to an empty bytearray
+                // since null is not allowed
```
steven-sheehy, line 123: We should still have the topic running hash in a synthetic transaction result, right?

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/StateChangeContext.java
```diff
@@ -0,0 +1,225 @@
...
+
+    private void processTokenStateChange(MapUpdateChange mapUpdate) {
+        var token = mapUpdate.getValue().getTokenValue();
+        tokenIds.add(token.getTokenId());
+        tokenTotalSupplies.put(token.getTokenId(), token.getTotalSupply());
+    }
+
+    private void processTopicStateChange(MapUpdateChange mapUpdate) {
+        var topic = mapUpdate.getValue().getTopicValue();
+        topicIds.add(topic.getTopicId());
```
steven-sheehy, line 215: We only need to add to here for creates, right? So can optimize just to put this when topic.sequenceNumber==0?

steven-sheehy (changes requested):

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/AbstractBlockItemTransformer.java
```diff
@@ -33,22 +39,47 @@ public TransactionRecord getTransactionRecord(BlockItem blockItem, TransactionBo
...
+    /**
+     * Process contract call transaction output. Contract call transaction output is special that every child
+     * transaction of a contract call transaction can have one.
+     *
+     * @param blockItem
+     * @param recordItemBuilder
+     */
+    private void processContractCallOutput(BlockItem blockItem, RecordItem.RecordItemBuilder recordItemBuilder) {
+        var outputs = blockItem.transactionOutputs();
+        if (!outputs.containsKey(TransactionCase.CONTRACT_CALL)) {
```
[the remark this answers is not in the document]
xin-hedera: that was how consensus nodes externalize child contract create many releases ago. In case of child contract create, it should have its own synthetic contract create transaction with a `CreateContract` output and no `CallContract` output.

For ethereum transaction, depends whether it calls a contract or creates a contract, consensus nodes will set  `EtherreumOutput`'s `ethereum_call_result` / `ethereum_create_result`. There shouldn't be a `CallContractOutput` for ethereum transaction at all.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/BlockItemTransformerFactory.java
```diff
@@ -5,21 +5,24 @@
...
 import java.util.Map;
 import java.util.function.Function;
 import java.util.stream.Collectors;
+import org.springframework.util.CollectionUtils;
 
 @Named
 public class BlockItemTransformerFactory {
 
     private final BlockItemTransformer defaultTransformer;
+    private final Map<Long, StateChangeContext> stateChangeContexts = new HashMap<>();
```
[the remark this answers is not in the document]
xin-hedera: Sure. Just note that needs to be a mutable class so when processing the child block item, we can parse and set the context. And when processing the parent, we can delete the context.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/TokenMintTransformer.java
```diff
@@ -3,29 +3,40 @@
...
+            StateChangeContext stateChangeContext,
+            TransactionBody transactionBody) {
         if (!blockItem.successful()) {
             return;
         }
 
-        updateTotalSupply(blockItem.stateChanges(), transactionRecordBuilder);
-
         var tokenTransferLists = blockItem.transactionResult().getTokenTransferListsList();
+        var serialNumbers = new ArrayList<Long>();
```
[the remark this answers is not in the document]
xin-hedera: Strictly speaking computation complexity is lower with one final sort. Only in special cases e.g., most items are inserted in its natural order, using TreeSet v.s. sorting once at the end have more or less the same performance.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/BlockItemTransformerFactory.java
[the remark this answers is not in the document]
steven-sheehy: Right, so can't be a record like I mentioned.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/reader/block/ProtoBlockFileReader.java
```diff
@@ -145,9 +147,10 @@ private void readEventTransactions(ReaderContext context) {
                             "Missing transaction result in block file " + context.getFilename());
                 }
 
-                var transactionOutputs = new ArrayList<TransactionOutput>();
+                var transactionOutputs = new EnumMap<TransactionCase, TransactionOutput>(TransactionCase.class);
                 while ((protoBlockItem = context.readBlockItemFor(TRANSACTION_OUTPUT)) != null) {
-                    transactionOutputs.add(protoBlockItem.getTransactionOutput());
+                    var transactionOutput = protoBlockItem.getTransactionOutput();
+                    transactionOutputs.put(transactionOutput.getTransactionCase(), transactionOutput);
```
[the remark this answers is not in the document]
xin-hedera: yup. the two contract creates are two synthetic child transactions, and both such contract create have a single `CreateContractOutput` transaction output.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/NodeCreateTransformer.java
```diff
@@ -2,36 +2,35 @@
...
-import com.hedera.hapi.block.stream.output.protoc.StateIdentifier;
 import com.hedera.mirror.common.domain.transaction.BlockItem;
+import com.hedera.mirror.common.domain.transaction.RecordItem;
 import com.hedera.mirror.common.domain.transaction.TransactionType;
 import com.hederahashgraph.api.proto.java.TransactionBody;
-import com.hederahashgraph.api.proto.java.TransactionRecord;
 import jakarta.inject.Named;
+import lombok.CustomLog;
 
+@CustomLog
```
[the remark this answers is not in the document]
xin-hedera: sure

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/ContractUpdateTransformer.java
xin-hedera: I believe so. Since it'll be bug to have a failed contract update / delete to change the contract's state. With that being said, this might be an edge case that we can't fully reconstruct the transaction receipt. However, in transaction handler `EntityIdService` should be able to do evm address mapping if needed.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/StateChangeContext.java
xin-hedera: Because `Topic` is larger and immutable, we have to recreate it in case of future HCS system contract support that submitting multiple topic messages to a topic in a contract call becomes possible.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/StateChangeContext.java
xin-hedera: for a contract all parent-child transaction group, all changes to a single topic (e.g., create, update, multiple submit messages), will have only one MapUpdate state change in the parent transaction's state changes list. That guarantees the uniqueness of the corresponding entity ids in the list.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/StateChangeContext.java
xin-hedera: the comparators only compare the `num` part. So it won't work if we have multiple realms. I don't see there's a way to make it work without adding necessary information to various transaction outputs.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/StateChangeContext.java
xin-hedera: topic running hash isn't in transaction result. we need a protobuf change. that's based on my hypothesis when looking at a consensus submit message transaction's state changes and the consensus node's `Topic` protobuf class. After all, we don't have HCS system contract support yet.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/BlockItemTransformerFactory.java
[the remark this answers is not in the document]
xin-hedera: Moved `StateChangeContext` into `BlockItem` so the factory becomes stateless.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/StateChangeContext.java
steven-sheehy: In theory, we should not have state changes for any entities that are not part of the base realm. So sorting by num might be okay for now.

steven-sheehy (approved):
LGTM

PR #10533, opened 2025-03-03
Prepare mainnet na for 0.124.1

jnels124:
* deploy 0.124.1 to mainnet na


xin-hedera (approved):

steven-sheehy (approved):

edwin-greene (approved):

steven-sheehy (dismissed):

jnels124:
Memory was previously configured to 6gb but looking at charts in grafana we haven't exceeded 2gb usage. The default of 4gb should be fine to start.

steven-sheehy:
I think we should set it back to 6gb. We have plenty of memory on the nodes.

edwin-greene (approved):

PR #10542, opened 2025-03-03
HIP-1056 Handle standalone state changes in genesis block

### LINKED ISSUE #10416: HIP-1056 Support genesis block system entity initialization
### Problem

For a new network which begins with block streams, the genesis block will externalize hundreds of system entities through standalone state changes in round 1 - without preceding event transaction / transaction result / transaction outputs.

Mirrornode needs to handle such state changes, create synthetic transactions, outputs and results if needed so  the block transformer can properly transform the info to synthetic transaction and record as in record stream files.

### Solution

check such synthetic transactions and transaction records in record stream and implement the same in block stream pipeline

### Alternatives

_No response_

xin-hedera:
**Description**:
This PR adds support to handle standalone state changes in genesis block

- Properly read and throw away standalone state changes in genesis block
- Workaround BlockHeader.first_transaction_consensus_time not set

**Related issue(s)**:

Fixes #10416 

**Notes for reviewer**:

**Checklist**


hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/reader/block/ProtoBlockFileReader.java
```diff
@@ -71,6 +72,11 @@ public BlockFile read(StreamFileData streamFileData) {
             blockFile.setHash(context.getBlockRootHashDigest().digest());
             blockFile.setSize(bytes.length);
 
+            if (blockFile.getCount() == 0) {
+                blockFile.setConsensusStart(context.getLastMetaTimestamp());
+                blockFile.setConsensusEnd(context.getLastMetaTimestamp());
```
steven-sheehy, line 77: This only sets consensus start/end for empty round. Where do we set these for non-empty rounds? The previous place they were set was removed.

steven-sheehy: Guessing it's `BlockFileBuilder.onNewTransaction()`?

xin-hedera: Yup

xin-hedera: I can remove that. that method was there since originally BlockFile was built and then immediately returned, setting consensus start / end inline seemed more natural.

Now there's already extra setters called after `builder.build()`, we should set consensus start / end explicitly in just one place.

xin-hedera: I'll move this PR to draft, and merge the contract block item transformer PR first since it'll become more convenient with the `Collection` to `List` change to utilize List.getFirst() / getLast() to set consensus start / end.

steven-sheehy (approved):
LGTM

jnels124 (approved):
LGTM

edwin-greene (approved):
Looks good

Issue #10419, opened 2025-02-18
Add tests for update token custom fees for modularized


kselveliev:
### Problem

Mono code does not support update token custom fees method but it is supported by the modularized. We need to add integration tests to cover it. For reference check https://hips.hedera.com/hip/hip-1010 and UpdateTokenFeeScheduleTest in services.

### Solution

Add new integration tests against modularized code to cover update token custom fees.

PR #10544, opened 2025-03-04
Add update token custom fee test cases

sdimitrov9:
**Description**:
Add update token custom fee test cases.Mono code does not support this functionality so the test will run only under the modularized flag.

Relate to #10419

**Checklist**

### DIFF
--- a/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceOpcodeTracerTest.java
+++ b/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceOpcodeTracerTest.java
@@ -23,13 +23,20 @@ import com.hedera.mirror.web3.evm.contracts.execution.traceability.OpcodeTracerO
 import com.hedera.mirror.web3.evm.store.accessor.EntityDatabaseAccessor;
 import com.hedera.mirror.web3.repository.EntityRepository;
 import com.hedera.mirror.web3.service.model.ContractDebugParameters;
+import com.hedera.mirror.web3.state.MirrorNodeState;
 import com.hedera.mirror.web3.utils.ContractFunctionProviderRecord;
 import com.hedera.node.app.service.evm.contracts.execution.HederaEvmTransactionProcessingResult;
 import io.micrometer.core.instrument.Counter;
 import io.micrometer.core.instrument.Meter.MeterProvider;
+import jakarta.annotation.PostConstruct;
 import jakarta.annotation.Resource;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
 import lombok.SneakyThrows;
 import org.apache.tuweni.bytes.Bytes;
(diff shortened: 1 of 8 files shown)


hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceOpcodeTracerTest.java
```diff
@@ -269,4 +276,24 @@ protected void accountBalanceRecordsPersist(EntityId accountId, Long timestamp,
     protected Entity getEntity(EntityId entityId) {
         return entityRepository.findById(entityId.getId()).get();
     }
+
+    protected void activateModularizedFlag() throws InvocationTargetException, IllegalAccessException {
```
bilyana-gospodinova, line 280: This method does more than just to activate the flag - it initializes the state. There are 4 more tests that do the same - the new methods can be invoked in these tests.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallServiceERCTokenReadOnlyFunctionsTest.java
```diff
@@ -698,21 +692,9 @@ void decimalsNegative() {
...
-
-        Method postConstructMethod = Arrays.stream(MirrorNodeState.class.getDeclaredMethods())
-                .filter(method -> method.isAnnotationPresent(PostConstruct.class))
-                .findFirst()
-                .orElseThrow(() -> new IllegalStateException("@PostConstruct method not found"));
-
-        postConstructMethod.setAccessible(true); // Make the method accessible
-        postConstructMethod.invoke(state);
+
+        activateModularizedFlagAndInitializeState();
```
bilyana-gospodinova, line 697: That's better but we should surround it with a try-catch-finally block in order to make sure we revert the original property configuration at the end of the test in case of exception. I see that this was missed in the original test code but please add it here.

sdimitrov9: Done

bilyana-gospodinova (dismissed):
Looks good but we need to see why the checks are failing.

hedera-mirror-web3/src/test/solidity/ModificationPrecompileTestContract.sol
```diff
@@ -321,6 +321,60 @@ contract ModificationPrecompileTestContract is HederaTokenService {
...
+        if (responseCode != HederaResponseCodes.SUCCESS) {
+            revert("Failed to update fungible token custom fees");
+        }
+
+        (newFixedFees, newFractionalFees, newRoyaltyFees) = getCustomFeesForToken(token);
+
+        return (newFixedFees, newFractionalFees, newRoyaltyFees);
+    }
+
+    function updateNonFungibleTokenCustomFeesExternal(address token, IHederaTokenService.FixedFee[] memory fixedFees, IHederaTokenService.RoyaltyFee[] memory royaltyFees) external
```
steven-sheehy, line 369: I assume this will be tested in a follow up? So why add to the contract in this one?

sdimitrov9: Correct, it will be added in another PR. Now it is removed.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
```diff
@@ -575,6 +600,28 @@ protected String getAddressFromEvmAddress(final byte[] evmAddress) {
...
+    protected void activateModularizedFlagAndInitializeState()
+            throws InvocationTargetException, IllegalAccessException {
+        mirrorNodeEvmProperties.setModularizedServices(true);
+
+        Method postConstructMethod = Arrays.stream(MirrorNodeState.class.getDeclaredMethods())
+                .filter(method -> method.isAnnotationPresent(PostConstruct.class))
+                .findFirst()
+                .orElseThrow(() -> new IllegalStateException("@PostConstruct method not found"));
+
+        postConstructMethod.setAccessible(true);
```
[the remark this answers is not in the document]
sdimitrov9: The flag check in the MirrorNodeState#init is necessary, otherwise the acceptance tests are going to experience failures.

steven-sheehy (approved):

PR #10554, opened 2025-03-06
Use shard and realm properties in web3 API

### LINKED ISSUE #10412: `EntityIdUtils` ignores realm and shard
### Problem

The web3 module incorrectly uses just long entity num in places instead of using shard, realm and num.

### Solution

* Change any method in `EntityIdUtils` to use shard and realm when encoding from an object to/from bytes or numerics
* Remove `BitPackUtils`
* Change `EntityNum` to internally use `EntityId`
* Investigate other areas like `AllowanceId`

### Alternatives

_No response_

jnels124:
**Description**:

**Related issue(s)**:

Fixes #10412 

**Notes for reviewer**:

**Checklist**


hedera-mirror-web3/src/main/java/com/hedera/services/utils/EntityNum.java
```diff
@@ -16,32 +17,31 @@
...
-        if (!isValidNum(l)) {
-            return MISSING_NUM;
-        }
-        final var value = codeFromNum(l);
-        return new EntityNum(value);
+        return new EntityNum(EntityId.of(l));
     }
 
     public static EntityNum fromAccountId(final AccountID grpc) {
+        // Should this remain restricted?
```
jnels124, line 44: This function returns true if shard and realm are both zero.

hedera-mirror-common/src/test/java/com/hedera/mirror/common/util/DomainUtilsTest.java
```diff
@@ -350,6 +383,24 @@ void toEvmAddressContractID() throws Exception {
...
+        var id = TokenID.newBuilder()
+                .setShardNum(entityId.getShard())
+                .setRealmNum(entityId.getRealm())
+                .setTokenNum(entityId.getNum())
+                .build();
+
+        String expected = "000001FF000000000000FFFF0000003FFFFFFFFF";
+        String expectedEmpty = "0000000000000000000000000000000000000000";
+        assertThat(DomainUtils.toEvmAddress(id)).asHexString().isEqualTo(expected);
+        assertThat(DomainUtils.toEvmAddress(AccountID.getDefaultInstance()))
```
jnels124, line 398: should be `TokenID.getDefaultInstance()`

hedera-mirror-web3/src/main/java/com/hedera/services/utils/EntityNum.java
kselveliev: I think we should probably change this to shard/realm from common properties. This is old mono logic that will be deleted after completion of modularized evm.

kselveliev (dismissed):

hedera-mirror-common/src/main/java/com/hedera/mirror/common/util/DomainUtils.java
```diff
@@ -295,6 +313,14 @@ public static byte[] toEvmAddress(EntityId contractId) {
         return toEvmAddress((int) contractId.getShard(), contractId.getRealm(), contractId.getNum());
     }
 
+    public static byte[] toEvmAddress(long id) {
+        return toEvmAddress(EntityId.of(id));
+    }
+
+    public static byte[] toEvmAddress(long shard, long realm, long num) {
+        return toEvmAddress((int) shard, realm, num);
```
bilyana-gospodinova, line 321: I have some concerns about casting the `long` directly to an `int`. It seems to me that it would be safer to check if the value would fit into an `int` first in order to prevent some runtime unexpected exceptions. For reference there is a similar check in `EntityId.encode(...)`.

jnels124: I will take another look at this. Thanks

hedera-mirror-common/src/test/java/com/hedera/mirror/common/util/DomainUtilsTest.java
```diff
@@ -332,6 +331,40 @@ void toEvmAddressEntityId() {
...
+    @Test
+    void toEvmAddressAccountId() {
+        var entityId = EntityId.of(Long.MAX_VALUE);
+        var id = AccountID.newBuilder()
+                .setShardNum(entityId.getShard())
+                .setRealmNum(entityId.getRealm())
+                .setAccountNum(entityId.getNum())
+                .build();
+
+        String expected = "000001FF000000000000FFFF0000003FFFFFFFFF";
```
filev94, line 343: You can create a constant for this variable as it is present at least 3 times

hedera-mirror-common/src/test/java/com/hedera/mirror/common/util/DomainUtilsTest.java
```diff
@@ -350,6 +383,24 @@ void toEvmAddressContractID() throws Exception {
...
+    void toEvmAddressTokenId() {
+        var entityId = EntityId.of(Long.MAX_VALUE);
+        var id = TokenID.newBuilder()
+                .setShardNum(entityId.getShard())
+                .setRealmNum(entityId.getRealm())
+                .setTokenNum(entityId.getNum())
+                .build();
+
+        String expected = "000001FF000000000000FFFF0000003FFFFFFFFF";
+        String expectedEmpty = "0000000000000000000000000000000000000000";
```
filev94, line 396: Whenever possible, try to use final var for non-primitives

hedera-mirror-common/src/test/java/com/hedera/mirror/common/util/DomainUtilsTest.java
```diff
@@ -4,10 +4,7 @@
 
 import static com.hedera.mirror.common.util.CommonUtils.nextBytes;
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.junit.jupiter.api.Assertions.assertAll;
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertNull;
-import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.*;
```
filev94, line 7: I think there is a rule not to use wildcard imports

hedera-mirror-common/src/main/java/com/hedera/mirror/common/util/DomainUtils.java
jnels124: Removed this as the only thing relying on this is ` Id.asEvmAddress` which calls a utility function in entity id utils. I have changed that function to convert the id to entityId before getting the evm address. This will result in an `InvalidEntityException` if the id has an invalid shard, realm, or num

hedera-mirror-web3/src/main/java/com/hedera/services/utils/BitPackUtils.java
```diff
@@ -1,55 +0,0 @@
...
-
-    /**
-     * Checks if the given long is not a number in the allowed range
-     *
-     * @param num given long number to check
-     * @return true if valid, else returns false
-     */
-    public static boolean isValidNum(long num) {
-        return num >= 0 && num <= MAX_NUM_ALLOWED;
-    }
```
steven-sheehy, line 54: Where do we check this now?

hedera-mirror-web3/src/main/java/com/hedera/services/utils/EntityNum.java
```diff
@@ -16,60 +18,68 @@
...
         return fromLong(grpc.getTokenNum());
     }
 
+    /**
+     * This method is unsafe to use before the spring context is fully initialized.
+     * */
     static boolean areValidNums(final long shard, final long realm) {
-        return shard == 0 && realm == 0;
+        var commonProps = CommonProperties.getInstance();
+        return shard == commonProps.getShard() && realm == commonProps.getRealm();
```
steven-sheehy, line 69: Why do we need to validate that they're from the configured shard and realm here? This would mean we can't support tables that might have entities across different realms or shards.

hedera-mirror-web3/src/main/java/com/hedera/services/utils/EntityNum.java
```diff
@@ -16,60 +18,68 @@
...
 
-    public int intValue() {
-        return value;
+    public static EntityNum fromId(Id id) {
+        return new EntityNum(EntityId.of(id.shard(), id.realm(), id.num()));
     }
 
-    public long longValue() {
-        return numFromCode(value);
+    public long getId() {
```
steven-sheehy, line 76: Don't think we should expose the encoded ID. Should update the caller to use `EntityNum` directly. It looks like the only user is `CryptoOpsUsage` which is just using the allowances map to calculate cost.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/common/Web3UnitTest.java
```diff
@@ -0,0 +1,13 @@
...
+package com.hedera.mirror.web3.common;
+
+import com.hedera.mirror.common.CommonProperties;
+import org.junit.jupiter.api.BeforeAll;
+
+public abstract class Web3UnitTest {
+    @BeforeAll
+    static void setUp() {
+        new CommonProperties().init();
+    }
```
steven-sheehy, line 12: Not a fan of this whole class and having to remember to extend unit tests from it. We can just change `CommonProperties.getInstance()` to create one if it doesn't exist and to always set it in `PostConstruct` even when not null.

Or if you feel strongly that's a bad approach, we should call init in a global, centralized place in JUnit so that it occurs only once per JVM. The way to do this looks to be [LauncherSessionListener](https://junit.org/junit5/docs/current/user-guide/#launcher-api-launcher-session-listeners-custom).

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/services/utils/EntityNum.java
```diff
@@ -16,60 +18,68 @@
...
+        return new EntityNum(DomainUtils.fromEvmAddress(address.toArrayUnsafe()));
     }
 
     public static EntityNum fromInt(final int i) {
         return new EntityNum(i);
     }
 
     public static EntityNum fromLong(final long l) {
-        if (!isValidNum(l)) {
-            return MISSING_NUM;
```
steven-sheehy, line 38: There's some logic that depends upon MISSING_NUM in TransferLogic. Not sure if this removal could cause it problems.

hedera-mirror-web3/src/main/java/com/hedera/services/utils/BitPackUtils.java
jnels124: Calls now pass through `EntityId.of()` which will validate but throw on an invalid shard,realm,num. It will return fine for 0 id but not for invalid shard, realm, or num. Should we modify the method it uses now(uses EntityNum.fromId) to return an empty id (0.0.0) for these invalid cases?

hedera-mirror-web3/src/main/java/com/hedera/services/utils/EntityNum.java
jnels124: I wasn't sure about  this change initially and the comments on this pr suggested it change to use the configured shard and realm

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/common/Web3UnitTest.java
jnels124: I agree and wanted to do something globally but wasn't sure of the approach so did this initially

hedera-mirror-web3/src/main/java/com/hedera/services/utils/EntityNum.java
jnels124: TransferLogic now calls the fromId method which could result in an invalid entity exception if the id is not valid. If it is id:0 it will return empty but for invalid values in shard,realm, or num it will throw an InvalidEntityException. Should we catch that exception on calls from here and return the empty entity num instead?

hedera-mirror-web3/src/main/java/com/hedera/services/utils/EntityNum.java
steven-sheehy: Probably doesn't matter at this point as this code will go away before we have to consider inter-realm transactions. Let's keep with validating against props.

hedera-mirror-web3/src/main/java/com/hedera/services/utils/EntityNum.java
bilyana-gospodinova: I think we should catch the exception and return the empty entity num in order not to change the flow.

steven-sheehy (approved):

PR #10557, opened 2025-03-06
Change entity id encoding

### LINKED ISSUE #10410: Change EntityId encoding
### Problem

The mirror node should be able to support hundreds of billions of entities but the current EntityId encoding limits it to 2^32 entities.

### Solution

* Increase `EntityId.num` to 38 bits
* Decrease `EntityId.shard` to 10 bits. This will require using the currently unused signed bit.

### Alternatives

_No response_

xin-hedera:
**Description**:

This PR changes entity id encoding to the following

- shard, 10 bits
- realm, 16 bits
- num, 38 bits

**Related issue(s)**:

Fixes #10410 

**Notes for reviewer**:

**Checklist**


hedera-mirror-rest/entityId.js
```diff
@@ -63,9 +63,12 @@ class EntityId {
         this.encodedId = null;
       } else {
         this.encodedId =
-          this.shard <= maxSafeShard
-            ? this.shard * shardScale + this.realm * realmScale + this.num
-            : (BigInt(this.shard) << shardOffset) | (BigInt(this.realm) << numBits) | BigInt(this.num);
+          this.shard === 0 && this.realm <= maxSafeRealm
```
steven-sheehy, line 66: Why do we compare shard against 0 here? Is this code path just an optimization to not use BigInt and bit shifting for the common zero realm/shard scenario?

xin-hedera: yes. this an optimization that with the new encoding, if shard is 0 and realm <= maxSafeRealm, we can fit the encoded id into a `Number` since the value will be <= `Number.MAX_SAFE_INTEGER`. It was there for the old encoding too.

steven-sheehy (dismissed):
LGTM

jnels124 (approved):
LGTM

steven-sheehy (approved):

PR #10562, opened 2025-03-06
Support non-zero realm in monitor

### LINKED ISSUE #10413: Monitor uses a hardcoded operator with a non-zero realm
### Problem

Monitor uses a hardcoded genesis account 0.0.2 as its operator and should be adjusted to work with non-zero realms.

### Solution

* On startup, retrieve the actual shard and realm from `/api/v1/networks?limit=1` response field `node_account_id`. If the operator ID is the default of `0.0.2`, adjust it to the retrieved `shard.realm.2`. If the operator ID is not the default value, verify the shard and realm match.
* Change obtainer id in transaction suppliers similarly

### Alternatives

_No response_

xin-hedera:
**Description**:

**Related issue(s)**:

Fixes #10413 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-monitor/build.gradle.kts
+++ b/hedera-mirror-monitor/build.gradle.kts
@@ -9,6 +9,11 @@ plugins {
 
 dependencies {
     implementation(platform("org.springframework.cloud:spring-cloud-dependencies"))
+    implementation(project(":common")) {
+        exclude("com.google.protobuf", "protobuf-java")
+        exclude("org.springframework.boot", "spring-boot-starter-data-jpa")
+        exclude("org.web3j", "core")
+    }
     implementation("com.fasterxml.jackson.core:jackson-databind")
     implementation("com.google.guava:guava")
     implementation("com.hedera.hashgraph:sdk")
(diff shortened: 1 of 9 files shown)


hedera-mirror-monitor/src/main/java/com/hedera/mirror/monitor/publish/generator/ConfigurableTransactionGenerator.java
```diff
@@ -114,15 +116,21 @@ private TransactionSupplier<?> convert() {
...
+            var validator = validatorFactory.getValidator();
+            var validations = validator.validate(supplier);
+
+            if (!validations.isEmpty()) {
+                throw new ConstraintViolationException(validations);
+            }
+
+            if (supplier instanceof AccountDeleteTransactionSupplier accountDeleteTransactionSupplier) {
+                String accountId = accountDeleteTransactionSupplier.getTransferAccountId();
+                accountDeleteTransactionSupplier.setTransferAccountId(accountIdValidator.validate(accountId));
```
steven-sheehy, line 132: I think instead of this approach we should set this using the value from `OperatorProperties` if it's the default value (or we can set it to empty for default). That provides more functionality in the case when we change the default operator to some non-default value we don't have to remember to also change it here.

Then we can get rid of the `AccountIdValidator` and only perform that logic inside OperatorProperties. It's not clear to me if HAPI disallows obtainer IDs for a different realm so probably don't need to validate it locally and can let HAPI validate it.

xin-hedera: updated

edwin-greene (approved):
Looks good to me

PR #10588, opened 2025-03-10
Adapt web3 startup with modularized for non genesis startup

### LINKED ISSUE #10555: Disable account balance verification on startup
### Problem

On web3 container startup when the DB is not empty (as it will be on an environment that is not local) the following exception occurs on state initialization:

```
Error creating bean with name 'mirrorNodeState': Invocation of init method failed
...
Caused by: java.lang.IllegalStateException: Total balance of all accounts does not match the total float: actual: 0 vs expected: 5000000000000000000
2025-03-06 14:10:21     at com.hedera.node.app.service.token.impl.schemas.V0490TokenSchema.createGenesisSchema(V0490TokenSchema.java:194)
2025-03-06 14:10:21     at com.hedera.node.app.service.token.impl.schemas.V0490TokenSchema.migrate(V0490TokenSchema.java:116)
2025-03-06 14:10:21     at com.hedera.mirror.web3.state.components.SchemaRegistryImpl.migrate(SchemaRegistryImpl.java:127)
2025-03-06 14:10:21     at com.hedera.mirror.web3.state.components.ServiceMigratorImpl.lambda$doMigrations$0(ServiceMigratorImpl.java:69)
...
```

The reason for that is that on startup all of the account balances are summed up and the value is expected to be equal to 50,000,000,000 HBARs (comes from config). Currently the account size is not returned correctly and no actual sum is performed.

### Solution

Disable the account balance verification since it is not applicable for the mirror node.

### Alternatives

_No response_

kselveliev:
**Description**:
Seems like during state initialization when we do migrations we have the options to pass previous version (`SoftwareVersion`).

Currently we are passing null as previous version which later on during the migrations is checked if null then do all genesis startup logic and validations which is pretty big operation and takes a lot of time.
This is a good approach to do on empty db like integration tests startup.

But after testing with mainnet snapshots we are getting into trouble since our db is already populated with all system accounts etc and we get into failures.

This pr assigns a previous version if db is not empty if completely empty then we should proceed with genesis logic otherwise couple of our singletons querying for record files will start failing.

Changes:
`MirrorNodeState` - add logic to set previous version during startup on non empty db.
tests..

**Related issue(s)**:

Fixes #10555

**Notes for reviewer**:
During startup with non genesis we need to set latest recordFile in context since it's being read in our custom singleton implementations during the init.

**Checklist**

### DIFF
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/MirrorNodeState.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/MirrorNodeState.java
@@ -11,7 +11,6 @@ import static java.util.Objects.requireNonNull;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.hedera.hapi.node.base.Key;
-import com.hedera.hapi.node.base.SemanticVersion;
 import com.hedera.hapi.node.base.SignatureMap;
 import com.hedera.hapi.node.transaction.ThrottleDefinitions;
 import com.hedera.mirror.common.domain.transaction.RecordFile;
@@ -25,7 +24,6 @@ import com.hedera.mirror.web3.state.core.MapReadableStates;
 import com.hedera.mirror.web3.state.core.MapWritableKVState;
 import com.hedera.mirror.web3.state.core.MapWritableStates;
 import com.hedera.mirror.web3.state.singleton.SingletonState;
-import com.hedera.node.app.config.BootstrapConfigProviderImpl;
 import com.hedera.node.app.fees.FeeService;
 import com.hedera.node.app.ids.EntityIdService;
 import com.hedera.node.app.records.BlockRecordService;
@@ -106,16 +104,11 @@ public class MirrorNodeState implements State {
         }
 
         Optional<RecordFile> latest = recordFileRepository.findLatest();
-        final var bootstrapConfig = new BootstrapConfigProviderImpl().getConfiguration();
+        final var bootstrapConfig = mirrorNodeEvmProperties.getVersionedConfiguration();
         final var currentSemanticVersion =
                 bootstrapConfig.getConfigData(VersionConfig.class).servicesVersion();
         final var currentVersion = new ServicesSoftwareVersion(currentSemanticVersion);
-        final var previousVersion = latest.isEmpty()
-                ? null
-                : new ServicesSoftwareVersion(SemanticVersion.newBuilder()
-                        .minor(currentSemanticVersion.minor() - 1)
-                        .build());
-
+        final var previousVersion = latest.isEmpty() ? null : currentVersion;
         ContractCallContext.run(ctx -> {
             latest.ifPresent(ctx::setRecordFile);
             registerServices(servicesRegistry);


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/MirrorNodeState.java
```diff
@@ -101,15 +105,25 @@ private void init() {
...
+        Optional<RecordFile> latest = recordFileRepository.findLatest();
+        final var bootstrapConfig = new BootstrapConfigProviderImpl().getConfiguration();
+        final var currentSemanticVersion =
+                bootstrapConfig.getConfigData(VersionConfig.class).servicesVersion();
+        final var currentVersion = new ServicesSoftwareVersion(currentSemanticVersion);
+        final var previousVersion = latest.isEmpty()
+                ? null
+                : new ServicesSoftwareVersion(SemanticVersion.newBuilder()
+                        .minor(currentSemanticVersion.minor() - 1)
+                        .build());
```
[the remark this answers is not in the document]
bilyana-gospodinova: Will this be an issue if the previous version results to a version that does not actually exist? Have you tested it? Or if the current version has minor == 0?

bilyana-gospodinova (dismissed):
LGTM

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/MirrorNodeState.java
steven-sheehy, line 117: Think this means we would always attempt to run migrations from current-1 to current on startup instead of only doing that once. Are we sure that's what we want? Maybe we should set them equal or use a stateful previous?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/MirrorNodeState.java
```diff
@@ -101,15 +105,25 @@ private void init() {
             return;
         }
 
+        Optional<RecordFile> latest = recordFileRepository.findLatest();
+        final var bootstrapConfig = new BootstrapConfigProviderImpl().getConfiguration();
```
steven-sheehy, line 109: Can we use mirrorNodeEvmProperties.getConfiguration() here?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/MirrorNodeState.java
```diff
@@ -101,15 +105,25 @@ private void init() {
...
+        final var currentSemanticVersion =
+                bootstrapConfig.getConfigData(VersionConfig.class).servicesVersion();
```
steven-sheehy, line 111: Does this have an accurate version reflecting the library version? I see in VersionConfig it defaults to 0.0.0.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/MirrorNodeState.java
kselveliev: We can addressed.

steven-sheehy (approved):

PR #10603, opened 2025-03-12
Update Contract Create Transformer for failed transaction with result

### LINKED ISSUE #10590: Contract Create BlockItem contents differ from Record Item: Receipt Contract Id
### Description

Certain blocks on testnet when transformed through the Contract Create Transformer yield different record items than the record stream and when parsed lead to differences in persistence. The block items contain an extra contract id in the transaction record receipt, where as the corresponding record items do not have a contract id in the receipt. 

### Steps to reproduce

1. Transform block `000000000000000000000000000036394120.blk.gz`. 

The consensus timestamp of this difference is: `1741220051009841000`
The transformed record item has this contract id in it's receipt: `5659374`
This leads to differences in the contract_result, contract_transaction and contract_transaction_hash tables.

The corresponding record file is: `2025-03-06T00_14_10.095626000Z.rcd.gz`.

### Additional context

_No response_

### Hedera network

testnet

### Version

main

### Operating system

None

edwin-greene:
**Description**:
Updates the transformer to not populated the contract id in the receipt if the transaction is unsucessful.

**Related issue(s)**:

Fixes #10590 

**Notes for reviewer**:
Tested with testnet block that exhibited this problem. 

**Checklist**

### DIFF
--- a/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/ContractCreateTransformer.java
+++ b/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/ContractCreateTransformer.java
@@ -26,7 +26,7 @@ final class ContractCreateTransformer extends AbstractBlockItemTransformer {
 
         var recordBuilder = recordItemBuilder.transactionRecordBuilder();
         recordBuilder.setContractCreateResult(contractCreate.getContractCreateResult());
-        if (contractCreate.getContractCreateResult().hasContractID()) {
+        if (contractCreate.getContractCreateResult().hasContractID() && blockItem.isSuccessful()) {
             recordBuilder
                     .getReceiptBuilder()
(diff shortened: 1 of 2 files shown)


hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/transformer/ContractCreateTransformer.java
```diff
@@ -26,7 +26,7 @@ protected void doTransform(BlockItemTransformation blockItemTransformation) {
 
         var recordBuilder = recordItemBuilder.transactionRecordBuilder();
         recordBuilder.setContractCreateResult(contractCreate.getContractCreateResult());
-        if (contractCreate.getContractCreateResult().hasContractID()) {
+        if (contractCreate.getContractCreateResult().hasContractID() && blockItem.isSuccessful()) {
```
steven-sheehy, line 29: Why would there be any contract id in the output if it wasn't successful?

PR #10615, opened 2025-03-13
Handle optional block stream transaction output

### LINKED ISSUE #10614: Block item transformer assumed transaction output always exists when successful
### Description

Various block item transformers assumed the transaction output for the handled transaction should exist in the block stream when the transaction is successful. Turns out it only exists when there's valid data defined in proto message to emit.

For example, when a successful crypto transfer transaction doesn't incur custom fees, there won't be a `CryptoTransferOutput` transaction output block item. 

### Steps to reproduce

Check some transactions in block stream files.

### Additional context

_No response_

### Hedera network

_No response_

### Version

v0.126.0-SNAPSHOT

### Operating system

None

xin-hedera:
**Description**:

Fix the bug that block item transformers fail when there is no optional transaction output

- Handle transaction output when it exists
- Log cloud bucket path prefix if set

**Related issue(s)**:

Fixes #10614

**Notes for reviewer**:

**Checklist**


steven-sheehy (approved):

jnels124 (approved):
LGTM

PR #10654, opened 2025-03-17
Adapt acceptance tests and web3 to be able to force 100% traffic to mod

kselveliev:
**Description**:
This PR adds a temporary parameter `Is-Modularized` in the web3 header.
This is needed because there is a difference between gasUsed and some methods return results, when run against mono or modularized.

We introduced a traffic percentage which can split % of the requests agains mono and % agains modularized. 
Problem is that when the web3 is deployed with a trafic split not equal to 0.0 or 1.0 acceptance tests can start failing when run against the web3 because its expects result for modularized evm but the call was processed by mono.

To fix this this new temporary `Is-Modularized` header forces traffic to go 100% to modularizedEvm ONLY when the `evmProperties.isModularized=true`.

Changes applied:

Test Module:

MirrorNodeClient - adapt web3 post request to pass the `Is-Modularized` when we run the tests with modularized 

Web3:
`ContractController` - read new `Is-Modularized` header and force traffic to modularized if set to true.

**Related issue(s)**:

Fixes #10654

**Checklist**


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/ContractController.java
```diff
@@ -98,6 +102,14 @@ private ContractExecutionParameters constructServiceParameters(ContractCallReque
 
         boolean isModularized = evmProperties.directTrafficThroughTransactionExecutionService();
 
+        boolean isModularizedRequest = Boolean.parseBoolean(isModularizedHeader);
+        // Temporary workaround to ensure modularized services are fully available when enabled.
+        // This prevents flakiness in acceptance tests, as directTrafficThroughTransactionExecutionService()
+        // can distribute traffic between the old and new logic.
+        if (isModularizedRequest && evmProperties.isModularizedServices()) {
```
sdimitrov9, line 109: I would think about allowing the header to fully override the traffic flow for simplicity and to avoid being misled by having the header 'true' and still go trough the mono logic (in case header is true and isModularizedServices is false) and vice versa:


        final var block = request.getBlock();
        //Adding the ability to distribute the whole traffic to either mono or modularized logic by overriding the
        //traffic direction calculation logic by using the Is-Modularized header.
        boolean isModularized = isModularizedHeader == null ?
                evmProperties.directTrafficThroughTransactionExecutionService() :
                Boolean.parseBoolean(isModularizedHeader);

        return ContractExecutionParameters.builder()
                .block(block)
                .callData(data)
                .callType(callType)
                .gas(request.getGas())
                .isEstimate(request.isEstimate())
                .isModularized(isModularized)
                .isStatic(isStaticCall)
                .receiver(receiver)
                .sender(sender)
                .value(request.getValue())
                .build();`

bilyana-gospodinova (approved):
LGTM

jnels124 (approved):
LGTM

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/ContractController.java
```diff
@@ -98,6 +102,14 @@ private ContractExecutionParameters constructServiceParameters(ContractCallReque
 
         boolean isModularized = evmProperties.directTrafficThroughTransactionExecutionService();
 
+        boolean isModularizedRequest = Boolean.parseBoolean(isModularizedHeader);
```
jnels124, line 105: On second thought. Should we allow this to also direct to mono? If we send,  `"Is-Modularized": "false"`. We should be able to guarantee mono but current implementation wouldn't respect that

jnels124: maybe could just change to 
`boolean isModularized = StringUtils.isEmpty(isModularizedHeader) ?evmProperties.directTrafficThroughTransactionExecutionService(); : false`

kselveliev: Yeah you are correct that setting "Is-Modularized": "false" won't be respected.

Still looking into it to check of any potential side effects if we try to guarantee mono execution.

 There are some other places in the code which are checking both `evmProperties.isModularized` and the `isModularized` param in `ContractExecutionParameters`. But still going for mono seems a bit counterIntuitive when we are having  `evmProperties.isModularized` to true. The two other ways of guaranteeing mono is just setting the percentage config to 0.0 or just  evmProperties.isModularized=false. This is why I'm a bit hesitant of this change.

steven-sheehy: Think it makes sense to have `is-modularized=false` guarantee mono path, but not critical to have. Since it sounds like it might introduce regressions we can keep as is. This is all temporary so it doesn't have to be ideal.

steven-sheehy (approved):

PR #10663, opened 2025-03-18
Fix flaky eth call acceptance test

### LINKED ISSUE #10618: Fix flaky test in call.feature acceptance test
### Description

While running acceptance tests I noticed that the test `I burn NFT and get the total supply and balance` in CallFeature.java fails from time to time with 400 Bad Request. It is possible that this test requires @Retry annotation but before going there, debugging is needed to check if something else is happening that is preventing the successful completion of this test.

### Steps to reproduce

Run all acceptance tests multiple times

### Additional context

_No response_

### Hedera network

other

### Version

v0.126.0-SNAPSHOT

### Operating system

None

kselveliev:
**Description**:
This PR adds a retry mechanism for `I burn NFT and get the total supply and balance`.

The test is flaky and is executed right after the token mint operation and then we wait:
```
Given I mint a NFT
Then the mirror node should return status 200 for the HAPI transaction
I burn NFT and get the total supply and balance
```

Seems like even though we wait for successful hapi transaction the next token burn simulations fails because it does not yet find the nft in db and fails.(This happens in very very rare cases, have seen similar problems with other flaky tests in the past). Simulated that with an integration test if we removed the nft persist method. Will cause the same error as in the acceptance test.

Fixes #10618 

**Checklist**

### DIFF
--- a/hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/CallFeature.java
+++ b/hedera-mirror-test/src/test/java/com/hedera/mirror/test/e2e/acceptance/steps/CallFeature.java
@@ -64,6 +64,9 @@ import java.util.List;
 import lombok.CustomLog;
 import lombok.Getter;
 import lombok.RequiredArgsConstructor;
+import org.springframework.retry.annotation.Backoff;
+import org.springframework.retry.annotation.Retryable;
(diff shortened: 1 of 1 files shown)


bilyana-gospodinova (approved):
LGTM

filev94 (approved):
LGTM

PR #10670, opened 2025-03-19
Fix the wrongful modularized flag flip

### LINKED ISSUE #10669: Fix issue with reflection usage to change modularized flag
### Description

While debugging web3 tests i noticed that sometimes the tests change from old flow to new flow. For some of the tests that is okay, but then again those same tests change the initial flag with which the test suite was run.

We need to check the tests that change the modularized flag to true in a scenario when modularized flag is already set to true in the test suites.

Because of the change, in a finally block the flag is set to false meaning the next tests will execute with modularized flag set to false, which is not something we want.

### Steps to reproduce

Run the tests in debug mode with these env variables: 
CI=true;HEDERA_MIRROR_WEB3_EVM_MODULARIZEDSERVICES=true;HEDERA_MIRROR_WEB3_EVM_MODULARIZEDTRAFFICPERCENT=1.0;MIRROR_NODE_SCHEMA=v1;SPRING_PROFILES_ACTIVE=v1

### Additional context

_No response_

### Hedera network

other

### Version

v0.126.0-SNAPSHOT

### Operating system

Other

bilyana-gospodinova:
**Description**:
One of the integration tests (`updateCustomFeesForTokenWithFixedFee`) at the end of execution always used to set the `modularizedServices` flag to `false`. Now the flag is set to its previous value as it should be.

Fixes #10669

### DIFF
--- a/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
+++ b/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
@@ -703,10 +703,6 @@ public abstract class AbstractContractCallServiceTest extends Web3IntegrationTes
         mirrorNodeEvmProperties.setProperties(propertiesMap);
     }
 
-    protected void deactivateModularizedFlag() {
-        mirrorNodeEvmProperties.setModularizedServices(false);
-    }
-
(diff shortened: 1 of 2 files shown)


kselveliev (approved):

edwin-greene (approved):
Looks good

jnels124 (approved):

PR #10678, opened 2025-03-20
Update batch transaction design

### LINKED ISSUE #10597: Update batch transaction design for block streams
### Problem

There have been changes to the implementation of batch transactions in block streams.

### Solution

- Document change of proto Transaction to signed transaction bytes
- Include design for block item processing with de-duplication of inner transactions

### Alternatives

_No response_

jnels124:
**Description**:
This pr updates the design for hip-551 (batch transactions) to handle deduplication of inner transactions in the block stream

**Related issue(s)**:

Fixes #10597

**Notes for reviewer**:

**Checklist**


docs/design/batch-transactions.md
```diff
@@ -91,6 +91,113 @@ public class SqlEntityListener implements EntityListener, RecordStreamFileListen
...
+                    .transactionOutputs(Collections.unmodifiableMap(transactionOutputs))
+                    .stateChanges(Collections.unmodifiableList(stateChangesList))
+                    .previous(context.getLastBlockItem())
+                    .build();
+
+            context.getBlockFile().item(blockItem);
+            context.setLastBlockItem(blockItem);
+
+            // read child transactions
+            readEventTransactions(context);
```
jnels124, line 194: can add a note to indicate on the last item in the batch we can peek at the next block item and if it is a event transaction, we need to see if it is part of the batch (a child transaction of last item in batch).  Don't think that is really necessary as the items will still be processed correctly. Just  the call stack would be kind of "weird".. It would be the call from read batch transactions that potentially finishes reading out the round even if its inner transactions are not part of the ending events.

docs/design/batch-transactions.md
```diff
@@ -91,6 +87,99 @@ public class SqlEntityListener implements EntityListener, RecordStreamFileListen
...
+
+    /**
+     * Iterate over inner transactions and create block items for each inner transaction (along with preceding and child
+     * transactions)
+     *
+     * @param stateChangesList list of state changes for entire batch
+     * @param context
+     * @param transactionBody transaction body of batch transaction
+     * */
+    private void readInnerTransactions(ArrayList<StateChanges> stateChangesList,
```
xin-hedera, line 160: I'd prefer to hide the complexity in the context and reuse existing code

1. in `Context.setLastBlockItem`, if it's an atomic batch transaction, keep the atomic transaction body and reset atomic transaction `index` to 0 (these are new fields in `Context` class)
2. Add a new function `Context.getApplicationTransaction`, which would first try to read an `EVENT_TRANSACTION` block item and extract the application transaction, or get the current transaction from the atomic transaction body (also advance the `index`)
3. in `readEventTransactions`, replace the top level while loop with
  `` while context.getApplicationTransaction() != null`

xin-hedera (changes requested):

docs/design/batch-transactions.md
jnels124: I like that approach better.

xin-hedera (approved):
LGTM

steven-sheehy (approved):

PR #10680, opened 2025-03-20
Don't print whole account data when payer account is invalid

### LINKED ISSUE #10620: Fix error response from modularized calls with invalid input
### Description

When executing a modularized call against the web3. If you accidentally switch the `from` and `to` parameters
we can get into this logic in the modularized library ->
```
if (account == null || account.deleted() || account.smartContract()) {
    throw new IllegalStateException(
            "Category " + category + " payer account should have been rejected " + account);
} 
```
And this will print the whole account toString in the logs.

Image

### Steps to reproduce

Switch from and to parameters during contract call. It's a corner case may not work with every call really depends on the from and to addresses.

### Additional context

_No response_

### Hedera network

_No response_

### Version

v0.125.0

### Operating system

macOS

sdimitrov9:
**Description**:
When payer account is not valid during the check in the DispatchValidator, the whole account data is printed as part of the exception message, which results in similar output:
invalidpayer

The outcome of this PR will be to print only the Account alias 
Screenshot 2025-03-21 at 9 13 19

**Related issue(s)**:

Fixes #10620

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/GenericControllerAdvice.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/GenericControllerAdvice.java
@@ -73,7 +73,7 @@ class GenericControllerAdvice extends ResponseEntityExceptionHandler {
 
     @ExceptionHandler
     private ResponseEntity<?> mirrorEvmTransactionError(final MirrorEvmTransactionException e, WebRequest request) {
-        log.error(
+        log.warn(
                 "Mirror EVM transaction error: {}, detail: {}, data: {}, isModularized: {}",
                 e.getMessage(),
                 e.getDetail(),
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/GenericControllerAdvice.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/GenericControllerAdvice.java
@@ -73,8 +73,8 @@ class GenericControllerAdvice extends ResponseEntityExceptionHandler {
 
     @ExceptionHandler
     private ResponseEntity<?> mirrorEvmTransactionError(final MirrorEvmTransactionException e, WebRequest request) {
-        log.warn(
-                "Mirror EVM transaction error: {}, detail: {}, data: {}, isCallModularized: {}",
+        log.error(
+                "Mirror EVM transaction error: {}, detail: {}, data: {}, isModularized: {}",
                 e.getMessage(),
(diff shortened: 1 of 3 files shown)


filev94:
Do we need to create a ticket for services to propagate a better message?

sdimitrov9:
> Do we need to create a ticket for services to propagate a better message?

Already talked to Neeha and Zhivko and we will come up with detailed approach next monday, probably in the form of a ticket.

kselveliev (dismissed):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/GenericControllerAdvice.java
```diff
@@ -73,8 +73,8 @@ private ResponseEntity<?> badRequest(final Exception e, final WebRequest request
 
     @ExceptionHandler
     private ResponseEntity<?> mirrorEvmTransactionError(final MirrorEvmTransactionException e, WebRequest request) {
-        log.warn(
-                "Mirror EVM transaction error: {}, detail: {}, data: {}, isCallModularized: {}",
+        log.error(
```
steven-sheehy, line 76: Change back to warn. If the user submits bad data and we return bad request then it should only be a warning. Only if it's some server issue should it be an error.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/ContractCallService.java
```diff
@@ -123,7 +123,8 @@ protected HederaEvmTransactionProcessingResult doProcessCall(
                 result = mirrorEvmTxProcessor.execute(params, estimatedGas);
             }
         } catch (IllegalStateException | IllegalArgumentException e) {
-            throw new MirrorEvmTransactionException(e.getMessage(), EMPTY, EMPTY);
+            log.error("Error processing contract call: ", e);
```
steven-sheehy, line 126: Why do we need this log line that prints a stacktrace? We should not print stacktraces or even extra log lines for bad user input as logs are expensive to store at scale.

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/controller/GenericControllerAdvice.java
sdimitrov9: Fixed

steven-sheehy (approved):

PR #10687, opened 2025-03-21
Implement retry and caching for throttle definitions v2

### LINKED ISSUE #10686: Contract call against modularized can fail with unparseable throttle definitions
### Description

Initialization of transaction executor can fail when parsing throttle definitions in DB

### Steps to reproduce

Happens when we try to execute request agains modularized evm on integration.
Needs a retry mechanism and caching to fix the issue.

### Additional context

_No response_

### Hedera network

other

### Version

v0.126.0

### Operating system

macOS

kselveliev:
This PR implements a throttle definition parsing mechanism with retry and caching.
If not valid definitions is found in DB we resolve to system file loader.

Mechanism
Retry and parse file data up to 10 times(Similar to RatesAndFeesLoader). Return the first valid from DB. Cache the result
In case of empty DB data - resolve to systemFileLoader -  Cache the result
In case of 10 failures - resolve to systemFileLoader - Cache the result

`SystemFileLoader` - add the retry mechanism and cache throttles similar to RatesAndFeesLoader. Method can be reused when we add retry and caching for rates and fees again for modularized
`SystemFileLoaderTest` - adds tests

Fixes #10686

**Checklist**


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/throttle/ThrottleDefinitionsManager.java
```diff
@@ -0,0 +1,94 @@
...
+import lombok.RequiredArgsConstructor;
+import org.springframework.cache.annotation.CacheConfig;
+import org.springframework.cache.annotation.Cacheable;
+import org.springframework.retry.support.RetryTemplate;
+
+@Named
+@CacheConfig(cacheManager = CACHE_MANAGER_SYSTEM_FILE)
+@CustomLog
+@RequiredArgsConstructor
+public class ThrottleDefinitionsManager {
```
steven-sheehy, line 31: I'd prefer this logic live in `SystemFileLoader` so it can be reused for other system files. Also so we don't have to have conditional logic for 123 both in `FileReadableKVState` and in `SystemFileLoader`. The retry logic should be generic similar to `RatesAndFeesLoader.getFileData()` can be passed a generic parser.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/throttle/ThrottleDefinitionsManager.java
```diff
@@ -0,0 +1,94 @@
...
+        AtomicReference<FileData> successfulFileData = new AtomicReference<>(null);
+
+        try {
+            retryTemplate.execute(context -> {
+                Optional<FileData> fileDataOptional = fileDataRepository.getFileAtTimestamp(fileId, nanoSeconds.get());
+
+                if (fileDataOptional.isEmpty()) {
+                    // Stop retrying if no file exists at the timestamp
+                    return Optional.empty();
+                }
```
steven-sheehy, line 56: Rewrite this so we don't have the atomic reference and chain the optional to map and have an orElseGet default similar to RatesAndFeesLoader.getFileData()

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/throttle/ThrottleDefinitionsManager.java
```diff
@@ -0,0 +1,94 @@
...
+@CacheConfig(cacheManager = CACHE_MANAGER_SYSTEM_FILE)
+@CustomLog
+@RequiredArgsConstructor
+public class ThrottleDefinitionsManager {
+
+    private final FileDataRepository fileDataRepository;
+    private final SystemFileLoader systemFileLoader;
+
+    @Named("customThrottleParser")
+    private final ThrottleParser throttleParser;
```
steven-sheehy, line 37: Not sure we need to use the ThrottleParser since we don't need to validate the contents of the throttle. We can probably get away with just `ThrottleDefinitions.PROTOBUF.parse()`.

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/throttle/ThrottleConfig.java
```diff
@@ -0,0 +1,16 @@
...
+import com.hedera.node.app.throttle.ThrottleParser;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class ThrottleConfig {
+
+    @Bean
+    public ThrottleParser customThrottleParser() {
+        return new ThrottleParser();
```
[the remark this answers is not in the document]
kselveliev: Removed this class after refactoring

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/FileReadableKVState.java
```diff
@@ -51,6 +55,10 @@ protected File readFromDataSource(@Nonnull FileID key) {
         final var fileEntityId = toEntityId(key);
         final var fileId = fileEntityId.getId();
 
+        if (fileId == 123) {
```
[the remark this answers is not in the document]
kselveliev: Addressed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/throttle/ThrottleConfig.java
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package com.hedera.mirror.web3.state.throttle;
+
+import com.hedera.node.app.throttle.ThrottleParser;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class ThrottleConfig {
```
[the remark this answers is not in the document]
kselveliev: The class got removed after latest commit

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/throttle/ThrottleDefinitionsManager.java
```diff
@@ -0,0 +1,94 @@
...
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicReference;
+import lombok.CustomLog;
+import lombok.RequiredArgsConstructor;
+import org.springframework.cache.annotation.CacheConfig;
+import org.springframework.cache.annotation.Cacheable;
+import org.springframework.retry.support.RetryTemplate;
+
+@Named
+@CacheConfig(cacheManager = CACHE_MANAGER_SYSTEM_FILE)
```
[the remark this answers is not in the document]
kselveliev: Addressed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/throttle/ThrottleDefinitionsManager.java
kselveliev: Added the logic in SystemFileLoader, made the retry logic as close as possible to RatesAndFeesLoader.
We should be able reuse it for fees and exchange rates just passing the codec object now.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/throttle/ThrottleDefinitionsManager.java
kselveliev: Addressed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/throttle/ThrottleDefinitionsManager.java
kselveliev: Removed ThrottleParser, now just passing ThrottleDefinitions.PROTOBUF and in loadWithRetry() we accept it as a generic codec and call the parse method.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/FileReadableKVState.java
```diff
@@ -34,26 +36,36 @@ public class FileReadableKVState extends AbstractReadableKVState<FileID, File> {
...
     protected File readFromDataSource(@Nonnull FileID key) {
         final var timestamp = ContractCallContext.get().getTimestamp();
         final var fileEntityId = toEntityId(key);
         final var fileId = fileEntityId.getId();
+        final var currentTimestamp = getCurrentTimestamp();
+
+        if (SystemEntity.THROTTLE_DEFINITIONS
+                .getScopedEntityId(commonProperties)
+                .equals(fileEntityId)) {
+            return systemFileLoader.loadThrottles(fileId, key, currentTimestamp);
```
steven-sheehy, line 63: I think what we want is for `SystemLoader` to take care of loading system files from db or classpath. With current change some methods load from db while others only load from classpath as a fallback. I think we should take the time now to refactor it to make `load()` load from db and be done first and not as a fallback for all system files. This also avoids having to check fileID in both kv state and in loader.

```suggestion
        var systemFile = systemFileLoader.load(key, currentTimestamp);
        if (systemFile != null) {
            return systemFile;
        }
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/SystemFileLoader.java
```diff
@@ -33,6 +51,54 @@ public class SystemFileLoader {
         return getSystemFiles().get(key);
     }
 
+    @Cacheable(
+            cacheManager = CACHE_MANAGER_SYSTEM_FILE_MODULARIZED,
+            cacheNames = CACHE_NAME_THROTTLE,
+            key = "'now'",
+            unless = "#result == null")
+    public File loadThrottles(final long fileId, final FileID key, final long currentTimestamp) {
```
steven-sheehy, line 59: We don't need both fileId and FileID. Change to just FileID and convert internally.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/SystemFileLoader.java
```diff
@@ -33,6 +51,54 @@ public class SystemFileLoader {
...
+                                    e);
+                            nanoSeconds.set(fileData.getConsensusTimestamp() - 1);
+                            throw new InvalidFileException(e);
+                        }
+                    })
+                    .orElseGet(() -> load(key)));
+        } catch (InvalidFileException e) {
+            log.warn("All retry attempts failed for fileId {}: {}", fileId, e.getMessage());
+            return load(key);
+        }
```
steven-sheehy, line 99: We should be able to pass a `RecoveryCallback` lambda as the second arg to `retryTemplate.execute()` instead of catching. Also, we can do orElseThrow instead of orElseGet to stay DRY

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/SystemFileLoader.java
```diff
@@ -33,6 +51,54 @@ public class SystemFileLoader {
         return getSystemFiles().get(key);
     }
 
+    @Cacheable(
+            cacheManager = CACHE_MANAGER_SYSTEM_FILE_MODULARIZED,
+            cacheNames = CACHE_NAME_THROTTLE,
```
steven-sheehy, line 56: If we want to have different caches for each file type but a single load() method then we probably can't use the annotation approach and have to inject the `CacheManager` to lookup the individual caches. But if they're all the same spec we can probably use the same cache name for each since FileIDs are already unique.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/state/SystemFileLoaderTest.java
```diff
@@ -13,15 +21,64 @@
...
 
-    private final SystemFileLoader systemFileLoader = new SystemFileLoader(new MirrorNodeEvmProperties());
+    private static final long TEST_FILE_ID = 100L;
+    private static final FileID TEST_FILE_ID_PROTO =
+            FileID.newBuilder().fileNum(TEST_FILE_ID).build();
+    private static final File TEST_SYSTEM_FILE = File.newBuilder()
+            .fileId(TEST_FILE_ID_PROTO)
+            .contents(Bytes.wrap(new byte[] {4, 5, 6}))
+            .build();
+    private static final long TEST_FILE_ID_THROTTLES = 123L;
```
steven-sheehy, line 51: We shouldn't use hardcoded well known IDs. Use the SystemEntity enum and convert to an ID using common props

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/state/SystemFileLoaderTest.java
```diff
@@ -13,15 +21,64 @@
...
+    private static final com.hederahashgraph.api.proto.java.ThrottleDefinitions throttleDefinitions =
+            com.hederahashgraph.api.proto.java.ThrottleDefinitions.newBuilder()
+                    .addThrottleBuckets(com.hederahashgraph.api.proto.java.ThrottleBucket.newBuilder()
+                            .build())
+                    .build();
+
+    private static final FileData throttleDefinitionsFileData = FileData.builder()
+            .consensusTimestamp(200L)
+            .fileData(throttleDefinitions.toByteArray())
+            .build();
```
steven-sheehy, line 69: The previous version of the class had no static test data. I prefer that approach if at all possible since it keeps the test input dynamic and less coupling between test methods.

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/SystemFileLoader.java
```diff
@@ -33,6 +51,54 @@ public class SystemFileLoader {
...
+     * Load file data with retry logic and parsing. This method will attempt to load and parse file data,
+     * retrying with earlier versions if parsing fails.
+     *
+     * @param fileId The ID of the file to load
+     * @param key The FileID object representing the file
+     * @param currentTimestamp The current timestamp to start loading from
+     * @param codec The codec to use for parsing
+     * @return The parsed file data, or the default value if no valid data is found
+     */
+    public <T> File loadWithRetry(final long fileId, final FileID key, final long currentTimestamp, Codec<T> codec) {
```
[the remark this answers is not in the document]
kselveliev: Addressed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/SystemFileLoader.java
kselveliev: Addressed

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/state/SystemFileLoaderTest.java
kselveliev: Addressed

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/state/SystemFileLoaderTest.java
kselveliev: Removed almost all static data, except for the corruptData

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/SystemFileLoader.java
kselveliev: I thought we might be able to to this directly  and call  loadFromDb and beforehand checking if it is a systemFile-> 

```
Cacheable(
        cacheManager = CACHE_MANAGER_SYSTEM_FILE_MODULARIZED,
        cacheNames = CACHE_NAME_MODULARIZED,
        key = "#key.toString()",
        unless = "#result == null")
public File loadFromDb(final FileID key, final long currentTimestamp) {
 
    return loadWithRetry(key, currentTimestamp, ThrottleDefinitions.PROTOBUF);
}
```

but then in load with retry we need to know which protobuf object to pass ThrottleDefinitions.PROTOBUF
and here it gets complicated.

The other approach is to inject the cache 

and do a cache.get() and cache.put() manually but we lose the thread safety that the annotation provides and this might also get complicated

So probably we might want to leave it as it is .

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/FileReadableKVState.java
kselveliev: Me might want to avoid this due to concerns described here https://github.com/hiero-ledger/hiero-mirror-node/pull/10687#discussion_r2010600147

bilyana-gospodinova (dismissed):
LGTM

bilyana-gospodinova (dismissed):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/SystemFileLoader.java
steven-sheehy: I don't see the complexity, it seems really straightforward. Just put the genesis files and corresponding codecs into a lazy map by FileID. I think all the files can have the same cache spec for now since both rates and fees had the same in mono.

```java

    @Getter(lazy = true)
    private final Map<FileID, SystemFile> systemFiles = loadAll();

    private record SystemFile(File genesisFile, Codec<?> codec) {
    }

    @Cacheable(cacheManager = CACHE_MANAGER_SYSTEM_FILE_MODULARIZED, cacheNames = CACHE_NAME_MODULARIZED, key = "#key", unless = "#result == null")
    public @Nullable File load(@Nonnull FileID key, long consensusTimestamp) {
        if (key.shardNum() != commonProperties.getShard() || key.realmNum() != commonProperties.getRealm()) {
            return null;
        }

        var systemFile = getSystemFiles().get(key);

        if (systemFile == null) {
            return null;
        }

        return loadWithRetry(key, consensusTimestamp, systemFile);
    }

    private File loadWithRetry(final FileID key, final long currentTimestamp, SystemFile systemFile) {
```

Take care around the `key` param to Cacheable since FileID does not have an overridden toString(). Not sure the exact syntax but it should just use FileID directly which should invoke its equals and hashCode.

kselveliev: Applied the above suggestion had to do some small adaptations also in the loadAll method.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/SystemFileLoader.java
```diff
@@ -2,57 +2,146 @@
...
-                load(SystemEntity.EXCHANGE_RATE, fileSchema.genesisExchangeRates(configuration)),
-                load(SystemEntity.NETWORK_PROPERTY, fileSchema.genesisNetworkProperties(configuration)),
-                load(
-                        SystemEntity.HAPI_PERMISSION,
-                        Bytes.EMPTY), // genesisHapiPermissions() fails to load files from the classpath
-                load(SystemEntity.THROTTLE_DEFINITION, fileSchema.genesisThrottleDefinitions(configuration)));
-
-        return files.stream().collect(Collectors.toMap(File::fileId, Function.identity()));
+                new SystemFile(load(SystemEntity.ADDRESS_BOOK_101, Bytes.EMPTY), null),
+                new SystemFile(load(SystemEntity.ADDRESS_BOOK_102, Bytes.EMPTY), null),
```
steven-sheehy, line 129: Use NodeAddressBook.PROTOBUF

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/SystemFileLoader.java
```diff
@@ -2,57 +2,146 @@
...
+    @Cacheable(
+            cacheManager = CACHE_MANAGER_SYSTEM_FILE_MODULARIZED,
+            cacheNames = CACHE_NAME_MODULARIZED,
+            key = "#key",
+            unless = "#result == null")
+    public @Nullable File load(@Nonnull FileID key, long consensusTimestamp) {
+        if (key.shardNum() != commonProperties.getShard() || key.realmNum() != commonProperties.getRealm()) {
+            return null;
+        }
+
```
steven-sheehy, line 70: Looking at this again, this can be removed as redundant. The map will already filter out non-relevant shards and realms.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/SystemFileLoader.java
```diff
@@ -2,57 +2,146 @@
...
     public @Nullable File load(@Nonnull FileID key) {
-        return getSystemFiles().get(key);
+        var systemFile = getSystemFiles().get(key);
+        return systemFile != null ? systemFile.genesisFile() : null;
+    }
+
+    @Cacheable(
+            cacheManager = CACHE_MANAGER_SYSTEM_FILE_MODULARIZED,
+            cacheNames = CACHE_NAME_MODULARIZED,
+            key = "#key",
```
steven-sheehy, line 64: We should have a test that verifies the cache. Insert an entry, load() it, then delete it from db, load it again and ensure it's still equal.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/FileReadableKVState.java
```diff
@@ -50,10 +50,15 @@ protected File readFromDataSource(@Nonnull FileID key) {
...
+        if (systemFileLoader.isSystemFile(key)) {
+            return systemFileLoader.load(key, currentTimestamp);
+        }
 
         return timestamp
                 .map(t -> fileDataRepository.getFileAtTimestamp(fileId, t))
-                .orElseGet(() -> fileDataRepository.getFileAtTimestamp(fileId, getCurrentTimestamp()))
+                .orElseGet(() -> fileDataRepository.getFileAtTimestamp(fileId, currentTimestamp))
                 .map(fileData -> mapToFile(fileData, key, timestamp))
                 .orElseGet(() -> systemFileLoader.load(key));
```
steven-sheehy, line 63: In what situations would this load get called? It seems like the previous load should handle every scenario and this call and its method in SystemFileLoader can be removed now.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/SystemFileLoader.java
kselveliev: Addressed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/FileReadableKVState.java
kselveliev: Removed

steven-sheehy: only its invocation was removed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/SystemFileLoader.java
kselveliev: Added an integration test to verify the cache, instead of delete from db I did add new fileData with the same id and then verify the cache would return the first data instead.

Could not figure out a way to make the delete work, domainBuilder does not support it and fileDataRepo.delete does not do the job either in test env.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/FileReadableKVState.java
kselveliev: Maybe we can leave this for another PR for the sake of speed, most of the the tests in SystemFileLoaderTest rely their expected results to be obtained from this load method which only reads the classpath.

We'll need to refactor pretty much all of them and add more static data to be able to make them work.

steven-sheehy (approved):
LGTM. Please open a ticket to capture the cleanup effort.

bilyana-gospodinova (approved):

PR #10713, opened 2025-03-24
Added an optional query parameter to include tokenBalance in the getAccount query

### LINKED ISSUE #10543: feat: minor - allow tokenBalances=false in getAccount
### Problem

When I request `getAccount` for an account ID that has many tokens, I get a large response containing token balances. However, in many cases I do not need this information.

To reproduce:
```shell
curl -X 'GET' \
  'https://testnet.mirrornode.hedera.com/api/v1/accounts/0.0.1534?limit=2&order=asc&transactiontype=cryptotransfer&transactions=true' \
  -H 'accept: application/json'
```

### Solution

I request the addition of an optional query param to this API, `tokenBalances`, which accepts `true` (default) or `false`.
- If `tokenBalances=false`, then the `balance.tokens` array should be omitted from the response.
- If `tokenBalances=true`, then the `balance.tokens` array should be included in the response (current behaviour).

Example:
```shell
curl -X 'GET' \
  'https://testnet.mirrornode.hedera.com/api/v1/accounts/0.0.1534?limit=2&order=asc&transactiontype=cryptotransfer&transactions=true&tokenBalances=false' \
  -H 'accept: application/json'
```

Refs:
- https://testnet.mirrornode.hedera.com/api/v1/docs/#/accounts/getAccount

### Alternatives

Nil

HarshSawarkar:
**Description**:

Successfully added support for an optional query parameter in the Accounts query API.

**Related issue(s)**:

Fixes #10543 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-rest/accounts.js
+++ b/hedera-mirror-rest/accounts.js
@@ -35,9 +35,10 @@ const getEntityStakeQuery = (filter, isHistorical = false) => {
 /**
  * Processes one row of the results of the SQL query and format into API return format
  * @param {Object} row One row of the SQL query result
+ * @param includeTokenBalances include token balance info or not
  * @return {Object} Processed account record
  */
-const processRow = (row) => {
+const processRow = (row, includeTokenBalances = true) => {
   const alias = base32.encode(row.alias);
   const balance =
     row.balance === undefined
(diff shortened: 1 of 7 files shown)


hedera-mirror-rest/accounts.js
```diff
@@ -35,17 +35,18 @@ const getEntityStakeQuery = (filter, isHistorical = false) => {
...
+const processRow = (row, includeTokenBalances = true) => {
   const alias = base32.encode(row.alias);
   const balance =
     row.balance === undefined
       ? null
       : {
           balance: row.balance,
           timestamp: utils.nsToSecNs(row.balance_timestamp),
-          tokens: utils.parseTokenBalances(row.token_balances),
+          ...(includeTokenBalances ? {tokens: utils.parseTokenBalances(row.token_balances)} : {}),
```
jnels124, line 49: should pass `[]` instead of `{}`

jnels124 (changes requested):
We should also exclude querying for the token balance information in this case.

PR #10716, opened 2025-03-25
Support non-zero realm in importer

### LINKED ISSUE #10617: Use configured shard and realm in importer
### Problem

The importer makes assumption about zero realm and shard when interacting with system files

### Solution

* Change address book service to use file IDs with configured realm and shard
* Analyze other places to ensure no non-zero usage
* Change `DomainBuilder`, `RecordItemBuilder`, `BlockItemBuilder`, `BlockFileBuilder`, and `RecordFileBuilder` to use configured realm and shard
* Test non-zero realm and shard in config and fix tests

### Alternatives

_No response_

steven-sheehy:
**Description**:

* Change address book service to use configured realm and shard
* Change `DomainBuilder`, `RecordItemBuilder`, `BlockItemBuilder`, `BlockFileBuilder`, and `RecordFileBuilder` to use configured realm and shard
* Change alias and EVM address lookup to pass shard and realm
* Partially replace `SystemEntity` enum with `SystemEntities` bean
* Set redis `maxRetriesPerRequest` to 10 in REST API tests to decrease flakiness

**Related issue(s)**:

Fixes #10617

**Notes for reviewer**:

I will do a follow up to fully replace `SystemEntity` with `SystemEntities`. It'd be too large of a change to do in this PR.

**Checklist**

### DIFF
--- a/hedera-mirror-common/src/main/java/com/hedera/mirror/common/CommonConfiguration.java
+++ b/hedera-mirror-common/src/main/java/com/hedera/mirror/common/CommonConfiguration.java
@@ -2,11 +2,18 @@
 
 package com.hedera.mirror.common;
 
+import com.hedera.mirror.common.domain.SystemEntities;
 import org.springframework.boot.autoconfigure.domain.EntityScan;
 import org.springframework.boot.context.properties.ConfigurationPropertiesScan;
+import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 
 @Configuration
 @ConfigurationPropertiesScan(basePackages = "com.hedera.mirror")
 @EntityScan("com.hedera.mirror.common.domain")
-public class CommonConfiguration {}
+public class CommonConfiguration {
+    @Bean
+    SystemEntities systemEntities(CommonProperties commonProperties) {
+        return new SystemEntities(commonProperties);
+    }
+}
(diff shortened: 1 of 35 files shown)


hedera-mirror-common/src/main/java/com/hedera/mirror/common/CommonConfiguration.java
```diff
@@ -2,11 +2,18 @@
...
 import org.springframework.boot.context.properties.ConfigurationPropertiesScan;
+import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 
 @Configuration
 @ConfigurationPropertiesScan(basePackages = "com.hedera.mirror")
 @EntityScan("com.hedera.mirror.common.domain")
-public class CommonConfiguration {}
+public class CommonConfiguration {
+    @Bean
```
jnels124, line 15: Shouldn't common properties be injected as a param here?

jnels124 (approved):

hedera-mirror-common/src/main/java/com/hedera/mirror/common/CommonConfiguration.java
steven-sheehy: I tried that initially but got weird errors in `@WebMvcTest` like `ContractControllerTest`.

jnels124: maybe can add something like `@DependsOn({"commonProperties"})`?

jnels124: Looks like you can inject on the bean method but you need to add `@EnableConfigurationProperties(CommonProperties.class)` on the `CommonConfiguration` class

steven-sheehy: Thanks, went with that approach. Though it's less than ideal since `@ConfigurationPropertiesScan` is supposed to already `@EnableConfigurationProperties` and scan all properties so now we have to manually list all properties.

jnels124 (approved):

xin-hedera (approved):
LGTM

PR #10717, opened 2025-03-25
Use shard and realm for system files in REST

### LINKED ISSUE #10616: Use configured realm and shard for system files in REST
### Problem

The REST API assume a zero realm and shard in a number of places.

### Solution

* Verify `/api/v1/network/exchangerate` uses configured realm and shard
* Verify `/api/v1/network/fees` uses configured realm and shard
* Verify `/api/v1/network/nodes` uses configured realm and shard
* Fix `isValidAddressBookFileIdPattern()` and default `file.id`
* Fix anchor account `'0.0.0'` in `getAccounts()`
* Fix config for `unreleasedSupplyAccounts` to drop shard and realm and use configured
* Search for any other hardcoded realms or shards

### Alternatives

_No response_

jnels124:
**Description**:

Modify the REST API to ensure any system files account for non zero shard and realm in queries.

* Fix exchange rate query
* Fix fee schedule query
* Fix nodes query
* Use `shard.realm.0` for anchor account
* Use configured shard and realm for unreleased supply accounts

**Related issue(s)**:

Fixes #10616 

**Notes for reviewer**:

**Checklist**


hedera-mirror-rest/utils.js
```diff
@@ -1277,10 +1288,11 @@ const formatComparator = (comparator) => {
           comparator.value = parseInt(comparator.value, 16);
         }
         break;
-      case constants.filterKeys.FILE_ID:
+      case constants.filterKeys.FILE_ID: {
```
jnels124, line 1291: will remove `{}`

hedera-mirror-rest/__tests__/entityId.test.js
```diff
@@ -530,3 +530,38 @@ describe('computeContractIdPartsFromContractIdValue', () => {
...
+    {
+      input: '0.0.71eaa748d5252be68c1185588beca495459fdba4',
+      expected: EntityId.of(0, 0, null, '71eaa748d5252be68c1185588beca495459fdba4'),
+    },
+    {
+      input: '71eaa748d5252be68c1185588beca495459fdba4',
+      expected: EntityId.of(0, 0, null, '71eaa748d5252be68c1185588beca495459fdba4'),
+    },
+    {
+      input: '0.0.0x71eaa748d5252be68c1185588beca495459fdba4',
```
steven-sheehy, line 553: I don't think we accept this syntax now. Are you adding support for this? If so, I don't think we should as it's weird to have 0x in the middle of a string

hedera-mirror-rest/accounts.js
```diff
@@ -366,7 +370,7 @@ const getAccounts = async (req, res) => {
     },
   };
 
-  let anchorAcc = '0.0.0';
+  let anchorAcc = `${systemShard}.${systemRealm}.0`;
```
steven-sheehy, line 373: Can we have a test to make sure this works for non-zero?

hedera-mirror-rest/entityId.js
```diff
@@ -293,6 +303,7 @@ const cache = new quickLru({
  * @param {Function} error
  * @return {EntityId}
  */
+// Some calls from num here
```
xin-hedera, line 306:
```suggestion
```

hedera-mirror-rest/service/systemEntity.js
```diff
@@ -18,6 +30,22 @@ class SystemEntity {
   get treasuryAccount() {
     return this.#treasuryAccount;
   }
+
+  get exchangeRateFileId() {
```
steven-sheehy, line 34: I think we should drop the `Id` suffix from the method and field like the existing entries have it.

hedera-mirror-rest/topicmessage.js
```diff
@@ -93,7 +93,7 @@ const getMessageByTopicAndSequenceRequest = async (req, res) => {
   const seqNum = req.params.sequenceNumber;
   validateGetSequenceMessageParams(topicIdStr, seqNum);
 
-  const topicId = EntityId.parse(topicIdStr).getEncodedId();
+  const topicId = EntityId.parseIdParameterToEntityId(topicIdStr).getEncodedId();
```
steven-sheehy, line 96: What areas remain that would use `EntityId.parse()` but not be passed user input?

hedera-mirror-rest/utils.js
```diff
@@ -197,7 +201,14 @@ const isValidSlot = (slot) => slotPattern.test(slot);
...
 const isValidValueIgnoreCase = (value, validValues) => validValues.includes(value.toLowerCase());
 
-const addressBookFileIdPattern = ['101', '0.101', '0.0.101', '102', '0.102', '0.0.102'];
+const addressBookFileIdPattern = [
+  '101',
+  `${systemRealm}.101`,
+  `${systemShard}.${systemRealm}.101`,
+  '102',
+  `${systemRealm}.102`,
+  `${systemShard}.${systemRealm}.102`,
```
steven-sheehy, line 210: Probably we can add a method to `EntityId` to generate an array with the 3 forms. Then we can get use `SystemEntity.addressBookFile101().thatMethod()`

hedera-mirror-rest/entityId.js
```diff
@@ -265,16 +265,26 @@ const parseFromString = (id, error) => {
   return [shard, realm, BigInt(numOrEvmAddress), null];
 };
 
+const parseIdParameterToEntityId = (id) => {
```
xin-hedera, line 268: the main issues with the function replacing EntityId.parse is

- lack of error check
- not cached
- it assumes the id should be of from `shard.realm.num`, otherwise just prepends shard / realm

I'm also not sure why we need it, or why `EntityId.parse` has to be replaced in various places.

`EntityId.parse` does append system shard if it's a `realm.num` string. If the value to parse is a number, we have to treat it as encoded (which contains shard, realm, and num).

xin-hedera (changes requested):

steven-sheehy (changes requested):

hedera-mirror-rest/accounts.js
xin-hedera: perhaps we can just get rid of using `0.0.0` as the default anchor account since it's only needed when there's accounts returned from the query and then the anchor account is set to the last in the array.

hedera-mirror-rest/entityId.js
jnels124: EntityId.parse is being called when you do things like `/api/v1/contract/{id}' if id=0.0.27 everything parses correctly. If id=27 then EntityId.parse is treating 27 as the encoded id but that is not correct.  This method calls parse. It just ensures to call it with shard.realm.num form every time.

jnels124: The openapi specs indicate the value passed this way is the entitynum not the encoded id.

hedera-mirror-rest/__tests__/entityId.test.js
jnels124: it is accepted by contracts endpoints. But those endpoints dont actually use this method. Will look at this again

hedera-mirror-rest/topicmessage.js
jnels124: We use this method to get the entity id from the encoded id in the database

steven-sheehy: Maybe we should have them named `parseString()` and `parseNumber()` and assert the type in each? Then we only use former for user input and latter for db output?

jnels124: I was going to go with that approach too but was going to require more changes and testing etc. I can go with that approach though

steven-sheehy: Can just rename your `parseIdParameterToEntityId` to `parseString` and validate input as string. Should also verify non-negative numbers, number in range of shard/realm/num bits. Don't have to do the `parseNumber` change now.

hedera-mirror-rest/utils.js
jnels124: Can't access system entity here but will look at putting elsewhere

hedera-mirror-rest/entityId.js
xin-hedera: umm, I see. That's a bit controversial. Since the current implementation using `EntityId.parse` supports sth different: a number is treated as an encoded id. I think we should instead change `parseIdParameterToEntityId` to first append shard / realm if necessary, then delegate to `parse` for the remaining parsing.

hedera-mirror-rest/entityId.js
```diff
@@ -265,16 +265,26 @@ const parseFromString = (id, error) => {
...
+  const evmAddressOrNum = stripHexPrefix(idPieces[2]);
 
-  if (isEvmAddressAlias(evmAddress)) {
-    contractIdParts.create2_evm_address = evmAddress;
+  if (isEvmAddressAlias(evmAddressOrNum)) {
+    contractIdParts.create2_evm_address = evmAddressOrNum;
   } else {
-    contractIdParts.num = idPieces[2];
+    contractIdParts.shard = contractIdParts.shard || systemShard;
+    contractIdParts.realm = contractIdParts.realm || systemRealm;
```
xin-hedera, line 286: this would replace `shard` from the input with `systemShard` if `contractidParts.shard` is 0. believe if you change this line

```javascript
idPieces.unshift(...[null, null].slice(0, 3 - idPieces.length));
```

to

```javascript
idPieces.unshift(...[systemShard, systemRealm].slice(0, 3 - idPieces.length));
```

you'll no longer need the conditionals here.

hedera-mirror-rest/__tests__/entityId.test.js
jnels124: this seems to be present in several tests. I left it as it would require special logic to not handle that conversion and causes no issues I can see.

hedera-mirror-rest/topicmessage.js
jnels124: these validations happen in the call to parse

hedera-mirror-rest/entityId.js
xin-hedera: not sure. that's a valid concern. then you can fix it with

```javascript
  contractIdParts.shard = contractIdParts.shard ?? systemShard;
```

hedera-mirror-rest/accounts.js
```diff
@@ -366,10 +366,7 @@ const getAccounts = async (req, res) => {
     },
   };
 
-  let anchorAcc = '0.0.0';
-  if (ret.accounts.length > 0) {
-    anchorAcc = ret.accounts[ret.accounts.length - 1].account;
-  }
+  let anchorAcc = ret.accounts[ret.accounts.length - 1]?.account;
```
steven-sheehy, line 369: What happens when ret.accounts.length is zero?

hedera-mirror-rest/entityId.js
```diff
@@ -265,16 +265,40 @@ const parseFromString = (id, error) => {
...
 };
 
+const parseString = (id) => {
+  const pieces = piecesFromString(id);
+  return pieces && parse(pieces.filter((item) => item !== null).join('.'));
+};
+
+const piecesFromString = (id) => {
+  if (typeof id !== 'string') {
+    return null;
```
steven-sheehy, line 275: Shouldn't we throw an error her since this would be more of a programming error calling the wrong method?

hedera-mirror-rest/__tests__/entityId.test.js
steven-sheehy: So you're saying it's not possible via the actual API, only via EntityId?

jnels124: Not that I can see. It appears some restjava endpoints may allow it. Just wasn't sure if it should be removed from here or not. I can remove if that is preferred

hedera-mirror-rest/accounts.js
jnels124: then this gets set to undefined which is fine because it will not be used.

hedera-mirror-rest/entityId.js
jnels124: yes that is probably better

hedera-mirror-rest/entityId.js
```diff
@@ -336,10 +359,38 @@ const parse = (id, {allowEvmAddress, evmAddressType, isNullable, paramName} = {}
   return checkNullId(id, isNullable) || parseCached(`${id}`, allowEvmAddress, evmAddressType, error);
 };
 
+/**
+ * System Entities
+ * */
+const addressBookFile101 = of(systemShard, systemRealm, 101);
+const addressBookFile102 = of(systemShard, systemRealm, 102);
```
jnels124, line 366: preferable to export these as systemEntities object?, or individually as done?

hedera-mirror-rest/__tests__/entityId.test.js
steven-sheehy: Would prefer to remove and add logic to reject if input contains both '.' and '0x'. Later, we should fix rest-java to not allow.

hedera-mirror-rest/entityId.js
steven-sheehy: Think a separate SystemEntity object to match the other modules behavior, if it's easy enough.

hedera-mirror-rest/utils.js
```diff
@@ -197,9 +197,8 @@ const isValidSlot = (slot) => slotPattern.test(slot);
 
 const isValidValueIgnoreCase = (value, validValues) => validValues.includes(value.toLowerCase());
 
-const addressBookFileIdPattern = ['101', '0.101', '0.0.101', '102', '0.102', '0.0.102'];
 const isValidAddressBookFileIdPattern = (fileId) => {
-  return addressBookFileIdPattern.includes(fileId);
+  return EntityId.validAddressBookFileIds.includes(EntityId.parseString(fileId)?.getEncodedId());
```
xin-hedera, line 201: this function can be removed since it's no longer used.

hedera-mirror-rest/__tests__/entityId.test.js
jnels124: it will now reject that case and there is a test added

hedera-mirror-rest/entityId.js
```diff
@@ -265,16 +265,40 @@ const parseFromString = (id, error) => {
   return [shard, realm, BigInt(numOrEvmAddress), null];
 };
 
+const parseString = (id) => {
```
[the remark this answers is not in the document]
jnels124: was going to but none of the current calls used the options. Can add for future use though

hedera-mirror-rest/utils.js
jnels124: Forgot to remove that. Thanks.

xin-hedera (approved):
LGTM

steven-sheehy (approved):

PR #10795, opened 2025-04-01
Add tests for token custom fee modification precompiled functions

### LINKED ISSUE #10770: Unable to set fixed fee in hbar for fungible token
### Description

In order to set fixed fee in hbar for fungible token you have to leave the denomination token id unset, as stated in the documentation here: 
https://docs.hedera.com/hedera/sdks-and-apis/sdks/token-service/custom-token-fees

Image

However, leaving the denominationTokenId unset/null causes exception when mapping the fixed fees in the TokenReadableKVState class.

### Steps to reproduce

1.Persist fungible token
2.Associate fixed fee without denominationTokenId (or denominationTokenId set to null)
3.Call precompiled system contract for getting custom fees for the token. This will results in output like: 

Image

### Additional context

_No response_

### Hedera network

_No response_

### Version

v.0.126.0

### Operating system

None

sdimitrov9:
**Description**:

Add tests for the custom token fees modification precompiled functions.

**Related issue(s)**:

Fixes #10770 

**Checklist**

### DIFF
--- a/hedera-mirror-test/src/test/resources/solidity/contracts/PrecompileTestContract.sol
+++ b/hedera-mirror-test/src/test/resources/solidity/contracts/PrecompileTestContract.sol
@@ -68,7 +68,7 @@ contract PrecompileTestContract is HederaTokenService {
     {
         (int256 responseCode,IHederaTokenService.TokenInfo memory retrievedTokenInfo) = HederaTokenService.getTokenInfo(token);
         if (responseCode != HederaResponseCodes.SUCCESS) {
-            revert("Failed to get token info");
+            revert();
         }
         tokenInfo = retrievedTokenInfo;
     }
--- a/hedera-mirror-test/src/test/resources/solidity/contracts/PrecompileTestContract.sol
+++ b/hedera-mirror-test/src/test/resources/solidity/contracts/PrecompileTestContract.sol
@@ -68,7 +68,7 @@ contract PrecompileTestContract is HederaTokenService {
     {
         (int256 responseCode,IHederaTokenService.TokenInfo memory retrievedTokenInfo) = HederaTokenService.getTokenInfo(token);
         if (responseCode != HederaResponseCodes.SUCCESS) {
-            revert();
+            revert("Failed to get token info");
         }
         tokenInfo = retrievedTokenInfo;
     }
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/TokenReadableKVState.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/TokenReadableKVState.java
@@ -205,8 +205,7 @@ public class TokenReadableKVState extends AbstractReadableKVState<TokenID, Token
                     .get());
             final var denominatingTokenId = f.getDenominatingTokenId();
 
-            final var fixedFee =
-                    new FixedFee(f.getAmount(), denominatingTokenId == null ? null : toTokenId(denominatingTokenId));
+            final var fixedFee = new FixedFee(f.getAmount(), toTokenId(denominatingTokenId));
             var constructed = new CustomFee(
                     new OneOf<>(FeeOneOfType.FIXED_FEE, fixedFee), collector, f.isAllCollectorsAreExempt());
             fixedFees.add(constructed);
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/TokenReadableKVState.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/TokenReadableKVState.java
@@ -205,7 +205,8 @@ public class TokenReadableKVState extends AbstractReadableKVState<TokenID, Token
                     .get());
             final var denominatingTokenId = f.getDenominatingTokenId();
 
-            final var fixedFee = new FixedFee(f.getAmount(), toTokenId(denominatingTokenId));
+            final var fixedFee =
+                    new FixedFee(f.getAmount(), denominatingTokenId == null ? null : toTokenId(denominatingTokenId));
             var constructed = new CustomFee(
                     new OneOf<>(FeeOneOfType.FIXED_FEE, fixedFee), collector, f.isAllCollectorsAreExempt());
             fixedFees.add(constructed);
--- a/hedera-mirror-web3/src/main/java/com/hedera/services/utils/EntityIdUtils.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/services/utils/EntityIdUtils.java
@@ -189,6 +189,10 @@ public final class EntityIdUtils {
     }
 
     public static com.hedera.hapi.node.base.TokenID toTokenId(final EntityId entityId) {
+        if (entityId == null) {
+            return null;
+        }
+
         return com.hedera.hapi.node.base.TokenID.newBuilder()
                 .shardNum(entityId.getShard())
                 .realmNum(entityId.getRealm())
--- a/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
+++ b/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
@@ -2,6 +2,7 @@
 
 package com.hedera.mirror.web3.service;
 
+import static com.hedera.mirror.web3.evm.utils.EvmTokenUtils.toAddress;
 import static com.hedera.mirror.web3.service.model.CallServiceParameters.CallType.ETH_CALL;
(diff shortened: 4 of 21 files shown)


hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
```diff
@@ -292,27 +291,10 @@ protected Token fungibleTokenPersist() {
      * @param treasuryEntityId - the treasuryEntityId that has to be set in the token
      * @return Token object that is persisted in db
      */
-    protected Token fungibleTokenPersistWithTreasuryAccount(final EntityId treasuryEntityId) {
+    protected Token fungibleTokenPersist(final EntityId treasuryEntityId) {
         return fungibleTokenCustomizable(t -> t.treasuryAccountId(treasuryEntityId));
```
kselveliev, line 295: Probably we should revert this name changes since a lot of classes that are unrelated to the PR are changed due to it and it's pretty hard to track which are the most critical changes from the PR.

hedera-mirror-test/src/test/resources/solidity/contracts/PrecompileTestContract.sol
```diff
@@ -68,7 +68,7 @@ contract PrecompileTestContract is HederaTokenService {
     {
         (int256 responseCode,IHederaTokenService.TokenInfo memory retrievedTokenInfo) = HederaTokenService.getTokenInfo(token);
         if (responseCode != HederaResponseCodes.SUCCESS) {
-            revert();
+            revert("Failed to get token info");
```
bilyana-gospodinova, line 71: We should probably avoid doing such small changes in the test contracts in the hedera-mirror-test module. These contracts are used in the acceptance tests, deployed not regularly to mainnet and used by the k6 tests as well. If we get an error while testing with k6, there will be difference between the deployed contract and the message in the sources in here, which can cause confusion and make the issue root cause analysis harder. Also, this change is not relevant for the current PR. Please revert.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/TokenReadableKVState.java
```diff
@@ -205,7 +205,8 @@ private List<CustomFee> mapFixedFees(
                     .get());
             final var denominatingTokenId = f.getDenominatingTokenId();
 
-            final var fixedFee = new FixedFee(f.getAmount(), toTokenId(denominatingTokenId));
+            final var fixedFee =
+                    new FixedFee(f.getAmount(), denominatingTokenId == null ? null : toTokenId(denominatingTokenId));
```
bilyana-gospodinova, line 209: This null check should probably be in `toTokenId`.

hedera-mirror-test/src/test/resources/solidity/contracts/PrecompileTestContract.sol
bilyana-gospodinova: Also, the contracts in hedera-mirror-test are not compiled automatically with `web3j` as in the web3 module, so this change would require recompiling the contract and updating the existing artifact as well. Just fyi.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
sdimitrov9: Done

hedera-mirror-web3/src/main/java/com/hedera/services/utils/EntityIdUtils.java
```diff
@@ -189,6 +189,10 @@ public static com.hedera.hapi.node.base.TokenID toTokenId(final Long entityId) {
     }
 
     public static com.hedera.hapi.node.base.TokenID toTokenId(final EntityId entityId) {
+        if (entityId == null) {
```
bilyana-gospodinova, line 192: We should have a junit test for this as well.

sdimitrov9: Done

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallCustomFeesModificationTest.java
```diff
@@ -0,0 +1,601 @@
...
+
+    @AfterEach
+    void afterEach() {
+        mirrorNodeEvmProperties.setModularizedServices(isModularized);
+        mirrorNodeEvmProperties.setProperties(evmProperties);
+    }
+
+    @Test
+    void updateFungibleTokenFixedFeeInHBAR() throws Exception {
+        // 1.CREATE TOKEN WITH FIXED FEE IN HBAR
```
kselveliev, line 57: All of our integration tests follow this structure:
```
// Given
test data setup
// When
contract.call
// Then
assert result

```
I think we should do that also here. 
This does not really mean that we should remove all current code comments but we should try to keep to the above structure.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallCustomFeesModificationTest.java
```diff
@@ -0,0 +1,498 @@
...
+
+    @AfterEach
+    void afterEach() {
+        mirrorNodeEvmProperties.setModularizedServices(isModularized);
+        mirrorNodeEvmProperties.setProperties(evmProperties);
+    }
+
+    @Test
+    void updateFungibleTokenFixedFeeInHBAR() throws Exception {
+        // Given I create token with fixed fee in HBAR
```
sdimitrov9, line 56: I have left descriptive comments within the first test of the class to improve test comprehension and since the other tests are similar they only contain the given, when and then keywords as comments in the respective places, similar to the most of the other tests within the module.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallCustomFeesModificationTest.java
sdimitrov9: Done, see the below comments for more details.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallCustomFeesModificationTest.java
```diff
@@ -0,0 +1,484 @@
...
+        assertThat(updatedFixedFee.useCurrentTokenForPayment).isEqualTo(newFee.useCurrentTokenForPayment);
+        assertThat(updatedFixedFee.useHbarsForPayment).isEqualTo(newFee.useHbarsForPayment);
+        assertThat(updatedFixedFee.tokenId).isEqualTo(newFee.tokenId);
+        assertThat(updatedFixedFee.feeCollector).isEqualTo(getAddressFromEntityId(newCollector.toEntityId()));
+
+        verifyEthCallAndEstimateGas(updateFeesFunctionCall, modificationPrecompileTestContract, ZERO_VALUE);
+    }
+
+    @Test
+    void updateFungibleTokenFixedFeeInCustomToken() throws Exception {
```
bilyana-gospodinova, line 96: All of the update tests in this class follow this logic:
1. Persist the needed setup.
2. Make contract call 1 with `precompileTestContract.call_getCustomFeesForToken`.
3. Verify that the result is expected.
4. Make contract call 2 with `modificationPrecompileTestContract. call_updateFungibleTokenCustomFeesAndGetExternal*`.
5. Verify that the result is expected.

We already have test cases for 1, 2 and 3 in another class. We need only 1, 4 and 5 in here. With this the tests will be easier to read and will not be misleading if there is an issue with `precompileTestContract.call_getCustomFeesForToken` as the test will fail and naturally it would look like there is an issue with the update custom fees but the issue would be elsewhere. We don't need to make the same verifications in multiple tests.

sdimitrov9: You are right that 2 and 3 are supposed to be verified in another tests, however the goal of these 2 steps here is to verify that the setup in step 1 is correct, so if the test fails in the future it will be easier to see if the issues is coming from the setup or if it is an actual bug. In other words I would prefer to leave the tests the same way, because this will save us time investigating potential test failures in the future.

sdimitrov9: After discussion with the rest of the team, we decided to go without 2 and 3. Changes are now applied.

steven-sheehy (approved):
LGTM

PR #10822, opened 2025-04-03
Support legacy Transaction in block stream

### LINKED ISSUE #10552: Transform block transactions without signed transaction bytes
### Description

A block may contain transactions without signed transaction bytes, these currently do not get transformed correctly.

### Steps to reproduce

Example block on previewnet:
`000000000000000000000000000006005034.blk.gz `

### Additional context

_No response_

### Hedera network

previewnet

### Version

main

### Operating system

None

xin-hedera:
**Description**:

This PR adds support of legacy Transaction in block stream

- Get transaction id from transaction body parsed from `Transaction.bodyBytes` if set 
- Support transaction hash calculation when `Transaction.bodyBytes` and `Transaction.sigMap` are set

**Related issue(s)**:

Fixes #10552

**Notes for reviewer**:

The other combinations of deprecated fields will get rejected by consensus node with `INVALID_TRANSACTION` so we don't have to support.

Will find such a transaction and its block stream file, run transformer and make sure it passes and the transaction record matches what's in record streams.

*update*: created a transaction with Transaction.bodyBytes and Transaction.sigMap set in integration, verified block item transformer can handle it and the reconstructed transaction record / receipt match record streams.

**Checklist**


build.gradle.kts
```diff
@@ -92,7 +92,6 @@ dependencies {
         api("uk.org.webcompere:system-stubs-jupiter:2.1.7")
         api("org.web3j:core:4.12.2")
         api("tech.pegasys:jc-kzg-4844:1.0.0")
-        api("com.hedera.cryptography:hedera-cryptography-bls:0.1.1-SNAPSHOT")
```
xin-hedera, line 95: not needed, we only need SNAPSHOT maven repo in web3 build file so those dependencies can be found

steven-sheehy (approved):

PR #10861, opened 2025-04-08
HIP-1056 Add block stream metrics

### LINKED ISSUE #10742: HIP-1056 Block stream metrics
### Problem

The block file poller is missing some download metrics that record stream calculated.

### Solution

Copy metrics from `Downloader`:

* Add `hedera.mirror.importer.cloud.latency` to `BlockStreamPoller`
* Add `hedera.mirror.download.latency` to `BlockStreamPoller`
* Add `hedera.mirror.download.stream.verification` to `BlockStreamVerifier`
* Add `hedera.mirror.stream.close.latency` to `BlockStreamVerifier`
* Don't carry over `hedera.mirror.download.signature.verification`

### Alternatives

_No response_

filev94:
**Description**:
This PR adds metrics to BlockStreamVerifier and BlockStreamPoller:

- Added `hedera.mirror.importer.cloud.latency` to `BlockStreamPoller`
- Added `hedera.mirror.download.latency` to `BlockStreamPoller`
- Added `hedera.mirror.download.stream.verification` to `BlockStreamVerifier`
- Added `hedera.mirror.stream.close.latency` to `BlockStreamVerifier`

**Related issue(s)**:

Fixes #10742 

**Notes for reviewer**:

**Checklist**


hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/BlockStreamPoller.java
```diff
@@ -65,6 +99,12 @@ public void poll() {
                 var blockFile = blockFileReader.read(blockFileData);
                 blockFile.setNodeId(nodeId);
 
+                Instant cloudStorageTime = blockFileData.getLastModified();
+                Instant consensusEnd = Instant.ofEpochSecond(0, blockFile.getConsensusEnd());
+
+                cloudStorageLatencyMetric.record(Duration.between(consensusEnd, cloudStorageTime));
+                downloadLatencyMetric.record(Duration.between(consensusEnd, Instant.now()));
```
bilyana-gospodinova, line 106: By definition the `downloadLatencyMetric` is "The difference in time between the consensus time of the last transaction in the file and the time at which the file was downloaded and verified" but the verification is done some lines after this one and the time should be calculated after that. Please check this for the other metrics as well.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/BlockStreamPoller.java
```diff
@@ -35,6 +38,37 @@ final class BlockStreamPoller implements StreamPoller {
     private final BlockPollerProperties properties;
     private final StreamFileProvider streamFileProvider;
 
+    private final Timer cloudStorageLatencyMetric;
+    private final Timer downloadLatencyMetric;
+
```
kselveliev, line 43: Maybe we could create a class containing those metrics, and we could probably just inject it in the 3 different classes that use them `Downloader`, `BlockStreamPoller` and BlockStreamVerifier.
To avoid duplication.

filev94: I thought about that too. My concern was that the metrics might get mixed if its a singleton spring bean, because they can have different tags (Downloader -can have Record and Balance and Block classes have Block streamType). I can probably change the scope of the bean to Prototype and it might work but we'll have different beans for all the classes and we will need to manage the clean up process. I really don't want to create concrete classes in the constructor eg. new StreamMetrics(streamType). The other option would be factory pattern, but thought that was unnecessarily complex at this time. Maybe just creating helper methods in an Utility class would be best. Thoughts?

kselveliev: Got it. So in that case we should probably leave them as they are due to above concern.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/block/BlockStreamPoller.java
filev94: Good catch. I moved both the cloudStorageLatencyMetric and downloadLatencyMetric after the complete verification of the block. For the streamVerification in verifier - i am not sure there is a better way to handle it

steven-sheehy (approved):
LGTM

bilyana-gospodinova (approved):

xin-hedera (approved):
LGTM

Issue #10719, opened 2025-03-25
Adapt tests in web3 to support different shard and realm


filev94:
### Problem

We need to make sure that our unit/integration tests can work with shard and realm values different from 0.

### Solution

Make changes in DomainBuilder to support non-zero shard and realm - already implemented via changes for shard / realm in importer
Fix tests with hardcoded values for shard/realm or address, if any

steven-sheehy:
This will probably require 0.61 hedera app library first.

filev94:
> This will probably require 0.61 hedera app library first.

Yes, currently there is a provideNodeSelfIdProvider in hedera-app (DaggerExecutorComponent) that has an instance that provides the DEFAULT account

`public static final AccountID DEFAULT = newBuilder().build();`

and whenever the payer account  and the nodeAccount are compared, the shard and realm are different and throws exception.

steven-sheehy:
See https://github.com/hiero-ledger/hiero-mirror-node/pull/10874#discussion_r2035607472 for an area where we need to fix with this ticket.

steven-sheehy:
v0.61.2 library is merged to main which unblocks this ticket.

PR #10874, opened 2025-04-09
Bump hedera.app to version 0.61.1

### LINKED ISSUE #10873: Bump hedera.app to version 0.61.1
### Problem

There is a new version of the `hedera.app` - 0.61.1.

### Solution

Update the dependency and adapt the code.

### Alternatives

_No response_

bilyana-gospodinova:
**Description**:
Bump `hedera.app` to latest version - 0.61.1 and adapt the code for the dependency changes that come from it.

Fixes #10873

### DIFF
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/MirrorNodeState.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/MirrorNodeState.java
@@ -121,7 +121,8 @@ public class MirrorNodeState implements MerkleNodeState {
             10L,
             List.of(),
             Bytes.EMPTY,
-            List.of());
+            List.of(),
+            true);
     private static final Metrics NO_OP_METRICS = new NoOpMetrics();
 
     @PostConstruct
--- a/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
+++ b/hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
@@ -710,6 +710,23 @@ public abstract class AbstractContractCallServiceTest extends Web3IntegrationTes
         mirrorNodeEvmProperties.setProperties(propertiesMap);
     }
 
+    protected void persistRewardAccounts() {
+        domainBuilder
+                .entity()
+                .customize(e -> e.id(801L)
+                        .num(801L)
+                        .createdTimestamp(genesisRecordFile.getConsensusStart())
+                        .timestampRange(Range.atLeast(genesisRecordFile.getConsensusStart())))
+                .persist();
+        domainBuilder
+                .entity()
+                .customize(e -> e.id(800L)
+                        .num(800L)
+                        .createdTimestamp(genesisRecordFile.getConsensusStart())
+                        .timestampRange(Range.atLeast(genesisRecordFile.getConsensusStart())))
+                .persist();
+    }
+
     public enum KeyType {
         ADMIN_KEY(1),
         KYC_KEY(2),
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -49,7 +49,7 @@ dependencies {
         api("com.graphql-java-generator:graphql-java-client-runtime:2.9")
         api("com.graphql-java:graphql-java-extended-scalars:22.0")
         api("com.graphql-java:graphql-java-extended-validation:22.0")
-        api("com.hedera.hashgraph:app:0.61.0")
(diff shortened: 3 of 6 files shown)


kselveliev (dismissed):

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
```diff
@@ -710,6 +710,23 @@ protected void activateModularizedFlagAndInitializeState()
         mirrorNodeEvmProperties.setProperties(propertiesMap);
     }
 
+    protected void persistRewardAccounts() {
+        domainBuilder
+                .entity()
+                .customize(e -> e.id(801L)
```
xin-hedera, line 716: wouldn't `id(801L)` fail with non-zero realm? at least setting id to 801, with num = 801 and non-zero realm is wrong.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
```diff
@@ -710,6 +710,23 @@ protected void activateModularizedFlagAndInitializeState()
...
+                        .num(801L)
+                        .createdTimestamp(genesisRecordFile.getConsensusStart())
+                        .timestampRange(Range.atLeast(genesisRecordFile.getConsensusStart())))
+                .persist();
+        domainBuilder
+                .entity()
+                .customize(e -> e.id(800L)
+                        .num(800L)
+                        .createdTimestamp(genesisRecordFile.getConsensusStart())
+                        .timestampRange(Range.atLeast(genesisRecordFile.getConsensusStart())))
```
steven-sheehy, line 726: This doesn't properly handle non-zero realm. We should also avoid using hardcoded system IDs. Use `SystemEntity.stakingRewardAccount` and `nodeRewardAccount` and ensure shard and realm are set as well. You can also use `EntityId.toEntity()` if it's easier.

steven-sheehy (changes requested):

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
bilyana-gospodinova: I agree with this. However, this method was just moved from a child class into a parent and this change is not related to the bump to `hedera.app` 0.61.1 itself. Also, most of the tests have a similar setup with just a num and an id and this needs to be addressed for all tests in another PR that is focused just on fixing the shard/realm setup.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/MirrorNodeState.java
```diff
@@ -121,7 +121,8 @@ public class MirrorNodeState implements MerkleNodeState {
             10L,
             List.of(),
             Bytes.EMPTY,
-            List.of());
+            List.of(),
+            true);
```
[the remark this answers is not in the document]
bilyana-gospodinova: We will raise it as in the upcoming meeting with the services team.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
steven-sheehy: Added to existing #10719. We can handle it there.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
bilyana-gospodinova: I agree but I explained that this code was just moved from one place to another in the comment here - https://github.com/hiero-ledger/hiero-mirror-node/pull/10874/files#r2035607472. I can add the shard/realm in this method but this needs to be done on multiple places as well for the setup of the other tests too and I think this needs a separate PR.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
xin-hedera: sounds good. we should be able to catch such issues in the dedicated ticket and fix it there.

steven-sheehy (approved):

xin-hedera (approved):
LGTM

PR #10892, opened 2025-04-11
HIP-551 Add rest changes for batch transactions

### LINKED ISSUE #10515: HIP-551 Enhance `/api/v1/transactions/{id}`
### Problem

Need to query all inner transactions contained in a batch

### Solution

- When querying a transaction by transaction id, we need to do an additional query to retrieve all inner transactions.

### Alternatives

_No response_

jnels124:
**Description**:
This pr updates the rest api to support batch transactions defined by hip 551

* update transaction by transaction id to include inner transactions
* include batch key on all transaction responses
* updates parser to correctly populate parent for inner transactions 

**Related issue(s)**:

Fixes #10515 

**Notes for reviewer**:

**Checklist**


jnels124:
| transactionsId | /transactions/0.0.8270952-1742869224-651776719 | 1500 | 100.00 | 2409.74/s | 2409.74/s | 583.91ms | No | New code with new id. No inners |
| transactionsId | /transactions/0.0.8270952-1742869224-651776719 | 1500 | 99.95 | 1850.77/s | 1849.84/s | 787.99ms | No | New code with new id. With inners |
| transactionsId | /transactions/0.0.1082181-1725646545-819106593 | 1500 | 100.00 | 2839.95/s | 2839.95/s | 465.51ms | No | Old id new Code|

hedera-mirror-rest/transactions.js
```diff
@@ -786,18 +784,70 @@ const extractSqlFromTransactionsByIdOrHashRequest = async (transactionIdOrHash,
...
+  if (transactionKeys.length % 2 !== 0) {
+    throw new InvalidArgumentError('transaction keys must be in sequenced pairs of [payer,timestamp,payer,timestamp]');
+  }
+
+  for (let index = 0; index < transactionKeys.length; index += 2) {
+    let paramIndex = params.length;
+    const payer = transactionKeys[index];
+    const timestamp = BigInt(transactionKeys[index + 1]);
+
+    if (payer) {
```
xin-hedera, line 815: when will this evaluate to `false`?

hedera-mirror-rest/transactions.js
```diff
@@ -786,18 +784,70 @@ const extractSqlFromTransactionsByIdOrHashRequest = async (transactionIdOrHash,
...
+      payerAccountParams.push(`$${++paramIndex}`);
+      idConditions.push(`(${Transaction.PAYER_ACCOUNT_ID} = $${paramIndex} and 
+                               ${timestampField} = $${++paramIndex})`);
+      params.push(payer);
+      includePayer = transactionKeys;
```
xin-hedera, line 820: includePayer is assigned but never used otherwise

hedera-mirror-rest/transactions.js
```diff
@@ -786,18 +784,70 @@ const extractSqlFromTransactionsByIdOrHashRequest = async (transactionIdOrHash,
...
+    } else {
+      idConditions.push(`${timestampField} = $${++paramIndex}`);
+    }
+    params.push(timestamp);
+
+    if (!minTimestamp || timestamp < minTimestamp) {
```
xin-hedera, line 826: initialize minTimestamp to `constants.MAX_LONG`, then simplify this to
```suggestion
    if (timestamp < minTimestamp) {
```

hedera-mirror-rest/transactions.js
```diff
@@ -786,18 +784,70 @@ const extractSqlFromTransactionsByIdOrHashRequest = async (transactionIdOrHash,
...
+      minTimestamp = timestamp;
+    }
   }
-  mainConditions.unshift(...commonConditions);
+  const maxConsensusTimestamp =
+    timestampField === Transaction.CONSENSUS_TIMESTAMP
+      ? transactionKeys[transactionKeys.length - 1]
+      : minTimestamp + maxTransactionConsensusTimestampRangeNs;
```
xin-hedera, line 833: we need to track `maxTimestamp` and change this to
```suggestion
      : maxTimestamp + maxTransactionConsensusTimestampRangeNs;
```

hedera-mirror-rest/transactions.js
```diff
@@ -786,18 +784,70 @@ const extractSqlFromTransactionsByIdOrHashRequest = async (transactionIdOrHash,
...
+const getTransactionsByTransactionIdsSql = (transactionKeys, filters, timestampField) => {
+  let payerAccountParams = [];
+  let params = [];
+
+  let minTimestamp = 0;
+  let idConditions = [];
+  let includePayer = false;
+
+  if (transactionKeys.length % 2 !== 0) {
+    throw new InvalidArgumentError('transaction keys must be in sequenced pairs of [payer,timestamp,payer,timestamp]');
```
xin-hedera, line 807: nit,
```suggestion
    throw new InvalidArgumentError('transaction keys must be in sequenced pairs of [payer,timestamp]');
```

hedera-mirror-rest/transactions.js
```diff
@@ -852,6 +902,22 @@ const getTransactionsByIdOrHash = async (req, res) => {
     throw new NotFoundError();
   }
 
+  const innerTransactions = rows
+    .map((row) => row.inner_transactions)
+    .filter((innerTransactions) => innerTransactions)
+    .flat();
```
xin-hedera, line 908: no flattening needed, since it's now a dimensional array.

```suggestion
```

hedera-mirror-rest/transactions.js
```diff
@@ -852,6 +902,22 @@ const getTransactionsByIdOrHash = async (req, res) => {
...
+
+  if (!isTransactionHash && innerTransactions.length > 0) {
```
xin-hedera, line 910:
```suggestion
  const innerTransactions = rows.inner_transactions;
  
  if (!isTransactionHash && innerTransactions?.length > 0) {
```

xin-hedera (changes requested):

hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/transaction/BlockItem.java
```diff
@@ -89,13 +89,17 @@ private StateChangeContext createStateChangeContext() {
...
 
     private BlockItem parseParent() {
-        // set parent, parent-child items are assured to exist in sequential order of [Parent, Child1,..., ChildN]
         if (parentConsensusTimestamp != null && previous != null) {
             if (parentConsensusTimestamp == previous.consensusTimestamp) {
                 return previous;
             } else if (previous.parent != null && parentConsensusTimestamp == previous.parent.consensusTimestamp) {
                 // check older siblings parent, if child count is > 1 this prevents having to search to parent
                 return previous.parent;
+            } else if (previous.parent != null
```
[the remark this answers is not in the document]
jnels124: T1p will have a parent consensus timestamp = atomic batch consensus timestamp. I was going to traverse the ancestors till a match was made but there isn't really a good stopping point. Not that we should ever have to traverse far but was still hesitant to do it

hedera-mirror-rest/transactions.js
jnels124: when transaction hash without payer account id is called. They all share the same generator (this method)

hedera-mirror-rest/transactions.js
jnels124: oops meant to remove that.

hedera-mirror-rest/transactions.js
jnels124: do we though? For a batch, the batch is going to have some validstartns so all inner transactions will have to execute with in the batch.validstart + (35 min or whatever the max is).

hedera-mirror-rest/transactions.js
jnels124: this is collecting the array of inner transactions (from potentially many rows - though only one should ever have inner transactions currently) so think result of the map will be an [][]. Will double check.

hedera-mirror-rest/transactions.js
```diff
@@ -211,6 +214,12 @@ const formatTransactionRows = async (rows) => {
       valid_duration_seconds: utils.getNullableNumber(row.valid_duration_seconds),
       valid_start_timestamp: utils.nsToSecNs(validStartTimestamp),
     };
+
+    if (row.batch_key) {
+      formatted.batch_key = utils.toHexString(row.batch_key, true);
+    }
```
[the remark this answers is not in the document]
jnels124: It was stated that it would only be included in the response for an inner transaction. Maybe I mistook that. Will update

hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/transaction/BlockItem.java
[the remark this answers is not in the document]
xin-hedera: I see. Tried to look it up in the HIP, however obviously the diagram got lost in the repo.

hedera-mirror-rest/transactions.js
xin-hedera: isn't that we first get payer + consensus timestamp from `transaction_hash` table given a transaction hash? We need the payer here so the query only goes to one shard.

hedera-mirror-rest/transactions.js
xin-hedera: I don't think so. The only relevant description in the HIP I can find is:

```
Inner Transactions, by design:
...
4. Are deduplicated on an individual basis. If two inner transactions within the same batch have the same TransactionID, the transaction will be rejected.
```

If they all have the same valid start timestamp, with the rule above, a payer can't have more than 1 inner transaction in the batch, since otherwise multiple inner transactions end up with the same transaction id.

xin-hedera: I see what you mean by we can just use batch.validstartns + a max range.. btw, why can' we just store payer + consensus timestamp of all inner transactions?

hedera-mirror-rest/transactions.js
jnels124: yes we do exactly that but there are rows in transaction hash that do not have the payer account id. We have existing tests to validate that case as well.

hedera-mirror-rest/transactions.js
jnels124: We use valid start so that we can capture any child/preceding transactions for an inner whenever we are querying for things. We could use consensusTimestamp but would need to store more data in the array in the db.

hedera-mirror-rest/transactions.js
xin-hedera: umm, sounds like that is for v1. At least in v1, there might be rows with null pay_account_id.

hedera-mirror-rest/__tests__/transactions.test.js
```diff
@@ -592,6 +624,200 @@ describe('extractSqlFromTransactionsByIdOrHashRequest', () => {
...
+          query: getTransactionIdQuery('nonce = $7 and scheduled = $8', defaultTransactionKeys),
+          params: addTimestampIndices([...defaultParams, 3, false], 4, 5),
+          scheduled: false,
+          isTransactionHash: false,
+        },
+      },
+    ];
+
+    for (const testSpec of testSpecs) {
+      test(testSpec.name, async () => {
```
jnels124, line 800: will remove async call here

steven-sheehy (approved):

xin-hedera (approved):
LGTM

PR #10899, opened 2025-04-11
HIP-1081 Block Node Support Design

### LINKED ISSUE #10737: HIP-1081: Design block node support
### Problem

With block stream integration mostly complete, we can now turn our attention to block node support.

### Solution

* Create a design doc in markdown format
* Breakout individual tasks

### Alternatives

_No response_

xin-hedera:
**Description**:

This PR adds block node support design.

**Related issue(s)**:

Fixes #10737 

**Notes for reviewer**:

**Checklist**


docs/design/block-node-support.md
```diff
@@ -0,0 +1,170 @@
...
+    public void poll();
+}
+```
+
+### CompositeBlockStreamPoller
+
+`CompositeBlockStreamPoller` delegates to either `BlockFilePoller` or `BlockNodePoller` to poll blocks, depending on
+the configuration and the delegates' health.
+
+When `sourceType` is `BOTH`, `CompositeBlockStreamPoller` will first use `BlockNodePoller`. When one poller is
```
jnels124, line 160: Do we need special handling to detect a CN upgrade so we don't switch pollers just because there is an upgrade in progress?

jnels124 (approved):

docs/design/block-node-support.md
```diff
@@ -0,0 +1,170 @@
...
+```
+
+### BlockPollerProperties
+
+Add new configuration properties to the existing class
+
+```java
+public class BlockPollerProperties {
+    ...
+    private Collection<BlockNodeAddresses> blockNodes = Collections.emptyList();
```
steven-sheehy, line 74: I would like to see this at `hedera.mirror.importer.block.nodes`. We should probably move away from downloader terminology as we may later refactor our code to stream between data directly from grpc to db.

docs/design/block-node-support.md
```diff
@@ -0,0 +1,170 @@
...
+- Rename `BockFileReader` to `BlockStreamReader`, and `ProtoBlockFileReader` to `ProtoBlockStreamReader`
+
+### Interfaces and Classes
+
+### BlockNodeAddress
+
+```java
+public class BlockNodeAddress {
+    private String host;
+    private int port;
```
steven-sheehy, line 53: Any consideration for TLS? Is it assumed that block nodes will only be TLS or only plaintext? Or if they do use TLS it will be via CA signed certificates so it will work without extra cert fields?

docs/design/block-node-support.md
```diff
@@ -0,0 +1,170 @@
...
+}
+```
+
+### SourceType
+
+```java
+public enum SourceType {
```
steven-sheehy, line 60: We already have a SourceType enum

```suggestion
public enum BlockSourceType {
```

docs/design/block-node-support.md
```diff
@@ -0,0 +1,170 @@
...
+
+Note: The diagram amends the block streams data flow with block node support
+
+## Importer
+
+- Add new configuration properties to `BlockPollerProperties` to support block nodes
+- Add an abstract class `AbstractBlockStreamPoller`
+- Move block file downloading in `BlockStreamPoller` to a new class `BlockFilePoller`
+- Add new class `BlockNodePoller` to support streaming blocks from block nodes
+- Rename `BockFileReader` to `BlockStreamReader`, and `ProtoBlockFileReader` to `ProtoBlockStreamReader`
```
steven-sheehy, line 44:
```suggestion
- Rename `BlockFileReader` to `BlockStreamReader`, and `ProtoBlockFileReader` to `ProtoBlockStreamReader`
```

docs/design/block-node-support.md
```diff
@@ -0,0 +1,170 @@
...
+                call onBlockStream
+                return
+            catch error and continue to next node
+    }
+}
+```
+
+### BlockNodePoller
+
+`BlockNodePoller` streams blocks from a list of block nodes in round-robin. At any time, `BlockNodePoller` should only
```
steven-sheehy, line 136: We shouldn't use `poll` terminology for connecting to block nodes either in child class or in any parent. Prefer `BlockNodeSubscriber`. Maybe for parent classes/interface `StreamSource` or `BlockStreamSource`.

docs/design/block-node-support.md
```diff
@@ -0,0 +1,170 @@
...
+communicate with one block node. Upon any error during streaming and block processing, `BlockNodePoller` should clean up
+resources created for the connection with the block node and throw exception so `CompositeBlockStreamPoller` can
+maintain health status and switch to `BlockFilePoller` if needed.
```
steven-sheehy, line 139: I don't think we should be switching back and forth between file and node. Once we transition away from buckets we should never need to invoke a s3 API again. So we should have logic to automatically migrate between the two (marker file again?), but not go back and forth. Perhaps we should add the sourceType to the table so it's clear the next file is the same block node source?

docs/design/block-node-support.md
steven-sheehy, line 136: Probably we should mention it's round-robin starting at first entry in the config. And that it's assumed the earliest entries are preferred due to costs or latency reasons.

docs/design/block-node-support.md
```diff
@@ -0,0 +1,170 @@
...
+streaming and fail fast if none of the block nodes have the next block.
+
+It's possible that a streaming response from a block node only contains part of a block, `BlockNodePoller` needs to
+combine block items in the same block and only then calls `onBlockStream`.
+
+```java
+public class BlockNodePoller extends AbstractBlockStreamPoller {
+
+    @Override
+    public void poll();
```
steven-sheehy, line 151: Again, name should not be poll() for streaming as it's the exact opposite of polling.

docs/design/block-node-support.md
```diff
@@ -0,0 +1,170 @@
...
+}
+```
+
+### CompositeBlockStreamPoller
+
+`CompositeBlockStreamPoller` delegates to either `BlockFilePoller` or `BlockNodePoller` to poll blocks, depending on
+the configuration and the delegates' health.
+
+When `sourceType` is `BOTH`, `CompositeBlockStreamPoller` will first use `BlockNodePoller`. When one poller is
+unhealthy, e.g., it fails to download one block 5 times consecutively, `CompositeBLockStreamPoller` will switch to the
```
steven-sheehy, line 161: Again, should only work in one direction, IMO. file -> block node

docs/design/block-node-support.md
steven-sheehy, line 151: We should document the algorithm to connect including looping over each node and how we're going to at least for now stream items until we reach block proof. Then pass to onBlockStream. May also consider timeouts if we don't receive the expected items in a reasonable amount of time.

steven-sheehy (changes requested):

docs/design/block-node-support.md
xin-hedera: Not addressed in the HIP. I'll rename `BlockNodeAddress` to `BlockNodeProperties` so it becomes generic.

docs/design/block-node-support.md
```diff
@@ -0,0 +1,212 @@
...
+            call onBlockStream;
+    }
+}
+```
+
+`BlockNodeSubscriber` streams blocks from a list of block nodes in round-robin starting from the first node in the
+configuration. At any time, `BlockNodeSubscriber` should only communicate with one block node. `BlockNodeSubscriber`
+should set a timeout while waiting for block items of a block to complete. Upon any error during streaming and block
+processing, `BlockNodeSubscriber` should clean up resources created for the connection with the block node and throw
+exception so `CompositeBlockStreamSource` can maintain health status, and have `BlockNodeSubscriber` try the next block
```
steven-sheehy, line 179: How will CompositeBlockStreamSource maintain health status for block node subscriber? Shouldn't the looping over block nodes be added to the above pseudocode? Or how else would you track which block node to try next?

steven-sheehy (changes requested):

docs/design/block-node-support.md
```diff
@@ -0,0 +1,212 @@
...
+    private BlockSourceType sourceType = BlockSourceType.AUTO;
+}
+```
+
+This is renamed from `BlockPollerProperties`, with two new properties added.
+
+### BlockStream
+
+```java
+public record BlockStream(List<BlockItem> blockItems, String filename, Long nodeId) {}
```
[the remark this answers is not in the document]
xin-hedera: used in stateproof. we can drop the column when we remove stateproof alpha

docs/design/block-node-support.md
```diff
@@ -0,0 +1,212 @@
...
+}
+```
+
+### BlockSourceType
+
+```java
+public enum BlockSourceType {
+    AUTO,
+    BLOCK_NODE,
+    FILE
```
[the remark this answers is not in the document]
xin-hedera: yes, when block source is `FILE` and stream provider is local. extra work is needed if we want to save blocks streamed from block nodes and later load them by local provider.

docs/design/block-node-support.md
```diff
@@ -0,0 +1,212 @@
...
+
+`BlockStream` represents a block from a downloaded block file, or a block streamed from a block node. The `filename`
+is the downloaded block file's name, or generated from the block number with `.blk` suffix. `nodeId` indicates which
+node the block file is downloaded from, or `null` if data is streamed from a block node.
+
+### BlockStreamReader
+
+```java
+public interface BlockStreamReader {
+    BlockFile read(BlockStream blockStream);
```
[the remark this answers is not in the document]
steven-sheehy: I say we just don't support saving block file bytes. Probably all those downloader properties will go away anyway once record streams are migrated.

steven-sheehy (approved):

PR #10902, opened 2025-04-11
Support long zero address for token by id endpoint

### LINKED ISSUE #10799: Support long form EVM address in `/tokens/{id}` endpoint
### Problem

`/tokens/{id}` endpoint doesn't support long form EVM address and returns code 400. EVM users then have to convert the long form address to `shard.realm.num` form and it's less than optimal UX.

### Solution

Change the endpoint to accept long form EVM address. Note we should still reject create2 EVM address.

### Alternatives

_No response_

jnels124:
**Description**:
This pr modifies the rest api token by id endpoint to allow token id path param to be the long zero format

**Related issue(s)**:

Fixes #10799 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-mirror-rest/api/v1/openapi.yml
+++ b/hedera-mirror-rest/api/v1/openapi.yml
@@ -4388,13 +4388,13 @@ components:
       required: true
       description: Token id
       examples:
-        longZero:
+        tokenNumAlias:
           value: 0x000000000000000000000000000000000000046F
-        shardRealmLongZero:
+        shardRealmTokenNumAlias:
           value: 0.0.000000000000000000000000000000000000046F
-        accountNumOnly:
+        tokenNum:
           value: 1135
-        shardRealmAccountNum:
+        shardRealmTokenNum:
           value: 0.0.1135
       schema:
         $ref: "#/components/schemas/EntityId"
@@ -5012,8 +5012,8 @@ components:
         noValue:
           summary: --
           value: ""
-        longZero:
-          summary: Example of long address equals with no operator
+        tokenNumAlias:
+          summary: Example of token num alias equals with no operator
           value: 0000000000000000000000000000000000000064
         entityNumNoOperator:
           summary: Example of entityNum equals with no operator
--- a/hedera-mirror-rest/constants.js
+++ b/hedera-mirror-rest/constants.js
@@ -82,7 +82,7 @@ const EvmAddressType = {
   // can be either a NO_SHARD_REALM or OPTIONAL_SHARD_REALM
   ANY: 2,
   // long zero hedera account id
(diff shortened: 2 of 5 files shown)


mgoelswirlds (approved):
LGTM

jnels124:
Not sure if it makes sense to normalize the next links to be the shard.realm.num. It will be converted if the parameter is changed for the next page, but doesn't get converted if the value is the same for the next page.

hedera-mirror-rest/api/v1/openapi.yml
```diff
@@ -5003,6 +5012,9 @@ components:
         noValue:
           summary: --
           value: ""
+        longZero:
+          summary: Example of long address equals with no operator
```
steven-sheehy, line 5016: We should not call this long zero since for non-zero realm it's not all zeros. It's official name is account num alias as defined in HIP-583 so guess this would be token num alias.

hedera-mirror-rest/api/v1/openapi.yml
```diff
@@ -4387,6 +4387,15 @@ components:
       in: path
       required: true
       description: Token id
+      examples:
+        longZero:
```
steven-sheehy, line 4391: tokenNumAlias

hedera-mirror-rest/constants.js
```diff
@@ -81,6 +81,8 @@ const EvmAddressType = {
   OPTIONAL_SHARD_REALM: 1,
   // can be either a NO_SHARD_REALM or OPTIONAL_SHARD_REALM
   ANY: 2,
+  // long zero hedera account id
+  LONG_ZERO: 3,
```
steven-sheehy, line 85: NUM_ALIAS

steven-sheehy (changes requested):

hedera-mirror-rest/entityId.js
```diff
@@ -132,6 +132,9 @@ const isValidEvmAddress = (address, evmAddressType = constants.EvmAddressType.AN
   if (evmAddressType === constants.EvmAddressType.NO_SHARD_REALM) {
     return evmAddressRegex.test(address);
   }
+  if (evmAddressType === constants.EvmAddressType.LONG_ZERO) {
+    return (evmAddressRegex.test(address) || evmAddressShardRealmRegex.test(address)) && !isEvmAddressAlias(address);
```
jnels124, line 136: should we support `shard.real.numAlias` format given that the shard and realm is already encoded in the numalias?

steven-sheehy: Yes, it's weird but we support it in other APIs.

xin-hedera (approved):
LGTM

PR #10935, opened 2025-04-15
Adapt web3 integration tests to fully pass with shard/realm

### LINKED ISSUE #10719: Adapt tests in web3 to support different shard and realm

filev94:
**Description**:
This PR adapts the remaining tests in web3 module to pass with shard/realm

Notable changes
- tokenAllowance persist with correct values
- feeCollectorAccount correct setup in NativePrecompileTest
- Several places where EntityId.of(id) is used, which would not work with shard/realm

**Related issue(s)**:

Fixes #10719

**Notes for reviewer**:
This PR is dependent on this [PR](https://github.com/hiero-ledger/hiero-mirror-node/pull/10934). Currently I am comparing against the latter's branch. After the first is merged, i will rebase. This I am doing so whoever wants to test locally can do it and you can easily see the changes

**Checklist**


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/Utils.java
```diff
@@ -50,6 +51,10 @@ public static boolean isMirror(final byte[] address) {
             return false;
         }
 
+        if (Arrays.equals(address, Address.ZERO.toArray())) {
```
steven-sheehy, line 54: Shouldn't allocate an array every time. Use toArrayUnsafe or save as a static somewhere.

hedera-mirror-web3/src/main/java/com/hedera/node/app/service/evm/accounts/HederaEvmContractAliases.java
```diff
@@ -21,6 +22,10 @@ public static boolean isMirror(final byte[] address) {
             return false;
         }
 
+        if (Arrays.equals(address, Address.ZERO.toArray())) {
```
steven-sheehy, line 25: Same comment

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/Utils.java
filev94: Fixed

hedera-mirror-web3/src/main/java/com/hedera/node/app/service/evm/accounts/HederaEvmContractAliases.java
filev94: Fixed

steven-sheehy (approved):

jnels124 (approved):

PR #10964, opened 2025-04-16
Fix incorrect entity public key

### LINKED ISSUE #10904: Account with complex key unduly returned by `accounts?account.publickey={key}`
### Description

Calling this REST endpoint `https://mainnet-public.mirrornode.hedera.com/api/v1/accounts?account.publickey=0aa8e21064c61eab86e2a9c164565b4e7a9a4146106e0a6cd03a8c395a110e92` returns 
 about 100 accounts among which is `0.0.200`, which has a complex admin key (unrelated to the searched key) and hence should not match the query.

### Steps to reproduce

Call REST endpoint `https://mainnet-public.mirrornode.hedera.com/api/v1/accounts?account.publickey=0aa8e21064c61eab86e2a9c164565b4e7a9a4146106e0a6cd03a8c395a110e92`

### Additional context

_No response_

### Hedera network

mainnet

### Version

0.127.0

### Operating system

None

xin-hedera:
**Description**:

This PR fixes incorrect entity public key.

- Use empty string as sentinel value to clear public key in database when it can't get extracted from the key
- Add a database migration to fix existing incorrect public key 

**Related issue(s)**:

Fixes #10904 

**Notes for reviewer**:

It took 90 seconds for `entity` table, and 71 seconds for `entity_history` table in V2

```
UPDATE 136168
Time: 89959.835 ms (01:29.960)

...
UPDATE 3231
Time: 71165.985 ms (01:11.166)
```

Number of entities with incorrect public key per type

```
   type   | count
----------+--------
 ACCOUNT  |    330
 CONTRACT |    152
 FILE     |     78
 TOPIC    | 135569
 TOKEN    |     39
```

Majority of them are topics, and it's because we used to set key to empty byte array if it's not set in `ConsensusCreateTopic` transaction, and that cascades to an empty string as `publicKey` as returned by `DomainUtils.getPublicKey`. The migration script will correct those to `null`.

**Checklist**


hedera-mirror-common/src/main/java/com/hedera/mirror/common/domain/transaction/RecordItem.java
```diff
@@ -45,6 +46,7 @@
 @Builder(buildMethodName = "buildInternal")
 @AllArgsConstructor(access = PRIVATE)
 @CustomLog
+@ToString(onlyExplicitlyIncluded = true)
```
xin-hedera, line 49: only include `consensusTimestamp`in toString, this fixes the slowdown when debugging in IDEA

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.106.0__fix_entity_public_key.sql
```diff
@@ -0,0 +1,45 @@
...
+        -- ED25519 key - primitive, key list with one key, or 1/1 threshold key
+        regexp_match(hex_key, '^(1220|32240a221220|2a28080112240a221220)([A-Fa-f0-9]{64})$'),
+        array[null::text, null::text]
+      ) as key_info
+      from (select encode(key, 'hex') as hex_key) as i
+    ) as t) as correct_public_key,
+    lower(timestamp_range) as lower_timestamp
+  from entity_history
+)
+update entity_history
```
xin-hedera, line 39: for entity_history table, since there is no primary key, there might be risk that when there are multiple rows with the same `(id, lower(timestamp_range))`, the update becomes indeterministic. However, we don't have such rows in production database.

hedera-mirror-importer/src/main/resources/db/migration/v1/V1.106.0__fix_entity_public_key.sql
```diff
@@ -0,0 +1,45 @@
...
+      from (select encode(key, 'hex') as hex_key) as i
+    ) as t) as correct_public_key
+  from entity
+)
+update entity
+set public_key = correct_public_key
+from extracted
+where entity.id = extracted.id
+  and coalesce(correct_public_key, public_key) is not null
+  and coalesce(correct_public_key <> public_key, true);
```
xin-hedera, line 21: only update rows where correct_public_key and public_key are different

- at least one value is not null
- and the values are different when both are not null, or one is null (the coalesce `true` part)

steven-sheehy (dismissed):
LGTM

steven-sheehy (approved):

jnels124 (approved):
LGTM

PR #11004, opened 2025-04-24
Adapt state to check for new state keys in hedera-app

### LINKED ISSUE #11003: Modify mirror node state to throw an error if it sees a new state in hedera-app
### Problem

Currently hedera-app state contains many state implementations and singletons with different keys.
Right now there is no way to figure out if newer version of hedera-app contains states that need to be implemented.

### Solution

Adapt mirror node state implementation when doing it's startup state migrations to check if it gets a new state 
that is not implemented then throw an error.
This would indicate that we first need to add the new state implementation before we can upgrade the hedera-app library.

### Alternatives

_No response_

kselveliev:
**Description**:
This PR introduces helper methods and a `StateKeyRegistry` class that lists all supported state and singleton keys for the current hedera-app version. During schema initialization, we now validate that all keys are known.

If hedera-app is upgraded and introduces new state keys that aren't present in `StateKeyRegistry` or missing in `DefaultSingleton` list of keys, an `UnsupportedOperationException` will be thrown. This helps catch unimplemented state logic early when bumping hedera-app versions.

Fixes #11003

**Checklist**

### DIFF
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/StateKeyRegistry.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/StateKeyRegistry.java
@@ -8,24 +8,17 @@ import com.hedera.node.app.service.schedule.impl.schemas.V0570ScheduleSchema;
 import com.hedera.node.app.service.token.impl.schemas.V0490TokenSchema;
 import com.hedera.node.app.state.recordcache.schemas.V0490RecordCacheSchema;
 import com.hedera.node.app.state.recordcache.schemas.V0540RecordCacheSchema;
+import com.swirlds.state.spi.ReadableKVState;
+import jakarta.inject.Named;
+import java.util.Collection;
+import java.util.HashSet;
 import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
-public enum StateKeyRegistry {
-    ;
-    private static final Set<String> keys = Stream.of(
-                    // Implemented
-                    AccountReadableKVState.KEY,
-                    AirdropsReadableKVState.KEY,
-                    AliasesReadableKVState.KEY,
-                    ContractBytecodeReadableKVState.KEY,
-                    ContractStorageReadableKVState.KEY,
-                    FileReadableKVState.KEY,
-                    NftReadableKVState.KEY,
-                    TokenReadableKVState.KEY,
-                    TokenRelationshipReadableKVState.KEY,
-                    // Not implemented but not needed
+@Named
+public class StateKeyRegistry {
+    private static final Set<String> DEFAULT_IMPLEMENTATIONS = Stream.of(
                     V0490TokenSchema.STAKING_INFO_KEY,
                     V0490FileSchema.UPGRADE_DATA_KEY,
                     V0490RecordCacheSchema.TXN_RECORD_QUEUE,
@@ -40,10 +33,18 @@ public enum StateKeyRegistry {
                     V0490ScheduleSchema.SCHEDULES_BY_ID_KEY)
             .collect(Collectors.toSet());
 
-    public static boolean contains(final String stateKey) {
+    private final Set<String> stateKeys;
+
+    public StateKeyRegistry(final Collection<ReadableKVState<?, ?>> keyValues) {
+        this.stateKeys = keyValues.stream()
+                .map(ReadableKVState::getStateKey)
+                .collect(Collectors.toCollection(() -> new HashSet<>(DEFAULT_IMPLEMENTATIONS)));
+    }
+
+    public boolean contains(final String stateKey) {
         if (stateKey != null && stateKey.startsWith("UPGRADE_DATA")) {
-            return keys.contains(V0490FileSchema.UPGRADE_DATA_KEY);
+            return stateKeys.contains(V0490FileSchema.UPGRADE_DATA_KEY);
         }
-        return keys.contains(stateKey);
+        return stateKeys.contains(stateKey);
     }
 }
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/StateKeyRegistry.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/StateKeyRegistry.java
@@ -1,0 +1,48 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package com.hedera.mirror.web3.state.keyvalue;
+
+import com.hedera.node.app.service.file.impl.schemas.V0490FileSchema;
+import com.hedera.node.app.service.schedule.impl.schemas.V0490ScheduleSchema;
+import com.hedera.node.app.service.schedule.impl.schemas.V0570ScheduleSchema;
+import com.hedera.node.app.service.token.impl.schemas.V0490TokenSchema;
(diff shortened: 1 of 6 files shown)


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/StateKeyRegistry.java
```diff
@@ -0,0 +1,48 @@
...
+import com.hedera.node.app.service.schedule.impl.schemas.V0490ScheduleSchema;
+import com.hedera.node.app.service.schedule.impl.schemas.V0570ScheduleSchema;
+import com.hedera.node.app.service.token.impl.schemas.V0490TokenSchema;
+import com.hedera.node.app.state.recordcache.schemas.V0490RecordCacheSchema;
+import com.hedera.node.app.state.recordcache.schemas.V0540RecordCacheSchema;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class StateKeyRegistry {
```
sdimitrov9, line 15: We may consider replacing this class with an enum. This way we will achieve type safety by referencing predefined enum values and we will probably get rid of the most codacy issues related to this class.

kselveliev: Addressed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/StateKeyRegistry.java
```diff
@@ -0,0 +1,49 @@
...
+                    // Implemented
+                    AccountReadableKVState.KEY,
+                    AirdropsReadableKVState.KEY,
+                    AliasesReadableKVState.KEY,
+                    ContractBytecodeReadableKVState.KEY,
+                    ContractStorageReadableKVState.KEY,
+                    FileReadableKVState.KEY,
+                    NftReadableKVState.KEY,
+                    TokenReadableKVState.KEY,
+                    TokenRelationshipReadableKVState.KEY,
```
steven-sheehy, line 27: I think it would be better to encapsulate state creation in a single registry class. Instead of manually listing states we have implemented we can scan them from spring beans dynamically. Then we only need to list known ones with default implementations.

```java

@Named
public class StateRegistry {

    private static final Set<String> DEFAULT_IMPLEMENTATIONS = Set.of(...); // Only list non-beans

    private final Collection<SingletonState<?>> singletons;
    private final Collection<ReadableKVState<?,?>> keyValues;

    private final Map<String, Object> states = // populate with singletons and keyValues beans

    public Object lookup(StateDefinition definition) {
        var state = states.get(definition.stateKey());
        if (state != null) {
            return state;
        }

        if (!DEFAULT_IMPLEMENTATIONS.contains(definition.stateKey())) {
            throw new UnsupportedOperationException(...);
        }

        if (definition.queue()) {
            return new ConcurrentLinkedDeque<>();
        } else if (definition.singleton()) {
            return new DefaultSingleton();
        }

        return new ConcurrentHashMap<>();
    }
}
```

We'd have to adapt MirrorNodeState to work with ReadableKVState instead of just Map, but that also avoids allocating a map just to throw it away.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/StateKeyRegistry.java
```diff
@@ -0,0 +1,49 @@
...
+                    V0570ScheduleSchema.SCHEDULED_ORDERS_KEY,
+                    V0570ScheduleSchema.SCHEDULE_ID_BY_EQUALITY_KEY,
+                    V0570ScheduleSchema.SCHEDULED_USAGES_KEY,
+                    // Not implemented but needed
+                    V0490ScheduleSchema.SCHEDULES_BY_ID_KEY)
+            .collect(Collectors.toSet());
+
+    public static boolean contains(final String stateKey) {
+        if (stateKey != null && stateKey.startsWith("UPGRADE_DATA")) {
+            return keys.contains(V0490FileSchema.UPGRADE_DATA_KEY);
```
steven-sheehy, line 45: Why this special logic?

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/StateKeyRegistry.java
```diff
@@ -0,0 +1,49 @@
...
+import com.hedera.node.app.service.schedule.impl.schemas.V0570ScheduleSchema;
+import com.hedera.node.app.service.token.impl.schemas.V0490TokenSchema;
+import com.hedera.node.app.state.recordcache.schemas.V0490RecordCacheSchema;
+import com.hedera.node.app.state.recordcache.schemas.V0540RecordCacheSchema;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public enum StateKeyRegistry {
+    ;
```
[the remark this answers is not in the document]
kselveliev: It is required to end the empty enum constant section, otherwise we cannot compile the enum.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/StateKeyRegistry.java
kselveliev: Upgrade date keys is a bit more special in the format of `public static final String UPGRADE_DATA_KEY = "UPGRADE_DATA[FileID[shardNum=%d, realmNum=%d, fileNum=%d]]";`
And during state initialization it's checked 10 times for keys ranging from:
```
UPGRADE_DATA[FileID[shardNum=0, realmNum=0, fileNum=150]] -> 
...
UPGRADE_DATA[FileID[shardNum=0, realmNum=0, fileNum=159]]
```

Since we don't have custom implementation for it and it's probably irrelevant for our use I added this special logic.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/StateKeyRegistry.java
kselveliev: Applied the dynamic spring bean scanning suggestion.

kselveliev: I think we can work encapsulate state creation in a single registry class in a new task since it will require some more changes. 
Mainly related to **adapt MirrorNodeState to work with ReadableKVState instead of just Map**

Since by modifying the it we need to change` MirrorNodeState.getReadableStates` -> 
`else if (state instanceof Map map) {`
to:
`else if (state instanceof ReadableKVState kvState) {`

and then we have to put
`data.put(stateName, new MapReadableKVState(stateName, kvState));` instead of the Map

This would require changes to the constructor of `MapReadableKVState` from
`public MapReadableKVState(final String stateKey, final Map<K, V> backingStore) {`
to: 
`public MapReadableKVState(final String stateKey, final ReadableKVState<K, V> backingStore) {`

and then we have to do adaptions to test classes and retest the logic and research that we are not breaking anything.

I think this can be done but probably better be in a separate PR not to overcomplicate this one.

steven-sheehy (approved):

PR #11024, opened 2025-04-27
Split downloaded files by date into different folders

Neurone:
**Description**:

The mirror node is already able to read and parse files organized by folders representing days, but the organization of the files in such a way is up to the operators.

Example of already supported file organization

```shell
./data/streams/2025-04-21/recordstreams/record0.0.4/2025-04-21T23_22_12.797932000Z.rcd.gz 
./data/streams/2025-04-21/recordstreams/record0.0.4/sidecar/2025-04-21T23_22_10.184472599Z_01.rcd.gz 
```

The mirror node currently stores downloaded files in a single folder for each consensus node, and when downloading record files, signatures, and sidecars, this quickly makes file system management very difficult.

With this PR:

- With the file saving options enabled, the mirror node will separate files into different folders based on the day.
- The importer process will now stop with a graceful shutdown if the environment variable `HIERO_MIRROR_IMPORTER_DOWNLOADER_HALTONERROR` is set and there are errors saving the files (i.e. because the disk is full).

**Related issue(s)**:

Fixes N/A

**Notes for reviewer**:

During the development I noticed the warning [FromAsCasing: 'as' and 'FROM' keywords' casing do not match](https://docs.docker.com/reference/build-checks/from-as-casing/) when rebuilding the importer image, but this PR includes the fixes for the importer and the other images as well.

**Checklist**


hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/record/RecordDownloaderProperties.java
```diff
@@ -31,6 +31,8 @@ public class RecordDownloaderProperties implements DownloaderProperties {
 
     private boolean writeFiles = false;
 
+    private boolean writeSidecars = false;
```
steven-sheehy, line 34: I'm not convinced we need this property. The `hedera.mirror.importer.parser.record.sidecar.enabled` indicates both that it should be downloaded and parsed. That is, in general, it enables the use of sidecars on the mirror node. The `writeFiles` property then writes both record and sidecar files if sidecars are enabled. I don't see why someone would would want to download sidecars and write record files, but not write sidecar files.

Besides, sidecars are about to go away very soon with block streams so the less properties the better.

steven-sheehy (changes requested):
You probably already notice, but we're currently in the midst of renaming the top level folders for the first time in 5 years to migrate to hiero. So your PR might have difficulty merging with main.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/record/RecordDownloaderProperties.java
Neurone: The goal is to be able to download files (record and sidecars) and skip the parsing like I can do with record files. Currently, if I disable the parsing of the sidecars, there's no way to download them. The option is to give the users the same options they have for the files, and decoupling the download from the actual parsing.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/DownloaderProperties.java
```diff
@@ -25,6 +25,10 @@ public interface DownloaderProperties {
 
     void setWriteFiles(boolean keepFiles);
 
+    boolean isWriteSidecars();
+
+    void setWriteSidecars(boolean keepSidecars);
```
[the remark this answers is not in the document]
Neurone: Done.

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/Downloader.java
```diff
@@ -333,15 +333,17 @@ private boolean verifySignatures(Collection<StreamFileSignature> signatures) {
 
                 verify(streamFile, signature);
 
+                var archiveDestinationFolder = importerProperties
+                        .getStreamPath()
+                        .resolve(streamFileData.getFilename().substring(0, 10));
```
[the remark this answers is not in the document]
Neurone: I set up a `hedera.mirror.importer.downloader.groupByDay` option for this, I just set it as default = true :)

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/util/Utility.java
```diff
@@ -116,6 +116,7 @@ public static void archiveFile(String filename, byte[] contents, Path destinatio
             log.trace("Archived file to {}", destination);
         } catch (Exception e) {
             log.error("Error archiving file to {}", destination, e);
+            if (Boolean.parseBoolean(System.getProperty(HALT_ON_ERROR_PROPERTY))) System.exit(-1);
```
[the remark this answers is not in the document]
Neurone: I proposed using the `HIERO_MIRROR_IMPORTER_DOWNLOADER_HALTONERROR` env variable

hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/downloader/record/RecordDownloaderProperties.java
steven-sheehy: If your goal is to download only and skip all parsing, why can't you just set `hiero.mirror.importer.parser.record.enabled=false` to disable all parsing? Again, the property you added is only useful if you want to download sidecars and write record files, but not also write sidecar files.

Neurone: I set `hiero.mirror.importer.parser.record.enabled=false` indeed, but the current source code download sidecar files only if the parsing is enabled. If the parsing is disabled, the MN does not download the sidecars, and so I cannot save them.

```java
private void downloadSidecars(StreamFilename recordFilename, RecordFile recordFile, ConsensusNode node) {
  if (!sidecarProperties.isEnabled() || recordFile.getSidecars().isEmpty()) {
   return;
  }
  [...]
}
```

With this PR, the MN will still continue to not download the sidecars if parsing is disabled and isWriteSidecars is false. But if at least one of the two options is enabled (you want to parse, or you want to save) then the downloading is allowed.

Neurone: I introduced the variable `writeSidecars` because we already use `writeFiles` for the records and `writeSignatures` for the signatures. I thought it would be consistent to have a variable for the sidecars too, but I can change that to check `writeFiles` variable for both the records and the sidecars instead.

Neurone: I updated the PR to check the `writeFiles` variable in the `downloadSidecars` function, avoiding adding the new `writeSidecars` variable.

steven-sheehy (approved):
LGTM

xin-hedera (dismissed):
LGTM

Neurone:
@steven-sheehy @xin-hedera sorry guys, I pushed again to solve a conflict with a renamed Docker file, and this triggered the review process again.

steven-sheehy (approved):

PR #11062, opened 2025-04-29
Add airdrop to contract system contract tests

### LINKED ISSUE #10886: Add airdrop test similar to AirdropToContractSystemContractTest
### Problem

Currently there are some airdrop integration tests in the web3. We need to add more.

### Solution

Add test cases similar to AirdropToContractSystemContractTest in hiero-consensus-node.

### Alternatives

_No response_

filev94:
**Description**:
This PR adds tests similar to the ones in `AirdropToContractSystemContractTest` in hiero-consensus-node.

`ContractCallAirdropContractSystemContractTest.java` - adds new tests and parameterizes existing ones
`AbstractContractCallServiceTest.java` - added contractPersistCustomizable method

**Related issue(s)**:

Fixes #10886

**Checklist**


hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallAirdropToContractSystemContractTest.java
```diff
@@ -0,0 +1,804 @@
...
+        final var receiverContractEntityId = contractPersist(
+                receiverContractAddress, AssociateContract.BINARY, e -> e.maxAutomaticTokenAssociations(1));
+
+        final var tokenId = fungibleTokenSetup(sender);
+        tokenAccountPersist(tokenId, receiverContractEntityId.getId());
+        final var tokenAddress = toAddress(tokenId).toHexString();
+
+        final var airdropContract =
+                testWeb3jService.deployWithoutPersistWithValue(Airdrop::deploy, DEFAULT_DEPLOYED_CONTRACT_BALANCE);
+        contractPersist(airdropContract.getContractAddress(), Airdrop.BINARY, e -> e.balance(DEFAULT_BALANCE));
```
nickeynikolovv, line 67: I think that you don't have to persist a contract if you deploy it

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallAirdropToContractSystemContractTest.java
```diff
@@ -0,0 +1,804 @@
...
+                tokenAccountPersist(fungibleTokenId, receiverContractEntityId.getId());
+                tokenAccountPersist(nonFungibleTokenId, receiverContractEntityId.getId());
+            }
+
+            senders.add(getAddressFromEntity(sender));
+            receivers.add(receiverContractAddress);
+            serials.add(DEFAULT_SERIAL_NUMBER);
+        }
+
+        final var airdropContract = testWeb3jService.deployWithValue(Airdrop::deploy, MAX_DEPLOYED_CONTRACT_BALANCE);
```
kselveliev, line 183: deployWithValue internally would persist the contract but the line below also persists it
`contractPersist(airdropContract.getContractAddress(), Airdrop.BINARY, e -> e.balance(MAX_BALANCE));`
So we doing one unnecessary persist.

Also for all the methods in this class we can just use `testWeb3jService.deployWithValue(Airdrop::deploy, value)`since there is probably no need to custom persist.

hedera-mirror-web3/src/test/solidity/AssociateContract.sol
```diff
@@ -0,0 +1,24 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.0;
+
+import "./HederaTokenService.sol";
+import "./HederaResponseCodes.sol";
+
+contract AssociateContract is HederaTokenService {
+
+    function associateTokenToThisContract(address token) external {
```
kselveliev, line 9: Can we just use an empty contract for this tests. Since we are not calling the two methods of this contract?

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallAirdropToContractSystemContractTest.java
filev94: Unfortunately there is an issue if you have a contract that you manually persist and one that you normally deploy. If i do this change - the modularized flow doesn't even go to the TokenAirdropHandler but just immediately says the test passes. In the mono flow - it is not throwing exception. It seems that it wants both contracts to be either manually persisted or not, for some reason. The Empty contract needs to be manually persisted so i can control the maxAutomaticTokenAssociations, and in two tests i need to do the same for the airdropContract.

The change i can make is to replace deployWithValue with deployWithoutPersistWithValue.

hedera-mirror-web3/src/test/solidity/AssociateContract.sol
filev94: Renamed it to Empty.sol and removed the functions.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallAirdropToContractSystemContractTest.java
```diff
@@ -0,0 +1,804 @@
...
+        final var tokenAddresses = new ArrayList<String>();
+        for (int i = 0; i < 3; i++) {
+
+            final var tokenId = fungibleTokenSetup(sender);
+            final var tokenAddress = toAddress(tokenId).toHexString();
+
+            tokenAddresses.add(tokenAddress);
+        }
+
+        final var firstReceiverContract = testWeb3jService.deployWithoutPersist(AssociateContract::deploy);
```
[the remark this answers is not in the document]
filev94: I renamed it to Empty and removed the functions.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallAirdropToContractSystemContractTest.java
```diff
@@ -0,0 +1,804 @@
...
+        final var receiverContractAddress = receiverContract.getContractAddress();
+        final var receiverContractEntityId = contractPersist(
+                receiverContractAddress, AssociateContract.BINARY, e -> e.maxAutomaticTokenAssociations(1));
+
+        final var tokenId = fungibleTokenSetup(sender);
+        tokenAccountPersist(tokenId, receiverContractEntityId.getId());
+        final var tokenAddress = toAddress(tokenId).toHexString();
+
+        final var airdropContract =
+                testWeb3jService.deployWithoutPersistWithValue(Airdrop::deploy, DEFAULT_DEPLOYED_CONTRACT_BALANCE);
```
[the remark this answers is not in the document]
filev94: Check comment for explanation https://github.com/hiero-ledger/hiero-mirror-node/pull/11062#discussion_r2066518850

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallAirdropToContractSystemContractTest.java
```diff
@@ -0,0 +1,804 @@
...
+        // Then
+        if (mirrorNodeEvmProperties.isModularizedServices()) {
+            verifyEthCallAndEstimateGas(functionCall, airdropContract);
+        } else {
+            assertThrows(PrecompileNotSupportedException.class, functionCall::send);
+        }
+    }
+
+    @Test
+    @DisplayName("Can airdrop multiple tokens to a contract that is already associated to them")
```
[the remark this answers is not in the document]
filev94: Probably you have local changes with deployWithValue?

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallAirdropToContractSystemContractTest.java
kselveliev: Seems like we are not resetting the persistContract flag so we should keep the above changes

filev94: After discussion with Kris, seems like the issue is that the contractPersist flag in testweb3jService is set to false and the contract is not persisted when doing deployWithValue. For the sake of consistency, i prefer all the contracts to be deployWithoutPersist, furthermore in the last 2 tests it is required.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallAirdropToContractSystemContractTest.java
filev94: Check comment and the whole discussion for explanation https://github.com/hiero-ledger/hiero-mirror-node/pull/11062#discussion_r2066518850

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/AbstractContractCallServiceTest.java
```diff
@@ -639,6 +641,28 @@ protected void persistAirdropForNft(final Token token, final Entity sender, fina
                 .persist();
     }
 
+    protected void contractPersistCustomizable(
+            final String binary, final EntityId entityId, Consumer<Entity.EntityBuilder<?, ?>> customizer) {
+        final var contractBytes = Hex.decode(binary.replace(HEX_PREFIX, ""));
+        final var entity = domainBuilder
+                .entity(entityId)
+                .customize(e -> {
+                    e.type(CONTRACT).alias(null).evmAddress(null).maxAutomaticTokenAssociations(-1);
```
kselveliev, line 650: `.maxAutomaticTokenAssociations(-1)` should be passed from the caller in customizer since it is a very special case.

kselveliev (approved):

bilyana-gospodinova:
In the future please split the tests in multiple PRs so that they are easier to review. In here the feedback for 1 or 2 of the tests is applicable in the other tests as well. This way when the issues are addressed in the first smaller PR, the follow-up PRs will be a lot cleaner and will be easier to write as well without the need to copy the feedback from the review multiple times.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallAirdropToContractSystemContractTest.java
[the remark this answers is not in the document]
bilyana-gospodinova: I agree with @nickeynikolovv and I think that we can reuse one of the existing contracts as well. On a real environment we will have actual contracts with actual functions, so I think an integration test will be more valuable if it mocks the real environment as close as possible. By doing this sometimes we uncover some unexpected bugs as well.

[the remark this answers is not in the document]
filev94: Then why not use the AssociateContract that is connected to these tests?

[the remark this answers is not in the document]
bilyana-gospodinova: How is it connected?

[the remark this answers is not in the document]
filev94: In the hapi tests in hiero-consensus-node this is the contract that is used to associate the tokens. Also an empty contract is used in some of the tests as well. If I put another contract here, someone might wonder why I am using DynamicEthCalls in these tests for example. If i have to reuse, it should be non other than the Airdrop contracts like ClaimAirdrop / CancelAirdop.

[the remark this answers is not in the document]
filev94: No, I won't be using it. Okay, got it. Addressed.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallAirdropToContractSystemContractTest.java
```diff
@@ -0,0 +1,805 @@
...
+        final var receiverContract = testWeb3jService.deployWithoutPersist(ClaimAirdrop::deploy);
+        final var receiverContractAddress = receiverContract.getContractAddress();
+        final var receiverContractEntityId = contractPersist(receiverContractAddress, ClaimAirdrop.BINARY, e -> {});
+
+        final var tokenId = fungibleTokenSetup(sender);
+        tokenAccountPersist(tokenId, receiverContractEntityId.getId());
+        final var tokenAddress = toAddress(tokenId).toHexString();
+
+        final var airdropContract =
+                testWeb3jService.deployWithoutPersistWithValue(Airdrop::deploy, DEFAULT_DEPLOYED_CONTRACT_BALANCE);
```
bilyana-gospodinova, line 65: Do we really need a separate test suite? This class is very similar to the existing `ContractCallAirdropSystemContractTest`, it uses almost the same setup, the same contract. The main difference is in the receiver address which in this suite is a contract address. Can't we reuse the existing suite and make the tests parameterized by the receiver address?

filev94: There are similarities, yes, but i think that merging both tests suites can bring more confusion rather than make things simpler. I opted for a different test suite because of better clarity and easier reviewing of code. Maybe it is possible to merge them, but i think it would become a bit more complex and that would require refactoring of `ContractCallAirdropSystemContractTest.` There would have to be more ternary operators to distinguish between the account receiver and contract receiver.

nickeynikolovv (approved):
LGTM

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallAirdropToContractSystemContractTest.java
```diff
@@ -0,0 +1,805 @@
...
+        if (mirrorNodeEvmProperties.isModularizedServices()) {
+            verifyEthCallAndEstimateGas(functionCall, airdropContract);
+        } else {
+            assertThrows(PrecompileNotSupportedException.class, functionCall::send);
+        }
+    }
+
+    @Test
+    @DisplayName("Airdropped token with custom fees to be paid by the contract receiver should be paid by the sender")
+    void airdropToContractCustomFeePaidBySender() {
```
bilyana-gospodinova, line 254: This test is exactly the same as `airdropToContractCustomFeePaidBySenderWithNetOfTransfers` with one boolean parameter difference - both can be merged in one parameterized test.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallAirdropToContractSystemContractTest.java
```diff
@@ -0,0 +1,805 @@
...
+        if (mirrorNodeEvmProperties.isModularizedServices()) {
+            verifyEthCallAndEstimateGas(functionCall, airdropContract);
+        } else {
+            assertThrows(PrecompileNotSupportedException.class, functionCall::send);
+        }
+    }
+
+    @Test
+    @DisplayName("Can airdrop multiple tokens to contract that has free auto association slots")
+    void airdropMultipleTokensToContractWithFreeSlots() {
```
bilyana-gospodinova, line 523: This test is the same as `airdropMultipleTokensToContractWithNoFreeSlots` with 1 parameter difference - can be parameterized.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallAirdropToContractSystemContractTest.java
```diff
@@ -0,0 +1,805 @@
...
+
+    @Test
+    @DisplayName("Airdrop with multiple senders and multiple contract receivers")
+    void airdropMultipleSenderMultipleContractReceivers() {
+        // Given
+        final var airdropContract =
+                testWeb3jService.deployWithoutPersistWithValue(Airdrop::deploy, MAX_DEPLOYED_CONTRACT_BALANCE);
+        final var airdropContractAddress = airdropContract.getContractAddress();
+        contractPersist(airdropContractAddress, Airdrop.BINARY, e -> e.balance(MAX_BALANCE)
+                .maxAutomaticTokenAssociations(-1));
```
bilyana-gospodinova, line 639: The value of "-1" can be extracted as a constant in a utility class with a comment that this is the value for unlimited auto associations.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallAirdropToContractSystemContractTest.java
```diff
@@ -0,0 +1,805 @@
...
+        return Stream.of(Arguments.of("fungible", true), Arguments.of("non-fungible", false));
+    }
+
+    @BeforeEach
+    void setUp() {
+        persistRewardAccounts();
+    }
+
+    @Test
+    @DisplayName("Can airdrop fungible token to a contract that is already associated to it")
```
sdimitrov9, line 52: When naming tests a common convention is to describe the expected outcome directly. In other words, a more suitable name would be "Airdrop fungible token to a contract that is already associated to it".
This style is more declarative and is used in many test frameworks like Junit for example. I would suggest to follow this naming approach across all the tests.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallAirdropToContractSystemContractTest.java
```diff
@@ -0,0 +1,805 @@
...
+        // When
+        final var functionCall = airdropContract.send_tokenAirdrop(
+                tokenAddress,
+                getAddressFromEntity(sender),
+                receiverContractAddress,
+                DEFAULT_TOKEN_AIRDROP_AMOUNT,
+                DEFAULT_WEI_VALUE);
+
+        // Then
+        if (mirrorNodeEvmProperties.isModularizedServices()) {
```
sdimitrov9, line 77: If this functionality is not supported for mono, I would assume that executing the whole test setup just to assert we are getting notSupportedException may be a waste of time and resources. We should consider updating the BeforeEach method the following way:

`    @BeforeEach
    void setUp() {
        assumeTrue(mirrorNodeEvmProperties.isModularizedServices(),
                "Skipping test: tested functionality is supported only for modularized services.");
        persistRewardAccounts();
    }`

This way the tests are going to be skipped/ignored when modularizedServices is false.
Following this approach, we will need only the verifyEthCallAndEstimateGas(functionCall, airdropContract); statement allowing us to remove the rest of the if statement (for all of the tests in the class).

kselveliev: I think it's valid to assert notSupportedException cases since they can be hit by the user until we are 100% on modularized.

sdimitrov9: In that case we can just extract the whole if statement in a separate method and use it across the tests in order to avoid repeating the same block of code at multiple places:

```
    protected <T extends Contract> void verifyEthCallAndEstimateGasForModularized(RemoteFunctionCall<TransactionReceipt> functionCall, T contract) {
        if (mirrorNodeEvmProperties.isModularizedServices()) {
        } else {
            assertThrows(PrecompileNotSupportedException.class, functionCall::send);
        }
    }
```

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallAirdropToContractSystemContractTest.java
filev94: Addressed

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallAirdropToContractSystemContractTest.java
filev94: Addressed

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallAirdropToContractSystemContractTest.java
filev94: Addressed

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallAirdropToContractSystemContractTest.java
filev94: Addressed

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallAirdropToContractSystemContractTest.java
filev94: I created 2 methods -> verify and verifyWithException. Currently left them in the test suite, not in the abstract parent.

sdimitrov9: I think the names should be more specific. verify and verifyWithException leave too much field for interpretation.

kselveliev (dismissed):

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallAirdropToContractSystemContractTest.java
filev94: Changed to verifyContractCall and verifyContractCallWithException, as discussed

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallAirdropSystemContractTest.java
```diff
@@ -2,198 +2,242 @@
...
+import org.junit.jupiter.params.provider.CsvSource;
+import org.junit.jupiter.params.provider.MethodSource;
+import org.web3j.protocol.core.RemoteFunctionCall;
+import org.web3j.protocol.core.methods.response.TransactionReceipt;
 
 class ContractCallAirdropSystemContractTest extends AbstractContractCallServiceTest {
 
-    private static final BigInteger ZERO_VALUE = BigInteger.ZERO;
+    private static final BigInteger DEFAULT_WEI_VALUE = BigInteger.ZERO;
+    public static final long DEFAULT_BALANCE = 100_000_000L;
```
bilyana-gospodinova, line 40: These constants can be private.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallAirdropSystemContractTest.java
```diff
@@ -2,198 +2,242 @@
...
+    public static final BigInteger DEFAULT_DEPLOYED_CONTRACT_BALANCE = BigInteger.valueOf(DEFAULT_BALANCE);
+    public static final BigInteger DEPLOYED_BALANCE_1_BILLION = BigInteger.valueOf(1_000_000_000L);
+    public static final BigInteger DEPLOYED_BALANCE_10_BILLION = BigInteger.valueOf(10_000_000_000L);
+    public static final int NO_MAX_AUTO_ASSOCIATIONS = 0;
+    public static final int UNLIMITED_AUTOMATIC_TOKEN_ASSOCIATIONS = -1;
+    public static final String ACCOUNT = "account";
```
bilyana-gospodinova, line 46: This constant and the next one can be group into an enum and this way the we can avoid String comparison and remove the need of the `receiverData()` method.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallAirdropSystemContractTest.java
filev94: Addressed

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallAirdropSystemContractTest.java
filev94: I saw there is an EntityType enum where the contract and the account are already present so opted to use that instead of creating a new one. Also, since there are other entries there, decided to keep the receiverData() with the idea not to have to make 11 changes in case a new receiver is introduced and i use EnumSource with names.

steven-sheehy (approved):

kselveliev (approved):

PR #11088, opened 2025-05-02
Add validation for smart contract sender accounts

### LINKED ISSUE #11079: Warnings of type `Category NODE payer account should have resulted in failure upstream` are logged in grafana on mainnet
### Description

There are a lot warnings in web3 for `isModularized:true` in grafana logs similar to that one: 
`2025-04-30 11:50:35.755	
2025-04-30T08:50:35.755Z  WARN tomcat-handler-1521285 c.h.m.w.c.GenericControllerAdvice Mirror EVM transaction error: Category NODE payer account should have resulted in failure upstream db34c1ef944883f0e5a2fc18b6c1978b088bd31d, detail: , data: , isModularized: true `

Observed on mainnet

### Steps to reproduce

1. Observe the grafana logs for mainnet

### Additional context

_No response_

### Hedera network

_No response_

### Version

v0.128.2

### Operating system

None

bilyana-gospodinova:
**Description**:
This PR enhances the validation of the sender account. Currently, we only check if it exists and is not deleted. Now we are going to check if the sender is a smart contract and fail softly with `PAYER_ACCOUNT_NOT_FOUND` response code as it would otherwise fail when reaching the consensus node.

All of the entity to account mapping implementation was moved from `AccountReadableKVState` to a new `AbstractAliasedAccountReadableKVState` which is now the parent of `AccountReadableKVState` and `AliasesReadableKVState`. In both these child classes we read the same entity in the DB but by different key. By caching the entity in both caches (aliases and accounts) we ensure that we will make only a single call to the DB and this way we will lighten the load on the DB.

Fixes #11079 

**Checklist**

### DIFF
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/config/ModularizedConfiguration.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/config/ModularizedConfiguration.java
@@ -3,6 +3,7 @@
 package com.hedera.mirror.web3.evm.config;
 
 import com.hedera.mirror.web3.evm.properties.MirrorNodeEvmProperties;
+import com.hedera.mirror.web3.state.CacheManager;
 import com.hedera.mirror.web3.state.components.NoOpMetrics;
 import com.hedera.node.app.config.ConfigProviderImpl;
 import com.hedera.node.app.ids.AppEntityIdFactory;
@@ -33,5 +34,10 @@ public class ModularizedConfiguration {
     @Bean
     public EntityIdFactory entityIdFactory() {
         return new AppEntityIdFactory(mirrorNodeEvmProperties.getVersionedConfiguration());
+    }
+
+    @Bean
+    public CacheManager cacheManager() {
+        return new CacheManager();
     }
 }
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/config/ModularizedConfiguration.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/config/ModularizedConfiguration.java
@@ -3,7 +3,7 @@
 package com.hedera.mirror.web3.evm.config;
 
 import com.hedera.mirror.web3.evm.properties.MirrorNodeEvmProperties;
-import com.hedera.mirror.web3.state.CacheManager;
+import com.hedera.mirror.web3.state.AliasedAccountCacheManager;
 import com.hedera.mirror.web3.state.components.NoOpMetrics;
 import com.hedera.node.app.config.ConfigProviderImpl;
 import com.hedera.node.app.ids.AppEntityIdFactory;
@@ -37,7 +37,7 @@ public class ModularizedConfiguration {
     }
 
     @Bean
-    public CacheManager cacheManager() {
-        return new CacheManager();
+    public AliasedAccountCacheManager aliasedAccountCacheManager() {
+        return new AliasedAccountCacheManager();
     }
 }
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/config/ModularizedConfiguration.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/config/ModularizedConfiguration.java
@@ -3,7 +3,6 @@
 package com.hedera.mirror.web3.evm.config;
 
 import com.hedera.mirror.web3.evm.properties.MirrorNodeEvmProperties;
-import com.hedera.mirror.web3.state.AliasedAccountCacheManager;
 import com.hedera.mirror.web3.state.components.NoOpMetrics;
 import com.hedera.node.app.config.ConfigProviderImpl;
 import com.hedera.node.app.ids.AppEntityIdFactory;
@@ -34,10 +33,5 @@ public class ModularizedConfiguration {
     @Bean
     public EntityIdFactory entityIdFactory() {
         return new AppEntityIdFactory(mirrorNodeEvmProperties.getVersionedConfiguration());
-    }
-
-    @Bean
-    public AliasedAccountCacheManager aliasedAccountCacheManager() {
-        return new AliasedAccountCacheManager();
     }
 }
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/TransactionExecutionService.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/TransactionExecutionService.java
@@ -51,7 +51,8 @@ public class TransactionExecutionService {
 
     private static final Duration TRANSACTION_DURATION = new Duration(15);
     private static final long CONTRACT_CREATE_TX_FEE = 100_000_000L;
-    private static final String SENDER_NOT_FOUND = "Sender account not found";
+    private static final String SENDER_NOT_FOUND = "Sender account not found.";
+    private static final String SENDER_IS_SMART_CONTRACT = "Sender account is smart contract.";
 
     private final AccountReadableKVState accountReadableKVState;
     private final AliasesReadableKVState aliasesReadableKVState;
@@ -200,12 +201,15 @@ public class TransactionExecutionService {
 
         if (isMirror(senderAddress)) {
             final var accountID = accountIdFromEvmAddress(senderAddress);
-            if (!accountReadableKVState.contains(AccountID.newBuilder()
+            final var account = accountReadableKVState.get(AccountID.newBuilder()
                     .accountNum(accountID.getAccountNum())
                     .shardNum(accountID.getShardNum())
                     .realmNum(accountID.getRealmNum())
-                    .build())) {
-                throwInvalidAccountIdException(SENDER_NOT_FOUND);
+                    .build());
+            if (account == null) {
+                throwPayerAccountNotFoundException(SENDER_NOT_FOUND);
+            } else if (account.smartContract()) {
+                throwPayerAccountNotFoundException(SENDER_IS_SMART_CONTRACT);
             }
 
             var entityId = DomainUtils.fromEvmAddress(senderAddress.toArrayUnsafe());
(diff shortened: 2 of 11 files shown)


bilyana-gospodinova:
Putting the PR to draft as I think I missed a case.

bilyana-gospodinova:
Added the missing case.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/TransactionExecutionService.java
```diff
@@ -195,33 +195,44 @@ private AccountID getSenderAccountID(final CallServiceParameters params) {
         if (params.getSender().canonicalAddress().isZero() && params.getValue() == 0L) {
             return EntityIdUtils.toAccountId(systemEntity.treasuryAccount());
         }
-
         final var senderAddress = params.getSender().canonicalAddress();
+        AccountID accountIDNum = getSenderAccountIDAsNum(senderAddress);
```
filev94, line 199: Why not final var as the others?

filev94 (dismissed):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/TransactionExecutionService.java
```diff
@@ -195,33 +195,44 @@ private AccountID getSenderAccountID(final CallServiceParameters params) {
...
+        return accountIDNum;
+    }
 
-        return address;
+    private AccountID getSenderAccountIDAsNum(final Address senderAddress) {
+        final var senderAccountID = accountIdFromEvmAddress(senderAddress);
+        AccountID accountIDNum;
+        if (!isMirror(senderAddress)) {
+            // If the address is an alias we need to first check if it exists and get the AccountID as a num.
+            accountIDNum = aliasesReadableKVState.get(convertAddressToProtoBytes(senderAddress));
```
steven-sheehy, line 216: Is there a way internally to cache the `account = accountReadableKVState.get(accountIDNum)` when we lookup the alias successfully? Right now we're effectively loading the same row from the db twice for aliases.

bilyana-gospodinova: Currently, I think there is no direct way to do such caching. Perhaps in another task and PR we can improve that. Should I open a new task?

kselveliev (dismissed):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/TransactionExecutionService.java
```diff
@@ -195,33 +195,44 @@ private AccountID getSenderAccountID(final CallServiceParameters params) {
...
-        final var address = aliasesReadableKVState.get(convertAddressToProtoBytes(senderAddress));
-        if (address == null) {
-            throwInvalidAccountIdException(SENDER_NOT_FOUND);
-        }
+        return accountIDNum;
+    }
 
-        return address;
+    private AccountID getSenderAccountIDAsNum(final Address senderAddress) {
+        final var senderAccountID = accountIdFromEvmAddress(senderAddress);
```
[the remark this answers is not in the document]
bilyana-gospodinova: Comment addressed.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/AccountReadableKVState.java
```diff
@@ -92,178 +62,19 @@ protected Account readFromDataSource(@Nonnull AccountID key) {
...
-                        long treasuryAccountId = systemEntity.treasuryAccount().getId();
-                        return accountBalanceRepository
-                                .findHistoricalAccountBalanceUpToTimestamp(entity.getId(), t, treasuryAccountId)
-                                .orElse(0L);
-                    } else {
-                        return 0L;
+                .map(entity -> {
+                    final var account = accountFromEntity(entity, timestamp);
+                    // Associate the account alias with this entity in the cache, if any.
+                    if (account.alias().length() > 0) {
```
bilyana-gospodinova, line 68: Not ideal. It would had been better if we could use `markRead` directly from the parent. However, it can be used for classes only of the same type. In the accounts state, we need to cache the alias and in the aliases state we need to cache the account id as well so this is why need to make this caching in the mapping function.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/service/TransactionExecutionService.java
bilyana-gospodinova: Added caching in this PR.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/AbstractAliasedAccountReadableKVState.java
```diff
@@ -0,0 +1,244 @@
...
+import com.hedera.pbj.runtime.io.buffer.Bytes;
+import com.hedera.services.utils.EntityIdUtils;
+import jakarta.annotation.Nonnull;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+
+public abstract class AbstractAliasedAccountReadableKVState<K, V> extends AbstractReadableKVState<K, V> {
```
bilyana-gospodinova, line 42: Nothing new in this class. The implementation from `AccountReadableKVState` was just moved here, so we can call the `accountFromEntity` method in both `AccountReadableKVState` and `AliasesReadableKVState`.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/config/ModularizedConfiguration.java
```diff
@@ -34,4 +35,9 @@ public ConfigProviderImpl configProvider() {
     public EntityIdFactory entityIdFactory() {
         return new AppEntityIdFactory(mirrorNodeEvmProperties.getVersionedConfiguration());
     }
+
+    @Bean
+    public CacheManager cacheManager() {
+        return new CacheManager();
```
kselveliev, line 41: Naming it `cacheManager() `without a qualifier makes Spring treat it as the default cache manager. Since we already define multiple named CacheManager beans (e.g. for contracts, tokens, etc.), this could cause ambiguity or unintended behavior in caching. Check EvmConfiguration beans, we probably should use some more specific bean name to avoid possible issues.

bilyana-gospodinova: Renamed to `AliasedAccountCacheManager`.

kselveliev (approved):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/AbstractAliasedAccountReadableKVState.java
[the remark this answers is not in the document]
bilyana-gospodinova: Actually, the purpose of `AbstractAliasedAccountReadableKVState` changed a lot during development and maybe we don't need such parent class at all anymore but more like a separate helper class that builds an `Account` from an `Entity`.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/AliasesReadableKVState.java
```diff
@@ -6,27 +6,63 @@
...
+        this.aliasedAccountCacheManager = aliasedAccountCacheManager;
     }
 
     @Override
     protected AccountID readFromDataSource(@Nonnull ProtoBytes alias) {
         final var timestamp = ContractCallContext.get().getTimestamp();
         final var entity = commonEntityAccessor.get(alias.value(), timestamp);
-        return entity.map(e -> toAccountId(e.getShard(), e.getRealm(), e.getNum()))
+        return entity.map(e -> {
+                    final var accountID = toAccountId(e.getShard(), e.getRealm(), e.getNum());
```
steven-sheehy, line 60: Remove and use existing `account.accountId()`

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/AliasesReadableKVState.java
```diff
@@ -6,27 +6,63 @@
...
+                    final var account = accountFromEntity(e, timestamp);
+                    // Put the account in the account num cache.
+                    aliasedAccountCacheManager.putAccountNum(accountID, account);
```
steven-sheehy, line 63: This is maybe for a follow up, but couldn't we make our KV state implementations extend from `WritableKVStateBase` since the class is stateless due to the use of `ContractCallContext`? Then here we could call `AccountReadableKVState.put(account.accountId(), account)`?

This would be a much cleaner approach to warm the other cache by using the put methods on its state instead of using this hacky AliasAccountCacheManager. It would also simplify the setup in `MirrorNodeState` and make the system easier to reason about if there was only one state hierarchy for both readable and writable. Not sure about circular references issues though.

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/config/ModularizedConfiguration.java
```diff
@@ -34,4 +35,9 @@ public ConfigProviderImpl configProvider() {
     public EntityIdFactory entityIdFactory() {
         return new AppEntityIdFactory(mirrorNodeEvmProperties.getVersionedConfiguration());
     }
+
+    @Bean
+    public AliasedAccountCacheManager aliasedAccountCacheManager() {
```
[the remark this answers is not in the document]
bilyana-gospodinova: Correct - removed.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/AliasesReadableKVState.java
bilyana-gospodinova: Comment addressed.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/AliasesReadableKVState.java
bilyana-gospodinova: We won't be able to call directly `AccountReadableKVState.put(account.accountId(), account)` from here because the template here will resolve to `put(ProtoBytes, AccountID)`. In order to call the `AccountReadableKVState` from here we will need to add it as a dependency using composition and this will make both classes coupled and this is what I tried to avoid by adding the `AliasAccountCacheManager`.

steven-sheehy: We can consider using a spring event based approach to decouple in the future. Still think we should also change our kv states to inherit from writable to be more self contained in the future.

steven-sheehy (approved):
LGTM

PR #11090, opened 2025-05-02
chore: Create SECURITY.md

jwagantall:


steven-sheehy (changes requested):
Why create a separate `SECURITY.md` in every repo when we can have a single `SECURITY.md` for the entire organization at https://github.com/hiero-ledger/.github? See https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/creating-a-default-community-health-file

jwagantall:
> Why create a separate `SECURITY.md` in every repo when we can have a single `SECURITY.md` for the entire organization at https://github.com/hiero-ledger/.github? See https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/creating-a-default-community-health-file

Thank you Steven, this security file is a pointer to https://github.com/hiero-ledger/governance/blob/main/SECURITY.md
The main Security file is going to live in the governance repo and all repos will point to it.

jwagantall:
> > Thank you Steven, this security file is a pointer to https://github.com/hiero-ledger/governance/blob/main/SECURITY.md The main Security file is going to live in the governance repo and all repos will point to it.
> > More information on this policy's baseline is presented in our LFDT's guidelines: https://lf-decentralized-trust.github.io/governance/governing-documents/security.html
> 
> Hi Jessica. I understand the purpose of the file and where it currently lives. My suggestion is to copy it from https://github.com/hiero-ledger/governance/blob/main/SECURITY.md to https://github.com/hiero-ledger/.github/blob/main/SECURITY.md. Since this is a special file in GitHub it will automatically show up in every repository in the organization if it's placed in the .github repository. This is what I did when setting up the prior https://github.com/hashgraph/.github/blob/main/SECURITY.md for the Hashgraph organization. For example, hedera-transaction-tool has no direct SECURITY.md but it shows up when [viewing](https://github.com/hashgraph/hedera-transaction-tool?tab=security-ov-file#readme) the repo. Placing it in github.com/hiero-ledger/.github reduces the maintenance burden for new repositories and in case the contents of the file ever need to change.
> 
> The actual contents of the https://github.com/hiero-ledger/.github/blob/main/SECURITY.md could be either:
> 
> 1. The full security.md text from the governance repo
> 2. A pointer file like in this PR
> 
> In my opinion, the first approach is better as it would provide a better user experience to copy the full security.md text to the .github repo so users don't have to click twice through the pointer file to read the text. But if there's some legal aspect that I don't understand why it can't be copied, then it's still better to do option two as it is effectively the same result to the end user as this PR.

Thank you so much for your reply! My apologies I understand what you mean now. 
Moving it into the .github repo will be easier indeed based on your explanation.

So, after moving the file .github repo, then there will be no need to create all the pointers for each repo, right? And should also cover new repos as they get created. 

I like that suggestion :)

steven-sheehy:
> So, after moving the file .github repo, then there will be no need to create all the pointers for each repo, right? And should also cover new repos as they get created.

That's correct. :) Relevant github docs: https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/creating-a-default-community-health-file

jwagantall:
> > So, after moving the file .github repo, then there will be no need to create all the pointers for each repo, right? And should also cover new repos as they get created.
> 
> That's correct. :) Relevant github docs: https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/creating-a-default-community-health-file

Thank you! let me work on that. My apologies for the back and forth

PR #11103, opened 2025-05-05
Remove the temporary workaround for ThrottleAccumulator

### LINKED ISSUE #10589: Revert the temporary workaround for the throttle from hedera-app
### Problem

PR https://github.com/hiero-ledger/hiero-mirror-node/pull/10547 introduces a temporary workaround that disables the deterministic throttle from hedera-app. When the issue is fixed in services, we need to use it and delete the copied code.

### Solution

Delete the copied code in the mentioned PR and use the fix.
Delete `ThrottleAccumulator`

### Alternatives

_No response_

filev94:
**Description**:
This PR removes the copied class `ThrottleAccumulator` and makes changes to SystemFileLoader so that the fileData for addressBook101 and addressBook102 is filled. The nodeDetails from addressBook102 are needed to set the value of a capacitySplit for an ingestThrottle that we currently do not use.

**Related issue(s)**:

Fixes #10589 

**Checklist**


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/SystemFileLoader.java
```diff
@@ -55,6 +59,8 @@ public class SystemFileLoader {
     @Getter(lazy = true)
     private final Map<FileID, SystemFile> systemFiles = loadAll();
 
+    private byte[] mockAddressBook;
```
steven-sheehy, line 62: Use lombok lazy getter and can directly store as Bytes
```suggestion
    @Getter(lazy = true, value = AccessLevel.PRIVATE)
    private final Bytes mockAddressBook = createMockAddressBook();
```

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/SystemFileLoader.java
```diff
@@ -160,5 +166,48 @@ private long maxExpiry() {
...
+                            nodeId)))
+                    .setNodeAccountId(AccountID.newBuilder().setAccountNum(nodeId))
+                    .setNodeCertHash(ByteString.copyFromUtf8("nodeCertHash"))
+                    .setRSAPubKey("rsa+public/key");
+            builder.addNodeAddress(nodeAddressBuilder.build());
+        }
+        return builder.build();
+    }
+
+    private Bytes getBytes(FileData fileData, long fileId) {
```
steven-sheehy, line 193: Remove this method since if we have actual address book data from fileDataRepository we should use that instead of loading the data, discarding it, then using the mock data.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/SystemFileLoader.java
```diff
@@ -160,5 +166,48 @@ private long maxExpiry() {
...
+            long nodeId = 3 + i;
+            NodeAddress.Builder nodeAddressBuilder = NodeAddress.newBuilder()
+                    .setIpAddress(ByteString.copyFromUtf8("127.0.0." + nodeId))
+                    .setPortno((int) nodeId)
+                    .setNodeId(nodeId)
+                    .setMemo(ByteString.copyFromUtf8(String.format(
+                            "%s.%s.%s",
+                            properties.getCommonProperties().getShard(),
+                            properties.getCommonProperties().getRealm(),
+                            nodeId)))
```
steven-sheehy, line 184: No need to populate the memo, nodeCertHash, or RSAPubKey

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/SystemFileLoader.java
filev94: Removed them

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/SystemFileLoader.java
filev94: If we have actual fileData, the fileBytes.length will be greater than 0 and it won't create the mockAddressBook and assign it to fileBytes. It will use the real data. I added an integration test to cover this scenario as well.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/SystemFileLoader.java
```diff
@@ -160,5 +166,48 @@ private long maxExpiry() {
         return Instant.now().getEpochSecond() + maxLifetime;
     }
 
+    @SuppressWarnings("deprecation")
+    private com.hederahashgraph.api.proto.java.NodeAddressBook addressBookMockup(int size) {
+
+        com.hederahashgraph.api.proto.java.NodeAddressBook.Builder builder =
+                com.hederahashgraph.api.proto.java.NodeAddressBook.newBuilder();
+        for (int i = 0; i < size; ++i) {
```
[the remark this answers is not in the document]
filev94: We dont, i actually left it like this because i thought i might need to create multiple or eventually create a helper class. Removed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/SystemFileLoader.java
```diff
@@ -160,5 +168,30 @@ private long maxExpiry() {
         return Instant.now().getEpochSecond() + maxLifetime;
     }
 
+    @SuppressWarnings("deprecation")
```
bilyana-gospodinova, line 171: Can't we use the `serviceEndpoint` field instead of the deprecated `ipAddress` and `portno`?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/SystemFileLoader.java
```diff
@@ -160,5 +168,30 @@ private long maxExpiry() {
...
+        builder.addNodeAddress(nodeAddressBuilder.build());
+        return builder.build().toByteArray();
+    }
+
+    private Bytes getBytes(FileData fileData, long fileId) {
+        byte[] fileBytes = fileData.getFileData();
+        boolean isAddressBookFile = fileId == systemEntity.addressBookFile101().getId()
+                || fileId == systemEntity.addressBookFile102().getId();
+        if (isAddressBookFile && (fileBytes == null || fileBytes.length == 0)) {
+            fileBytes = getMockAddressBook();
```
kselveliev, line 190: Seems like this change bypasses the original retry logic and aggregation query, which are essential for accurately reconstructing file contents from historical data. `getFileAtTimestamp` already handles aggregation. Observe `getFileAtTimestamp` query.

If it’s absolutely necessary to use mocks for those specific files (e.g., address books), we should probably add a targeted `if` condition in (loadWithRetry) to handle only those cases explicitly and not go to retryTemplate.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
```diff
@@ -351,6 +353,7 @@ private Map<String, String> buildTransactionProperties() {
         props.put("nodes.gossipFqdnRestricted", "false");
         props.put("tss.hintsEnabled", "false");
         props.put("tss.historyEnabled", "false");
+        props.put("executor.disableThrottles", "true");
```
steven-sheehy, line 356: nit: sort

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/SystemFileLoader.java
filev94: When the fileData is empty for addressBook it is parsed successfully and it doesn't retry. The other option i can think of is to get the file from the retryTemplate and check if it's addressBook and the contents.length == 0 and then return the genesisFile().

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/SystemFileLoader.java
steven-sheehy: My point is `fileDataRepository.getFileAtTimestamp()` returns an `Optional` so will only invoke `getBytes()` with a valid row from the database. Since it's an actual row, we should always use it and parse it. It's possible that row has a null or empty bytes, but that should not cause us to use the mock address book but to instead try the next timestamp in the retry loop.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/SystemFileLoader.java
kselveliev: Understood, can continue in the discussion above https://github.com/hiero-ledger/hiero-mirror-node/pull/11103#discussion_r2077193074

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/SystemFileLoader.java
filev94: When an addressBook has an empty bytes it parses successfully and it does not retry. If for some reason the fileData is null it gets the genesisFile(). For addressBook files eventually i could throw the exception InvalidFileException when bytes are empty/null to make it retry and after the retryTemplate is finished i could check the file's contents - if they are still empty then return the mock data

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/evm/properties/MirrorNodeEvmProperties.java
filev94: Addressed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/SystemFileLoader.java
filev94: Addressed. Also added shard and realm in the accountId for the nodeAccount just to be safe.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/SystemFileLoader.java
filev94: Added the filter and added parameterized integration tests to cover similar scenario for most files. I've left out the hapi permissions and network property as they are a bit different.

steven-sheehy (approved):

kselveliev (dismissed):

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/state/SystemFileLoaderIntegrationTest.java
```diff
@@ -46,6 +64,80 @@ class SystemFileLoaderIntegrationTest extends Web3IntegrationTest {
...
+                            .build())
+                    .build())
+            .build();
+
+    private static final ServicesConfigurationList NETWORK_PROPERTY = ServicesConfigurationList.newBuilder()
+            .addNameValue(Setting.newBuilder().setName("name").build())
+            .build();
+
+    private static final byte[] EMPTY_BYTES = new byte[0];
+    private static final int ADDRESS_BOOK_101 = 101;
```
bilyana-gospodinova, line 117: Why don't you use the constants from `SystemEntity` as `EntityId`s instead of defining them from scratch and creating the `EntityId`s from an int in the tests?

filev94: Addressed

steven-sheehy (approved):

kselveliev (approved):

bilyana-gospodinova (approved):

PR #11123, opened 2025-05-08
Add schedule KV state and tests

### LINKED ISSUE #10927: Add integration tests for ContractSignScheduleTest
### Problem

Currently our code does not have any HederaScheduleService tests so we need to add them.

### Solution

The class you can check in hiero-consensus-node is `ContractSignScheduleTest,` located in `package com.hedera.services.bdd.suites.contract.precompile.schedule` of test-clients module. We need to add tests similar to those.

### Alternatives

_No response_

kselveliev:
**Description**:
This PR adds `ScheduleReadableKVState`.
Adds repositories for `Schedule` and `TransactionSignature` entities.
Adds integrations tests for Schedule sign operations.
Copies schedule solidity contracts from consensus node repo.

`ScheduleRepository` 
`TransactionSignatureRepository` - also support for historical queries by consensus timestamp.
`TransactionSignatureRepositoryTest` - adds test for TransactionSignatureRepository methods.
`ScheduleReadableKVState` - gets schedule info.
`ContractCallContractSignScheduleTest` - adds integration tests for various scenarios for schedule sign
 .sol files are copied from consensus node

**Related issue(s)**:

Fixes #10927 

**Checklist**

### DIFF
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TransactionSignatureRepository.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TransactionSignatureRepository.java
@@ -1,0 +1,12 @@
+package com.hedera.mirror.web3.repository;
+
+import com.hedera.mirror.common.domain.entity.EntityId;
+import com.hedera.mirror.common.domain.transaction.TransactionSignature;
+import java.util.List;
+import org.springframework.data.repository.CrudRepository;
+import org.springframework.stereotype.Repository;
+
+@Repository
+public interface TransactionSignatureRepository extends CrudRepository<TransactionSignature, TransactionSignature.Id> {
+    List<TransactionSignature> findByEntityId(EntityId entityId);
+}
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TransactionSignatureRepository.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TransactionSignatureRepository.java
@@ -1,3 +1,5 @@
+// SPDX-License-Identifier: Apache-2.0
+
 package com.hedera.mirror.web3.repository;
 
 import com.hedera.mirror.common.domain.entity.EntityId;
@@ -9,4 +11,7 @@ import org.springframework.stereotype.Repository;
 @Repository
 public interface TransactionSignatureRepository extends CrudRepository<TransactionSignature, TransactionSignature.Id> {
     List<TransactionSignature> findByEntityId(EntityId entityId);
+
+    List<TransactionSignature> findByEntityIdAndConsensusTimestampLessThanEqual(
+            EntityId entityId, long consensusTimestamp);
 }
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/ScheduleReadableKVState.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/ScheduleReadableKVState.java
@@ -1,14 +1,17 @@
 package com.hedera.mirror.web3.state.keyvalue;
 
+import com.hedera.hapi.node.base.Key;
 import com.hedera.hapi.node.base.ScheduleID;
 import com.hedera.hapi.node.base.TransactionID;
 import com.hedera.hapi.node.scheduled.SchedulableTransactionBody;
 import com.hedera.hapi.node.state.schedule.Schedule;
 import com.hedera.hapi.node.transaction.TransactionBody;
 import com.hedera.mirror.common.domain.entity.Entity;
+import com.hedera.mirror.common.domain.entity.EntityId;
 import com.hedera.mirror.common.domain.entity.EntityType;
 import com.hedera.mirror.web3.common.ContractCallContext;
 import com.hedera.mirror.web3.repository.ScheduleRepository;
+import com.hedera.mirror.web3.repository.TransactionSignatureRepository;
 import com.hedera.mirror.web3.state.CommonEntityAccessor;
 import com.hedera.node.app.service.schedule.impl.schemas.V0490ScheduleSchema;
 import com.hedera.pbj.runtime.ParseException;
@@ -16,6 +19,8 @@ import com.hedera.pbj.runtime.io.buffer.Bytes;
 import com.hedera.services.utils.EntityIdUtils;
 import jakarta.annotation.Nonnull;
 import jakarta.inject.Named;
+import java.util.List;
+import java.util.stream.Collectors;
 
 @Named
 public class ScheduleReadableKVState extends AbstractReadableKVState<ScheduleID, Schedule> {
@@ -24,10 +29,14 @@ public class ScheduleReadableKVState extends AbstractReadableKVState<ScheduleID,
 
     private final CommonEntityAccessor commonEntityAccessor;
 
-    protected ScheduleReadableKVState(ScheduleRepository scheduleRepository, CommonEntityAccessor commonEntityAccessor) {
+    private final TransactionSignatureRepository transactionSignatureRepository;
+
+    protected ScheduleReadableKVState(ScheduleRepository scheduleRepository, CommonEntityAccessor commonEntityAccessor,
+            TransactionSignatureRepository transactionSignatureRepository) {
         super(V0490ScheduleSchema.SCHEDULES_BY_ID_KEY);
         this.scheduleRepository = scheduleRepository;
         this.commonEntityAccessor = commonEntityAccessor;
+        this.transactionSignatureRepository = transactionSignatureRepository;
     }
 
     @Override
@@ -58,16 +67,53 @@ public class ScheduleReadableKVState extends AbstractReadableKVState<ScheduleID,
                 .scheduleId(scheduleID)
                 .payerAccountId(EntityIdUtils.toAccountId(schedule.getPayerAccountId()))
                 .schedulerAccountId(EntityIdUtils.toAccountId(schedule.getCreatorAccountId()))
-                // up to this point it should be correct
                 .deleted(entity.getDeleted())
                 .memo(entity.getMemo())
-               // .adminKey(entity.getKey())
-
                 .scheduledTransaction(SchedulableTransactionBody.PROTOBUF.parse(Bytes.wrap(schedule.getTransactionBody())))
                 .originalCreateTransaction(TransactionBody.newBuilder().transactionID(TransactionID.newBuilder()
                                 .accountID(EntityIdUtils.toAccountId(schedule.getCreatorAccountId()))))
-                //.signatories(entity.getKey())
-                .signatories()
+                .signatories(getSignatories(schedule.getScheduleId()))
                 .build();
     }
+
+    private List<Key> getSignatories(final Long scheduleId) {
+        // Here we should also start taking into a account the timestamp and make historical query
+        final var signatures = transactionSignatureRepository.findByEntityId(EntityId.of(scheduleId));
+
+        return signatures.stream()
+                .map(signature -> {
+                    Key.KeyOneOfType keyType = fromProtoOrdinal(signature.getType());
+
+                    return switch (keyType) {
+                        // Might need to add logic for KEY_LIST and THRESHOLD_KEY check
+                        // private static void accumulateNewSignatories( in AbstractScheduleHandler
+                        case ED25519 -> Key.newBuilder()
+                                .ed25519(Bytes.wrap(signature.getPublicKeyPrefix()))
+                                .build();
+
+                        case ECDSA_SECP256K1 -> Key.newBuilder()
+                                .ecdsaSecp256k1(Bytes.wrap(signature.getPublicKeyPrefix()))
+                                .build();
+                        default -> throw new IllegalArgumentException("Unsupported key type in getSignatories: " + keyType);
+                    };
+                })
+                .collect(Collectors.toList());
+    }
+
+    /**
+     * Converts a proto ordinal to the corresponding Key.KeyOneOfType enum value.
+     *
+     * @param protoOrdinal the proto ordinal to convert
+     * @return the corresponding Key.KeyOneOfType enum value
+     * @throws IllegalArgumentException if the proto ordinal does not correspond to any Key.KeyOneOfType
+     */
+    private Key.KeyOneOfType fromProtoOrdinal(int protoOrdinal) {
+        for (Key.KeyOneOfType kind : Key.KeyOneOfType.values()) {
+            if (kind.protoOrdinal() == protoOrdinal) {
+                return kind;
+            }
+        }
+        throw new IllegalArgumentException("Unknown protoOrdinal: " + protoOrdinal);
+    }
+
 }
--- a/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/ScheduleReadableKVState.java
+++ b/hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/ScheduleReadableKVState.java
@@ -1,0 +1,73 @@
+package com.hedera.mirror.web3.state.keyvalue;
+
+import com.hedera.hapi.node.base.ScheduleID;
+import com.hedera.hapi.node.base.TransactionID;
+import com.hedera.hapi.node.scheduled.SchedulableTransactionBody;
+import com.hedera.hapi.node.state.schedule.Schedule;
+import com.hedera.hapi.node.transaction.TransactionBody;
+import com.hedera.mirror.common.domain.entity.Entity;
+import com.hedera.mirror.common.domain.entity.EntityType;
+import com.hedera.mirror.web3.common.ContractCallContext;
+import com.hedera.mirror.web3.repository.ScheduleRepository;
+import com.hedera.mirror.web3.state.CommonEntityAccessor;
+import com.hedera.node.app.service.schedule.impl.schemas.V0490ScheduleSchema;
+import com.hedera.pbj.runtime.ParseException;
+import com.hedera.pbj.runtime.io.buffer.Bytes;
+import com.hedera.services.utils.EntityIdUtils;
+import jakarta.annotation.Nonnull;
+import jakarta.inject.Named;
+
+@Named
+public class ScheduleReadableKVState extends AbstractReadableKVState<ScheduleID, Schedule> {
+
+    private final ScheduleRepository scheduleRepository;
+
+    private final CommonEntityAccessor commonEntityAccessor;
+
+    protected ScheduleReadableKVState(ScheduleRepository scheduleRepository, CommonEntityAccessor commonEntityAccessor) {
+        super(V0490ScheduleSchema.SCHEDULES_BY_ID_KEY);
+        this.scheduleRepository = scheduleRepository;
+        this.commonEntityAccessor = commonEntityAccessor;
+    }
+
+    @Override
+    protected Schedule readFromDataSource(@Nonnull ScheduleID key) {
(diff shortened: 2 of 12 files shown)


hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallContractSignScheduleTest.java
```diff
@@ -0,0 +1,445 @@
...
+    void signScheduleWithEcdsaKey() throws Exception {
+        // Given
+        final var contract = testWeb3jService.deploy(HRC755Contract::deploy);
+        final var scheduleEntity = persistScheduleEntity();
+        final var messageHash = getMessageHash(scheduleEntity);
+
+        final var keyPair = Keys.createEcKeyPair();
+        final var signatureMapBytes = getSignatureMapBytesEcdsa(messageHash, keyPair);
+
+        final var ecdsaKey = Key.newBuilder()
```
filev94, line 52: You can create methods that build the ecdsaKey, signerAccount and domainBuilder.schedule() inside this class since they are copied between 4 and 6 times and this will help with reducing the length of the tests as well.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallContractSignScheduleTest.java
```diff
@@ -0,0 +1,445 @@
...
+import lombok.RequiredArgsConstructor;
+import org.bouncycastle.jcajce.provider.digest.Keccak;
+import org.junit.jupiter.api.Test;
+import org.testcontainers.shaded.org.bouncycastle.asn1.x509.SubjectPublicKeyInfo;
+import org.web3j.crypto.ECKeyPair;
+import org.web3j.crypto.Keys;
+import org.web3j.utils.Numeric;
+
+@RequiredArgsConstructor
+class ContractCallContractSignScheduleTest extends AbstractContractCallServiceTest {
```
nickeynikolovv, line 40: You have duplication of 'Contract' in the class name

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallContractSignScheduleTest.java
```diff
@@ -0,0 +1,445 @@
...
+        // Given
+        final var contract = testWeb3jService.deploy(HRC755Contract::deploy);
+        final var scheduleEntity = persistScheduleEntity();
+        final var messageHash = getMessageHash(scheduleEntity);
+
+        final var keyPair = Keys.createEcKeyPair();
+        final var signatureMapBytes = getSignatureMapBytesEcdsa(messageHash, keyPair);
+
+        final var ecdsaKey = Key.newBuilder()
+                .setECDSASecp256K1(convertToCompressedPublicKey(keyPair.getPublicKey()))
```
nickeynikolovv, line 53: We already have method for compressing the key and returning bytes - AbstractContractCallServiceTest - getProtobufKeyECDSA

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallContractSignScheduleTest.java
```diff
@@ -0,0 +1,445 @@
...
+                .build()
+                .toByteArray();
+        final var signerAccount = accountEntityPersistCustomizable(e -> e.type(EntityType.ACCOUNT)
```
nickeynikolovv, line 56: we already have method for persisting account with specific or null  alias and evm address - AbstractContractCallServiceTest - persistAccountWithEvmAddressAndPublicKey

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallContractSignScheduleTest.java
```diff
@@ -0,0 +1,445 @@
...
+    @Test
+    void signScheduleWithEd25519Key() throws Exception {
+        // Given
+        final var contract = testWeb3jService.deploy(HRC755Contract::deploy);
+        final var scheduleEntity = persistScheduleEntity();
+
+        final var keyPairGenerator = KeyPairGenerator.getInstance(ED_25519);
+        final var keyPairEd = keyPairGenerator.generateKeyPair();
+        var publicKey = keyPairEd.getPublic();
+        var publicKeyCompressed = convertToCompressedPublicKey(publicKey);
```
nickeynikolovv, line 277: We have the same method as `convertToCompressedPublicKey ` in ContratCallsIsAuthorizedTest -> getProtobufKeyEd25519. Maybe it will be useful if  you can extract that method in AbstractContractCallServiceTest because it is duplicated now in the both suites

nickeynikolovv (changes requested):

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallContractSignScheduleTest.java
kselveliev: Comment addressed

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallContractSignScheduleTest.java
kselveliev: Comment addressed

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallContractSignScheduleTest.java
kselveliev: Comment addressed

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallContractSignScheduleTest.java
kselveliev: I used a new method here since `persistAccountWithEvmAddressAndPublicKey` only sets null for evm address but autogenerates alias(), but in my case I needed both to be null.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/ScheduleReadableKVState.java
```diff
@@ -0,0 +1,133 @@
...
+        final var scheduleId = EntityIdUtils.toEntityId(key);
+
+        final var timestamp = ContractCallContext.get().getTimestamp();
+        final var entity = commonEntityAccessor.get(scheduleId, timestamp).orElse(null);
+
+        if (entity == null || entity.getType() != EntityType.SCHEDULE) {
+            return null;
+        }
+
+        return scheduleRepository
```
bilyana-gospodinova, line 59: I am concerned that we have 2 calls to the DB for a single schedule entry. Can't we merge these calls in a single query that joins the tables and returns the required fields?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/ScheduleReadableKVState.java
```diff
@@ -0,0 +1,133 @@
...
+                .scheduledTransaction(
+                        SchedulableTransactionBody.PROTOBUF.parse(Bytes.wrap(schedule.getTransactionBody())))
+                .originalCreateTransaction(TransactionBody.newBuilder()
+                        .transactionID(TransactionID.newBuilder()
+                                .accountID(EntityIdUtils.toAccountId(schedule.getCreatorAccountId()))))
+                .signatories(getSignatories(schedule.getScheduleId(), timestamp))
+                .build();
+    }
+
+    private List<Key> getSignatories(final Long scheduleId, final Optional<Long> timestamp) {
```
bilyana-gospodinova, line 92: We can use a `Supplier` here and read the signatures lazily like we did in the `Account` model.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/repository/TransactionSignatureRepositoryTest.java
```diff
@@ -0,0 +1,55 @@
...
+
+    @Test
+    void findByEntityIdSuccessful() {
+        final var transactionSignature = domainBuilder.transactionSignature().persist();
+        assertThat(transactionSignatureRepository.findByEntityId(transactionSignature.getEntityId()))
+                .contains(transactionSignature);
+    }
+
+    @Test
+    void findByEntityIdNotSuccessful() {
```
bilyana-gospodinova, line 25: nit: Instead of "NotSuccessful" you can add how exactly the test is expected to fail.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/ScheduleReadableKVState.java
kselveliev: Worth discussing. We are already using a similar pattern with TokenReadableKVState where we first query the entity then query token table.

The current separate queries should be pretty fast in isolation and combining them would require introducing a new projection class, which brings some additional complexity.

I’d lean toward keeping the current approach unless we have clear evidence that it's significantly less efficient than a combined query.

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/repository/TransactionSignatureRepositoryTest.java
kselveliev: Comment addressed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TransactionSignatureRepository.java
```diff
@@ -0,0 +1,17 @@
...
+import com.hedera.mirror.common.domain.transaction.TransactionSignature;
+import java.util.List;
+import org.springframework.data.repository.CrudRepository;
+import org.springframework.stereotype.Repository;
+
+@Repository
+public interface TransactionSignatureRepository extends CrudRepository<TransactionSignature, TransactionSignature.Id> {
+    List<TransactionSignature> findByEntityId(EntityId entityId);
+
+    List<TransactionSignature> findByEntityIdAndConsensusTimestampLessThanEqual(
```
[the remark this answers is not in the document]
kselveliev: Comment addressed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/ScheduleReadableKVState.java
steven-sheehy: It does make sense from a performance standpoint since the entity and schedule row will be colocated on the same worker node. But schedule doesn't have a history table to use with timestamp so it's a bit awkward since entity does. Probably we can keep as is and explore in a follow up along with other ones like token.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/ScheduleReadableKVState.java
```diff
@@ -0,0 +1,133 @@
...
+        final var entityId = EntityId.of(scheduleId);
+        final var signatures = timestamp
+                .map(ts ->
+                        transactionSignatureRepository.findByEntityIdAndConsensusTimestampLessThanEqual(entityId, ts))
+                .orElseGet(() -> transactionSignatureRepository.findByEntityId(entityId));
+
+        return signatures.stream()
+                .map(signature -> {
+                    Key.KeyOneOfType keyType = fromProtoOrdinal(signature.getType());
```
steven-sheehy, line 101: Why are we mapping to Key type? This should be `SignaturePair.SignatureCase`. See `EntityRecordItemListener.insertTransactionSignatures`.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/ScheduleReadableKVState.java
```diff
@@ -0,0 +1,133 @@
...
+                    return switch (keyType) {
+                        case ED25519 ->
```
steven-sheehy, line 103: We should map to all types in SignatureCase.

steven-sheehy (changes requested):

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TransactionSignatureRepository.java
```diff
@@ -0,0 +1,26 @@
...
+
+package com.hedera.mirror.web3.repository;
+
+import com.hedera.mirror.common.domain.entity.EntityId;
+import com.hedera.mirror.common.domain.transaction.TransactionSignature;
+import java.util.List;
+import org.springframework.data.repository.CrudRepository;
+import org.springframework.stereotype.Repository;
+
+@Repository
```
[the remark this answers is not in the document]
kselveliev: Comment addressed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/ScheduleReadableKVState.java
```diff
@@ -0,0 +1,133 @@
...
+import com.hedera.services.utils.EntityIdUtils;
+import jakarta.annotation.Nonnull;
+import jakarta.inject.Named;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+@Named
+public class ScheduleReadableKVState extends AbstractReadableKVState<ScheduleID, Schedule> {
```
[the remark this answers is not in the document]
kselveliev: Comment addressed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/ScheduleReadableKVState.java
kselveliev: Correct. Comment addressed.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TransactionSignatureRepository.java
```diff
@@ -0,0 +1,26 @@
...
+    /**
+     * Retrieves transaction signatures for a specific entity ID, filtering by a consensus timestamp.
+     * This method returns all signatures for the given entity that were valid at or before the provided consensus timestamp.
+     *
+     * @param entityId           the ID of the entity to fetch transaction signatures for.
+     * @param consensusTimestamp the consensus timestamp to filter the signatures by.
+     * @return a list of `TransactionSignature` entities for the specified entity ID,
+     *         filtered by the consensus timestamp.
+     */
+    List<TransactionSignature> findByEntityIdAndConsensusTimestampLessThanEqual(
```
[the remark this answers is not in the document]
kselveliev: Should we do this a separate PR?

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/ScheduleReadableKVState.java
kselveliev: I see that It should also be possible to sign schedule with Thresholdkey and Key_List.
This is the signature accumulation logic in consensus node.
```
switch (key.key().kind()) {
    case ED25519, ECDSA_SECP256K1 -> {
        if (signingCryptoKeys.contains(key)) {
            signatories.add(key);
        }
    }
    case KEY_LIST -> key.keyListOrThrow()
            .keys()
            .forEach(k -> accumulateNewSignatories(signatories, signingCryptoKeys, k));
    case THRESHOLD_KEY -> key.thresholdKeyOrThrow()
            .keysOrThrow()
            .keys()
            .forEach(k -> accumulateNewSignatories(signatories, signingCryptoKeys, k));
}

```
But based on our transaction signature and looking at EntityRecordItemListener logic in importer.
I’m not really sure should how can we map to the last 2 types.
This is why I only checked Ed and ECDSA.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/repository/TransactionSignatureRepository.java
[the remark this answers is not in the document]
steven-sheehy: No, your changes are not performant without this change. It's just adding two sql files (v1 and v2) with two lines in each.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/ScheduleReadableKVState.java
```diff
@@ -0,0 +1,124 @@
...
+        return Suppliers.memoize(() -> {
+            final var entityId = EntityId.of(scheduleId);
+            final var signatures = timestamp
+                    .map(ts -> transactionSignatureRepository.findByEntityIdAndConsensusTimestampLessThanEqual(
+                            entityId, ts))
+                    .orElseGet(() -> transactionSignatureRepository.findByEntityId(entityId));
+
+            return signatures.stream()
+                    .map(signature -> {
+                        SignaturePair.SignatureCase signatureCase =
```
steven-sheehy, line 106: nit: Use var

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/ScheduleReadableKVState.java
steven-sheehy: Please check with consensus node team on what we should do. I think we should also confirm if ContractID key is needed here.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/ScheduleReadableKVState.java
kselveliev: Addressed

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/keyvalue/ScheduleReadableKVState.java
steven-sheehy: I reached out. Only simple keys can sign. But we also need to add contractID and delegatable contract ID.

steven-sheehy (dismissed):
LGTM

hedera-mirror-web3/src/test/java/com/hedera/mirror/web3/service/ContractCallContractSignScheduleTest.java
nickeynikolovv: The method `persistAccountWithEvmAddressAndPublicKey` sets  .evmAddress(evmAddress) and .alias(evmAddress), so if you pass null as `evmAddress` they both will be set to null. I think the two methods are the same

```
    protected Entity persistAccountWithEvmAddressAndPublicKey(byte[] evmAddress, byte[] publicKey) {
        return accountEntityPersistCustomizable(e -> e.alias(evmAddress)
                .evmAddress(evmAddress)
                .key(publicKey)
                .type(EntityType.ACCOUNT)
                .balance(DEFAULT_ACCOUNT_BALANCE));
    }
```

kselveliev: Right. Addressed.

nickeynikolovv (approved):
LGTM

filev94 (approved):

PR #11134, opened 2025-05-09
Set an implicit ContractID key on contracts with no key

### LINKED ISSUE #10988: DB migration needed for `null` contract keys
### Description

As part of fixing [this bug](https://github.com/hiero-ledger/hiero-consensus-node/issues/15990) in smart contracts the last bullet point remains where a migration is needed in the mirror node. For all contracts with a null key, a default key needs to be set, similar to the fix from smart contracts. The key needs to be created from `Contract{shard, realm, num}`. The contract transaction handlers should also be updated to set the contract ID if it's an empty key.

### Steps to reproduce

1. Find a contract with null contract key.
2. Call a function from that contract that creates a child transaction (such as mint token). In the inner child transaction the payer is set to be the contract. The exception is the one from [this issue](https://github.com/hiero-ledger/hiero-mirror-node/issues/10900). It will not be reproduced anymore after merging [PR](https://github.com/hiero-ledger/hiero-mirror-node/pull/10968) which is a temporary workaround. After creating a DB migration, this workaround can be removed.

### Additional context

_No response_

### Hedera network

_No response_

### Version

v.0.107.0

### Operating system

None

steven-sheehy:
**Description**:

* Disallow blank issues when reporting an issue
* Migrate existing contracts without a key to a `ContractID` key
* Set an implicit `ContractID` key on contracts with no key during create

**Related issue(s)**:

Fixes #10988

**Notes for reviewer**:

Only 9730 contracts impacted on mainnet.

**Checklist**

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,3 @@
+# SPDX-License-Identifier: Apache-2.0
+
+blank_issues_enabled: false
--- a/importer/src/main/java/org/hiero/mirror/importer/migration/ContractNullKeyMigration.java
+++ b/importer/src/main/java/org/hiero/mirror/importer/migration/ContractNullKeyMigration.java
@@ -1,0 +1,63 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package org.hiero.mirror.importer.migration;
+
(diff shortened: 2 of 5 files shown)


jnels124 (approved):
LGTM

xin-hedera (approved):
LGTM

bilyana-gospodinova (approved):

PR #11140, opened 2025-05-09
Fix hbar leak from ethereum feature tests

### LINKED ISSUE #11139: Fix ethereum.feature hbar leaking due to not deleting the signer account
### Problem

In ethereum.feature tests we generate a signer account that is hollow and at the end we don't delete it. That means that the hbars that are left in that account are gone. 

### Solution

Improve the flow to not lose hbars

### Alternatives

_No response_

nickeynikolovv:
**Description**:
In ethereum.feature tests we generate a signer account that is hollow and at the end we don't delete it. That means that the hbars that are left in that account are gone.
Because the account is hollow, in order to delete it we have to finalize it first. That's why in that PR a hollow account creation is removed and a real account with ECDSA key and alias is created. That way we are able to delete the account at the end of execution

Fixes #11139 

**Notes for reviewer**:

**Checklist**


nickeynikolovv:
Changes are added in another PR

PR #11151, opened 2025-05-12
Fix Rest API response cache handler

### LINKED ISSUE #11026: REST API response cache handler set incorrect http headers
### Description

There are several issues in REST API response cache handler

- when handling a client request with `if-none-match` header and the content doesn't change thus etag matches, expressjs HttpResponse.send sets status to 304 and removes body and `Content-Type` / `Content-Length` / `Transfer-Encoding` headers; this causes subsequent responses from the cache handler with incorrect `Content-Type` header
- `Content-Length` is set to the size of uncompressed data though the body is compressed and there is `content-encoding: gzip` response header
- gzip the response body when the client says no gzip by request header `accept-encoding: gzip;q=0,deflate,sdch`

### Steps to reproduce

check the description

### Additional context

_No response_

### Hedera network

testnet

### Version

v0.128.0

### Operating system

None

xin-hedera:
**Description**:

This PR fixes issues in Rest API response cache handler

- Handle response headers properly when data is cached from a 304 not modified response
- Set content length properly, compressed or uncompressed
- Use `negotiator` to determine whether to `gzip` response given `accept-encoding` from request
- Set headers properly for 304 response
- Set headers properly for HEAD request
- Add compress threshold to avoid larger output than input
- Refactor `CachedApiResponse` to encapsulate compression / decompression, and get compressed / uncompressed content length
- Add version suffix to cache key
- base64url encode gzipped content for smaller JSON stringified data footprint in cache
- Add integration tests
- Fix test module hanging because `Cache` redis client keeps retrying 

**Related issue(s)**:

Fixes #11026 

**Notes for reviewer**:

- why base64url (no padding, save another 1-2 bytes) encoding

```javascript
const obj = {data: Buffer.from(Array(10).fill(95))};
> JSON.stringify(obj).length;
65
> JSON.stringify({data: obj.data.toString('base64url')}).length;
25
```

- why set compress threshold at 150: grok said for string, output becomes smaller then input at 100-150 bytes 

**Checklist**


jnels124 (approved):

rest/model/cachedApiResponse.js
```diff
@@ -1,14 +1,37 @@
...
+      this.length = body.length;
+    }
+  }
+
+  getBody() {
+    return this.compressed ? Buffer.from(this.body, 'base64url') : this.body;
+  }
+
+  getLength() {
+    return this.compressed ? Math.floor((this.body.length * 3) / 4) : this.body.length;
```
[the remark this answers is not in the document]
xin-hedera: They are used in the negotiate function.

we need to save uncompressed length when body is compressed to avoid the expensive decompress to get length logic for HEAD requests

rest/middleware/responseCacheHandler.js
```diff
@@ -1,90 +1,135 @@
...
+  responseHeadersLabel,
+} from '../constants';
+
 const CACHE_CONTROL_HEADER = 'cache-control';
 const CACHE_CONTROL_REGEX = /^.*max-age=(\d+)/;
+const CACHE_KEY_VERSION_SUFFIX = '-v1';
+const CONDITIONAL_HEADER = 'if-none-match';
 const CONTENT_ENCODING_HEADER = 'content-encoding';
+const CONTENT_LENGTH_HEADER = 'content-length';
+const CONTENT_TYPE_HEADER = 'content-type';
```
steven-sheehy, line 23: Duplicates contentTypeHeader in constants.js. Does express provide any constants for these headers?

steven-sheehy (approved):
LGTM

rest/middleware/responseCacheHandler.js
xin-hedera: nope, express uses string literals in its code allover.

PR #11155, opened 2025-05-12
Cache null values on findStorage calls

sdimitrov9:
**Description**:
If many requests for the same non-existent value happen in a short burst, caching null avoids overwhelming the DB during that time and in the mean time any potential change in the DB will be picked up very soon since the cache entry TTL is 1 second.

Data about the contractState cache for the last 7 days:

Screenshot 2025-05-12 at 19 06 49

**Related issue(s)**:

Fixes #

**Checklist**


steven-sheehy (approved):

xin-hedera (approved):
LGTM

PR #11161, opened 2025-05-12
Cache null values on findStorage calls (0.129)

steven-sheehy:
**Description**:

Cherry pick of #11155 to `release/0.129`.

If many requests for the same non-existent value happen in a short burst, caching null avoids overwhelming the DB during that time and in the mean time any potential change in the DB will be picked up very soon since the cache entry TTL is 1 second.

**Related issue(s)**:

**Notes for reviewer**:

**Checklist**


xin-hedera (approved):
LGTM

jnels124 (approved):

PR #11184, opened 2025-05-16
Remove copied readable and writable kv states

### LINKED ISSUE #10487: Port the changes in the platform API for the mirror node state cache implementation
### Problem

There are currently some classes that were copied from the platform API temporarily for the state cache implementation - `ReadableKVStateBase` and `WritableKVStateBase`. This works currently but is hard to maintain.

### Solution

An upstream PR was created and merged - https://github.com/hiero-ledger/hiero-consensus-node/pull/18012 so when it becomes a part of a release we need to remove the copied classes and use Guava's `ForwardingMap` and pass it in the newly added parent constructor. This way we will delegate our `ScopedValue` data structures from the current approach to the `ForwardingMap's` methods and achieve the same results but make the code more maintainable.

### Alternatives

_No response_

bilyana-gospodinova:
**Description**:
In this PR the copied `ReadableKVStateBase` and `WritableKVStateBase` from the platform API are now removed.

Fixes #10487

**Checklist**


hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/state/core/MapReadableKVState.java
```diff
@@ -32,7 +33,7 @@ public class MapReadableKVState<K, V> extends ReadableKVStateBase<K, V> {
      * @param backingStore The backing store to use
      */
     public MapReadableKVState(@Nonnull final String stateKey, @Nonnull final Map<K, V> backingStore) {
-        super(stateKey);
+        super(stateKey, (ConcurrentMap<K, V>) backingStore);
```
bilyana-gospodinova, line 36: Here I am not sure if we need to do some special handling.

hedera-mirror-web3/src/main/java/com/hedera/mirror/web3/common/ContractCallContext.java
```diff
@@ -62,7 +64,7 @@ public class ContractCallContext {
     private CachingStateFrame<Object> stackBase;
 
     @Getter(AccessLevel.NONE)
-    private Map<String, Map<Object, Object>> readCache = new HashMap<>();
+    private Map<String, ConcurrentMap<Object, Object>> readCache = new HashMap<>();
```
bilyana-gospodinova, line 67: Changed to `ConcurrentMap` in order to match the upstream interface in `ReadableKVStateBase`.

PR #11237, opened 2025-05-21
Ensure treasury entity exists for balance generation

### LINKED ISSUE #9065: Calculate accurate historical balances for new accounts for partial mirror node
### Problem

We require `0.0.2` to exist to act as a special sentinel account for our balance snapshot creation, but for partial mirror nodes it will not exist. This causes inaccurate historical balances even for new accounts created after the start date.

### Solution

Insert a fake `shard.realm.2` to the database. Possibly we should also set a special field like memo to indicate this should not be returned via the API. Also set timestamp_range=[0,).

### Alternatives

_No response_

jnels124:
**Description**:
* ensure existence of treasury account before processing historical balances 

**Related issue(s)**:

Fixes #9065

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/importer/src/main/java/org/hiero/mirror/importer/parser/record/historicalbalance/HistoricalBalanceService.java
+++ b/importer/src/main/java/org/hiero/mirror/importer/parser/record/historicalbalance/HistoricalBalanceService.java
@@ -13,11 +13,12 @@ import jakarta.inject.Named;
 import java.time.Instant;
 import java.util.Optional;
 import java.util.concurrent.atomic.AtomicBoolean;
+import lombok.AccessLevel;
 import lombok.CustomLog;
+import lombok.Getter;
 import org.hiero.mirror.common.domain.StreamType;
 import org.hiero.mirror.common.domain.SystemEntity;
 import org.hiero.mirror.common.domain.balance.AccountBalanceFile;
-import org.hiero.mirror.common.domain.entity.Entity;
 import org.hiero.mirror.common.domain.entity.EntityType;
 import org.hiero.mirror.common.domain.transaction.RecordFile;
 import org.hiero.mirror.importer.db.TimePartitionService;
@@ -46,8 +47,10 @@ import org.springframework.transaction.support.TransactionTemplate;
 @CustomLog
 @Named
 public class HistoricalBalanceService {
-    private static final AtomicBoolean TREASURY_EXISTS = new AtomicBoolean(false);
     private static final String ACCOUNT_BALANCE_TABLE_NAME = "account_balance";
+
+    @Getter(AccessLevel.PACKAGE)
+    private final AtomicBoolean treasuryExists = new AtomicBoolean(false);
 
     private final AccountBalanceFileRepository accountBalanceFileRepository;
     private final AccountBalanceRepository accountBalanceRepository;
@@ -120,11 +123,7 @@ public class HistoricalBalanceService {
             }
 
             final long treasuryAccountId = systemEntity.treasuryAccount().getId();
-
-            if (!TREASURY_EXISTS.get()) {
-                checkTreasuryAccount();
-                TREASURY_EXISTS.set(true);
-            }
+            checkTreasuryAccount();
 
             log.info("Generating historical balances after processing record file with consensusEnd {}", consensusEnd);
             transactionTemplate.executeWithoutResult(t -> {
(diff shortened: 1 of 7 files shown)


importer/src/main/java/org/hiero/mirror/importer/parser/record/historicalbalance/HistoricalBalanceService.java
```diff
@@ -206,4 +219,25 @@ private boolean shouldGenerate(long consensusEnd) {
                                 >= properties.getMinFrequency().toNanos())
                         .isPresent();
     }
+
+    private void checkTreasuryAccount() {
```
jnels124, line 223: Adding as a migration is probably better but seems a little odd as well.

importer/src/main/java/org/hiero/mirror/importer/parser/record/historicalbalance/HistoricalBalanceService.java
```diff
@@ -42,7 +46,7 @@
 @CustomLog
 @Named
 public class HistoricalBalanceService {
-
+    private static final AtomicBoolean TREASURY_EXISTS = new AtomicBoolean(false);
```
steven-sheehy, line 49: It doesn't need to be static since it's a singleton. Also will make testing easier.

importer/src/main/java/org/hiero/mirror/importer/parser/record/historicalbalance/HistoricalBalanceService.java
```diff
@@ -113,6 +120,12 @@ public void onRecordFileParsed(RecordFileParsedEvent event) {
             }
 
             final long treasuryAccountId = systemEntity.treasuryAccount().getId();
+
+            if (!TREASURY_EXISTS.get()) {
+                checkTreasuryAccount();
+                TREASURY_EXISTS.set(true);
```
steven-sheehy, line 126:
```suggestion
            if (treasuryExists.compareAndSet(false, true)) {
                checkTreasuryAccount();
```

importer/src/main/java/org/hiero/mirror/importer/parser/record/historicalbalance/HistoricalBalanceService.java
```diff
@@ -113,6 +120,12 @@ public void onRecordFileParsed(RecordFileParsedEvent event) {
             }
 
             final long treasuryAccountId = systemEntity.treasuryAccount().getId();
+
+            if (!TREASURY_EXISTS.get()) {
+                checkTreasuryAccount();
```
steven-sheehy, line 125: Should push the if statement into checkTreasuryAccount() so it's encapsulated

importer/src/main/java/org/hiero/mirror/importer/parser/record/historicalbalance/HistoricalBalanceService.java
```diff
@@ -206,4 +219,25 @@ private boolean shouldGenerate(long consensusEnd) {
                                 >= properties.getMinFrequency().toNanos())
                         .isPresent();
     }
+
+    private void checkTreasuryAccount() {
+        var treasuryAccountEntityId = systemEntity.treasuryAccount();
+        var treasuryAccount = entityRepository.findById(treasuryAccountEntityId.getId());
+        if (treasuryAccount.isEmpty()) {
```
steven-sheehy, line 226:
```suggestion
        if (entityRepository.existsById(treasuryAccountEntityId.getId())) {
```

importer/src/main/java/org/hiero/mirror/importer/parser/record/historicalbalance/HistoricalBalanceService.java
```diff
@@ -206,4 +219,25 @@ private boolean shouldGenerate(long consensusEnd) {
...
+            log.warn("Treasury account {} not found in entity table, creating it", treasuryAccountEntityId);
```
steven-sheehy, line 227: Move to after save and log full entity `"Created sentinel treasury account: {}", entity")`

importer/src/main/java/org/hiero/mirror/importer/parser/record/historicalbalance/HistoricalBalanceService.java
```diff
@@ -206,4 +219,25 @@ private boolean shouldGenerate(long consensusEnd) {
...
+            entityRepository.save(Entity.builder()
+                    .balance(0L)
+                    .createdTimestamp(0L)
+                    .declineReward(true)
+                    .deleted(false)
+                    .id(treasuryAccountEntityId.getId())
```
steven-sheehy, line 233: Remove id, num, shard, and realm and use `treasuryAccountEntityId.toEntityId()`

rest/accounts.js
```diff
@@ -136,7 +136,7 @@ const getEntityBalanceQuery = (
   const {query: limitQuery, params: limitParams, order} = limitAndOrderQuery;
 
   const whereCondition = [
-    `e.type in ('ACCOUNT', 'CONTRACT')`,
+    `e.type in ('ACCOUNT', 'CONTRACT') and e.memo != 'SENTINEL_AUTO_GENERATED'`,
```
steven-sheehy, line 139: We need to check num == 2 as well so users can't hide their account from APIs by setting a secret memo.

importer/src/main/java/org/hiero/mirror/importer/parser/record/historicalbalance/HistoricalBalanceService.java
```diff
@@ -206,4 +219,25 @@ private boolean shouldGenerate(long consensusEnd) {
...
+                    .memo("SENTINEL_AUTO_GENERATED")
```
steven-sheehy, line 234: Would prefer a bit more explanatory memo.

```suggestion
                    .memo("Mirror node created synthetic treasury account")
```

steven-sheehy (changes requested):

importer/src/main/java/org/hiero/mirror/importer/parser/record/historicalbalance/HistoricalBalanceService.java
jnels124: didn't go this route in case the lookup/save errors for some reason. Rolling back the set value is a little awkward in that case.

steven-sheehy: Guess there's a risk the save fails and it doesn't try again when using compareAndSet

PR #11287, opened 2025-05-29
Remove copied readable and writable kv state classes v2

### LINKED ISSUE #10487: Port the changes in the platform API for the mirror node state cache implementation

kselveliev:
**Description**:
In this PR the copied `ReadableKVStateBase` and `WritableKVStateBase` from the platform API are now removed. A new implementation of Guava's `ForwardingConcurrentMap` is added that now delegates the `Map` function calls to our scoped transaction's readable and writable caches.

Fixes #10487

**Checklist**


kselveliev:
Needs more changes since some accepance tests are failing

web3/src/main/java/org/hiero/mirror/web3/state/core/MapReadableKVState.java
```diff
@@ -32,7 +35,12 @@ public class MapReadableKVState<K, V> extends ReadableKVStateBase<K, V> {
      * @param backingStore The backing store to use
      */
     public MapReadableKVState(@Nonnull final String stateKey, @Nonnull final Map<K, V> backingStore) {
-        super(stateKey);
+        super(stateKey, new ForwardingConcurrentMap<K, V>() {
+            @Override
+            protected ConcurrentMap<K, V> delegate() {
```
kselveliev, line 40: This makes acceptance test pass, but we need to fix a couple of failing non contract call tests

bilyana-gospodinova: The tests are fixed.

kselveliev:
LGTM, needs other approvers since I'm the creator of the PR.

PR #11292, opened 2025-05-30
Acceptance tests should not encode shard and realm in EVM addresses

### LINKED ISSUE #11265: Acceptance tests should not encode shard and realm in EVM addresses
### Problem

EVM addresses in account num format (i.e. long-zero) should not have the shard and realm encoded within the 20 bytes.

### Solution

* Update `TestUtil.asAddress(*)` to not call `toSolidityAddress()` until it can be fixed upstream
* Update places that use `ACCOUNT_ID_FORMAT` to use `DomainUtils.toEvmAddress(long)`
* Double check #10414 and #10630 to see if anything else should be undone
* Update `deleteTokenInvalidAddressEstimateGas`

### Alternatives

_No response_

filev94:
This PR makes changes in the acceptance tests so that toSolidityAddress() is not used until it is fixed upstream. Instead of the latter, I am using DomainUtils.toEvmAddress(long num). Modified asAddress methods to accommodate for these changes as well as when the switch back to toSolidityAddress needs to happen, change in small amount of places should be made.

**Related issue(s)**:

Fixes #11265 

**Checklist**


test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/EstimatePrecompileFeature.java
```diff
@@ -960,31 +917,29 @@ public void nestedFreezeAndUnfreezeNFTEstimateGas() {
...
-        var data = encodeData(ESTIMATE_PRECOMPILE, DELETE_TOKEN, asAddress(nonFungibleKycUnfrozenTokenId));
+        var data = encodeData(ESTIMATE_PRECOMPILE, DELETE_TOKEN, nonFungibleKycUnfrozenTokenAddress);
 
         validateGasEstimation(data, DELETE_TOKEN, estimatePrecompileContractSolidityAddress);
     }
 
     @Then("I call estimateGas with delete function for invalid token address")
     public void deleteTokenInvalidAddressEstimateGas() {
         String address = Hex.toHexString(ByteBuffer.allocate(20)
-                .putInt((int) commonProperties.getShard())
```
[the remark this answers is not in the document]
filev94: Addressed. Instead of shard & realm, putting 0 instead.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/util/TestUtil.java
```diff
@@ -75,56 +78,46 @@ public static String hexToAscii(String hexStr) {
...
-        final var address = tokenClient
-                .getSdkClient()
-                .getExpandedOperatorAccountId()
-                .getAccountId()
-                .toSolidityAddress();
-        final var addressBytes = Bytes.fromHexString(address.startsWith("0x") ? address : "0x" + address);
-        final var addressAsInteger = addressBytes.toUnsignedBigInteger();
-        return Address.wrap(Address.toChecksumAddress(addressAsInteger));
+    public static Address asAddress(final long num) {
+        return wrapEvmAddress(DomainUtils.toEvmAddress(num));
```
steven-sheehy: Can we skip DomainUtils.toEvmAddress() and directly convert long to BigInteger?

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/client/AbstractNetworkClient.java
```diff
@@ -204,6 +205,9 @@ public void validateAddress(final String actualAddress) {
                 .isEqualTo(
                         account.getPublicKey().isECDSA()
                                 ? account.getPublicKey().toEvmAddress().toString()
-                                : account.getAccountId().toSolidityAddress());
+                                : asAddress(account.getAccountId())
+                                        .toString()
+                                        .toLowerCase()
```
steven-sheehy: If we create separate helper methods for string addresses and use that everywhere then we don't have to call `toString().toLowerCase()` everywhere and don't have to change them again later.

```suggestion
                                : asHexAddress(account.getAccountId())
```

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/util/TestUtil.java
filev94: We can, however i saw that Jesse made changes in his PR regarding DomainUtils.toEvmAddress(num) and I was thinking that after i merge his changes here i could reuse them instead of duplicating them.

His changes are like this:

```
public static byte[] toEvmAddress(long num) {
        final var evmAddress = new byte[EVM_ADDRESS_LENGTH];
        final var buffer = ByteBuffer.wrap(evmAddress);
        buffer.putInt(0);
        buffer.putLong(0);
        buffer.putLong(num);
        return evmAddress;
}
```

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/client/AbstractNetworkClient.java
filev94: Added asHexAddress(AccountId accountId)

steven-sheehy: It was added, but not everywhere that uses `asAddress(...).toString()` in the diff was changed to use `asHexAddress()`

filev94: I guess I misunderstood your comment. Apologies. It is not needed to change the asAddress(...).toString(). I've tested with the proposed fix in the TestUtil comment - to change the asAddress methods to utilize toSolidityAddress and the tests pass without the need to revert anything.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/util/TestUtil.java
```diff
@@ -75,56 +75,50 @@ public static String hexToAscii(String hexStr) {
...
+        return asAddress(num);
     }
 
-    public static Address asAddress(ContractClient contractClient) {
-        final var address = contractClient.getClientAddress();
-        final var addressBytes = Bytes.fromHexString(address.startsWith("0x") ? address : "0x" + address);
-        final var addressAsInteger = addressBytes.toUnsignedBigInteger();
-        return Address.wrap(Address.toChecksumAddress(addressAsInteger));
+    public static Address asAddress(final long num) {
+        return Address.wrap(Address.toChecksumAddress(new BigInteger(Longs.toByteArray(num))));
```
steven-sheehy: Doesn't this work?

```suggestion
        return Address.wrap(Address.toChecksumAddress(BigInteger.valueOf(num)));
```

steven-sheehy (approved):

PR #11319, opened 2025-06-03
Utilize child error transactions in web3

### LINKED ISSUE #11243: Utilize child transaction errors
### Problem

In TransactionExecutionSerivce.execute method when we get the receipt list we only grab the first object from it which is the parent transaction. There are cases where this parent transaction can return a SUCCESS result but the child transactions (the next objects in the list) can contain error messages - check tests added in this [PR](https://github.com/hiero-ledger/hiero-mirror-node/pull/11242) Currently we are ignoring them and it might lead to some false positives. Also when the parent transaction has status CONTRACT_REVERT_EXECUTED, usually child transactions contain the real error message that was thrown in services. 

### Solution

It is possible that we could enrich our logs so that when there is a CONTRACT_REVERT_EXECUTED or any other exception we can log the real reason why that happened (if it doesn't exist in the parent transaction) by checking the child transactions and their statuses. This, of course, can lead to enriching our tests as well.

Image

Image

Image

### Alternatives

_No response_

filev94:
This PR makes changes to utilize child transaction errors in web3 modularized flow. There are cases where we have CONTRACT_REVERT_EXECUTED with no message or detail in the parent transaction but the real reason it failed in services is present in the child transactions. Added unique statuses to childTransactionErrors list and propagate them to GenericResponse. There is a case where the parent transaction is SUCCESS but it has child transactions that failed - logging a warning in that case.

**Related issue(s)**:

Fixes #11243 

**Checklist**


web3/src/main/java/org/hiero/mirror/web3/exception/MirrorEvmTransactionException.java
```diff
@@ -25,6 +27,7 @@ public class MirrorEvmTransactionException extends EvmException {
     private final String data;
     private final Boolean isCallModularized;
     private final transient HederaEvmTransactionProcessingResult result;
+    private final List<String> childTransactionErrors = new ArrayList<>();
```
bilyana-gospodinova: This collection can be a `LinkedList` since we don't really need to fetch a particular entry by index.

web3/src/main/java/org/hiero/mirror/web3/service/TransactionExecutionService.java
```diff
@@ -110,23 +135,34 @@ private HederaEvmTransactionProcessingResult buildSuccessResult(
...
-            var errorMessage = getErrorMessage(result).orElse(Bytes.EMPTY);
-            var detail = maybeDecodeSolidityErrorStringToReadableMessage(errorMessage);
+            final var errorMessage = getErrorMessage(result).orElse(Bytes.EMPTY);
+            final var detail = maybeDecodeSolidityErrorStringToReadableMessage(errorMessage);
+
+            final var childTransactionErrors = new ArrayList<String>();
+
+            for (int i = 1; i < transactionRecords.size(); i++) {
+                final var record = transactionRecords.get(i).transactionRecord();
+                addErrorMessage(isContractCreate, record, childTransactionErrors);
```
bilyana-gospodinova: Won't we add the successful child transaction statuses as well this way?

web3/src/main/java/org/hiero/mirror/web3/exception/MirrorEvmTransactionException.java
filev94: Addressed

web3/src/main/java/org/hiero/mirror/web3/service/TransactionExecutionService.java
filev94: If the errorMessage of said transaction is not blank, then yes. So I changed it to skip successful child transactions entirely.

filev94: At first i was storing errorMessage because there is a high chance they are more unique and it will be easier to debug IMO, but decided to change to storing statuses, so we can have REVERT_SUCCESS and others that do not have errorMessage and potentially would miss out on them.

web3/src/main/java/org/hiero/mirror/web3/service/TransactionExecutionService.java
```diff
@@ -240,4 +263,19 @@ private OperationTracer[] getOperationTracers() {
                 ? new OperationTracer[] {opcodeTracer}
                 : new OperationTracer[] {mirrorOperationTracer};
     }
+
+    private List<String> populateChildTransactionErrors(List<SingleTransactionRecord> transactionRecords) {
+        final var childTransactionErrors = new LinkedHashSet<String>();
+
+        for (int i = 1; i < transactionRecords.size(); i++) {
+            final var record = transactionRecords.get(i).transactionRecord();
```
steven-sheehy: We don't know if the list supports O(1) random access to use get(i) efficiently.

```suggestion
        for (final var record : transactionRecords) {
```

web3/src/main/java/org/hiero/mirror/web3/service/TransactionExecutionService.java
```diff
@@ -240,4 +263,19 @@ private OperationTracer[] getOperationTracers() {
...
+            final var status = record.receiptOrThrow().status();
+            if (status == com.hedera.hapi.node.base.ResponseCodeEnum.SUCCESS) {
+                continue;
+            }
+            childTransactionErrors.add(status.protoName());
+        }
+
+        return new ArrayList<>(childTransactionErrors);
```
steven-sheehy: Change everywhere to use `SequencedCollection` so you can directly return the `childTransactionErrors` set and avoid allocating and copying an extra list.

web3/src/main/java/org/hiero/mirror/web3/service/TransactionExecutionService.java
```diff
@@ -77,28 +80,42 @@ public HederaEvmTransactionProcessingResult execute(final CallServiceParameters
...
+        final var parentTransaction = transactionRecords.getFirst().transactionRecord();
+        final var childTransactionErrors = populateChildTransactionErrors(transactionRecords);
+
+        final var result = getTransactionResult(parentTransaction, isContractCreate);
+
+        if (!childTransactionErrors.isEmpty()) {
+            // there are some child transactions that failed but parent is SUCCESS, logging a warning
+            final var contractId = result.contractID();
+            log.warn(
+                    "Child transaction errors present for contract: {} with successful parent transaction, isModularized: {}, errors: {}",
```
steven-sheehy: Remove isModularized since it's implied when log statement source is `TransactionExecutionService`.

web3/src/main/java/org/hiero/mirror/web3/service/TransactionExecutionService.java
```diff
@@ -240,4 +263,19 @@ private OperationTracer[] getOperationTracers() {
                 ? new OperationTracer[] {opcodeTracer}
                 : new OperationTracer[] {mirrorOperationTracer};
     }
+
+    private List<String> populateChildTransactionErrors(List<SingleTransactionRecord> transactionRecords) {
+        final var childTransactionErrors = new LinkedHashSet<String>();
```
steven-sheehy: Prefer setting to null and only lazily creating when first entry added. Reason is with current code we'd be allocating both a set and list for every transaction when only a tiny percent will have child errors. At bottom we could do `return childTransactionErrors != null ? childTransactionErrors : List.of();`

filev94: Addressed

web3/src/main/java/org/hiero/mirror/web3/service/TransactionExecutionService.java
filev94: Addressed

web3/src/main/java/org/hiero/mirror/web3/service/TransactionExecutionService.java
filev94: Addressed

web3/src/main/java/org/hiero/mirror/web3/service/TransactionExecutionService.java
filev94: Used iterator since I want to skip the first singleTransactionRecord, which is the parent.

web3/src/main/java/org/hiero/mirror/web3/exception/MirrorEvmTransactionException.java
```diff
@@ -57,6 +60,17 @@ public MirrorEvmTransactionException(
...
 
+    public MirrorEvmTransactionException(
+            final String message,
+            final String detail,
+            final String hexData,
+            final HederaEvmTransactionProcessingResult result,
+            final Boolean isCallModularized,
+            final SequencedCollection<String> childTransactionErrors) {
+        this(message, detail, hexData, result, isCallModularized);
+        this.childTransactionErrors.addAll(childTransactionErrors);
```
steven-sheehy: Why do we need to copy? Can't we just assign?

filev94: Addressed, also initialized it to List.of() instead of LinkedList

steven-sheehy (approved):

PR #11320, opened 2025-06-03
HIP-1081 BlockNodeSubscriber

### LINKED ISSUE #11025: HIP-1081 Implement `BlockNodeSubscriber`
### Problem

To support streaming block stream items from block node, we need to implement `BlockNodeSubscriber` as outlined in HIP-1081 block node support design

### Solution

- Add configuration property classes `BlockNodeProperties` and `BlockSourceType`, add them to `BlockStreamProperties`
- Implement `BlockNodeSubscriber`
  - pick block node to stream block items from by priority group then round robin, also maintain the next block node to try; optionally instead of round-robin, for the same priority group, consider sorting the nodes by latency and always pick lowest latency node first
  - logic to reassemble block items belonging to the same block
- Implement `AUTO` mode switch in `CompositeBlockStreamSource`

### Alternatives

_No response_

xin-hedera:
**Description**:

**Related issue(s)**:

Fixes #11025 

**Notes for reviewer**:

**Checklist**


docs/design/block-node-support.md
```diff
@@ -76,27 +80,50 @@ public enum BlockSourceType {
 
 See [`CompositeBlockStreamSource`](#CompositeBlockStreamSource) for how `AUTO` works.
 
-### BlockStreamProperties
+### BlockProperties
```
xin-hedera, line 83: renamed most class with `BlockStream` prefix to just `Block`, since I added `StreamProperties` for common block node properties, it's confusing to have a `StreamProperties` field inside `BlockStreamProperties`.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/StreamProperties.java
```diff
@@ -0,0 +1,39 @@
...
+import lombok.Data;
+import org.hibernate.validator.constraints.time.DurationMin;
+import org.springframework.validation.annotation.Validated;
+
+@Data
+@Validated
+public class StreamProperties {
+
+    @Min(1000)
+    private int maxBlockItems = 800_000;
```
xin-hedera, line 20: we can adjust the default value when we can get a more realistic number from perfnet, once the finalized block stream protobuf implemented in consensus node.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/StreamProperties.java
```diff
@@ -0,0 +1,39 @@
...
+
+    @Min(16)
+    private int maxBufferSize = 128;
```
xin-hedera: we need a large enough buffer for catch up mode so we don't frequently overrun the buffer and cancel the subscription unnecessarily.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockNodeSubscriber.java
```diff
@@ -0,0 +1,112 @@
...
+
+@Named
+public final class BlockNodeSubscriber extends AbstractBlockSource {
+
+    private static final long UNKNOWN_NODE_ID = -1;
+
+    private final Map<Integer, List<BlockNode>> nodes = new TreeMap<>();
+
+    @Getter(lazy = true, value = AccessLevel.PRIVATE)
+    private final Scheduler scheduler = createScheduler();
```
jnels124: nit: Couldn't we just create the scheduler inline on the `.publishOn`? Not sure lazily creating the scheduler provides any benefit and makes the code more verbose

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockNode.java
```diff
@@ -0,0 +1,230 @@
...
+
+    public Flux<StreamedBlock> stream(long blockNumber) {
+        var context = new AtomicReference<SubscriptionContext>();
+        return Flux.<StreamedBlock>create(sink -> {
+                    var channel = channelBuilder.build();
+                    var call = channel.newCall(
+                            BlockStreamSubscribeServiceGrpc.getSubscribeBlockStreamMethod(), CallOptions.DEFAULT);
+                    context.set(new SubscriptionContext(call, channel));
+                    var request = SubscribeStreamRequest.newBuilder()
+                            .setEndBlockNumber(INFINITE_END_BLOCK_NUMBER)
```
jnels124: Do we have a follow up task to support a stopping point similar to support for recordStreams?

jnels124 (dismissed):

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockNodeSubscriber.java
jnels124: ah never mind I see the benefit. In case there is resubscribes don't have to create the pool again

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockNode.java
xin-hedera: think we should. will create a ticket.

docs/configuration.md
```diff
@@ -40,7 +40,16 @@ value, it is recommended to only populate overridden properties in the custom `a
...
+| `hiero.mirror.importer.block.nodes[].host`                                      |                                                      | The hostname or IP of the block node server.                                                                                                                                                                                                                       |
+| `hiero.mirror.importer.block.nodes[].port`                                      |                                                      | The port of the block node server.                                                                                                                                                                                                                                 |
```
steven-sheehy: Does block node not have a default port?

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockProperties.java
```diff
@@ -2,27 +2,38 @@
...
 @Validated
-public class BlockStreamProperties {
+public class BlockProperties {
 
     private boolean enabled = false;
 
     @NotNull
     private Duration frequency = Duration.ofMillis(100L);
 
+    private Collection<BlockNodeProperties> nodes = Collections.emptyList();
```
steven-sheehy, line 31:
```suggestion
    @Valid
    private Collection<BlockNodeProperties> nodes = Collections.emptyList();
```

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockSource.java
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package org.hiero.mirror.importer.downloader.block;
+
+interface BlockSource {
+
+    /*
+     * Gets blocks from the source. An implementation can either download block files from cloud storage, or stream
+     * blocks from a block node.
```
steven-sheehy: We should omit details about different implementations.
```suggestion
     * Gets blocks from the source.
```

importer/src/main/java/org/hiero/mirror/importer/downloader/block/StreamProperties.java
```diff
@@ -0,0 +1,36 @@
...
+
+@Data
+@Validated
+public class StreamProperties {
+
+    @Min(1000)
+    private int maxBlockItems = 800_000;
+
+    @Min(1)
+    private int maxStreamResponseSizeMB = 8;
```
steven-sheehy: Prefer the use of DataSize
```suggestion
    @NotNull
    @DataSizeUnit(DataUnit.KILOBYTES)
    private DataSize maxStreamResponseSize = DataSize.ofMegabytes(8);
```

importer/src/main/java/org/hiero/mirror/importer/downloader/block/StreamProperties.java
```diff
@@ -0,0 +1,36 @@
...
+    @Min(1)
+    private int maxSubscribeAttempts = 3;
+
+    @DurationMin(seconds = 10)
+    @NotNull
+    private Duration readmitDelay = Duration.ofMinutes(1);
+
+    @DurationMin(millis = 100)
+    @NotNull
+    private Duration statusTimeout = Duration.ofMillis(400);
```
steven-sheehy: Prefer `responseTimeout`

importer/build.gradle.kts
```diff
@@ -12,6 +18,9 @@ dependencies {
     implementation("com.fasterxml.jackson.dataformat:jackson-dataformat-csv")
     implementation("commons-io:commons-io")
     implementation("io.github.mweirauch:micrometer-jvm-extras")
+    implementation("io.grpc:grpc-inprocess")
```
steven-sheehy: Shouldn't this be testImplementation?

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockFileSource.java
```diff
@@ -26,7 +26,7 @@
 import org.hiero.mirror.importer.util.Utility;
 
 @Named
-final class BlockFileSource extends AbstractBlockStreamSource {
+public final class BlockFileSource extends AbstractBlockSource {
```
steven-sheehy: Why does it need to be public?

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockNodeSubscriber.java
```diff
@@ -0,0 +1,96 @@
...
+import org.hiero.mirror.importer.reader.block.BlockStream;
+import org.hiero.mirror.importer.reader.block.BlockStreamReader;
+import org.springframework.beans.factory.DisposableBean;
+
+@Named
+public final class BlockNodeSubscriber extends AbstractBlockSource implements DisposableBean {
+
+    private static final long UNKNOWN_NODE_ID = -1;
+
+    private final Map<Integer, List<BlockNode>> nodes = new TreeMap<>();
```
steven-sheehy: This is a bit complicated and could be simplified. We should have `BlockNode` implement `Comparable` that uses priority internally. Then we can just use an ArrayList that we sort once. If we need to optimize retrieval when there's a large number of inactive nodes we could consider a PriorityQueue.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockNodeSubscriber.java
```diff
@@ -0,0 +1,96 @@
...
+        nodes.values().stream().flatMap(List::stream).forEach(BlockNode::destroy);
+    }
+
+    @Override
+    public void get() {
+        long blockNumber = getNextBlockNumber();
+        var node = getNode(blockNumber);
+        if (node == null) {
+            throw new BlockStreamException("No block node can provide block " + blockNumber);
+        }
```
steven-sheehy: Could this be removed to inside getNode() so it never returns null? Then add nonnull

importer/src/main/java/org/hiero/mirror/importer/downloader/block/CompositeBlockSource.java
```diff
@@ -0,0 +1,103 @@
...
+import lombok.RequiredArgsConstructor;
+import org.hiero.mirror.common.domain.transaction.BlockFile;
+import org.hiero.mirror.importer.leader.Leader;
+import org.springframework.context.annotation.Primary;
+import org.springframework.scheduling.annotation.Scheduled;
+
+@CustomLog
+@Named
+@Primary
+public final class CompositeBlockSource implements BlockSource {
```
steven-sheehy:
```suggestion
final class CompositeBlockSource implements BlockSource {
```

importer/src/main/java/org/hiero/mirror/importer/downloader/block/CompositeBlockSource.java
```diff
@@ -0,0 +1,103 @@
...
+
+    private final SourceHealth blockFileSourceHealth;
+    private final SourceHealth blockNodeSubscriberSourceHealth;
+    private final BlockStreamVerifier blockStreamVerifier;
+    private SourceHealth current;
```
steven-sheehy: Probably we should use AtomicReference just to be safe.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/CompositeBlockSource.java
```diff
@@ -0,0 +1,103 @@
...
+        }
+    }
+
+    private SourceHealth getSourceHealth() {
+        return switch (properties.getSourceType()) {
+            case AUTO -> {
+                boolean switched = blockStreamVerifier
+                        .getLastBlockFilename()
+                        .map(BlockFile::isStreamedFilename)
+                        .orElse(false);
```
steven-sheehy: Is this the best approach? Perhaps the file format could change. Should we add a column to record file to indicate source is either S3 or block node?

importer/src/main/java/org/hiero/mirror/importer/downloader/block/CompositeBlockSource.java
```diff
@@ -0,0 +1,103 @@
...
+                if (switched) {
+                    yield blockNodeSubscriberSourceHealth;
+                }
+
+                if (properties.getNodes().isEmpty()) {
+                    yield blockFileSourceHealth;
```
steven-sheehy, line 70: Maybe we should throw an error instead of assume they just want files?

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockNode.java
```diff
@@ -0,0 +1,250 @@
...
+    private final StreamProperties streamProperties;
+
+    BlockNode(BlockNodeProperties properties, StreamProperties streamProperties) {
+        var endpoint = properties.getEndpoint();
+        var channelBuilder = properties.isInProcess()
+                ? InProcessChannelBuilder.forName(endpoint)
+                : ManagedChannelBuilder.forTarget(endpoint);
+        this.channel = channelBuilder
+                .maxInboundMessageSize(streamProperties.getMaxStreamResponseSize())
+                .usePlaintext()
```
steven-sheehy: Not sure we should assume plaintext. Maybe we should only use plaintext if port is not 443?

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockNode.java
```diff
@@ -0,0 +1,250 @@
...
+    private int errors = 0;
+    private final BlockNodeProperties properties;
+    private final AtomicReference<Instant> readmitTime = new AtomicReference<>(Instant.now());
+    private final StreamProperties streamProperties;
+
+    BlockNode(BlockNodeProperties properties, StreamProperties streamProperties) {
+        var endpoint = properties.getEndpoint();
+        var channelBuilder = properties.isInProcess()
+                ? InProcessChannelBuilder.forName(endpoint)
+                : ManagedChannelBuilder.forTarget(endpoint);
```
steven-sheehy: I don't think it's a good idea to have in process on the classpath just for tests. We can create a `ManagedChannelFactory` that only gets overridden in tests to always create in-process channels.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockNode.java
```diff
@@ -0,0 +1,250 @@
...
+final class BlockNode {
+
+    private static final long INFINITE_END_BLOCK_NUMBER = -1;
+    private static final ServerStatusRequest SERVER_STATUS_REQUEST = ServerStatusRequest.getDefaultInstance();
+
+    @Getter
+    private boolean active = true;
+
+    private final ManagedChannel channel;
+    private int errors = 0;
```
steven-sheehy: Use AtomicInteger

docs/configuration.md
xin-hedera: nope, not in the HIP. in local node, it's mapped to port 8080

importer/build.gradle.kts
xin-hedera: used it in `BlockNode.java`, though it's mainly for testing.

docs/configuration.md
steven-sheehy: I meant in the hiero-block-node. Looks like it's [8080](https://github.com/hiero-ledger/hiero-block-node/blob/74025c95111a6d8be2c56c4e031393bcaa58f33a/charts/block-node-server/values.yaml#L41) as well. We should use that as our default.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/CompositeBlockSource.java
xin-hedera: reason is the default is `AUTO` and the logic is for when there's only block files in cloud storage.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockFileSource.java
xin-hedera: think it's some changes forced it to be public, and later those changes were removed...

importer/src/main/java/org/hiero/mirror/importer/downloader/block/CompositeBlockSource.java
xin-hedera: perhaps create a ticket to track? once it goes canonical, we should not expect filename change.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockNode.java
```diff
@@ -0,0 +1,250 @@
...
+            var request = SubscribeStreamRequest.newBuilder()
+                    .setEndBlockNumber(INFINITE_END_BLOCK_NUMBER)
+                    .setStartBlockNumber(blockNumber)
+                    .build();
+            grpcCall.set(ClientCalls.blockingV2ServerStreamingCall(
+                    channel,
+                    BlockStreamSubscribeServiceGrpc.getSubscribeBlockStreamMethod(),
+                    CallOptions.DEFAULT,
+                    request));
+            var timeout = new TimeoutContext(blockTimeout);
```
steven-sheehy: Can't we combine the concept of the TimeoutContext and BlockAssembler and have a single object?

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockNode.java
```diff
@@ -0,0 +1,250 @@
...
+                            onStreamedBlock.accept(streamedBlock);
+                        }
+                    }
+                    case STATUS -> {
+                        var status = response.getStatus();
+                        if (status == SubscribeStreamResponse.Code.READ_STREAM_SUCCESS) {
+                            // The server may end the stream gracefully for various reasons, and this shouldn't be
+                            // treated as an error.
+                            log.info("Block server ended the subscription with {}", status);
+                            break outer;
```
steven-sheehy: If it ended it gracefully shouldn't we try to check another block node then instead of trying to read from the same one again?

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockNode.java
```diff
@@ -0,0 +1,250 @@
...
+                    }
+                    default -> throw new BlockStreamException("Unknown response case " + response.getResponseCase());
+                }
+            }
+        } catch (Exception ex) {
+            onError();
+
+            if (ex instanceof BlockStreamException blockStreamException) {
+                throw blockStreamException;
+            }
```
steven-sheehy: Better to add another specific catch and duplicate the onError() call

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockNode.java
```diff
@@ -0,0 +1,250 @@
...
+                            break outer;
+                        }
+
+                        throw new BlockStreamException("Received status " + response.getStatus() + " from block node");
+                    }
+                    default -> throw new BlockStreamException("Unknown response case " + response.getResponseCase());
+                }
+            }
+        } catch (Exception ex) {
+            onError();
```
steven-sheehy, line 134: Pass ex to onError and print out its message each time its called

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockNode.java
```diff
@@ -0,0 +1,250 @@
...
+            active = true;
+        }
+
+        return this;
+    }
+
+    private void onError() {
+        if (++errors >= streamProperties.getMaxSubscribeAttempts()) {
+            log.warn(
+                    "Failed to stream blocks from {} {} times consecutively, mark it inactive",
```
steven-sheehy: Add error message to end and reword.

```suggestion
                    "Marking connection to {} as inactive after {} attempts: {}"
```

steven-sheehy (changes requested):

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockNode.java
xin-hedera: debatable. `SUCCESS` should mean there's nothing wrong at the server side. Also given that we do infinite subscription, the most common scenario would be the server has a internal policy about the max number of blocks willing to stream per single subscription request.

Given the node selection algorithm, plus if we incorporate latency later, we should still have the node as one of the active nodes to consider.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockNode.java
xin-hedera: the exception will be logged in `CompositeBlockSource` so we should avoid double logging.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockNode.java
xin-hedera: was that way I just don't like to call onError() in two catch blocks.

importer/build.gradle.kts
```diff
@@ -34,14 +41,18 @@ dependencies {
     implementation("software.amazon.awssdk:netty-nio-client")
     implementation("software.amazon.awssdk:s3")
     implementation("software.amazon.awssdk:sts")
+    protobuf("org.hiero.block:block-node-protobuf-sources:0.13.0-rc1")
```
xin-hedera: have to specify version here

steven-sheehy: I'm not sure if specifying it here will allow dependabot to update it. Why can't we put in root build gradle?

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockNodeProperties.java
```diff
@@ -0,0 +1,39 @@
...
+@Data
+@Validated
+public class BlockNodeProperties implements Comparable<BlockNodeProperties> {
+
+    private static final Comparator<BlockNodeProperties> COMPARATOR = Comparator.comparing(
+                    BlockNodeProperties::getPriority)
+            .thenComparing(BlockNodeProperties::getHost)
+            .thenComparing(BlockNodeProperties::getPort);
+
+    @NotEmpty
```
steven-sheehy: `@NotBlank`?

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockNodeProperties.java
```diff
@@ -0,0 +1,39 @@
...
+    private String host;
+
+    @Max(65535)
+    @Min(1)
```
steven-sheehy: Technically 0 is a valid port
```suggestion
    @Min(0)
```

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockNodeSubscriber.java
```diff
@@ -0,0 +1,83 @@
...
+                .map(blockNodeProperties ->
+                        new BlockNode(channelBuilderProvider, blockNodeProperties, properties.getStream()))
+                .sorted()
+                .toList();
+    }
+
+    @Override
+    public void destroy() {
+        nodes.forEach(BlockNode::destroy);
+    }
```
steven-sheehy: Prefer `AutoCloseable` of both subscriber and node. Spring supports close method so prefer not coupling to spring DisposableBean.
```suggestion
    public void close() {
        nodes.forEach(BlockNode::close);
        nodes.clear();
    }
```

importer/build.gradle.kts
xin-hedera: It just doesn’t work. When using api and version in the main build file and then building importer the version just isn’t picked up here.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/CompositeBlockSource.java
```diff
@@ -0,0 +1,105 @@
...
+            case BLOCK_NODE -> blockNodeSubscriberSourceHealth;
+            case FILE -> blockFileSourceHealth;
+        };
+    }
+
+    @Getter
+    @RequiredArgsConstructor
+    private static class SourceHealth {
+
+        private int errors = 0;
```
steven-sheehy: Should use AtomicInteger here just in case.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockProperties.java
```diff
@@ -2,27 +2,39 @@
...
-public class BlockStreamProperties {
+public class BlockProperties {
 
     private boolean enabled = false;
 
     @NotNull
     private Duration frequency = Duration.ofMillis(100L);
 
+    @Valid
+    private Collection<BlockNodeProperties> nodes = Collections.emptyList();
```
steven-sheehy, line 31: Add `@NotNull`

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockNode.java
```diff
@@ -0,0 +1,266 @@
...
+            blockNodeProperties -> {
+                var builder = ManagedChannelBuilder.forTarget(blockNodeProperties.getEndpoint());
+                if (blockNodeProperties.getPort() != 443) {
+                    builder.usePlaintext();
+                } else {
+                    builder.useTransportSecurity();
+                }
+
+                return builder;
+            };
```
steven-sheehy: This duplicates code in ManagedChannelBuilderProviderImpl and seems to be unused?

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockNodeSubscriber.java
```diff
@@ -0,0 +1,83 @@
...
+        }
+
+        throw new BlockStreamException("No block node can provide block " + blockNumber);
+    }
+
+    private BlockStream toBlockStream(BlockNode.StreamedBlock streamedBlock) {
+        var blockItems = streamedBlock.blockItems();
+        var blockHeader = blockItems.getFirst().getBlockHeader();
+        var filename = BlockFile.getFilename(blockHeader.getNumber(), false);
+        return new BlockStream(blockItems, null, filename, streamedBlock.loadStart(), UNKNOWN_NODE_ID);
```
steven-sheehy: Can we not just convert directly to BlockStream in BlockNode and delete the intermediary `BlockNode.StreamedBlock`?

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockNode.java
```diff
@@ -0,0 +1,266 @@
...
+                        }
+
+                        throw new BlockStreamException("Received status " + response.getStatus() + " from block node");
+                    }
+                    default -> throw new BlockStreamException("Unknown response case " + response.getResponseCase());
+                }
+
+                errors.set(0);
+
+                if (serverSuccess) {
```
steven-sheehy: Can we not move serverSuccess outside loop and add `&& !serverSuccess` to while?

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockNode.java
```diff
@@ -0,0 +1,266 @@
...
+            if (blockItems.isEmpty()) {
+                log.warn("Received empty BlockItemSet from block node");
+                return null;
+            }
+
+            var firstItemCase = blockItems.getFirst().getItemCase();
+            append(blockItems, firstItemCase);
+
+            if (firstItemCase == BLOCK_HEADER || firstItemCase == RECORD_FILE) {
+                loadStart = System.currentTimeMillis();
```
steven-sheehy, line 197: Can we not assume that we start in BlockAssembler constructor and make the field final? I don't think the time gap between request first block and receive first block will be high?

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockNode.java
```diff
@@ -0,0 +1,266 @@
...
+                    this,
+                    streamProperties.getMaxSubscribeAttempts());
+        }
+    }
+
+    private class BlockAssembler {
+
+        private long loadStart;
+        private final List<List<BlockItem>> pending = new ArrayList<>();
+        private int pendingCount = 0;
```
steven-sheehy, line 177: Why can't we flatten the block items into a single dimenson list when we receive them? The fact that their aggregated during send is only for performance reasons upstream. Then we can ditch pendingCount as well.

steven-sheehy (changes requested):

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockNodeSubscriber.java
xin-hedera: ~that's solely for `loadStart`, for blocks from block nodes, loadStart is recorded when the block header of a block is received.~

think that was intended to leave non-streaming related logic out of `BlockNode` as much as possible.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockNode.java
xin-hedera: the assembler's lifecycle is the whole streaming subscription, so it can span over multiple streamed blocks. we have to record `loadStart` for each completed block, so loadStart can't be final.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockNode.java
xin-hedera: performance consideration on our end

- avoid flattening when the whole block is in a single stream response
- defend ourselves against malicious block node which tries to stream a huge block (many block items) in multiple responses, so we don't waste cycles to keep flattening them.

importer/build.gradle.kts
steven-sheehy: Can we specify version in `extra.apply` in root build.gradle.kts and use that here?

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockNode.java
steven-sheehy: I'm not quite convinced. We protect against such a scenario by limiting the stream response size. If we also need to protected against too many block items in a response then we should throw an exception up front if it exceeds that threshold. I think it will be cleander code and slightly faster performance for happy path since we don't have to allocate a separate array and flatten at the end.

xin-hedera: The performance difference is negligible. And for a good behaving block node, it's actually the opposite because a good block node should stream a complete block in a single stream response most of the time, the logic here would avoid the unnecessary flattening. For a malicious node, we also spend the least amount of memory possible to accumulate the stream responses and throw an exception once the number of items exceeds the max, without the overhead of flattening at all.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockNode.java
xin-hedera: https://github.com/hiero-ledger/hiero-mirror-node/issues/11435

PR #11321, opened 2025-06-03
Drop shard / realm in long form evm address encoding and validation

### LINKED ISSUE #11246: REST API account num alias should not encode shard and realm
### Problem

In #10598, the change was made to verify EVM addresses in account num format (i.e. long-zero) should have the shard and realm encoded within the 20 bytes. This requirement has now been dropped.

### Solution

* Change `entityid.js` to no longer verify the first 12 bytes against the shard and realm. Account num alias should now always be zero-prefixed regardless of it's in a different network/realm/ledger
* Change `entityService.js` to no longer lookup addresses or alias with the shard and realm
* Check anywhere else in rest and in #10598

### Alternatives

_No response_

### LINKED ISSUE #11317: Log errors when js rest module readiness check fails
### Problem

As a developer, I want to have log message to indicate why JS rest module readiness check fails. 

### Solution

terminus `createTerminus` function has a `logger` option, when configured, will log a message, and the error details if any checks fail.

```javascript
const options = {
  ...,
  logger // [optional] logger function to be called with errors. Example logger call: ('error happened during shutdown', error). See terminus.js for more details.
};
```

### Alternatives

_No response_

xin-hedera:
**Description**:

- Remove shard / realm check and encoding in `entityId.js`
- Remove shard / realm conditions in SQL query in `entityService.js`
- Remove shard / realm conditions in `contractController.js` and `contractService.js`
- Log readiness check errors

**Related issue(s)**:

Fixes #11246
Fixes #11317

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/rest/entityId.js
+++ b/rest/entityId.js
@@ -96,7 +96,6 @@ class EntityId {
       return null;
     }
 
-    // shard, realm, and num take 4, 8, and 8 bytes respectively from the left
     return this.num === null ? null : toHexString(this.num, true, 40);
   }
 
@@ -212,13 +211,13 @@ const parseFromEncodedId = (id, error) => {
 };
 
 /**
- * Parses shard, realm, num from EVM address string.
+ * Parses prefix and num from EVM address string.
  * @param {string} evmAddress
  * @return {[string, bigint]}
  */
 const parseFromEvmAddress = (evmAddress) => {
-  // The first 24 chars is the prefix and the last 16 is the num
   const hexDigits = _.last(stripHexPrefix(evmAddress).split('.'));
+  // The first 24 chars is the prefix and the last 16 is the num
   return [
     hexDigits.slice(0, 24),
     BigInt(constants.HEX_PREFIX + hexDigits.slice(24, 40)), // num
--- a/rest/entityId.js
+++ b/rest/entityId.js
@@ -38,6 +38,9 @@ const encodedEntityIdRegex = /^-?\d{1,19}$/;
 const evmAddressShardRealmRegex = /^(\d{1,4}\.)?(\d{1,5}\.)?[A-Fa-f0-9]{40}$/;
 const evmAddressRegex = /^(0x)?[A-Fa-f0-9]{40}$/;
 
+// 0x prefix + 12-byte 0s
+const longFormEvmAddressPrefix = `${constants.HEX_PREFIX}${'00'.repeat(12)}`;
+
 class EntityId {
   /**
    * Creates an EntityId instance
@@ -94,14 +97,7 @@ class EntityId {
     }
 
     // shard, realm, and num take 4, 8, and 8 bytes respectively from the left
-    return this.num === null
-      ? null
-      : [
-          constants.HEX_PREFIX,
-          toHex(this.shard).padStart(8, '0'),
-          toHex(this.realm).padStart(16, '0'),
-          toHex(this.num).padStart(16, '0'),
-        ].join('');
+    return this.num === null ? null : `${longFormEvmAddressPrefix}${toHex(this.num).padStart(16, '0')}`;
   }
 
   toString() {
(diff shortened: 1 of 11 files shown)


jnels124 (dismissed):

rest/entityId.js
```diff
@@ -156,8 +152,9 @@ const isEvmAddressAlias = (evmAddress) => {
   if (!isValidEvmAddress(evmAddress)) {
     return false;
   }
+
   const parts = parseFromEvmAddress(evmAddress);
```
steven-sheehy, line 151: We need to update or remove `parseFromEvmAddress`. It should not need to slice addresses at all that we then compare against zero. We should instead add a `isAccountNumAlias()` that checks the first 12 bytes are zero.

rest/entityId.js
```diff
@@ -94,14 +97,7 @@ class EntityId {
...
     // shard, realm, and num take 4, 8, and 8 bytes respectively from the left
-    return this.num === null
-      ? null
-      : [
-          constants.HEX_PREFIX,
-          toHex(this.shard).padStart(8, '0'),
-          toHex(this.realm).padStart(16, '0'),
-          toHex(this.num).padStart(16, '0'),
-        ].join('');
+    return this.num === null ? null : `${longFormEvmAddressPrefix}${toHex(this.num).padStart(16, '0')}`;
```
steven-sheehy: Push this logic into `toHex()`.

Even better, can't we use `utils.toHexString(Buffer.from(num.toString(16), 'utf8'), true, 40)`? Not sure if there's a simpler way to convert number directly to buffer.

steven-sheehy (changes requested):

rest/entityId.js
xin-hedera: there's multiple back and forth unnecessary conversions if we want to use `toHexString`, perhaps we should optimize toHexString

rest/entityId.js
xin-hedera: kept the method, and changed it to slice into two parts, we still need to extract the num part for long form address from all accepted formats.

rest/entityId.js
```diff
@@ -221,14 +216,13 @@ const parseFromEncodedId = (id, error) => {
...
- * @return {bigint[3]}
+ * @return {bigint[2]}
  */
 const parseFromEvmAddress = (evmAddress) => {
   // extract shard from index 0->8, realm from 8->23, num from 24->40 and parse from hex to decimal
   const hexDigits = _.last(stripHexPrefix(evmAddress).split('.'));
   return [
-    BigInt(constants.HEX_PREFIX + hexDigits.slice(0, 8)), // shard
-    BigInt(constants.HEX_PREFIX + hexDigits.slice(8, 24)), // realm
+    hexDigits.slice(0, 24), // shard + realm
```
steven-sheehy: Remove the comment as it's no longer encoding shard and realm in the address.

rest/entityId.js
```diff
@@ -96,12 +99,7 @@ class EntityId {
     // shard, realm, and num take 4, 8, and 8 bytes respectively from the left
     return this.num === null
       ? null
-      : [
-          constants.HEX_PREFIX,
-          toHex(this.shard).padStart(8, '0'),
-          toHex(this.realm).padStart(16, '0'),
-          toHex(this.num).padStart(16, '0'),
-        ].join('');
+      : `${constants.HEX_PREFIX}${longFormEvmAddressPrefix}${toHexString(this.num, false, 16)}`;
```
steven-sheehy: Why concatenate three strings when should be able to use:

```suggestion
      : toHexString(this.num, true, 40);
```

xin-hedera: choice between readability and slightly better performance.

jnels124 (dismissed):

rest/entityId.js
```diff
@@ -94,14 +97,7 @@ class EntityId {
     }
 
     // shard, realm, and num take 4, 8, and 8 bytes respectively from the left
```
steven-sheehy: Update or remove inaccurate comment.

rest/entityId.js
```diff
@@ -221,14 +214,13 @@ const parseFromEncodedId = (id, error) => {
 /**
  * Parses shard, realm, num from EVM address string.
```
steven-sheehy: Update or remove inaccurate comment.

PR #11329, opened 2025-06-05
Adjusting gasPerSecond and gasLimitRefundPercent based on metrics

### LINKED ISSUE #11244: Adjust gasPerSecond and gasLimitRefundPercent based upon metrics
### Problem

In #8390, we disabled gasLimitRefundPercent and increased gasPerSecond until we had metrics, but we forgot to revisit the problem.

### Solution

Analyze metrics in production environments and use the data to adjust gasLimitRefundPercent and gasPerSecond appropriately

### Alternatives

_No response_

filev94:
This PR makes changes to gasPerSecond and gasLimitRefundPercent to adjust them based on metrics GasConsumed and GasLimit, taken from mainnet clusters in the time period - 05/27/2025 - 06/06/2025. Based on analysis, following observations can be made:

Gas Consumed
- **Peak usage is ~67.2M** on 05/30/2025 between 20:23:00 and 20:25:00 UTC+3
- **95% of calls consume under ~61M**.

Gas Limit
- Clients often **over-request gas**
- Some request **billions**, but only consume a fraction

Based on this, I've made changes so that the gasPerSecond is 75 Million, which is around 12% buffer above the max value. Could be a bit tight and potential change to 85 Mill or even 100 Mill is possible. Restored the gasLimitRefundPercent to its previous value of 20 to prevent over-abuse

**Related issue(s)**:

Fixes #11244 

**Notes for reviewer**:
Currently the traffic on mainnet is 100% mono code.

**Checklist**


filev94:
Currently I've proposed 75 Million as the value for gasPerSecond, but it might be a bit tight, compared to 67 Mill. Can adjust to 85 Mill or even 100 Mill.

bilyana-gospodinova:
I agree with Ivan - the `gasPerSecond` property should be way higher than 75 000 000 or 100 000 000 as it counts the passed gas limit in the request (and not the actual gas used). I think we should keep the `gasLimitRefundPercent` set to 100% as it is now, because roughly more than 90% of the passed gas limit remains unused. 
We can create a task to improve the throttle mechanism by taking into account the actual gas used by the transactions instead of the gas limit. This way the throttle will be more meaningful. Another suggestion for improvement is that we can consume 250 000 gas per request from the bucket (which is an average for the actual gas used) instead of the passed gas limit (by default 15 000 000) and at the end of the transaction execution we can adjust the consumed gas in the bucket by adding/removing gas. For example:
1. If the transaction took 280 000 gas - we need to consume 30 000 more at the end of execution.
2. If the transaction took 200 000 gas - we need to return 50 000 gas to the bucket.

filev94:
Thanks for the feedback! Adjusted the gasPerSecond to be 7.5 B to cover 500 rps and refund is back at 100%. Created a task for improvement of said throttle mechanism - https://github.com/hiero-ledger/hiero-mirror-node/issues/11366

bilyana-gospodinova (approved):

steven-sheehy (approved):

IvanKavaldzhiev (approved):
LGTM

PR #11382, opened 2025-06-12
Add jumbo transaction tests

### LINKED ISSUE #11354: HIP-1086 Test jumbo transactions gas estimate
### Problem

Currently we don't have a test for gas estimate for the newly added jumbo transactions.

### Solution

Add at least one test to cover the scenario.

### Alternatives

_No response_

kselveliev:
**Description**:
This PR adds integration tests for jumbo transactions.
There are upstream changes in 0.63 that enable us to remove `props.put("jumboTransactions.isEnabled", "false");`

Fixes #11354

**Checklist**

### DIFF
--- a/web3/src/test/java/org/hiero/mirror/web3/service/ContractCallJumboTransactionTest.java
+++ b/web3/src/test/java/org/hiero/mirror/web3/service/ContractCallJumboTransactionTest.java
@@ -4,13 +4,13 @@ package org.hiero.mirror.web3.service;
 
 import static com.hedera.hapi.node.base.ResponseCodeEnum.TRANSACTION_OVERSIZE;
 import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
+import static org.assertj.core.api.AssertionsForClassTypes.assertThatThrownBy;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 
 import com.hedera.node.config.data.JumboTransactionsConfig;
 import org.hiero.mirror.common.exception.MirrorNodeException;
 import org.hiero.mirror.web3.exception.MirrorEvmTransactionException;
 import org.hiero.mirror.web3.web3j.generated.JumboTransaction;
-import org.hiero.mirror.web3.web3j.generated.NestedCalls;
 import org.junit.jupiter.api.Test;
 
 class ContractCallJumboTransactionTest extends AbstractContractCallServiceTest {
@@ -24,7 +24,7 @@ class ContractCallJumboTransactionTest extends AbstractContractCallServiceTest {
     void testJumboTransactionHappyPath() {
         // Given
         final var jumboPayload = new byte[JUMBO_PAYLOAD];
-        final var contract = testWeb3jService.deploy(JumboTransaction::deploy);
+        final var contract = testWeb3jService.deployWithInput(JumboTransaction::deploy, new byte[0]);
         // When
         final var functionCall = contract.send_consumeLargeCalldata(jumboPayload);
         // Then
@@ -35,11 +35,31 @@ class ContractCallJumboTransactionTest extends AbstractContractCallServiceTest {
     void testJumboContractCreate() {
         // Given
         testWeb3jService.setUseContractCallDeploy(true);
+        final var jumboPayload = new byte[JUMBO_PAYLOAD];
         // When
-        // 12 kb deploy
-        final var contract = testWeb3jService.deploy(NestedCalls::deploy);
+        final var contract = testWeb3jService.deployWithInput(JumboTransaction::deploy, jumboPayload);
         // Then
         assertThat(contract.getContractAddress()).isNotNull();
+    }
+
+    @Test
+    void testJumboContractCreateOverMaxSize() {
+        // Given
+        testWeb3jService.setUseContractCallDeploy(true);
+        final int maxDataSize = mirrorNodeEvmProperties
+                .getVersionedConfiguration()
+                .getConfigData(JumboTransactionsConfig.class)
+                .ethereumMaxCallDataSize();
+        byte[] jumboPayload;
+        if (!mirrorNodeEvmProperties.isModularizedServices()) {
+            // mono requires a little bit larger call data for deploy to fail
+            jumboPayload = new byte[maxDataSize + JUMBO_PAYLOAD];
+        } else {
+            jumboPayload = new byte[maxDataSize];
+        }
+        // Then
+        assertThatThrownBy(() -> testWeb3jService.deployWithInput(JumboTransaction::deploy, jumboPayload))
+                .isInstanceOf(Exception.class);
     }
 
     @Test
@@ -50,7 +70,7 @@ class ContractCallJumboTransactionTest extends AbstractContractCallServiceTest {
                 .getConfigData(JumboTransactionsConfig.class)
                 .ethereumMaxCallDataSize();
         final var jumboPayload = new byte[maxDataSize];
-        final var contract = testWeb3jService.deploy(JumboTransaction::deploy);
+        final var contract = testWeb3jService.deployWithInput(JumboTransaction::deploy, new byte[0]);
         // When
         final var functionCall = contract.call_consumeLargeCalldata(jumboPayload);
         // Тhen
--- a/web3/src/test/java/org/hiero/mirror/web3/service/ContractCallJumboTransactionTest.java
+++ b/web3/src/test/java/org/hiero/mirror/web3/service/ContractCallJumboTransactionTest.java
@@ -12,8 +12,11 @@ import org.junit.jupiter.api.Test;
 
 public class ContractCallJumboTransactionTest extends AbstractContractCallServiceTest {
 
-    private static final int JUMBO_PAYLOAD = 64 * 128;
-    private static final int OVERSIZED_JUMBO_PAYLOAD = 128 * 1024;
+    private static final int KILOBYTE = 1024;
+
+    // Jumbo payload: any payload over 6 KiB and up to 128 KiB
+    private static final int JUMBO_PAYLOAD = 64 * KILOBYTE;
+    private static final int OVERSIZED_JUMBO_PAYLOAD = 128 * KILOBYTE;
 
     @Test
     void testJumboTransactionHappyPath() {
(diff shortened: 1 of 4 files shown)


web3/src/test/java/org/hiero/mirror/web3/service/ContractCallJumboTransactionTest.java
```diff
@@ -0,0 +1,43 @@
...
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import org.hiero.mirror.common.exception.MirrorNodeException;
+import org.hiero.mirror.web3.web3j.generated.JumboTransaction;
+import org.junit.jupiter.api.Test;
+
+public class ContractCallJumboTransactionTest extends AbstractContractCallServiceTest {
+
+    private static final int JUMBO_PAYLOAD = 64 * 128;
```
bilyana-gospodinova: Can you please add some comments explaining these numbers? In future we might forget what each them means.

web3/src/test/java/org/hiero/mirror/web3/service/ContractCallJumboTransactionTest.java
```diff
@@ -0,0 +1,43 @@
...
+        // When
+        final var functionCall = contract.send_consumeLargeCalldata(jumboPayload);
+        // Then
+        verifyEthCallAndEstimateGas(functionCall, contract);
+    }
+
+    @Test
+    void testJumboTransactionOverMaxSize() {
+        // Given
+        if (!mirrorNodeEvmProperties.isModularizedServices()) {
```
bilyana-gospodinova, line 32: Does this mean this is another breaking change and we need to add it to the doc and the informational HIP?

bilyana-gospodinova, line 54: Shouldn't this transaction fail in mono too?

web3/src/test/java/org/hiero/mirror/web3/service/ContractCallJumboTransactionTest.java
kselveliev: Added a little bit of explanation. And this line was actually wrong, should have been 64 * 1024. Addressed.

web3/src/test/java/org/hiero/mirror/web3/service/ContractCallJumboTransactionTest.java
kselveliev: It fails but only modularized seems to support the TRANSACTION_OVERSIZE status for over 128kb.
We just fail with Insufficient gas in mono.

bilyana-gospodinova: Then the test can run for mono as well and the if-check `mirrorNodeEvmProperties.isModularizedServices()` can be moved at the end of test to verify the `TRANSACTION_OVERSIZE` status only in mod. For mono the verification can just be that the code throws an exception.

kselveliev: Addressed

bilyana-gospodinova (dismissed):

web3/src/test/java/org/hiero/mirror/web3/service/ContractCallJumboTransactionTest.java
```diff
@@ -0,0 +1,48 @@
...
+import org.hiero.mirror.web3.web3j.generated.JumboTransaction;
+import org.junit.jupiter.api.Test;
+
+public class ContractCallJumboTransactionTest extends AbstractContractCallServiceTest {
+
+    private static final int KILOBYTE = 1024;
+
+    // Jumbo payload: any payload over 6 KiB and up to 128 KiB
+    private static final int JUMBO_PAYLOAD = 64 * KILOBYTE;
+    private static final int OVERSIZED_JUMBO_PAYLOAD = 128 * KILOBYTE;
```
steven-sheehy: We should pull this max size from hedera-app config as it may change in the future.

web3/src/test/java/org/hiero/mirror/web3/service/ContractCallJumboTransactionTest.java
```diff
@@ -0,0 +1,48 @@
...
+import static com.hedera.hapi.node.base.ResponseCodeEnum.TRANSACTION_OVERSIZE;
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import org.hiero.mirror.common.exception.MirrorNodeException;
+import org.hiero.mirror.web3.exception.MirrorEvmTransactionException;
+import org.hiero.mirror.web3.web3j.generated.JumboTransaction;
+import org.junit.jupiter.api.Test;
+
+public class ContractCallJumboTransactionTest extends AbstractContractCallServiceTest {
```
steven-sheehy: nit:
```suggestion
final class ContractCallJumboTransactionTest extends AbstractContractCallServiceTest {
```

steven-sheehy (commented):
Do we in the future need to consider switching TransactionExecutionService to use EthereumTransaction instead of create and call?

web3/src/test/java/org/hiero/mirror/web3/service/ContractCallJumboTransactionTest.java
steven-sheehy: You used our property. We should use JumboTransactionsConfig

web3/src/test/java/org/hiero/mirror/web3/service/ContractCallJumboTransactionTest.java
kselveliev: Addressed

web3/src/test/java/org/hiero/mirror/web3/service/ContractCallJumboTransactionTest.java
kselveliev: Problem is after we bumped to 63.5 unless we convert to EthereumTransaction the only data check will be coming from what we pass as max size from "executor.maxSignedTxnSize". 
We should consider switching to EthereumCall in order to make use of the jumbo configs.

steven-sheehy: I don't think that's true. It's using `JumboTransactionsConfig.ethereumMaxCallDataSize` in `ContextTransactionProcessor` to throw the `TRANSACTION_OVERSIZE` that you're catching in your test. So we should use that property in the test as well.

web3/src/test/java/org/hiero/mirror/web3/service/ContractCallJumboTransactionTest.java
```diff
@@ -0,0 +1,48 @@
...
+        // When
+        final var functionCall = contract.send_consumeLargeCalldata(jumboPayload);
+        // Then
+        verifyEthCallAndEstimateGas(functionCall, contract);
+    }
+
+    @Test
+    void testJumboTransactionOverMaxSize() {
+        // Given
+        final int maxDataSize = (int) mirrorNodeEvmProperties.getMaxDataSize().toBytes();
```
steven-sheehy: I think we should probably remove our property now and purely rely upon the jumbo config. Unless there's some scenario where it does not get validated against that config? We can probably remove the logic in `ContractController.validateContractData()` as well. We still need the logic in `JacksonConfiguration` but it should use the jumbo config.

web3/src/test/java/org/hiero/mirror/web3/service/ContractCallJumboTransactionTest.java
```diff
@@ -0,0 +1,48 @@
...
+import org.hiero.mirror.web3.web3j.generated.JumboTransaction;
+import org.junit.jupiter.api.Test;
+
+class ContractCallJumboTransactionTest extends AbstractContractCallServiceTest {
+
+    private static final int KILOBYTE = 1024;
+
+    // Jumbo payload: any payload over 6 KiB and up to 128 KiB
+    private static final int JUMBO_PAYLOAD = 64 * KILOBYTE;
+
```
steven-sheehy, line 22: Add a test for large create as well

web3/src/test/java/org/hiero/mirror/web3/service/ContractCallJumboTransactionTest.java
kselveliev: Ahh yes I double checked that. Using the JumboTransactionsConfig.ethereumMaxCallDataSize now.

web3/src/test/java/org/hiero/mirror/web3/service/ContractCallJumboTransactionTest.java
kselveliev: Addressed

web3/src/test/java/org/hiero/mirror/web3/service/ContractCallJumboTransactionTest.java
kselveliev: I agree — we probably can remove that property. My initial intention for this PR was just to add tests, but removing the config might require a bit of investigation to ensure it doesn't introduce any other issues, along with potentially adding a few more tests. I can follow up with those changes in a separate PR.

web3/src/test/java/org/hiero/mirror/web3/service/ContractCallJumboTransactionTest.java
```diff
@@ -0,0 +1,64 @@
...
+        verifyEthCallAndEstimateGas(functionCall, contract);
+    }
+
+    @Test
+    void testJumboContractCreate() {
+        // Given
+        testWeb3jService.setUseContractCallDeploy(true);
+        // When
+        // 12 kb deploy
+        final var contract = testWeb3jService.deploy(NestedCalls::deploy);
```
steven-sheehy: I don't think we should use `NestedCalls`. I think we should a param to the constructor of `JumboTransaction.sol` that takes a `bytes` param of variable length. Then we should have a test at `JUMBO_PAYLOAD` size that succeeds and one at `ethereumMaxCallDataSize` that fails.

web3/src/test/java/org/hiero/mirror/web3/service/ContractCallJumboTransactionTest.java
steven-sheehy: That's fine. Please open a ticket to track and we can work next sprint.

web3/src/test/java/org/hiero/mirror/web3/service/ContractCallJumboTransactionTest.java
kselveliev: Addressed

steven-sheehy (approved):

PR #11388, opened 2025-06-13
Add workaround for long zero addresses to be permitted by the EVM

### LINKED ISSUE #11381: Investigate successful workflow when aliased account is referenced by its shard/realm/num id
The mono workflow inside web3 didn't have stricter validations for the address format that was passed for execution. The modularized workflow, however, utilizes a hedera-app library, which reuses the business logic inside consensus nodes, which is - when an account has an alias, it should be referenced only by its alias and it should be taken in priority.

This discrepancy might be inconvenient for some users, so this task aims to find a way for shard/realm/num addresses to work even in modularized workflow.

IvanKavaldzhiev:
**Description**:
Allow modularized workflow to work even if an account having an alias is referenced by its long zero (shard/realm/num) address. This will serve as a workaround for one of the breaking changes connected to the migration from mono to modularized workflow inside web3.

In order the workaround to be utilized, 2 classed needed to be overriden in the classpath. These are `DispatchingEvmFrameState` and `ConversionUtils`.

The only change in `DispatchingEvmFrameState` is inside `isNotPriority`, where based on a feature flag we return that the address is always a priority in case the feature flag is activated and if it's not then the actual check is performed.

The only change in `ConversionUtils` is inside `accountNumberForEvmReference`. If the feature flag is activated then even if an account having an alias is compared with a passed long zero address, then no `NON_CANONICAL_REFERENCE_NUMBER` error will be thrown. If the feature flag is disabled and the comparison has a mismatch, then the error will be thrown.

**Related issue(s)**:

Fixes #11381 

**Checklist**


kselveliev:
We probably should be able to add this behind a feature flag

IvanKavaldzhiev:
> We probably should be able to add this behind a feature flag

Added feature flag.

web3/src/main/java/org/hiero/mirror/web3/evm/properties/OverrideClasspathProperties.java
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package org.hiero.mirror.web3.evm.properties;
+
+import jakarta.annotation.PostConstruct;
+import lombok.Value;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.validation.annotation.Validated;
+
+@Value
```
bilyana-gospodinova: If we configure `allowLongZeroAddresses` to `true` in the "application.yml", won't this `@Value` annotation break the Spring bean initialization since there is no setter method?

web3/src/test/java/org/hiero/mirror/web3/service/ContractCallServiceTest.java
```diff
@@ -370,13 +372,22 @@ void estimateGasForViewCall() {
         verifyEthCallAndEstimateGas(functionCall, contract);
     }
 
-    @Test
-    void transferFunds() {
+    @ParameterizedTest
+    @ValueSource(booleans = {true, false})
+    void transferFunds(boolean longZeroAddressAllowed) {
```
bilyana-gospodinova, line 377: Would you please add more tests for this change? We need to verify a scenario where some of the handlers and system contracts is involved as in this case the code flow is different and the address might not be resolved correctly.
For example (modified test from `ContractCallServicePrecompileModificationTest`):
```
@ParameterizedTest
@ValueSource(booleans = {true, false})
void associateToken(final Boolean single) throws Exception {
    // Given
    final var notAssociatedAccount = accountEntityWithEvmAddressPersist(); // the changed code is here

    final var token = fungibleTokenPersist();
    final var contract = testWeb3jService.deploy(ModificationPrecompileTestContract::deploy);

    // When
    final var functionCall = single
            ? contract.call_associateTokenExternal(
                    getAddressFromEntity(notAssociatedAccount),
                    toAddress(token.getTokenId()).toHexString())
            : contract.call_associateTokensExternal(
                    getAddressFromEntity(notAssociatedAccount),
                    List.of(toAddress(token.getTokenId()).toHexString()));

    // Then
    verifyEthCallAndEstimateGas(functionCall, contract, ZERO_VALUE);
    verifyOpcodeTracerCall(functionCall.encodeFunctionCall(), contract);
}
```

IvanKavaldzhiev: Added more tests covering different use cases like calling precompile with function argument, calling HAS precompile, etc.

web3/src/main/java/org/hiero/mirror/web3/evm/properties/OverrideClasspathProperties.java
IvanKavaldzhiev: Changed to @ Data to be compliant with TraceProperties as well.

web3/src/test/java/org/hiero/mirror/web3/service/ContractCallServicePrecompileModificationTest.java
```diff
@@ -172,14 +173,23 @@ void approveNFT(final Boolean approve) throws Exception {
...
 
         final var token = nonFungibleTokenPersist();
         final var tokenId = token.getTokenId();
 
+        String spenderAddress;
+        System.setProperty(ALLOW_LONG_ZERO_ADDRESSES, Boolean.toString(longZeroAddressAllowed));
+        if (longZeroAddressAllowed) {
+            spenderAddress = getAddressFromEntity(spender);
+        } else {
+            spenderAddress = getEvmAddressBytesFromEntity(spender).toHexString();
```
bilyana-gospodinova: We have a `getAliasFromEntity()` method that does this directly.

web3/src/test/java/org/hiero/mirror/web3/service/ContractCallServiceTest.java
bilyana-gospodinova: Great! I see that `ConversionUtils` is changed to use shard and realm in some methods. Can you please add test coverage for that as well? This can be combined with a test that creates an NFT with an auto-renew account that has an alias but is passed by a long-zero address.

kselveliev:
Will be useful to increase PR description to point out the the main changes that we need to review in the overriden classes.

web3/src/main/java/org/hiero/mirror/web3/evm/properties/OverrideClasspathProperties.java
```diff
@@ -0,0 +1,23 @@
...
+import lombok.Data;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.validation.annotation.Validated;
+
+@Data
+@Validated
+@ConfigurationProperties(prefix = "hiero.mirror.web3.evm.classpath")
+public class OverrideClasspathProperties {
+
+    public static final String ALLOW_LONG_ZERO_ADDRESSES = "ALLOW_LONG_ZERO_ADDRESSES";
```
steven-sheehy:
```suggestion
    public static final String ALLOW_LONG_ZERO_ADDRESSES = "HIERO_MIRROR_WEB3_MODULARIZED_ALLOWLONGZEROADDRESS";
```

docs/configuration.md
```diff
@@ -716,6 +716,7 @@ value, it is recommended to only populate overridden properties in the custom `a
...
 | `hiero.mirror.web3.evm.network`                              | TESTNET                                            | Which network to use. Can be either `MAINNET`, `PREVIEWNET`, `TESTNET` or `OTHER`                                                                                                                |
 | `hiero.mirror.web3.evm.feesTokenTransferUsageMultiplier`     | 380                                                | Used to calculate token transfer fees                                                                                                                                                            |
+| `hiero.mirror.web3.evm.classpath.allowLongZeroAddresses`     | false                                              | Whether to enable long zero addresses to be permitted by the EVM                                                                                                                                 |
```
steven-sheehy: We shouldn't document it since it's private and we want to remove it soon.

web3/src/main/java/org/hiero/mirror/web3/evm/properties/OverrideClasspathProperties.java
```diff
@@ -0,0 +1,23 @@
...
+
+import jakarta.annotation.PostConstruct;
+import lombok.Data;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.validation.annotation.Validated;
+
+@Data
+@Validated
+@ConfigurationProperties(prefix = "hiero.mirror.web3.evm.classpath")
+public class OverrideClasspathProperties {
```
steven-sheehy: I don't think we need this class or the ability to map from yaml. We can just have it as a private env variable.

web3/src/test/java/org/hiero/mirror/web3/service/ContractCallServiceTest.java
IvanKavaldzhiev: The class is copied using the class from hedera-app:63.5, so the only difference is the one mentioned in the PR description.


I updated the mintNft test to also utilise aliased auto-renew account, so that we can also cover this path.

web3/src/test/java/org/hiero/mirror/web3/service/ContractCallServicePrecompileModificationTest.java
IvanKavaldzhiev: Updated

web3/src/main/java/org/hiero/mirror/web3/evm/properties/OverrideClasspathProperties.java
IvanKavaldzhiev: Renamed

docs/configuration.md
IvanKavaldzhiev: Removed

web3/src/main/java/org/hiero/mirror/web3/evm/properties/OverrideClasspathProperties.java
IvanKavaldzhiev: Updated

steven-sheehy (approved):

Issue #10744, opened 2025-03-29
Run unit and integration tests with non-zero realms


steven-sheehy:
### Problem

The unit and integration tests all mostly use default zero shard and realm except for the few that specifically test non-zero added recently.

### Solution

* Temporarily change default shard and realm in `CommonProperties` and in `CommonIntegrationTest` and run each module's test.
* Run tests in each module and fix any issues
* Refactor tests to avoid hard-coding entity IDs (explicitly or implicitly) to be zero realm and shard
* Split into many refactor PRs by module or package if necessary

mgoelswirlds:
Modules that have tests passing with NZR:

- [x] Common
- [x] [Graphql](#10866)
- [x] gRPC
- [x] [Importer](#10903)
- [x] Monitor
- [x] Rest java
- [x] [web3](#10719)
- [x] Rest

PR #11410, opened 2025-06-16
configure tests to work with non-zero shard/realm

jnels124:
**Description**:
* update tests to not use hard coded shard and realm 

**Related issue(s)**:

Part Of #10744 

**Notes for reviewer**:

**Checklist**


web3/src/main/java/com/hedera/services/store/contracts/precompile/TokenCreateWrapper.java
```diff
@@ -209,14 +210,16 @@ private FixedFee.Builder asBuilder() {
...
-                    FixedFee.newBuilder()
+                case USE_CURRENTLY_CREATED_TOKEN -> {
+                    final var commonProperties = CommonProperties.getInstance();
+                    yield FixedFee.newBuilder()
                             .setAmount(amount)
                             .setDenominatingTokenId(TokenID.newBuilder()
-                                    .setShardNum(0L)
-                                    .setRealmNum(0L)
+                                    .setShardNum(commonProperties.getShard())
+                                    .setRealmNum(commonProperties.getRealm())
```
steven-sheehy: Isn't this just some `0.0.0` sentinel value that doesn't change with shard and realm? The protobuf docs state `If this value is set to `0.0.0` in the `tokenCreate` transaction, it SHALL be replaced with the `TokenID` of the newly created token.`

web3/src/test/java/com/hedera/services/ledger/TransferLogicTest.java
```diff
@@ -48,19 +49,17 @@
...
+    private static final DomainBuilder domainBuilder = new DomainBuilder();
 
     private final long initialAllowance = 100L;
 
-    private final AccountID payer = AccountID.newBuilder().setAccountNum(12345L).build();
-    private final AccountID owner = AccountID.newBuilder().setAccountNum(12347L).build();
-    private final TokenID fungibleTokenID =
-            TokenID.newBuilder().setTokenNum(1234L).build();
+    private final AccountID payer = domainBuilder.entityNum(12345L).toAccountID();
+    private final AccountID owner = domainBuilder.entityNum(12347L).toAccountID();
```
steven-sheehy: Ideally we'd update tests from static data to dynamic data e.g. `domainBuilder.entityId().toAccountID()`. Are we sure these only work with these numbers? Or is it too much effort to change to dynamic?

Same comment for the rest of the PR.

jnels124: can make them more dynamic in most if not all cases. Will adjust

web3/src/main/java/com/hedera/services/store/contracts/precompile/TokenCreateWrapper.java
jnels124: The value changes based on shard/realm configuration. Will double check this.

steven-sheehy (approved):

xin-hedera (approved):
LGTM

PR #11411, opened 2025-06-17
Add workaround for INSUFFICIENT_PAYER_BALANCE in modularized

### LINKED ISSUE #11355: Research how we can mitigate breaking changes from the new modularized flow
### Problem

There are some breaking changes in the new modularized code, e.g. long-zero/alias addresses mismatch.

### Solution

Research if there is a way to mitigate this (+ maybe add a property to enable/disable this) for a more graceful migration to the new code for our users.

### Alternatives

_No response_

### LINKED ISSUE #11380: Investigate workaround for INSUFFICIENT_PAYER_BALANCE errors in modularized workflow
Mono code workflow inside web3 did have broader validations and sender's balance wasn't taken into account when transactions were simulated. Inside the modularized workflow, which utilizes hedera-app library, the same validations are conducted as in consensus nodes. This means sender's balance is taken into account.

Try to find a way to mitigate this issue inside modularized workflow, so that it doesn't throw this error.

kselveliev:
**Description**:
This PR add a config which when enabled will avoid `INSUFFICIENT_PAYER_BALANCE` errors which is a new breaking change in modularized codebase.

How it works:
Enable the overridePayerBalanceValidation flag.

Configure a minimum account balance using `MirrorNodeEvmProperties.minimumAccountBalance`.

When `overridePayerBalanceValidation` is enabled, and an account is being read from `AccountReadableKVState`, the following logic applies:

If the account’s actual balance in the database is less than the configured `minimumAccountBalance`, the system overrides the balance and sets it to the configured minimum.

This ensures the payer balance validation in hedera-app passes, as long as the minimum is set high enough.

Exception:
This override does not occur during `balanceOf` calls, where the real account balance must be returned.

Fixes #11380 
Fixes #11355

Acceptance test ethCallReentrancyCallFunction fails which does:
```
 function reentrancyCallWithGas(address _to, uint256 _amount) external returns (uint256, uint256)  {
        uint256 balanceBefore = address(this).balance;
        payable(_to).call{value: _amount, gas: 2300}("");
        uint256 balanceAfter = address(this).balance;

        return (balanceBefore, balanceAfter);
    }
```
balanceBefore is returned with the overriden since not coming from CustomBalanceOperation -
address(this).balance does not resolve to CustomBalanceOperation

**Checklist**


web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AbstractAliasedAccountReadableKVState.java
```diff
@@ -155,7 +160,20 @@ private Supplier<Long> getAccountBalance(final Entity entity, final Optional<Lon
                         return 0L;
                     }
                 })
-                .orElseGet(() -> Objects.requireNonNullElse(entity.getBalance(), 0L)));
+                .orElseGet(() -> {
+                    if (!overrideClasspathProperties.isOverridePayerBalanceValidation()) {
+                        return entity.getBalance();
+                    }
+                    boolean isBalanceCall = ContractCallContext.get().isBalanceCall();
+                    Long currentBalance = entity.getBalance();
```
IvanKavaldzhiev: These 2 fields can be final

web3/src/main/java/com/hedera/node/app/service/contract/impl/exec/operations/CustomBalanceOperation.java
```diff
@@ -0,0 +1,62 @@
...
+import org.hyperledger.besu.evm.frame.MessageFrame;
+import org.hyperledger.besu.evm.gascalculator.GasCalculator;
+import org.hyperledger.besu.evm.internal.UnderflowException;
+import org.hyperledger.besu.evm.internal.Words;
+import org.hyperledger.besu.evm.operation.BalanceOperation;
+
+/**
+ * A Hedera customization of the Besu {@link org.hyperledger.besu.evm.operation.BalanceOperation}.
+ */
+public class CustomBalanceOperation extends BalanceOperation {
```
bilyana-gospodinova, line 25: Would you please document in a comment what is the change in this class compared to the original one?

web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AbstractAliasedAccountReadableKVState.java
kselveliev: Addressed

web3/src/main/java/com/hedera/node/app/service/contract/impl/exec/operations/CustomBalanceOperation.java
kselveliev: Addressed

web3/src/test/java/org/hiero/mirror/web3/service/AbstractContractCallServiceTest.java
```diff
@@ -859,11 +866,14 @@ protected ByteString convertToCompressedPublicKey(final PublicKey publicKey) {
      * @param publicKey - the public key to be set to the account
      */
     protected Entity persistAccountWithEvmAddressAndPublicKey(byte[] evmAddress, byte[] publicKey) {
+        final var accountBalance = mirrorNodeEvmProperties.isOverridePayerBalanceValidation()
```
IvanKavaldzhiev: This code is repetitive. Even though it's a single line it can be extracted in a private method.

In addition if we override the payer balance validation, won't we need the DEFAULT_ACCOUNT_BALANCE instead of the DEFAULT_SMALL_ACCOUNT_BALANCE, since want to use a bigger value that passes all such validations?

web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AbstractAliasedAccountReadableKVState.java
```diff
@@ -155,7 +156,20 @@ private Supplier<Long> getAccountBalance(final Entity entity, final Optional<Lon
                         return 0L;
                     }
                 })
-                .orElseGet(() -> Objects.requireNonNullElse(entity.getBalance(), 0L)));
+                .orElseGet(() -> {
+                    if (!mirrorNodeEvmProperties.isOverridePayerBalanceValidation()) {
+                        return entity.getBalance();
+                    }
+                    boolean isBalanceCall = ContractCallContext.get().isBalanceCall();
+                    Long currentBalance = entity.getBalance();
```
IvanKavaldzhiev: `Long currentBalance = entity.getBalance();` can be moved above the if statement, since in any case we need the currentBalance. Currently we call entity.getBalance() twice, without the need to.

kselveliev: Addressed

web3/src/test/java/org/hiero/mirror/web3/service/AbstractContractCallServiceTest.java
kselveliev: Refactored common logic into a helper method.

kselveliev: The goal is to simulate a low account balance in the database so that the call fails due to insufficient payer balance.

However, when the `overridePayerBalanceValidation` property is enabled, the system ignores the DB value and instead uses `mirrorNodeEvmProperties.getMinimumAccountBalance().` This overrides the low balance and causes the call to succeed, which is what we want to test.
This is why we want to set lower balance for the accounts when the property is enabled.

When the property is disabled, the default high balance (`DEFAULT_ACCOUNT_BALANCE`) is used, ensuring the payer balance is always sufficient and the tests pass as expected.

IvanKavaldzhiev: Got it.

web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AbstractAliasedAccountReadableKVState.java
```diff
@@ -155,7 +156,22 @@ private Supplier<Long> getAccountBalance(final Entity entity, final Optional<Lon
                         return 0L;
                     }
                 })
-                .orElseGet(() -> Objects.requireNonNullElse(entity.getBalance(), 0L)));
+                .orElseGet(() -> {
+                    final Long currentBalance = entity.getBalance();
+                    if (!mirrorNodeEvmProperties.isOverridePayerBalanceValidation()) {
+                        return currentBalance;
```
steven-sheehy:
```suggestion
                        return Objects.requireNonNullElse(currentBalance, 0L);
```

web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AbstractAliasedAccountReadableKVState.java
```diff
@@ -155,7 +156,22 @@ private Supplier<Long> getAccountBalance(final Entity entity, final Optional<Lon
...
+                    }
+
+                    final ContractCallContext context = ContractCallContext.get();
+                    final boolean isBalanceCall = ContractCallContext.get().isBalanceCall();
```
steven-sheehy:
```suggestion
                    final boolean isBalanceCall = context.isBalanceCall();
```

web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AbstractAliasedAccountReadableKVState.java
```diff
@@ -155,7 +156,22 @@ private Supplier<Long> getAccountBalance(final Entity entity, final Optional<Lon
...
+                    final long minimumBalance = mirrorNodeEvmProperties.getMinimumAccountBalance();
+
+                    if (!isBalanceCall && (currentBalance == null || currentBalance < minimumBalance)) {
+                        return mirrorNodeEvmProperties.getMinimumAccountBalance();
```
steven-sheehy:
```suggestion
                        return minimumBalance;
```

web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AbstractAliasedAccountReadableKVState.java
```diff
@@ -155,7 +156,22 @@ private Supplier<Long> getAccountBalance(final Entity entity, final Optional<Lon
...
+                    }
+                    context.setBalanceCall(false);
```
steven-sheehy: Why do we set it to false here? Is it because it's guaranteed the payer is always loaded first? Maybe worth adding a comment.

steven-sheehy (changes requested):

web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AbstractAliasedAccountReadableKVState.java
kselveliev: Addressed

web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AbstractAliasedAccountReadableKVState.java
kselveliev: Addressed

web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AbstractAliasedAccountReadableKVState.java
kselveliev: Addressed

web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AbstractAliasedAccountReadableKVState.java
kselveliev: I'm not sure whether the payer is loaded first.
Our goal is to always reset isBalanceCall to false.
It should only be true when the readable state is accessed through the CustomBalanceOperation flow. In all other cases, we must ensure it remains false.
If we don't reset it, a subsequent call might read the payer from state and return their actual balance, which could lead to an INSUFFICIENT_PAYER_BALANCE error.

I will add a comment.

web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AbstractAliasedAccountReadableKVState.java
```diff
@@ -155,7 +156,22 @@ private Supplier<Long> getAccountBalance(final Entity entity, final Optional<Lon
...
+                    final Long currentBalance = entity.getBalance();
+                    if (!mirrorNodeEvmProperties.isOverridePayerBalanceValidation()) {
+                        return Objects.requireNonNullElse(currentBalance, 0L);
+                    }
+
+                    final ContractCallContext context = ContractCallContext.get();
+                    final boolean isBalanceCall = context.isBalanceCall();
+                    final long minimumBalance = mirrorNodeEvmProperties.getMinimumAccountBalance();
+
+                    if (!isBalanceCall && (currentBalance == null || currentBalance < minimumBalance)) {
```
kselveliev: Will need to add `entity.getType() != CONTRACT ` here to cover corner case where if we call `address(this).balance` in contract it will return the overriden balance instead of the one in database

web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AbstractAliasedAccountReadableKVState.java
steven-sheehy: Then does it need to be set in a finally block so it occurs regardless?

kselveliev: It's better to be in finally block. Addressed

PR #11421, opened 2025-06-18
Fix `block.timestamp` to return the correct value

### LINKED ISSUE #11409: EVM calls against a historical state are using the "to" timestamp of a block, instead of the "from" timestamp
### Description

As stated in the title, a wrong block timestamp seems to be used for historical call execution.

The contract at `0xA27EceB53d076EF2E663de960A1a6193A5843C89` in the example below exposes a simple `block.timestamp` passthrough method (see "steps to reproduce for details").
When I run a call query against a past block, say 20888413, it seems to be using the "to" timestamp of that block. Example:
```
> curl https://testnet.mirrornode.hedera.com/api/v1/blocks/20888413
{"count":11,"hapi_version":"0.62.6","hash":"0xc87750d77b42521dbf63b1b97eaeec6dddac3de1f2e4cb9a38fec4f75c7f9526ffea18cc8e28c24b3a1038bba809abf5","name":"2025-06-13T13_25_22.039109000Z.rcd.gz","number":20888413,"previous_hash":"0x90a310e7ccecfb83795c1f7e57184e77e0cb1135f5c49a857d17f3731318fc7a236b22a684d9381b0570a69b40cbfbde","size":4245,"timestamp":{"from":"1749821122.039109000","to":"1749821123.832888516"},"gas_used":0,"logs_bloom":"0x"}

> curl https://testnet.mirrornode.hedera.com/api/v1/contracts/call -X POST --data '{"block": "0x13EBB5D", "data": "0x796b89b9", "from": "0x913537797aE5274B10eB88F63bcB530C617aD162", "to": "0xA27EceB53d076EF2E663de960A1a6193A5843C89", "value": 0}' -H 'Content-Type: application/json'
{"result":"0x00000000000000000000000000000000000000000000000000000000684c26c3"}
# 1749821123 decimal = the "to" timestamp
```
I've got another method there that returns a block number (signature `0x42cbb15c`) and it returns an expected value of 0x13EBB5D, which indicates that the correct record file is loaded into the call context.

Potential cause:
> We currently have two entirely different code paths for EVM calls. On testnet it's our new modularized flow that uses hedera-app library while on mainnet it's the old mono code. The StaticBlockMetaSource is from mono. In mod, BlockInfoSingleton is the source. Looks like we probably have a bug in our singleton.

Related issue: https://github.com/hiero-ledger/hiero-consensus-node/issues/17162

### Steps to reproduce

1. Deploy a sample contract https://github.com/hashgraph/hedera-smart-contracts/blob/main/contracts/solidity/block/BlockInfo.sol
2. Call `getBlockTimestamp()` on a historical block (using `/v1/contracts/call`)
3. Observe that the result matches the "to" timestamp, instead of the "from" timestamp of the specified block

### Additional context

_No response_

### Hedera network

testnet

### Version

not sure

### Operating system

None

bilyana-gospodinova:
**Description**:
When calling `block.timestamp` the consensus start timestamp of the passed block needs to be returned. The consensus end timestamp was returned instead. This PR fixes this behaviour.

Fixes #11409 

**Checklist**

### DIFF
--- a/common/src/main/java/org/hiero/mirror/common/util/DomainUtils.java
+++ b/common/src/main/java/org/hiero/mirror/common/util/DomainUtils.java
@@ -218,6 +218,10 @@ public class DomainUtils {
         return convertToNanosMax(timestamp.getSeconds(), timestamp.getNanos());
     }
 
+    public static long getUnixSeconds(long nanoseconds) {
+        return nanoseconds / NANOS_PER_SECOND;
+    }
+
     /**
      * Cleans a string of invalid characters that would cause it to fail when inserted into the database. In particular,
      * PostgreSQL does not allow the null character (0x0000) to be inserted.
--- a/common/src/main/java/org/hiero/mirror/common/util/DomainUtils.java
+++ b/common/src/main/java/org/hiero/mirror/common/util/DomainUtils.java
@@ -218,10 +218,6 @@ public class DomainUtils {
         return convertToNanosMax(timestamp.getSeconds(), timestamp.getNanos());
     }
 
-    public static long getUnixSeconds(long nanoseconds) {
-        return nanoseconds / NANOS_PER_SECOND;
-    }
-
     /**
      * Cleans a string of invalid characters that would cause it to fail when inserted into the database. In particular,
      * PostgreSQL does not allow the null character (0x0000) to be inserted.
--- a/web3/src/main/java/org/hiero/mirror/web3/state/singleton/BlockInfoSingleton.java
+++ b/web3/src/main/java/org/hiero/mirror/web3/state/singleton/BlockInfoSingleton.java
@@ -29,7 +29,7 @@ public class BlockInfoSingleton implements SingletonState<BlockInfo> {
         return BlockInfo.newBuilder()
                 .blockHashes(Bytes.EMPTY)
                 .consTimeOfLastHandledTxn(endTimestamp)
-                .firstConsTimeOfCurrentBlock(endTimestamp)
+                .firstConsTimeOfCurrentBlock(startTimestamp)
                 .firstConsTimeOfLastBlock(startTimestamp)
                 .lastBlockNumber(recordFile.getIndex() - 1) // Library internally increments last by one for current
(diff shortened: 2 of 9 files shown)


kselveliev (dismissed):

IvanKavaldzhiev (approved):
LGTM

common/src/main/java/org/hiero/mirror/common/util/DomainUtils.java
```diff
@@ -218,6 +218,10 @@ public static Long timestampInNanosMax(Timestamp timestamp) {
         return convertToNanosMax(timestamp.getSeconds(), timestamp.getNanos());
     }
 
+    public static long getUnixSeconds(long nanoseconds) {
```
xin-hedera: Is this only used in tests? Then it should live in a test utility class.

bilyana-gospodinova: Correct, I moved the method into a test utility.

xin-hedera (approved):
LGTM

steven-sheehy (approved):
LGTM

web3/src/main/java/org/hiero/mirror/web3/state/singleton/BlockInfoSingleton.java
```diff
@@ -29,7 +29,7 @@ public BlockInfo get() {
         return BlockInfo.newBuilder()
                 .blockHashes(Bytes.EMPTY)
                 .consTimeOfLastHandledTxn(endTimestamp)
-                .firstConsTimeOfCurrentBlock(endTimestamp)
+                .firstConsTimeOfCurrentBlock(startTimestamp)
                 .firstConsTimeOfLastBlock(startTimestamp)
```
lukasz-hashgraph, line 33: Is it correct that both the `firstConsTimeOfCurrentBlock` and `firstConsTimeOfLastBlock` are now being set to the same value? Sounds like it's either one or the other that's wrong 🤔

PR #11423, opened 2025-06-18
Generating table usage report in java modules

### LINKED ISSUE #11286: In Java modules, research a way to intercept calls to repository broken down by endpoint
In Java modules, research a way to intercept calls to repository broken down by endpoint

filev94:
This PR introduces support for tracking database table usage by API endpoint during test execution. The goal is to generate a Markdown report per module after tests run, showing which repositories and tables are used by which endpoints. The said report is generated if one test, a whole test suite or all tests are ran in a module.

The logic is implemented in the common module and used across test scopes in all modules.

`ApiTrackingFilter.java `-  storing the current API endpoint URI 
`RepositoryUsageTrackerAspect.java `- class used to intercept the calls to repository and extract the table name that is used.
`TestExecutionTracker.java` - tracks whether tests are currently running. Main goal is to act as a rail guard so the report is generated only in test scope
`MarkdownReportGenerator.java` - class used to generate the table usage report
**Related issue(s)**:

Fixes #11286

**Notes for reviewer**:
There are custom repositories that dont implement CrudRepository, JpaRepository or any other Spring Data interface. In those cases, the current code is unable to extract the table (it is shown as UNKNOWN_TABLE), however, the repository and the method are logged in the report. For Jooq repositories, the data is double logged. Once with correct table name from the customImpl repository, and once with UNKNOWN_TABLE for the interface that extends JooqRepository. This behavior can also serve as a useful indicator for refactoring opportunities.
**Checklist**

### DIFF
--- a/common/src/test/java/org/hiero/mirror/common/aspect/RepositoryUsageTrackerAspect.java
+++ b/common/src/test/java/org/hiero/mirror/common/aspect/RepositoryUsageTrackerAspect.java
@@ -1,0 +1,124 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package org.hiero.mirror.common.aspect;
+
+import jakarta.persistence.EntityManager;
+import jakarta.persistence.metamodel.EntityType;
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
+import java.util.LinkedHashSet;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import org.aspectj.lang.ProceedingJoinPoint;
+import org.aspectj.lang.annotation.Around;
+import org.aspectj.lang.annotation.Aspect;
+import org.hiero.mirror.common.filter.ApiTrackingFilter;
+import org.hiero.mirror.common.util.TestExecutionTracker;
+import org.springframework.data.repository.CrudRepository;
+import org.springframework.data.repository.Repository;
+
+@Aspect
+@AllArgsConstructor
+public class RepositoryUsageTrackerAspect {
+
+    private static final String CRUD_REPOSITORY = "CrudRepository";
+    private static final String UNKNOWN_TABLE = "UNKNOWN_TABLE";
+
+    private final EntityManager entityManager;
+
+    @Getter
+    private static final Map<String, Map<String, Set<String>>> apiTableQueries = new ConcurrentHashMap<>();
+
+    @Around("execution(* org.hiero.mirror..repository.*.*(..))")
+    public Object trackRepositoryCall(final ProceedingJoinPoint joinPoint) throws Throwable {
+        if (!TestExecutionTracker.isTestRunning()) { // guard rail: only run during test execution
+            return joinPoint.proceed();
+        }
+
+        final var endpoint =
+                Optional.ofNullable(ApiTrackingFilter.getCurrentEndpoint()).orElse("UNKNOWN_ENDPOINT");
+
+        final var repository = joinPoint.getTarget();
+        final var tableName = resolveEntityTableName(repository);
+
+        var method = joinPoint.getSignature().toShortString();
+
+        if (method.contains(CRUD_REPOSITORY)) {
+            // If the invoked method belongs to CrudRepository (e.g., findById, deleteAll),
+            // it will appear as CrudRepository.methodName in the report.
+            // To improve traceability, CrudRepository is replaced with the actual repository class name
+            // that invoked the method.
+            method = method.replace(CRUD_REPOSITORY, getRepositoryName(repository));
+        }
+
+        apiTableQueries
+                .computeIfAbsent(endpoint, k -> new ConcurrentHashMap<>())
+                .computeIfAbsent(tableName, k -> new LinkedHashSet<>())
+                .add(method);
+
+        return joinPoint.proceed();
+    }
+
+    private String resolveEntityTableName(final Object repository) {
+        final var entityClass = extractEntityClassFromRepository(repository);
+        if (entityClass != null) {
+            return resolveJpaTableName(entityClass);
+        }
+        return UNKNOWN_TABLE;
+    }
+
+    private Class<?> extractEntityClassFromRepository(Object repositoryClass) {
+        for (Type iface : repositoryClass.getClass().getGenericInterfaces()) {
+            if (iface instanceof Class<?> ifaceClass) {
+                for (Type superIface : ifaceClass.getGenericInterfaces()) {
+                    if (superIface instanceof ParameterizedType paramType) {
+                        Type rawType = paramType.getRawType();
+                        if (rawType instanceof Class<?> rawClass && Repository.class.isAssignableFrom(rawClass)) {
+
+                            final var entityType = paramType.getActualTypeArguments()[0];
+                            if (entityType instanceof Class<?> entityClass) {
+                                return entityClass;
+                            }
+                        }
+                    }
+                }
+            }
+        }
+        return null;
+    }
+
+    private String convertCamelCaseToSnakeRegex(final String input) {
+        return input.replaceAll("([A-Z])(?=[A-Z])", "$1_")
+                .replaceAll("([a-z])([A-Z])", "$1_$2")
+                .toLowerCase();
+    }
+
+    private String resolveJpaTableName(final Class<?> entityClass) {
+        final var name = entityManager.getMetamodel().getEntities().stream()
+                .filter(e -> e.getJavaType().equals(entityClass))
+                .map(EntityType::getName)
+                .findFirst()
+                .orElse(UNKNOWN_TABLE);
+
+        return convertCamelCaseToSnakeRegex(name);
+    }
+
+    private String getRepositoryName(final Object repositoryInstance) {
+        final var genericInterfaces = repositoryInstance.getClass().getGenericInterfaces();
+
+        for (final var genericInterface : genericInterfaces) {
+            if (genericInterface instanceof Class<?> repoInterface) {
+
+                if (CrudRepository.class.isAssignableFrom(repoInterface)) {
+                    return repoInterface.getSimpleName();
+                }
+            }
+        }
+
+        return "UNKNOWN_REPOSITORY"; // Default fallback
+    }
+}
--- a/common/src/test/java/org/hiero/mirror/common/aspect/RepositoryUsageTrackerAspect.java
+++ b/common/src/test/java/org/hiero/mirror/common/aspect/RepositoryUsageTrackerAspect.java
@@ -5,7 +5,6 @@ package org.hiero.mirror.common.aspect;
 import jakarta.persistence.EntityManager;
 import jakarta.persistence.metamodel.EntityType;
 import java.lang.reflect.ParameterizedType;
-import java.lang.reflect.Type;
 import java.util.LinkedHashSet;
 import java.util.Map;
 import java.util.Optional;
@@ -71,12 +70,12 @@ public class RepositoryUsageTrackerAspect {
         return UNKNOWN_TABLE;
     }
 
-    private Class<?> extractEntityClassFromRepository(Object repositoryClass) {
-        for (Type iface : repositoryClass.getClass().getGenericInterfaces()) {
+    private Class<?> extractEntityClassFromRepository(final Object repositoryClass) {
+        for (final var iface : repositoryClass.getClass().getGenericInterfaces()) {
             if (iface instanceof Class<?> ifaceClass) {
-                for (Type superIface : ifaceClass.getGenericInterfaces()) {
+                for (final var superIface : ifaceClass.getGenericInterfaces()) {
                     if (superIface instanceof ParameterizedType paramType) {
-                        Type rawType = paramType.getRawType();
+                        final var rawType = paramType.getRawType();
                         if (rawType instanceof Class<?> rawClass && Repository.class.isAssignableFrom(rawClass)) {
 
                             final var entityType = paramType.getActualTypeArguments()[0];
--- a/common/src/test/java/org/hiero/mirror/common/config/TableUsageReportConfiguration.java
+++ b/common/src/test/java/org/hiero/mirror/common/config/TableUsageReportConfiguration.java
@@ -1,0 +1,23 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package org.hiero.mirror.common.config;
+
+import jakarta.persistence.EntityManager;
+import org.hiero.mirror.common.aspect.RepositoryUsageTrackerAspect;
+import org.hiero.mirror.common.filter.ApiTrackingFilter;
+import org.springframework.boot.test.context.TestConfiguration;
+import org.springframework.context.annotation.Bean;
+
+@TestConfiguration(proxyBeanMethods = false)
+public class TableUsageReportConfiguration {
+
+    @Bean
+    public RepositoryUsageTrackerAspect repositoryUsageTrackerAspect(EntityManager entityManager) {
+        return new RepositoryUsageTrackerAspect(entityManager);
+    }
+
+    @Bean
+    public ApiTrackingFilter apiTrackingFilter() {
+        return new ApiTrackingFilter();
+    }
+}
--- a/common/src/test/java/org/hiero/mirror/common/config/TableUsageReportConfiguration.java
+++ b/common/src/test/java/org/hiero/mirror/common/config/TableUsageReportConfiguration.java
@@ -12,7 +12,7 @@ import org.springframework.context.annotation.Bean;
 public class TableUsageReportConfiguration {
 
     @Bean
-    public RepositoryUsageTrackerAspect repositoryUsageTrackerAspect(EntityManager entityManager) {
+    public RepositoryUsageTrackerAspect repositoryUsageTrackerAspect(final EntityManager entityManager) {
         return new RepositoryUsageTrackerAspect(entityManager);
     }
 
--- a/common/src/test/java/org/hiero/mirror/common/filter/ApiTrackingFilter.java
+++ b/common/src/test/java/org/hiero/mirror/common/filter/ApiTrackingFilter.java
@@ -1,0 +1,29 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package org.hiero.mirror.common.filter;
+
+import jakarta.servlet.Filter;
+import jakarta.servlet.FilterChain;
+import jakarta.servlet.ServletException;
+import jakarta.servlet.ServletRequest;
+import jakarta.servlet.ServletResponse;
+import jakarta.servlet.http.HttpServletRequest;
+import java.io.IOException;
+
+public class ApiTrackingFilter implements Filter {
+
+    private static final ThreadLocal<String> currentEndpoint = new ThreadLocal<>();
+
+    @Override
+    public void doFilter(final ServletRequest request, final ServletResponse response, final FilterChain chain)
+            throws IOException, ServletException {
+        final var httpRequest = (HttpServletRequest) request;
+        currentEndpoint.set(httpRequest.getRequestURI());
+        chain.doFilter(request, response);
+        currentEndpoint.remove();
+    }
+
+    public static String getCurrentEndpoint() {
+        return currentEndpoint.get();
+    }
+}
--- a/common/src/test/java/org/hiero/mirror/common/util/MarkdownReportGenerator.java
+++ b/common/src/test/java/org/hiero/mirror/common/util/MarkdownReportGenerator.java
@@ -1,0 +1,39 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package org.hiero.mirror.common.util;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import lombok.CustomLog;
+import org.hiero.mirror.common.aspect.RepositoryUsageTrackerAspect;
+
+@CustomLog
+public class MarkdownReportGenerator {
+
+    public static void generateTableUsageReport() {
+        final var apiTableQueries = RepositoryUsageTrackerAspect.getApiTableQueries();
+        final var path = Paths.get("build/table-usage-report.md");
+
+        try (final var writer = Files.newBufferedWriter(path)) {
+            writer.write("# API Repository Calls & Database Tables Report\n\n");
+
+            for (final var entry : apiTableQueries.entrySet()) {
+                writer.write("## API Endpoint: " + entry.getKey() + "\n");
+
+                for (final var tableEntry : entry.getValue().entrySet()) {
+                    writer.write("### Table: " + tableEntry.getKey() + "\n");
+
+                    for (final var query : tableEntry.getValue()) {
+                        writer.write("- " + query + "\n");
+                    }
+                    writer.write("\n");
+                }
+                writer.write("\n");
+            }
+
+        } catch (IOException e) {
+            log.warn("Unexpected error occurred: {}", e.getMessage());
+        }
+    }
+}
--- a/common/src/test/java/org/hiero/mirror/common/util/TestExecutionTracker.java
+++ b/common/src/test/java/org/hiero/mirror/common/util/TestExecutionTracker.java
@@ -1,0 +1,28 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package org.hiero.mirror.common.util;
+
+import lombok.Getter;
+import org.junit.platform.engine.TestExecutionResult;
+import org.junit.platform.launcher.TestExecutionListener;
+import org.junit.platform.launcher.TestIdentifier;
+
+public class TestExecutionTracker implements TestExecutionListener {
+
+    @Getter
+    private static boolean testRunning = false;
+
+    @Override
+    public void executionStarted(TestIdentifier testIdentifier) {
+        if (testIdentifier.isTest()) {
+            testRunning = true;
+        }
+    }
+
+    @Override
+    public void executionFinished(TestIdentifier testIdentifier, TestExecutionResult result) {
+        if (testIdentifier.isTest()) {
+            testRunning = false;
+        }
+    }
+}
--- a/common/build.gradle.kts
+++ b/common/build.gradle.kts
@@ -20,6 +20,7 @@ dependencies {
     api("org.springframework.boot:spring-boot-starter-data-jpa")
     api("org.springframework.boot:spring-boot-starter-validation")
     api("org.web3j:core")
+    api("jakarta.servlet:jakarta.servlet-api")
     testImplementation("org.hyperledger.besu:evm")
     testImplementation("org.springframework.boot:spring-boot-testcontainers")
     testImplementation("org.testcontainers:postgresql")
--- a/common/src/test/java/org/hiero/mirror/common/GlobalTestSetup.java
+++ b/common/src/test/java/org/hiero/mirror/common/GlobalTestSetup.java
@@ -2,6 +2,8 @@
 
 package org.hiero.mirror.common;
 
+import org.hiero.mirror.common.util.MarkdownReportGenerator;
+import org.hiero.mirror.common.util.TestExecutionTracker;
 import org.junit.platform.engine.TestExecutionResult;
 import org.junit.platform.launcher.LauncherSession;
 import org.junit.platform.launcher.LauncherSessionListener;
@@ -14,7 +16,9 @@ public class GlobalTestSetup implements LauncherSessionListener, TestExecutionLi
 
     @Override
     public void launcherSessionOpened(LauncherSession session) {
-        session.getLauncher().registerTestExecutionListeners(this);
+        final var laucher = session.getLauncher();
+        laucher.registerTestExecutionListeners(this);
+        laucher.registerTestExecutionListeners(new TestExecutionTracker());
     }
 
     @Override
@@ -41,5 +45,10 @@ public class GlobalTestSetup implements LauncherSessionListener, TestExecutionLi
         var commonProperties = CommonProperties.getInstance();
         commonProperties.setShard(originalCommonProperties.getShard());
         commonProperties.setRealm(originalCommonProperties.getRealm());
+    }
+
+    @Override
+    public void launcherSessionClosed(LauncherSession session) {
+        MarkdownReportGenerator.generateTableUsageReport();
     }
 }
--- a/common/src/test/java/org/hiero/mirror/common/GlobalTestSetup.java
+++ b/common/src/test/java/org/hiero/mirror/common/GlobalTestSetup.java
@@ -16,9 +16,7 @@ public class GlobalTestSetup implements LauncherSessionListener, TestExecutionLi
 
     @Override
     public void launcherSessionOpened(LauncherSession session) {
-        final var laucher = session.getLauncher();
-        laucher.registerTestExecutionListeners(this);
-        laucher.registerTestExecutionListeners(new TestExecutionTracker());
+        session.getLauncher().registerTestExecutionListeners(this, new TestExecutionTracker());
     }
 
     @Override
--- a/common/src/test/java/org/hiero/mirror/common/config/CommonIntegrationTest.java
+++ b/common/src/test/java/org/hiero/mirror/common/config/CommonIntegrationTest.java
@@ -18,7 +18,7 @@ import org.springframework.cache.CacheManager;
 import org.springframework.context.annotation.Import;
 import org.springframework.test.context.jdbc.Sql;
 
-@Import(CommonTestConfiguration.class)
+@Import({CommonTestConfiguration.class, TableUsageReportConfiguration.class})
 @Sql(executionPhase = Sql.ExecutionPhase.BEFORE_TEST_METHOD, scripts = "classpath:cleanup.sql")
 @Sql(executionPhase = Sql.ExecutionPhase.AFTER_TEST_METHOD, scripts = "classpath:cleanup.sql")
 @SpringBootTest


common/src/test/java/org/hiero/mirror/common/aspect/RepositoryUsageTrackerAspect.java
```diff
@@ -0,0 +1,147 @@
...
+        if (entityClass != null) {
+            return resolveJpaTableName(entityClass);
+        }
+        return UNKNOWN_TABLE;
+    }
+
+    private Class<?> extractEntityClassFromRepository(final Object repositoryInstance) {
+        final var interfaces = repositoryInstance.getClass().getGenericInterfaces();
+
+        for (final var iface : interfaces) {
```
IvanKavaldzhiev: I think a better name would be `interface`

common/src/test/java/org/hiero/mirror/common/aspect/RepositoryUsageTrackerAspect.java
```diff
@@ -0,0 +1,147 @@
...
+        final var genericInterfaces = repositoryInstance.getClass().getGenericInterfaces();
+
+        for (final var genericInterface : genericInterfaces) {
+            if (genericInterface instanceof Class<?> repoInterface
+                    && CrudRepository.class.isAssignableFrom(repoInterface)) {
+                return repoInterface.getSimpleName();
+            }
+        }
+
+        return "UNKNOWN_REPOSITORY"; // Default fallback
```
IvanKavaldzhiev: Comment can be on the top line

common/src/test/java/org/hiero/mirror/common/util/TestExecutionTracker.java
```diff
@@ -0,0 +1,28 @@
...
+import org.junit.platform.launcher.TestExecutionListener;
+import org.junit.platform.launcher.TestIdentifier;
+
+public class TestExecutionTracker implements TestExecutionListener {
+
+    @Getter
+    private static boolean testRunning;
+
+    @Override
+    public void executionStarted(TestIdentifier testIdentifier) {
```
IvanKavaldzhiev: Argument can be final

common/src/test/java/org/hiero/mirror/common/util/TestExecutionTracker.java
```diff
@@ -0,0 +1,28 @@
...
+        if (testIdentifier.isTest()) {
+            testRunning = true;
+        }
+    }
+
+    @Override
+    public void executionFinished(TestIdentifier testIdentifier, TestExecutionResult result) {
```
IvanKavaldzhiev: All arguments can be final

common/src/test/java/org/hiero/mirror/common/config/TableUsageReportConfiguration.java
```diff
@@ -0,0 +1,23 @@
...
+import org.hiero.mirror.common.aspect.RepositoryUsageTrackerAspect;
+import org.hiero.mirror.common.filter.ApiTrackingFilter;
+import org.springframework.boot.test.context.TestConfiguration;
+import org.springframework.context.annotation.Bean;
+
+@TestConfiguration(proxyBeanMethods = false)
+public class TableUsageReportConfiguration {
+
+    @Bean
+    public RepositoryUsageTrackerAspect repositoryUsageTrackerAspect(final EntityManager entityManager) {
```
bilyana-gospodinova, line 15: These beans are defined but the only methods that are called on the instances are static getter methods.

common/src/test/java/org/hiero/mirror/common/aspect/RepositoryUsageTrackerAspect.java
```diff
@@ -0,0 +1,147 @@
...
+            // If the invoked method belongs to CrudRepository (e.g., findById, deleteAll),
+            // it will appear as CrudRepository.methodName in the report.
+            // To improve traceability, CrudRepository is replaced with the actual repository class name
+            // that invoked the method.
+            method = method.replace(CRUD_REPOSITORY, getRepositoryName(repository));
+        }
+
+        apiTableQueries
+                .computeIfAbsent(endpoint, k -> new ConcurrentHashMap<>())
+                .computeIfAbsent(tableName, k -> new LinkedHashSet<>())
```
kselveliev: If we are aiming for thread safety here should this become -> 
```
apiTableQueries
    .computeIfAbsent(endpoint, k -> new ConcurrentHashMap<>())
    .computeIfAbsent(tableName, k -> ConcurrentHashMap.newKeySet())
    .add(method);
```
assuming we don't care about order since new LinkedHashSet<>() is not thread safe?

common/src/test/java/org/hiero/mirror/common/aspect/RepositoryUsageTrackerAspect.java
filev94: `interface` is a reserved keyword and can't be used as variable name

common/src/test/java/org/hiero/mirror/common/aspect/RepositoryUsageTrackerAspect.java
filev94: Addressed

common/src/test/java/org/hiero/mirror/common/util/TestExecutionTracker.java
filev94: Addressed

common/src/test/java/org/hiero/mirror/common/util/TestExecutionTracker.java
filev94: Addressed

common/src/test/java/org/hiero/mirror/common/aspect/RepositoryUsageTrackerAspect.java
filev94: Addressed

kselveliev (approved):

common/src/test/java/org/hiero/mirror/common/aspect/RepositoryUsageTrackerAspect.java
```diff
@@ -0,0 +1,148 @@
...
+@AllArgsConstructor
+public class RepositoryUsageTrackerAspect {
+
+    private static final String CRUD_REPOSITORY = "CrudRepository";
+    private static final String UNKNOWN_TABLE = "UNKNOWN_TABLE";
+
+    private final EntityManager entityManager;
+
+    @Getter
+    private static final Map<String, Map<String, Set<String>>> apiTableQueries = new ConcurrentHashMap<>();
```
xin-hedera: per our coding style and Codacy analysis, the variable name should be `API_TABLE_QUERIES` and moved up to the section with other `private static final` variables.

common/src/test/java/org/hiero/mirror/common/aspect/RepositoryUsageTrackerAspect.java
```diff
@@ -0,0 +1,148 @@
...
+
+        apiTableQueries
+                .computeIfAbsent(endpoint, k -> new ConcurrentHashMap<>())
+                .computeIfAbsent(tableName, k -> ConcurrentHashMap.newKeySet())
+                .add(method);
+
+        return joinPoint.proceed();
+    }
+
+    private String resolveEntityTableName(final Object repository) {
```
xin-hedera: this works for simple repository methods with either generated queries or simple queries, however not the methods with complicated query in the `Query` annotation which actually uses multiple tables. There are plenty of such methods in web3, e.g.,  `AccountBalanceRepository.findHistoricalAccountBalanceUpToTimestamp`

common/src/test/java/org/hiero/mirror/common/aspect/RepositoryUsageTrackerAspect.java
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    private String resolveJpaTableName(final Class<?> entityClass) {
+        final var name = entityManager.getMetamodel().getEntities().stream()
+                .filter(e -> e.getJavaType().equals(entityClass))
+                .map(EntityType::getName)
+                .findFirst()
+                .orElse(UNKNOWN_TABLE);
+
+        return convertCamelCaseToSnakeRegex(name);
```
xin-hedera: use guava `CaseFormat` to do the conversion, see importer `Utility.toSnakeCase`

common/src/test/java/org/hiero/mirror/common/aspect/RepositoryUsageTrackerAspect.java
```diff
@@ -0,0 +1,148 @@
...
+
+    private static final String CRUD_REPOSITORY = "CrudRepository";
+    private static final String UNKNOWN_TABLE = "UNKNOWN_TABLE";
+
+    private final EntityManager entityManager;
+
+    @Getter
+    private static final Map<String, Map<String, Set<String>>> apiTableQueries = new ConcurrentHashMap<>();
+
+    @Around("execution(* org.hiero.mirror..repository.*.*(..))")
```
xin-hedera: this doesn't work with

- generated upsert queries in importer, e.g., `UpsertQueryGenerator`; maybe it's fine, we can maintain it manually and these only change when we add new tables to persist new domain objects
- `JooqRepository` in rest-java

common/src/test/java/org/hiero/mirror/common/util/TestExecutionTracker.java
```diff
@@ -0,0 +1,28 @@
...
+
+import lombok.Getter;
+import org.junit.platform.engine.TestExecutionResult;
+import org.junit.platform.launcher.TestExecutionListener;
+import org.junit.platform.launcher.TestIdentifier;
+
+public class TestExecutionTracker implements TestExecutionListener {
+
+    @Getter
+    private static boolean testRunning;
```
xin-hedera: the static variable may get accessed by different concurrent test worker threads and then creates issues.

filev94: Changed to ThreadLocal<Boolean> testRunning = ThreadLocal.withInitial(() -> false);

common/src/test/java/org/hiero/mirror/common/aspect/RepositoryUsageTrackerAspect.java
filev94: Addressed

common/src/test/java/org/hiero/mirror/common/aspect/RepositoryUsageTrackerAspect.java
filev94: Addressed

common/src/test/java/org/hiero/mirror/common/aspect/RepositoryUsageTrackerAspect.java
filev94: Added code to check for @Query annotations and extractTableNamesFromSql. Tested with complex queries:
```
updateEntityStake
findHistoricalAccountBalanceUpToTimestamp
```

 and others

It has limitations, of course. It will not be able to capture things like

```
var namedParameterJdbcTemplate = new NamedParameterJdbcTemplate(jdbcTemplate);
        namedParameterJdbcTemplate.update(CREATE_ENTITY_STATE_START_SQL, params);
        jdbcTemplate.execute(CREATE_TABLE_INDEX_DDL);
```
                     
  where CREATE_ENTITY_STATE_START_SQL is a constant in the custom repo class.
  
  There are cases where CTE name shadows real table: 

  ```
with nft_history as (
              insert into nft_history (account_id, created_timestamp, delegating_spender, deleted, metadata,
                serial_number, spender, token_id, timestamp_range)
```
                
and additional handling would be required for such cases
                
Tried using third party library (jsqlparser) to parse the sql but it actually scans the queries and checks them if they are correct or not and fails.

File: common/src/test/java/org/hiero/mirror/common/aspect/RepositoryUsageTrackerAspect.java (hunk 10 above)
Comment (filev94) [reply common/src/test/java/org/hiero/mirror/common/aspect/RepositoryUsageTrackerAspect.java]: For JooqRepository it double logs the data like this: 

```
### Table: nft_allowance
- NftAllowanceRepositoryCustom.findAll(..)

### Table: UNKNOWN_TABLE
- NftAllowanceRepositoryCustomImpl.findAll(..)
```

For upsert queries in importer - you are correct. Some custom solution is probably needed for them.

File: common/src/test/java/org/hiero/mirror/common/filter/ApiTrackingFilter.java [hunk 12]
```diff
@@ -0,0 +1,29 @@
...
+import jakarta.servlet.FilterChain;
+import jakarta.servlet.ServletException;
+import jakarta.servlet.ServletRequest;
+import jakarta.servlet.ServletResponse;
+import jakarta.servlet.http.HttpServletRequest;
+import java.io.IOException;
+
+public class ApiTrackingFilter implements Filter {
+
+    private static final ThreadLocal<String> currentEndpoint = new ThreadLocal<>();
```
steven-sheehy: Does this work fine with web3 even though we use virtual threads and ScopedValue? I believe it should but please double check.

common/src/test/java/org/hiero/mirror/common/filter/ApiTrackingFilter.java
```diff
@@ -0,0 +1,29 @@
...
+
+import jakarta.servlet.Filter;
+import jakarta.servlet.FilterChain;
+import jakarta.servlet.ServletException;
+import jakarta.servlet.ServletRequest;
+import jakarta.servlet.ServletResponse;
+import jakarta.servlet.http.HttpServletRequest;
+import java.io.IOException;
+
+public class ApiTrackingFilter implements Filter {
```
steven-sheehy: I think some controllers use mocks for the service layer like web3. So this wouldn't catch all the possible tables per API. Should we attempt to add some integration controller test that stresses all the repositories? Or fake it by setting the endpoint inside the contract call service integration tests?

common/src/test/java/org/hiero/mirror/common/config/TableUsageReportConfiguration.java
```diff
@@ -0,0 +1,23 @@
...
+        return new RepositoryUsageTrackerAspect(entityManager);
+    }
+
+    @Bean
+    public ApiTrackingFilter apiTrackingFilter() {
```
steven-sheehy: Can make all classes and methods non-public. Can make class final.

common/src/test/java/org/hiero/mirror/common/aspect/RepositoryUsageTrackerAspect.java
```diff
@@ -0,0 +1,298 @@
...
+     * @return the snake_case table name, or a default unknown table constant if not resolvable
+     */
+    private String resolveJpaTableName(final Class<?> entityClass) {
+        final var name = entityManager.getMetamodel().getEntities().stream()
+                .filter(e -> e.getJavaType().equals(entityClass))
+                .map(EntityType::getName)
+                .findFirst()
+                .orElse(UNKNOWN_TABLE);
+
+        return toSnakeCase(name);
```
steven-sheehy: Could add `map(this::toSnakeCase)` to the stream and avoid the var name.

common/src/test/java/org/hiero/mirror/common/aspect/RepositoryUsageTrackerAspect.java
```diff
@@ -0,0 +1,298 @@
...
+        final var endpoint =
+                Optional.ofNullable(ApiTrackingFilter.getCurrentEndpoint()).orElse("UNKNOWN_ENDPOINT");
+
+        final var repository = joinPoint.getTarget();
+        final var repositoryClass = repository.getClass();
+        final var methodName = joinPoint.getSignature().getName();
+        final var methodArgs = joinPoint.getArgs();
+
+        // Find the repository interface method that matches the join point method
+        final var repositoryMethod = findRepositoryMethod(repositoryClass, methodName, methodArgs);
```
steven-sheehy: Using reflection to intercept and lookup repository info is brittle. We should leverage Spring specific technologies to offload this behavior so that it's maintainable, even if Spring internally might use reflection. We can use `RepositoryProxyPostProcessor` to get the `RepositoryInformation` class and intercept calls using that info. See https://stackoverflow.com/questions/26258158/how-to-instrument-advice-a-spring-data-jpa-repository. If you need additional metadata you can use `applicationContext.getBeansOfType(JpaRepositoryFactoryBean.class)` to gain access to additional JPA metadata.

steven-sheehy (changes requested):

common/src/test/java/org/hiero/mirror/common/util/MarkdownReportGenerator.java
```diff
@@ -0,0 +1,48 @@
...
+
+    public static void generateTableUsageReport() {
+        final var apiTableQueries = RepositoryUsageTrackerAspect.getAPI_TABLE_QUERIES();
+        final var path = Paths.get("build/table-usage-report.md");
+
+        try (final var writer = Files.newBufferedWriter(path)) {
+            writer.write("# API Repository Calls & Database Tables Report" + NEWLINE + NEWLINE);
+
+            for (final var entry : apiTableQueries.entrySet()) {
+                writer.write("## API Endpoint: " + entry.getKey() + NEWLINE);
```
steven-sheehy: Would prefer a more succinct report:

```
# Table Usage Report

| Endpoint | Table | Source        |
------------------------------------
| /foo     | bar   | barRepo.baz() |
```

common/src/test/java/org/hiero/mirror/common/config/TableUsageReportConfiguration.java
filev94: Addressed

common/src/test/java/org/hiero/mirror/common/aspect/RepositoryUsageTrackerAspect.java
filev94: Addressed. Thought it would be good idea to put it in DomainUtils instead of being private method in this class.

common/src/test/java/org/hiero/mirror/common/aspect/RepositoryUsageTrackerAspect.java
filev94: Used ProxyPostProcessor instead of reflection. Created TrackingRepositoryProxyPostProcesso, TrackingRepositoryFactoryBean and RepositoryUsageInterceptor. This handles well the repository calls, even jooq repositories - they are not double logged now. 

Regarding jdbcTemplate - i am working on a JdbcTemplateTrackingAspect (haven't pushed that change) but currently it logs too many things that are not relevant and I am trying to see if they can be removed.

common/src/test/java/org/hiero/mirror/common/util/MarkdownReportGenerator.java
filev94: Changed to suggested report. One thing to note, if one endpoint has many tables, I've preferred not to put the same endpoint name in all the cells in the Endpoint column, but rather chose to leave them blank.

image

common/src/test/java/org/hiero/mirror/common/filter/ApiTrackingFilter.java
filev94: Probably adding integration controller tests would be the best course of action. If we fake the endpoint, tomorrow we have to remember that this was done and this might cause us some unforeseen issues. 

Regarding the integration controller tests - I'd prefer if they are not in this PR, so i went ahead and created a task for this -> https://github.com/hiero-ledger/hiero-mirror-node/issues/11504

Also, the issue with unknown_endpoint is common for all modules.

common/src/test/java/org/hiero/mirror/common/filter/ApiTrackingFilter.java
filev94: Tried using 

` private static final ScopedValue<String> CURRENT_ENDPOINT = ScopedValue.newInstance();` 

and

```
ScopedValue.where(CURRENT_ENDPOINT, uri).run(() -> {
            try {
                chain.doFilter(request, response);
            } catch (IOException | ServletException e) {
                throw new RuntimeException(e);
            }
        });
```

and

```
public static String getCurrentEndpoint() {
        return CURRENT_ENDPOINT.isBound()
                ? CURRENT_ENDPOINT.get()
                : "UNKNOWN_ENDPOINT";
    }

```

here and running the tests, the report that was generated in web3 was the same as with ThreadLocal. Not sure if ScopedValue will work with the other modules -> there the application context fails to load with different issues and I haven't figured out how to fix it as of now.

steven-sheehy: ScopedValue requires `--enable-preview` JVM flag, so we wouldn't want to enable that outside web3. We can just use ThreadLocal just wanted to verify it works.

common/src/test/java/org/hiero/mirror/common/filter/ApiTrackingFilter.java
steven-sheehy: Thinking further on it, adding additional integration tests that somehow touch every table would be a lot of work and duplicate test setup. Think the other approach would be quicker and easier.

common/src/test/java/org/hiero/mirror/common/config/TrackingRepositoryFactoryBean.java
```diff
@@ -0,0 +1,29 @@
...
+
+package org.hiero.mirror.common.config;
+
+import jakarta.persistence.EntityManager;
+import java.io.Serializable;
+import org.springframework.data.jpa.repository.JpaRepository;
+import org.springframework.data.jpa.repository.support.JpaRepositoryFactoryBean;
+import org.springframework.data.repository.core.support.RepositoryFactorySupport;
+
+public final class TrackingRepositoryFactoryBean<R extends JpaRepository<T, I>, T, I extends Serializable>
```
steven-sheehy:
```suggestion
final class TrackingRepositoryFactoryBean<R extends JpaRepository<T, I>, T, I extends Serializable>
```

common/src/test/java/org/hiero/mirror/common/config/TableUsageReportConfiguration.java
```diff
@@ -0,0 +1,24 @@
...
+
+    import jakarta.persistence.EntityManager;
+    import org.hiero.mirror.common.filter.ApiTrackingFilter;
+    import org.springframework.boot.test.context.TestConfiguration;
+    import org.springframework.context.annotation.Bean;
+    import org.springframework.data.jpa.repository.config.EnableJpaRepositories;
+
+    @TestConfiguration(proxyBeanMethods = false)
+    @EnableJpaRepositories(basePackages = "org.hiero.mirror", repositoryFactoryBeanClass = TrackingRepositoryFactoryBean.class)
+    public class TableUsageReportConfiguration {
```
steven-sheehy:
```suggestion
    class TableUsageReportConfiguration {
```

common/src/test/java/org/hiero/mirror/common/config/TrackingRepositoryProxyPostProcessor.java
```diff
@@ -0,0 +1,21 @@
...
+
+import jakarta.persistence.EntityManager;
+import lombok.AllArgsConstructor;
+import org.hiero.mirror.common.interceptor.RepositoryUsageInterceptor;
+import org.springframework.aop.framework.ProxyFactory;
+import org.springframework.data.repository.core.RepositoryInformation;
+import org.springframework.data.repository.core.support.RepositoryProxyPostProcessor;
+
+@AllArgsConstructor
+public final class TrackingRepositoryProxyPostProcessor implements RepositoryProxyPostProcessor {
```
steven-sheehy:
```suggestion
@RequiredArgsConstructor
final class TrackingRepositoryProxyPostProcessor implements RepositoryProxyPostProcessor {
```

common/src/test/java/org/hiero/mirror/common/interceptor/RepositoryUsageInterceptor.java
```diff
@@ -0,0 +1,165 @@
...
+import org.hiero.mirror.common.util.DomainUtils;
+import org.hiero.mirror.common.util.SqlParsingUtil;
+import org.hiero.mirror.common.util.TestExecutionTracker;
+import org.springframework.data.jpa.repository.Query;
+import org.springframework.data.repository.core.RepositoryInformation;
+
+@CustomLog
+@AllArgsConstructor
+public class RepositoryUsageInterceptor implements MethodInterceptor {
+    private static final Map<String, Map<String, Set<String>>> API_TABLE_QUERIES = new ConcurrentHashMap<>();
```
steven-sheehy, line 28: We want the final report output to be sorted to make it easier to read. We also want it grouped by different criteria. It might be easier to just have a `List<TableUsage>` then do the grouping/sorting in the report generator.

common/src/test/java/org/hiero/mirror/common/util/MarkdownReportGenerator.java
steven-sheehy: We should actually group by endpoint name and join table names into a single row by comma or newline.

common/src/main/java/org/hiero/mirror/common/util/DomainUtils.java
```diff
@@ -332,6 +333,13 @@ public static boolean isLongZeroAddress(byte[] evmAddress) {
         return Arrays.equals(MIRROR_PREFIX, 0, 12, evmAddress, 0, 12);
     }
 
+    public static String toSnakeCase(final String text) {
```
bilyana-gospodinova, line 336: Can you add a junit test for this method?

common/src/test/java/org/hiero/mirror/common/util/MarkdownReportGenerator.java
```diff
@@ -0,0 +1,82 @@
...
+        }
+    }
+
+    private static String buildMethodList(final Set<String> methods) {
+        final var sb = new StringBuilder();
+        for (final var method : methods) {
+            sb.append("- ").append(escapeMarkdown(method)).append("<br>");
+        }
+        // Remove last <br>
+        if (sb.length() > 4) {
```
bilyana-gospodinova: It will be cleaner if you extract `"<br>"` into a variable (let's say `br`) an in here you call `br.length()` instead of a hardcoded number.

common/src/test/java/org/hiero/mirror/common/interceptor/RepositoryUsageInterceptor.java
```diff
@@ -0,0 +1,165 @@
...
+        var tableNames = (sql != null) ? SqlParsingUtil.extractTableNamesFromSql(sql) : Set.<String>of();
+
+        // Fallback to entity class if no table names found
+        if (tableNames.isEmpty()) {
+            final var entityClass = repositoryInformation.getDomainType();
+            final var tableName = resolveJpaTableName(entityClass);
+            tableNames = Set.of(tableName);
+        }
+
+        final String methodSignature = createMethodSignature(invocation);
```
IvanKavaldzhiev: Can be var

common/src/test/java/org/hiero/mirror/common/interceptor/RepositoryUsageInterceptor.java
```diff
@@ -0,0 +1,165 @@
...
+
+        // Track method under each resolved table for the current endpoint
+        final var endpointMap = API_TABLE_QUERIES.computeIfAbsent(endpoint, k -> new ConcurrentHashMap<>());
+        for (var tableName : tableNames) {
```
IvanKavaldzhiev: tableName can be final

common/src/test/java/org/hiero/mirror/common/interceptor/RepositoryUsageInterceptor.java
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    /**
+     * Resolves the JPA table name for the given entity class by querying the JPA metamodel.
+     * Converts the entity name to snake_case.
+     *
+     * @param entityClass the JPA entity class
+     * @return the resolved table name, or {@code "UNKNOWN_TABLE"} if not found
+     */
+    private String resolveJpaTableName(Class<?> entityClass) {
```
IvanKavaldzhiev: Argument can be final

common/src/test/java/org/hiero/mirror/common/util/SqlParsingUtil.java
```diff
@@ -0,0 +1,64 @@
...
+            while (matcher.find()) {
+                final var table = matcher.group(1) != null ? matcher.group(1) : matcher.group(2);
+                if (table != null && !cteNames.contains(table)) {
+                    tables.add(table);
+                }
+            }
+        }
+
+        // Handle CREATE INDEX ON table
+        var createIndexMatcher = CREATE_INDEX_PATTERN.matcher(loweredSql);
```
IvanKavaldzhiev: Can be final

filev94: Addressed

common/src/test/java/org/hiero/mirror/common/interceptor/RepositoryUsageInterceptor.java
filev94: Addressed

common/src/test/java/org/hiero/mirror/common/interceptor/RepositoryUsageInterceptor.java
filev94: Addressed

common/src/test/java/org/hiero/mirror/common/interceptor/RepositoryUsageInterceptor.java
filev94: Addressed

common/src/test/java/org/hiero/mirror/common/util/MarkdownReportGenerator.java
filev94: I created a constant for br , but since I made some changes in regards to [comment](https://github.com/hiero-ledger/hiero-mirror-node/pull/11423#discussion_r2178611470 ), this method was no longer used so I deleted it.

common/src/main/java/org/hiero/mirror/common/util/DomainUtils.java
filev94: Addressed

common/src/test/java/org/hiero/mirror/common/config/TrackingRepositoryFactoryBean.java
filev94: Addressed

common/src/test/java/org/hiero/mirror/common/config/TableUsageReportConfiguration.java
filev94: Addressed

common/src/test/java/org/hiero/mirror/common/config/TrackingRepositoryProxyPostProcessor.java
filev94: Addressed

common/src/test/java/org/hiero/mirror/common/interceptor/RepositoryUsageInterceptor.java
filev94: I made changes to report generator to group/sort -> group by endpoint and second table -> group by table names

common/src/test/java/org/hiero/mirror/common/util/MarkdownReportGenerator.java
filev94: Addressed, also created 2 csv files -> one with sources and one without

common/src/test/java/org/hiero/mirror/common/filter/ApiTrackingFilter.java
filev94: Faked the endpoints in integration tests

common/src/test/java/org/hiero/mirror/common/config/GrpcTestConfiguration.java
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package org.hiero.mirror.common.config;
+
+import org.hiero.mirror.common.interceptor.GrpcInterceptor;
+import org.springframework.boot.test.context.TestConfiguration;
+import org.springframework.context.annotation.Bean;
+
+@TestConfiguration
+public class GrpcTestConfiguration {
```
steven-sheehy: I don't like the idea of leaking grpc and graphql specific logic into the common module. Can we please move any grpc or graphql specific classes in common into their respective API specific modules?

web3/src/test/java/org/hiero/mirror/web3/service/OpcodeServiceTest.java
```diff
@@ -64,6 +65,11 @@ class OpcodeServiceTest extends AbstractContractCallServiceOpcodeTracerTest {
     private final OpcodeService opcodeService;
     private final EvmEncodingFacade evmEncoder;
 
+    @BeforeEach
+    void configure() {
+        setOpcodeEndpoint();
```
IvanKavaldzhiev, line 70: Shouldn't we reset the endpoint in @ AfterEach method?

filev94: Since this class has a parent - AbstractContractCallServiceTest - it is cleared in the @AfterEach there

common/src/test/java/org/hiero/mirror/common/util/Constants.java
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package org.hiero.mirror.common.util;
+
+import lombok.experimental.UtilityClass;
+
+@UtilityClass
+public class Constants {
```
steven-sheehy: I think we can delete this class and move these constants into EndpointContext

common/src/test/java/org/hiero/mirror/common/interceptor/GrpcInterceptor.java
```diff
@@ -0,0 +1,23 @@
...
+import net.devh.boot.grpc.server.interceptor.GrpcGlobalServerInterceptor;
+import org.hiero.mirror.common.util.EndpointContext;
+
+@GrpcGlobalServerInterceptor
+public class GrpcInterceptor implements ServerInterceptor {
+
+    @Override
+    public <ReqT, RespT> ServerCall.Listener<ReqT> interceptCall(
+            ServerCall<ReqT, RespT> call, Metadata headers, ServerCallHandler<ReqT, RespT> next) {
+        final var fullMethod = call.getMethodDescriptor().getFullMethodName();
```
steven-sheehy: We should also add the service name so it looks like `NetworkService/getNodes`

common/src/test/java/org/hiero/mirror/common/util/CsvGenerator.java
```diff
@@ -0,0 +1,61 @@
...
+                    }
+                }
+            }
+        } catch (final IOException e) {
+            throw new UncheckedIOException(e);
+        }
+    }
+
+    private static String escapeCsv(String value) {
+        if (value == null) return "";
```
steven-sheehy: Note sure why spotless isn't flagging this.

```suggestion
        if (value == null) {
            return "";
        }
```

web3/src/test/java/org/hiero/mirror/web3/service/AbstractContractCallServiceOpcodeTracerTest.java
```diff
@@ -45,6 +47,10 @@
 
 abstract class AbstractContractCallServiceOpcodeTracerTest extends AbstractContractCallServiceHistoricalTest {
 
+    protected static void setOpcodeEndpoint() {
```
steven-sheehy, line 50: Remove calls to `setOpcodeEndpoint()` and just let junit invoke it automatically. If this doesn't work change to non-static `@BeforeEach`.

```suggestion
    @BeforeAll
    private static void setOpcodeEndpoint() {
```

graphql/src/test/java/org/hiero/mirror/graphql/controller/AccountControllerTest.java
```diff
@@ -7,16 +7,22 @@
...
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.CsvSource;
 import org.springframework.boot.test.autoconfigure.graphql.tester.AutoConfigureHttpGraphQlTester;
+import org.springframework.boot.test.context.SpringBootTest;
 import org.springframework.graphql.ResponseError;
 import org.springframework.graphql.test.tester.HttpGraphQlTester;
 
+@SpringBootTest(
+        webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT,
+        classes = {GraphqlApplication.class, GraphqlTestConfiguration.class})
```
steven-sheehy: We should add to `GraphqlIntegrationTest` or create an `AbstractControllerTest`

common/src/test/java/org/hiero/mirror/common/util/EndpointNormalizer.java
```diff
@@ -0,0 +1,33 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package org.hiero.mirror.common.util;
+
+import java.util.Map;
+import java.util.regex.Pattern;
+import lombok.experimental.UtilityClass;
+
+@UtilityClass
+public final class EndpointNormalizer {
```
steven-sheehy, line 10: Remove this class and use BEST_MATCHING_PATTERN_ATTRIBUTE to get the URI pattern. See https://github.com/hiero-ledger/hiero-mirror-node/blob/f0cc1cd4d1ac5c56b229b900c0b8116e69dc93e4/web3/src/main/java/org/hiero/mirror/web3/config/MetricsFilter.java#L62

steven-sheehy (changes requested):

web3/src/test/java/org/hiero/mirror/web3/service/AbstractContractCallServiceOpcodeTracerTest.java
filev94: This is needed because there are test suites like ContractCallDynamicCallsTest where we would call both /contracts/call and /opcodes and if i do it in @ BeforeEach or @BeforeAll i would overwrite the /contracts/call.

common/src/test/java/org/hiero/mirror/common/config/GrpcTestConfiguration.java
filev94: Moved grpc and graphql specific files to their corresponding modules

common/src/test/java/org/hiero/mirror/common/util/Constants.java
filev94: Addressed

common/src/test/java/org/hiero/mirror/common/interceptor/GrpcInterceptor.java
filev94: Addressed

common/src/test/java/org/hiero/mirror/common/util/CsvGenerator.java
filev94: Addressed

graphql/src/test/java/org/hiero/mirror/graphql/controller/AccountControllerTest.java
filev94: Moved it to GraphqlIntegrationTest

common/src/test/java/org/hiero/mirror/common/util/EndpointNormalizer.java
filev94: Unfortunately BEST_MATCHING_PATTERN_ATTRIBUTE does not work with Filter and i had to implement a HandlerInterceptor, but then that came with other problems -> either importer application context would not start or it wouldn't connect to Redis.

steven-sheehy: > Unfortunately BEST_MATCHING_PATTERN_ATTRIBUTE does not work with Filter

Not sure what you mean since we're currently using it in the MetricsFilter linked above. But we can get this in and iterate on it.

steven-sheehy (approved):
LGTM

PR #11433, opened 2025-06-20
Add github action for release branches comparison

### LINKED ISSUE #11422: Create a mechanism to compare the copied from hedera-app files
### Problem

Currently, we have some classes that are copied from hedera-app and adapted for our needs. Such classes are:
- com.hedera.hapi.node.state.token.Account
- com.hedera.hapi.node.state.token.Token
- com.hedera.hapi.node.state.token.TokenRelation
- com.swirlds.state.spi.ReadableKVStateBase
- com.swirlds.state.spi.WritableKVStateBase
- com.hedera.hapi.node.state.file.File
- com.hedera.hapi.node.state.schedule.Schedule

Since they will not be removed soon, we need to have an automated way to verify if these files have been changed and flag if we need to adapt our implementations as well.

### Solution

We can calculate the files' checksums, save them somewhere (like a file, for example), take the latest files from upstream, calculate their checksums again and compare the values. This way if the checksums have changed, we will know that we need to check manually what are the new changes upstream and adapt our code if necessary.

### Alternatives

_No response_

sdimitrov9:
**Description**:
This PR adds github action that compares the following files of the consensus node if the hedera app version specified in the PR and the one of the base branch of the mirror node differentiate: 

  com.hedera.hapi.node.state.token.Account
  com.hedera.hapi.node.state.token.Token
  com.hedera.hapi.node.state.token.TokenRelation
  com.swirlds.state.spi.ReadableKVStateBase
  com.swirlds.state.spi.WritableKVStateBase
  com.hedera.hapi.node.state.file.File
  com.hedera.hapi.node.state.schedule.Schedule

The action is triggered on PRs changing the hedera app version in the build.gradle.kts file.
The steps are the following:
```
1. Checkout Code: Mirror node repo is cloned into the runner, so we can execute git relation operations.
2. Check if hedera app version changed in the PR: Checks the PR diff of the build.gradle.kts against the base branch. If hedera app dependency line is changed the consecutive steps are executed, otherwise they are skipped and the action passes.
3. Install JDK
4. Setup Gradle
5. Run diff script: Target files are compared and if differences are found the step fails.
```
The step doing the comparison timeouts after 30 min in case something hangs.

**Related issue(s)**:

Fixes #11422 

**Checklist**

### DIFF
--- a/.github/workflows/releases-comparator.yml
+++ b/.github/workflows/releases-comparator.yml
@@ -2,7 +2,6 @@
 
 name: Consensus node release branches compare
 on:
-  workflow_dispatch:
   pull_request:
     branches:
       - main
@@ -25,6 +24,11 @@ jobs:
       pr_version: ${{ steps.check.outputs.pr_version }}
       base_version: ${{ steps.check.outputs.base_version }}
     steps:
+      - name: Harden Runner
+        uses: step-security/harden-runner@002fdce3c6a235733a90a27c80493a3241e56863 # v2.12.1
+        with:
+          egress-policy: audit
+
       - name: Checkout Code
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
@@ -56,6 +60,11 @@ jobs:
     needs: check-version
     if: needs.check-version.outputs.changed == 'true'
     steps:
+      - name: Harden Runner
+        uses: step-security/harden-runner@002fdce3c6a235733a90a27c80493a3241e56863 # v2.12.1
+        with:
+          egress-policy: audit
+
       - name: Checkout Code
         uses: actions/checkout@v4.2.2
 
--- a/.github/workflows/releases-comparator.yml
+++ b/.github/workflows/releases-comparator.yml
@@ -21,16 +21,44 @@ jobs:
     name: Compare latest two releases of consensus node
     runs-on: hiero-mirror-node-linux-large
     steps:
+      - name: Checkout Code
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - name: Check if hedera app version changed in the PR
+        id: check-hedera-app-line
+        run: |
+          set -ex
+
+          PR_VERSION=$(grep -o 'api("com\.hedera\.hashgraph:app:[^"]*")' build.gradle.kts | sed -E 's/.*app:([^"]*)".*/\1/')
+          PR_VERSION_TAG="v$PR_VERSION"
+
+          git fetch origin ${{ github.base_ref }} --depth=1
+          git show origin/${{ github.base_ref }}:build.gradle.kts > base_build.gradle.kts
+          BASE_VERSION=$(grep -o 'api("com\.hedera\.hashgraph:app:[^"]*")' base_build.gradle.kts | sed -E 's/.*app:([^"]*)".*/\1/')
+          BASE_VERSION_TAG="v$BASE_VERSION"
+
+          echo "pr_version=$PR_VERSION_TAG" >> $GITHUB_OUTPUT
+          echo "base_version=$BASE_VERSION_TAG" >> $GITHUB_OUTPUT
+
+          if [ "$PR_VERSION_TAG" != "$BASE_VERSION_TAG" ]; then
+            echo "changed=true" >> $GITHUB_OUTPUT
+          else
+            echo "changed=false" >> $GITHUB_OUTPUT
+          fi
+
       - name: Install JDK
+        if: steps.check-hedera-app-line.outputs.changed == 'true'
         uses: actions/setup-java@c5195efecf7bdfc987ee8bae7a71cb8b11521c00 # v4.7.1
         with:
           distribution: "temurin"
           java-version: 21
 
       - name: Setup Gradle
+        if: steps.check-hedera-app-line.outputs.changed == 'true'
         uses: gradle/actions/setup-gradle@ac638b010cf58a27ee6c972d7336334ccaf61c96 # v4
 
       - name: Run diff script
+        if: steps.check-hedera-app-line.outputs.changed == 'true'
         run: |
           set -ex
 
@@ -50,31 +78,18 @@ jobs:
           TMP_DIR_ONE=$(mktemp -d)
           TMP_DIR_TWO=$(mktemp -d)
 
-          # Function to clean the temp directories after script finishes
-          cleanup() {
-            rm -rf "$TMP_DIR_ONE" "$TMP_DIR_TWO"
-          }
-          trap cleanup EXIT
-
-          TAGS=($(git ls-remote --tags "$CONSENSUS_NODE_REPO" | \
-                awk '{print $2}' | \
-                grep 'refs/tags/v0' | \
-                grep -v '\^{}' | \
-                sed 's#refs/tags/##' | \
-                sort -Vr))
-
-          TAG_ONE="${TAGS[0]}"
-          TAG_TWO="${TAGS[1]}"
+          TAG_ONE="${{ steps.check-hedera-app-line.outputs.pr_version }}"
+          TAG_TWO="${{ steps.check-hedera-app-line.outputs.base_version }}"
 
           echo "Comparing consensus node release versions $TAG_ONE and $TAG_TWO ..."
 
           git clone --depth 1 --branch "$TAG_ONE" "$CONSENSUS_NODE_REPO" "$TMP_DIR_ONE" > /dev/null 2>&1
           cd "$TMP_DIR_ONE"
-          ./gradlew build -q -x test -x spotlessCheck
+          ./gradlew hapi:generatePbjSource > /dev/null 2>&1
 
           git clone --depth 1 --branch "$TAG_TWO" "$CONSENSUS_NODE_REPO" "$TMP_DIR_TWO" > /dev/null 2>&1
           cd "$TMP_DIR_TWO"
-          ./gradlew build -q -x test -x spotlessCheck > /dev/null 2>&1
+          ./gradlew hapi:generatePbjSource > /dev/null 2>&1
 
           # Flag to track for differences
           CHANGED=false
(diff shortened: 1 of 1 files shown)


.github/workflows/releases-comparator.yml
```diff
@@ -0,0 +1,120 @@
...
+    branches:
+      - main
+      - release/**
+    tags:
+      - v*
+defaults:
+  run:
+    shell: bash
+permissions:
+  contents: write
```
IvanKavaldzhiev: I'm not sure this is necessary. Does the script work if we set the permission to read?

.github/workflows/releases-comparator.yml
```diff
@@ -0,0 +1,120 @@
...
+      - main
+      - release/**
+    paths:
+      - "build.gradle.kts"
+  push:
+    branches:
+      - main
+      - release/**
+    tags:
+      - v*
```
xin-hedera: why do we want to run it again on pushes to branches and tags?

.github/workflows/releases-comparator.yml
```diff
@@ -0,0 +1,120 @@
...
+defaults:
+  run:
+    shell: bash
+permissions:
+  contents: write
+env:
+  LC_ALL: C.UTF-8
+jobs:
+  run-diff-script:
+    name: Compare latest two releases of consensus node
```
xin-hedera: we should compare the tags, not the release branches, only when in our main build file that the `com.hedera.hashgraph:app` version changes. And we should extract the before and after versions from the diff of the build file, instead of comparing the tags extracted from the script.

xin-hedera (changes requested):

.github/workflows/releases-comparator.yml
sdimitrov9: We don't need the write since we are not pushing any changes via this github action. Changed to read.

.github/workflows/releases-comparator.yml
sdimitrov9: Removed

.github/workflows/releases-comparator.yml
```diff
@@ -0,0 +1,114 @@
...
+
+          # Temp directories to store cloned repos
+          TMP_DIR_ONE=$(mktemp -d)
+          TMP_DIR_TWO=$(mktemp -d)
+
+          # Function to clean the temp directories after script finishes
+          cleanup() {
+            rm -rf "$TMP_DIR_ONE" "$TMP_DIR_TWO"
+          }
+          trap cleanup EXIT
```
xin-hedera: this is not needed since the workspace is destroyed after the job completes and there's no subsequent steps may have conflict with files left in the temp dirs.

.github/workflows/releases-comparator.yml
```diff
@@ -0,0 +1,114 @@
...
+                sort -Vr))
+
+          TAG_ONE="${TAGS[0]}"
+          TAG_TWO="${TAGS[1]}"
+
+          echo "Comparing consensus node release versions $TAG_ONE and $TAG_TWO ..."
+
+          git clone --depth 1 --branch "$TAG_ONE" "$CONSENSUS_NODE_REPO" "$TMP_DIR_ONE" > /dev/null 2>&1
+          cd "$TMP_DIR_ONE"
+          ./gradlew build -q -x test -x spotlessCheck
```
xin-hedera: is there not a similar `generateProto` task for PBJ? that's a more lightweight task.

xin-hedera: looks like it's `pbjCompiler`, see here: https://github.com/hashgraph/pbj/blob/9c555f5a4b5106a7ee5e6ae81bbcf4084d5c5fbd/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/PbjCompilerTask.java#L15

sdimitrov9: Thanks, I think hapi:generatePbjSource will do the work. 
Change is now applied.

.github/workflows/releases-comparator.yml
sdimitrov9: Updated. Now there is conditional step that check specifically if the hedera app version is changed in the PR. If changed, the versions from the base branch and from the PR are saved and passed to the next steps.

.github/workflows/releases-comparator.yml
sdimitrov9: Removed.

.github/workflows/releases-comparator.yml
```diff
@@ -0,0 +1,129 @@
...
+          BASE_VERSION_TAG="v$BASE_VERSION"
+
+          echo "pr_version=$PR_VERSION_TAG" >> $GITHUB_OUTPUT
+          echo "base_version=$BASE_VERSION_TAG" >> $GITHUB_OUTPUT
+
+          if [ "$PR_VERSION_TAG" != "$BASE_VERSION_TAG" ]; then
+            echo "changed=true" >> $GITHUB_OUTPUT
+          else
+            echo "changed=false" >> $GITHUB_OUTPUT
+          fi
```
xin-hedera: we can move these steps into a separate job, and have this job depends on the new job, so we don't have to repeatedly have the `steps.check-hedera-app-line.outputs.changed == 'true'` condition for the subsequent steps.

.github/workflows/releases-comparator.yml
```diff
@@ -0,0 +1,129 @@
...
+
+          echo "Comparing consensus node release versions $TAG_ONE and $TAG_TWO ..."
+
+          git clone --depth 1 --branch "$TAG_ONE" "$CONSENSUS_NODE_REPO" "$TMP_DIR_ONE" > /dev/null 2>&1
+          cd "$TMP_DIR_ONE"
+          ./gradlew hapi:generatePbjSource > /dev/null 2>&1
+
+          git clone --depth 1 --branch "$TAG_TWO" "$CONSENSUS_NODE_REPO" "$TMP_DIR_TWO" > /dev/null 2>&1
+          cd "$TMP_DIR_TWO"
+          ./gradlew hapi:generatePbjSource > /dev/null 2>&1
```
xin-hedera: can make the logic into a function and call it for the two tags

.github/workflows/releases-comparator.yml
```diff
@@ -0,0 +1,129 @@
...
+              branch_two_file_path="$TMP_DIR_TWO/platform-sdk/swirlds-state-api/src/main/java/$input_file_path"
+            else
+              branch_one_file_path="$TMP_DIR_ONE/hapi/hapi/build/generated/source/pbj-proto/main/java/$input_file_path"
+              branch_two_file_path="$TMP_DIR_TWO/hapi/hapi/build/generated/source/pbj-proto/main/java/$input_file_path"
+            fi
+
+            if ! diff -q "$branch_one_file_path" "$branch_two_file_path" >/dev/null 2>&1; then
+              echo "File changed: $input_file_path"
+              echo "----------------------------------------"
+              diff -u "$branch_two_file_path" "$branch_one_file_path" || true
```
xin-hedera: nit: should just diff once for one pair of files

build.gradle.kts
```diff
@@ -48,7 +48,7 @@ dependencies {
         api("com.graphql-java-generator:graphql-java-client-runtime:2.9")
         api("com.graphql-java:graphql-java-extended-scalars:22.0")
         api("com.graphql-java:graphql-java-extended-validation:22.0")
-        api("com.hedera.hashgraph:app:0.63.5")
```
xin-hedera: checked in by accident?

sdimitrov9: Temporary changed to test the behaviour of the github action. Now should be ok.

.github/workflows/releases-comparator.yml
sdimitrov9: Done

.github/workflows/releases-comparator.yml
sdimitrov9: Done

.github/workflows/releases-comparator.yml
```diff
@@ -0,0 +1,144 @@
+# SPDX-License-Identifier: Apache-2.0
+
+name: Consensus node release branches compare
+on:
+  workflow_dispatch:
+  pull_request:
```
xin-hedera, line 5: wrong indentation and somehow still works? The `pull_request:` block should be an item under `workflow_dispatch`

.github/workflows/releases-comparator.yml
```diff
@@ -0,0 +1,144 @@
...
+jobs:
+  check-version:
+    name: Check if hedera app version changed in the PR
+    runs-on: hiero-mirror-node-linux-large
+    outputs:
+      changed: ${{ steps.check.outputs.changed }}
+      pr_version: ${{ steps.check.outputs.pr_version }}
+      base_version: ${{ steps.check.outputs.base_version }}
+    steps:
+      - name: Checkout Code
```
xin-hedera, line 32: add the following as the first step in both jobs
```
      - name: Harden Runner
        uses: step-security/harden-runner@002fdce3c6a235733a90a27c80493a3241e56863 # v2.12.1
        with:
          egress-policy: audit
```

.github/workflows/releases-comparator.yml
xin-hedera: think I misread it.

.github/workflows/releases-comparator.yml
```diff
@@ -0,0 +1,144 @@
+# SPDX-License-Identifier: Apache-2.0
+
+name: Consensus node release branches compare
+on:
+  workflow_dispatch:
```
xin-hedera:
```suggestion
```

sdimitrov9: Added to be able to start the action from the Github UI, but I guess we don't need it anymore. Removed.

.github/workflows/releases-comparator.yml
sdimitrov9: Done

xin-hedera (approved):
LGTM

steven-sheehy (approved):

PR #11434, opened 2025-06-20
Merge the two log entries into one

### LINKED ISSUE #11054: Double logging of errors does not give us more information
### Problem

In the grafana logs when we encounter an error we log it twice and this does not give us any more details. 

Once in the GenericControllerAdvice, and through the LoggingFilter

`
2025-04-28T15:29:01.183Z  WARN tomcat-handler-761989 c.h.m.w.c.GenericControllerAdvice Mirror EVM transaction error: CONTRACT_REVERT_EXECUTED, detail: , data: 0x24a01144, isModularized: true |  
-- | --
  |   | 2025-04-28T15:29:01.184Z  INFO tomcat-handler-761989 c.h.m.w.c.LoggingFilter 54.210.178.49 POST /api/v1/contracts/call in 15 ms: 400 CONTRACT_REVERT_EXECUTED - {"data":"0xb3596f0700000000000000000000000000000000000000000000000000000000003991ed","estimate":false,"to":"0x08460a5eb624cc5e112802B3d5f7f96ef59c68e1"}

`

### Solution

Either we need to enrich the logs so that they can be useful or remove one of them.

### Alternatives

_No response_

bilyana-gospodinova:
**Description**:

Fixes #11054 

**Notes for reviewer**:
Before:
```
2025-06-20T13:49:45.735Z  WARN Test worker o.h.m.w.c.GenericControllerAdvice Mirror EVM transaction error: CONTRACT_REVERT_EXECUTED, detail: Custom revert message, data: 0x08c379a000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000015437573746f6d20726576657274206d6573736167650000000000000000000000, isModularized: null, childTransactionErrors: [] 
2025-06-20T13:49:45.769Z  INFO Test worker o.h.m.w.c.LoggingFilter 127.0.0.1 POST /api/v1/contracts/call in 128 ms (mod=false): 400 CONTRACT_REVERT_EXECUTED - {"block":"latest","data":"0xa26388bb","estimate":false,"from":"0x00000000000000000000000000000000000004e2","gas":10000000,"gasPrice":78282329,"to":"0x00000000000000000000000000000000000004e4","value":23} 
```

After:
`2025-06-20T13:40:10.336Z  INFO Test worker o.h.m.w.c.LoggingFilter 127.0.0.1 POST /api/v1/contracts/call in 141 ms (mod=true): 400 CONTRACT_REVERT_EXECUTED - {"block":"latest","data":"0xa26388bb","estimate":false,"from":"0x00000000000000000000000000000000000004e2","gas":10000000,"gasPrice":78282329,"to":"0x00000000000000000000000000000000000004e4","value":23}, result: {Mirror EVM transaction error: CONTRACT_REVERT_EXECUTED, detail: Custom revert message, data: 0x08c379a000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000015437573746f6d20726576657274206d6573736167650000000000000000000000, childTransactionErrors: []} `

**Checklist**


docs/configuration.md
```diff
@@ -724,7 +724,7 @@ value, it is recommended to only populate overridden properties in the custom `a
...
 | `hiero.mirror.web3.evm.entityNumBuffer`                      | 1000                                               | A buffer added to the max known entity number to prevent contract call simulations from generating IDs that may conflict with entities persisted by the importer module.                         |
-| `hiero.mirror.web3.maxPayloadLogSize`                        | 300                                                | The maximum number of bytes to use to log the request payload.                                                                                                                                   |
+| `hiero.mirror.web3.maxPayloadLogSize`                        | 600                                                | The maximum number of bytes to use to log the request payload.                                                                                                                                   |
```
steven-sheehy: We should not double our log size as this will double our log cloud costs. Since this config is used for both input and output so when there's no error input will now always be up to 600 bytes. We should add a `maxResponseLogSize` and set it to 100 or so. I think we can output a full error message and only limit the `MirrorEvmTransactionException.data` to 100 chars.

steven-sheehy: Guess the description is wrong. We should swap `bytes` for `characters`. Can you please adjust that since you're modifying this area?

web3/src/main/java/org/hiero/mirror/web3/config/LoggingFilter.java
```diff
@@ -78,13 +80,16 @@ private void logRequest(HttpServletRequest request, HttpServletResponse response
...
+        var contentBuilder = new StringBuilder();
         int maxPayloadLogSize = web3Properties.getMaxPayloadLogSize();
         var wrapper = WebUtils.getNativeRequest(request, ContentCachingRequestWrapper.class);
 
         if (wrapper != null) {
-            content = StringUtils.deleteWhitespace(wrapper.getContentAsString());
+            contentBuilder.append(StringUtils.deleteWhitespace(wrapper.getContentAsString()));
         }
+        var exceptionMessage = getExceptionMessage(request);
+        contentBuilder.append(", result: {").append(exceptionMessage).append("}");
```
steven-sheehy: Believe this will break replay functionality. We should move it to the message right before ` - ` so request is last. In general, think we can revert changes to this class with my other recommendation.

web3/src/main/java/org/hiero/mirror/web3/controller/GenericControllerAdvice.java
```diff
@@ -73,14 +74,11 @@ private ResponseEntity<?> badRequest(final Exception e, final WebRequest request
...
-                e.getMessage(),
-                e.getDetail(),
-                e.getData(),
-                e.getIsCallModularized(),
-                e.getChildTransactionErrors());
+        final var exceptionMessage = String.format(
+                "Mirror EVM transaction error: %s, detail: %s, data: %s, childTransactionErrors: %s",
+                e.getMessage(), e.getDetail(), e.getData(), e.getChildTransactionErrors());
         request.setAttribute(WebUtils.ERROR_EXCEPTION_ATTRIBUTE, e, SCOPE_REQUEST);
+        request.setAttribute(EXCEPTION_MESSAGE, exceptionMessage, SCOPE_REQUEST);
```
steven-sheehy: I think we should leverage the existing `ERROR_EXCEPTION_ATTRIBUTE` to attach the error so that it gets output in the message section in LoggingFilter. Then this requires no change to LoggingFilter and doesn't break replay. To do this, we should customize the `MirrorEvmTransactionException.getMessage()` to have the necessary details of `exceptionMessage`.

web3/src/main/java/org/hiero/mirror/web3/controller/GenericControllerAdvice.java
```diff
@@ -73,14 +74,11 @@ private ResponseEntity<?> badRequest(final Exception e, final WebRequest request
...
-        log.warn(
-                "Mirror EVM transaction error: {}, detail: {}, data: {}, isModularized: {}, childTransactionErrors: {}",
-                e.getMessage(),
-                e.getDetail(),
-                e.getData(),
-                e.getIsCallModularized(),
-                e.getChildTransactionErrors());
+        final var exceptionMessage = String.format(
+                "Mirror EVM transaction error: %s, detail: %s, data: %s, childTransactionErrors: %s",
+                e.getMessage(), e.getDetail(), e.getData(), e.getChildTransactionErrors());
```
steven-sheehy: We should put `data` at the end since it can be large. We should also cap the data to say 100 characters as per config. We shouldn't limit message, detail, child errors size.

steven-sheehy: I think frequently some of these are empty. We should append each item only if non-empty to reduce the size. e.g. we shouldn't have `childTransactionErrors: []`.

bilyana-gospodinova: Good idea. I added the non-empty check.

docs/configuration.md
bilyana-gospodinova: Sure. I changed the description.

web3/src/main/java/org/hiero/mirror/web3/config/LoggingFilter.java
bilyana-gospodinova: I removed this line and now the result is added in the `MirrorEvmTransactionException`.

web3/src/main/java/org/hiero/mirror/web3/controller/GenericControllerAdvice.java
bilyana-gospodinova: I added a new method - `getFullMessage()` in `MirrorEvmTransactionException` in order to address the comment and I removed the added request attribute. I preferred not to modify the existing `getMessage()` method as it is used in a lot of places and this might break some existing logs.

web3/src/main/java/org/hiero/mirror/web3/exception/MirrorEvmTransactionException.java
```diff
@@ -78,6 +78,19 @@ public Bytes messageBytes() {
         return Bytes.of(message.getBytes(StandardCharsets.UTF_8));
     }
 
+    public String getFullMessage() {
```
kselveliev, line 81: I think we can add some unit tests for this new method.

steven-sheehy (dismissed):

web3/src/main/java/org/hiero/mirror/web3/exception/MirrorEvmTransactionException.java
bilyana-gospodinova: Sure, added some test cases.

steven-sheehy (approved):

PR #11457, opened 2025-06-23
Set a default postgresql image tag in charts 

xin-hedera:
**Description**:

This PR sets a default postgresql iamge tag in `hedera-mirror` charts:

- Set postgresql image tag to `17.5.0-debian-12-r10`

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/charts/marketplace/gcp/release.sh
+++ b/charts/marketplace/gcp/release.sh
@@ -23,7 +23,7 @@ fi
 
 annotation="com.googleapis.cloudmarketplace.product.service.name=services/hedera-mirror-node-mirror-node-public.cloudpartnerservices.goog"
 bats_tag="1.11.1"
-postgresql_tag="16.6.0-debian-12-r3"
+postgresql_tag="17.5.0-debian-12-r10"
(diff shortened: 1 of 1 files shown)


charts/hedera-mirror/values.yaml
```diff
@@ -194,6 +194,7 @@ postgresql:
     extraEnvVarsSecret: mirror-passwords
     image:
       debug: true
+      tag: 17.5.0-debian-12-r10
```
steven-sheehy, line 197: Need to also update in marketplace release.sh

xin-hedera: that'll break existing marketplace deployments since it's an upgrade from pg16 to pg17 and it doesn't seem the users can override it. Should we just keep it?

steven-sheehy: Marketplace isn't really upgradable. Once they deploy they're responsible to manage upgrades themselves. So we've always kept the images in sync.

steven-sheehy (approved):

PR #11464, opened 2025-06-24
Update transfer synthetic logs to lookup evm address

### LINKED ISSUE #10856: Mirror Node Returns Long Zero Address in Logs for HTS Token Mint
When minting an HTS token, the relay fetches transaction logs using the contracts/results/logs endpoint from the Mirror Node. However, in the topics array of the emitted Transfer event, the address of the account appears as a long zero address instead of its EVM address (when available).

This differs from tokens created via smart contracts, where the correct EVM address is returned. The inconsistency causes issues in third-party tools like BlockScout that rely on standard EVM log formats.

Here is an example transaction minting an HTS token, respectively doing a transfer and a transfer event is emitted. This transfer event appears in the topics array 
https://hashscan.io/testnet/transaction/1742987700.661533638

and the logs in response of the relay
`"logs": [
            {
                "address": "0x0000000000000000000000000000000000582649",
                "blockHash": "0xf7ac5123d60e821e93dee28ceb9e37590b5571a94cc15e95c6179538da0c8ff2",
                "blockNumber": "0x10a9cc4",
                "data": "0x00000000000000000000000000000000000000000000000000000000000003e8",
                "logIndex": "0x0",
                "removed": false,
                "topics": [
                    "0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef",
                    "0x0000000000000000000000000000000000000000000000000000000000000000",
                    "0x0000000000000000000000000000000000000000000000000000000000219b4a"
                ],
                "transactionHash": "0xb8a3a7107b597e2ac85f2b1ffb6208a4ca3c7397ab355e5ea48c810f6c81c5ab",
                "transactionIndex": "0xa"
            }
        ],`

Expected:
Return the EVM address of the account in the topics (when available), instead of the long zero-prefixed format.

This will ensure compatibility with EVM-based tools and make log handling more consistent.

jnels124:
**Description**:
* add new entity listener and record stream file listener
* lookup evm address for sender/receiver on transfer synthetic logs
* add migration to populate evm address for sender/receiver on existing transfer synthetic logs

**Related issue(s)**:

Fixes #10856

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/common/src/main/java/org/hiero/mirror/common/util/DomainUtils.java
+++ b/common/src/main/java/org/hiero/mirror/common/util/DomainUtils.java
@@ -285,6 +285,39 @@ public class DomainUtils {
         return null;
     }
 
+    public static EntityId fromTrimmedEvmAddress(byte[] evmAddress) {
+        if (evmAddress == null) {
+            return null;
+        }
+
+        var padding = new byte[EVM_ADDRESS_LENGTH - evmAddress.length];
+
+        return fromEvmAddress(Bytes.concat(padding, evmAddress));
+    }
+
+    public static byte[] trim(byte[] data) {
+        if (ArrayUtils.isEmpty(data)) {
+            return data;
+        }
+
+        int i = 0;
+        for (; i < data.length; i++) {
+            if (data[i] != 0) {
+                break;
+            }
+        }
+
+        if (i == 0) {
+            return data;
+        }
+
+        if (i == data.length) {
+            return ArrayUtils.EMPTY_BYTE_ARRAY;
+        }
+
+        return ArrayUtils.subarray(data, i, data.length);
+    }
+
     public static byte[] toEvmAddress(ContractID contractId) {
         if (contractId == null || contractId == ContractID.getDefaultInstance()) {
             throw new InvalidEntityException("Invalid ContractID");
--- a/importer/src/main/java/org/hiero/mirror/importer/config/CacheProperties.java
+++ b/importer/src/main/java/org/hiero/mirror/importer/config/CacheProperties.java
@@ -20,6 +20,8 @@ public class CacheProperties {
 
     private boolean enabled = true;
 
+    private String evmAddress = "maximumSize=500000,recordStats";
+
     @NotBlank
     private String timePartition = "maximumSize=50,expireAfterWrite=1d,recordStats";
 
--- a/importer/src/main/java/org/hiero/mirror/importer/domain/EntityIdServiceImpl.java
+++ b/importer/src/main/java/org/hiero/mirror/importer/domain/EntityIdServiceImpl.java
@@ -14,18 +14,12 @@ import com.hederahashgraph.api.proto.java.AccountID;
 import com.hederahashgraph.api.proto.java.ContractID;
 import jakarta.annotation.Nonnull;
 import jakarta.inject.Named;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
-import java.util.Set;
 import java.util.concurrent.Callable;
 import java.util.function.Function;
-import java.util.stream.Collectors;
 import lombok.CustomLog;
 import org.apache.commons.codec.binary.Hex;
-import org.hiero.mirror.common.domain.entity.AbstractEntity;
 import org.hiero.mirror.common.domain.entity.Entity;
 import org.hiero.mirror.common.domain.entity.EntityId;
 import org.hiero.mirror.common.domain.entity.EntityType;
@@ -43,16 +37,12 @@ public class EntityIdServiceImpl implements EntityIdService {
     private static final Optional<EntityId> EMPTY = Optional.of(EntityId.EMPTY);
 
     private final Cache aliasCache;
-    private final Cache evmCache;
 
     private final EntityRepository entityRepository;
 
     public EntityIdServiceImpl(
-            @Qualifier(CACHE_ALIAS) CacheManager aliasCachManager,
-            @Qualifier(CACHE_EVM_ADDRESS) CacheManager evmAddressCacheManager,
-            EntityRepository entityRepository) {
+            @Qualifier(CACHE_ALIAS) CacheManager aliasCachManager, EntityRepository entityRepository) {
         this.aliasCache = aliasCachManager.getCache(CACHE_NAME);
-        this.evmCache = evmAddressCacheManager.getCache(CACHE_NAME);
         this.entityRepository = entityRepository;
     }
 
@@ -111,35 +101,6 @@ public class EntityIdServiceImpl implements EntityIdService {
     @Override
     public Optional<EntityId> lookup(ContractID... contractIds) {
         return doLookups(contractIds, this::lookup);
-    }
-
-    public Map<Long, byte[]> lookupEvmAddresses(Set<Long> entityIds) {
-        Map<Long, byte[]> result = new HashMap<>();
-        Set<Long> missing = new HashSet<>();
-
-        for (Long id : entityIds) {
-            byte[] cached = evmCache.get(id, byte[].class);
-            if (cached != null) {
-                result.put(id, cached);
-            } else {
-                missing.add(id);
-            }
-        }
-
-        if (!missing.isEmpty()) {
-            Map<Long, byte[]> dbResults = entityRepository.findEvmAddressesByIds(missing).stream()
-                    .collect(Collectors.toMap(AbstractEntity::getNum, entity -> {
-                        byte[] evm = entity.getEvmAddress();
-                        return evm != null ? evm : DomainUtils.toEvmAddress(entity.getNum());
-                    }));
-
-            dbResults.forEach((id, evmAddress) -> {
-                evmCache.put(id, evmAddress);
-                result.put(id, evmAddress);
-            });
-        }
-
-        return result;
     }
 
     private @Nonnull Optional<EntityId> cacheLookup(ByteString key, Callable<Optional<EntityId>> loader) {
--- a/importer/src/main/java/org/hiero/mirror/importer/domain/EntityIdServiceImpl.java
+++ b/importer/src/main/java/org/hiero/mirror/importer/domain/EntityIdServiceImpl.java
@@ -5,8 +5,7 @@ package org.hiero.mirror.importer.domain;
 import static org.hiero.mirror.common.util.DomainUtils.EVM_ADDRESS_LENGTH;
 import static org.hiero.mirror.common.util.DomainUtils.fromBytes;
 import static org.hiero.mirror.common.util.DomainUtils.toBytes;
-import static org.hiero.mirror.importer.config.CacheConfiguration.CACHE_ALIAS;
-import static org.hiero.mirror.importer.config.CacheConfiguration.CACHE_NAME;
+import static org.hiero.mirror.importer.config.CacheConfiguration.*;
 import static org.hiero.mirror.importer.util.Utility.aliasToEvmAddress;
 
 import com.google.protobuf.ByteString;
@@ -15,12 +14,18 @@ import com.hederahashgraph.api.proto.java.AccountID;
 import com.hederahashgraph.api.proto.java.ContractID;
 import jakarta.annotation.Nonnull;
 import jakarta.inject.Named;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.Set;
 import java.util.concurrent.Callable;
 import java.util.function.Function;
+import java.util.stream.Collectors;
 import lombok.CustomLog;
 import org.apache.commons.codec.binary.Hex;
+import org.hiero.mirror.common.domain.entity.AbstractEntity;
 import org.hiero.mirror.common.domain.entity.Entity;
 import org.hiero.mirror.common.domain.entity.EntityId;
 import org.hiero.mirror.common.domain.entity.EntityType;
@@ -37,11 +42,17 @@ public class EntityIdServiceImpl implements EntityIdService {
 
     private static final Optional<EntityId> EMPTY = Optional.of(EntityId.EMPTY);
 
-    private final Cache cache;
+    private final Cache aliasCache;
+    private final Cache evmCache;
+
     private final EntityRepository entityRepository;
 
-    public EntityIdServiceImpl(@Qualifier(CACHE_ALIAS) CacheManager cacheManager, EntityRepository entityRepository) {
-        this.cache = cacheManager.getCache(CACHE_NAME);
+    public EntityIdServiceImpl(
+            @Qualifier(CACHE_ALIAS) CacheManager aliasCachManager,
+            @Qualifier(CACHE_EVM_ADDRESS) CacheManager evmAddressCacheManager,
+            EntityRepository entityRepository) {
+        this.aliasCache = aliasCachManager.getCache(CACHE_NAME);
+        this.evmCache = evmAddressCacheManager.getCache(CACHE_NAME);
         this.entityRepository = entityRepository;
     }
 
@@ -102,9 +113,38 @@ public class EntityIdServiceImpl implements EntityIdService {
         return doLookups(contractIds, this::lookup);
     }
 
+    public Map<Long, byte[]> lookupEvmAddresses(Set<Long> entityIds) {
+        Map<Long, byte[]> result = new HashMap<>();
+        Set<Long> missing = new HashSet<>();
+
+        for (Long id : entityIds) {
+            byte[] cached = evmCache.get(id, byte[].class);
+            if (cached != null) {
+                result.put(id, cached);
+            } else {
+                missing.add(id);
+            }
+        }
+
+        if (!missing.isEmpty()) {
+            Map<Long, byte[]> dbResults = entityRepository.findEvmAddressesByIds(missing).stream()
+                    .collect(Collectors.toMap(AbstractEntity::getNum, entity -> {
+                        byte[] evm = entity.getEvmAddress();
+                        return evm != null ? evm : DomainUtils.toEvmAddress(entity.getNum());
+                    }));
+
+            dbResults.forEach((id, evmAddress) -> {
+                evmCache.put(id, evmAddress);
+                result.put(id, evmAddress);
+            });
+        }
+
+        return result;
+    }
+
     private @Nonnull Optional<EntityId> cacheLookup(ByteString key, Callable<Optional<EntityId>> loader) {
         try {
-            return Objects.requireNonNullElse(cache.get(key, loader), Optional.empty());
+            return Objects.requireNonNullElse(aliasCache.get(key, loader), Optional.empty());
         } catch (Cache.ValueRetrievalException e) {
             Utility.handleRecoverableError("Error looking up alias or EVM address {} from cache", key, e);
             return Optional.empty();
@@ -139,14 +179,14 @@ public class EntityIdServiceImpl implements EntityIdService {
 
         switch (type) {
             case ACCOUNT -> {
-                cache.put(alias, entityId);
+                aliasCache.put(alias, entityId);
 
                 // Accounts can have an alias and an EVM address so warm the cache with both
                 if (entity.getAlias() != null && entity.getEvmAddress() != null) {
-                    cache.put(fromBytes(entity.getEvmAddress()), entityId);
+                    aliasCache.put(fromBytes(entity.getEvmAddress()), entityId);
                 }
             }
-            case CONTRACT -> cache.put(alias, entityId);
+            case CONTRACT -> aliasCache.put(alias, entityId);
             default -> Utility.handleRecoverableError("Invalid Entity: {} entity can't have alias", type);
         }
     }
--- a/importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListener.java
+++ b/importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListener.java
@@ -2,39 +2,57 @@
 
 package org.hiero.mirror.importer.parser.contractlog;
 
+import static org.hiero.mirror.common.util.DomainUtils.*;
+import static org.hiero.mirror.importer.config.CacheConfiguration.CACHE_EVM_ADDRESS;
+import static org.hiero.mirror.importer.config.CacheConfiguration.CACHE_NAME;
 import static org.hiero.mirror.importer.parser.contractlog.AbstractSyntheticContractLog.TRANSFER_SIGNATURE;
 
+import io.micrometer.core.annotation.Timed;
 import jakarta.inject.Named;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import lombok.RequiredArgsConstructor;
+import java.util.stream.Collectors;
+import lombok.CustomLog;
 import org.hiero.mirror.common.domain.contract.ContractLog;
 import org.hiero.mirror.common.domain.entity.EntityId;
 import org.hiero.mirror.common.domain.transaction.RecordFile;
-import org.hiero.mirror.common.util.DomainUtils;
-import org.hiero.mirror.importer.domain.EntityIdService;
 import org.hiero.mirror.importer.parser.record.RecordStreamFileListener;
 import org.hiero.mirror.importer.parser.record.entity.EntityListener;
-import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.hiero.mirror.importer.parser.record.entity.EntityProperties;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.cache.Cache;
+import org.springframework.cache.CacheManager;
 import org.springframework.core.annotation.Order;
+import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
+import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
 
 @Named
 @Order(2)
-@RequiredArgsConstructor
-@ConditionalOnProperty(
-        name = "hiero.mirror.importer.parser.record.entity.persist.synthetic-contract-log-evm-address-lookup",
-        havingValue = "true")
+@CustomLog
 public class SyntheticLogListener implements EntityListener, RecordStreamFileListener {
-    private final EntityIdService entityIdService;
+    private static final String LOOKUP_QUERY =
+            "select evm_address, num from entity where id in (:ids) and length(evm_address) > 0";
+    private final NamedParameterJdbcTemplate jdbcTemplate;
+    private final Cache evmCache;
+    private final EntityProperties entityProperties;
     private final List<SyntheticLogUpdater> syntheticLogParticipants = new ArrayList<>();
-    private final Set<Long> evmAddressLookupIds = new HashSet<>();
+    private final Set<EntityId> evmAddressLookupIds = new HashSet<>();
 
-    // TODO ADD Metrics
+    public SyntheticLogListener(
+            NamedParameterJdbcTemplate jdbcTemplate,
+            EntityProperties entityProperties,
+            @Qualifier(CACHE_EVM_ADDRESS) CacheManager cacheManager) {
+        this.jdbcTemplate = jdbcTemplate;
+        this.evmCache = cacheManager.getCache(CACHE_NAME);
+        this.entityProperties = entityProperties;
+    }
 
     @Override
+    @Timed
     public void onEnd(RecordFile recordFile) {
         updateSyntheticContractLogs();
         syntheticLogParticipants.clear();
@@ -43,9 +61,10 @@ public class SyntheticLogListener implements EntityListener, RecordStreamFileLis
 
     @Override
     public void onContractLog(ContractLog contractLog) {
-        if (contractLog.getTopic0() == TRANSFER_SIGNATURE) {
-            var senderId = DomainUtils.fromEvmAddress(contractLog.getTopic1());
-            var receiverId = DomainUtils.fromEvmAddress(contractLog.getTopic2());
+        if (entityProperties.getPersist().isSyntheticContractLogEvmAddressLookup()
+                && contractLog.getTopic0() == TRANSFER_SIGNATURE) {
+            var senderId = fromTrimmedEvmAddress(contractLog.getTopic1());
+            var receiverId = fromTrimmedEvmAddress(contractLog.getTopic2());
             if (!(senderId == null && receiverId == null)) {
                 var updater = new SyntheticLogUpdater(senderId, receiverId, contractLog);
                 syntheticLogParticipants.add(updater);
@@ -55,33 +74,83 @@ public class SyntheticLogListener implements EntityListener, RecordStreamFileLis
     }
 
     private void updateSyntheticContractLogs() {
-        var entityMap = entityIdService.lookupEvmAddresses(evmAddressLookupIds);
+        var entityMap = lookupEvmAddresses();
         syntheticLogParticipants.forEach(updater -> updater.updateContractLog(entityMap));
     }
 
-    private record SyntheticLogUpdater(EntityId sender, EntityId receiver, ContractLog contractLog) {
+    private Map<Long, byte[]> lookupEvmAddresses() {
+        Map<Long, byte[]> result = new HashMap<>(evmAddressLookupIds.size());
+        Set<EntityId> missing = new HashSet<>();
 
-        public List<Long> getSearchIds() {
-            var ids = new ArrayList<Long>();
-
-            if (receiver != null) {
-                ids.add(receiver.getId());
+        try {
+            // Try cache for all entity IDs
+            for (var entityId : evmAddressLookupIds) {
+                var num = entityId.getNum();
+                var cached = evmCache.get(num, byte[].class);
+                if (cached != null) {
+                    result.put(num, cached);
+                } else {
+                    missing.add(entityId);
+                }
             }
 
-            if (sender != null) {
-                ids.add(sender.getId());
+            // Query all missing Keys
+            var dbResults = findEvmAddressesByIds(missing).stream()
+                    .collect(Collectors.toMap(EvmAddressMapping::num, EvmAddressMapping::evmAddress));
+
+            // Fill in results from DB (or default if still missing)
+            for (var entityId : missing) {
+                long num = entityId.getNum();
+                byte[] evmAddress = dbResults.getOrDefault(num, trim(toEvmAddress(entityId)));
+                evmCache.put(num, evmAddress);
+                result.put(num, evmAddress);
+            }
+        } catch (Exception e) {
+            log.error("Error looking up EVM addresses for entity IDs: {}", evmAddressLookupIds, e);
+        }
+
+        return result;
+    }
+
+    private List<EvmAddressMapping> findEvmAddressesByIds(Set<EntityId> ids) {
+        if (ids.isEmpty()) {
+            return List.of();
+        }
+        var params = new MapSqlParameterSource()
+                .addValue("ids", ids.stream().map(EntityId::getId).toList());
+
+        return jdbcTemplate.query(LOOKUP_QUERY, params, (rs, rowNum) -> {
+            byte[] evmAddress = rs.getBytes("evm_address");
+            long num = rs.getLong("num");
+            return new EvmAddressMapping(evmAddress, num);
+        });
+    }
+
+    record EvmAddressMapping(byte[] evmAddress, long num) {}
+
+    private record SyntheticLogUpdater(EntityId sender, EntityId receiver, ContractLog contractLog) {
+
+        public List<EntityId> getSearchIds() {
+            var ids = new ArrayList<EntityId>();
+
+            if (receiver != null && receiver != EntityId.EMPTY) {
+                ids.add(receiver);
+            }
+
+            if (sender != null && sender != EntityId.EMPTY) {
+                ids.add(sender);
             }
 
             return ids;
         }
 
         public void updateContractLog(Map<Long, byte[]> entityEvmAddresses) {
-            if (sender != null && entityEvmAddresses.containsKey(sender.getId())) {
-                contractLog.setTopic1(entityEvmAddresses.get(sender.getId()));
+            if (!EntityId.isEmpty(sender) && entityEvmAddresses.containsKey(sender.getNum())) {
+                contractLog.setTopic1(entityEvmAddresses.get(sender.getNum()));
             }
 
-            if (receiver != null && entityEvmAddresses.containsKey(receiver.getId())) {
-                contractLog.setTopic2(entityEvmAddresses.get(receiver.getId()));
+            if (!EntityId.isEmpty(receiver) && entityEvmAddresses.containsKey(receiver.getNum())) {
+                contractLog.setTopic2(entityEvmAddresses.get(receiver.getNum()));
             }
         }
     }
--- a/importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListener.java
+++ b/importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListener.java
@@ -1,0 +1,88 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package org.hiero.mirror.importer.parser.contractlog;
+
+import static org.hiero.mirror.importer.parser.contractlog.AbstractSyntheticContractLog.TRANSFER_SIGNATURE;
+
+import jakarta.inject.Named;
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import lombok.RequiredArgsConstructor;
+import org.hiero.mirror.common.domain.contract.ContractLog;
+import org.hiero.mirror.common.domain.entity.EntityId;
+import org.hiero.mirror.common.domain.transaction.RecordFile;
+import org.hiero.mirror.common.util.DomainUtils;
+import org.hiero.mirror.importer.domain.EntityIdService;
+import org.hiero.mirror.importer.parser.record.RecordStreamFileListener;
+import org.hiero.mirror.importer.parser.record.entity.EntityListener;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.core.annotation.Order;
+
+@Named
+@Order(2)
+@RequiredArgsConstructor
+@ConditionalOnProperty(
+        name = "hiero.mirror.importer.parser.record.entity.persist.synthetic-contract-log-evm-address-lookup",
+        havingValue = "true")
+public class SyntheticLogListener implements EntityListener, RecordStreamFileListener {
+    private final EntityIdService entityIdService;
+    private final List<SyntheticLogUpdater> syntheticLogParticipants = new ArrayList<>();
+    private final Set<Long> evmAddressLookupIds = new HashSet<>();
+
+    // TODO ADD Metrics
+
+    @Override
+    public void onEnd(RecordFile recordFile) {
+        updateSyntheticContractLogs();
+        syntheticLogParticipants.clear();
+        evmAddressLookupIds.clear();
+    }
+
+    @Override
+    public void onContractLog(ContractLog contractLog) {
+        if (contractLog.getTopic0() == TRANSFER_SIGNATURE) {
+            var senderId = DomainUtils.fromEvmAddress(contractLog.getTopic1());
+            var receiverId = DomainUtils.fromEvmAddress(contractLog.getTopic2());
+            if (!(senderId == null && receiverId == null)) {
+                var updater = new SyntheticLogUpdater(senderId, receiverId, contractLog);
+                syntheticLogParticipants.add(updater);
+                evmAddressLookupIds.addAll(updater.getSearchIds());
+            }
+        }
+    }
+
+    private void updateSyntheticContractLogs() {
+        var entityMap = entityIdService.lookupEvmAddresses(evmAddressLookupIds);
+        syntheticLogParticipants.forEach(updater -> updater.updateContractLog(entityMap));
+    }
+
+    private record SyntheticLogUpdater(EntityId sender, EntityId receiver, ContractLog contractLog) {
+
+        public List<Long> getSearchIds() {
+            var ids = new ArrayList<Long>();
+
+            if (receiver != null) {
+                ids.add(receiver.getId());
+            }
+
+            if (sender != null) {
+                ids.add(sender.getId());
+            }
+
+            return ids;
+        }
+
+        public void updateContractLog(Map<Long, byte[]> entityEvmAddresses) {
+            if (sender != null && entityEvmAddresses.containsKey(sender.getId())) {
+                contractLog.setTopic1(entityEvmAddresses.get(sender.getId()));
+            }
+
+            if (receiver != null && entityEvmAddresses.containsKey(receiver.getId())) {
+                contractLog.setTopic2(entityEvmAddresses.get(receiver.getId()));
+            }
+        }
+    }
+}
--- a/importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/topic/TopicMessageLookupEntityListener.java
+++ b/importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/topic/TopicMessageLookupEntityListener.java
@@ -18,7 +18,7 @@ import org.hiero.mirror.importer.parser.record.entity.ParserContext;
 import org.springframework.core.annotation.Order;
 
 @Named
-@Order(3)
+@Order(4)
 @ConditionOnEntityRecordParser
 @RequiredArgsConstructor
 public class TopicMessageLookupEntityListener implements EntityListener {
--- a/importer/src/main/java/org/hiero/mirror/importer/repository/EntityRepository.java
+++ b/importer/src/main/java/org/hiero/mirror/importer/repository/EntityRepository.java
@@ -2,7 +2,6 @@
 
 package org.hiero.mirror.importer.repository;
 
-import java.util.List;
 import java.util.Optional;
 import org.hiero.mirror.common.domain.entity.Entity;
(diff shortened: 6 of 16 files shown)


importer/src/test/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListenerTest.java
```diff
@@ -0,0 +1,288 @@
...
+    }
+
+    @Test
+    void alreadyHasEvmAddress() {
+        var contractLog = domainBuilder
+                .contractLog()
+                .customize(cl -> cl.topic0(TRANSFER_SIGNATURE).topic1(EVM_1).topic2(EVM_2))
+                .get();
+
+        listener.onContractLog(contractLog);
```
jnels124, line 207: will add onEnd call.

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListener.java
```diff
@@ -0,0 +1,157 @@
...
+    public void onEnd(RecordFile recordFile) {
+        updateSyntheticContractLogs();
+        syntheticLogParticipants.clear();
+        evmAddressLookupIds.clear();
+    }
+
+    @Override
+    public void onContractLog(ContractLog contractLog) {
+        if (entityProperties.getPersist().isSyntheticContractLogEvmAddressLookup()
+                && contractLog.getTopic0() == TRANSFER_SIGNATURE) {
```
jnels124: Will change to use Arrays.equals

common/src/test/java/org/hiero/mirror/common/util/DomainUtilsTest.java
```diff
@@ -436,4 +435,51 @@ void isMirror(long shard, long realm, String hexAddress, boolean result) throws
...
+    })
+    void testFromTrimmedEvmAddress(String inputHex, Long expectedNum) {
+        var input = inputHex == null ? null : Hex.decodeHex(inputHex);
+
+        var result = DomainUtils.fromTrimmedEvmAddress(input);
+
+        if (expectedNum == null) {
+            assertNull(result);
+        } else {
+            assertEquals(expectedNum, result.getNum());
```
steven-sheehy: Use assertThat here and above

importer/src/main/java/org/hiero/mirror/importer/domain/EntityIdServiceImpl.java
```diff
@@ -5,8 +5,7 @@
 import static org.hiero.mirror.common.util.DomainUtils.EVM_ADDRESS_LENGTH;
 import static org.hiero.mirror.common.util.DomainUtils.fromBytes;
 import static org.hiero.mirror.common.util.DomainUtils.toBytes;
-import static org.hiero.mirror.importer.config.CacheConfiguration.CACHE_ALIAS;
-import static org.hiero.mirror.importer.config.CacheConfiguration.CACHE_NAME;
+import static org.hiero.mirror.importer.config.CacheConfiguration.*;
```
steven-sheehy: Should not have wildcard imports. In fact, think all changes to this class could be reverted.

common/src/test/java/org/hiero/mirror/common/util/DomainUtilsTest.java
```diff
@@ -4,10 +4,7 @@
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.hiero.mirror.common.util.CommonUtils.nextBytes;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNull;
-import static org.junit.Assert.assertThrows;
-import static org.junit.jupiter.api.Assertions.assertAll;
+import static org.junit.jupiter.api.Assertions.*;
```
steven-sheehy: Avoid wildcard imports. May double check IDE settings.

importer/src/main/java/org/hiero/mirror/importer/config/CacheProperties.java
```diff
@@ -20,6 +20,8 @@ public class CacheProperties {
 
     private boolean enabled = true;
 
+    private String evmAddress = "maximumSize=500000,recordStats";
```
steven-sheehy, line 23: Need to document.

common/src/main/java/org/hiero/mirror/common/util/DomainUtils.java
```diff
@@ -285,6 +285,39 @@ public static EntityId fromEvmAddress(byte[] evmAddress) {
         return null;
     }
 
+    public static EntityId fromTrimmedEvmAddress(byte[] evmAddress) {
+        if (evmAddress == null) {
+            return null;
+        }
+
+        var padding = new byte[EVM_ADDRESS_LENGTH - evmAddress.length];
```
steven-sheehy, line 296: We should handle evmAddress.length > EVM_ADDRESS_LENGTH

common/src/main/java/org/hiero/mirror/common/util/DomainUtils.java
```diff
@@ -285,6 +285,39 @@ public static EntityId fromEvmAddress(byte[] evmAddress) {
...
+
+        return fromEvmAddress(Bytes.concat(padding, evmAddress));
```
steven-sheehy, line 298: We can optimize evmAddress.length == EVM_ADDRESS_LENGTH by returning evmAddress without padding.

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListener.java
```diff
@@ -0,0 +1,158 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package org.hiero.mirror.importer.parser.contractlog;
+
+import static org.hiero.mirror.common.util.DomainUtils.*;
```
steven-sheehy: wildcard

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListener.java
```diff
@@ -0,0 +1,158 @@
...
+import org.springframework.cache.Cache;
+import org.springframework.cache.CacheManager;
+import org.springframework.core.annotation.Order;
+import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
+import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
+
+@Named
+@Order(2)
+@CustomLog
+public class SyntheticLogListener implements EntityListener, RecordStreamFileListener {
```
steven-sheehy:
```suggestion
final class SyntheticLogListener implements EntityListener, RecordStreamFileListener {
```

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListener.java
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    private List<EvmAddressMapping> findEvmAddressesByIds(Set<EntityId> ids) {
+        if (ids.isEmpty()) {
+            return List.of();
+        }
+        var params = new MapSqlParameterSource()
+                .addValue("ids", ids.stream().map(EntityId::getId).toList());
+
+        return jdbcTemplate.query(LOOKUP_QUERY, params, (rs, rowNum) -> {
```
steven-sheehy: We should use a repository method. JdbcTemplate should generally be avoided outside migrations.

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListener.java
steven-sheehy: We can put `@ConditionalOnProperty` and put the new property in case the listener itself slows down processing. Then we can remove property elsewhere.

importer/src/main/resources/db/migration/v1/V1.110.0__fix_transfer_synthetic_contract_logs.sql
```diff
@@ -0,0 +1,48 @@
...
+),
+cl_ext as (
+    select *,
+           encode(topic1, 'hex') as topic1_hex,
+           encode(topic2, 'hex') as topic2_hex
+    from contract_log
+    where topic0 = '\xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef'
+      and (
+          octet_length(topic1) < 20 or
+          octet_length(topic2) < 20 or
```
steven-sheehy: An EVM address can be less than 20 bytes (due to zero prefix stripping) but not a long-zero EVM address. Think we should explicitly check that the `octet_length(topic1) <= 8` for this scenario.

importer/src/test/java/org/hiero/mirror/importer/migration/UpdateSyntheticContractLogsMigrationTest.java
```diff
@@ -0,0 +1,114 @@
...
+import org.hiero.mirror.importer.config.Owner;
+import org.hiero.mirror.importer.repository.ContractLogRepository;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.jdbc.core.JdbcTemplate;
+import org.testcontainers.shaded.com.google.common.primitives.Longs;
+
+@Tag("migration")
+public class UpdateSyntheticContractLogsMigrationTest extends ImporterIntegrationTest {
```
steven-sheehy:
```suggestion
final class UpdateSyntheticContractLogsMigrationTest extends ImporterIntegrationTest {
```

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListener.java
```diff
@@ -0,0 +1,158 @@
...
+    @Override
+    public void onContractLog(ContractLog contractLog) {
+        if (entityProperties.getPersist().isSyntheticContractLogEvmAddressLookup()
+                && Arrays.equals(contractLog.getTopic0(), TRANSFER_SIGNATURE)) {
+            var senderId = fromTrimmedEvmAddress(contractLog.getTopic1());
+            var receiverId = fromTrimmedEvmAddress(contractLog.getTopic2());
+            if (!(senderId == null && receiverId == null)) {
+                var updater = new SyntheticLogUpdater(senderId, receiverId, contractLog);
+                syntheticLogParticipants.add(updater);
+                evmAddressLookupIds.addAll(updater.getSearchIds());
```
steven-sheehy: I think we can omit this logic in `onContractLog` and instead leverage `ParserContext.get(ContractLog.class)` in a `StreamFileListener.onEnd()` and just accumulate at the end. I really don't want to have stateful classes anywhere outside ParserContext. That way we also don't need `SyntheticLogUpdater`.

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListener.java
```diff
@@ -0,0 +1,158 @@
...
+                result.put(num, evmAddress);
+            }
+        } catch (Exception e) {
+            log.error("Error looking up EVM addresses for entity IDs: {}", evmAddressLookupIds, e);
+        }
+
+        return result;
+    }
+
+    private List<EvmAddressMapping> findEvmAddressesByIds(Set<EntityId> ids) {
```
steven-sheehy: Postgresql has a limit to how many items in an in clause. We should paginate this querying so it doesn't exceed this limit and throw an error.

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListener.java
```diff
@@ -0,0 +1,158 @@
...
+            var dbResults = findEvmAddressesByIds(missing).stream()
+                    .collect(Collectors.toMap(EvmAddressMapping::num, EvmAddressMapping::evmAddress));
+
+            // Fill in results from DB (or default if still missing)
+            for (var entityId : missing) {
+                long num = entityId.getNum();
+                byte[] evmAddress = dbResults.getOrDefault(num, trim(toEvmAddress(entityId)));
+                evmCache.put(num, evmAddress);
+                result.put(num, evmAddress);
+            }
```
steven-sheehy: This method is overly complex and has too many loops. We can simplify with a Caffeine `LoadingCache.getAll(keys)` which internally handles separating the keys into cache hits and misses and calls the loader with the misses.

common/src/main/java/org/hiero/mirror/common/util/DomainUtils.java
```diff
@@ -285,6 +285,39 @@ public static EntityId fromEvmAddress(byte[] evmAddress) {
         return null;
     }
 
+    public static EntityId fromTrimmedEvmAddress(byte[] evmAddress) {
```
IvanKavaldzhiev: Argument can be final

common/src/main/java/org/hiero/mirror/common/util/DomainUtils.java
```diff
@@ -285,6 +285,39 @@ public static EntityId fromEvmAddress(byte[] evmAddress) {
...
+        if (evmAddress == null) {
+            return null;
+        }
+
+        var padding = new byte[EVM_ADDRESS_LENGTH - evmAddress.length];
+
+        return fromEvmAddress(Bytes.concat(padding, evmAddress));
+    }
+
+    public static byte[] trim(byte[] data) {
```
IvanKavaldzhiev: Argument can be final

importer/src/main/java/org/hiero/mirror/importer/domain/EntityIdServiceImpl.java
```diff
@@ -38,10 +37,12 @@ public class EntityIdServiceImpl implements EntityIdService {
     private static final Optional<EntityId> EMPTY = Optional.of(EntityId.EMPTY);
 
     private final Cache cache;
+
     private final EntityRepository entityRepository;
 
-    public EntityIdServiceImpl(@Qualifier(CACHE_ALIAS) CacheManager cacheManager, EntityRepository entityRepository) {
-        this.cache = cacheManager.getCache(CACHE_NAME);
+    public EntityIdServiceImpl(
```
IvanKavaldzhiev: All arguments can be final

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListener.java
```diff
@@ -0,0 +1,158 @@
...
+public class SyntheticLogListener implements EntityListener, RecordStreamFileListener {
+    private static final String LOOKUP_QUERY =
+            "select evm_address, num from entity where id in (:ids) and length(evm_address) > 0";
+    private final NamedParameterJdbcTemplate jdbcTemplate;
+    private final Cache evmCache;
+    private final EntityProperties entityProperties;
+    private final List<SyntheticLogUpdater> syntheticLogParticipants = new ArrayList<>();
+    private final Set<EntityId> evmAddressLookupIds = new HashSet<>();
+
+    public SyntheticLogListener(
```
IvanKavaldzhiev: All arguments here and below can be final

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListener.java
jnels124: Did that originally but will have to add the property to the default yaml and these persist properties aren't generally defaulted there.

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListener.java
jnels124: We would have to loop through all contract logs on the context to find relevant ones, not just the synthetics. I can do that though

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListener.java
jnels124: the in clause limit is much less than what our tps allows. Will batch to 10k max

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListener.java
jnels124: that doesn't work with the spring wrapper. We could use the native cache though

common/src/test/java/org/hiero/mirror/common/util/DomainUtilsTest.java
jnels124: Think my settings were lost on an upgrade. will fix

importer/src/main/java/org/hiero/mirror/importer/domain/EntityIdServiceImpl.java
jnels124: I agree with this but we don't do that in most of the importer code when possible. I can change though.

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListener.java
```diff
@@ -0,0 +1,158 @@
...
+    @Timed
+    public void onEnd(RecordFile recordFile) {
+        updateSyntheticContractLogs();
+        syntheticLogParticipants.clear();
+        evmAddressLookupIds.clear();
+    }
+
+    @Override
+    public void onContractLog(ContractLog contractLog) {
+        if (entityProperties.getPersist().isSyntheticContractLogEvmAddressLookup()
```
xin-hedera: suggest adding transient `boolean synthetic` property to `ContractLog`, and skip non-synthetic contract logs here.

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListener.java
```diff
@@ -0,0 +1,158 @@
...
+            }
+        }
+    }
+
+    private void updateSyntheticContractLogs() {
+        var entityMap = lookupEvmAddresses();
+        syntheticLogParticipants.forEach(updater -> updater.updateContractLog(entityMap));
+    }
+
+    private Map<Long, byte[]> lookupEvmAddresses() {
```
xin-hedera: entities created with evm address in the same block are not handled.

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListener.java
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    record EvmAddressMapping(byte[] evmAddress, long num) {}
+
+    private record SyntheticLogUpdater(EntityId sender, EntityId receiver, ContractLog contractLog) {
+
+        public List<EntityId> getSearchIds() {
+            var ids = new ArrayList<EntityId>();
+
+            if (receiver != null && receiver != EntityId.EMPTY) {
```
xin-hedera: use `EntityId.isEmpty`

importer/src/main/resources/db/migration/v1/V1.110.0__fix_transfer_synthetic_contract_logs.sql
```diff
@@ -0,0 +1,48 @@
...
+          substr(topic1, 1, 12) = repeat(e'\\000', 12)::bytea or
+          substr(topic2, 1, 12) = repeat(e'\\000', 12)::bytea
```
xin-hedera: this is not needed. for releases < 0.133.0, the entity id topics in the synthetic transfer logs are always of 8 bytes. for releases >= 0.133.0, the leading 0s are trimmed.

importer/src/main/resources/db/migration/v1/V1.110.0__fix_transfer_synthetic_contract_logs.sql
```diff
@@ -0,0 +1,48 @@
...
+    topic2 = coalesce(e2.evm_address, cl.topic2)
+from cl_ext
+left join entity_addresses e1 on e1.num = (
+    ('x' || (
+        case
+            when length(right(cl_ext.topic1_hex, 16)) % 2 = 1
+            then '0' || right(cl_ext.topic1_hex, 16)
+            else right(cl_ext.topic1_hex, 16)
+        end
+    ))::bit(64)::bigint
```
xin-hedera: this can be simplified with the `topic length <= 8` change, sth like `('x' || lpad(topic1_hex,16,'0'))::bit(64)::bigint`

importer/src/main/resources/db/migration/v1/V1.110.0__fix_transfer_synthetic_contract_logs.sql
```diff
@@ -0,0 +1,48 @@
+with entity_addresses as (
+    select num, evm_address
+    from entity
+    where length(evm_address) > 0
+),
```
xin-hedera, line 5: don't think this provides value or better performance. the query plan will likely choose the primary key

importer/src/main/resources/db/migration/v1/V1.110.0__fix_transfer_synthetic_contract_logs.sql
```diff
@@ -0,0 +1,48 @@
...
+          substr(topic1, 1, 12) = repeat(e'\\000', 12)::bytea or
+          substr(topic2, 1, 12) = repeat(e'\\000', 12)::bytea
+      )
+)
+
+update contract_log cl
+set topic1 = coalesce(e1.evm_address, cl.topic1),
+    topic2 = coalesce(e2.evm_address, cl.topic2)
+from cl_ext
+left join entity_addresses e1 on e1.num = (
```
xin-hedera: there's no index on `num`, and so far in mainnet, we always have `entity.id == entity.num`

```suggestion
left join entity_addresses e1 on e1.id = (
```

xin-hedera (changes requested):

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListener.java
steven-sheehy: Yes, LoadingCache is a Caffeine API

importer/src/main/resources/db/migration/v1/V1.110.0__fix_transfer_synthetic_contract_logs.sql
jnels124: Can we make this change and assume it will never be ran on shard/realm != 0? I left it this way because that and the migration still completes in < 10 minutes with mainnet data

importer/src/main/resources/db/migration/v1/V1.110.0__fix_transfer_synthetic_contract_logs.sql
jnels124: this makes the data pushed to workers smaller. Can make the change though

importer/src/main/resources/db/migration/v1/V1.110.0__fix_transfer_synthetic_contract_logs.sql
xin-hedera: yeah. If a third party runs a really old release on a non-zero realm network, they will find out later they have to restart with a compatible version.

If it's a release with the fix and ran on a non-zero realm network, the tables should be empty and the migration shouldn't change any data.

importer/src/main/resources/db/migration/v1/V1.110.0__fix_transfer_synthetic_contract_logs.sql
jnels124: actually this is needed for the coalesce. We dont want to overwrite the topic fields with an empty evm address

xin-hedera: yeah, we should add the length check in the left join when removing the CTE. However think we should keep the CTE as for citus it'll perform better. And we should then keep the `num = entity num from topic` condition in the left join too.

importer/src/main/resources/db/migration/v1/V1.110.0__fix_transfer_synthetic_contract_logs.sql
xin-hedera: still need to add the condition `octet_length(topic) <= 8` to each of the left joins

importer/src/test/java/org/hiero/mirror/importer/migration/UpdateSyntheticContractLogsMigrationTest.java
xin-hedera: add the following annotations
```
@DisablePartitionMaintenance
@DisableRepeatableSqlMigration
```

the migration test may fail at certain point without disabling those.

importer/src/test/java/org/hiero/mirror/importer/migration/UpdateSyntheticContractLogsMigrationTest.java
```diff
@@ -0,0 +1,114 @@
...
+    @Owner
+    @Resource
+    private JdbcTemplate ownerJdbcTemplate;
```
xin-hedera: already in parent class

```suggestion
```

importer/src/test/java/org/hiero/mirror/importer/migration/UpdateSyntheticContractLogsMigrationTest.java
```diff
@@ -0,0 +1,114 @@
...
+
+    @Resource
+    private ContractLogRepository contractLogRepository;
+
+    @Resource
+    private DomainBuilder domainBuilder;
+
+    @Getter
+    @Value("#{environment.matchesProfiles('!v2')}")
+    private boolean v1;
```
xin-hedera: already in parent classes

```suggestion
```

importer/src/test/java/org/hiero/mirror/importer/migration/UpdateSyntheticContractLogsMigrationTest.java
xin-hedera: there should be a `ContextConfiguration` annotation with a class setting flyway target right before the versioned migration this class tests against. There's no problem now however if there's ever a later migration changes the schemas this migration depends on, it'll start to fail.

https://github.com/hiero-ledger/hiero-mirror-node/blob/b06f49b4fad73486817c161c895ab89a36ebbaeb/importer/src/test/java/org/hiero/mirror/importer/migration/FixAsyncJavaMigrationHistoryMigrationTest.java#L28

importer/src/test/java/org/hiero/mirror/importer/parser/contractlog/SyntheticContractLogTest.java
```diff
@@ -56,14 +57,14 @@ void booleanToBytes() {
             00000000, ''
             '', ''
             """)
-    void trim(String inputHex, String expectedHex) {
```
xin-hedera, line 59: should move these trim test cases to `DomainUtilsTest.java`

importer/src/test/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListenerTest.java
```diff
@@ -0,0 +1,290 @@
...
+import org.mockito.Captor;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+import org.springframework.cache.Cache;
+import org.springframework.cache.CacheManager;
+import org.springframework.jdbc.core.RowMapper;
+import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
+import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
+
+@Slf4j
```
xin-hedera:
```suggestion
```

importer/src/test/java/org/hiero/mirror/importer/parser/contractlog/SyntheticContractLogTest.java
jnels124: tests were already added there. Can remove these if we don't want verified here

xin-hedera: yes we should remove these duplicate tests.

importer/src/main/java/org/hiero/mirror/importer/domain/EntityIdServiceImpl.java
IvanKavaldzhiev: I see, then there is no problem.

docs/configuration.md
```diff
@@ -54,7 +54,7 @@ value, it is recommended to only populate overridden properties in the custom `a
...
-| `hiero.mirror.importer.cache.timePartition`                                     | maximumSize=50,expireAfterWrite=1d,recordStats       | The Caffeine cache specification for time partition lookups.                                                                                                                                                                                                       |
+| `hiero.mirror.importer.cache.evmAddress`                                        | maximumSize=500000,recordStats                       | The Caffeine cache specification for encoded entity id to trimmed evm address or long zero mapping                                                                                                                                                                 |     | `hiero.mirror.importer.cache.timePartition` | maximumSize=50,expireAfterWrite=1d,recordStats | The Caffeine cache specification for time partition lookups. |
```
xin-hedera: looks like two lines ended up to be one single line

importer/src/main/java/org/hiero/mirror/importer/repository/EntityRepository.java
```diff
@@ -2,22 +2,26 @@
...
 public interface EntityRepository extends CrudRepository<Entity, Long> {
-
     @Query(value = "select id from entity where alias = ?1 and deleted <> true", nativeQuery = true)
     Optional<Long> findByAlias(byte[] alias);
 
     @Query(value = "select id from entity where evm_address = ?1 and deleted <> true", nativeQuery = true)
     Optional<Long> findByEvmAddress(byte[] evmAddress);
 
+    @Query(value = "select evm_address,id from entity where id in (?1)", nativeQuery = true)
+    List<EvmAddressMapping> findEvmAddressesByIds(Iterable<? extends Long> ids);
```
xin-hedera, line 23: why not just return `List<Entity>`? Also the query should exclude entities without evm address, reason is it should be faster for the application code to find out from the query result which entities don't have evm address, compared to returning all entity info with a null evm address indicating the same.

importer/src/main/resources/application.yml
```diff
@@ -34,6 +34,11 @@ hiero:
           checksum: 2
         tokenAccountBalanceMigration:
           checksum: 3
+      parser:
+        record:
+          entity:
+            persist:
+              synthetic-contract-log-evm-address-lookup: true
```
xin-hedera:
```suggestion
              syntheticContractLogEvmAddressLookup: true
```

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListener.java
```diff
@@ -0,0 +1,185 @@
...
+    private final CacheProperties cacheProperties;
+    private final EntityRepository entityRepository;
+
+    @Override
+    @Timed
+    public void onEnd(RecordFile recordFile) {
+        final var candidates = parserContext.get(ContractLog.class).stream()
+                .map(this::toLogUpdater)
+                .filter(Objects::nonNull)
+                .toList();
```
xin-hedera: should avoid the looping here, prefer the old approach that `onContractLog` adds synthetic transfer log into a list.

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListener.java
```diff
@@ -0,0 +1,185 @@
...
+            return Collections.emptyMap();
+        }
+
+        Map<Long, byte[]> loadedFromDb = new HashMap<>(keys.size());
+
+        if (keys.size() <= MAX_CACHE_LOAD_ENTRIES) {
+            processBatch(loadedFromDb, entityRepository.findEvmAddressesByIds(keys));
+        } else {
+            var i = 0;
+            final var keyList = new ArrayList<>(keys);
```
xin-hedera: can use guava `Iterables.partition`

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListener.java
```diff
@@ -0,0 +1,185 @@
...
+                result.put(mapping.getId(), trim(mapping.getEvmAddress()));
+            }
+        }
+    }
+
+    private SyntheticLogUpdater toLogUpdater(ContractLog contractLog) {
+        if (contractLog.isSyntheticTransfer()) {
+            var senderId = fromTrimmedEvmAddress(contractLog.getTopic1());
+            var receiverId = fromTrimmedEvmAddress(contractLog.getTopic2());
+            if (!(EntityId.isEmpty(senderId) && EntityId.isEmpty(receiverId))) {
```
xin-hedera: this is redundant since there's empty check in `addNewEntityToCache`
```suggestion
```

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListener.java
```diff
@@ -0,0 +1,185 @@
...
+    private void addNewEntityToCache(EntityId entityId, byte[] defaultValue) {
+        if (!EntityId.isEmpty(entityId)) {
+            var contextEntity = parserContext.get(Entity.class, entityId.getId());
+            if (contextEntity != null && !contextEntity.hasHistory()) {
+                var evmAddress = contextEntity.getEvmAddress();
+                if (evmAddress != null && evmAddress.length > 0) {
+                    getEvmCache().put(entityId.getId(), trim(evmAddress));
+                } else {
+                    getEvmCache().put(entityId.getId(), defaultValue);
+                }
```
xin-hedera: this is wrong. an empty evm address in a partial `Entity` object != the entity doesn't have evm address. We can add a has `createdTimestamp` check though.

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListener.java
```diff
@@ -0,0 +1,185 @@
...
+        }
+        return null;
+    }
+
+    private void addNewEntityToCache(EntityId entityId, byte[] defaultValue) {
+        if (!EntityId.isEmpty(entityId)) {
+            var contextEntity = parserContext.get(Entity.class, entityId.getId());
+            if (contextEntity != null && !contextEntity.hasHistory()) {
+                var evmAddress = contextEntity.getEvmAddress();
+                if (evmAddress != null && evmAddress.length > 0) {
```
xin-hedera: `!ArrayUtils.isEmpty(...)`

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListener.java
```diff
@@ -0,0 +1,185 @@
...
+                return new SyntheticLogUpdater(senderId, receiverId, contractLog);
+            }
+        }
+        return null;
+    }
+
+    private void addNewEntityToCache(EntityId entityId, byte[] defaultValue) {
+        if (!EntityId.isEmpty(entityId)) {
+            var contextEntity = parserContext.get(Entity.class, entityId.getId());
+            if (contextEntity != null && !contextEntity.hasHistory()) {
```
xin-hedera: it doesn't matter if it has history or not, if there's evm address from the beginning, it should get merged to all subsequent Entity objects.
```suggestion
            if (contextEntity != null) {
```

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListener.java
```diff
@@ -0,0 +1,185 @@
...
+                } else {
+                    getEvmCache().put(entityId.getId(), defaultValue);
+                }
+            }
+        }
+    }
+
+    private record SyntheticLogUpdater(EntityId sender, EntityId receiver, ContractLog contractLog) {
+
+        public List<Long> getSearchIds() {
```
xin-hedera: should avoid allocating an list (and sometimes it's empty), then add all elements to the hash set, perhaps just change it to `public void populateSearchIds(Set<Long> ids)`

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListener.java
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    private SyntheticLogUpdater toLogUpdater(ContractLog contractLog) {
+        if (contractLog.isSyntheticTransfer()) {
+            var senderId = fromTrimmedEvmAddress(contractLog.getTopic1());
+            var receiverId = fromTrimmedEvmAddress(contractLog.getTopic2());
+            if (!(EntityId.isEmpty(senderId) && EntityId.isEmpty(receiverId))) {
+
+                addNewEntityToCache(senderId, contractLog.getTopic1());
+                addNewEntityToCache(receiverId, contractLog.getTopic2());
```
xin-hedera: nit, it can be further optimized to only set senderId / receiverId for the new `SyntheticLogUpdater` when lookup is needed.

xin-hedera (changes requested):

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListener.java
jnels124: Lookup is needed for all synthetic transfer contract logs as we always set the long zero. I can add a check though in case that changes

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListener.java
jnels124: I thought evm address can only be set on initial creation. Will fix

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListener.java
xin-hedera: we should not do look up for an entity within parser context that we are sure it either has evm address or doesn't. in other words, we need the lookup for entities that we are not sure if it has evm address or not

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListener.java
jnels124: if we call this method, there will be at least one item added to it. We only create the updater in cases where there is at least one of the two populated.

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListener.java
jnels124: this is a check to not create the updater if it isn't necessary

importer/src/main/java/org/hiero/mirror/importer/repository/EntityRepository.java
jnels124: we can do that but will have to return the whole entity object because jpa can not map to partial object. the evm address check was left off so we get all requested keys in order to populate the cache. Can change that though and populate that another way (though this may be the most efficient. Will double check)

jnels124: I will change the log updater to be a class (to access the parent class cache) and can update the cache with relevant values (the trimmed long zero if there is no entry for the provided key)

jnels124: Actually that doesn't work because the loadAll on the cache will already loop through all keys and add if the result doesn't have an entry for each key. Still think it would be better to have the db return all results to avoid the loop there. That loop exists for partial mirror nodes that may not have all entities

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/topic/TopicMessageLookupEntityListener.java
```diff
@@ -18,7 +18,7 @@
 import org.springframework.core.annotation.Order;
 
 @Named
-@Order(3)
+@Order(4)
```
steven-sheehy, line 21: Can revert and revisit order changes across all listeners with change to only use record file listener.

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListener.java
```diff
@@ -0,0 +1,202 @@
...
+@CustomLog
+@RequiredArgsConstructor
+@ConditionalOnProperty(
+        name = "hiero.mirror.importer.parser.record.entity.persist.synthetic-contract-log-evm-address-lookup",
+        havingValue = "true")
+final class SyntheticLogListener implements EntityListener, RecordStreamFileListener {
+    static final int MAX_CACHE_LOAD_ENTRIES = 30000;
+
+    @Getter(lazy = true)
+    private final LoadingCache<Long, byte[]> evmCache = buildCache();
```
jnels124, line 49: If wanted, we can still put this in cache configuration but loader function will need to be there too. Then we can use .nativeCache() to get access to methods we need (getAll)

importer/src/test/java/org/hiero/mirror/importer/parser/record/entity/ParserContextTest.java
```diff
@@ -54,6 +56,16 @@ void get() {
...
     }
 
+    @Test
+    void getTransient() {
+        assertThat(parserContext.getTransient(Entity.class)).isEmpty();
+        assertThatThrownBy(() -> parserContext.getTransient(null)).isInstanceOf(NullPointerException.class);
+
+        var domain = domainBuilder.entity().get();
+        parserContext.merge(domain.getId(), domain, (a, b) -> a);
+        assertThat(parserContext.get(Entity.class, domain.getId())).isEqualTo(domain);
```
jnels124, line 66: will fix

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListener.java
xin-hedera: don't think this is addressed

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListener.java
```diff
@@ -0,0 +1,181 @@
...
+            if (!EntityId.isEmpty(sender)) {
+                parserContext.getEvmAddressLookupIds().add(sender.getId());
+            }
+        }
+
+        public void updateContractLog(Map<Long, byte[]> entityEvmAddresses) {
+            updateTopicField(
+                    sender, entityEvmAddresses.get(sender.getId()), contractLog::setTopic1, contractLog::getTopic1);
+            updateTopicField(
+                    receiver, entityEvmAddresses.get(receiver.getId()), contractLog::setTopic2, contractLog::getTopic2);
```
xin-hedera: can sender or receiver be `null`?

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListener.java
```diff
@@ -0,0 +1,181 @@
...
+            if (mapping.getEvmAddress() == null || mapping.getEvmAddress().length == 0) {
+                result.put(mapping.getId(), trim(toEvmAddress(EntityId.of(mapping.getId()))));
+            } else {
+                result.put(mapping.getId(), trim(mapping.getEvmAddress()));
+            }
+        }
+    }
+
+    private SyntheticLogUpdater toLogUpdater(ContractLog contractLog) {
+        if (contractLog.isSyntheticTransfer()) {
```
xin-hedera: can remove it now since the caller checks `isSyntheticTransfer`
```suggestion
```

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListener.java
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    @Override
+    public void onContractLog(ContractLog contractLog) {
+        if (contractLog.isSyntheticTransfer()) {
+            var updater = toLogUpdater(contractLog);
+            if (updater != null) {
+                updater.populateSearchIds();
+                parserContext.addTransient(updater);
+            }
```
xin-hedera: nit, can probably remove `toLogUpdater` and move the logic here, it's not that many lines of code and combining makes it easier to read.

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListener.java
```diff
@@ -0,0 +1,181 @@
...
+        }
+
+        return loadedFromDb;
+    }
+
+    private void processBatch(Map<Long, byte[]> result, List<EvmAddressMapping> evmAddressMappings) {
+        for (var mapping : evmAddressMappings) {
+            if (mapping.getEvmAddress() == null || mapping.getEvmAddress().length == 0) {
+                result.put(mapping.getId(), trim(toEvmAddress(EntityId.of(mapping.getId()))));
```
xin-hedera: this is dead code now because the repository method will only return entities with evm address

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListener.java
```diff
@@ -0,0 +1,181 @@
...
+                List<? extends Long> batch = batches.next();
+                processBatch(loadedFromDb, entityRepository.findEvmAddressesByIds(batch));
+            }
+        }
+
+        return loadedFromDb;
+    }
+
+    private void processBatch(Map<Long, byte[]> result, List<EvmAddressMapping> evmAddressMappings) {
+        for (var mapping : evmAddressMappings) {
```
xin-hedera, line 117: nit, move `entityRepository.findEvmAddressesByIds(keys)` into `processBatch`

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListener.java
```diff
@@ -0,0 +1,181 @@
...
+                if (cachedResult != null) {
+                    setter.accept(cachedResult);
+                } else {
+                    var contextEntity = parserContext.get(Entity.class, key.getId());
+                    if (contextEntity != null && !ArrayUtils.isEmpty(contextEntity.getEvmAddress())) {
+                        var trimmedEvmAddress = trim(contextEntity.getEvmAddress());
+                        getEvmCache().put(key.getId(), trimmedEvmAddress);
+                        setter.accept(trimmedEvmAddress);
+                    } else {
+                        getEvmCache().put(key.getId(), defaultValue.get());
```
xin-hedera: good to have a comment why we are sure the entity doesn't have evm address

xin-hedera (changes requested):

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListener.java
xin-hedera: never mind, the logic is changed slightly so all good.

importer/src/main/java/org/hiero/mirror/importer/repository/EntityRepository.java
```diff
@@ -2,22 +2,26 @@
...
 public interface EntityRepository extends CrudRepository<Entity, Long> {
-
     @Query(value = "select id from entity where alias = ?1 and deleted <> true", nativeQuery = true)
     Optional<Long> findByAlias(byte[] alias);
 
     @Query(value = "select id from entity where evm_address = ?1 and deleted <> true", nativeQuery = true)
     Optional<Long> findByEvmAddress(byte[] evmAddress);
 
+    @Query(value = "select evm_address,id from entity where id in (?1) and length(evm_address) > 0", nativeQuery = true)
+    List<EvmAddressMapping> findEvmAddressesByIds(Iterable<? extends Long> ids);
```
steven-sheehy, line 23: Need a repository test

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/ParserContext.java
```diff
@@ -20,9 +20,16 @@
  */
 @Named
 public class ParserContext {
+    @Getter
+    private List<Long> evmAddressLookupIds = new ArrayList<>();
```
steven-sheehy: The rest of the context doesn't expose the data structures without wrapping them in unmodifiable. We should do the same here.

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/topic/TopicMessageLookupEntityListener.java
steven-sheehy: Think this was a comment on an early commit and might not be applicable. Please just re-check orders are correct with latest.

jnels124: oops missed this one and will double check

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticLogListener.java
jnels124: yes definitely can. Will fix

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/topic/TopicMessageLookupEntityListener.java
jnels124: ordering is correct.

steven-sheehy (dismissed):

xin-hedera (dismissed):
LGTM

xin-hedera (approved):
LGTM

PR #11479, opened 2025-06-26
Generate table usage for nodejs REST API

### LINKED ISSUE #11266: Generate a report of table usage
### Problem

We'd like to know which modules and APIs use which tables for documentation and refactoring purposes.

### Solution

* In REST API, add a hook in spec test execution or query execution that captures which tables are queried broken down by endpoint
* In Java modules, research a way to intercept calls to repository broken down by endpoint
* Generate a markdown report in each project build directory with the results

### Alternatives

_No response_

xin-hedera:
**Description**:

This PR adds support of generating table usage for nodejs REST APIs

- Add SQL query tracking during spec based integration tests
- Analyze table usage from the saved queries and write info to file in spec test module `afterAll` hook
- Generate report in `globalTeardown`
- Patch error stacktrace for correct function name

**Related issue(s)**:

Close #11266

**Notes for reviewer**:

Example markdown report (trimmed down)

##### By Endpoint

| Endpoint | Tables |
|----------|--------|
| /api/v1/accounts | entity,entity_stake,token_account|
| /api/v1/accounts/{id} | account_balance,crypto_transfer,entity,entity_history,entity_stake,entity_stake_history,staking_reward_transfer,token_account,token_balance,token_transfer,transaction|
| /api/v1/accounts/{id}/allowances/crypto | crypto_allowance,entity|
| /api/v1/accounts/{id}/allowances/tokens | entity,token_allowance|

##### By Table

| Table | Endpoints |
|-------|-----------|
| account_balance | /api/v1/accounts/{id},/api/v1/balances,/api/v1/network/supply,/api/v1/tokens/{id}/balances |
| address_book | /api/v1/network/nodes,/api/v1/transactions/{id}/stateproof |
| address_book_entry | /api/v1/network/nodes,/api/v1/transactions/{id}/stateproof |
| address_book_service_endpoint | /api/v1/network/nodes |

Sample csv:

```
Endpoint,Source,Table
/api/v1/accounts,getAccounts (accounts.js:361),entity
/api/v1/accounts,getAccounts (accounts.js:361),entity_stake
/api/v1/accounts,getAccounts (accounts.js:361),token_account
/api/v1/accounts/{id},func (timestampRange.js:51),transaction
/api/v1/accounts/{id},getAccountBalanceTimestampRange (balances.js:189),account_balance
```

Note if the query calls a UDF, the logic isn't able to find the tables the UDF accesses, that's reflected in get transaction by transaciton hash API

The overhead when running all tests is about 7.5% locally, 67s v.s. 72s

**Checklist**

### DIFF
--- a/rest/__tests__/globalTeardown.js
+++ b/rest/__tests__/globalTeardown.js
@@ -1,8 +1,97 @@
 // SPDX-License-Identifier: Apache-2.0
 
+import fs from 'fs';
 import log4js from 'log4js';
 
+import {TABLE_USAGE_OUTPUT_DIR} from './testutils.js';
+
+const CSV_HEADER = 'Endpoint,Source,Table\n';
+const MARKDOWN_ENDPOINT_HEADER = `### By Endpoint\n
+| Endpoint | Tables |
+|----------|--------|\n`;
+const MARKDOWN_TABLE_HEADER = `\n### By Table\n
+| Table | Endpoints |
+|-------|-----------|\n`;
+const REPORT_FILENAME = 'table-usage';
+
+const createTableUsageReport = () => {
+  if (process.env.GENERATE_TABLE_USAGE === 'false' || !fs.existsSync(TABLE_USAGE_OUTPUT_DIR)) {
+    return;
+  }
+
+  const tableUsage = {};
+  for (const dirent of fs.readdirSync(TABLE_USAGE_OUTPUT_DIR, {withFileTypes: true})) {
+    if (!dirent.isFile() || !dirent.name.endsWith('.json')) {
+      continue;
+    }
+
+    try {
+      Object.assign(
+        tableUsage,
+        JSON.parse(fs.readFileSync(`${TABLE_USAGE_OUTPUT_DIR}/${dirent.name}`, {encoding: 'utf8'}))
+      );
+    } catch (err) {
+      console.error(`Unable to read file ${dirent.name} - ${err}`);
+    }
+  }
+
+  writeReport(tableUsage, `${REPORT_FILENAME}.csv`, writeCsvReport);
+  writeReport(tableUsage, `${REPORT_FILENAME}.md`, writeMarkdownReport);
+};
+
+const writeCsvReport = (data, writeStream) => {
+  writeStream.write(CSV_HEADER);
+  for (const endpoint of Object.keys(data).sort()) {
+    const callerTables = data[endpoint];
+    for (const caller of Object.keys(callerTables).sort()) {
+      for (const table of callerTables[caller]) {
+        writeStream.write(`${endpoint},${caller},${table}\n`);
+      }
+    }
+  }
+};
+
+const writeMarkdownReport = (data, writeStream) => {
+  // By endpoint
+  writeStream.write(MARKDOWN_ENDPOINT_HEADER);
+  const usageByTable = {};
+  for (const endpoint of Object.keys(data).sort()) {
+    const callerTables = data[endpoint];
+    const tables = Array.from(new Set(Object.values(callerTables).flat())).sort();
+    writeStream.write(`| ${endpoint} | ${tables.join(',')}|\n`);
+
+    tables.forEach((table) => {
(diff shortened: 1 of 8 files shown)


filev94:
Regarding the output format, wouldn’t it be more useful if we could clearly associate which tables are accessed by which repositories for each endpoint? Maybe something like this?

image

steven-sheehy:
If it's easy enough, we could produce two tables in the same report: grouped by endpoint and grouped by table. For each, we should make the grouping column the first column and there should only be one distinct entry per row in that column.

rest/__tests__/globalTeardown.js
```diff
@@ -1,8 +1,93 @@
...
+
+const CSV_HEADER = 'Endpoint,Source,Table\n';
+const MARKDOWN_ENDPOINT_HEADER = `| Endpoint | Tables |
+|----------|--------|\n`;
+const MARKDOWN_TABLE_HEADER = `| Table | Endpoints |
+|-------|-----------|\n`;
+const REPORT_FILENAME = 'table-usage';
+
+const createTableUsageReport = () => {
+  if (process.env.NO_GENERATE_TABLE_USAGE === 'true' || !fs.existsSync(TABLE_USAGE_OUTPUT_DIR)) {
```
steven-sheehy: We should use affirmatively named variables.

```suggestion
  if (process.env.GENERATE_TABLE_USAGE === 'false' || !fs.existsSync(TABLE_USAGE_OUTPUT_DIR)) {
```

rest/__tests__/globalTeardown.js
```diff
@@ -1,8 +1,93 @@
...
+
+  // By table
+  writeStream.write('\n### By Table\n\n');
+  writeStream.write(MARKDOWN_TABLE_HEADER);
+
+  for (const table of Object.keys(usageByTable).sort()) {
+    writeStream.write(`| ${table} | ${usageByTable[table].sort().join(',')} |\n`);
+  }
+
+  writeStream.close();
```
steven-sheehy: Probably should be done in a finally block. Same with csv report.

rest/__tests__/globalTeardown.js
```diff
@@ -1,8 +1,93 @@
...
+};
+
+const writeMarkdownReport = (tableUsage) => {
+  const writeStream = fs.createWriteStream(`${TABLE_USAGE_OUTPUT_DIR}/${REPORT_FILENAME}.md`);
+
+  // By endpoint
+  writeStream.write('### By Endpoint\n\n');
+  writeStream.write(MARKDOWN_ENDPOINT_HEADER);
```
steven-sheehy: Kind of weird to have inline section header and constant table header. Should either combine into single constant or make both inline. My vote is just make table header an inline string to since it's only used here and improves readability.

rest/__tests__/globalTeardown.js
```diff
@@ -1,8 +1,93 @@
...
+  const tableUsage = {};
+  for (const dirent of fs.readdirSync(TABLE_USAGE_OUTPUT_DIR, {withFileTypes: true})) {
+    if (!dirent.isFile() || !dirent.name.endsWith('.json')) {
+      continue;
+    }
+
+    try {
+      Object.assign(
+        tableUsage,
+        JSON.parse(fs.readFileSync(`${TABLE_USAGE_OUTPUT_DIR}/${dirent.name}`, {encoding: 'utf8'}))
```
steven-sheehy, line 31: Is it guaranteed that separate jest processes won't write to the same endpoint? If not, won't there be issues in overwriting some entries?

xin-hedera: guaranteed per jest runtime model and how our spec based tests are organized.

One spec based integration test module (*.spec.test.js file) runs all spec files for a specific group, and a test js file as a module can only and will only run by a single jest worker. The json file is only written in the top level afterAll hook which should run exactly once.

steven-sheehy (approved):
LGTM

jnels124 (approved):

PR #11511, opened 2025-07-03
Upgrade to express 5.1.0

### LINKED ISSUE #10864: Upgrade to Express 5.1
### Problem

Express 5.1 has been released and several components in our app need to be updated

### Solution

[migration docs](https://expressjs.com/en/guide/migrating-5.html).

Note that the provided `@expressjs/codemod` doesn't seem to update our codebase appropriately. 

Known issues with upgrade:
- `@awaitjs/express` is no longer necessary and is not compatible with express 5.1. All uses of this library should be replaced and express methods used directly. 
- handlers don't appear to automatically call next() after the handler is complete. This is because no response is sent from these handlers and is sent by response handler middleware instead
- Any routes with regex matches need to be evaluated and verified and updated. We have atleast one (in ContractRoutes) that needs to be updated.

- [ ] Successfully migrated to 5.1 w/ passing test run
- [ ] Passing performance tests

### Alternatives

_No response_

xin-hedera:
**Description**:

This PR upgrades REST module to express 5.1.0

- Upgrade to express 5.1.0
- Apply @awaitjs/express logic to extend express to automatically call next()
- Add conflict resolution for `/contracts/{id}/results/logs` and `/contracts/{id}/results/{consensusTimestamp}`

**Related issue(s)**:

Fixes #10864

**Notes for reviewer**:

k6 tests show comparable results. Only outlier is `schedules`, a 50% reduction in pass RPS compared to v0.133.0-rc1, however almost the same compared to v0.132.0-rc2. 

**Checklist**


rest/routes/contractRoute.js
```diff
@@ -1,23 +1,25 @@
...
-router.getAsync('/results/logs', ContractController.getContractLogs);
-router.getAsync('/results/:transactionIdOrHash', ContractController.getContractResultsByTransactionIdOrHash);
-router.getAsync('/results/:transactionIdOrHash/actions', ContractController.getContractActions);
+router.getExt('/', ContractController.getContracts);
+router.getExt('/:contractId', ContractController.getContractById);
+router.getExt('/:contractId/results', ContractController.getContractResultsById);
+router.getExt('/:contractId/state', ContractController.getContractStateById);
+router.getExt('/:contractId/results/logs', ContractController.getContractLogsById);
+// must add after '/:contractId/results/logs' for proper conflict resolution
+router.getExt('/:contractId/results/:consensusTimestamp', ContractController.getContractResultsByTimestamp);
```
xin-hedera, line 18: express 5.x doesn't support regex in path variable. this also trap us into path conflict between `/:contractId/results/logs` and `/:contractId/results/:consensusTimestamp`

steven-sheehy (approved):

PR #11512, opened 2025-07-03
Support endBlockNumber

### LINKED ISSUE #11435: HIP-1081 Support endBlockNumber
### Problem

As a mirrornode operator, I want to configure importer to ingest blocks up to a known block number.

### Solution

Add a `endBlockNumber` config properties
- the default is null, means ingest blocks infinitely
- if set to a valid non-negative value, it's the last block importer will ingest and parse

### Alternatives

_No response_

steven-sheehy:
**Description**:

* Add `hiero.mirror.importer.endBlockNumber` property
* Fix `The cache 'default' is not recording statistics` for `contractSlots` cache in web3
* Integrate new property into both `BlockFileSource` and `BlockNodeSubscriber`
* Reduce error logs in `BlockFileSource` when missing file in bucket for node

**Related issue(s)**:

Fixes #11435

**Notes for reviewer**:

**Checklist**


importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockNode.java
```diff
@@ -40,16 +41,15 @@ final class BlockNode implements AutoCloseable, Comparable<BlockNode> {
     private static final long INFINITE_END_BLOCK_NUMBER = -1;
```
jnels124: dont think this is used anymore

steven-sheehy: Thanks, I've removed it.

importer/src/main/java/org/hiero/mirror/importer/ImporterProperties.java
```diff
@@ -36,6 +37,13 @@ public class ImporterProperties {
     @NotNull
     private Path dataPath = Paths.get(".", "data");
 
+    @EqualsAndHashCode.Exclude
+    @Getter(lazy = true)
+    private final Path streamPath = dataPath.resolve(STREAMS);
+
+    @Positive
+    private long endBlockNumber = Long.MAX_VALUE;
```
xin-hedera: this may create issue depending on how a block node implements the spec for `end_block_number` in `SubscribeStreamRequest`.

<li>A block node implementation MAY reject a request for a block that is not yet available.</li>
<li>A block node implementation MAY accept future block numbers.</li>

Link: https://github.com/hiero-ledger/hiero-block-node/blob/ebcb0a32c86ad5f9d9a26f3bdd496b1b7e084e62/protobuf-sources/src/main/proto/block-node/api/block_stream_subscribe_service.proto#L28

yup the two items are conflict of each other, we can raise the concern, or revert back to using -1 for infinite block streaming.

steven-sheehy: Changed to use null for infinite

xin-hedera (approved):
LGTM

PR #11513, opened 2025-07-03
Update rest tests for non-zero shard/realm

### LINKED ISSUE #10744: Run unit and integration tests with non-zero realms

jnels124:
**Description**:
* replace hardcoded shard/realm in spec tests at runtime
* update integration tests to work with configured shard/realm

**Related issue(s)**:

Fixes #10744 

**Notes for reviewer**:

**Checklist**


rest/__tests__/accounts.deduplication.timestamp-notfound.test.js
```diff
@@ -23,18 +24,25 @@ describe('Accounts deduplicate timestamp not found tests', () => {
   const balanceTimestamp2 = tenDaysInToPreviousMonth + nanoSecondsPerSecond * 4n;
   const timestampRange1 = middleOfPreviousMonth + nanoSecondsPerSecond * 7n;
 
+  const entityId2 = EntityId.parseString('2');
```
steven-sheehy, line 27: Can use `EntityId.systemEntity.treasuryAccount` here and in other files with `entityId2`.

rest/viewmodel/networkNodeViewModel.js
```diff
@@ -25,7 +25,7 @@ class NetworkNodeViewModel {
     this.max_stake = utils.asNullIfDefault(nodeStake.maxStake, -1);
     this.memo = addressBookEntry.memo;
     this.min_stake = utils.asNullIfDefault(nodeStake.minStake, -1);
-    this.node_id = addressBookEntry.nodeId;
+    this.node_id = EntityId.parse(addressBookEntry.nodeId).num;
```
steven-sheehy: Revert this as a node_id is not an EntityId. It's a separate incrementing number maintained by consensus nodes without a shard and realm.

rest/utils.js
```diff
@@ -1362,6 +1362,8 @@ const formatComparator = (comparator) => {
         comparator.value = math.min(Number(comparator.value), responseLimit.max);
         break;
       case constants.filterKeys.NODE_ID:
+        comparator.value = EntityId.parseString(comparator.value).getEncodedId();
```
steven-sheehy: Revert

rest/__tests__/accounts.deduplication.timestamp-notfound.test.js
jnels124: a lot of these aren't really used as a treasury account so I left them. Can update though

rest/__tests__/accountAlias.test.js
```diff
@@ -18,50 +18,40 @@ describe('AccountAlias', () => {
...
             expect(AccountAlias.fromString(spec.input)).toEqual(spec.expected);
           } else {
-            expect(() => AccountAlias.fromString(spec.input)).toThrowErrorMatchingSnapshot();
+            try {
+              AccountAlias.fromString(spec.input);
+              throw new Error('Expected an error to be thrown');
+            } catch (error) {
+              expect(error).toBeInstanceOf(InvalidArgumentError);
+              expect(error.message).toMatch(/^Invalid accountAlias string \d+\.\d+\.AABBCC22$/);
+            }
```
xin-hedera, line 53: `toThrow` can't chain other matchers, however we can make it slightly better:

```suggestion
  const matcher =  expect(() => AccountAlias.fromString(spec.input));
  matcher.toThrow(InvalidArgumentError);
  matcher.toThrow(/^Invalid accountAlias string \d+\.\d+\.AABBCC22$/);
```

rest/__tests__/integrationUtils.js
```diff
@@ -6,17 +6,93 @@ import {exec} from 'child_process';
...
+    return value;
+  }
+  const transformedValue = transformShardRealmValues(value);
+  return EntityId.parseString(`${transformedValue}`).getEncodedId();
+};
+
 const applyResponseJsonMatrix = (spec, key) => {
   if (spec.responseJsonMatrix?.[key]) {
-    spec.responseJson = {
+    spec.responseJson = transformShardRealmValues({
```
xin-hedera, line 95: don't think this is needed. spec is already transformed before matrix functions are applied.

rest/__tests__/service/networkNodeService.test.js
```diff
@@ -11,6 +11,15 @@ import EntityId from '../../entityId';
 setupIntegrationTest();
 
 const defaultNodeFilter = 'abe.node_id = $2';
+
+const bigIntRange = (from, to) => {
```
xin-hedera, line 15: nit, can do

```javascript
  new Array(Number(to - from)).fill(from).map((v, i) => v + BigInt(i));
```

rest/__tests__/service/networkNodeService.test.js
```diff
@@ -133,49 +142,53 @@ describe('NetworkNodeService.getNetworkNodesWithFiltersQuery tests', () => {
...
-    file_id: 102,
+    file_id: EntityId.systemEntity.addressBookFile102.toString(),
     node_count: 4,
   },
 ];
 
+const nodeAccount3 = EntityId.parseString('3');
+const nodeAccount4 = EntityId.parseString('4');
+const nodeId0 = EntityId.parseString('0');
+const nodeId1 = EntityId.parseString('1');
```
xin-hedera: node id shouldn't depend on shard / realm.

rest/__tests__/service/networkNodeService.test.js
```diff
@@ -628,9 +651,9 @@ describe(`NetworkNodeService.getSupply`, () => {
     const accounts = [];
     const timestamp = utils.nowInNs();
     EntityId.systemEntity.unreleasedSupplyAccounts.forEach((range) => {
-      const from = range.from.getEncodedId();
-      const to = range.to.getEncodedId();
-      _.range(from, to + 1).forEach((id) => {
+      const from = range.from.num;
+      const to = range.to.num;
+      bigIntRange(BigInt(from), BigInt(to) + 1n).forEach((id) => {
```
xin-hedera: let's make `bigIntRange`'s to inclusive to avoid adding 1 everywhere.

rest/__tests__/integrationUtils.js
```diff
@@ -6,17 +6,93 @@ import {exec} from 'child_process';
...
+
+    if (/^0\.[A-Z0-9]{40,}$/i.test(obj)) {
+      return obj.replace(/^0\./, `${systemRealm}.`);
+    }
+
+    if (/^0\.\d+$/.test(obj)) {
+      return obj.replace(/^0\./, `${systemRealm}.`);
+    }
+
+    let possibleUrlParts = obj.split(/[?&]/);
```
xin-hedera: if all URLs start w/ `/api/v1` we can do explicit `obj.startsWith(apiPrefix)`

rest/__tests__/integrationDomainOps.js
```diff
@@ -529,6 +535,8 @@ const addAddressBook = async (addressBookInput) => {
       ? Buffer.from(addressBookInput.file_data, 'utf8')
       : Buffer.from(addressBook.file_data);
 
+  addressBook.file_id = encodedIdFromSpecValue(addressBook.file_id);
```
xin-hedera, line 538: if most functions at the end call `insertDomainObject` to insert one domain object to a table, should do the encoding in `insertDomainObject`? That is, for example, given such a map:

```javascript
const entityIdColumns = {
  crypto_allowance:  [
    'owner',
    'payer_account_id',
  ],
  ...
};
```

`insertDomainObject` converts all such columns with a non-null value.

xin-hedera (changes requested):

rest/__tests__/integrationUtils.js
jnels124: it is needed and fails any of the secondary tests

rest/__tests__/integrationDomainOps.js
jnels124: tried this initially but had some issues. will try again.

jnels124: if it is preferable can have a list of columns where we do the conversion but we still have other things like proto objects that need to be decoded changed and re-encoded. Not sure we want to do that all here in insert domain object or not.

rest/viewmodel/networkNodeViewModel.js
```diff
@@ -25,7 +25,7 @@ class NetworkNodeViewModel {
     this.max_stake = utils.asNullIfDefault(nodeStake.maxStake, -1);
     this.memo = addressBookEntry.memo;
     this.min_stake = utils.asNullIfDefault(nodeStake.minStake, -1);
-    this.node_id = addressBookEntry.nodeId;
+    this.node_id = this.node_id = addressBookEntry.nodeId;
```
steven-sheehy:
```suggestion
    this.node_id = addressBookEntry.nodeId;
```

steven-sheehy (approved):
LGTM

rest/__tests__/integrationDomainOps.js
xin-hedera: since that approach will reduce the number of changed lines and easier to read. however it's not a blocking issue.

xin-hedera (approved):
LGTM

PR #11535, opened 2025-07-08
Fix contract log indexes

### LINKED ISSUE #10944: Incorrect contract log index
### Description

contract result log index should be counted across a block, not a single transaction.

Official doc: https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_getfilterchanges

Some examples

- https://etherscan.io/tx/0xcf833ccd3f7ef7b4e39966b72a3d1a1fcdd6c0e339241ca03102ac3bacf47468#eventlog
- https://etherscan.io/tx/0x295425b3d5a1c7344a8fdde9e8004035897509d5bae953a88d588f0b7049e547#eventlog
- https://etherscan.io/tx/0x6a5dc82b3df0db293459bf13a6689d7709b75648a0662a99000c83974889ec69#eventlog
- https://etherscan.io/tx/0x19757f097b0c89024441184ed29b2246fb86fd7528ec9de7c38e2a626d1c9ee1#eventlog

Note these transactions are in the same block.

mirrornode rest api response for logs in a block:

https://mainnet.mirrornode.hedera.com/api/v1/contracts/results/logs?timestamp=gte:1744816320.039101950&timestamp=lte:1744816321.846722006&order=asc

### Steps to reproduce

check the description

### Additional context

_No response_

### Hedera network

mainnet

### Version

v0.127.0

### Operating system

None

bilyana-gospodinova:
**Description**:
This PR adds a DB migration that updates the persisted contract log indexes to increment on block level and not - on transaction level as it was before. This is needed for Ethereum compatibility. For the new contract logs that will be imported a change in the importer was made as well - the record file now keeps a log index between the different processed record items and they are persisted with the correct index directly.

Fixes #10944

**Checklist**


common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
```diff
@@ -143,6 +142,14 @@ public int getAndIncrementLogIndex() {
         return logIndex.getAndIncrement();
     }
 
+    public int getLogIndex() {
+        return logIndex.get();
+    }
+
+    public void setLogIndex(int index) {
+        logIndex.set(index);
+    }
```
IvanKavaldzhiev: Won't @ Getter - without PRIVATE and @ Setter lombok annotations be sufficient?

importer/src/main/java/org/hiero/mirror/importer/parser/record/RecordFileParser.java
```diff
@@ -158,10 +158,15 @@ protected void doParse(RecordFile recordFile) {
             aggregator.accept(recordItem);
 
             if (dateRangeFilter.filter(recordItem.getConsensusTimestamp())) {
+                recordItem.setLogIndex(recordFile.getLogIndex());
                 recordItemListener.onItem(recordItem);
                 recordMetrics(recordItem);
                 count.incrementAndGet();
             }
+            // The log index is incremented when the record item is processed, so
+            // set it directly here in the record file and order to preserve the index
```
IvanKavaldzhiev: Probably you meant - "in order to preserve"?

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
bilyana-gospodinova: They try to set a whole `AtomicInteger` as an object instead of the value of it.

importer/src/main/java/org/hiero/mirror/importer/parser/record/RecordFileParser.java
bilyana-gospodinova: Correct - fixed.

importer/src/main/java/org/hiero/mirror/importer/domain/ContractResultServiceImpl.java
```diff
@@ -289,7 +289,7 @@ private void processContractLogs(
             contractLog.setConsensusTimestamp(contractResult.getConsensusTimestamp());
             contractLog.setContractId(contractId);
             contractLog.setData(DomainUtils.toBytes(contractLoginfo.getData()));
-            contractLog.setIndex(index);
```
xin-hedera, line 292: since we no longer need `index`, can change the for loop to

```java
        for (var contractLogInfo : functionResult.getLogInfoList()) {
            ...   
        }
```

importer/src/test/java/org/hiero/mirror/importer/migration/ContractLogIndexesMigrationTest.java
```diff
@@ -0,0 +1,75 @@
...
+            recordFilePersist(index);
+        }
+        // Persist contract logs
+        // First block - contract log indexes are shuffled.
+        final var contractLogFirstRecordFile0 = contractLogPersist(2, 25);
+        final var contractLogFirstRecordFile1 = contractLogPersist(0, 55);
+        final var contractLogFirstRecordFile2 = contractLogPersist(1, 45);
+
+        // Second block - empty
+        // Third block - the indexes are all zeroes
```
IvanKavaldzhiev: Do we need explicitly 3 blocks of setup and skipping the second, while we can just have 2?

bilyana-gospodinova: I just wanted to add coverage for this corner case as well in order to make sure we would not run into some unexpected behaviour. In a live environment, there might be blocks that don't have a single contract log in them.

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordFile.java
```diff
@@ -119,6 +120,12 @@ public class RecordFile implements StreamFile<RecordItem> {
 
     private int version;
 
+    // The log index is used to keep track of the contract log count in each block.
+    @EqualsAndHashCode.Exclude
+    @JsonIgnore
+    @Transient
+    private final AtomicInteger logIndex = new AtomicInteger(0);
```
xin-hedera: think we can remove `logIndex` here and in `RecordItem`, and

- add it to `SqlEntityListener`
- in `SqlEntityListener.onContractLog`, set the index and increment the atomic integer
- in `SqlEntityListener.onEnd`, reset `logIndex` to 0

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexesMigration.java
```diff
@@ -0,0 +1,70 @@
...
+                                ) - 1) as new_calculated_index
+                            from
+                                contract_log cl_inner
+                            join
+                                record_file rf on cl_inner.consensus_timestamp >= rf.consensus_start
+                                              and cl_inner.consensus_timestamp <= rf.consensus_end
+                        ) as subquery
+                    where
+                        cl.consensus_timestamp = subquery.consensus_timestamp
+                        and cl.index = subquery.old_contract_log_index;
```
xin-hedera: This should be a versioned SQL migration, in `src/main/resources/db/migration/v1` and  `src/main/resources/db/migration/v2` respectively.

Do we know how long it takes to complete?

importer/src/test/java/org/hiero/mirror/importer/migration/ContractLogIndexesMigrationTest.java
```diff
@@ -0,0 +1,75 @@
...
+        assertThat(findIndex(contractLogThirdRecordFile2.getConsensusTimestamp()))
+                .isEqualTo(2);
+    }
+
+    private Integer findIndex(final long consensusTimestamp) {
+        var query = "select index from contract_log where consensus_timestamp = ?";
+        return jdbcOperations.queryForObject(query, Integer.class, consensusTimestamp);
+    }
+
+    private ContractLog contractLogPersist(final int contractLogIndex, final long consensusTimestamp) {
```
xin-hedera: should use `domainBuilder.timestamp()` for all consensus timestamps.

importer/src/test/java/org/hiero/mirror/importer/migration/ContractLogIndexesMigrationTest.java
```diff
@@ -0,0 +1,75 @@
...
+    @Test
+    void migrateSuccessful() {
+        // Given
+        // Persist record files
+        for (int index = 0; index < 3; index++) {
+            recordFilePersist(index);
+        }
+        // Persist contract logs
+        // First block - contract log indexes are shuffled.
+        final var contractLogFirstRecordFile0 = contractLogPersist(2, 25);
```
xin-hedera: we should have at least one instance of multiple logs in a single contract function result (with same consensus timestamp)

importer/src/test/java/org/hiero/mirror/importer/migration/ContractLogIndexesMigrationTest.java
```diff
@@ -0,0 +1,75 @@
...
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
+
+import lombok.RequiredArgsConstructor;
+import org.hiero.mirror.common.domain.contract.ContractLog;
+import org.hiero.mirror.importer.ImporterIntegrationTest;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+@RequiredArgsConstructor
+@Tag("migration")
```
xin-hedera: we should set a flyway migration target, with `@ContextConfiguration`, see https://github.com/hiero-ledger/hiero-mirror-node/blob/main/importer/src/test/java/org/hiero/mirror/importer/migration/FixAsyncJavaMigrationHistoryMigrationTest.java

xin-hedera (changes requested):

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordFile.java
steven-sheehy: We should not add state to `SqlEntityListener`. I think the `RecordFile` approach makes the most sense and we already have precedent with other transient fields on it.

xin-hedera: fine with that. however, we should still remove `logIndex` in RecordFile, and add a local variable in `RecordFileParser.doParse` to track it. The current approach will keep incrementing the index if a record file is reparsed (when transaction timeout happens).

steven-sheehy: If we add a local variable to RecordFileParser.doParse, how will it be made available to the ContractLog which is only seen in a few places like ContractResultServiceImpl?

xin-hedera: part of the logic is already in the changes:

- before calling EntityRecordItemListner.onItem, do `recordItem.setLogIndex(logIndex.get())`
- after that, do `logIndex.set(recordItem.getLogIndex())`

importer/src/main/java/org/hiero/mirror/importer/domain/ContractResultServiceImpl.java
bilyana-gospodinova: Correct, I updated the for-loop.

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexesMigration.java
bilyana-gospodinova: Changed the migration as suggested.

importer/src/test/java/org/hiero/mirror/importer/migration/ContractLogIndexesMigrationTest.java
bilyana-gospodinova: Changed to use the generated timestamp from the domain builder.

importer/src/test/java/org/hiero/mirror/importer/migration/ContractLogIndexesMigrationTest.java
bilyana-gospodinova: Good point. I extended the test to cover this case.

importer/src/test/java/org/hiero/mirror/importer/migration/ContractLogIndexesMigrationTest.java
bilyana-gospodinova: Addressed.

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordFile.java
bilyana-gospodinova: Thanks for the feedback. I didn't consider the case when a record file might get reparsed. I removed the log index from the record file and changed the implementation as per the last comment.

importer/src/test/java/org/hiero/mirror/importer/migration/ContractLogIndexesMigrationTest.java
```diff
@@ -0,0 +1,139 @@
...
+        var file = TestUtils.getResource("db/migration/" + migrationFilepath);
+        ownerJdbcTemplate.execute(FileUtils.readFileToString(file, StandardCharsets.UTF_8));
+    }
+
+    static class Initializer implements ApplicationContextInitializer<ConfigurableApplicationContext> {
+
+        @Override
+        public void initialize(ConfigurableApplicationContext configurableApplicationContext) {
+            var environment = configurableApplicationContext.getEnvironment();
+            String version = environment.acceptsProfiles(Profiles.of("v2")) ? "2.15.1" : "1.110.1";
```
xin-hedera: the target should be the version right before the one this class tests against. can address it after the other PR with migration merges or this PR gets 1.110.0 / 2.15.0 instead

bilyana-gospodinova: I see. I will update the versions on this line once the other PR with versions 1.110.0 / 2.15.0 gets merged.

importer/src/main/resources/db/migration/v1/V1.110.1__fix_contract_log_indexes.sql
```diff
@@ -0,0 +1,23 @@
+with calculated_indexes as (
```
xin-hedera: we'll have to optimize it, just the read-only part (the select) has taken more than 1 hour when I aborted it. May have to make it an async java migration.

kselveliev (dismissed):

importer/src/main/resources/db/migration/v1/V1.110.1__fix_contract_log_indexes.sql
bilyana-gospodinova: I made the migration of the async type utilizing the monthly partitioning as discussed.

importer/src/test/java/org/hiero/mirror/importer/migration/ContractLogIndexesMigrationTest.java
bilyana-gospodinova: This is no longer relevant as the migration is now async and there are no manually defined versions.

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexMigration.java
```diff
@@ -0,0 +1,151 @@
...
+                    order by cl.consensus_timestamp desc, cl.index desc
+            """;
+
+    private static final String UPDATE_CONTRACT_LOG_INDEX =
+            """
+                     update contract_log cl
+                     set "index" = :newIndex
+                     where cl.consensus_timestamp = :consensusTimestamp
+                         and cl.index = :index
+            """;
```
xin-hedera: we should do the select and update in one query, and also limit the timestamp with the same range in the update part. That is necessary for v2 as both `contract_log` and `record_file` tables are time partitioned.

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexMigration.java
```diff
@@ -0,0 +1,151 @@
...
+                importerProperties.getMigration(),
+                new NamedParameterJdbcTemplate(jdbcTemplate),
+                dbProperties.getSchema());
+        this.recordFileRepository = recordFileRepository;
+        this.transactionOperations = transactionOperations;
+    }
+
+    @Override
+    protected Long getInitial() {
+        return recordFileRepository
```
xin-hedera: there are quite some nuances in how our async java migrations work.

`getInitial` should provide data so that if the last run of the given async java migration doesn't complete (for instance importer was stopped), the new run should resume from where it left, instead of starting from the beginning.

The query here would definitely cause duplicate work in such scenario and we want to avoid it.

Normally, we can efficiently figure out the initial value from existing data in the tables. However, for contract log index, we'll need a temp table to store the consensus timestamp after which the migration has fixed the contract log index so far. The migration should  keep updating it in each `migratePartial`, and drop the table in `onSuccess`.

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexMigration.java
```diff
@@ -0,0 +1,151 @@
...
+    @Getter
+    private final TransactionOperations transactionOperations;
+
+    private static final RowMapper<RecordFile> RECORD_FILE_ROW_MAPPER = new DataClassRowMapper<>(RecordFile.class);
+
+    private static final String SELECT_RECORD_FILES =
+            """
+                    select *
+                    from record_file
+                    where consensus_end > :consensusEndLowerBound and consensus_end <= :consensusEndUpperBound
```
xin-hedera: this is inefficient and returns too much data

```sql
select
  (select consensus_start from record_file where consensus_end > :consensusEndLowerBound and consensus_end <= :consensusEndUpperBound order by consensus_end limit 1) as min_consensus_timestamp,
  (select consensus_start from record_file where consensus_end > :consensusEndLowerBound and consensus_end <= :consensusEndUpperBound order by consensus_end desc limit 1) as max_consensus_timestamp
```

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexMigration.java
```diff
@@ -0,0 +1,151 @@
...
+                    Map.of(
+                            "consensusTimestamp", consensusTimestamp,
+                            "index", index,
+                            "newIndex", newIndex));
+        }
+
+        final var nextConsensusEnd = lastConsensusEnd - INTERVAL;
+        if (nextConsensusEnd < 0) {
+            return Optional.of(0L);
+        }
```
xin-hedera: not needed. if `nextConsensusEnd` ever becomes negative, the next run of `migrationPartial` will not find any record files.

importer/src/main/java/org/hiero/mirror/importer/parser/record/RecordFileParser.java
```diff
@@ -158,10 +159,15 @@ protected void doParse(RecordFile recordFile) {
             aggregator.accept(recordItem);
 
             if (dateRangeFilter.filter(recordItem.getConsensusTimestamp())) {
+                recordItem.setLogIndex(logIndex.get());
```
steven-sheehy: Should the log index match the actual block contents or the mirror node's date range filtering of that block? If the former, need to move it out of the if statement.

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexMigration.java
bilyana-gospodinova: I am not sure we can combine the select and the update query. The reason for that is the `index` column that we are updating - it is part of the primary key in this table and in some cases a combined query leads to a duplicate key error. An example for that is one of the test cases - `ContractLogIndexMigration#migrateSuccessful`:
```
final var contractLogFirstRecordFile0 =
                contractLogPersist(2, recordFiles.get(0).getConsensusStart());
final var contractLogFirstRecordFile1 =
        contractLogPersist(0, recordFiles.get(0).getConsensusEnd());
final var contractLogFirstRecordFile2 =
        contractLogPersist(1, recordFiles.get(0).getConsensusEnd());
```
In this example `contractLogFirstRecordFile0` has a new index of 0 and it is updated successfully as the `consensus_timestamp` is the earliest in this block.
`contractLogFirstRecordFile1` needs to take index 1 and the log file is in the same transaction as `contractLogFirstRecordFile2` and they both have the same `consensus_timestamp` but `contractLogFirstRecordFile2` currently has index 1 and this leads to duplicate key error.

The combined query that I work with:
```
with calculated_indexes as (
      select
          cl.consensus_timestamp,
          cl.index as old_contract_log_index,
          (row_number() over (
           partition by rf.consensus_end
           order by cl.consensus_timestamp asc, cl.index asc
      ) - 1) as new_calculated_index
      from contract_log cl
      join record_file rf on cl.consensus_timestamp >= rf.consensus_start
                        and cl.consensus_timestamp <= rf.consensus_end
      where cl.consensus_timestamp >= :consensusStart and cl.consensus_timestamp <= :lastConsensusEnd
      order by cl.consensus_timestamp desc, cl.index desc
  )
  update contract_log cl
  set "index" = ci.new_calculated_index
  from calculated_indexes ci
  where cl.consensus_timestamp = ci.consensus_timestamp
      and cl.index = ci.old_contract_log_index;
```
Perhaps I am missing something. Do you see how we can mitigate this?

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexMigration.java
bilyana-gospodinova: Thanks, this was a helpful clarification! I added a temporary table and adapted `getInitial` and `migratePartial`. I used `FixCryptoAllowanceAmountMigration` as a reference.

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexMigration.java
xin-hedera: did you hit the issue when running the query? I don't see why we will hit the uniqueness constraint of primary key, even through we are changing the `index`'s scope to be within the same record file, `(consensus_timestamp, index)` are still unique. Unless there's bug in the SQL itself, given when we have a row with `(t1, i1)` and we try to update a different row `(t2,i2)` to be also `(t1,i1)`.

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexMigration.java
bilyana-gospodinova: I agree your way is much better. Addressed.

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexMigration.java
bilyana-gospodinova: Removed.

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexMigration.java
xin-hedera: saw the issue now. we can make the SQL work still, the key is sort the `caculated_indexes` by consensus_timestamp desc and old_index desc, that way, we won't hit the duplicate key issue.

assume these are the consensus timestamp and index in contract_log table: (1, 0), (2, 0), (2, 1), (2,2)

and `cl_tmp` table has the correct indexes.

```sql
create table cl_tmp (consensus_timestamp bigint, index integer, old_index integer);
insert into cl_tmp (consensus_timestamp, index, old_index) values (1, 0, 0), (2,1,0), (2,2,1), (2,3,2);
```

this will fail

```sql
update contract_log as cl
set index = t.index
from cl_tmp as t
where cl.consensus_timestamp = t.consensus_timestamp and cl.index = t.old_index;
```

this will succeed

```sql
with patch as (
  select *
  from cl_tmp
  order by consensus_timestamp desc, old_index desc
)
update contract_log as cl
set index = t.index
from patch as t
where cl.consensus_timestamp = t.consensus_timestamp and cl.index = t.old_index
```

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexMigration.java
```diff
@@ -0,0 +1,211 @@
...
+        return "Recalculate contract log indexes on block level.";
+    }
+
+    @VisibleForTesting
+    Optional<RecordFileSlice> getRecordFilesMinAndMaxTimestamp(final Long lastConsensusEnd) {
+        var params = Map.of(
+                "consensusEndUpperBound", lastConsensusEnd, "consensusEndLowerBound", lastConsensusEnd - INTERVAL);
+        final var minTimestamp = new AtomicLong();
+        final var maxTimestamp = new AtomicLong();
+        namedParameterJdbcTemplate.query(SELECT_RECORD_FILES_MIN_AND_MAX_TIMESTAMP, params, rs -> {
```
xin-hedera: use `Optional.ofNullable(queryForObjectOrNull(SELECT_RECORD_FILES_MIN_AND_MAX_TIMESTAMP, params, RecordFileSlice.class)`.

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexMigration.java
xin-hedera: unfortunately that doesn't work in citus. let's first avoid unnecessary update by excluding logs whose index doesn't have to change:

```sql
with rf as (
  select consensus_start, consensus_end
  from record_file
  where consensus_end >= :consensusStart and consensus_end <= :lastConsensusEnd
  order by consensus_end
), updated_index as (
  select
    cl.contract_id,
    cl.consensus_timestamp,
    cl.index as old_index,
    (row_number() over (
        partition by rf.consensus_end
        order by cl.consensus_timestamp asc, cl.index asc
    ) - 1) as new_index
  from contract_log as cl
  join rf on cl.consensus_timestamp >= rf.consensus_start
        and cl.consensus_timestamp <= rf.consensus_end
  where cl.consensus_timestamp >= :consensusStart and cl.consensus_timestamp <= :lastConsensusEnd
  order by cl.consensus_timestamp desc, cl.index desc
)
select *
from updated_index
where old_index <> new_index
```

Note:

- the `rf` CTE, it improves performance in citus
- the returned data also include `contract_id`, that may make the update in citus faster since `contract_log` is distributed on `contract_id` column

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexMigration.java
```diff
@@ -0,0 +1,211 @@
...
+
+    // 30-day intervals of data to take advantage of the monthly partitioning of record_file and contract_log tables.
+    static final long INTERVAL = Duration.ofDays(30).toNanos();
+
+    @Getter
+    private final TransactionOperations transactionOperations;
+
+    private final JdbcTemplate jdbcTemplate;
+
+    private static final String CREATE_TEMPORARY_RECORD_FILE_TABLE =
```
xin-hedera: a cleaner way is

- override `performSynchronousSteps()` and return `false` if `record_file` table is empty
- in `getInitial()`
  - first do 
    ```sql
    create table if not exists  processed_record_file_temp (consensus_end bigint not null);
    ```
  - then return
    ```sql
     select coalesce(
       (select consensus_end from processed_record_file_temp limit 1),
       (select consensus_end from record_file order by consensus_end desc limit 1)
    );
     ```

Then we don't have to do `isMigrationTableExists` check.

It might be better to not spread the logic over to `performSynchronousSteps()`, then we can change the select query with coalesce to

```sql
     select coalesce(
       (select consensus_end from processed_record_file_temp limit 1),
       (select consensus_end from record_file order by consensus_end desc limit 1),
       0
    );
```

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexMigration.java
```diff
@@ -0,0 +1,211 @@
...
+        }
+
+        return getCurrentRecordFileEndTimestamp();
+    }
+
+    @Nonnull
+    @Override
+    protected Optional<Long> migratePartial(Long lastConsensusEnd) {
+        // Get record files for an interval of time.
+        final var recordFileSlice = getRecordFilesMinAndMaxTimestamp(lastConsensusEnd);
```
xin-hedera: think we no longer need the function `getRecordFilesMinAndMaxTimestamp` once changing to call 

`queryForObjectOrNull(SELECT_RECORD_FILES_MIN_AND_MAX_TIMESTAMP, params, RecordFileSlice.class)`

the other benefit is we only need to calculate `nextConsensusEnd = lastConsensusEnd - INTERVAL` once.

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexMigration.java
bilyana-gospodinova: I can't use this directly as `queryForObject()` expects the result as a single column and in this case there are 2 columns returned and I get an exception. I split the query into two separate statements and used the suggested method and I think it looks cleaner this way.

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexMigration.java
bilyana-gospodinova: `getRecordFilesMinAndMaxTimestamp()` has still some logic in it and I prefer to keep it for now.

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexMigration.java
bilyana-gospodinova: Thanks. Comment addressed.

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexMigration.java
bilyana-gospodinova: I addressed the last comment. However, I could not really understand what exactly did not work on Citus.

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexMigration.java
xin-hedera: oh my bad.

we can do this, instead of the complicated logic in the function

- add this function to the parent class

```java
    public <O> O queryForObjectOrNull(String sql, SqlParameterSource paramSource, RowMapper<O> rowMapper) {
        try {
            return namedParameterJdbcTemplate.queryForObject(sql, paramSource, rowMapper);
        } catch (EmptyResultDataAccessException ex) {
            return null;
        }
    }
```

- create static final RowMapper for the data class, e.g, `new DataClassRowMapper<>(RecordFileSlice.class)` and use it with the new function

bilyana-gospodinova: I have already tried this. However, in case there is no such entry, an exception is thrown when trying to cast `null` to `long` inside `queryForObject(...)`. To mitigate this I need to catch type mismatch exception and I thought this is too error-prone and removed the code.

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexMigration.java
xin-hedera: in postgres, when updating a table using data from another relation / table / subquery, the order of the rows updated should follow the order of the rows in the from item. that's why I proposed to use the desc order so given a consensus timestamp t1, to update the following tuples

`(t1, 0), (t1, 1), (t1, 2)`

to

`(t1, 1), (t1, 2), (t1, 3)`

the ascending order will have duplicate key violation when trying to update `(t1, 0)` to `(t1, 1)` and on the opposite, the descending order won't have such issue because every time for the same `t1`, the new index is a value not in any existing rows.

However, that doesn't work in citus, it's likely when citus runs the distributed SQL it just doesn't honor or preserve the order.

bilyana-gospodinova: Oh, I see. Good to know, thanks.

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexMigration.java
xin-hedera: we can change the query to find the timestamp range to

```sql
select consensus_start as min_consensus_timestamp, max_consensus_timestamp
from record_file
join (
  select i.consensus_end
  from record_file as i
  where i.consensus_end > :consensusEndLowerBound and i.consensus_end <= :consensusEndUpperBound
  order by i.consensus_end desc
  limit 1
) as t(max_consensus_timestamp) on true
where consensus_end > :consensusEndLowerBound and consensus_end <= :consensusEndUpperBound
order by consensus_end
limit 1;
```

The query will return no data if there's no matching record file in the range `(lowerBounder, upperBound]` so we don't have to deal with `null` value.

Pair the query with the proposed `public <O> O queryForObjectOrNull(String sql, SqlParameterSource paramSource, RowMapper<O> rowMapper) `, you should be able to remove the function `getRecordFilesMinAndMaxTimestamp()`

bilyana-gospodinova:
image
Tested different slice intervals against a random timestamp and real data on Postgre DB. It seems that the smaller the interval the better the performance. This can be seen in the last row which calculates an approximate ratio between the processed rows and the time needed to process them. The query is the select statement that recalculates the indexes (`SELECT_CONTRACT_LOG_INDEXES`). I have excluded the last where clause ("where old_contract_log_index <> new_calculated_index") in order to have a clear view of the actually processed rows (by the partitioning, the join and the order clauses) as this "where" clause would "hide" some of the rows that have been processed but not returned due to the where clause.
Considering the numbers I think we should decrease the interval from 30 days to 1 hour.

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexMigration.java
bilyana-gospodinova: Removed `getRecordFilesMinAndMaxTimestamp()`.

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexMigration.java
bilyana-gospodinova: Comment addressed.

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexMigration.java
```diff
@@ -0,0 +1,189 @@
...
+            """
+                    create table if not exists processed_record_file_temp(
+                        consensus_end bigint not null
+                    );
+            """;
+
+    private static final String SELECT_LAST_CONSENSUS_END_TIMESTAMP =
+            """
+                    select coalesce(
+                       (select consensus_end from processed_record_file_temp limit 1),
```
xin-hedera: since for each iteration, the SQL to save the new `consensus_end` has to be an `insert`, an explicit ascending order is needed here


```suggestion
                       (select consensus_end from processed_record_file_temp order by consensus_end limit 1),
```

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexMigration.java
```diff
@@ -0,0 +1,189 @@
...
+                       (select consensus_end from record_file order by consensus_end desc limit 1),
+                       0
+                    );
+            """;
+
+    private static final String UPDATE_TEMPORARY_RECORD_FILE_TABLE =
```
xin-hedera: change this to `insert into processed_record_file_temp :lastProcessedRecordFileEndTimestamp;`

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexMigration.java
```diff
@@ -0,0 +1,189 @@
...
+                      where rf.consensus_end > :consensusEndLowerBound and rf.consensus_end <= :consensusEndUpperBound
+                      order by rf.consensus_end desc
+                      limit 1
+                    ) as t(max_consensus_timestamp) on true
+                    where consensus_end > :consensusEndLowerBound and consensus_end <= :consensusEndUpperBound
+                    order by consensus_end
+                    limit 1;
+            """;
+
+    private static final String SELECT_CONTRACT_LOG_INDEXES =
```
xin-hedera: do select and update all in SQL and in a transaction to reduce the number of roundtrips between db server and application:

```sql
            begin;

            set citus.max_intermediate_result_size = -1;
            set local temp_buffers = '64MB';
            create temp table if not exists contract_log_migration(like contract_log);

            with rf as (
                select consensus_start, consensus_end
                from record_file
                where consensus_end >= :consensusStart and consensus_end <= :lastConsensusEnd
                order by consensus_end
            ), updated_index as (
                select
                    cl.contract_id,
                    cl.consensus_timestamp,
                    cl.index as old_index,
                    (row_number() over (
                     partition by rf.consensus_end
                     order by cl.consensus_timestamp, cl.index
                ) - 1) as new_index
                from contract_log cl
                join rf on cl.consensus_timestamp >= rf.consensus_start
                                  and cl.consensus_timestamp <= rf.consensus_end
                where cl.consensus_timestamp >= :consensusStart and cl.consensus_timestamp <= :lastConsensusEnd
                order by cl.consensus_timestamp, cl.index
            ), changed as (
              select *
              from updated_index
              where old_index <> new_index
            ), changed_stat as (
              select consensus_timestamp, max(old_index) as max_old, min(new_index) as min_new
              from changed
              group by consensus_timestamp
            ), non_conflict as (
              select *
              from changed
              where consensus_timestamp in (select consensus_timestamp from changed_stat where max_old < min_new)
            ), fix_non_conflict as (
              update contract_log as cl
              set index = new_index
              from non_conflict as n
              where cl.consensus_timestamp = n.consensus_timestamp and cl.index = n.old_index
                and cl.consensus_timestamp >= :consensusStart and cl.consensus_timestamp <= :lastConsensusEnd
            ), conflict as (
              select *
              from changed
              where consensus_timestamp in (select consensus_timestamp from changed_stat where max_old >= min_new)
            ), deleted as (
              delete from contract_log as cl
              using conflict as c
              where cl.consensus_timestamp = c.consensus_timestamp and cl.index = c.old_index
                and cl.consensus_timestamp >= :consensusStart and cl.consensus_timestamp <= :lastConsensusEnd
              returning cl.*, c.new_index
            )
            insert into contract_log_migration (
              bloom,
              consensus_timestamp,
              contract_id,
              data,
              index,
              payer_account_id,
              root_contract_id,
              topic0,
              topic1,
              topic2,
              topic3,
              transaction_hash,
              transaction_index
            )
            select
              bloom,
              consensus_timestamp,
              contract_id,
              data,
              new_index,
              payer_account_id,
              root_contract_id,
              topic0,
              topic1,
              topic2,
              topic3,
              transaction_hash,
              transaction_index
            from deleted;

            insert into contract_log
            select * from contract_log_migration;

            commit;
```

The query does the following:

- calculate correct contract log indexex given the timestamp range
- determine which contract log row's index needs to fix: the `changed` CTE
- determine which rows can change in place without violating primary key constraint: given a consensus timestamp, if the largest old index is smaller than the smallest new index, it's safe to update in place
- update those without conflict in place
- determine which rows (grouped by consensus timestamp) have overlapping old indexes and new indexes, delete them from `contract_log`, and insert the rows with correct index into the temp table `contract_log_migration`
- in a separate query, insert all rows from `contract_log_migration` into `contract_log`

The last step has to be a separate query because if it's done in another CTE or the main query with the CTE deletes the rows, the changes of the deletion aren't visible to the `insert` and it'll still fail with duplicate primary key violation.

There is a citus specific setting `set citus.max_intermediate_result_size = -1;`, so the clean way is make this a template and only add the setting for v2.

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexMigration.java
```diff
@@ -0,0 +1,189 @@
...
+        return endTimestamp;
+    }
+
+    @Nonnull
+    @Override
+    protected Optional<Long> migratePartial(Long lastConsensusEnd) {
+        // Get record files for an interval of time.
+        var recordFileSliceParams = new MapSqlParameterSource()
+                .addValue("consensusEndUpperBound", lastConsensusEnd)
+                .addValue("consensusEndLowerBound", lastConsensusEnd - INTERVAL);
```
xin-hedera: `lastConsensusEnd - INTERVAL` appeared twice, so make it a variable.

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexMigration.java
```diff
@@ -0,0 +1,189 @@
...
+        final var nextConsensusEnd = lastConsensusEnd - INTERVAL;
+        return Optional.of(nextConsensusEnd);
+    }
+
+    @Override
+    public String getDescription() {
+        return "Recalculate contract log indexes on block level.";
+    }
+
+    record RecordFileSlice(long minConsensusTimestamp, long maxConsensusTimestamp) {}
```
xin-hedera: Codacy suggests to make it `static`

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexMigration.java
```diff
@@ -0,0 +1,189 @@
...
+
+        final var nextConsensusEnd = lastConsensusEnd - INTERVAL;
+        return Optional.of(nextConsensusEnd);
+    }
+
+    @Override
+    public String getDescription() {
+        return "Recalculate contract log indexes on block level.";
+    }
+
```
xin-hedera, line 257: let's implement `performSynchronousSteps()` and return `false` to skip the migration when `PersistProperties.isContracts()` is false or `PersistProperties.isContractResults()` is false.

importer/src/test/java/org/hiero/mirror/importer/migration/ContractLogIndexMigrationTest.java
```diff
@@ -0,0 +1,151 @@
...
+import org.hiero.mirror.common.domain.contract.ContractLog;
+import org.hiero.mirror.common.domain.transaction.RecordFile;
+import org.hiero.mirror.importer.EnabledIfV1;
+import org.hiero.mirror.importer.repository.ContractLogRepository;
+import org.hiero.mirror.importer.repository.RecordFileRepository;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+@EnabledIfV1
```
xin-hedera: this should run for both v1 and v2
```suggestion
```

importer/src/test/java/org/hiero/mirror/importer/migration/ContractLogIndexMigrationTest.java
```diff
@@ -0,0 +1,151 @@
...
+
+    @Getter
+    private final ContractLogIndexMigration migration;
+
+    private final RecordFileRepository recordFileRepository;
+    private final ContractLogRepository contractLogRepository;
+
+    @AfterEach
+    void temporaryTableCleanedUp() {
+        assertThat(tableExists("processed_record_file_temp")).isFalse();
```
xin-hedera: the assertion should be part of each test, it's weird to do assertion in `@AfterEach` hook.

importer/src/test/java/org/hiero/mirror/importer/migration/ContractLogIndexMigrationTest.java
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    @Test
+    void migrationOnEmptyDB() {
+        // given, when
+        migration.migrateAsync();
```
xin-hedera: please merge from `main`, then call `runMigration()` from parent class.

```suggestion
        runMigration();
```

also need to add `waitForCompletion()` before assertions.

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexMigration.java
```diff
@@ -0,0 +1,189 @@
...
+@Named
+public class ContractLogIndexMigration extends AsyncJavaMigration<Long> {
+
+    @Getter
+    private final TransactionOperations transactionOperations;
+
+    private final JdbcTemplate jdbcTemplate;
+
+    static final long INTERVAL = Duration.ofHours(1).toNanos();
```
xin-hedera: hourly iteration is too excessive, tested with 7 days, it took 55 minutes. I can test with smaller intervals, however should be at least 1 day imo.

xin-hedera: can also make it configuration and default to 7 days. example: https://github.com/hiero-ledger/hiero-mirror-node/blob/5142c7610b4dc70299d0e192ae724e7cf59e8e7b/importer/src/main/java/org/hiero/mirror/importer/migration/BackfillAndDeduplicateBalanceMigration.java#L164

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexMigration.java
bilyana-gospodinova: It does but I get a warning in the IDE that the modifier is redundant. Also, in other similar classes the `static` modifier is omitted as well, so I guess it would be better not to add it.

importer/src/test/java/org/hiero/mirror/importer/migration/ContractLogIndexMigrationTest.java
bilyana-gospodinova: Removed

importer/src/test/java/org/hiero/mirror/importer/migration/ContractLogIndexMigrationTest.java
bilyana-gospodinova: Moved into tests.

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexMigration.java
bilyana-gospodinova: Implemented.

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexMigration.java
bilyana-gospodinova: Added into a local variable.

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexMigration.java
bilyana-gospodinova: Addressed.

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexMigration.java
bilyana-gospodinova: Changed.

importer/src/test/java/org/hiero/mirror/importer/migration/ContractLogIndexMigrationTest.java
bilyana-gospodinova: Done, thanks.

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexMigration.java
bilyana-gospodinova: Thanks, @xin-hedera! This looks very elegant! I made the query into a template and I now add the citus property only for v2. Added also a temporary table cleanup on commit.

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexMigration.java
```diff
@@ -0,0 +1,261 @@
...
+
+    private static final RowMapper<RecordFileSlice> ROW_MAPPER = new DataClassRowMapper<>(RecordFileSlice.class);
+
+    @Lazy
+    protected ContractLogIndexMigration(
+            final Environment environment,
+            final DBProperties dbProperties,
+            final ImporterProperties importerProperties,
+            final @Owner JdbcTemplate jdbcTemplate,
+            final TransactionOperations transactionOperations,
```
xin-hedera: should avoid potential issue that users are mixed in the same db transaction. note

- `transactionOperations` is used by parent class to run `migratePartial()` by child class 

```java
getTransactionOperations().execute(t -> migratePartial(previous.get()))
```

- inside `migratePartial`, `jdbcTemplate` is used to run SQL queries
- in this migration, `transactionOperations` bean uses `mirror_importer`, and the `@Owner jdbcTemplate` uses `mirror_node`

Please implement `getTransactionOperations()` the same way in `BackfillAndDeduplicateBalanceMigration.java`

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexMigration.java
bilyana-gospodinova: I set the interval to 7 days for now.

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexMigration.java
```diff
@@ -0,0 +1,261 @@
...
+                sliceStartTimestamp,
+                sliceEndTimestamp);
+
+        final var params = Map.of(
+                "lastConsensusEnd", sliceEndTimestamp,
+                "consensusStart", sliceStartTimestamp);
+        namedParameterJdbcTemplate.update(getVersionedContractUpdateQuery(), params);
+
+        namedParameterJdbcTemplate.update(
+                INSERT_LAST_PROCESSED_TIMESTAMP, Map.of("lastProcessedRecordFileEndTimestamp", sliceEndTimestamp));
```
xin-hedera: nit, we can move the query into `UPDATE_CONTRACT_LOG_INDEXES`

bilyana-gospodinova: Moved.

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexMigration.java
bilyana-gospodinova: Comment addressed.

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexMigration.java
```diff
@@ -0,0 +1,261 @@
...
+                        topic3,
+                        transaction_hash,
+                        transaction_index
+                    from deleted;
+
+                    insert into contract_log
+                    select * from contract_log_migration;
+
+                    insert into processed_record_file_temp(consensus_end)
+                    values(:lastConsensusEnd);
```
xin-hedera:
```suggestion
                    values(:consensusStart);
```

bilyana-gospodinova: Nice catch - addressed.

xin-hedera (dismissed):
LGTM

importer/src/main/java/org/hiero/mirror/importer/migration/AsyncJavaMigration.java
```diff
@@ -148,6 +149,14 @@ public <O> O queryForObjectOrNull(String sql, SqlParameterSource paramSource, Cl
         }
     }
 
+    public <O> O queryForObjectOrNull(String sql, SqlParameterSource paramSource, RowMapper<O> rowMapper) {
```
steven-sheehy: Can also change the other similar method.

```suggestion
    protected final <O> O queryForObjectOrNull(String sql, SqlParameterSource paramSource, RowMapper<O> rowMapper) {
```

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexMigration.java
```diff
@@ -0,0 +1,264 @@
...
+import org.springframework.jdbc.core.JdbcTemplate;
+import org.springframework.jdbc.core.RowMapper;
+import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
+import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
+import org.springframework.jdbc.datasource.DataSourceTransactionManager;
+import org.springframework.transaction.support.TransactionOperations;
+import org.springframework.transaction.support.TransactionTemplate;
+
+@Named
+public class ContractLogIndexMigration extends AsyncJavaMigration<Long> {
```
steven-sheehy:
```suggestion
final class ContractLogIndexMigration extends AsyncJavaMigration<Long> {
```

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
```diff
@@ -73,7 +73,6 @@ public class RecordItem implements StreamItem {
     private final int transactionType;
 
     @EqualsAndHashCode.Exclude
-    @Getter(PRIVATE)
     private final AtomicInteger logIndex = new AtomicInteger(0);
```
steven-sheehy: It's wasteful to create a separate AtomicInteger for each item when a single one shared across all items would suffice. It's also less brittle. Change this to be non-final, remove `getLogIndex()` and change `setLogIndex()` to accept an `AtomicInteger`.

importer/src/main/java/org/hiero/mirror/importer/parser/record/RecordFileParser.java
```diff
@@ -158,10 +159,15 @@ protected void doParse(RecordFile recordFile) {
...
                 recordItemListener.onItem(recordItem);
                 recordMetrics(recordItem);
                 count.incrementAndGet();
             }
+            // The log index is incremented when the record item is processed, so
+            // set it directly here in the record file in order to preserve the index
+            // and pass it to the next record item.
+            logIndex.set(recordItem.getLogIndex());
```
steven-sheehy: Can remove this with suggestion to have atomic integer setter.

steven-sheehy (changes requested):

importer/src/main/java/org/hiero/mirror/importer/parser/record/RecordFileParser.java
bilyana-gospodinova: I think the former makes more sense. However, in this case the log index will be used only inside `recordItemListener.onItem(recordItem);` which gets called and incremented only inside the data range filtering if-statement above, so I guess in this case it does not really matter where we put this line.

However, when you pointed the data filtering out I started to think about the following corner case: what if we have a record file with start and end timestamps [0; 1000] and the start date in the importer properties is 500 then only the record items in range [500; 1000] will be processed and only their indexes will be calculated and they will be wrong with the current implementation. The issue is that we are not going to be able to calculate the starting index at timestamp 500 without parsing the whole record file. Is this case possible? Do we currently parse partially some record items from a record file?

importer/src/main/java/org/hiero/mirror/importer/migration/AsyncJavaMigration.java
bilyana-gospodinova: Changed

importer/src/main/java/org/hiero/mirror/importer/migration/ContractLogIndexMigration.java
bilyana-gospodinova: Addressed

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
bilyana-gospodinova: Addressed

importer/src/main/java/org/hiero/mirror/importer/parser/record/RecordFileParser.java
bilyana-gospodinova: Removed

xin-hedera (approved):
LGTM

PR #11563, opened 2025-07-11
Fix bulk load storage slots

### LINKED ISSUE #11551: Bulk slot loading returns inconsistent response
### Description

Customers have complained that the new bulk slot loading and caching mechanism added in 0.133 causes results to the same contract with different data to return inconsistent responses.

### Steps to reproduce

1. Enable bulk slot loading
2. Run the following 4 calls in sequence
```json
{
    "to": "0x00000000000000000000000000000000003c4196",
    "data": "0x351fb47800000000000000000000000027784a45fdee4d8f355237a4e92f259c1d7d7617fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffe"
}
{
    "data": "0x351fb47800000000000000000000000027784a45fdee4d8f355237a4e92f259c1d7d7617ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff",
    "to": "0x00000000000000000000000000000000003c4196"
}
{
    "data": "0x351fb47800000000000000000000000027784a45fdee4d8f355237a4e92f259c1d7d76170000000000000000000000000000000000000000000000000000000000000000",
    "to": "0x00000000000000000000000000000000003c4196"
}
{
    "data": "0x351fb47800000000000000000000000027784a45fdee4d8f355237a4e92f259c1d7d76170000000000000000000000000000000000000000000000000000000000000001",
    "to": "0x00000000000000000000000000000000003c4196"
}
```
Only the middle two (-1 and 0) should return data, the others will return a 0 array.

### Additional context

_No response_

### Hedera network

mainnet

### Version

v0.133.0

### Operating system

None

kselveliev:
**Description**:
Fix bulk load storage slots.

Added a couple of integration tests that test concurrent calls to cover specific corner cases and also eviction corner case.

Fixes #11551

**Checklist**

### DIFF
--- a/web3/src/main/java/org/hiero/mirror/web3/service/ContractStateServiceImpl.java
+++ b/web3/src/main/java/org/hiero/mirror/web3/service/ContractStateServiceImpl.java
@@ -86,14 +86,15 @@ final class ContractStateServiceImpl implements ContractStateService {
         final var contractSlotsCache = ((CaffeineCache) this.contractSlotsCache.get(
                 contractId, () -> cacheManagerSlotsPerContract.getCache(contractId.toString())));
         // Cached slot keys for contract, whose slot values are not present in the contractStateCache
-        final var cachedSlots = new ArrayList<byte[]>();
         contractSlotsCache.putIfAbsent(ByteBuffer.wrap(key), EMPTY_VALUE);
-        (contractSlotsCache).getNativeCache().asMap().keySet().forEach(slot -> {
-            final var slotBytes = ((ByteBuffer) slot).array();
-            cachedSlots.add(slotBytes);
-        });
+        final var cachedSlots = contractSlotsCache.getNativeCache().asMap().keySet();
 
-        final var contractSlotValues = contractStateRepository.findStorageBatch(contractId.getId(), cachedSlots);
+        final var contractSlotValues = contractStateRepository.findStorageBatch(
+                contractId.getId(),
+                cachedSlots.stream()
+                        .map(slot -> ((ByteBuffer) slot).array())
+                        .toList()
+        );
         boolean isKeyEvictedFromCache =
                 !cachedSlots.contains(ByteBuffer.wrap(key).array());
         byte[] cachedValue = null;
--- a/web3/src/main/java/org/hiero/mirror/web3/service/ContractStateServiceImpl.java
+++ b/web3/src/main/java/org/hiero/mirror/web3/service/ContractStateServiceImpl.java
@@ -83,25 +83,16 @@ final class ContractStateServiceImpl implements ContractStateService {
      * @return slotKey-value pairs for contractId
      */
     private Optional<byte[]> findStorageBatch(final EntityId contractId, final byte[] key) {
-        final var contractSlotsCache = this.contractSlotsCache.get(
-                contractId, () -> cacheManagerSlotsPerContract.getCache(contractId.toString()));
+        final var contractSlotsCache = ((CaffeineCache) this.contractSlotsCache.get(
+                contractId, () -> cacheManagerSlotsPerContract.getCache(contractId.toString())));
         // Cached slot keys for contract, whose slot values are not present in the contractStateCache
         final var cachedSlots = new ArrayList<byte[]>();
         synchronized (contractSlotsCache) {
             contractSlotsCache.putIfAbsent(ByteBuffer.wrap(key), EMPTY_VALUE);
-            ((CaffeineCache) contractSlotsCache)
-                    .getNativeCache()
-                    .asMap()
-                    .keySet()
-                    .forEach(slot -> {
-                        final var slotBytes = ((ByteBuffer) slot).array();
-                        final var value =
-                                contractStateCache.putIfAbsent(generateCacheKey(contractId, slotBytes), EMPTY_VALUE);
-
-                        if (value == null || value.get().equals(EMPTY_VALUE)) {
-                            cachedSlots.add(slotBytes);
-                        }
-                    });
+            (contractSlotsCache).getNativeCache().asMap().keySet().forEach(slot -> {
+                final var slotBytes = ((ByteBuffer) slot).array();
+                cachedSlots.add(slotBytes);
+            });
         }
 
         final var contractSlotValues = contractStateRepository.findStorageBatch(contractId.getId(), cachedSlots);
@@ -111,7 +102,6 @@ final class ContractStateServiceImpl implements ContractStateService {
             final byte[] slotKey = contractSlotValue.getSlot();
             final byte[] slotValue = contractSlotValue.getValue();
             contractStateCache.put(generateCacheKey(contractId, slotKey), slotValue);
-            contractSlotsCache.put(ByteBuffer.wrap(slotKey), EMPTY_VALUE);
 
             if (Arrays.equals(slotKey, key)) {
                 cachedValue = slotValue;
--- a/web3/src/main/java/org/hiero/mirror/web3/service/ContractStateServiceImpl.java
+++ b/web3/src/main/java/org/hiero/mirror/web3/service/ContractStateServiceImpl.java
@@ -85,18 +85,24 @@ final class ContractStateServiceImpl implements ContractStateService {
     private Optional<byte[]> findStorageBatch(final EntityId contractId, final byte[] key) {
         final var contractSlotsCache = this.contractSlotsCache.get(
                 contractId, () -> cacheManagerSlotsPerContract.getCache(contractId.toString()));
-        contractSlotsCache.putIfAbsent(ByteBuffer.wrap(key), EMPTY_VALUE);
-
         // Cached slot keys for contract, whose slot values are not present in the contractStateCache
         final var cachedSlots = new ArrayList<byte[]>();
-        ((CaffeineCache) contractSlotsCache).getNativeCache().asMap().keySet().forEach(slot -> {
-            final var slotBytes = ((ByteBuffer) slot).array();
-            final var value = contractStateCache.putIfAbsent(generateCacheKey(contractId, slotBytes), EMPTY_VALUE);
+        synchronized (contractSlotsCache) {
+            contractSlotsCache.putIfAbsent(ByteBuffer.wrap(key), EMPTY_VALUE);
+            ((CaffeineCache) contractSlotsCache)
+                    .getNativeCache()
+                    .asMap()
+                    .keySet()
+                    .forEach(slot -> {
+                        final var slotBytes = ((ByteBuffer) slot).array();
+                        final var value =
+                                contractStateCache.putIfAbsent(generateCacheKey(contractId, slotBytes), EMPTY_VALUE);
 
-            if (value == null) {
-                cachedSlots.add(slotBytes);
-            }
-        });
+                        if (value == null || value.get().equals(EMPTY_VALUE)) {
+                            cachedSlots.add(slotBytes);
+                        }
+                    });
+        }
 
         final var contractSlotValues = contractStateRepository.findStorageBatch(contractId.getId(), cachedSlots);
         byte[] cachedValue = null;
--- a/web3/src/main/java/org/hiero/mirror/web3/service/ContractStateServiceImpl.java
+++ b/web3/src/main/java/org/hiero/mirror/web3/service/ContractStateServiceImpl.java
@@ -94,7 +94,8 @@ final class ContractStateServiceImpl implements ContractStateService {
         });
 
         final var contractSlotValues = contractStateRepository.findStorageBatch(contractId.getId(), cachedSlots);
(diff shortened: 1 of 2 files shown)


web3/src/test/java/org/hiero/mirror/web3/service/ContractStateServiceTest.java
```diff
@@ -2,6 +2,7 @@
 
 package org.hiero.mirror.web3.service;
 
+import static net.i2p.crypto.eddsa.Utils.hexToBytes;
```
bilyana-gospodinova: We can probably use some other import like `org.apache.commons.codec.binary.Hex.decodeHex(String)` similar to other classes that we have.

web3/src/main/java/org/hiero/mirror/web3/service/ContractStateServiceImpl.java
```diff
@@ -83,20 +83,17 @@ public Optional<byte[]> findStorageByBlockTimestamp(
...
+        final var contractSlotsCache = ((CaffeineCache) this.contractSlotsCache.get(
+                contractId, () -> cacheManagerSlotsPerContract.getCache(contractId.toString())));
         // Cached slot keys for contract, whose slot values are not present in the contractStateCache
         final var cachedSlots = new ArrayList<byte[]>();
-        ((CaffeineCache) contractSlotsCache).getNativeCache().asMap().keySet().forEach(slot -> {
-            final var slotBytes = ((ByteBuffer) slot).array();
-            final var value = contractStateCache.putIfAbsent(generateCacheKey(contractId, slotBytes), EMPTY_VALUE);
-
-            if (value == null) {
+        synchronized (contractSlotsCache) {
```
kselveliev: Proves needed to avoid corner case when eviction is being done.

web3/src/main/java/org/hiero/mirror/web3/service/ContractStateServiceImpl.java
```diff
@@ -83,20 +83,17 @@ public Optional<byte[]> findStorageByBlockTimestamp(
...
-                contractId, () -> cacheManagerSlotsPerContract.getCache(contractId.toString()));
-        contractSlotsCache.putIfAbsent(ByteBuffer.wrap(key), EMPTY_VALUE);
-
+        final var contractSlotsCache = ((CaffeineCache) this.contractSlotsCache.get(
+                contractId, () -> cacheManagerSlotsPerContract.getCache(contractId.toString())));
         // Cached slot keys for contract, whose slot values are not present in the contractStateCache
         final var cachedSlots = new ArrayList<byte[]>();
-        ((CaffeineCache) contractSlotsCache).getNativeCache().asMap().keySet().forEach(slot -> {
-            final var slotBytes = ((ByteBuffer) slot).array();
-            final var value = contractStateCache.putIfAbsent(generateCacheKey(contractId, slotBytes), EMPTY_VALUE);
```
kselveliev, line 94: This is unneeded, now removed.
Puts empty values in contract state cache and does pretty bad in concurrent environment.

web3/src/main/java/org/hiero/mirror/web3/service/ContractStateServiceImpl.java
```diff
@@ -105,7 +102,6 @@ private Optional<byte[]> findStorageBatch(final EntityId contractId, final byte[
             final byte[] slotKey = contractSlotValue.getSlot();
             final byte[] slotValue = contractSlotValue.getValue();
             contractStateCache.put(generateCacheKey(contractId, slotKey), slotValue);
-            contractSlotsCache.put(ByteBuffer.wrap(slotKey), EMPTY_VALUE);
```
kselveliev, line 108: Also unneeded.

web3/src/main/java/org/hiero/mirror/web3/service/ContractStateServiceImpl.java
sdimitrov9: Consider removing the synchronized block and modify the return statement of the method to:

```
return Optional.ofNullable(cachedValue)
                .or(() -> contractStateRepository.findStorage(contractId.getId(), key));
```

This way we should be able to avoid the problem cause by the corner case where the cache eviction clears the key that was just added by the `contractSlotsCache.putIfAbsent(ByteBuffer.wrap(key), EMPTY_VALUE);`

sdimitrov9: The above approach would deal with the corner case, but it will do an unnecessary query in cases were the slot key is not present in the db at all, so it will impact performance. Another approach should be investigated.

sdimitrov9: Consider adding         
`boolean evicted = !cachedSlots.contains(ByteBuffer.wrap(key).array());`
after the findStorageBatch statement and the following if statement before the final return:
```
if (cachedValue == null && evicted) {
            return contractStateRepository.findStorage(contractId.getId(), key);
        }
```

kselveliev: Addressed

web3/src/main/java/org/hiero/mirror/web3/service/ContractStateServiceImpl.java
```diff
@@ -83,35 +83,35 @@ public Optional<byte[]> findStorageByBlockTimestamp(
...
 
             if (Arrays.equals(slotKey, key)) {
                 cachedValue = slotValue;
             }
         }
 
+        // If the cache key was evicted and hasn't been requested since, the cached value will be null.
+        // In that case, fall back to the original query.
+        if (cachedValue == null && evicted) {
```
sdimitrov9: I think If slot key is evicted cachedValue will always be null. In that case the if condition should be just:
`if (evicted) { ....}`
Considering the complexity of the code it might be reasonable to rename evicted to isKeyEvictedFromCache for better clarity.

web3/src/test/java/org/hiero/mirror/web3/service/ContractStateServiceTest.java
kselveliev: Addressed

web3/src/main/java/org/hiero/mirror/web3/service/ContractStateServiceImpl.java
kselveliev: Correct, done

sdimitrov9 (approved):

web3/src/main/java/org/hiero/mirror/web3/service/ContractStateServiceImpl.java
```diff
@@ -83,35 +83,36 @@ public Optional<byte[]> findStorageByBlockTimestamp(
...
-        final var contractSlotsCache = this.contractSlotsCache.get(
-                contractId, () -> cacheManagerSlotsPerContract.getCache(contractId.toString()));
-        contractSlotsCache.putIfAbsent(ByteBuffer.wrap(key), EMPTY_VALUE);
-
+        final var contractSlotsCache = ((CaffeineCache) this.contractSlotsCache.get(
+                contractId, () -> cacheManagerSlotsPerContract.getCache(contractId.toString())));
         // Cached slot keys for contract, whose slot values are not present in the contractStateCache
         final var cachedSlots = new ArrayList<byte[]>();
-        ((CaffeineCache) contractSlotsCache).getNativeCache().asMap().keySet().forEach(slot -> {
+        contractSlotsCache.putIfAbsent(ByteBuffer.wrap(key), EMPTY_VALUE);
```
xin-hedera: `cachedSlots` has to be a `HashSet<ByteBuffer>` for `contains` to work, also for efficient `contains` check

```suggestion
        final var wrappedkey = ByteBuffer.wrap(key);
        contractSlotsCache.putIfAbsent(wrappedkey, EMPTY_VALUE);
        final var cachedSlots = contractSlotsCache.getNativeCache().asMap().keySet();
        ...
        
        boolean isKeyEvictedFromCache = !cachedSlots.contains(wrappedKey);
```

web3/src/main/java/org/hiero/mirror/web3/service/ContractStateServiceImpl.java
```diff
@@ -83,35 +83,36 @@ public Optional<byte[]> findStorageByBlockTimestamp(
...
 
             if (Arrays.equals(slotKey, key)) {
                 cachedValue = slotValue;
             }
         }
 
+        // If the cache key was evicted and hasn't been requested since, the cached value will be null.
+        // In that case, fall back to the original query.
+        if (isKeyEvictedFromCache) {
+            return contractStateRepository.findStorage(contractId.getId(), key);
```
xin-hedera, line 119: cache the slot value

xin-hedera (changes requested):

web3/src/main/java/org/hiero/mirror/web3/service/ContractStateServiceImpl.java
```diff
@@ -83,35 +83,36 @@ public Optional<byte[]> findStorageByBlockTimestamp(
...
-                contractId, () -> cacheManagerSlotsPerContract.getCache(contractId.toString()));
-        contractSlotsCache.putIfAbsent(ByteBuffer.wrap(key), EMPTY_VALUE);
-
+        final var contractSlotsCache = ((CaffeineCache) this.contractSlotsCache.get(
+                contractId, () -> cacheManagerSlotsPerContract.getCache(contractId.toString())));
         // Cached slot keys for contract, whose slot values are not present in the contractStateCache
         final var cachedSlots = new ArrayList<byte[]>();
-        ((CaffeineCache) contractSlotsCache).getNativeCache().asMap().keySet().forEach(slot -> {
+        contractSlotsCache.putIfAbsent(ByteBuffer.wrap(key), EMPTY_VALUE);
+        (contractSlotsCache).getNativeCache().asMap().keySet().forEach(slot -> {
```
steven-sheehy: Unnecessary parenthesis
```suggestion
        contractSlotsCache.getNativeCache().asMap().keySet().forEach(slot -> {
```

web3/src/main/java/org/hiero/mirror/web3/service/ContractStateServiceImpl.java
steven-sheehy: The repository has `@Cacheable` so this should already be cached before returning, unless there's some bug

web3/src/main/java/org/hiero/mirror/web3/service/ContractStateServiceImpl.java
steven-sheehy: Maybe instead we can just compare the key inside the first loop to see if it's present?

sdimitrov9: I think If we do the check in the first loop the performance won’t change, considering the Caffeine cache is using ConcurrentHashMap as an underlying structure. Since it is unordered the check via iteration would has O(n/2) complexity on average similar to the ArrayList contains(). Also the lambda expression would become more complex.
On the other hand,  the contains() method  of the HashSet would have constant time complexity on average compared to O(n) in the ArrayList case.

IvanKavaldzhiev: The `contains` in ConcurrentHashMap should be also constant.

web3/src/main/java/org/hiero/mirror/web3/service/ContractStateServiceImpl.java
kselveliev: Addressed

web3/src/main/java/org/hiero/mirror/web3/service/ContractStateServiceImpl.java
xin-hedera: missed that

web3/src/main/java/org/hiero/mirror/web3/service/ContractStateServiceImpl.java
sdimitrov9: It is, but iterating through the structure isn't.

xin-hedera: Agree we can also do the `isKeyEvictedFromCache` check inside the loop. We can't avoid the loop because `contractStateRepository.findStorageBatch` takes `List<byte[]>` while the keyset in the underlying `cache.asMap()` is a `Set<ByteBuffer>`.

kselveliev: Refactored the code to work with the approach that Steven suggested initially.
Comparing the key inside the loop to check if it's present.

xin-hedera (approved):
LGTM

PR #11564, opened 2025-07-11
Update K8s version upgrade script

### LINKED ISSUE #10973: Enhance K8s Upgrade Script
### Problem

The current upgrade script only works well for updating the citus node pools. It should be able to upgrade the master as well as the default node pool

### Solution

- Need to failover to coord-0 if it is not primary
- Support a custom linux config for node pools 
- Detect whether there are non-citus pools involved in the upgrade and upgrade those pools before scaling down citus
- support upgrading the master 
- gcloud cli only supports upgrading one minor version at a time for the master. Need logic to walk through all upgrades of necessary versions

### Alternatives

_No response_

jnels124:
**Description**:
* add support to upgrade cluster master
* add version validation
* ensure primary pod is always first in stateful set when pausing
* support providing linux config
* do non citus pool upgrades first for less downtime
* update release automation workflow to prevent benign failure when cutting pre-releases

**Related issue(s)**:

Fixes #10973 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/.github/workflows/release-automation.yml
+++ b/.github/workflows/release-automation.yml
@@ -120,6 +120,7 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Output Release Notes
+        if: ${{ steps.version_parser.outputs.prerelease == '' }}
         continue-on-error: true
         run: cat changelog.md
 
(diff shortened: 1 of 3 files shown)


xin-hedera (approved):
LGTM

PR #11583, opened 2025-07-14
Bump com.hedera.hashgraph:app to 0.64.0

### LINKED ISSUE #11542: Add BLOCK_STREAM_INFO singleton
### Problem

When trying to debug web3 tests with shard & realm and hedera app version 0.64.0-SNAPSHOT noticed that they all started failing with 

```
MirrorEvmTransactionException(message=Unknown singleton state key: BLOCK_STREAM_INFO, detail=, data=, dataDecoded=)
java.lang.RuntimeException: MirrorEvmTransactionException(message=Unknown singleton state key: BLOCK_STREAM_INFO, detail=, data=, dataDecoded=)
```

### Solution

We will have to add the BLOCK_STREAM_INFO singleton in our code

### Alternatives

_No response_

IvanKavaldzhiev:
**Description**:
This PR bumps hedera.app to 0.64.0. The protobuf is kept at 0.63.5 since it wasn't updated at this release. It seems this does not interfere with bumping the hedera.app itself.

The PR also aligns some of the classpath overriden classes with the latest changes from hedera.app.

**Related issue(s)**:

Fixes #11542 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/.github/workflows/releases-comparator.yml
+++ b/.github/workflows/releases-comparator.yml
@@ -108,10 +108,9 @@ jobs:
             local target_dir="$2"
 
             git clone --depth 1 --branch "$tag" "$CONSENSUS_NODE_REPO" "$target_dir" > /dev/null 2>&1
-            (
-              cd "$target_dir"
-              ./gradlew hapi:generatePbjSource > /dev/null 2>&1
-            )
+            cd "$target_dir"
+            ./gradlew hapi:generatePbjSource > /dev/null 2>&1
+            cd - > /dev/null
           }
 
           generate_pbj_source "$TAG_ONE" "$TMP_DIR_ONE"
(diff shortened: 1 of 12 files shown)


steven-sheehy (changes requested):
Actually, I see some problems with the release comparator workflow:

* I see `File changed: com/hedera/node/app/service/contract/impl/exec/processors/CustomMessageCallProcessor.java` but no change to it in this PR
* `ConversionUtils` was modified in this PR but it's not noted in the workflow as changing. Probably we need to add it to the whitelisted files in the workflow.
* The workflow uses `set -x` which prints every command and is too verbose.

web3/src/main/java/org/hiero/mirror/web3/state/singleton/BlockStreamInfoSingleton.java
```diff
@@ -0,0 +1,23 @@
...
+
+package org.hiero.mirror.web3.state.singleton;
+
+import static com.hedera.node.app.blocks.schemas.V0560BlockStreamSchema.BLOCK_STREAM_INFO_KEY;
+
+import com.hedera.hapi.node.state.blockstream.BlockStreamInfo;
+import jakarta.inject.Named;
+
+@Named
+public class BlockStreamInfoSingleton
```
steven-sheehy:
```suggestion
final class BlockStreamInfoSingleton
```

steven-sheehy: We should have a simple test for it.

IvanKavaldzhiev: Added

IvanKavaldzhiev: Fixed

.github/workflows/releases-comparator.yml
```diff
@@ -79,8 +79,6 @@ jobs:
 
       - name: Run diff script
         run: |
-          set -ex
```
steven-sheehy, line 82: `set -euo pipefail`

.github/workflows/releases-comparator.yml
```diff
@@ -93,6 +91,7 @@ jobs:
             "com/swirlds/state/spi/WritableKVStateBase.java"
             "com/hedera/node/app/service/contract/impl/exec/processors/CustomMessageCallProcessor.java"
```
steven-sheehy, line 96: Need to identify why this changed between versions but no change in this PR to account for that.

steven-sheehy, line 96: Need to add `CustomBalanceOperation` and `CustomCallOperation` (we renamed ours but still we should check the original) to this list as well and verify if any change.

steven-sheehy (changes requested):

.github/workflows/releases-comparator.yml
IvanKavaldzhiev: Added

IvanKavaldzhiev: It was an issue caused by missing the correct paths to search these files in. The default result was reporting there was a change in the file. Correct path is now added.

.github/workflows/releases-comparator.yml
IvanKavaldzhiev: Added

PR #11585, opened 2025-07-14
Add a workflow to deploy a specific version to preprod environment

### LINKED ISSUE #11484: Add a GitHub workflow that triggers a deployment of a specific version
### Problem

As part of CITR, ops needs the ability to deploy a specific version to a specific environment.

### Solution

* Add a `workflow_dispatch` GitHub Action that takes in an environment name (e.g. `performance`) and a version `v0.132.0` and updates the deploy branch for that environment to trigger a deploy.
* No user interaction should be necessary
* Ops will trigger this workflow via some external process.
* As a safety check, we should exclude public environments

### Alternatives

_No response_

xin-hedera:
**Description**:

This PR adds a workflow to deploy a specific version to preprod environment

**Related issue(s)**:

Fixes #11484 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/.github/workflows/deploy-preprod.yaml
+++ b/.github/workflows/deploy-preprod.yaml
@@ -1,0 +1,82 @@
+# SPDX-License-Identifier: Apache-2.0
+
+name: Deploy to Preprod Environment
+
+on:
+  workflow_dispatch:
+    inputs:
+      environment:
+        description: "Environment to deploy to "
+        required: true
+        type: string
+      version:
+        description: "Version to deploy, e.g., 0.132.0"
+        required: true
+        type: string
+
+permissions:
+  contents: write
+
+defaults:
+  run:
+    shell: bash
+
+env:
+  LC_ALL: C.UTF-8
+
+jobs:
+  deploy:
+    name: Deploy
+    runs-on: hiero-mirror-node-linux-medium
+    steps:
+      - name: Harden Runner
+        uses: step-security/harden-runner@6c439dc8bdf85cadbbce9ed30d1c7b959517bc49 # v2.12.2
+        with:
+          egress-policy: audit
+
+      - name: Validate environment
+        run: |
+          [[ ! "${{ github.event.inputs.environment }}" =~ ^(common|integration|performance-citus)$ ]] || \
+          { echo "Deploying to ${{ github.event.inputs.environment }} is not allowed" && false; }
+
+      - name: Validate version
+        run: curl -fsSL -o /dev/null ${{ github.server_url }}/${{ github.repository }}/releases/tags/v${{ github.event.inputs.version }}
+
+      - name: Checkout Repository
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          ref: deploy
+          token: ${{ secrets.HEDERA_BOT_TOKEN }}
+
+      - name: Install yq
+        env:
+          BINARY: yq_linux_amd64
+          TARGET: /home/runner/.local/bin
+          VERSION: v4.45.1
+        run: |
+          mkdir -p "${TARGET}"
+          cd "${TARGET}"
+          curl -sL https://github.com/mikefarah/yq/releases/download/${VERSION}/${BINARY}.tar.gz --output - | tar -xvz
+          mv ${BINARY} yq
+
+      - name: Update deployment version
+        run: yq -i '.spec.chart.spec.version |= "${{ github.event.inputs.version }}"' clusters/preprod/${{ github.event.inputs.environment }}/helmrelease.yaml
+
+      - name: Import GPG Key
+        id: gpg_importer
+        uses: step-security/ghaction-import-gpg@c86c374c0659a6c2d1284bccf8af889e73ce8fe0 # v6.3.0
+        with:
+          git_commit_gpgsign: true
+          git_tag_gpgsign: true
+          git_user_signingkey: true
+          gpg_private_key: ${{ secrets.GPG_PRIVATE_KEY }}
+          passphrase: ${{ secrets.GPG_PASSPHRASE }}
+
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-preprod.yaml
```diff
@@ -0,0 +1,82 @@
...
+          TARGET: /home/runner/.local/bin
+          VERSION: v4.45.1
+        run: |
+          mkdir -p "${TARGET}"
+          cd "${TARGET}"
+          curl -sL https://github.com/mikefarah/yq/releases/download/${VERSION}/${BINARY}.tar.gz --output - | tar -xvz
+          mv ${BINARY} yq
+
+      - name: Update deployment version
+        run: yq -i '.spec.chart.spec.version |= "${{ github.event.inputs.version }}"' clusters/preprod/${{ github.event.inputs.environment }}/helmrelease.yaml
```
jnels124: not sure this needs to work for integration env but if so I dont think this will work because that key doesnt exist

jnels124 (approved):

.github/workflows/deploy-preprod.yaml
xin-hedera: common / integration / performance-citus are in the not allowed env list and checked in the validate environment step

.github/workflows/deploy-preprod.yaml
```diff
@@ -0,0 +1,82 @@
+# SPDX-License-Identifier: Apache-2.0
+
+name: Deploy to Preprod Environment
+
+on:
+  workflow_dispatch:
+    inputs:
+      environment:
+        description: "Environment to deploy to "
```
steven-sheehy:
```suggestion
        description: "Environment"
```

.github/workflows/deploy-preprod.yaml
```diff
@@ -0,0 +1,82 @@
...
+        required: true
+        type: string
+      version:
+        description: "Version to deploy, e.g., 0.132.0"
```
steven-sheehy:
```suggestion
        description: "Version"
```

.github/workflows/deploy-preprod.yaml
```diff
@@ -0,0 +1,82 @@
...
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          ref: deploy
+          token: ${{ secrets.HEDERA_BOT_TOKEN }}
+
+      - name: Install yq
+        env:
+          BINARY: yq_linux_amd64
+          TARGET: /home/runner/.local/bin
+          VERSION: v4.45.1
```
steven-sheehy: Can we use the latest v4.46.1?

.github/workflows/deploy-preprod.yaml
```diff
@@ -0,0 +1,82 @@
...
+        with:
+          egress-policy: audit
+
+      - name: Validate environment
+        run: |
+          [[ ! "${{ github.event.inputs.environment }}" =~ ^(common|integration|performance-citus)$ ]] || \
+          { echo "Deploying to ${{ github.event.inputs.environment }} is not allowed" && false; }
+
+      - name: Validate version
+        run: curl -fsSL -o /dev/null ${{ github.server_url }}/${{ github.repository }}/tree/v${{ github.event.inputs.version }}
```
steven-sheehy: We should support versions with or without `v` prefix.

.github/workflows/deploy-preprod.yaml
```diff
@@ -0,0 +1,82 @@
...
+    runs-on: hiero-mirror-node-linux-medium
+    steps:
+      - name: Harden Runner
+        uses: step-security/harden-runner@6c439dc8bdf85cadbbce9ed30d1c7b959517bc49 # v2.12.2
+        with:
+          egress-policy: audit
+
+      - name: Validate environment
+        run: |
+          [[ ! "${{ github.event.inputs.environment }}" =~ ^(common|integration|performance-citus)$ ]] || \
```
steven-sheehy: We explicitly must support performance environment for CITR. We should be able to take as input `performance` without `-citus` and it add it if necessary.

steven-sheehy: We should reject public environments as well up front. Otherwise I assume it will fail later during yq stage with a less than desirable error message.

steven-sheehy (changes requested):

.github/workflows/deploy-preprod.yaml
```diff
@@ -0,0 +1,93 @@
...
+      - name: Harden Runner
+        uses: step-security/harden-runner@6c439dc8bdf85cadbbce9ed30d1c7b959517bc49 # v2.12.2
+        with:
+          egress-policy: audit
+
+      - name: Validate environment
+        run: |
+          set -euo pipefail
+
+          ENVIRONMENT="${{ github.event.inputs.environment }}"
```
steven-sheehy: > Using variable interpolation `${{...}}` with `github` context data in a `run:` step could allow an attacker to inject their own code into the runner.

Move to `env:` block.

.github/workflows/deploy-preprod.yaml
```diff
@@ -0,0 +1,93 @@
...
+          [[ ! "$ENVIRONMENT" =~ ^(common|integration|mainnet.*|previewnet.*|testnet.*)$ ]] || \
+            { echo "Deploying to $ENVIRONMENT is not allowed" && false; }
+          [[ "$ENVIRONMENT" = "performance" ]] && ENVIRONMENT="${ENVIRONMENT}-citus" || true
+          echo ENVIRONMENT=${ENVIRONMENT} >> $$GITHUB_ENV
+
+      - name: Validate version
+        run: |
+          set -euo pipefail
+
+          VERSION="${{ github.event.inputs.version }}"
```
steven-sheehy: Move to `env:` block

PR #11588, opened 2025-07-15
Add drop sql script to support version rollback

### LINKED ISSUE #11483: Create a drop.sql to support version rollbacks
### Problem

As part of CITR, ops needs the capability to reset an environment _and_ rollback its version. The current `cleanup.sql` only truncates and can't handle scenarios where the schema changes.

### Solution

* Add a `drop.sql` next to `cleanup.sql`
* Implement `drop.sql` like the previous [deleted](https://github.com/hiero-ledger/hiero-mirror-node/pull/7350/files) script or by looping through PostgreSQL tables/types/etc and dropping them.
* Test thoroughly that an existing database can be dropped and restarted with an older version of the importer with schema changes.

### Alternatives

_No response_

sdimitrov9:
**Description**:
Add drop sql script to support version rollback.

Tested in local kubernetes cluster against citus. After applying the script the importer still manages to recover the data successfully.

**Related issue(s)**:

Fixes #11483 

**Checklist**

### DIFF
--- a/importer/src/main/resources/db/scripts/drop.sql
+++ b/importer/src/main/resources/db/scripts/drop.sql
@@ -104,12 +104,47 @@ begin
         execute format('drop type if exists public.%I cascade', obj.typname);
     end loop;
 
-    -- drop functions
-    drop function if exists f_file_create(_file_name character varying) cascade;
-    drop function if exists f_file_complete(_file_id bigint, _file_hash character varying, _prev_hash character varying) cascade;
-    drop function if exists f_entity_create(_shard bigint, _realm bigint, _num bigint, _type_id integer, _exp_time_sec bigint, _exp_time_nanos bigint, _exp_time_ns bigint, _auto_renew bigint, _admin_key bytea, _key bytea, _proxy_acc_id bigint) cascade;
-    drop function if exists nanos_to_timestamptz(nanos bigint) cascade;
-    drop function if exists timestamptz_to_nanos(ts timestamp with time zone) cascade;
+    -- drop all plpgsql functions from the public schema
+    for obj in
+        select
+            n.nspname as schema_name,
+            p.proname as function_name,
+            pg_get_function_identity_arguments(p.oid) as args
+        from pg_proc p
+        join pg_namespace n on p.pronamespace = n.oid
+        join pg_language l on p.prolang = l.oid
+        where n.nspname = 'public'
+          and l.lanname = 'plpgsql'
+          and p.prokind = 'f'
+    loop
+        execute format(
+            'drop function if exists %I.%I(%s) cascade',
+            obj.schema_name,
+            obj.function_name,
+            obj.args
+        );
+    end loop;
+
+    -- drop all plpgsql procedures from the public schema
+    for obj in
+        select
+            n.nspname as schema_name,
+            p.proname as procedure_name,
+            pg_get_function_identity_arguments(p.oid) as args
+        from pg_proc p
+        join pg_namespace n on p.pronamespace = n.oid
+        join pg_language l on p.prolang = l.oid
+        where n.nspname = 'public'
+          and l.lanname = 'plpgsql'
+          and p.prokind = 'p'
+    loop
+        execute format(
+            'drop procedure if exists %I.%I(%s) cascade',
+            obj.schema_name,
+            obj.procedure_name,
+            obj.args
+        );
+    end loop;
 
 end;
 $$;
--- a/importer/src/main/resources/db/scripts/drop.sql
+++ b/importer/src/main/resources/db/scripts/drop.sql
@@ -35,19 +35,17 @@ begin
         execute 'alter default privileges for role mirror_node in schema public revoke all on sequences from mirror_api';
         execute 'alter default privileges for role mirror_node in schema public revoke all on functions from mirror_api';
         execute 'alter default privileges for role mirror_node in schema public revoke all on types from mirror_api';
-    end if;
 
-    -- revoke all privileges on mirror_node_time_partitions from mirror_api only if it exists
-    if exists (
-        select 1
-        from pg_class c
-        join pg_namespace n on n.oid = c.relnamespace
-        where c.relname = 'mirror_node_time_partitions'
-          and n.nspname = 'public'
-    ) and exists (
-        select 1 from pg_roles where rolname = 'mirror_api'
-    ) then
-        execute 'revoke all privileges on public.mirror_node_time_partitions from mirror_api';
+        -- revoke privileges on mirror_node_time_partitions only if the table exists
+        if exists (
+            select 1
+            from pg_class c
+            join pg_namespace n on n.oid = c.relnamespace
+            where c.relname = 'mirror_node_time_partitions'
+              and n.nspname = 'public'
+        ) then
+            execute 'revoke all privileges on public.mirror_node_time_partitions from mirror_api';
+        end if;
     end if;
 
     drop role if exists mirror_api;
--- a/importer/src/main/resources/db/scripts/drop.sql
+++ b/importer/src/main/resources/db/scripts/drop.sql
@@ -57,8 +57,7 @@ begin
         select table_schema, table_name
         from information_schema.views
         where table_schema = 'public'
-          and table_name not like 'citus%'
-          and table_name not like 'pg_%'
+          and table_name like 'mirror_%'
     loop
         execute format('drop view if exists %I.%I cascade', obj.table_schema, obj.table_name);
     end loop;
@@ -103,20 +102,6 @@ begin
           )
     loop
         execute format('drop type if exists public.%I cascade', obj.typname);
-    end loop;
-
-    -- drop all sequences
-    for obj in
-        select sequence_name from information_schema.sequences where sequence_schema = 'public'
-    loop
-        execute format('drop sequence if exists public.%I cascade', obj.sequence_name);
-    end loop;
-
-    -- drop all indexes
-    for obj in
-        select indexname from pg_indexes where schemaname = 'public'
-    loop
-        execute format('drop index if exists public.%I cascade', obj.indexname);
     end loop;
 
     -- drop functions
--- a/importer/src/main/resources/db/scripts/drop.sql
+++ b/importer/src/main/resources/db/scripts/drop.sql
@@ -4,60 +4,28 @@ do $$
 declare
     obj record;
 begin
-    -- drop all foreign key constraints in public
+    -- drop all tables from public and temporary schema
     for obj in
-        select conname, conrelid::regclass
-        from pg_constraint c
-        join pg_class r on c.conrelid = r.oid
-        join pg_namespace n on r.relnamespace = n.oid
-        where contype = 'f'
-          and n.nspname = 'public'
+        select schemaname, tablename
+        from pg_tables
+        where schemaname in ('public', 'temporary')
     loop
-        execute format('alter table %s drop constraint %I', obj.conrelid, obj.conname);
-    end loop;
-
-    -- drop all tables from public schema
-    for obj in
-        select tablename from pg_tables where schemaname = 'public'
-    loop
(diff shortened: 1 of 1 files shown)


importer/src/main/resources/db/scripts/drop.sql
```diff
@@ -0,0 +1,130 @@
...
+    for obj in
+        select indexname from pg_indexes where schemaname = 'public'
+    loop
+        execute format('drop index if exists public.%I cascade', obj.indexname);
+    end loop;
+
+    -- drop functions
+    drop function if exists f_file_create(_file_name character varying) cascade;
+    drop function if exists f_file_complete(_file_id bigint, _file_hash character varying, _prev_hash character varying) cascade;
+    drop function if exists f_entity_create(_shard bigint, _realm bigint, _num bigint, _type_id integer, _exp_time_sec bigint, _exp_time_nanos bigint, _exp_time_ns bigint, _auto_renew bigint, _admin_key bytea, _key bytea, _proxy_acc_id bigint) cascade;
```
steven-sheehy: These 3 functions were removed in a previous migration and don't exist on the latest schema. Should remove these lines.

importer/src/main/resources/db/scripts/drop.sql
```diff
@@ -0,0 +1,130 @@
...
+    drop function if exists nanos_to_timestamptz(nanos bigint) cascade;
+    drop function if exists timestamptz_to_nanos(ts timestamp with time zone) cascade;
```
steven-sheehy: These are specific to v2, which is fine since you have if exists. However, we are missing some other functions and we should avoid hardcoding these. Instead loop over something like:
```
select * from information_schema.routines where specific_schema = 'public' and external_language = 'PLPGSQL';
```

importer/src/main/resources/db/scripts/drop.sql
```diff
@@ -0,0 +1,130 @@
...
+    -- drop all sequences
+    for obj in
+        select sequence_name from information_schema.sequences where sequence_schema = 'public'
+    loop
+        execute format('drop sequence if exists public.%I cascade', obj.sequence_name);
+    end loop;
+
+    -- drop all indexes
+    for obj in
+        select indexname from pg_indexes where schemaname = 'public'
```
steven-sheehy: If we've dropped all tables it should have cascaded to drop all indexes already.

importer/src/main/resources/db/scripts/drop.sql
```diff
@@ -0,0 +1,130 @@
...
+                and c.relkind = 'v'
+                and d.deptype != 'i'
+          )
+    loop
+        execute format('drop type if exists public.%I cascade', obj.typname);
+    end loop;
+
+    -- drop all sequences
+    for obj in
+        select sequence_name from information_schema.sequences where sequence_schema = 'public'
```
steven-sheehy: If we've dropped all tables it should have cascaded to drop all sequences already.

importer/src/main/resources/db/scripts/drop.sql
```diff
@@ -0,0 +1,130 @@
...
+
+    -- revoke all privileges on mirror_node_time_partitions from mirror_api only if it exists
+    if exists (
+        select 1
+        from pg_class c
+        join pg_namespace n on n.oid = c.relnamespace
+        where c.relname = 'mirror_node_time_partitions'
+          and n.nspname = 'public'
+    ) and exists (
+        select 1 from pg_roles where rolname = 'mirror_api'
```
steven-sheehy: We already do this check above, so could move this in side that same block.

importer/src/main/resources/db/scripts/drop.sql
```diff
@@ -0,0 +1,130 @@
...
+
+    drop role if exists mirror_api;
+
+    -- drop all views only in the public schema that do not start with 'citus' or 'pg_'
+    for obj in
+        select table_schema, table_name
+        from information_schema.views
+        where table_schema = 'public'
+          and table_name not like 'citus%'
+          and table_name not like 'pg_%'
```
steven-sheehy: Instead of black listing it would be safer to whitelist just mirror views. All of them start with `mirror_` right?

```suggestion
          and table_name like 'mirror_%'
```

importer/src/main/resources/db/scripts/drop.sql
sdimitrov9: Addressed

importer/src/main/resources/db/scripts/drop.sql
sdimitrov9: Addressed

importer/src/main/resources/db/scripts/drop.sql
sdimitrov9: Addressed

importer/src/main/resources/db/scripts/drop.sql
sdimitrov9: Addressed

importer/src/main/resources/db/scripts/drop.sql
sdimitrov9: Addressed

importer/src/main/resources/db/scripts/drop.sql
sdimitrov9: Fixed

importer/src/main/resources/db/scripts/drop.sql
```diff
@@ -0,0 +1,148 @@
...
+    for obj in
+        select conname, conrelid::regclass
+        from pg_constraint c
+        join pg_class r on c.conrelid = r.oid
+        join pg_namespace n on r.relnamespace = n.oid
+        where contype = 'f'
+          and n.nspname = 'public'
+    loop
+        execute format('alter table %s drop constraint %I', obj.conrelid, obj.conname);
+    end loop;
```
xin-hedera: we don't have foreign keys, so shouldn't need it?

importer/src/main/resources/db/scripts/drop.sql
```diff
@@ -0,0 +1,148 @@
...
+    loop
+        execute format('drop table if exists public.%I cascade', obj.tablename);
+    end loop;
+
+    -- drop all tables from temporary schema
+    for obj in
+        select tablename from pg_tables where schemaname = 'temporary'
+    loop
+        execute format('drop table if exists temporary.%I cascade', obj.tablename);
+    end loop;
```
xin-hedera: combine to one loop

`select schemaname, tablename from pg_tables where schemaname in ('public', 'temporary')`

`execute format('drop table if exists %I.%I cascade', obj.schenaname, obj.tablename);`

importer/src/main/resources/db/scripts/drop.sql
```diff
@@ -0,0 +1,148 @@
...
+        -- revoke privileges on mirror_node_time_partitions only if the table exists
+        if exists (
+            select 1
+            from pg_class c
+            join pg_namespace n on n.oid = c.relnamespace
+            where c.relname = 'mirror_node_time_partitions'
+              and n.nspname = 'public'
+        ) then
+            execute 'revoke all privileges on public.mirror_node_time_partitions from mirror_api';
+        end if;
```
xin-hedera:
```suggestion
        execute 'drop owned by mirror_api cascade';
```

importer/src/main/resources/db/scripts/drop.sql
```diff
@@ -0,0 +1,148 @@
...
+    end if;
+
+    drop role if exists mirror_api;
```
xin-hedera: perhaps also move this into the if block after `execute 'drop owned by mirror_api cascade'`

importer/src/main/resources/db/scripts/drop.sql
```diff
@@ -0,0 +1,148 @@
...
+
+    -- drop all views only in the public schema that do not start with 'citus' or 'pg_'
```
xin-hedera: comment doesn't match code

importer/src/main/resources/db/scripts/drop.sql
```diff
@@ -0,0 +1,148 @@
...
+    for obj in
+        select table_schema, table_name
+        from information_schema.views
+        where table_schema = 'public'
+          and table_name like 'mirror_%'
+    loop
+        execute format('drop view if exists %I.%I cascade', obj.table_schema, obj.table_name);
+    end loop;
```
xin-hedera: the table_schema here is always `public`, so don't have to select it.

importer/src/main/resources/db/scripts/drop.sql
```diff
@@ -0,0 +1,148 @@
...
+    -- drop user-defined types in public schema that are not:
+    -- 1. Named with pg_/citus/gbtree
+    -- 2. Used by views
+    -- 3. Owned by extensions (like pg_trgm)
+    for obj in
+        select t.typname
+        from pg_type t
+        join pg_namespace n on n.oid = t.typnamespace
+        where n.nspname = 'public'
+          and t.typtype in ('c', 'e', 'r', 'd')
```
xin-hedera: should we just care about the types: `'e', 'd'`?

importer/src/main/resources/db/scripts/drop.sql
```diff
@@ -0,0 +1,148 @@
...
+          )
+          -- exclude types used by views
+          and not exists (
+              select 1
+              from pg_depend d
+              join pg_class c on d.refobjid = t.oid
+              where d.classid = 'pg_class'::regclass
+                and c.relkind = 'v'
+                and d.deptype != 'i'
+          )
```
xin-hedera: instead of all these conditions, can't we just limit the type owner to `mirror_node`?

importer/src/main/resources/db/scripts/drop.sql
```diff
@@ -0,0 +1,148 @@
...
+          and l.lanname = 'plpgsql'
+          and p.prokind = 'p'
+    loop
+        execute format(
+            'drop procedure if exists %I.%I(%s) cascade',
+            obj.schema_name,
+            obj.procedure_name,
+            obj.args
+        );
+    end loop;
```
xin-hedera: 1. should not just limit to `lanname = 'plpgsql'`
2. can use `drop routine if exists ...` since it can drop function / procedure

importer/src/main/resources/db/scripts/drop.sql
sdimitrov9: Fixed

importer/src/main/resources/db/scripts/drop.sql
sdimitrov9: Fixed

importer/src/main/resources/db/scripts/drop.sql
sdimitrov9: Fixed

importer/src/main/resources/db/scripts/drop.sql
sdimitrov9: Done

importer/src/main/resources/db/scripts/drop.sql
sdimitrov9: Done

importer/src/main/resources/db/scripts/drop.sql
sdimitrov9: Done

importer/src/main/resources/db/scripts/drop.sql
sdimitrov9: Removed

importer/src/main/resources/db/scripts/drop.sql
sdimitrov9: Addressed

importer/src/main/resources/db/scripts/drop.sql
sdimitrov9: Modified

xin-hedera (approved):
LGTM

jnels124 (approved):

steven-sheehy (approved):
LGTM

PR #11592, opened 2025-07-16
Enable slot caching

kselveliev:
**Description**:
This PR enables contract slot caching after reggression fixex

**Checklist**


web3/src/main/java/org/hiero/mirror/web3/repository/properties/CacheProperties.java
```diff
@@ -27,7 +27,7 @@ public class CacheProperties {
     @NotBlank
     private String contractState = "expireAfterWrite=2s,maximumSize=25000,recordStats";
```
kselveliev, line 29: We probably could also increase the expireAfterWrite back to 5 seconds

steven-sheehy: Let's keep at 2s to reduce risk of breaking customers

steven-sheehy (approved):

xin-hedera (approved):
LGTM

PR #11594, opened 2025-07-16
Fix failing web3 tests with non-zero shard and realm

### LINKED ISSUE #11456: Update Contract Call Tests to work with non-zero shard and realm
### Problem

After updating to tests to work with non zero shard and realm in web3, there are still some contract call tests failing. 

### Solution

* Configure shard/realm with non zero values
* run web3 test task
* Notice the test failures
* Fix remaining test failures

### Alternatives

_No response_

IvanKavaldzhiev:
**Description**:
This PR fixes the remaining failing web3 tests with non-zero shard and realm. They were only mono code based tests.

**Related issue(s)**:

Fixes #11456 

**Notes for reviewer**:

Dependant on #11542 

**Checklist**

### DIFF
--- a/.github/workflows/deploy-preprod.yaml
+++ b/.github/workflows/deploy-preprod.yaml
@@ -1,0 +1,94 @@
+# SPDX-License-Identifier: Apache-2.0
+
+name: Deploy to Preprod Environment
+
+on:
+  workflow_dispatch:
+    inputs:
+      environment:
+        description: "Environment"
+        required: true
+        type: string
+      version:
+        description: "Version"
+        required: true
+        type: string
+
+permissions:
+  contents: write
+
+defaults:
+  run:
+    shell: bash
+
+env:
+  LC_ALL: C.UTF-8
+
+jobs:
+  deploy:
+    name: Deploy
+    env:
+      ENVIRONMENT: ${{ github.event.inputs.environment }}
+      VERSION: ${{ github.event.inputs.version }}
+    runs-on: hiero-mirror-node-linux-medium
+    steps:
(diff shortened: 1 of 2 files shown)


web3/src/main/java/com/hedera/services/store/contracts/precompile/impl/TokenCreatePrecompile.java
```diff
@@ -402,7 +402,7 @@ private static TokenCreateWrapper getTokenCreateWrapperFungible(
                 isFungible,
                 tokenName,
                 tokenSymbol,
-                tokenTreasury.getAccountNum() != 0 ? tokenTreasury : null,
+                (tokenTreasury.getShardNum() != 0 || tokenTreasury.getRealmNum() != 0) ? tokenTreasury : null,
```
xin-hedera: why is this needed? The logic now becomes `tokenTreasury` is only valid when it has non-zero shard or non-zero realm.

The original logic should work regardless of shard / realm value since an entity whose number = 0 is not valid in any shard / realm network.

kselveliev:
Probably needs a bit more adjustments since web3 check is failing

web3/src/main/java/com/hedera/services/store/contracts/precompile/impl/TokenCreatePrecompile.java
IvanKavaldzhiev: Correct, the tests are failing with shard/realm=0. I reverted the changes.

xin-hedera (approved):
LGTM

PR #11599, opened 2025-07-16
SSL mode support for importer

### LINKED ISSUE #11439: SSL mode support for importer
Can we have `hiero.mirror.importer.db.sslMode` like we have for other components?

ashumahajan:
**Description**:

Add SSL mode support for importer

* Add `hiero.mirror.importer.db.sslMode=DISABLE` property
* Added property `sslMode` in `DBProperties` class
* Added documentation for the property in `hiero.mirror.importer.db.sslMode` `docs/configuration.md` file

**Related issue(s)**:

Fixes #11439 

**Checklist**

- [x] Documented in configuration.md file for the added property
- [x] Tested locally against DB using sslMode and also ensured that the property works with our current setup having DISABLE value


importer/src/main/java/org/hiero/mirror/importer/db/DBProperties.java
```diff
@@ -51,6 +51,9 @@ public class DBProperties {
     @NotBlank
     private String username = "";
 
+    @NotBlank
+    private String sslMode = "";
```
steven-sheehy: Need to set the default. We can use the jdbc enum to provide value validation as well.

```suggestion
    @NotNull
    private SslMode sslMode = SslMode.DISABLE;
```

docs/configuration.md
```diff
@@ -74,6 +74,7 @@ value, it is recommended to only populate overridden properties in the custom `a
...
 | `hiero.mirror.importer.db.username`                                             | mirror_importer                                      | The Importer username the processor uses to connect to the database                                                                                                                                                                                                |
+| `hiero.mirror.importer.db.sslMode`                                              | DISABLE                                              | The ssl level of protection against Eavesdropping, Man-in-the-middle (MITM) and Impersonation on the db connection. Accepts either DISABLE, ALLOW, PREFER, REQUIRE, VERIFY_CA or VERIFY_FULL.                                                                      |
```
steven-sheehy: The last `|` doesn't align with the other rows. This doesn't impact rendering but means the next person to touch the file may cause all table rows to resize. We should align the columns either by shortening the description or increase the column size of other rows to match. Preferably shorten.

importer/src/main/resources/application.yml
```diff
@@ -97,6 +98,7 @@ spring:
       data-source-properties:
         #loggerLevel: TRACE
         logUnclosedConnections: true
+        sslmode: ${hiero.mirror.importer.db.sslMode}
```
steven-sheehy: [sslmode](https://jdbc.postgresql.org/documentation/use/) is a string with values like `verify-full` but our property is an enum with values like `VERIFY_FULL` as per your docs. Are we sure that this conversion between the two works properly? Do we need to make ours a string with only lower kebab-case values?

steven-sheehy (changes requested):

importer/src/main/resources/application.yml
ashumahajan: You're right sslMode values like VERIFY_FULL are uppercase enum values from org.postgresql.jdbc.SslMode, and that's what we currently use. When I tested, I tested agaist REQUIRE (that does not verify the CA) I'll update the documentation to clarify this and reflect the correct format.

Also, for verify-full to work as expected, we also need to provide the sslrootcert property (i.e., path to the trusted root certificate). I'll make sure we capture this requirement and add support for it in the configuration as well.

docs/configuration.md
ashumahajan: I will fix that

importer/src/main/java/org/hiero/mirror/importer/db/DBProperties.java
ashumahajan: Thanks! Will take care of it

importer/src/main/resources/application.yml
steven-sheehy: I'm not saying to test specifically verify full or to document how to use it. Instead I'm concerned by the format difference of upper camel case vs kebab case and which works eith the jdbc driver

importer/src/main/resources/application.yml
```diff
@@ -17,6 +17,7 @@ hiero:
         schema: public
         tempSchema: temporary
         username: mirror_importer
+        sslMode: disable
```
steven-sheehy: nit: sort inline alphabetically

importer/src/main/java/org/hiero/mirror/importer/db/DBProperties.java
```diff
@@ -51,6 +52,9 @@ public class DBProperties {
     @NotBlank
     private String username = "";
 
+    @NotNull
+    private SslMode sslMode = SslMode.DISABLE;
```
steven-sheehy: nit: Sort before tempSchema

importer/src/main/resources/application.yml
steven-sheehy: nit: Sort after reWriteBatchedInserts

docs/configuration.md
```diff
@@ -74,6 +74,7 @@ value, it is recommended to only populate overridden properties in the custom `a
...
 | `hiero.mirror.importer.db.username`                                             | mirror_importer                                      | The Importer username the processor uses to connect to the database                                                                                                                                                                                                |
+| `hiero.mirror.importer.db.sslMode`                                              | disable                                              | The ssl level of protection against Eavesdropping, Man-in-the-middle (MITM) and Impersonation on the db connection. Accepts either disable, allow, prefer, require, verify-ca or verify-full.                                                                      |
```
steven-sheehy, line 76: nit: Sort before schema and can you add tempSchema?

steven-sheehy (changes requested):

PR #11601, opened 2025-07-17
Add workaround for sender is a smart contract simulations

### LINKED ISSUE #11547: Contract call fails when from address is a smart contract–based wallet
### Description

When the from address in an eth_call is smart contract–based wallet, the simulation fails with SENDER_IS_SMART_CONTRACT. Base issue https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/3832#issuecomment-2964994308

### Steps to reproduce

Make an contract call to web3 where the from is a smart contract to reproduce.

### Additional context

_No response_

### Hedera network

_No response_

### Version

v0.131.0

### Operating system

None

kselveliev:
**Description**:

This PR adds a workaround for smart contract senders to execute contract call simulations.

Previously we failed fast with SENDER_IS_SMART_CONTRACT status but this causes problems for smart accounts.

Modularized codebase does not allow top level payer to be a smart contract.

Workaround is: if the from address is a smart contract we just assign the treasury account as a sender.

We have the same logic when the from is the zero address  -> 

```
 if (params.getSender().canonicalAddress().isZero() && params.getValue() == 0L) {
            return EntityIdUtils.toAccountId(systemEntity.treasuryAccount());
 }
```

Added `isAssociated` test with the workaround flow.

Fixes #11547 

**Checklist**

### DIFF
--- a/web3/src/test/java/org/hiero/mirror/web3/service/TransactionExecutionServiceTest.java
+++ b/web3/src/test/java/org/hiero/mirror/web3/service/TransactionExecutionServiceTest.java
@@ -11,7 +11,6 @@ import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
-import com.hedera.hapi.node.base.AccountID;
 import com.hedera.hapi.node.base.ResponseCodeEnum;
 import com.hedera.hapi.node.contract.ContractFunctionResult;
 import com.hedera.hapi.node.state.token.Account;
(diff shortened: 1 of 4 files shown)


web3/src/test/java/org/hiero/mirror/web3/service/TransactionExecutionServiceTest.java
```diff
@@ -203,29 +202,6 @@ void testExecuteContractCallInvalidSender(final Address senderAddress) {
                     .isInstanceOf(MirrorEvmTransactionException.class)
                     .hasMessage(PAYER_ACCOUNT_NOT_FOUND.name());
         }
-
-        @MockitoSettings(strictness = Strictness.LENIENT)
-        @ParameterizedTest
-        @MethodSource("invalidSenderAddress")
-        void testExecuteContractCallInvalidSenderContract(final Address senderAddress) {
```
IvanKavaldzhiev, line 210: Can't we reverse the expected behaviour of the test and keep the test, instead of deleting it? In that it will assure as a unit test that smart contract as a sender is working inside this class.

kselveliev: That's a reasonable comment, the issue that I saw was that this test resides in ->
```
 Nested
    class InvalidSenderNegativeTest {
```

While this case is now not negative.
So my aim was just to cover it with 2 positive integration tests related to realistic user scenarios for calling isAssociated function.

PR #11607, opened 2025-07-18
HIP-1081 Prioritize block nodes by latency

### LINKED ISSUE #11271: HIP-1081 Prioritize block nodes by latency
### Problem

In order to reduce block streaming latency, mirrornode should add logic to measure block nodes latency and pick the fastest block node in the priority group to stream block items from. 

### Solution

TBA

### Alternatives

_No response_

### LINKED ISSUE #11546: Add tests that check the latency prioritisation
### Problem

add tests that check the block node to mirror node communication, considering the latency of the nodes

### Solution

add tests that check the block node to mirror node communication, considering the latency of the nodes

### Alternatives

_No response_

xin-hedera:
**Description**:

This PR implements latency based block node scheduling

- Add three block node schedulers: `LATENCY`, `PRIORITY`, and `PRIORITY_THEN_LATENCY`
- Add `LatencyService` to measure block node streaming latency in background for latency aware schedulers 

**Related issue(s)**:

Fixes #11271 
Fixes #11546

**Notes for reviewer**:

**Checklist**


docs/configuration.md
```diff
@@ -44,6 +44,9 @@ value, it is recommended to only populate overridden properties in the custom `a
...
 | `hiero.mirror.importer.block.persistBytes`                                      | false                                                | Whether to persist the block stream file bytes to the database.                                                                                                                                                                                                    |
+| `hiero.mirror.importer.block.scheduling.latencySchedulingEnabled`               | true                                                 | Whether to enable latency based block node scheduling.                                                                                                                                                                                                             |
```
steven-sheehy: An enum might be better so we can plug in different strategies in the future like PRIORITY, LATENCY, PRIORITY_THEN_LATENCY, etc.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockNode.java
```diff
@@ -38,12 +38,22 @@
 @CustomLog
 final class BlockNode implements AutoCloseable, Comparable<BlockNode> {
 
-    private static final Comparator<BlockNode> COMPARATOR = Comparator.comparing(blockNode -> blockNode.properties);
+    private static final Comparator<BlockNode> PRIORITY_COMPARATOR =
+            Comparator.comparing(b -> b.properties.getPriority());
+    private static final Comparator<BlockNode> COMPARATOR = PRIORITY_COMPARATOR
```
steven-sheehy: Why do we need PRIORITY_COMPARATOR as a separate variable?

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockNode.java
```diff
@@ -105,14 +115,17 @@ public void streamBlocks(
                     request));
             SubscribeStreamResponse response;
 
-            boolean serverSuccess = false;
-            while (!serverSuccess
+            boolean shouldStop = false;
```
steven-sheehy: Could use positive terminology
```suggestion
            boolean running = true;
```

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockNodeSubscriber.java
```diff
@@ -24,13 +31,15 @@ final class BlockNodeSubscriber extends AbstractBlockSource implements AutoClose
         nodes = properties.getNodes().stream()
                 .map(blockNodeProperties ->
                         new BlockNode(channelBuilderProvider, blockNodeProperties, properties.getStream()))
-                .sorted()
-                .toList();
+                .collect(
+                        TreeMultimap::create,
+                        (map, blockNode) -> map.put(blockNode.getProperties().getPriority(), blockNode),
+                        TreeMultimap::putAll);
```
steven-sheehy: We might want different strategies that ignore priority and only look at latency. It's probably better to keep it in a list since the tree wouldn't work well for such a scenario. Performance wise, until we get to thousands of block nodes it wouldn't matter.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockNodeSubscriber.java
```diff
@@ -43,20 +52,72 @@ public void get() {
...
+        var blockFile = onBlockStream(blockStream);
+
+        var scheduling = properties.getScheduling();
+        if (!scheduling.isLatencySchedulingEnabled()) {
+            return false;
+        }
+
+        long previousPostProcessingLatency = lastPostProcessingLatency.getAndSet(System.currentTimeMillis() - start);
+        int priority = current.get().getProperties().getPriority();
+        if (previousPostProcessingLatency
```
steven-sheehy: There's a lot of logic here that violates the single responsibility principle. Can we refactor this logic into a strategy interface?

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockNodeSubscriber.java
```diff
@@ -2,116 +2,56 @@
...
     @Override
     public void close() {
-        nodes.forEach(BlockNode::close);
-        executor.shutdown();
+        scheduler.close();
     }
 
     @Override
-    protected void doGet(final long blockNumber) {
+    protected void doGet(final long blockNumber, final Long endBlockNumber) {
```
ashumahajan, line 39: endBlockNumber is used only in BlockNodeSubscriber and BlockFileSource does not even use it. Not sure of why doGet method definition should be changed when BlockNodeSubscriber already has reference to the commonDownloaderProperties. Through which endBlockNumber is extracted in AbstractBlockSource.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/Latency.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+final class Latency {
+
+    private static final int HISTORY_SIZE = 5;
+
+    @Getter
+    private long average = Long.MIN_VALUE;
+
+    private int count = 0;
+    private final long[] history = new long[HISTORY_SIZE];
```
ashumahajan: BlockingQueue might be better in this case, as we don't need to track last 5 values

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockNodeSubscriber.java
xin-hedera: just a preference so `BlockNodeSubscriber` doesn't have to get the end block number from config.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/Latency.java
xin-hedera: have a hard time to understand the feedback.

- thread-safety should be guaranteed by the caller so we don't need a blocking queue here
- we need to track the last 5 values to calculate the average. yup we can do it differently though

docs/configuration.md
```diff
@@ -44,6 +44,13 @@ value, it is recommended to only populate overridden properties in the custom `a
...
+| `hiero.mirror.importer.block.scheduler.rescheduleLatencyThreshold`              | 50ms                                                 | The threshold to meet for lower latency block nodes to trigger a reschedule.                                                                                                                                                                                       |
+| `hiero.mirror.importer.block.scheduler.type`                                    | PRIORITY                                             | The scheduler type. Can be `LATENCY`, `PRIORITY`, or `PRIORITY_THEN_LATENCY`.                                                                                                                                                                                      |
```
steven-sheehy: Shouldn't we default to PRIORITY_THEN_LATENCY? Otherwise all this latency work will never be used in practice.

docs/configuration.md
```diff
@@ -44,6 +44,13 @@ value, it is recommended to only populate overridden properties in the custom `a
...
 | `hiero.mirror.importer.block.persistBytes`                                      | false                                                | Whether to persist the block stream file bytes to the database.                                                                                                                                                                                                    |
+| `hiero.mirror.importer.block.scheduler.latencyService.backlog`                  | 1                                                    | The backlog size of pending latency measuring tasks. Note a max number of backlog plus 1 tasks can be scheduled when the service is idle.                                                                                                                          |
```
steven-sheehy: Prefer simpler `latency` for property names instead of implementation names.

```suggestion
| `hiero.mirror.importer.block.scheduler.latency.backlog`                  | 1                                                    | The backlog size of pending latency measuring tasks. Note a max number of backlog plus 1 tasks can be scheduled when the service is idle.                                                                                                                          |
```

importer/src/main/java/org/hiero/mirror/importer/downloader/block/Latency.java
steven-sheehy: This class seems like overkill and may not smooth the averages as we like. Why can't we drop the class and use a simple exponential moving average in the calling class that's stored as a double or AtomicDouble?

importer/src/main/java/org/hiero/mirror/importer/downloader/block/SchedulerProperties.java
```diff
@@ -0,0 +1,51 @@
...
+import jakarta.validation.constraints.NotNull;
+import java.time.Duration;
+import lombok.Data;
+import org.hibernate.validator.constraints.time.DurationMin;
+import org.hiero.mirror.importer.downloader.block.scheduler.SchedulerType;
+import org.springframework.validation.annotation.Validated;
+
+@Data
+@Validated
+public final class SchedulerProperties {
```
steven-sheehy: Prefer this as a separate `@ConfigurationProperties` instead of nested to reduce coupling. Also move to scheduler package.

jnels124 (dismissed):

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockNode.java
```diff
@@ -148,24 +167,31 @@ public void streamBlocks(
...
-            while (!serverSuccess && (response = grpcCall.read(assembler.timeout(), TimeUnit.MILLISECONDS)) != null) {
+            boolean running = true;
+            while (running && (response = grpcCall.read(assembler.timeout(), TimeUnit.MILLISECONDS)) != null) {
                 switch (response.getResponseCase()) {
                     case BLOCK_ITEMS -> assembler.onBlockItemSet(response.getBlockItems());
-                    case END_OF_BLOCK -> assembler.onEndOfBlock(response.getEndOfBlock());
+                    case END_OF_BLOCK -> {
+                        running = !assembler.onEndOfBlock(response.getEndOfBlock());
+                        if (!running) {
+                            log.info("Cancel the subscription to try rescheduling");
```
steven-sheehy: This log sounds like a suggestion to the operator to take some action. Do you mean to say `Cancelling the subscription...` to indicate the code is doing the action?
```suggestion
                            log.info("Cancel the subscription to try rescheduling");
```

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockNode.java
```diff
@@ -190,9 +216,13 @@ public int compareTo(final BlockNode other) {
         return COMPARATOR.compare(this, other);
     }
 
+    public synchronized void recordLatency(long latency) {
+        this.latency.record(latency);
```
steven-sheehy: To be safe, check if latency is positive so record doesn't throw an error

importer/src/main/java/org/hiero/mirror/importer/downloader/block/scheduler/SchedulerSupplier.java
```diff
@@ -0,0 +1,48 @@
...
+    private final MeterRegistry meterRegistry;
+
+    @Override
+    public Scheduler get() {
+        final var blockNodeProperties = blockProperties.getNodes();
+        final var schedulerProperties = blockProperties.getScheduler();
+        final var streamProperties = blockProperties.getStream();
+        return switch (schedulerProperties.getType()) {
+            case LATENCY ->
+                new LatencyScheduler(
```
steven-sheehy, line 29: Why do we need this class? Why can't all 3 types be beans that we inject? Can add a `SchedulerType getType()` to each so the calling class can generically map them.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/scheduler/PriorityScheduler.java
```diff
@@ -0,0 +1,34 @@
...
+    PriorityScheduler(
+            final Collection<BlockNodeProperties> blockNodeProperties,
+            final ManagedChannelBuilderProvider channelBuilderProvider,
+            final MeterRegistry meterRegistry,
+            final StreamProperties streamProperties) {
+        nodes = blockNodeProperties.stream()
+                .map(properties -> new BlockNode(
+                        channelBuilderProvider, this::drainGrpcBuffer, meterRegistry, properties, streamProperties))
+                .sorted()
+                .toList();
```
steven-sheehy: Prefer immutable list so returned iterator can't remove.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/scheduler/AbstractScheduler.java
```diff
@@ -0,0 +1,87 @@
...
+import org.slf4j.LoggerFactory;
+
+abstract class AbstractScheduler implements Scheduler {
+
+    protected final Logger log = LoggerFactory.getLogger(getClass());
+
+    private final ExecutorService executor;
+
+    protected AbstractScheduler() {
+        executor = Executors.newSingleThreadExecutor();
```
steven-sheehy, line 46: Need to wait for shutdown and close executor in close()

importer/src/main/java/org/hiero/mirror/importer/downloader/block/scheduler/AbstractScheduler.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    private static boolean hasBlock(final AtomicLong nextBlockNumber, final BlockNode node) {
+        final var blockRange = node.getBlockRange();
+        if (blockRange.isEmpty()) {
+            return false;
+        }
+
+        if (nextBlockNumber.get() == EARLIEST_AVAILABLE_BLOCK_NUMBER) {
+            nextBlockNumber.set(blockRange.lowerEndpoint());
```
steven-sheehy: I don't like that the interface can have a side effect of changing the parameter. Is there another way we can pass primitive long and move this logic higher up?

importer/src/main/java/org/hiero/mirror/importer/downloader/block/scheduler/AbstractLatencyAwareScheduler.java
```diff
@@ -0,0 +1,97 @@
...
+abstract class AbstractLatencyAwareScheduler extends AbstractScheduler {
+
+    protected final LatencyService latencyService;
+    protected final SchedulerProperties schedulerProperties;
+
+    protected final List<BlockNode> candidates = new CopyOnWriteArrayList<>();
+    protected final AtomicReference<@Nullable BlockNode> current = new AtomicReference<>();
+    protected final AtomicLong lastScheduledTime = new AtomicLong(0);
+
+    protected long lastPostProcessingLatency;
```
steven-sheehy: Can be private. Should also be atomic or volatile to be safe.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/scheduler/LatencyService.java
```diff
@@ -0,0 +1,145 @@
...
+     */
+    void setNodes(Collection<BlockNode> nodes) {
+        cancelAll();
+
+        long bornGeneration = generation.incrementAndGet();
+        nodes.forEach(blockNode -> tasks.add(new Task(bornGeneration, blockNode)));
+    }
+
+    @Scheduled(fixedDelayString = "#{@blockProperties.getScheduler().getLatencyService().getFrequency().toMillis()}")
+    public void schedule() {
```
steven-sheehy: Is it possible this takes longer than frequency and multiple invocations occur simultaneously? If so, might need synchronized.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/scheduler/AbstractLatencyAwareScheduler.java
```diff
@@ -0,0 +1,97 @@
...
+
+    @Override
+    public BlockNode getNode(final AtomicLong blockNumber) {
+        try {
+            latencyService.cancelAll();
+            current.set(super.getNode(blockNumber));
+            candidates.clear();
+            candidates.addAll(getCandidates());
+            latencyService.setNodes(candidates);
```
steven-sheehy: latencyService.cancelAll() called twice: once directly and other via setNodes.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/scheduler/LatencyService.java
```diff
@@ -0,0 +1,145 @@
...
+    /**
+     * Set the block nodes to asynchronously measure latency for
+     *
+     * @param nodes - Block nodes
+     */
+    void setNodes(Collection<BlockNode> nodes) {
+        cancelAll();
+
+        long bornGeneration = generation.incrementAndGet();
+        nodes.forEach(blockNode -> tasks.add(new Task(bornGeneration, blockNode)));
```
steven-sheehy, line 69: Is it possible to optimize to skip currently connected node since we can measure that directly elsewhere?

importer/src/main/java/org/hiero/mirror/importer/downloader/block/scheduler/LatencyService.java
```diff
@@ -0,0 +1,145 @@
...
+
+            final long nextBlockNumber = blockStreamVerifier.getNextBlockNumber();
+            if (nextBlockNumber < 0 || !node.getBlockRange().contains(nextBlockNumber)) {
+                return;
+            }
+
+            log.info("Measuring {}'s latency by streaming block {}", node, nextBlockNumber);
+            final var timeout =
+                    blockProperties.getScheduler().getLatencyService().getTimeout();
+            node.streamBlocks(nextBlockNumber, nextBlockNumber, this::measureLatency, timeout);
```
steven-sheehy: The next block might not be available. Should we stream a couple previous blocks that we know are present? This would omit time waiting for next block and spread the latency across a few blocks.

steven-sheehy (changes requested):

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockNode.java
xin-hedera: latency is allowed to be non-positive. It's measured as the block complete time (from wall clock) - consensus end. When wall clock time is too much behind, the measurement can be negative.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/scheduler/SchedulerSupplier.java
xin-hedera: Then we'll create 3 executors (in the constructor), one for each bean, and only one is used depending on the configured scheduler type.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/scheduler/LatencyService.java
xin-hedera: it's fixed delay, so it should get called after the set period since the last invocation.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/scheduler/LatencyService.java
xin-hedera: the candidates picked in AbstractLatencyAwareScheduler and set to the latency service doesn't include the current node.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/scheduler/LatencyService.java
xin-hedera: The latency is measured as wall clock time - a block's consensus end. that determines we should always try to stream the next block to measure latency. Otherwise, a lower latency node's measured latency can get artificially increased by a lot.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/Latency.java
xin-hedera: made it a separate class so we can unit test it. will change it to EMA.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/scheduler/Latency.java
```diff
@@ -0,0 +1,44 @@
...
+    private final AtomicBoolean stale = new AtomicBoolean(false);
+
+    @Override
+    public int compareTo(final Latency other) {
+        return COMPARATOR.compare(this, other);
+    }
+
+    double getAverage() {
+        return stale.get() ? Double.MAX_VALUE : average;
+    }
```
ashumahajan, line 29: record() writes average under synchronized(this), but getAverage() holds no lock. The volatile read of stale.get() establishes happens-before only for stale itself; average is written after the stale.set(false) volatile write inside record(), so the volatile ordering does not cover it. A reader thread (main scheduling thread calling shouldReschedule()) can see stale values of average written by the executor thread running Task.measureLatency().

  Fix: Make getAverage() synchronized, or declare `average` as volatile, or store the EMA as an AtomicLong using Double.doubleToLongBits / longBitsToDouble.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/scheduler/LatencyService.java
```diff
@@ -0,0 +1,157 @@
...
+                    return;
+                }
+
+                log.info("Measuring {}'s latency by streaming block {}", node, nextBlockNumber);
+                final var timeout = latencyServiceProperties.getTimeout();
+                node.streamBlocks(nextBlockNumber, nextBlockNumber, this::measureLatency, timeout);
+                lastMeasuredBlockNumber = nextBlockNumber;
+                skipped = 0;
+            } catch (Exception ex) {
+                // ignore
```
ashumahajan, line 142: I think we should atleast log the exception so failures are diagnosable

importer/src/main/java/org/hiero/mirror/importer/downloader/block/scheduler/LatencyService.java
```diff
@@ -0,0 +1,157 @@
...
+import org.springframework.scheduling.annotation.Scheduled;
+
+/**
+ * A latency service to measure a selection of block nodes' streaming latency in background. The latency measuring tasks
+ * are scheduled in fixed delay, and limited to up to 1 + backlog at a time.
+ */
+@CustomLog
+@Named
+@RequiredArgsConstructor
+public final class LatencyService implements AutoCloseable {
```
ashumahajan, line 33: No test coverage for LatencyService

importer/src/main/java/org/hiero/mirror/importer/downloader/block/scheduler/SchedulerSupplier.java
```diff
@@ -0,0 +1,49 @@
...
+import jakarta.inject.Named;
+import java.util.function.Supplier;
+import lombok.RequiredArgsConstructor;
+import org.hiero.mirror.importer.downloader.block.BlockNodeDiscoveryService;
+import org.hiero.mirror.importer.downloader.block.BlockProperties;
+import org.hiero.mirror.importer.downloader.block.ManagedChannelBuilderProvider;
+
+@Named
+@RequiredArgsConstructor
+public final class SchedulerSupplier implements Supplier<Scheduler> {
```
ashumahajan, line 15: We can annotate each scheduler class with @Bean in a @Configuration class,  conditioned on scheduler.type. This way we can remove SchedulerSupplier entirely

xin-hedera: don't think we can do conditional beans with graalvm, and believe we actually removed a lot of conditional beans for that reason, @jnels124 ?

importer/src/main/java/org/hiero/mirror/importer/downloader/block/scheduler/LatencyService.java
xin-hedera: don't think there's much value and it's mostly noise. This is used to measure latency, not in the critical path of streaming blocks.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/scheduler/LatencyService.java
```diff
@@ -0,0 +1,157 @@
...
+     * @param nodes - Block nodes
+     */
+    void setNodes(final Collection<BlockNode> nodes) {
+        cancelAll();
+
+        final long bornGeneration = generation.incrementAndGet();
+        nodes.forEach(blockNode -> tasks.add(new Task(bornGeneration, blockNode)));
+    }
+
+    @Scheduled(fixedDelayString = "#{@latencyServiceProperties.getFrequency().toMillis()}")
```
steven-sheehy, line 72: Please double check if `spring.task.scheduling.pool.size` needs to increase.

steven-sheehy (dismissed):
LGTM

jnels124 (dismissed):

importer/src/main/java/org/hiero/mirror/importer/downloader/block/scheduler/SchedulerSupplier.java
jnels124: That is correct. We can not have beans created conditionally for graal vm

importer/src/main/java/org/hiero/mirror/importer/downloader/block/scheduler/Latency.java
xin-hedera: made `average` volatile

importer/src/main/java/org/hiero/mirror/importer/downloader/block/scheduler/LatencyService.java
xin-hedera: there is integration test coverage from several mutlinode latency tests classes. Added simple unit test for the class iteself.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/scheduler/LatencyService.java
xin-hedera: bumped it up to 7

ashumahajan (approved):
LGTM

steven-sheehy (approved):

PR #11608, opened 2025-07-18
Add github workflow to test non-zero shard/realm 

### LINKED ISSUE #11540: Create Github Action to run tests with non-zero shard and realm
### Problem

Need to run periodic tests with non-zero shard and realm

### Solution

- Have a scheduled workflow to test main branch with non-zero shard and realm
- Allow selecting a different branch for manual execution

### Alternatives

_No response_

xin-hedera:
**Description**:

This PR adds a workflow to test non-zero shard/realm

- Add github workflow to test non-zero shard/realm
- Don't fail-fast in jobs with matrix

**Related issue(s)**:

Fixes #11540 

**Notes for reviewer**:

There are tests failing in web3: https://github.com/hiero-ledger/hiero-mirror-node/actions/runs/16380493136?pr=11608

**Checklist**

### DIFF
--- a/.github/workflows/non-zero-realm.yml
+++ b/.github/workflows/non-zero-realm.yml
@@ -2,9 +2,6 @@
 
 name: "Non-Zero Realm"
 on:
-  push:
-    branches:
-      - 11540-create-github-action-to-run-tests-with-non-zero-shard-and-realm
   schedule:
     - cron: "0 0 * * *" # Daily at midnight
   workflow_dispatch:


.github/workflows/nzr-test.yml
```diff
@@ -0,0 +1,92 @@
...
+          - grpc
+          - importer
+          - monitor
+          - rest
+          - rest-java
+          - rosetta
+          - web3
+        schema:
+          - v1
+          - v2
```
steven-sheehy: I don't think we need to test v2 with NZR since if it works with v1 it should with v2 since it's the same code. Can remove.

.github/workflows/nzr-test.yml
```diff
@@ -0,0 +1,92 @@
...
+          MIRROR_NODE_SCHEMA: ${{ matrix.schema}}
+          SPRING_PROFILES_ACTIVE: ${{ matrix.schema}}
+        run: ./gradlew :${{matrix.project}}:build --scan ${{ secrets.GRADLE_ARGS }}
+
+      - name: Build and test web3 with modularized code
+        env:
+          HIERO_MIRROR_WEB3_EVM_MODULARIZEDSERVICES: "true"
+          HIERO_MIRROR_WEB3_EVM_MODULARIZEDTRAFFICPERCENT: "1.0"
+        if: ${{ matrix.project == 'web3' && matrix.schema == 'v1'}}
+        run: ./gradlew :${{ matrix.project }}:build
```
steven-sheehy: I think we can just only test NZR with modularized and remove this step.

.github/workflows/nzr-test.yml
steven-sheehy, line 1: Prefer just `non-zero-realm.yml`

.github/workflows/nzr-test.yml
```diff
@@ -0,0 +1,92 @@
+# SPDX-License-Identifier: Apache-2.0
+
+name: "Test non-zero shard and realm network"
```
steven-sheehy:
```suggestion
name: "Non-Zero Realm"
```

.github/workflows/nzr-test.yml
```diff
@@ -0,0 +1,92 @@
...
+    name: Build
+    env:
+      HIERO_MIRROR_COMMON_REALM: 1
+      HIERO_MIRROR_COMMON_SHARD: 1
+    runs-on: hiero-mirror-node-linux-large
+    strategy:
+      fail-fast: false
+      matrix:
+        project:
+          - common
```
steven-sheehy: Should we simplify it to be a single threaded test without a matrix by using `org.gradle.parallel=false`?

steven-sheehy (commented):
Is it possible to quickly fix any of the test failures in this PR or do we need a separate ticket?

.github/workflows/nzr-test.yml
xin-hedera: that may take too long. not a big deal for scheduled runs, however it may be frustrating for runs triggered for a specific branch.

.github/workflows/non-zero-realm.yml
```diff
@@ -0,0 +1,74 @@
...
+        uses: step-security/harden-runner@6c439dc8bdf85cadbbce9ed30d1c7b959517bc49 # v2.12.2
+        with:
+          egress-policy: audit
+
+      - name: Checkout Code
+        if: ${{ github.event_name != 'workflow_dispatch' }}
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - name: Checkout Code from Branch
+        if: ${{ github.event_name == 'workflow_dispatch' }}
```
steven-sheehy: Is it possible to combine and have a conditional inside the `with.ref`?

.github/workflows/non-zero-realm.yml
```diff
@@ -0,0 +1,74 @@
...
+        uses: actions/setup-java@c5195efecf7bdfc987ee8bae7a71cb8b11521c00 # v4.7.1
+        with:
+          distribution: temurin
+          java-version: 21
+
+      - name: Setup Gradle
+        uses: gradle/actions/setup-gradle@ac638b010cf58a27ee6c972d7336334ccaf61c96 # v4
+
+      - name: Execute Gradle
+        run: ./gradlew :${{matrix.project}}:build --scan ${{ secrets.GRADLE_ARGS }}
```
steven-sheehy: We can probably omit the build scan for this.

```suggestion
        run: ./gradlew :${{matrix.project}}:build
```

.github/workflows/non-zero-realm.yml
xin-hedera: possible, let me confirm if the logic in the checkout action uses `github.ref` by default.

steven-sheehy (approved):
LGTM

PR #11627, opened 2025-07-21
Set up environment for block node to mirror node communication tests

nickeynikolovv:
**Description**:
This PR adds set up for an environment that can be used for testing block node communication to mirror node

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/importer/src/test/java/org/hiero/mirror/importer/blockNode/BaseClass.java
+++ b/importer/src/test/java/org/hiero/mirror/importer/blockNode/BaseClass.java
@@ -1,19 +1,18 @@
+// SPDX-License-Identifier: Apache-2.0
+
 package org.hiero.mirror.importer.blockNode;
 
 import io.grpc.ManagedChannel;
 import io.grpc.ManagedChannelBuilder;
-import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import org.apache.commons.lang3.StringUtils;
 import org.hiero.block.api.protoc.BlockAccessServiceGrpc;
 import org.hiero.block.api.protoc.BlockNodeServiceGrpc;
 import org.hiero.mirror.common.config.CommonTestConfiguration.FilteringConsumer;
 import org.junit.jupiter.api.AfterEach;
-import org.junit.jupiter.api.BeforeEach;
 import org.slf4j.LoggerFactory;
 import org.testcontainers.containers.BindMode;
 import org.testcontainers.containers.GenericContainer;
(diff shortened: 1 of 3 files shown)


importer/src/test/java/org/hiero/mirror/importer/blockNode/BaseClass.java
```diff
@@ -0,0 +1,267 @@
...
+import org.testcontainers.images.builder.ImageFromDockerfile;
+import org.testcontainers.utility.DockerImageName;
+
+public class BaseClass {
+    protected static GenericContainer<?> blockNodeContainer;
+    protected static PostgreSQLContainer<?> dbContainer;
+    protected static GenericContainer<?> importerContainer;
+    protected static GenericContainer<?> simulatorContainer;
+    protected static GenericContainer<?> blockNodeSecondContainer;
+    protected static GenericContainer<?> simulatorSecondContainer;
```
steven-sheehy, line 32: This is way too many containers. We should only really need the simulator container and then use regular classes like BlockNodeSubscriber to connect to it. We already have a db container in integration tests which we would extend.

importer/src/test/java/org/hiero/mirror/importer/blockNode/BaseClass.java
```diff
@@ -0,0 +1,267 @@
...
+    protected static int blockNodeMetricsPort;
+
+    /** gRPC channel for connecting to Block Node */
+    protected static ManagedChannel channel;
+
+    /** gRPC client stub for BlockAccessService */
+    protected static BlockAccessServiceGrpc.BlockAccessServiceBlockingStub blockAccessStub;
+
+    /** gRPC client stub for BlockNodeService */
+    protected static BlockNodeServiceGrpc.BlockNodeServiceBlockingStub blockServiceStub;
```
steven-sheehy, line 51: We shouldn't be invoking stubs directly, it should be done by our existing code in src/main/java.

importer/src/test/java/org/hiero/mirror/importer/blockNode/Tests.java
```diff
@@ -0,0 +1,32 @@
...
+    @Test
+    @DisplayName("Test")
+    void checkImporterConnectionToBlockNode() throws InterruptedException {
+        var network = Network.newNetwork();
+        var dbContainer = createDBContainer(network);
+        dbContainer.start();
+        var blockNodeContainer = createBlockNodeContainer(network);
+        blockNodeContainer.start();
+        var simulatorContainer = createSimulatorContainer(network);
+        simulatorContainer.start();
```
steven-sheehy, line 22: Did we see what knobs are available to adjust the simulator? It looks like it's only a basic configuration file and rest is file based block. If so, that's not really going to work for our needs. 

We need more fine-grained control of the block node to mirror node communication. Probably we need to stick with the approach in `BlockNodeSubscriberTest` but extract it to a separate `BlockNodeServer`. It would have methods to setup expected interactions and ideally allow swapping of the protocol layer from in-process to socket.

PR #11632, opened 2025-07-21
Add gas used metric to web3 hpa

### LINKED ISSUE #11596: Enhance Web3 HPA
### Problem

Several k6 tests can be improved with more effective hpa scaling

### Solution

Add scaling logic to scale based on gas used

### Alternatives

_No response_

jnels124:
**Description**:
* add gas used metric to web3 hpa

**Related issue(s)**:

Fixes #11596 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/charts/hedera-mirror-common/values.yaml
+++ b/charts/hedera-mirror-common/values.yaml
@@ -234,6 +234,17 @@ prometheus-adapter:
         name:
           as: "requests_per_second"
         metricsQuery: "sum(rate(<<.Series>>{<<.LabelMatchers>>}[3m])) by (<<.GroupBy>>)"
+      - seriesQuery: 'hiero_mirror_web3_evm_gas_used_total{application="web3",namespace!="",pod!=""}'
+        resources:
+          overrides:
+            namespace:
+              resource: "namespace"
+            pod:
+              resource: "pod"
+        name:
+          matches: "hiero_mirror_web3_evm_gas_used_total"
+          as: "evm_gas_used_per_second"
+        metricsQuery: 'sum by(namespace,pod)(rate(hiero_mirror_web3_evm_gas_used_total{application="web3",namespace!="",pod!=""}[3m]))'
 
 prometheus:
   alertmanager:
--- a/charts/hedera-mirror-common/values.yaml
+++ b/charts/hedera-mirror-common/values.yaml
@@ -242,9 +242,8 @@ prometheus-adapter:
             pod:
               resource: "pod"
         name:
-          matches: "hiero_mirror_web3_evm_gas_used_total"
           as: "evm_gas_used_per_second"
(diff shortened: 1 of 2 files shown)


charts/hedera-mirror-web3/values.yaml
```diff
@@ -104,6 +104,13 @@ hpa:
         target:
           type: Utilization
           averageUtilization: 80
+    - type: Pods
+      pods:
+        metric:
+          name: evm_gas_used_per_second
+        target:
+          type: AverageValue
+          averageValue: 7M
```
steven-sheehy: Is this number what we see per pod during k6 runs?

charts/hedera-mirror-common/values.yaml
```diff
@@ -234,6 +234,17 @@ prometheus-adapter:
...
+        resources:
+          overrides:
+            namespace:
+              resource: "namespace"
+            pod:
+              resource: "pod"
+        name:
+          matches: "hiero_mirror_web3_evm_gas_used_total"
+          as: "evm_gas_used_per_second"
+        metricsQuery: 'sum by(namespace,pod)(rate(hiero_mirror_web3_evm_gas_used_total{application="web3",namespace!="",pod!=""}[3m]))'
```
steven-sheehy: We should be able to use `sum(rate(<<.Series>>{<<.LabelMatchers>>}[3m])) by (<<.GroupBy>>)` to stay DRY

charts/hedera-mirror-common/values.yaml
```diff
@@ -234,6 +234,17 @@ prometheus-adapter:
...
         metricsQuery: "sum(rate(<<.Series>>{<<.LabelMatchers>>}[3m])) by (<<.GroupBy>>)"
+      - seriesQuery: 'hiero_mirror_web3_evm_gas_used_total{application="web3",namespace!="",pod!=""}'
+        resources:
+          overrides:
+            namespace:
+              resource: "namespace"
+            pod:
+              resource: "pod"
+        name:
+          matches: "hiero_mirror_web3_evm_gas_used_total"
```
steven-sheehy: Is this needed?

charts/hedera-mirror-web3/values.yaml
jnels124: the value here should allow pods to scale to max hpa during k6 test rampup without causing excessive scaling for mainnet traffic. The amount of gas used is much larger than this value when VUS gets to 500.

charts/hedera-mirror-common/values.yaml
jnels124: we can do that but figured would be explicit since this metric is not reused by additional apps. I can update to use the template approach

charts/hedera-mirror-common/values.yaml
jnels124: I will double check. That was added as I was following examples in documentation

jnels124: Looks like this will not be needed when I switch to the templated format requested in another comment

steven-sheehy (approved):

PR #11639, opened 2025-07-22
Add contract nonce check to Acceptance testsNonce

### LINKED ISSUE #6395: HIP-729 Add contract nonce check to Acceptance tests
### Problem

As per HIP-729 contract nonces have been introduced and so we are recording its value in entity.nonce.
Currently the acceptance tests do not reflect that.

### Solution

- Add nonce check to contract acceptance tests.

### Alternatives

_No response_

ashumahajan:
**Description**:
Add contract nonce check to Acceptance tests

* Added nonce property to Contract object in openapi.yml
* Added nonceVal in ContractFeature to track nonce value within the test scenario
* Added steps to verify parent nonce 

**Related issue(s)**:

Fixes #6395 

**Checklist**

### DIFF
--- a/rest/api/v1/openapi.yml
+++ b/rest/api/v1/openapi.yml
@@ -2002,6 +2002,12 @@ components:
         memo:
           example: contract memo
           type: string
+        nonce:
+          description: The nonce of the wrapped ethereum transaction
+          nullable: true
+          format: int64
+          type: integer
+          example: 1
         obtainer_id:
           $ref: "#/components/schemas/EntityId"
         permanent_removal:
--- a/test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/BaseContractFeature.java
+++ b/test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/BaseContractFeature.java
@@ -29,6 +29,7 @@ public abstract class BaseContractFeature extends AbstractFeature {
         assertThat(address).isNotBlank().isNotEqualTo("0x").isNotEqualTo("0x0000000000000000000000000000000000000000");
         assertThat(mirrorContract.getTimestamp()).isNotNull();
         assertThat(mirrorContract.getTimestamp().getFrom()).isNotNull();
+        assertThat(mirrorContract.getNonce()).isNotNull();
 
         if (contractClient
                 .getSdkClient()
--- a/test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/ContractFeature.java
+++ b/test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/ContractFeature.java
@@ -60,6 +60,7 @@ public class ContractFeature extends BaseContractFeature {
     private String create2ChildContractEntityId;
     private AccountId create2ChildContractAccountId;
     private ContractId create2ChildContractContractId;
+    private Long nonceVal = 0L;
 
     @Value("classpath:solidity/artifacts/contracts/Parent.sol/Parent.json")
     private Resource parentContract;
@@ -283,6 +284,12 @@ public class ContractFeature extends BaseContractFeature {
         assertNotEquals(ACCOUNT_EMPTY_KEYLIST, mirrorAccountResponse.getKey().getKey());
     }
 
+    @And("the mirror node Rest API should verify that the parent contract have correct nonce")
+    public void verifyContractNonce() {
+        var mirrorContract = verifyContractFromMirror(false);
+        assertThat(mirrorContract.getNonce()).isEqualTo(nonceVal);
+    }
+
     @When("I successfully delete the child contract by calling it and causing it to self destruct")
     public void deleteChildContractUsingSelfDestruct() {
         executeSelfDestructTransaction();
@@ -291,7 +298,8 @@ public class ContractFeature extends BaseContractFeature {
     @Override
     protected ContractResponse verifyContractFromMirror(boolean isDeleted) {
         var mirrorContract = super.verifyContractFromMirror(isDeleted);
-
+        // initial nonce value for the parent contract
+        nonceVal = mirrorContract.getNonce();
         assertThat(mirrorContract.getAdminKey()).isNotNull();
         assertThat(mirrorContract.getAdminKey().getKey())
                 .isEqualTo(contractClient
@@ -351,6 +359,8 @@ public class ContractFeature extends BaseContractFeature {
                 new ContractFunctionParameters().addUint256(BigInteger.valueOf(transferAmount));
 
         executeContractCallTransaction(deployedParentContract.contractId(), "createChild", parameters, null);
+        // increment expected nonce value since a child contract is created
+        nonceVal++;
     }
 
     private ExecuteContractResult executeGetChildContractBytecodeTransaction() {
@@ -370,7 +380,11 @@ public class ContractFeature extends BaseContractFeature {
                 .addBytes(childContractBytecodeFromParent)
                 .addUint256(BigInteger.valueOf(salt));
 
-        return executeContractCallTransaction(deployedParentContract.contractId(), "create2Deploy", parameters, null);
+        ExecuteContractResult create2Deploy =
+                executeContractCallTransaction(deployedParentContract.contractId(), "create2Deploy", parameters, null);
+        // increment expected nonce value since a child contract is created
+        nonceVal++;
+        return create2Deploy;
     }
 
     // This is a function call on the CREATE2 created child contract, not the parent.
--- a/test/src/test/resources/features/contract/contract.feature
+++ b/test/src/test/resources/features/contract/contract.feature
@@ -1,51 +1,52 @@
 @contractbase @fullsuite
 Feature: Contract Base Coverage Feature
 
-    @critical @release @acceptance
-    Scenario Outline: Validate Contract Flows
+  @critical @release @acceptance
+  Scenario Outline: Validate Contract Flows
 
-        Using the Java SDK and the mirror node REST API, create, update and call a smart contract and
-        verify proper operation. Use that same parent contract to create a second child contract using
-        the CREATE2 flow, namely by retrieving the child contract bytecode and using it, and a salt value,
-        to compute the child's EVM address and then deploying the contract to that address.
+  Using the Java SDK and the mirror node REST API, create, update and call a smart contract and
+  verify proper operation. Use that same parent contract to create a second child contract using
+  the CREATE2 flow, namely by retrieving the child contract bytecode and using it, and a salt value,
+  to compute the child's EVM address and then deploying the contract to that address.
 
-        Once the child EVM address is known, issue a crypto transfer transaction to that EVM address, creating
-        a hollow account. Verify the account exists and indeed is hollow, but that a contract does not yet
-        exist at that EVM address. Then, using a parent contract function, deploy the child contract using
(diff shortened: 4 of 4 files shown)


test/src/test/resources/features/contract/contract.feature
```diff
@@ -40,6 +40,7 @@ Feature: Contract Base Coverage Feature
         Then the mirror node REST API should return status <httpStatusCode> for the contract transaction
         And the mirror node REST API should retrieve the child contract when using evm address
         And the mirror node REST API should verify the account is no longer hollow
+        And the mirror node Rest API should verify that the parent contract have correct nonce
```
jnels124: nit: `the mirror node Rest API should verify the parent contract has correct nonce`

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/ContractFeature.java
```diff
@@ -291,7 +298,8 @@ public void deleteChildContractUsingSelfDestruct() {
     @Override
     protected ContractResponse verifyContractFromMirror(boolean isDeleted) {
         var mirrorContract = super.verifyContractFromMirror(isDeleted);
-
+        // initial nonce value for the parent contract
+        nonceVal = mirrorContract.getNonce();
```
jnels124: should we set this in the first step where we create the contract?

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/BaseContractFeature.java
jnels124, line 1: We should consider adding the nonce to the DeployedContract record here

test/src/test/resources/features/contract/contract.feature
jnels124, line 1: I believe we also need to verify this information in the ethereum.feature

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/BaseContractFeature.java
jnels124: If we do this it will be easier to verify the correct nonce below and we can remove the extra step added to the ContractFeature.java

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/ContractFeature.java
ashumahajan: I added this there so that I don't make another api call in the testcase. This is the first time we get contract info from the mirrornode

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/BaseContractFeature.java
ashumahajan: I am not sure, how can we add nonce value to the DeployedContract record. I thought nonce is added by the EVM

rest/api/v1/openapi.yml
```diff
@@ -2002,6 +2002,12 @@ components:
         memo:
           example: contract memo
           type: string
+        nonce:
+          description: The nonce of the wrapped ethereum transaction
```
xin-hedera:
```suggestion
          description: The nonce of the contract
```

rest/api/v1/openapi.yml
```diff
@@ -2002,6 +2002,12 @@ components:
...
+          nullable: true
```
xin-hedera, line 2007: it shouldn't be null

```suggestion
```

test/src/test/resources/features/contract/contract.feature
ashumahajan: As per my understanding nonce value is incremented with deploying child contract by calling CREATE and CREATE2 functions. Should we add CREATE and CREATE2 functions call in ethereum.feature as well?

xin-hedera: ideally we should try to verify all contract nonces, regardless of parent or child.

ashumahajan: @xin-hedera Do you mean verify child contract nonce when they are created? I can check if nonce is non-null. I didn't see the test case where a child contract is used to create its child contract.

xin-hedera: contract nonce is always non-null and starts from 1. I just mean instead of checking the nonce increases when a contract creates a child, we could just check the nonce wherever possible.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/BaseContractFeature.java
jnels124: I just mean the DeployedContract defined in the BaseContractFeature.java. Maybe can add a constructor to the record and have logic to retrieve the nonce from mirrornode and set it if the value passed in is null.

test/src/test/resources/features/contract/contract.feature
jnels124: we have createChild calls that occur in the EthereumFeature.java and I believe these should increase the nonce as well.

rest/api/v1/openapi.yml
ashumahajan: I put this nullable: true, since the change has not made changes to contracts created before the fix and thus those contracts will have null value. Please confirm if that is incorrect assumption, I will change it.

test/src/test/resources/features/contract/contract.feature
ashumahajan: Thanks @jnels124 ! Will do so.

rest/api/v1/openapi.yml
xin-hedera: when the consensus node implemented contract nonce, the nonce is added to `ContractFunctionResult`: https://github.com/hashgraph/hedera-protobufs/blob/8c27786cec93abab974309074feaef9b48a695b7/services/contract_types.proto#L238

that's for both newly created contract and when a contract creates a child contract so its nonce increases by 1.

So nonce should not be null.

xin-hedera: I see the data in our db contradicts with what I understand, I'll research more.

xin-hedera: yup you are right old contracts before adding contract nonce all have null nonce values.

test/src/test/resources/features/contract/contract.feature
ashumahajan: @jnels124 Made changes as requested

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/BaseContractFeature.java
ashumahajan: The initial value for the nonce from the contract mirrornode api in method verifyContractFromMirror, which is executed before creation of the child

rest/api/v1/openapi.yml
ashumahajan: Updated as per suggestion

test/src/test/resources/features/contract/contract.feature
ashumahajan: @jnels124 I also added testcase to verify child contract nonce

test/src/test/resources/features/contract/contract.feature
ashumahajan: updated as per comment

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/ContractFeature.java
```diff
@@ -291,7 +299,7 @@ public void deleteChildContractUsingSelfDestruct() {
     @Override
     protected ContractResponse verifyContractFromMirror(boolean isDeleted) {
         var mirrorContract = super.verifyContractFromMirror(isDeleted);
-
+        // initial nonce value for the parent contract
```
jnels124: nit: remove

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/EthereumFeature.java
```diff
@@ -94,12 +96,24 @@ public void verifyDeployedContractMirror() {
         verifyContractExecutionResultsByTransactionId();
     }
 
+    @And("the mirror node Rest API should verify the ethereum parent contract has correct nonce")
```
jnels124: nit:  `the mirror node Rest API should verify the contracts have correct nonce`

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/ContractFeature.java
```diff
@@ -283,6 +285,12 @@ public void verifyMirrorAPIFullAccountResponse() {
         assertNotEquals(ACCOUNT_EMPTY_KEYLIST, mirrorAccountResponse.getKey().getKey());
     }
 
+    @And("the mirror node Rest API should verify the parent contract has correct nonce")
+    public void verifyContractNonce() {
+        verifyNonceForParentContract();
```
jnels124, line 290: remove duplicate call

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/BaseContractFeature.java
```diff
@@ -91,4 +96,23 @@ protected void verifyContractExecutionResults(ContractResult contractResult) {
...
                 .isEqualTo(FeatureInputHandler.evmAddress(deployedParentContract.contractId()));
     }
+
+    protected void verifyNonceForParentContract() {
+        var mirrorContract = verifyContractFromMirror(false);
+        assertThat(mirrorContract.getNonce()).isEqualTo(nonceVal);
+    }
+
+    protected void verifyNonceForChildContracts() {
+        childContracts.stream().forEach(x -> {
```
xin-hedera: nit,

```java
  var nonces = childContracts.stream()
    .map(mirrorClient::getContractInfo)
    .map(ContractResponse::getNonce)
    .toList();
  assertThat(nonces).containsOnly(1L); 
```

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/EthereumFeature.java
ashumahajan: Updated

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/ContractFeature.java
```diff
@@ -350,7 +357,13 @@ private void executeCreateChildTransaction(int transferAmount) {
         ContractFunctionParameters parameters =
                 new ContractFunctionParameters().addUint256(BigInteger.valueOf(transferAmount));
 
-        executeContractCallTransaction(deployedParentContract.contractId(), "createChild", parameters, null);
+        ExecuteContractResult executeContractResult =
+                executeContractCallTransaction(deployedParentContract.contractId(), "createChild", parameters, null);
+        String childAddress = executeContractResult.contractFunctionResult().getAddress(0);
+
+        // add contract Id to the list for verification of nonce on mirror node
+        assertThat(deployedParentContract.contractId().toEvmAddress()).isNotEqualTo(childAddress);
```
xin-hedera: why would the parent contract id equal to `childAddress`? Or what we get is not the child addess?

xin-hedera (approved):
LGTM

steven-sheehy (approved):

PR #11645, opened 2025-07-23
Fix race condition in async java migrations

### LINKED ISSUE #11643: Race condition in async java migrations
### Description

There is a potential race condition in our async java migrations:

- flyway will insert a migration info row to `flyway_schema_history` table after `JavaMigration.migrate()` completes without throwing exception
- in our async java migration, the `migrateAsync` is kicked off in `JavaMigration.migrate()` and scheduled onto a shared single threaded executor; and upon completion, `onSuccess()` will try to update the checksum of the latest migration info row to the value set in `application.yaml` or the default value 1

The `flyway_schema_history` insert and update operations run in different threads and break the expectation that the row update operation in `onSuccess()` should run strictly after the insert operation in flyway core.

### Steps to reproduce

as described

### Additional context

_No response_

### Hedera network

other

### Version

v0.134.0

### Operating system

None

xin-hedera:
**Description**:

This PR fixes the race condition in async java migrations:

- Run async part of async java migrations after flyway migrate operation completes

**Related issue(s)**:

Fixes #11643 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/importer/src/main/java/org/hiero/mirror/importer/migration/AsyncJavaMigration.java
+++ b/importer/src/main/java/org/hiero/mirror/importer/migration/AsyncJavaMigration.java
@@ -12,6 +12,9 @@ import java.util.Optional;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import org.apache.commons.lang3.BooleanUtils;
+import org.flywaydb.core.api.callback.Callback;
+import org.flywaydb.core.api.callback.Context;
+import org.flywaydb.core.api.callback.Event;
 import org.springframework.dao.EmptyResultDataAccessException;
 import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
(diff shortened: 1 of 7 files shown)


importer/src/main/java/org/hiero/mirror/importer/migration/AsyncJavaMigration.java
```diff
@@ -118,10 +124,15 @@ public Integer getChecksum() {
         return lastChecksum;
     }
 
-    protected abstract TransactionOperations getTransactionOperations();
+    @Override
+    public void handle(Event event, Context context) {
+        if (event != Event.AFTER_MIGRATE_OPERATION_FINISH || !shouldMigrate.get()) {
```
xin-hedera, line 129: flyway emits the `AFTER_MIGRATE_OPERATION_FINISH` event after the migrate operation finishes *successfully*, i.e., if any migration determined to apply fails, flyway will not emit such event.

steven-sheehy (approved):

jnels124 (approved):

PR #11654, opened 2025-07-24
Update TestUtil to use toEvmAddress() from SDK

### LINKED ISSUE #11646: Update TestUtil to use sdk toEvmAddress method
### Problem

We had to make updates to TestUtil in test module for deprecated skd `toSolidityAddress` method.

```
 /*
       When toSolidityAddress() is fixed upstream to not encode shard & realm
       we can change below asAddress methods to something like this.

       final var address = accountId.getAccountId().toSolidityAddress();
       return asAddress(address);

       Possibly some .replace("0x", "") can be removed after the switch back to toSolidityAddress()
       getClientAddress() - can be changed as well
       accountAmount(), nftAmount(), asAddressArray() - places where these are used probably should be reverted as to not have to go through asAddress 2 times
    */
```

### Solution

In current v.2.60 version of sdk we can now use their new `toEvmAddress` method.
https://github.com/hiero-ledger/hiero-sdk-java/releases/tag/v2.60.0

Update testUtil to use toEvmAddress.

### Alternatives

_No response_

kselveliev:
**Description**:

This PR adapts acceptance tests to use newly added `toEvmAddress()` method from latest java sdk.
Reverts workarounds from https://github.com/hiero-ledger/hiero-mirror-node/pull/11292 and use `toEvmAddress()` instead of deprecated `toSolidityAddress()`.

Fixes #11646

**Checklist**

### DIFF
--- a/test/src/test/java/org/hiero/mirror/test/e2e/acceptance/client/AbstractNetworkClient.java
+++ b/test/src/test/java/org/hiero/mirror/test/e2e/acceptance/client/AbstractNetworkClient.java
@@ -3,7 +3,6 @@
 package org.hiero.mirror.test.e2e.acceptance.client;
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.hiero.mirror.test.e2e.acceptance.util.TestUtil.asHexAddress;
 
 import com.hedera.hashgraph.sdk.AccountBalanceQuery;
 import com.hedera.hashgraph.sdk.AccountId;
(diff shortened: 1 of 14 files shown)


bilyana-gospodinova (approved):

jnels124 (approved):
Maybe we can add `HEX_PREFIX` constant to test util in use in places where we have `0x`. There is a couple tests to modify to get to work with non-zero realm where I would like to use the value as well.

Issue #3162, opened 2022-01-17
HIP-32 Alias on all transactions


steven-sheehy:
### Problem

The first phase of [HIP-32](https://hips.hedera.com/HIP/hip-32.html) Auto Account Creation added alias to crypto transfer transactions only. In the next phase, alias will be added to all transactions that take an `AccountID` as input.

### Solution

* Handle `alias` updates on `CryptoUpdate`
* Replace any uses of `EntityId.of(AccountID)` that pulled from `TransactionBody` with `EntityIdService.lookup(AccountID)`
* Deprecate `EntityId.of(AccountID)` as we want to ensure the service is used going forward and also because we want to remove protobuf dependencies from common module.

steven-sheehy:
Blocked on services implementation

hendrikebbers:
@steven-sheehy is this a good first issue that a newcomer could solve? If yes, can you help us by following our new guidelines at https://github.com/hiero-ledger/governance/blob/main/guidelines/good-first-issues.md

YYuan02:
@steven-sheehy @hendrikebbers Hi, I would like to work on this issue. Can you assign to me?

steven-sheehy:
Hi @YYuan02. Sure I've assigned it to you. This could impact a lot of code so you might want to do it in smaller chunks to make reviews easier.

PR #11658, opened 2025-07-24
Remove use of lazy

### LINKED ISSUE #11631: Remove use of `@Lazy`
### Problem

Spring's `@Lazy` annotation does not work with `bootBuildImage` or with GraalVM native image.

### Solution

* Remove use of `@Lazy` everywhere
* Use `ObjectProvider` wrapper around `JdbcOperations`, repositories, etc to provide programmatic creation and retrieval of troublesome dependencies
* Add `protected final NamedParameterJdbcoperations getJdbcOperations()` to `AsyncJavaMigration` that hides use of `ObjectProvider`

### Alternatives

_No response_

HamzaElzarw-2022:
**Description**:

This PR addresses compatibility issues with Spring's @Lazy annotation when using bootBuildImage and GraalVM native image generation.

**Related issue(s)**:

Fixes #11631

### DIFF
--- a/importer/src/main/java/org/hiero/mirror/importer/migration/AsyncJavaMigration.java
+++ b/importer/src/main/java/org/hiero/mirror/importer/migration/AsyncJavaMigration.java
@@ -12,7 +12,9 @@ import java.util.Optional;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import org.apache.commons.lang3.BooleanUtils;
+import org.springframework.beans.factory.ObjectProvider;
 import org.springframework.dao.EmptyResultDataAccessException;
+import org.springframework.jdbc.core.JdbcTemplate;
 import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
 import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
 import org.springframework.jdbc.core.namedparam.SqlParameterSource;
@@ -60,17 +62,29 @@ abstract class AsyncJavaMigration<T> extends RepeatableMigration {
             where f.installed_rank = last.installed_rank
             """;
 
-    protected final NamedParameterJdbcTemplate namedParameterJdbcTemplate;
+    private final ObjectProvider<JdbcTemplate> jdbcTemplateProvider;
+    private NamedParameterJdbcTemplate namedParameterJdbcTemplate;
     private final String schema;
     private final AtomicBoolean complete = new AtomicBoolean(false);
 
     protected AsyncJavaMigration(
             Map<String, MigrationProperties> migrationPropertiesMap,
-            NamedParameterJdbcTemplate namedParameterJdbcTemplate,
+            ObjectProvider<JdbcTemplate> jdbcTemplateProvider,
             String schema) {
         super(migrationPropertiesMap);
-        this.namedParameterJdbcTemplate = namedParameterJdbcTemplate;
+        this.jdbcTemplateProvider = jdbcTemplateProvider;
         this.schema = schema;
+    }
+
+    protected final JdbcTemplate getJdbcTemplate() {
+        return jdbcTemplateProvider.getObject();
+    }
+
+    protected final NamedParameterJdbcTemplate getNamedParameterJdbcTemplate() {
+        if (namedParameterJdbcTemplate == null) {
+            namedParameterJdbcTemplate = new NamedParameterJdbcTemplate(getJdbcTemplate());
+        }
+        return namedParameterJdbcTemplate;
     }
 
     /**
@@ -126,7 +140,7 @@ abstract class AsyncJavaMigration<T> extends RepeatableMigration {
 
     public <O> O queryForObjectOrNull(String sql, SqlParameterSource paramSource, Class<O> requiredType) {
         try {
-            return namedParameterJdbcTemplate.queryForObject(sql, paramSource, requiredType);
+            return getNamedParameterJdbcTemplate().queryForObject(sql, paramSource, requiredType);
         } catch (EmptyResultDataAccessException ex) {
             return null;
         }
@@ -201,13 +215,13 @@ abstract class AsyncJavaMigration<T> extends RepeatableMigration {
     }
 
     private boolean hasFlywaySchemaHistoryTable() {
-        var exists = namedParameterJdbcTemplate.queryForObject(
+        var exists = getNamedParameterJdbcTemplate().queryForObject(
                 CHECK_FLYWAY_SCHEMA_HISTORY_EXISTENCE_SQL, Map.of("schema", schema), Boolean.class);
         return BooleanUtils.isTrue(exists);
     }
 
     private boolean isAsyncJavaMigrationHistoryFixed() {
-        var fixed = namedParameterJdbcTemplate
+        var fixed = getNamedParameterJdbcTemplate()
                 .getJdbcTemplate()
                 .queryForObject(ASYNC_JAVA_MIGRATION_HISTORY_FIXED, Boolean.class);
         return BooleanUtils.isTrue(fixed);
@@ -215,7 +229,7 @@ abstract class AsyncJavaMigration<T> extends RepeatableMigration {
 
     private void onSuccess() {
         var paramSource = getSqlParamSource().addValue("checksum", getSuccessChecksum());
-        namedParameterJdbcTemplate.update(UPDATE_CHECKSUM_SQL, paramSource);
+        getNamedParameterJdbcTemplate().update(UPDATE_CHECKSUM_SQL, paramSource);
     }
 
     @VisibleForTesting
--- a/importer/src/main/java/org/hiero/mirror/importer/migration/BackfillAndDeduplicateBalanceMigration.java
+++ b/importer/src/main/java/org/hiero/mirror/importer/migration/BackfillAndDeduplicateBalanceMigration.java
@@ -16,13 +16,12 @@ import org.hiero.mirror.importer.config.Owner;
 import org.hiero.mirror.importer.db.DBProperties;
 import org.hiero.mirror.importer.db.TimePartitionService;
 import org.hiero.mirror.importer.exception.InvalidDatasetException;
+import org.springframework.beans.factory.ObjectProvider;
 import org.springframework.boot.convert.DurationStyle;
-import org.springframework.context.annotation.Lazy;
 import org.springframework.context.annotation.Profile;
 import org.springframework.jdbc.core.JdbcTemplate;
 import org.springframework.jdbc.core.namedparam.EmptySqlParameterSource;
 import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
-import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
 import org.springframework.jdbc.datasource.DataSourceTransactionManager;
 import org.springframework.transaction.support.TransactionOperations;
 import org.springframework.transaction.support.TransactionTemplate;
@@ -128,9 +127,7 @@ public class BackfillAndDeduplicateBalanceMigration extends AsyncJavaMigration<L
     private static final String SELECT_LAST_CONSENSUS_TIMESTAMP_SQL =
             "select balance from account_balance_old where consensus_timestamp = -1 and account_id = -1";
 
-    private final JdbcTemplate jdbcTemplate;
-
-    private final TimePartitionService timePartitionService;
+    private final ObjectProvider<TimePartitionService> timePartitionServiceProvider;
 
     @Getter(lazy = true)
     private final TransactionOperations transactionOperations = transactionOperations();
@@ -138,18 +135,16 @@ public class BackfillAndDeduplicateBalanceMigration extends AsyncJavaMigration<L
     private Long lastConsensusTimestamp;
     private long minFrequency;
 
-    @Lazy
     public BackfillAndDeduplicateBalanceMigration(
             DBProperties dbProperties,
             ImporterProperties importerProperties,
-            @Owner JdbcTemplate jdbcTemplate,
-            TimePartitionService timePartitionService) {
+            @Owner ObjectProvider<JdbcTemplate> jdbcTemplateProvider,
+            ObjectProvider<TimePartitionService> timePartitionServiceProvider) {
         super(
                 importerProperties.getMigration(),
-                new NamedParameterJdbcTemplate(jdbcTemplate),
+                jdbcTemplateProvider,
                 dbProperties.getSchema());
-        this.jdbcTemplate = jdbcTemplate;
-        this.timePartitionService = timePartitionService;
+        this.timePartitionServiceProvider = timePartitionServiceProvider;
     }
 
     @Override
@@ -171,7 +166,7 @@ public class BackfillAndDeduplicateBalanceMigration extends AsyncJavaMigration<L
             return NO_BALANCE;
         }
 
-        return jdbcTemplate.queryForObject(SELECT_INITIAL_CONSENSUS_TIMESTAMP_SQL, Long.class, lastConsensusTimestamp);
+        return getJdbcTemplate().queryForObject(SELECT_INITIAL_CONSENSUS_TIMESTAMP_SQL, Long.class, lastConsensusTimestamp);
     }
 
     @Override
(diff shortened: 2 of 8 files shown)


importer/src/test/java/org/hiero/mirror/importer/migration/FixCryptoAllowanceAmountMigrationTest.java
```diff
@@ -268,7 +275,12 @@ void skipMigration(boolean trackAllowance) {
         var entityProps = new EntityProperties(new SystemEntity(CommonProperties.getInstance()));
         entityProps.getPersist().setTrackAllowance(trackAllowance);
         var allowanceAmountMigration = new FixCryptoAllowanceAmountMigration(
-                dbProperties, entityProps, new ImporterProperties(), ownerJdbcTemplate);
+                dbProperties, entityProps, new ImporterProperties(), new ObjectProvider<JdbcTemplate>() {
+            @Override
+            public JdbcTemplate getObject() {
+                return ownerJdbcTemplate;
+            }
+        });
```
steven-sheehy: Since it appears twice and is now more complicated to construct, can we move this to a helper method `createMigration()`?

importer/src/main/java/org/hiero/mirror/importer/migration/SyntheticCryptoTransferApprovalMigration.java
```diff
@@ -151,22 +151,24 @@ join contractresults cr using (consensus_timestamp, payer_account_id),
...
     public SyntheticCryptoTransferApprovalMigration(
             DBProperties dbProperties,
-            RecordFileRepository recordFileRepository,
+            ObjectProvider<RecordFileRepository> recordFileRepositoryProvider,
             ImporterProperties importerProperties,
-            NamedParameterJdbcTemplate jdbcTemplate,
-            TransactionOperations transactionOperations) {
-        super(importerProperties.getMigration(), jdbcTemplate, dbProperties.getSchema());
-        this.recordFileRepository = recordFileRepository;
+            @Owner ObjectProvider<JdbcTemplate> jdbcTemplateProvider,
```
steven-sheehy: We shouldn't need the owner db connection since we didn't use it previously.

HamzaElzarw-2022:
Hi @steven-sheehy,
I'm continuing to remove Lazy from all the classes. Could you please approve the workflow so I can check the build status and ensure everything is working as expected?

I'd also really appreciate it if you could take a quick look at the code and let me know if the refactoring approach aligns with what's expected.

Thanks!

importer/src/main/java/org/hiero/mirror/importer/migration/AccountEvmAddressMigration.java
```diff
@@ -36,13 +36,13 @@ private void updateAlias(boolean history) {
                 "select id, alias from entity%s where evm_address is null and length(alias) = 35", suffix);
         var update = String.format("update entity%s set evm_address = :evmAddress where id = :id", suffix);
 
-        jdbcOperations.query(query, rs -> {
+        jdbcOperationsProvider.getObject().query(query, rs -> {
```
steven-sheehy: We can refactor to call `getObject()` once and use it at each invocation.

```suggestion
        final var jdbcOperations = jdbcOperationsProvider.getObject();
        jdbcOperations.query(query, rs -> {
```

importer/src/main/java/org/hiero/mirror/importer/migration/AsyncJavaMigration.java
```diff
@@ -63,21 +65,33 @@ with last as (
             where f.installed_rank = last.installed_rank
             """;
 
-    protected final NamedParameterJdbcTemplate namedParameterJdbcTemplate;
+    private final ObjectProvider<JdbcTemplate> jdbcTemplateProvider;
+    private NamedParameterJdbcTemplate namedParameterJdbcTemplate;
```
steven-sheehy: We can just store `NamedParameterJdbcOperations` and use `namedParameterJdbcOperations.getJdbcOperations()` when needed.

```suggestion
    private final ObjectProvider<NamedParameterJdbcOperation> namedParameterJdbcOperation;
```

importer/src/main/java/org/hiero/mirror/importer/migration/AsyncJavaMigration.java
```diff
@@ -63,21 +65,33 @@ with last as (
...
-        this.namedParameterJdbcTemplate = namedParameterJdbcTemplate;
+        this.jdbcTemplateProvider = jdbcTemplateProvider;
         this.schema = schema;
     }
 
+    protected final JdbcTemplate getJdbcTemplate() {
+        return jdbcTemplateProvider.getObject();
+    }
+
+    protected final NamedParameterJdbcTemplate getNamedParameterJdbcTemplate() {
```
steven-sheehy: We should only use their Operations interfaces everywhere and not reference the template implementation. We haven't done a good job of that in places, but let's take the time now to do it properly. That is, for fields, methods, etc always use `NamedParameterJdbcOperations` over `NamedParameterJdbcTemplate` and `JdbcOperations` over `JdbcTemplate`.
```suggestion
    protected final JdbcOperations getJdbcOperations() {
        return jdbcOperationsProvider.getObject();
    }

    protected final NamedParameterJdbcOperations getNamedParameterJdbcOperations() {
```

importer/src/main/java/org/hiero/mirror/importer/migration/BackfillAndDeduplicateBalanceMigration.java
```diff
@@ -128,28 +127,21 @@ select coalesce(max(consensus_timestamp), 0)
...
             ImporterProperties importerProperties,
-            @Owner JdbcTemplate jdbcTemplate,
-            TimePartitionService timePartitionService) {
-        super(
-                importerProperties.getMigration(),
-                new NamedParameterJdbcTemplate(jdbcTemplate),
-                dbProperties.getSchema());
-        this.jdbcTemplate = jdbcTemplate;
-        this.timePartitionService = timePartitionService;
+            @Owner ObjectProvider<JdbcTemplate> jdbcTemplateProvider,
```
steven-sheehy: Same comment for every other sub-class

```suggestion
            @Owner ObjectProvider<JdbcOperations> jdbcOperationsProvider,
```

importer/src/main/java/org/hiero/mirror/importer/migration/BackfillBlockMigration.java
```diff
@@ -37,21 +37,23 @@ public class BackfillBlockMigration extends AsyncJavaMigration<Long> {
...
     public BackfillBlockMigration(
             DBProperties dbProperties,
             ImporterProperties importerProperties,
-            NamedParameterJdbcTemplate jdbcTemplate,
-            RecordFileRepository recordFileRepository,
-            TransactionOperations transactionOperations) {
-        super(importerProperties.getMigration(), jdbcTemplate, dbProperties.getSchema());
-        this.recordFileRepository = recordFileRepository;
-        this.transactionOperations = transactionOperations;
+            @Owner ObjectProvider<JdbcTemplate> jdbcTemplateProvider,
```
steven-sheehy: Shouldn't need `@Owner`.

importer/src/main/java/org/hiero/mirror/importer/migration/BackfillEthereumTransactionHashMigration.java
```diff
@@ -140,18 +142,24 @@ private void backfillTables(List<MigrationEthereumTransaction> patchedTransactio
...
-                patchedTransactions,
-                patchedTransactions.size(),
-                (ps, transaction) -> {
-                    ps.setLong(1, transaction.getConsensusTimestamp());
-                    ps.setBytes(2, transaction.getHash());
-                });
-
-        jdbcTemplate.batchUpdate(UPDATE_ETHEREUM_HASH_SQL, patchedTransactions, patchedTransactions.size(), PSS);
+        jdbcTemplateProvider
+                .getObject()
```
steven-sheehy: get the jdbc operations once and store in a variable.

importer/src/main/java/org/hiero/mirror/importer/migration/TopicMessageLookupMigration.java
```diff
@@ -137,38 +129,35 @@ not exists(select 1 from topic_message_lookup where partition = ? limit 1) and
...
+    private final ObjectProvider<TransactionOperations> transactionOperationsProvider;
 
     private Collection<Set<Long>> topicByShard;
 
-    @Lazy
     @SuppressWarnings("java:S107")
     protected TopicMessageLookupMigration(
             EntityProperties entityProperties,
-            JdbcTemplate jdbcTemplate,
+            @Owner ObjectProvider<JdbcTemplate> jdbcTemplateProvider,
```
steven-sheehy: Remove Owner

importer/src/main/java/org/hiero/mirror/importer/migration/BackfillTransactionHashMigration.java
```diff
@@ -252,13 +250,26 @@ private Strategy getStrategy() {
...
 
+    private JdbcTemplate getJdbcTemplate() {
+        return jdbcTemplateProvider.getObject();
+    }
+
+    private NamedParameterJdbcTemplate getNamedParameterJdbcTemplate() {
+        if (namedParameterJdbcTemplate != null) {
+            return namedParameterJdbcTemplate;
+        }
+        namedParameterJdbcTemplate = new NamedParameterJdbcTemplate(getJdbcTemplate());
```
steven-sheehy: Remove in favor of using `NamedParameterJdbcOperations.getJdbcOperations()`.

importer/src/main/java/org/hiero/mirror/importer/migration/BackfillTransactionHashMigration.java
```diff
@@ -252,13 +250,26 @@ private Strategy getStrategy() {
...
+        var transactionManager = new DataSourceTransactionManager(
+                Objects.requireNonNull(getJdbcTemplate().getDataSource()));
         return new TransactionTemplate(transactionManager);
     }
 
+    private JdbcTemplate getJdbcTemplate() {
+        return jdbcTemplateProvider.getObject();
+    }
+
+    private NamedParameterJdbcTemplate getNamedParameterJdbcTemplate() {
```
steven-sheehy: Rename fields and methods to use operations interface

steven-sheehy (changes requested):
Looking pretty good so far. I've triggered the CI tests. Note that you can also run them locally via `./gradlew :importer:build` and it should run the relevant tests.

steven-sheehy:
We added a new `ContractLogIndexMigration` that also needs `@Lazy` removed.

HamzaElzarw-2022:
> We added a new ContractLogIndexMigration that also needs @Lazy removed.

Got it @steven-sheehy , i will make sure to rebase and revise @Lazy usages before merging.

HamzaElzarw-2022:
@steven-sheehy In some classes, when I replace `JdbcTemplate` with `JdbcOperations`, issues arise from calls to methods that only exist in `JdbcTemplate` inherited from `JdbcAccessor`, which are being used directly.

for example in `BackfillAndDeduplicateBalanceMigration.java`:

```
private TransactionOperations transactionOperations() {
    var jdbcTemplate = (JdbcTemplate) getJdbcOperations();
    var transactionManager = new DataSourceTransactionManager(Objects.requireNonNull(
            jdbcTemplate.getDataSource()));
    return new TransactionTemplate(transactionManager);
}
```

Here, I rely on casting `JdbcOperations` to `JdbcTemplate`. Is this an acceptable approach, or should I leave direct usage of `JdbcTemplate` in such cases?

Additionally, while investigating this, I noticed that in `JdbcAccessor`, it states that the data source must be explicitly set before calling `getDataSource()` which is not happening any where in the class. I'm not sure if this poses a problem in our context, but I wanted to flag it just in case.

steven-sheehy:
> Here, I rely on casting JdbcOperations to JdbcTemplate. Is this an acceptable approach, or should I leave direct usage of JdbcTemplate in such cases?

That's fine if it's just done for accessing the data source in a few places.

> Additionally, while investigating this, I noticed that in JdbcAccessor, it states that the data source must be explicitly set before calling getDataSource() which is not happening any where in the class. I'm not sure if this poses a problem in our context, but I wanted to flag it just in case.

I don't think this is a concern since these templates should be constructed using a DataSource constructor.

HamzaElzarw-2022:
Hello @steven-sheehy, All constructor-level usages of `@Lazy` have been removed, and the PR is ready for review.

There is one remaining usage of `@Lazy` on the class `RedisTopicListener` in the grpc module. Since removing it doesn't appear to cause any issues, I'm unsure whether it should be removed as well. Please advise.

importer/src/main/java/org/hiero/mirror/importer/migration/AsyncJavaMigration.java
```diff
@@ -64,21 +67,34 @@ with last as (
...
+            @Override
+            public NamedParameterJdbcOperations getObject() {
+                return new NamedParameterJdbcTemplate(jdbcOperationsProvider.getObject());
+            }
+        };
         this.schema = schema;
     }
 
+    protected final JdbcOperations getJdbcOperations() {
+        return namedParameterJdbcOperationsProvider.getObject().getJdbcOperations();
```
steven-sheehy:
```suggestion
        return getJdbcOperations().getJdbcOperations();
```

importer/src/main/java/org/hiero/mirror/importer/migration/BackfillTransactionHashMigration.java
```diff
@@ -252,13 +256,22 @@ private Strategy getStrategy() {
     }
 
     private TransactionTemplate getTransactionTemplate() {
+        var jdbcTemplate = (JdbcTemplate) getJdbcOperations();
         var transactionManager = new DataSourceTransactionManager(Objects.requireNonNull(jdbcTemplate.getDataSource()));
         return new TransactionTemplate(transactionManager);
     }
 
+    private JdbcOperations getJdbcOperations() {
+        return namedParameterJdbcOperationsProvider.getObject().getJdbcOperations();
```
steven-sheehy:
```suggestion
        return getNamedParameterJdbcOperations().getJdbcOperations();
```

steven-sheehy:
> There is one remaining usage of @Lazy on the class RedisTopicListener in the grpc module. Since removing it doesn't appear to cause any issues, I'm unsure whether it should be removed as well. Please advise.

Yes, please remove it as well. There's also one lazy in a test migration that can be removed.

importer/src/test/java/org/hiero/mirror/importer/migration/TokenAccountBalanceMigrationTest.java
```diff
@@ -66,12 +68,39 @@ class TokenAccountBalanceMigrationTest extends ImporterIntegrationTest {
...
     void beforeEach() {
         timestamp = new AtomicLong(domainBuilder.timestamp());
+
+        ObjectProvider<AccountBalanceFileRepository> accountBalanceFileRepositoryProvider =
+                new ObjectProvider<AccountBalanceFileRepository>() {
+                    @Override
+                    public AccountBalanceFileRepository getObject() {
+                        return accountBalanceFileRepository;
+                    }
+                };
```
steven-sheehy: Add the below to `ImporterIntegrationTest`:

```java
    public static final <T> ObjectProvider<T> objectProvider(T t) {
        return new ObjectProvider<T>() {
            @Override
            public T getObject() throws BeansException {
                return t;
            }
        };
    }
```

Then we can reduce the 5-6 lines down to one for everywhere in the diff that we use an anonymous ObjectProvider. Also can use var.
```suggestion
        var accountBalanceFileRepositoryProvider = objectProvider(accountBalanceFileRepository);
```

HamzaElzarw-2022: Thanks! that looks cool :)

HamzaElzarw-2022:
Removing `@Lazy` from the `ErrataMigration` class caused an error: "No bean found for autowire" for the `Resource balanceOffsets` parameter. Since this field is configured using `@Value`, I removed it from the constructor to resolve the issue. However, this change meant the field could no longer be declared as `final`.

steven-sheehy:
Looks good, but seems there's an issue with running importer build with spring.profiles.active=v2 if you can check locally.

HamzaElzarw-2022:
Hi @steven-sheehy, I've been looking into why the Importer v2 build is failing, but I haven't been able to find the issue so far. Do you have any insights or suggestions on what might be going wrong?

steven-sheehy:
There are some migrations that only run with v2 profile, like `TopicMessageLookupMigration`. Need to add `ObjectProvider` around `RecordFileRepository` and `TimePartitionService` in that migration. Set spring.profiles.active=v2 in application.yml and re-run migration tests.

HamzaElzarw-2022:
Thanks @steven-sheehy! I added the missing `ObjectProvider`, importer v2 build should work fine now.

steven-sheehy (approved):
LGTM. Thanks for you contribution! 🥂

HamzaElzarw-2022:
Thanks @steven-sheehy, I would love to work on any other issues you think will be suitable.

steven-sheehy:
Awesome, thanks for the offer. You're welcome to look through the open tickets and see if anything is interesting to you. As far as good first issues, I only see a few right now like https://github.com/hiero-ledger/hiero-mirror-node/issues/10534 and https://github.com/hiero-ledger/hiero-mirror-node/issues/10688. And we're always open to improvements in our docs. I'll try to see if anything else comes to mind.

steven-sheehy:
@HamzaElzarw-2022 Also #3162 is a good option.

PR #11686, opened 2025-07-28
Runbook script improvements

### LINKED ISSUE #11662: Enhance Citus Runbook Scripts
### Problem

After recent updates to citus runbooks there are still some adjustments needed.

### Solution

1. Adjust restore snapshot script to skip rotating to coord-0 and making sure replica is caught up
2. Ensure scripts are idempotent and able to be re-ran when there is a failure in the middle of the script on a previous run
3. Make master version upgrade optional in upgrade k8s script
4. increase shutdown timeout to 10 minutes and add retries 
5. Detect pods stuck in scale down state. Verify annotation exists, then delete manually

### Alternatives

_No response_

jnels124:
**Description**:
* skip clean shutdown when restoring from snapshot
* ensure all steps can be reran if previous run resulted in failure
* force delete pods that don't scale down
* make master version upgrade optional
* increase timeout for clean shutdown and allow retries
* ensure gke version upgrade is bumping to later version

**Related issue(s)**:

Fixes #11662 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/docs/runbook/change-citus-node-pool-machine-type.md
+++ b/docs/runbook/change-citus-node-pool-machine-type.md
@@ -6,7 +6,7 @@ Need to Change Machine Type for Citus Node Pool(s)
 
 ## Prerequisites
 
-- Have `jq` installed
+- Have `jq` and `yq` installed
 - kubectl is pointing to the cluster you want to change the machine type for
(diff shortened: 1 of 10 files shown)


xin-hedera (approved):
LGTM

PR #11704, opened 2025-08-01
Add flag to enable and disable endpoints exposure through ingress

### LINKED ISSUE #11644: Support enabled flag for individual rest-java routes in Helm
### Problem

We need a way to rollout the new Java-based APIs gradually while still keeping the legacy JavaScript APIs as a fallback.

### Solution

* Support the feature in both API Gateway and Ingress
* Add an `ingress.hosts.paths.enabled` flag
* Since `gateway.rules` is templated, we can probably just add an `{{ if false }}` around rules, but needs testing
* Investigate feasibility of percentage-based routing

### Alternatives

_No response_

sdimitrov9:
**Description**:
Add support for enabling/disabling js endpoints rewritten in rest-java api. Falling back to js api in case of endpoint in rest-java is disabled,

Example disabling rest-java endpoint and and falling back to its rest service analogues:
`helm upgrade -i -n test --create-namespace mirror charts/hedera-mirror -f charts/hedera-mirror/values-prod.yaml -f charts/hedera-mirror/values-v2.yml -f values-minikube.yml --set restjava.routes.networkStake = false`

**Related issue(s)**:

Fixes #11644 

**Checklist**


charts/hedera-mirror-rest-java/values.yaml
```diff
@@ -129,9 +137,13 @@ ingress:
       paths:
         # the rest of /api/v1/* is still handled by the Node.js based REST API logic, except for these paths
         - path: '/api/v1/accounts/(\d+\.){0,2}(\d+|(0x)?[A-Fa-f0-9]{40}|(?:[A-Z2-7]{8})*(?:[A-Z2-7]{2}|[A-Z2-7]{4,5}|[A-Z2-7]{7,8}))/allowances/nfts'
+          enabled: true
```
xin-hedera: think the general problem of the approach is it doesn't provide convenience compared to not adding the `enabled` property.

the reason is helm template doesn't merge lists, it's always overridden so the last such in the value sources is honored.

In other words, if we have to disable one, we'd have to provide all other paths we want to keep...

charts/hedera-mirror-rest/values.yaml
```diff
@@ -71,10 +71,196 @@ gateway:
...
+        {{- if true }}
+        - path:
+            type: RegularExpression
+            value: '/api/v1/blocks/((0x)?[0-9a-fA-F]+|\d+)$'
+        {{- end }}
+        {{- if true }}
         - path:
             type: PathPrefix
-            value: /api/v1
+            value: /api/v1/blocks
```
xin-hedera: I don't think we need these changes, at least for traefik in our production environment, it gives longer path higher priority. Given a request path starting with `/api/v1`, if there's no more-specific path matching, the fallback is the js rest api.

charts/hedera-mirror-rest-java/values.yaml
sdimitrov9: Updated, now you can disable/enable endpoint exposure through the rest-java service on helm upgrade execution using : `--set restjava.gateway.endpoints.topics.enabled=false `

charts/hedera-mirror-rest/values.yaml
sdimitrov9: Addressed

xin-hedera (dismissed):
approve to unblock the PR

kselveliev (approved):

charts/hedera-mirror-rest-java/values.yaml
```diff
@@ -71,19 +84,27 @@ gateway:
           kind: "{{ .Values.gateway.target.kind }}"
           name: "{{ include \"hedera-mirror-rest-java.fullname\" $ }}"
           port: 80
-      matches:
+      matches: |
+        {{- if .Values.gateway.endpoints.accounts.allowances.nfts.enabled }}
```
steven-sheehy: We should omit existing routes. We only need the flag for new routes like networkStake that can appear in both JS and Java modules. Probably you can just add the entry for networkStake here that is in the other PR.

charts/hedera-mirror-rest-java/values.yaml
```diff
@@ -47,6 +47,19 @@ envFrom: []
...
+        nfts:
+          enabled: true
+      airdrops:
+        outstanding:
+          enabled: true
+        pending:
+          enabled: true
+    topics:
+      enabled: true
+
```
steven-sheehy: Prefer a simpler, flatten property path outside of both ingress and gateway like `routes.networkStake=false`

charts/hedera-mirror-rest-java/templates/ingress.yaml
```diff
@@ -19,17 +19,29 @@ metadata:
...
+          {{- if eq .path "/api/v1/accounts/(\\d+\\.){0,2}(\\d+|(0x)?[A-Fa-f0-9]{40}|(?:[A-Z2-7]{8})*(?:[A-Z2-7]{2}|[A-Z2-7]{4,5}|[A-Z2-7]{7,8}))/allowances/nfts" }}
+            {{- $pathEnabled = $.Values.gateway.endpoints.accounts.allowances.nfts.enabled -}}
+            {{- else if eq .path "/api/v1/accounts/(\\d+\\.){0,2}(\\d+|(0x)?[A-Fa-f0-9]{40}|(?:[A-Z2-7]{8})*(?:[A-Z2-7]{2}|[A-Z2-7]{4,5}|[A-Z2-7]{7,8}))/airdrops/outstanding" }}
+            {{- $pathEnabled = $.Values.gateway.endpoints.accounts.airdrops.outstanding.enabled -}}
+            {{- else if eq .path "/api/v1/accounts/(\\d+\\.){0,2}(\\d+|(0x)?[A-Fa-f0-9]{40}|(?:[A-Z2-7]{8})*(?:[A-Z2-7]{2}|[A-Z2-7]{4,5}|[A-Z2-7]{7,8}))/airdrops/pending" }}
+            {{- $pathEnabled = $.Values.gateway.endpoints.accounts.airdrops.pending.enabled -}}
+            {{- else if eq .path "/api/v1/topics/(\\d+\\.){0,2}\\d+$" }}
+            {{- $pathEnabled = $.Values.gateway.endpoints.topics.enabled -}}
+          {{- end }}
+          {{- if $pathEnabled }}
```
steven-sheehy: We should omit existing paths and not repeat regexes. We can add a `condition` to each path. It should default to true if omitted.

```suggestion
  {{- if or (empty .condition) (eq (tpl .condition $) "true") }}
```

Then in values.yaml
```yaml
        - path: '/api/v1/network/stake'
          condition: "{{ .Values.routes.networkStake }}"
```

steven-sheehy (changes requested):

charts/hedera-mirror-rest-java/values.yaml
sdimitrov9: Addressed.

charts/hedera-mirror-rest-java/templates/ingress.yaml
sdimitrov9: Addressed.

charts/hedera-mirror-rest-java/values.yaml
sdimitrov9: Addressed.

steven-sheehy (approved):
LGTM

PR #11723, opened 2025-08-04
Remote write db metrics mainnet-eu

jnels124:
remote write db metrics mainnet-eu


steven-sheehy:
Closing in favor of #11730

PR #11761, opened 2025-08-11
Add github workflow for running acceptance tests against solo in CI/CD

### LINKED ISSUE #8513: Test acceptance tests in CI
### Problem

Acceptance tests frequently fail later in the lifecycle due to only verifying them during deployment to environments.

### Solution

Create a new workflow that:
* Builds a hedera-mirror-test image
* Deploys a solo cluster that contains a single consensus node, minio, and mirror node
* Run acceptance test via helm test

### Alternatives

_No response_

IvanKavaldzhiev:
**Description**:
Adding support for running acceptance tests in CI/CD using solo. For now the idea is the tests to run using latest consensus and mirror node tags.

**Related issue(s)**:

Fixes #8513 

**Notes for reviewer**:

**Checklist**


.github/workflows/acceptance.yaml
```diff
@@ -0,0 +1,137 @@
...
+          java-version: 21
+
+      - name: Install Solo CLI via npm
+        run: |
+          npm install -g @hashgraph/solo
+
+      - name: Cleanup previous runs
+        run: |
+          kind delete cluster -n "${SOLO_CLUSTER_NAME}" || echo "No previous cluster to delete"
+          rm -rf ~/.solo || echo "No previous solo config to remove"
```
xin-hedera: this is not needed. gh workflow jobs always run in a clean environment.

.github/workflows/acceptance.yaml
```diff
@@ -0,0 +1,137 @@
...
+          sudo apt-get install -y --no-install-recommends wget
+
+      - name: Install Helm
+        uses: azure/setup-helm@b9e51907a09c216f16ebe8536097933489208112 # v4.3.0
+
+      - name: Setup Kind
+        uses: helm/kind-action@a1b0e391336a6ee6713a0583f8c6240d70863de3 # v1.12.0
+        with:
+          kubectl_version: v1.32.3
+          version: v0.27.0
```
xin-hedera: latest kind release is v0.29.0

.github/workflows/acceptance.yaml
```diff
@@ -0,0 +1,137 @@
...
+      - name: Install Solo CLI via npm
+        run: |
+          npm install -g @hashgraph/solo
+
+      - name: Cleanup previous runs
+        run: |
+          kind delete cluster -n "${SOLO_CLUSTER_NAME}" || echo "No previous cluster to delete"
+          rm -rf ~/.solo || echo "No previous solo config to remove"
+
+      - name: Setup Solo Cluster
```
xin-hedera, line 80: to test the mirrornode modules and run acceptance tests from the code of the triggering commit, we'll have to 
- build all relevant docker images, not just the `hedera-mirror-rest` image
- push them to a local docker registry
- config the kind cluster to use the local docker registry
- set global.image.registry, test.image.registry, and the correct tag when helm install / upgrade `hedera-mirror` chart

links for references:

- https://docs.docker.com/build/ci/github-actions/local-registry/
- https://kind.sigs.k8s.io/docs/user/local-registry/

.github/workflows/acceptance.yaml
```diff
@@ -0,0 +1,137 @@
...
+
+          solo init || { echo "Failed to initialize Solo"; exit 1; }
+          solo cluster-ref connect --cluster-ref kind-"${SOLO_CLUSTER_NAME}" --context kind-"${SOLO_CLUSTER_NAME}" || { echo "Failed to connect to cluster"; exit 1; }
+          solo deployment create --deployment "${SOLO_DEPLOYMENT}" --namespace "${SOLO_CLUSTER_NAME}" || { echo "Failed to create deployment"; exit 1; }
+          solo deployment add-cluster --deployment "${SOLO_DEPLOYMENT}" --cluster-ref kind-"${SOLO_CLUSTER_NAME}" --num-consensus-nodes 1 || { echo "Failed to add cluster to deployment"; exit 1; }
+          solo cluster-ref setup --cluster-ref kind-"${SOLO_CLUSTER_NAME}" || { echo "Failed to setup cluster"; exit 1; }
+          solo node keys --gossip-keys --tls-keys --deployment "${SOLO_DEPLOYMENT}" -i node1 || { echo "Failed to generate keys"; exit 1; }
+          solo network deploy --deployment "${SOLO_DEPLOYMENT}" || { echo "Failed to deploy network"; exit 1; }
+          solo node setup --deployment "${SOLO_DEPLOYMENT}" -i node1 || { echo "Failed to setup node"; exit 1; }
+          solo node start --deployment "${SOLO_DEPLOYMENT}" -i node1 || { echo "Failed to start node"; exit 1; }
```
xin-hedera: can we do `solo quick start single` and then helm upgrade?

.github/workflows/acceptance.yaml
```diff
@@ -0,0 +1,137 @@
...
+
+      - name: Wait for services to be ready
+        run: |
+          echo "Waiting for services to be ready..."
+          kubectl wait --for=condition=ready pod -l app=mirror -n "${SOLO_NAMESPACE}" --timeout=300s || echo "Some pods may not be ready, continuing with tests"
+
+      - name: Run acceptance tests
+        run: |
+          echo "Running acceptance tests..."
+          helm test mirror -n "${SOLO_NAMESPACE}" --logs --filter name=mirror-acceptance || { echo "Acceptance tests failed"; exit 1; }
```
xin-hedera: think we need a longer than default timeout here, the default is 5 minutes.

.github/workflows/acceptance.yaml
steven-sheehy: Think that would be too slow and costly. I think for PRs we just want to test changes to the test folder against main images for other components.

.github/workflows/acceptance.yaml
IvanKavaldzhiev: Agree, deleted.

.github/workflows/acceptance.yaml
IvanKavaldzhiev: I removed the explicitly set version. According to the kind-action documentation, 0.29.0 should be used by default, when not specified.

.github/workflows/acceptance.yaml
IvanKavaldzhiev: Theoretically yes, but we discussed with Steven this might be waste of resources, since mirror node will be started by this single one command, then reinitialized with the helm upgrade. We can think of using an environment variable specifying a `values.yaml` file to use when deploying mirror-node with this one line command. This should be a changed adapted in the `solo` repo.

.github/workflows/acceptance.yaml
IvanKavaldzhiev: Added

.github/workflows/acceptance.yaml
```diff
@@ -0,0 +1,150 @@
...
+
+      - name: Install Solo CLI via npm
+        run: |
+          npm install -g @hashgraph/solo
+
+      - name: Build Mirror Node Test JAR
+        run: ./gradlew test:build -x test
+
+      - name: Build Mirror Node Test Docker image
+        run: docker build -t hiero-mirror-node:pr-test -f test/Dockerfile ./test
```
xin-hedera: the image won't be available for the kind cluster, you'll need [kind load docker-image](https://kind.sigs.k8s.io/docs/user/quick-start/#loading-an-image-into-your-cluster)

steven-sheehy: Ivan: Why can't we just run gradlew acceptance test against the cluster and avoid the image?

xin-hedera: we'll have to port-forward consensus node port 50211. btw, my previous comment is wrong, the image is loaded in a later step.

.github/workflows/acceptance.yaml
```diff
@@ -0,0 +1,189 @@
...
+          REST_JAVA_URL="http://${SOLO_DEPLOYMENT}-restjava.${SOLO_NAMESPACE}.svc.cluster.local"
+          WEB3_URL="http://${SOLO_DEPLOYMENT}-web3.${SOLO_NAMESPACE}.svc.cluster.local"
+          NETWORK_URL="network-node1-svc.${SOLO_NAMESPACE}.svc.cluster.local:50211"
+
+          # Export for use in next steps
+          echo "GRPC_URL=${GRPC_URL}" >> $GITHUB_ENV
+          echo "REST_URL=${REST_URL}" >> $GITHUB_ENV
+          echo "REST_JAVA_URL=${REST_JAVA_URL}" >> $GITHUB_ENV
+          echo "WEB3_URL=${WEB3_URL}" >> $GITHUB_ENV
+          echo "NETWORK_URL=${NETWORK_URL}" >> $GITHUB_ENV
```
xin-hedera: consensus node grpc service is exposed at `localhost:50211`, and if you add `--enable-ingress` to `solo mirror-node deploy`, all APIs including mirrornode grpc API can be accessed at `localhost:8081`

.github/workflows/acceptance.yaml
```diff
@@ -0,0 +1,189 @@
...
+
+      - name: Run acceptance tests
+        run: |
+          ./gradlew :test:acceptance --info \
+            -Dcucumber.filter.tags=@acceptance \
+            -Dhiero.mirror.test.acceptance.mirrorNodeAddress=${GRPC_URL} \
+            -Dhiero.mirror.test.acceptance.network=OTHER \
+            -Dhiero.mirror.test.acceptance.nodes[0].accountId=${SOLO_NODE_ACCOUNT_ID} \
```
xin-hedera: I think we can just use `0.0.3` as node account id. if it changes, fix it then.

.github/workflows/acceptance.yaml
```diff
@@ -0,0 +1,189 @@
...
+            -Dhiero.mirror.test.acceptance.nodes[0].host=network-node1-svc.${SOLO_NAMESPACE}.svc.cluster.local \
+            -Dhiero.mirror.test.acceptance.nodes[0].port=50211 \
+            -Dhiero.mirror.test.acceptance.operatorId=${OPERATOR_ID} \
+            -Dhiero.mirror.test.acceptance.operatorKey=${OPERATOR_KEY} \
```
xin-hedera: does solo not use the default genesis key for 0.0.2?

IvanKavaldzhiev: Yes, I removed the explicit assignment here

.github/workflows/acceptance.yaml
IvanKavaldzhiev: Removed

.github/workflows/acceptance.yaml
IvanKavaldzhiev: I enabled ingress

.github/workflows/acceptance.yaml
```diff
@@ -0,0 +1,150 @@
...
+
+      - name: Setup Node
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
+        with:
+          node-version: 21
+
+      - name: Install Wget
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends wget
```
steven-sheehy: Why do we need wget?

.github/workflows/acceptance.yaml
```diff
@@ -0,0 +1,150 @@
...
+        with:
+          egress-policy: audit
+
+      - name: Checkout
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - name: Setup Node
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
+        with:
+          node-version: 21
```
steven-sheehy: It looks like solo prefers node 20
```suggestion
          node-version: 20
```

.github/workflows/acceptance.yaml
```diff
@@ -0,0 +1,150 @@
...
+
+      - name: Install Solo CLI via npm
+        run: |
+          npm install -g @hashgraph/solo
+
+      - name: Build Mirror Node Test JAR
+        run: ./gradlew test:build -x test
+
+      - name: Build Mirror Node Test Docker image
+        run: docker build -t gcr.io/mirrornode/hedera-mirror-test:"${{ github.base_ref }}" -f test/Dockerfile ./test
```
steven-sheehy: Can omit `-f test/Dockerfile` as it's the default.

.github/workflows/acceptance.yaml
```diff
@@ -0,0 +1,150 @@
...
+
+      - name: Setup Solo Cluster
+        run: |
+          kind create cluster -n "${SOLO_CLUSTER_NAME}" || { echo "Failed to create Kind cluster"; exit 1; }
```
steven-sheehy: Add `set -ex` at beginning and remove all ` || { echo "message"; exit; }` at the end of every command.

.github/workflows/acceptance.yaml
```diff
@@ -0,0 +1,161 @@
...
+            enabled: true
+          EOF
+
+          solo mirror-node deploy --deployment "${SOLO_DEPLOYMENT}" -f=temp-values.yaml --cluster-ref kind-"${SOLO_CLUSTER_NAME}" --pinger || { echo "Failed to deploy mirror node"; exit 1; }
+
+          echo "Solo cluster setup completed successfully!"
+
+      - name: Run acceptance tests
+        run: |
+          helm test mirror -n "${SOLO_NAMESPACE}" --logs --filter name=mirror-acceptance --timeout 20m
```
steven-sheehy: You will get `unable to get pod logs` with `--logs`. Need to set `test.annotations.helm.sh/hook-delete-policy=before-hook-creation` to make it work.

.github/workflows/acceptance.yaml
```diff
@@ -0,0 +1,161 @@
...
+          for pod in $(kubectl get pods -n "${SOLO_NAMESPACE}" -o jsonpath='{.items[*].metadata.name}' 2>/dev/null || echo ""); do
+            if [ -n "$pod" ]; then
+              echo "--- Logs for pod: $pod ---"
+              kubectl logs "$pod" -n "${SOLO_NAMESPACE}" --all-containers=true --tail=1000 || echo "Could not get logs for pod $pod"
+            fi
+          done
+
+      - name: Cleanup
+        if: always()
+        run: |
```
steven-sheehy: Add `continue-on-error: true`

.github/workflows/acceptance.yaml
```diff
@@ -0,0 +1,161 @@
...
+          echo "Cleaning up resources..."
+          kind delete cluster -n "${SOLO_CLUSTER_NAME}" || echo "Failed to delete cluster"
+          rm -rf ~/.solo || echo "Failed to remove solo config"
+          echo "Cleanup completed"
```
steven-sheehy: Can simplify. No need to remove files as actions run in a container or vm
```suggestion
        run: kind delete cluster -n "${SOLO_CLUSTER_NAME}"
```

.github/workflows/acceptance.yaml
```diff
@@ -0,0 +1,161 @@
...
+          solo mirror-node deploy --deployment "${SOLO_DEPLOYMENT}" -f=temp-values.yaml --cluster-ref kind-"${SOLO_CLUSTER_NAME}" --pinger || { echo "Failed to deploy mirror node"; exit 1; }
+
+          echo "Solo cluster setup completed successfully!"
+
+      - name: Run acceptance tests
+        run: |
+          helm test mirror -n "${SOLO_NAMESPACE}" --logs --filter name=mirror-acceptance --timeout 20m
+
+      - name: Show Pod Logs on Failure
+        if: ${{ failure() }}
```
steven-sheehy, line 120: Add continue-on-error: true

.github/workflows/acceptance.yaml
```diff
@@ -0,0 +1,161 @@
...
+        run: |
+          echo "--------------------------------------------------"
+          echo "Workflow failed. Collecting logs for debugging..."
+          echo "--------------------------------------------------"
+
+          echo "--- Describing all pods in namespace ${SOLO_NAMESPACE} ---"
+          kubectl describe pods -n "${SOLO_NAMESPACE}" || echo "Could not describe pods"
+
+          echo "--- Fetching logs for all pods in namespace ${SOLO_NAMESPACE} ---"
+          for pod in $(kubectl get pods -n "${SOLO_NAMESPACE}" -o jsonpath='{.items[*].metadata.name}' 2>/dev/null || echo ""); do
```
steven-sheehy: Use `-o name` instead.

.github/workflows/acceptance.yaml
```diff
@@ -0,0 +1,161 @@
...
+          solo node setup --deployment "${SOLO_DEPLOYMENT}" -i node1 --release-tag "${LATEST_NETWORK_TAG}" || { echo "Failed to setup node"; exit 1; }
+          solo node start --deployment "${SOLO_DEPLOYMENT}" -i node1 || { echo "Failed to start node"; exit 1; }
+
+          cat <<EOF > temp-values.yaml
+          global:
+            image:
+              pullPolicy: IfNotPresent
+              tag: "${{ github.base_ref }}"
+
+          web3:
```
steven-sheehy: Remove the extra newlines and sort web3 after test.

.github/workflows/acceptance.yaml
```diff
@@ -0,0 +1,161 @@
...
+            | tail -n1)
+
+          solo network deploy --deployment "${SOLO_DEPLOYMENT}" --release-tag "${LATEST_NETWORK_TAG}" || { echo "Failed to deploy network"; exit 1; }
+          solo node setup --deployment "${SOLO_DEPLOYMENT}" -i node1 --release-tag "${LATEST_NETWORK_TAG}" || { echo "Failed to setup node"; exit 1; }
+          solo node start --deployment "${SOLO_DEPLOYMENT}" -i node1 || { echo "Failed to start node"; exit 1; }
+
+          cat <<EOF > temp-values.yaml
+          global:
+            image:
+              pullPolicy: IfNotPresent
```
steven-sheehy: Can remove since this is the default

.github/workflows/acceptance.yaml
```diff
@@ -0,0 +1,161 @@
...
+              tag: "${{ github.base_ref }}"
+
+          web3:
+            config:
+              hiero:
+                mirror:
+                  web3:
+                    evm:
+                      modularizedServices: true
```
steven-sheehy: I don't think we need to necessarily test explicitly with modularized. We can just keep it at the default so when the default changes soon it will switch to modularized.

.github/workflows/acceptance.yaml
```diff
@@ -0,0 +1,161 @@
...
+
+          test:
+            image:
+              tag: pr-test
+            config:
+              hiero:
+                mirror:
+                  test:
+                    acceptance:
+                      network: OTHER
```
steven-sheehy: Could shorten lines by using env variable instead

.github/workflows/acceptance.yaml
```diff
@@ -0,0 +1,161 @@
+# SPDX-License-Identifier: Apache-2.0
+
+name: Acceptance
+
+on:
+  pull_request:
+    branches: [main, release/**]
+    paths:
+      - "test/**"
```
steven-sheehy: We're mixing array styles. Prefer branches to have array items on separate lines like paths. Same for push below.

.github/workflows/acceptance.yaml
```diff
@@ -0,0 +1,161 @@
...
+
+permissions:
+  contents: read
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  solo:
```
steven-sheehy:
```suggestion
  acceptance:
    name: Acceptance
```

.github/workflows/acceptance.yaml
```diff
@@ -0,0 +1,161 @@
+# SPDX-License-Identifier: Apache-2.0
+
+name: Acceptance
```
steven-sheehy:
```suggestion
name: End-to-end Tests
```

.github/workflows/acceptance.yaml
```diff
@@ -0,0 +1,161 @@
...
+          solo deployment create --deployment "${SOLO_DEPLOYMENT}" --namespace "${SOLO_NAMESPACE}" || { echo "Failed to create deployment"; exit 1; }
+          solo deployment add-cluster --deployment "${SOLO_DEPLOYMENT}" --cluster-ref kind-"${SOLO_CLUSTER_NAME}" --num-consensus-nodes 1 || { echo "Failed to add cluster to deployment"; exit 1; }
+          solo cluster-ref setup --cluster-ref kind-"${SOLO_CLUSTER_NAME}" || { echo "Failed to setup cluster"; exit 1; }
+          solo node keys --gossip-keys --tls-keys --deployment "${SOLO_DEPLOYMENT}" -i node1 || { echo "Failed to generate keys"; exit 1; }
+
+          LATEST_NETWORK_TAG=$(curl -s https://github.com/hiero-ledger/hiero-consensus-node/tags \
+            | grep -oE 'v[0-9]+\.[0-9]+\.[0-9]+(-[A-Za-z0-9._-]+)?(\.tar\.gz|\.zip)?' \
+            | sed -E 's/(\.tar\.gz|\.zip)$//' \
+            | sort -V \
+            | tail -n1)
```
steven-sheehy: Parsing HTML pages is brittle. We should depend upon the consensus node version in solo so that we get automatic updates. If you need to temporarily use some latest version due to some bug, can just hardcode it here for now and we'll remove it later.

steven-sheehy (changes requested):

.github/workflows/acceptance.yaml
IvanKavaldzhiev: Removed

.github/workflows/acceptance.yaml
IvanKavaldzhiev: Fixed

.github/workflows/acceptance.yaml
IvanKavaldzhiev: Removed

.github/workflows/acceptance.yaml
IvanKavaldzhiev: Added

.github/workflows/acceptance.yaml
IvanKavaldzhiev: Added

.github/workflows/acceptance.yaml
IvanKavaldzhiev: Added

.github/workflows/acceptance.yaml
IvanKavaldzhiev: Refactored

.github/workflows/acceptance.yaml
IvanKavaldzhiev: Added

.github/workflows/acceptance.yaml
IvanKavaldzhiev: Added

.github/workflows/acceptance.yaml
IvanKavaldzhiev: Fixed

.github/workflows/acceptance.yaml
IvanKavaldzhiev: Removed

.github/workflows/acceptance.yaml
IvanKavaldzhiev: Removed

.github/workflows/acceptance.yaml
IvanKavaldzhiev: Fixed

.github/workflows/acceptance.yaml
IvanKavaldzhiev: Refactored

.github/workflows/acceptance.yaml
IvanKavaldzhiev: Renamed

.github/workflows/acceptance.yaml
IvanKavaldzhiev: Renamed

.github/workflows/acceptance.yaml
IvanKavaldzhiev: Removed

.github/workflows/acceptance.yaml
```diff
@@ -0,0 +1,138 @@
...
+      SOLO_CLUSTER_SETUP_NAMESPACE: solo-mirror-acceptance-setup
+      SOLO_DEPLOYMENT: solo-deployment
+    steps:
+      - name: Harden Runner
+        uses: step-security/harden-runner@ec9f2d5744a09debf3a187a3f4f675c53b671911 # v2.13.0
+        with:
+          egress-policy: audit
+
+      - name: Checkout
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```
steven-sheehy, line 48: Missing:

```
        with:
          ref: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.branch || github.ref }}
```

IvanKavaldzhiev: Added

steven-sheehy (approved):
LGTM

PR #11767, opened 2025-08-13
Wrap contract service call in transaction

### LINKED ISSUE #11725: DB connection issues
### Description

Some intermittent DB connection issues are noticed in the Grafana logs. They do not seem related to slow queries, high query count per second or any other known pattern. The issue seems to occur even before August, 1st.

### Steps to reproduce

n/a

### Additional context

_No response_

### Hedera network

mainnet

### Version

v0.134.0

### Operating system

None

jnels124:
**Description**:
* wrap contact call service in transaction

**Related issue(s)**:

 Fixes #11725 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/web3/src/main/java/org/hiero/mirror/web3/service/ContractExecutionService.java
+++ b/web3/src/main/java/org/hiero/mirror/web3/service/ContractExecutionService.java
@@ -6,6 +6,7 @@ import com.google.common.base.Stopwatch;
 import com.hederahashgraph.api.proto.java.ResponseCodeEnum;
 import io.micrometer.core.instrument.MeterRegistry;
 import jakarta.inject.Named;
+import jakarta.transaction.Transactional;
 import java.util.Objects;
 import lombok.CustomLog;
 import org.apache.tuweni.bytes.Bytes;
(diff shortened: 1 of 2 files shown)


jnels124:
This will allow 3s of waiting for a connection and all work must complete within 10s for a total request time allowed of 13s

web3/src/main/java/org/hiero/mirror/web3/service/ContractExecutionService.java
```diff
@@ -46,7 +47,10 @@ public ContractExecutionService(
                 transactionExecutionService);
         this.binaryGasEstimator = binaryGasEstimator;
     }
-
+    /**
+     * Specififed as transactional to enforce timeouts across the entire call processing and to require all jpa calls to use the same connection
+     * */
+    @Transactional
```
xin-hedera, line 53: I think this caused serious performance regression previously. The problem is, all database queries in the same `processCall` are now wrapped in a top level `begin; ...; commit;` transaction. For user `mirror_web3`, pgbouncer pool mode is `transaction`, then we end up with pgbouncer holding connections for the call much longer v.s. previously transactions only last as long as specific repository method calls take. And many calls will just fail because no more `mirror_web3` connections available.

jnels124: With current config, each individual repository call is allowed to consume 10s individually. Only way to limit actual database query time for a whole request is to wrap in a transaction. k6 doesn't show a regression but that may definitely be different with mainnet traffic. The other option here to enforce a timeout for a single request is to change the controller to return a Callable (or similar) and configure spring async request timeout. The downside to that approach though is any currently running database calls would not be canceled and will still consume the connection it was using until the query is done.

jnels124: Do you know if previously the flush property was set to manual? Looking at docs etc it seems a fair amount of time is spent dirty checking because hibernate will perform dirty checks before each flush which happens automatically at the end of a transaction.

PR #11780, opened 2025-08-13
Fix partition maintenance

### LINKED ISSUE #11778: Fix partition maintenance functions
### Description

When partition maintenance is running on a db connection not using UTC, invalid partition names and bounds are generated

### Steps to reproduce

1. Configure TZ for importer and db to be non UTC timezone
2. delete the latest account_balance partition
3. restart the importer
4. Notice partition maintenance fails

### Additional context

_No response_

### Hedera network

_No response_

### Version

v0.132.0

### Operating system

None

jnels124:
**Description**:
* add utc timezone to v1 repeatable migration
* recreate partition maintenance functions in v2 via repeatable migration

**Related issue(s)**:

Fixes #11778 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/importer/src/main/resources/db/migration/v1/R__time_partition_maintenance.sql
+++ b/importer/src/main/resources/db/migration/v1/R__time_partition_maintenance.sql
@@ -47,6 +47,7 @@ $$;
 create or replace procedure create_mirror_node_time_partitions() language plpgsql as
 $$
 begin
+  set timezone to 'UTC';
   call create_time_partition_for_table('account_balance');
   call create_time_partition_for_table('token_balance');
 end;
--- a/importer/src/main/resources/db/migration/v2/R__time_partition_maintenance.sql
+++ b/importer/src/main/resources/db/migration/v2/R__time_partition_maintenance.sql
@@ -1,0 +1,36 @@
+create or replace procedure create_mirror_node_time_partitions() as
+$$
+declare
+    partition_info                    record;
+    created_partition                 boolean;
+    time_partition_pattern            varchar = '^(.*_timestamp|consensus_end)$';
+begin
+perform set_config('TimeZone', 'UTC', true);
+for partition_info in
+select distinct on (tp.parent_table) tp.parent_table,
+    to_timestamp(tp.to_value::bigint / 1000000000.0)                                      as next_from,
+    to_timestamp(tp.to_value::bigint / 1000000000.0) + interval ${partitiontimeinterval}  as next_to,
+    tp.partition                                                                          as current_partition
+from time_partitions tp
+where tp.partition_column::varchar ~ time_partition_pattern
+order by tp.parent_table, tp.from_value::bigint desc
+    loop
+    if current_timestamp + interval ${partitiontimeinterval} < partition_info.next_from then
+    raise log 'skipping partition creation for time partition % from % to %',
+    partition_info.parent_table, partition_info.next_from, partition_info.next_to;
+continue;
+end if;
+select create_time_partitions(partition_info.parent_table, interval ${partitiontimeinterval},
+                              partition_info.next_to, partition_info.next_from)
+into created_partition;
+if created_partition then
+                perform apply_vacuum_settings_from_child(partition_info.parent_table, partition_info.current_partition);
+end if;
+commit;
+raise log 'processed % for values from % to % created %',
+                       partition_info.parent_table, partition_info.next_from,
+                       partition_info.next_to, created_partition;
+end loop;
+end;
+$$
+language plpgsql;


importer/src/main/resources/db/migration/v1/R__time_partition_maintenance.sql
```diff
@@ -47,6 +47,7 @@ $$;
 create or replace procedure create_mirror_node_time_partitions() language plpgsql as
 $$
 begin
+  set timezone to 'UTC';
```
xin-hedera: should we try setting `spring.flyway.initSql` to `${hiero.mirror.importer.db.connectionInitSql}` instead?

jnels124: this will not work for the partition maintenance service. Originally i would expect `hiero.mirror.importer.db.connectionInitSql` to work in that case but it doesn't seem to work (probably just not set up properly for the owner template). However, in v2 that may cause issues anyways since this init sql is only ran on the hikari connection when it is created. Since we have pgbouncer in v2 I don't know that we can guarantee this is going to be carried over to those connections.

jnels124: Looks like pgbouncer has connect_query so maybe it is better to configure the owner template correctly to handle v1 and also configure pg bouner init query. I will investigate that some.

jnels124: Actually after looking `spring.flyway.initSql` may work for that service as i see the owner template is created with the flyway configuration. Still need to solve the pgbouncer side though

xin-hedera: yeah, guess so.

jnels124: Ok after some more investigation, we are not able to use `spring.flyway.initSql` as this only gets applied to connections flyway creates itself. In addition, we can't just add the timezone setting to the url and set the flyway url as there is no pooling configured to be used for flyway datasource and the jdbc driver calls set timezone (to jvm timezone) after the connection specified within the url is created.

jnels124: We can update the JDBCTemplateConfiguration to have its own hikari datasource for the owner template if we are ok with adding that? I would prefer this approach as it ensures we don't have any similar issues in future migrations

xin-hedera: think it's worth the effort so we have all connections set to UTC timezone. perhaps instead of doing that in `JDBCTemplateConfiguration`, we can customize flyway configuration to have a hikari managed datasource with user `mirror_node` in `ImporterConfiguration.flywayConfigurationCustomizer`?

charts/hedera-mirror/values.yaml
```diff
@@ -327,12 +327,14 @@ stackgres:
     instances: 1
     pgbouncer:
       pgbouncer:
+        connect_query: "set timezone to 'UTC'"
```
steven-sheehy: Shouldn't we also set something for our bitnami postgres to aid our solo deployments?

xin-hedera:
@jnels124 looks like there's sth wrong in v2 so chart install workflow job failed, likely caused by the pgbouncer changes

importer/src/main/resources/application.yml
```diff
@@ -104,6 +103,13 @@ spring:
   flyway:
     baselineOnMigrate: true
     connectRetries: 20
+    hikari:
+      connection-init-sql: ${hiero.mirror.importer.db.connectionInitSql}
+      data-source-properties: ${spring.datasource.hikari.data-source-properties}
+      maximum-pool-size: 10
```
xin-hedera: think we should set `minimum-idle: 0`, and a shorter `idle-timeout`, the default is 600000 (10 minutes).

minimumIdle by default is set to maximumPoolSize so makes it a fixed size pool, then these connections will just sit idle with nothing to run most of the time.

importer/src/main/java/org/hiero/mirror/importer/config/ImporterConfiguration.java
```diff
@@ -55,7 +71,12 @@ FlywayConfigurationCustomizer flywayConfigurationCustomizer() {
                     timestamp = 1588706343553042000L;
                 }
             }
+            var flywayDatasource = new HikariDataSource(hikariConfig);
+            var beanFactory = (DefaultListableBeanFactory) context.getAutowireCapableBeanFactory();
+            beanFactory.registerDisposableBean("flywayDataSource", flywayDatasource::close);
```
steven-sheehy: Since it implements Closeable Spring should already automatically close it on dispose. Should confirm and adjust.

steven-sheehy: It would be better to just create the `HikariDataSource` directly using `@Bean`. Can probably omit `flywayHikariConfig` bean and just directly create the data source.

importer/src/main/java/org/hiero/mirror/importer/config/ImporterConfiguration.java
```diff
@@ -45,7 +51,17 @@ LeaderService leaderService() {
...
 
     @Bean
-    FlywayConfigurationCustomizer flywayConfigurationCustomizer() {
+    @ConfigurationProperties(prefix = "spring.flyway.hikari")
+    @Owner
+    HikariConfig flywayHikariConfig(JdbcConnectionDetails connectionDetails) {
+        var config = new HikariConfig();
+        config.setJdbcUrl(connectionDetails.getJdbcUrl());
+        return config;
+    }
```
steven-sheehy: Do we really need to be able to configure all the properties separately at `spring.flyway.hikari` and have to duplicate them in the yml? Instead, can't we inject the existing HikariConfig and duplicate it?

```suggestion
    @Owner
    HikariDataSource flywayDataSource(JdbcConnectionDetails connectionDetails, HikariConfig hikariConfig) {
        var flywayConfig = new HikariConfig();
        hikariConfig.copyStateTo(flywayConfig);
        flywayConfig.setJdbcUrl(connectionDetails.getJdbcUrl());
        flywayConfig.setPoolName("flyway");
        return flywayConfig;
    }
```

importer/src/main/java/org/hiero/mirror/importer/config/ImporterConfiguration.java
xin-hedera: think it was done so in a previous commit, however, there might be issues as we need to figure out a way to mark the auto configured dataSource as `@Primary`

steven-sheehy: Can mark it with `@Fallback`

charts/hedera-mirror/values.yaml
xin-hedera: instead of setting timezone via pgboncer, we can change the server config by setting `timezone = UTC`, and can apply the same with bitnami postgres charts.

jnels124: Can you clarify what you mean? You mean just set the tz var in the env vars?

xin-hedera: I mean setting timezone = UTC in Postgres server config

importer/src/main/java/org/hiero/mirror/importer/config/ImporterConfiguration.java
jnels124: I can try that.

importer/src/main/java/org/hiero/mirror/importer/config/ImporterConfiguration.java
jnels124: I can try this but will only work if `@Fallback` works

importer/src/main/java/org/hiero/mirror/importer/config/ImporterConfiguration.java
jnels124: it doesn't close it unless it is a bean. It isn't a bean because it gets picked up as the default datasource. Will not need this if the `@Fallback ` annotation works to stop that

jnels124: `@Fallback` still prevents the autoconfigured datasource from being created but there is a property on the `@Bean` annotation we can use and use `@FlywayDataSource` to configure flyway to use this bean.

charts/hedera-mirror/values.yaml
```diff
@@ -191,6 +191,8 @@ postgresql:
         memory: 256Mi
   postgresql:
     existingSecret: mirror-passwords
+    extendedConfiguration: |
```
xin-hedera: believe we should use `postgresql.configuration`, `extendedConf` per the doc supports loading configuration from files and requires `volumePermissions.enabled` to be `true`

importer/src/main/java/org/hiero/mirror/importer/config/ImporterConfiguration.java
```diff
@@ -26,6 +32,23 @@ class ImporterConfiguration {
...
     private final ImporterProperties importerProperties;
 
+    @Bean(defaultCandidate = false)
+    @FlywayDataSource
+    DataSource flywayDataSource(
+            JdbcConnectionDetails connectionDetails, DBProperties dbProperties, HikariConfig hikariConfig) {
+        var flywayHikariConfig = new HikariConfig();
+        hikariConfig.copyStateTo(flywayHikariConfig);
+
+        flywayHikariConfig.setJdbcUrl(connectionDetails.getJdbcUrl());
```
xin-hedera: the jdbcUrl should have been already copied from the `hikariConfig` bean

jnels124: Ah yes I think you are right. I will confirm

charts/hedera-mirror/values.yaml
jnels124: we would have to provide the entire config with `postgresql.configuration`. I have a typo in the key above so can fix that if we want to keep that and enable volume permissions.  The other options is adding to a configmap via `extendedConfCM` Thoughts?

xin-hedera: that makes sense. overlooked that `configuration` completely replace the config file.

importer/src/main/java/org/hiero/mirror/importer/config/ImporterConfiguration.java
```diff
@@ -26,6 +32,23 @@ class ImporterConfiguration {
...
+        flywayHikariConfig.setUsername(dbProperties.getOwner());
+        flywayHikariConfig.setPassword(dbProperties.getOwnerPassword());
+        flywayHikariConfig.setMinimumIdle(0);
+        flywayHikariConfig.setIdleTimeout(60000);
+        flywayHikariConfig.setMaximumPoolSize(10);
+        flywayHikariConfig.setPoolName(hikariConfig.getPoolName() + "_flyway");
```
steven-sheehy: nit: Sort by setter

importer/src/main/resources/application.yml
```diff
@@ -95,7 +94,6 @@ spring:
     connectRetries: 20
     ignoreMigrationPatterns: ["*:missing", "*:ignored"]
     locations: "classpath:db/migration/common, classpath:db/migration/v1"
-    password: ${hiero.mirror.importer.db.ownerPassword}
```
steven-sheehy, line 98: Why do we remove user and password?

jnels124: because it is being configured in the importer configuration. We are setting up a different datasource (the hikari datasource) to use for flyway and put all of the config for that datasource there. Can confirm if this will still map to that datasource if we prefer to have username and password here.

jnels124: Just checked and that will not map to the datasource we create in importer configuration

charts/hedera-mirror/values.yaml
jnels124: Enabling volume permissions can have issues as well (especially adding to an existing deployment). Looking at the [startup script](https://github.com/bitnami/containers/blob/9a529b15ed928eef91ea33b78f2eaba70f972232/bitnami/postgresql/17/debian-12/rootfs/opt/bitnami/scripts/libpostgresql.sh#L1085), there is a way to configure the timezone through an env var. Tested in minikube and config works

steven-sheehy (approved):

xin-hedera (approved):
LGTM

PR #11784, opened 2025-08-15
Add design document to support Hiero hooks for mirror node

### LINKED ISSUE #11732: HIP-1195 Design document
### Problem

The mirror node needs to design our support for HIP-1195 Hooks.

### Solution

* Create a `docs/design/hooks.md`
* Update HIP with mirror node specific changes
* Create tasks that break down the work

### Alternatives

_No response_

ashumahajan:
**Description**:
Added design document to support Hiero hooks in mirror node [HIP-1195](https://hips.hedera.com/hip/hip-1195)

* Added hiero-hooks.md

**Related issue(s)**:

Fixes #11732 

**Checklist**

- [x] Documented (hiero-hooks.md)

### DIFF
--- a/docs/design/hiero-hooks.md
+++ b/docs/design/hiero-hooks.md
@@ -213,8 +213,6 @@ create table hook_storage_change
     primary key (owner_id, hook_id, key, consensus_timestamp)
 );
 
-create index hook_storage_change__hook_timestamp on hook_storage_change (hook_id, owner_id, consensus_timestamp);
-
 select create_distributed_table('hook_storage_change', 'owner_id', colocate_with = > 'entity');
 ```
 
@@ -253,9 +251,9 @@ public class Hook {
     private long contractId;
     private long createdTimestamp;
     private boolean deleted;
-    private ExtensionPoint extensionPoint;
-    private Long hookId;
-    private Long ownerId;
+    private HookExtensionPoint extensionPoint;
+    private long hookId;
+    private long ownerId;
     private HookType type;
 
     public static class Id implements Serializable {
@@ -348,7 +346,7 @@ for each hookDetails in transactionBody.hookCreationDetailsList:
             // Create historical change record for initial storage
             create HookStorageChange entity with:
                 - composite ID (hook_id, owner_id, storageUpdate.slot, consensus_timestamp)
-                - value_read = null (new storage entry)
+                - value_read = storageUpdate.value (same as written for new entries)
                 - value_written = storageUpdate.value
             save HookStorageChange via entityListener
 
@@ -579,41 +577,33 @@ domain entities and API responses.
 
 ### 3. Acceptance Tests
 
-#### Hook Creation and Retrieval
+#### Hook Lifecycle Testing
 
 ```gherkin
 Feature: Hook Management
 
-  Scenario: Complete hook lifecycle testing for account
-    Given I create an account
-    Given I creat a contract for hook
-    When I update account 'ALICE' to add hooks on the account:
+  Scenario Outline: Complete hook lifecycle testing for <owner_type>
+    Given I create a <owner_type>
+    Given I create a contract for hook
+    When I update <owner_type> to add hooks:
       | type   | extension_point        |
       | LAMBDA | ACCOUNT_ALLOWANCE_HOOK |
       | PURE   | ACCOUNT_ALLOWANCE_HOOK |
     And the mirror node processes the transactions
-    And I query mirror node REST API for account hooks
-    Then the response contains 2 hooks
-    And one hook has type "LAMBDA"
-    And one hook has type "PURE"
-
-
-  Scenario: Complete hook lifecycle testing for contract
-    Given I create a contract
-    When I update contract to add hooks on the contract:
-      | type   | extension_point        |
-      | LAMBDA | ACCOUNT_ALLOWANCE_HOOK |
-      | PURE   | ACCOUNT_ALLOWANCE_HOOK |
+    And I query mirror node REST API for <owner_type> hooks
+    Then the response contains 2 hooks with types "LAMBDA" and "PURE"
+    When I perform a CryptoTransfer that triggers hook execution for <owner_type>
     And the mirror node processes the transactions
-    And I query mirror node REST API for contract hooks
-    Then I call function with transfer that returns the balance
-    And I verify mirror node receives 2 ContractCall transaction to address '0.0.135'
-    And I query mirror node REST API tp get storage for contract hook 'LAMBDA'"
+    Then I verify mirror node receives 2 ContractCall transactions to address '0.0.365'
+    And I query mirror node REST API to get storage for <owner_type> hook 'LAMBDA'
     Then I receive storage entries
-    When I successfully update contract to delete hooks
-    And I successfully delete the Hook contracts
+    When I successfully update <owner_type> to delete hooks
+    And I successfully delete the hook contracts
 
-
+    Examples:
+      | owner_type |
+      | account    |
+      | contract   |
 ```
 
 ### 4. Test Data
--- a/docs/design/hiero-hooks.md
+++ b/docs/design/hiero-hooks.md
@@ -178,7 +178,7 @@ create type hook_extension_point as enum ('ACCOUNT_ALLOWANCE_HOOK');
 
 create table if not exists hook
 (
-    contract_id       bigint,
+    contract_id       bigint                not null,
     created_timestamp bigint                not null,
     hook_id           bigint                not null,
     owner_id          bigint                not null,
--- a/docs/design/hiero-hooks.md
+++ b/docs/design/hiero-hooks.md
@@ -178,49 +178,16 @@ create type hook_extension_point as enum ('ACCOUNT_ALLOWANCE_HOOK');
 
 create table if not exists hook
 (
-    contract_id
-    bigint
-    not
-    null,
-    created_timestamp
-    bigint
-    not
-    null,
-    hook_id
-    bigint
-    not
-    null,
-    owner_id
-    bigint
-    not
-    null,
-    extension_point
-    hook_extension_point
-    not
-    null
-    default
-    'ACCOUNT_ALLOWANCE_HOOK',
-    type
-    hook_type
-    not
-    null
-    default
-    'LAMBDA',
-    deleted
-    boolean
-    not
-    null
-    default
-    false,
-    admin_key
-    bytea,
+    contract_id       bigint                not null,
+    created_timestamp bigint                not null,
+    hook_id           bigint                not null,
+    owner_id          bigint                not null,
+    extension_point   hook_extension_point  not null default 'ACCOUNT_ALLOWANCE_HOOK',
+    type              hook_type             not null default 'LAMBDA',
+    deleted           boolean               not null default false,
+    admin_key         bytea,
 
-    primary
-    key
-(
-    owner_id,
-    hook_id
-)
+    primary key (owner_id, hook_id)
     );
 
 select create_distributed_table('hook', 'owner_id', colocate_with = > 'entity');
@@ -617,10 +584,10 @@ domain entities and API responses.
 ```gherkin
 Feature: Hook Management
 
-  Scenario: Complete hook lifecycle testing
+  Scenario: Complete hook lifecycle testing for account
     Given I create an account
     Given I creat a contract for hook
-    When I create hooks on the account:
+    When I update account 'ALICE' to add hooks on the account:
       | type   | extension_point        |
       | LAMBDA | ACCOUNT_ALLOWANCE_HOOK |
       | PURE   | ACCOUNT_ALLOWANCE_HOOK |
@@ -631,17 +598,14 @@ Feature: Hook Management
     And one hook has type "PURE"
 
 
-    # Test hook deletion
-    When I delete both hooks
+  Scenario: Complete hook lifecycle testing for contract
+    Given I create a contract
+    When I update contract to add hooks on the contract:
+      | type   | extension_point        |
+      | LAMBDA | ACCOUNT_ALLOWANCE_HOOK |
+      | PURE   | ACCOUNT_ALLOWANCE_HOOK |
     And the mirror node processes the transactions
-    And I query mirror node REST API for account hooks
-    Then the response contains 2 hooks with deleted=true
-
-  Scenario: Complete hook storage lifecycle
-    Given I create a contract with for LAMBDA hook
-    Given I create a contract with for PURE hook
-    Given I update 'ALICE' to add LAMBDA hook
-    Given I update 'ALICE' to add PURE hook
+    And I query mirror node REST API for contract hooks
     Then I call function with transfer that returns the balance
     And I verify mirror node receives 2 ContractCall transaction to address '0.0.135'
     And I query mirror node REST API tp get storage for contract hook 'LAMBDA'"
--- a/docs/design/hiero-hooks.md
+++ b/docs/design/hiero-hooks.md
@@ -192,17 +192,16 @@ For hook EVM storage data:
 -- add_hook_storage_table.sql
 create table hook_storage
 (
-    hook_id   bigint not null,
-    owner_id  bigint not null,
-    key       bytea  not null,
-    value     bytea  not null,
-    timestamp bigint not null,
+    hook_id             bigint not null,
+    owner_id            bigint not null,
+    key                 bytea  not null,
+    value               bytea  not null,
+    consensus_timestamp bigint not null,
 
     primary key (hook_id, owner_id, key)
 );
 
 create index hook_storage__hook_timestamp on hook_storage (hook_id, owner_id, timestamp);
-create index hook_storage__timestamp on hook_storage (timestamp);
 
 -- Distribute table for Citus using owner_id as distribution column (co-locate with hook table)
 select create_distributed_table('hook_storage', 'owner_id');
@@ -225,8 +224,8 @@ public class Hook {
     private Long contractId;
     private Long createdTimestamp;
     private Boolean deleted;
-    private String extensionPoint;
-    private String type;
+    private ExtensionPoint extensionPoint;
+    private HookType type;
 
     public static class HookId implements Serializable {
         private Long ownerId;
@@ -345,6 +344,9 @@ The mirror node should support **both approaches** simultaneously:
 - **Future-proofing**: Supports evolution of hook execution models
 
 ### 4. BlockItemTransformer Enhancement
+
+**Note**: Block Stream–specific changes are currently unavailable in HIP. The following represents an initial draft of
+potential changes, which will be revisited once further details are confirmed.
 
 The `BlockItemTransformer` will be enhanced to handle hook-related block items during processing:
 
--- a/docs/design/hiero-hooks.md
+++ b/docs/design/hiero-hooks.md
@@ -3,12 +3,13 @@
 ## Overview
 
 This design document outlines the implementation plan for supporting HIP-1195 hooks in the Hiero mirror node. The
-implementation will enable extraction, storage, and querying of hook data from blockchain transactions, supporting the
+implementation will enable extraction, storage, and querying of hook data from consensus node transactions, supporting
+the
 new REST APIs for hooks and hook storage.
 
 ## Goals
 
-- Enhance the database schema to store hook information, hook storage data, and historical state
+- Enhance the database schema to store hook information and hook storage data
 - Ingest hook creation, update, and deletion transactions from the record stream
 - Ingest hook storage updates via `LambdaSStoreTransactionBody` transactions and sidecars
 - Expose hook information via new REST APIs for account hooks and hook storage
@@ -30,7 +31,6 @@ The HIP-1195 implementation follows the established mirror node architecture pat
 1. **Transaction Processing**: Hook-related transactions are processed by dedicated transaction handlers
 2. **Database Storage**: Hook data is stored in normalized tables with proper indexing for efficient queries
 3. **REST APIs**: New endpoints expose hook information following existing API patterns
-4. **Historical Tracking**: Hook state changes are tracked using the existing history table pattern
 
 The implementation supports dual ingestion paths for hook storage updates:
 
@@ -75,25 +75,37 @@ Response format:
       "created_timestamp": "1726874345.123456789",
       "deleted": false,
       "extension_point": "ACCOUNT_ALLOWANCE_HOOK",
-      "hook_id": 1,
+      "hook_id": 5,
       "owner_id": "0.0.123",
       "type": "LAMBDA"
     }
   ],
   "links": {
-    "next": "/api/v1/accounts/0.0.123/hooks?timestamp=gt:1726874345.123456789&limit=1"
+    "next": "/api/v1/accounts/0.0.123/hooks?hook.id=lt:5&limit=1"
   }
 }
 ```
 
-**Note**: Hooks in the response are always sorted based on their `created_timestamp` in ascending order. The `timestamp`
-parameter in the `links.next` URL uses the format `timestamp=gt:1726874345.123456789` for pagination. _Team needs to
-review this approach._
+**Note**: Hooks in the response are always sorted based on their `hook_id` in descending order. The
+`hook.id`
+parameter in the `links.next` URL uses the format `hook.id=lt:5` for pagination.
+
+#### Query Parameters
+
+| Parameter | Type    | Description                                                                                 | Default | Validation                                                       |
+| --------- | ------- | ------------------------------------------------------------------------------------------- | ------- | ---------------------------------------------------------------- |
+| `hook.id` | string  | Filter hooks by hook ID. Supports comparison operators: `eq:`, `gt:`, `gte:`, `lt:`, `lte:` | none    | Must be a valid positive integer when using comparison operators |
+| `limit`   | integer | Maximum number of hooks to return                                                           | `25`    | Must be between 1 and 100                                        |
+| `order`   | string  | Sort order for results                                                                      | `desc`  | Must be either `asc` or `desc`                                   |
+
+**Examples:**
+
+- `/api/v1/accounts/0.0.123/hooks` - Get all hooks for account
+- `/api/v1/accounts/0.0.123/hooks?limit=10` - Get first 10 hooks
+- `/api/v1/accounts/0.0.123/hooks?hook.id=eq:5` - Get hook with ID 5
+- `/api/v1/accounts/0.0.123/hooks?hook.id=lt:10&limit=5` - Get 5 hooks with ID less than 10
 
 ### 2. Hook Storage API
-
-**Note**: We are currently trying to figure out whether to call it "state" or "storage". Will need to verify from
-Michael on what they support on their APIs.
 
 ```
 GET /api/v1/accounts/{idOrAliasOrEvmAddress}/hooks/{hook_id}/storage
@@ -118,6 +130,25 @@ Response format:
 }
 ```
 
+#### Query Parameters
+
+| Parameter | Type    | Description                                                                                               | Default | Validation                                                                             |
+| --------- | ------- | --------------------------------------------------------------------------------------------------------- | ------- | -------------------------------------------------------------------------------------- |
+| `key`     | string  | Filter storage entries by storage key. Supports comparison operators: `eq:`, `gt:`, `gte:`, `lt:`, `lte:` | none    | Must be a valid 32-byte hex string (64 hex characters) when using comparison operators |
+| `limit`   | integer | Maximum number of storage entries to return                                                               | `25`    | Must be between 1 and 100                                                              |
+| `order`   | string  | Sort order for results                                                                                    | `asc`   | Must be either `asc` or `desc`                                                         |
+
+**Examples:**
+
+- `/api/v1/accounts/0.0.123/hooks/1/storage` - Get all storage entries for hook
+- `/api/v1/accounts/0.0.123/hooks/1/storage?limit=10` - Get first 10 storage entries
+- `/api/v1/accounts/0.0.123/hooks/1/storage?key=eq:0x0000000000000000000000000000000000000000000000000000000000000001` -
+  Get specific storage entry
+-
+
+`/api/v1/accounts/0.0.123/hooks/1/storage?key=gt:0x0000000000000000000000000000000000000000000000000000000000000001&limit=5` -
+Get 5 storage entries with key greater than specified value
+
 ## Database Schema Design
 
 ### Query Requirements
@@ -125,11 +156,7 @@ Response format:
 Hooks can be fetched based on the following criteria:
 
 - **By owner_id**: Get all hooks owned by a specific account
-- **By contract_id**: Get all hooks associated with a specific contract
-- **By timestamp**: Support pagination and time-based filtering
-- **By hook_id**: Get specific hooks by their ID (combined with owner_id for uniqueness)
-- **By owner_id + created_timestamp**: Support efficient sorting and pagination of hooks by owner
-- **By contract_id + created_timestamp**: Support efficient sorting and pagination of hooks by contract
+- **By owner_id + hook_id**: Get specific hooks by their composite primary key
 
 The database indexes must support these query patterns efficiently for optimal performance.
 
@@ -138,122 +165,53 @@ The database indexes must support these query patterns efficiently for optimal p
 Create a new table to store hook information:
 
 ```sql
--- V1.111.0__add_hooks_support.sql
+-- add_hooks_support.sql
 create type hook_type as enum ('LAMBDA', 'PURE');
-create type extension_point as enum ('ACCOUNT_ALLOWANCE_HOOK');
+create type hook_extension_point as enum ('ACCOUNT_ALLOWANCE_HOOK');
 
 create table if not exists hook
 (
-    id
-    bigserial
-    primary
-    key,
-    owner_id
-    bigint
-    not
-    null,
-    hook_id
-    bigint
-    not
-    null,
-    admin_key
-    bytea,
-    contract_id
-    bigint,
-    created_timestamp
-    bigint
-    not
-    null,
-    deleted
-    boolean
-    not
-    null
-    default
-    false,
-    extension_point
-    extension_point
-    not
-    null,
-    type
-    hook_type
-    not
-    null,
-    timestamp_range
-    int8range
-    not
-    null,
+    owner_id           bigint                not null,
+    hook_id            bigint                not null,
+    admin_key          bytea,
+    contract_id        bigint,
+    created_timestamp  bigint                not null,
+    deleted            boolean               not null default false,
+    extension_point    hook_extension_point  not null,
+    type               hook_type             not null,
 
-    constraint
-    hook__owner_hook_id
-    unique
-(
-    owner_id,
-    hook_id
-)
-    );
-
--- Indexes to support the required query patterns
-create index if not exists hook__owner_id_timestamp on hook using gist (owner_id, timestamp_range);
-create index if not exists hook__owner_id_created_timestamp on hook (owner_id, created_timestamp);
-create index if not exists hook__contract_id_timestamp on hook using gist (contract_id, timestamp_range) where contract_id is not null;
-create index if not exists hook__contract_id_created_timestamp on hook (contract_id, created_timestamp) where contract_id is not null;
-```
-
-### 2. Hook History Table
-
-For historical tracking:
-
-```sql
-CREATE TABLE hook_history
-(
-    id                BIGINT    NOT NULL,
-    owner_id          BIGINT    NOT NULL,
-    hook_id           BIGINT    NOT NULL,
-    admin_key         BYTEA,
-    contract_id       BIGINT,
-    created_timestamp BIGINT    NOT NULL,
-    deleted           BOOLEAN DEFAULT FALSE,
-    extension_point   TEXT      NOT NULL,
-    type              TEXT      NOT NULL,
-    timestamp_range   INT8RANGE NOT NULL,
-
-    PRIMARY KEY (id, timestamp_range)
+    primary key (owner_id, hook_id)
 );
 
-CREATE INDEX hook_history__owner_id_timestamp ON hook_history (owner_id, timestamp_range);
+-- Index to support owner_id lookups (primary key already provides owner_id + hook_id lookups)
+create index if not exists hook__owner_id on hook (owner_id) where deleted = false;
+
+-- Distribute table for Citus using owner_id as distribution column
+select create_distributed_table('hook', 'owner_id');
 ```
 
-### 3. Hook Storage Table
+### 2. Hook Storage Table
 
 For hook EVM storage data:
 
 ```sql
--- V1.111.1__add_hook_storage_table.sql
-CREATE TABLE hook_storage
+-- add_hook_storage_table.sql
+create table hook_storage
 (
-    hook_id   BIGINT NOT NULL,
-    owner_id  BIGINT NOT NULL,
-    slot      BYTEA  NOT NULL,
-    value     BYTEA  NOT NULL,
-    timestamp BIGINT NOT NULL,
+    hook_id   bigint not null,
+    owner_id  bigint not null,
+    key       bytea  not null,
+    value     bytea  not null,
+    timestamp bigint not null,
 
-    PRIMARY KEY (hook_id, owner_id, slot),
-    CONSTRAINT fk_hook_storage__hook FOREIGN KEY (owner_id, hook_id) REFERENCES hook (owner_id, hook_id)
+    primary key (hook_id, owner_id, key)
 );
 
-CREATE INDEX hook_storage__hook_timestamp ON hook_storage (hook_id, owner_id, timestamp);
-CREATE INDEX hook_storage__timestamp ON hook_storage (timestamp);
-```
+create index hook_storage__hook_timestamp on hook_storage (hook_id, owner_id, timestamp);
+create index hook_storage__timestamp on hook_storage (timestamp);
 
-### 4. Transaction Table Enhancement
-
-Add hook_id tracking to the transaction table:
-
-```sql
--- V1.111.2__add_transaction_hook_id.sql
-ALTER TABLE transaction
-    ADD COLUMN hook_id BIGINT;
-CREATE INDEX transaction__hook_id ON transaction (hook_id) WHERE hook_id IS NOT NULL;
+-- Distribute table for Citus using owner_id as distribution column (co-locate with hook table)
+select create_distributed_table('hook_storage', 'owner_id');
 ```
 
 ## Importer Module Changes
@@ -266,24 +224,25 @@ Create new domain classes:
 
 ```java
 // common/src/main/java/org/hiero/mirror/common/domain/entity/Hook.java
-@Data
-@Entity
-@NoArgsConstructor
-@SuperBuilder(toBuilder = true)
-@Upsertable(history = true)
-public class Hook implements History {
+public class Hook {
     @Id
-    private Long id;
+    private HookId id;
 
-    private Long ownerId;
-    private Long hookId;
     private byte[] adminKey;
     private Long contractId;
     private Long createdTimestamp;
     private Boolean deleted;
     private String extensionPoint;
     private String type;
-    private Range<Long> timestampRange;
+
+    @Data
+    @Embeddable
+    @NoArgsConstructor
+    @AllArgsConstructor
+    public static class HookId implements Serializable {
+        private Long ownerId;
+        private Long hookId;
+    }
 
     // Getters, setters, and utility methods
 }
@@ -293,10 +252,6 @@ public class Hook implements History {
 
 ```java
 // common/src/main/java/org/hiero/mirror/common/domain/entity/HookStorage.java
-@Data
-@Entity
-@NoArgsConstructor
-@SuperBuilder(toBuilder = true)
 public class HookStorage {
     @Id
     private HookStorageId id;
@@ -311,93 +266,36 @@ public class HookStorage {
     public static class HookStorageId implements Serializable {
         private Long hookId;
         private Long ownerId;
-        private byte[] slot;
+        private byte[] key;
     }
 }
 ```
 
 ### 2. Transaction Handler Enhancements
 
-Modify existing transaction handlers to extract `HookCreationDetails`:
+Existing transaction handlers will be enhanced to extract hook information from transaction bodies:
 
-#### CryptoCreateTransactionHandler.java
+#### Hook Creation Processing
 
-```java
+The following transaction handlers will be modified to process `HookCreationDetails`:
 
-@Override
-private void doUpdateEntity(Entity entity, RecordItem recordItem) {
-    // ... existing code ...
-
-    processHooks(entity, recordItem);
-    entityListener.onEntity(entity);
-}
-
-private void processHooks(Entity entity, RecordItem recordItem) {
-    var transactionBody = recordItem.getTransactionBody().getCryptoCreateAccount();
-    if (!transactionBody.getHookCreationDetailsList().isEmpty()) {
-        for (var hookDetails : transactionBody.getHookCreationDetailsList()) {
-            createHook(entity, hookDetails, recordItem);
-        }
-    }
-}
-
-private void createHook(Entity entity, HookCreationDetails hookDetails, RecordItem recordItem) {
-    var hook = Hook.builder()
-            .ownerId(entity.getId())
-            .hookId(hookDetails.getHookId())
-            .extensionPoint(hookDetails.getExtensionPoint().name())
-            .createdTimestamp(recordItem.getConsensusTimestamp())
-            .deleted(false)
-            .timestampRange(Range.atLeast(recordItem.getConsensusTimestamp()))
-            .build();
-
-    // Set hook type and admin key based on hook type
-    switch (hookDetails.getHookCase()) {
-        case PURE_EVM_HOOK:
-            hook.setType("PURE");
-            var pureHook = hookDetails.getPureEvmHook();
-            if (pureHook.hasAdminKey()) {
-                hook.setAdminKey(pureHook.getAdminKey().toByteArray());
-            }
-            // Set contract ID from runtime bytecode or creation
-            break;
-        case LAMBDA_EVM_HOOK:
-            hook.setType("LAMBDA");
-            var lambdaHook = hookDetails.getLambdaEvmHook();
-            if (lambdaHook.hasAdminKey()) {
-                hook.setAdminKey(lambdaHook.getAdminKey().toByteArray());
-            }
-            // Set contract ID from runtime bytecode or creation
-            break;
-    }
-
-    entityListener.onHook(hook);
-}
-```
-
-Similar modifications will be made to:
-
+- `CryptoCreateTransactionHandler.java`
 - `CryptoUpdateTransactionHandler.java`
 - `ContractCreateTransactionHandler.java`
 - `ContractUpdateTransactionHandler.java`
 
-#### ContractCallTransactionHandler.java
+**Processing Logic:**
 
-For tracking hook-triggered contract calls:
-
-```java
-
-@Override
-private void doUpdateTransaction(Transaction transaction, RecordItem recordItem) {
-    var transactionBody = recordItem.getTransactionBody().getContractCall();
-
-    // Track hook ID that triggered this contract call
-    if (transactionBody.hasHookId()) {
-        transaction.setHookId(transactionBody.getHookId());
-    }
-
-    // ... existing code ...
-}
+```
+for each hookDetails in transactionBody.hookCreationDetailsList:
+    create Hook entity with:
+        - composite ID (owner_id, hook_id) from transaction
+        - extension point from protobuf enum
+        - hook type (PURE or LAMBDA)
+        - admin key if present
+        - contract_id = transaction entity ID if ContractCreate/Update, null if CryptoCreate/Update
+        - deleted = false
+    save Hook via entityListener
 ```
 
 ### 3. Entity Listener Enhancement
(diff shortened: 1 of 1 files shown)


jnels124 (approved):

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,912 @@
+# HIP-1195 Mirror Node Hooks Support - Design Document
+
+## Overview
+
+This design document outlines the implementation plan for supporting HIP-1195 hooks in the Hiero mirror node. The
+implementation will enable extraction, storage, and querying of hook data from blockchain transactions, supporting the
```
steven-sheehy: Prefer "consensus node" instead of blockchain

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,912 @@
...
+- Expose hook information via new REST APIs for account hooks and hook storage
+- Support efficient querying and pagination of hooks by owner, contract, and timestamp
+- Track hook usage in transactions for audit and analytics purposes
+
+## Non-Goals
+
+- Execute or validate hook logic (hooks are executed by consensus nodes)
+- Support real-time hook execution or simulation
+- Provide REST APIs for hook data
+- Store hook storage state
```
steven-sheehy: We do store and expose hook storage data, right? So this can be removed

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,912 @@
...
+      "type": "LAMBDA"
+    }
+  ],
+  "links": {
+    "next": "/api/v1/accounts/0.0.123/hooks?timestamp=gt:1726874345.123456789&limit=1"
+  }
+}
+```
+
+**Note**: Hooks in the response are always sorted based on their `created_timestamp` in ascending order. The `timestamp`
```
steven-sheehy: We should sort in descending so it always shows the latest like most of our endpoints. So we should update example to lt or specify order=asc.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,912 @@
...
+create table if not exists hook
+(
+    id
+    bigserial
+    primary
+    key,
+    owner_id
+    bigint
+    not
+    null,
```
steven-sheehy: This isn't really formatted properly to review.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,912 @@
...
+
+```sql
+-- V1.111.0__add_hooks_support.sql
+create type hook_type as enum ('LAMBDA', 'PURE');
+create type extension_point as enum ('ACCOUNT_ALLOWANCE_HOOK');
+
+create table if not exists hook
+(
+    id
+    bigserial
```
steven-sheehy: We use bigint

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,912 @@
...
+    primary
+    key,
```
steven-sheehy: As I understand it, hooks are not globally unique. The primary key has to be the account id plus the hook id.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,912 @@
...
+    hook_id           BIGINT    NOT NULL,
+    admin_key         BYTEA,
+    contract_id       BIGINT,
+    created_timestamp BIGINT    NOT NULL,
+    deleted           BOOLEAN DEFAULT FALSE,
+    extension_point   TEXT      NOT NULL,
+    type              TEXT      NOT NULL,
+    timestamp_range   INT8RANGE NOT NULL,
+
+    PRIMARY KEY (id, timestamp_range)
```
steven-sheehy: Use lowercase for all SQL

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,912 @@
...
+create index if not exists hook__contract_id_timestamp on hook using gist (contract_id, timestamp_range) where contract_id is not null;
+create index if not exists hook__contract_id_created_timestamp on hook (contract_id, created_timestamp) where contract_id is not null;
+```
+
+### 2. Hook History Table
+
+For historical tracking:
+
+```sql
+CREATE TABLE hook_history
```
steven-sheehy: A hook is immutable, right? So we don't need a history table.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,912 @@
...
+    false,
+    extension_point
+    extension_point
+    not
+    null,
+    type
+    hook_type
+    not
+    null,
+    timestamp_range
```
steven-sheehy: Remove timestamp_range since it's immutable and will only have created_timestamp.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,912 @@
...
+  ],
+  "links": {
+    "next": "/api/v1/accounts/0.0.123/hooks?timestamp=gt:1726874345.123456789&limit=1"
+  }
+}
+```
+
+**Note**: Hooks in the response are always sorted based on their `created_timestamp` in ascending order. The `timestamp`
+parameter in the `links.next` URL uses the format `timestamp=gt:1726874345.123456789` for pagination. _Team needs to
+review this approach._
```
steven-sheehy: I think it's highly likely that hooks could become mutable in the future when someone wants to update some field on it like admin key. Baking this assumption into the API that it will remain immutable seems ill-advised. We should instead use hook ID for sorting/paging even if it might mean potentially inaccurate paging if creating a new hook with an ID lower than the current max.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,912 @@
...
+  "owner_id": "0.0.123",
+  "storage": [
+    {
+      "key": "0x0000000000000000000000000000000000000000000000000000000000000000",
+      "timestamp": "1586567700.453054000",
+      "value": "0x00000000000000000000000000000000000000000000000000000000000003e8"
+    }
+  ],
+  "links": {
+    "next": "/api/v1/accounts/0.0.123/hooks/1/storage?key=gt:0x0000000000000000000000000000000000000000000000000000000000000001&limit=1"
```
steven-sheehy: Need to add a section that explicitly lists all query parameters, their description, defaults, and any validation. Same comment for both APIs.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,912 @@
...
+```
+
+**Note**: Hooks in the response are always sorted based on their `created_timestamp` in ascending order. The `timestamp`
+parameter in the `links.next` URL uses the format `timestamp=gt:1726874345.123456789` for pagination. _Team needs to
+review this approach._
+
+### 2. Hook Storage API
+
+**Note**: We are currently trying to figure out whether to call it "state" or "storage". Will need to verify from
+Michael on what they support on their APIs.
```
steven-sheehy: Believe we settled on appropriate verbiage?

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,912 @@
...
+### Query Requirements
+
+Hooks can be fetched based on the following criteria:
+
+- **By owner_id**: Get all hooks owned by a specific account
+- **By contract_id**: Get all hooks associated with a specific contract
+- **By timestamp**: Support pagination and time-based filtering
+- **By hook_id**: Get specific hooks by their ID (combined with owner_id for uniqueness)
+- **By owner_id + created_timestamp**: Support efficient sorting and pagination of hooks by owner
+- **By contract_id + created_timestamp**: Support efficient sorting and pagination of hooks by contract
```
steven-sheehy: We will only support lookup by owner_id + hook_id

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,912 @@
...
+The database indexes must support these query patterns efficiently for optimal performance.
+
+### 1. Hooks Table
+
+Create a new table to store hook information:
+
+```sql
+-- V1.111.0__add_hooks_support.sql
+create type hook_type as enum ('LAMBDA', 'PURE');
+create type extension_point as enum ('ACCOUNT_ALLOWANCE_HOOK');
```
steven-sheehy:
```suggestion
create type hook_extension_point as enum ('ACCOUNT_ALLOWANCE_HOOK');
```

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,912 @@
...
+    not
+    null,
+
+    constraint
+    hook__owner_hook_id
+    unique
+(
+    owner_id,
+    hook_id
+)
```
steven-sheehy: Remove constraint in favor of using composite primary key for these

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,912 @@
...
+    );
+
+-- Indexes to support the required query patterns
+create index if not exists hook__owner_id_timestamp on hook using gist (owner_id, timestamp_range);
+create index if not exists hook__owner_id_created_timestamp on hook (owner_id, created_timestamp);
+create index if not exists hook__contract_id_timestamp on hook using gist (contract_id, timestamp_range) where contract_id is not null;
+create index if not exists hook__contract_id_created_timestamp on hook (contract_id, created_timestamp) where contract_id is not null;
```
steven-sheehy: Remove all

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,912 @@
...
+
+CREATE INDEX hook_history__owner_id_timestamp ON hook_history (owner_id, timestamp_range);
+```
+
+### 3. Hook Storage Table
+
+For hook EVM storage data:
+
+```sql
+-- V1.111.1__add_hook_storage_table.sql
```
steven-sheehy: Remove hardcoding of migration number in design here and elsewhere.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,912 @@
...
+CREATE TABLE hook_storage
+(
+    hook_id   BIGINT NOT NULL,
+    owner_id  BIGINT NOT NULL,
+    slot      BYTEA  NOT NULL,
+    value     BYTEA  NOT NULL,
+    timestamp BIGINT NOT NULL,
+
+    PRIMARY KEY (hook_id, owner_id, slot),
+    CONSTRAINT fk_hook_storage__hook FOREIGN KEY (owner_id, hook_id) REFERENCES hook (owner_id, hook_id)
```
steven-sheehy: We don't use foreign keys.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,912 @@
...
+
+### 4. Transaction Table Enhancement
+
+Add hook_id tracking to the transaction table:
+
+```sql
+-- V1.111.2__add_transaction_hook_id.sql
+ALTER TABLE transaction
+    ADD COLUMN hook_id BIGINT;
+CREATE INDEX transaction__hook_id ON transaction (hook_id) WHERE hook_id IS NOT NULL;
```
steven-sheehy: Why do we need this? We definitely can't support another index on this able.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,912 @@
...
+
+For hook EVM storage data:
+
+```sql
+-- V1.111.1__add_hook_storage_table.sql
+CREATE TABLE hook_storage
+(
+    hook_id   BIGINT NOT NULL,
+    owner_id  BIGINT NOT NULL,
+    slot      BYTEA  NOT NULL,
```
steven-sheehy: They call it key in the HIP. There's a mapping between slot to key so they are different things.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,912 @@
...
+   state? This impacts the storage table design.
+
+## Future Enhancements
+
+### 1. Additional Extension Points
+
+- Support for new hook types as they're added
+- Dynamic extension point discovery: As new extension points are added to the network (beyond the initial
+  `ACCOUNT_ALLOWANCE_HOOK`), the mirror node should automatically recognize and process them without requiring code
+  changes. This could be achieved by:
```
steven-sheehy: This isn't really possible as protobuf is strongly typed and will require a dependency bump and changes to incorporate it.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,912 @@
...
+
+2. **Protobuf Changes and Record Stream Timeline**: Check with Michael Tinker on the availability timeline for protobuf
+   changes and record stream implementation for hook support.
+
+3. **API Naming Convention - Hook Storage vs State**: Should the API endpoints use "storage" or "state" terminology?
+   This should be aligned with what Michael Tinker is using in their consensus node APIs to maintain consistency across
+   the platform.
+
+4. **Historical Storage Queries**: Do we need to support querying historical states of hook storage, or only the current
+   state? This impacts the storage table design.
```
steven-sheehy: Yes, we will. Mirror node's job is to store all data changes over time. We don't necessarily have to expose it.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,912 @@
...
+
+## Future Enhancements
```
steven-sheehy: We should probably remove this section. This doc should be to capture the user's current requirements and provide a design to implement them. A future doc would be created for any new requirements.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,912 @@
...
+    List<Hook> findActiveHooksByContractAfterTimestamp(@Param("contractId") Long contractId,
+            @Param("timestamp") Long timestamp, Pageable pageable);
+
+    // Query by specific hook_id and owner_id
+    @Query("SELECT h FROM Hook h WHERE h.ownerId = :ownerId AND h.hookId = :hookId AND upper(h.timestampRange) IS NULL")
+    Optional<Hook> findActiveHook(@Param("ownerId") Long ownerId, @Param("hookId") Long hookId);
+
+    // Query with timestamp range filtering for pagination support
+    @Query("SELECT h FROM Hook h WHERE h.createdTimestamp > :timestampGt AND upper(h.timestampRange) IS NULL ORDER BY h.createdTimestamp")
+    List<Hook> findActiveHooksAfterTimestamp(@Param("timestampGt") Long timestampGt, Pageable pageable);
```
steven-sheehy: Design doc shouldn't have such detail as SQL queries. We can probably just note we will be adding the class but without any methods mentioned.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,912 @@
...
+            Pageable pageable);
+
+    // Query by contract_id
+    @Query("SELECT h FROM Hook h WHERE h.contractId = :contractId AND upper(h.timestampRange) IS NULL ORDER BY h.createdTimestamp")
+    List<Hook> findActiveHooksByContract(@Param("contractId") Long contractId);
+
+    // Query by contract_id with timestamp pagination
+    @Query("SELECT h FROM Hook h WHERE h.contractId = :contractId AND h.createdTimestamp > :timestamp AND upper(h.timestampRange) IS NULL ORDER BY h.createdTimestamp")
+    List<Hook> findActiveHooksByContractAfterTimestamp(@Param("contractId") Long contractId,
+            @Param("timestamp") Long timestamp, Pageable pageable);
```
steven-sheehy: We shouldn't be using Pageable as it's used with page numbers, not keyset pagination. We can try to use Scroll API but not sure if it's compatible with our code.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,912 @@
...
+Create new domain classes:
+
+#### Hook.java
+
+```java
+// common/src/main/java/org/hiero/mirror/common/domain/entity/Hook.java
+@Data
+@Entity
+@NoArgsConstructor
+@SuperBuilder(toBuilder = true)
```
steven-sheehy: We can omit all the annotations

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,912 @@
...
+
+private void processHooks(Entity entity, RecordItem recordItem) {
+    var transactionBody = recordItem.getTransactionBody().getCryptoCreateAccount();
+    if (!transactionBody.getHookCreationDetailsList().isEmpty()) {
+        for (var hookDetails : transactionBody.getHookCreationDetailsList()) {
+            createHook(entity, hookDetails, recordItem);
+        }
+    }
+}
+
```
steven-sheehy: We should omit java logic from design documents. If the logic is straightforward we can just describe it in a few sentences. If it's a bit more complex we should write simplified pseudo-code.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,912 @@
...
+```java
+
+@Override
+private void doUpdateTransaction(Transaction transaction, RecordItem recordItem) {
+    var transactionBody = recordItem.getTransactionBody().getContractCall();
+
+    // Track hook ID that triggered this contract call
+    if (transactionBody.hasHookId()) {
+        transaction.setHookId(transactionBody.getHookId());
```
steven-sheehy: There's no hookId on the contract call transaction body.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,912 @@
...
+
+@Override
+public void onHook(Hook hook) {
+    hookRepository.save(hook);
+}
+
+@Override
+public void onHookStorage(HookStorage hookStorage) {
+    hookStorageRepository.save(hookStorage);
+}
```
steven-sheehy: sql entity listener does not invoke a repository. We can just omit this as implementation detail.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,912 @@
...
+#### Transaction Type Addition
+
+The new transaction type needs to be added to the `TransactionType` enum:
+
+```java
+// Add to TransactionType.java
+LAMBDASSTORE(LambdaSStoreTransactionSupplier .class); // Replace 99 with appropriate proto ID
+```
+
+#### Transaction Supplier
```
steven-sheehy: The tx supplier and above transaction type should be in a dedicated monitor section instead of in the importer section.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,912 @@
...
+
+        // Process hook storage updates from sidecars
+        processHookStorageSidecars(recordItem);
+    }
+
+    protected void processHookStorageSidecars(RecordItem recordItem) {
+        var sidecarRecords = recordItem.getSidecarRecords();
+
+        for (var sidecar : sidecarRecords) {
+            if (sidecar.hasHookStorageUpdates()) {
```
steven-sheehy: No such field is defined in the HIP

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,912 @@
...
+```java
+// common/src/main/java/org/hiero/mirror/common/domain/transaction/Transaction.java
+@Data
+@Entity
+@NoArgsConstructor
+@SuperBuilder(toBuilder = true)
+public class Transaction {
+    // ... existing fields ...
+
+    private Long hookId;  // New field to track hook_id from LambdaSStoreTransactionBody and ContractCallTransactionBody
```
steven-sheehy: We should not add this field as it doesn't make sense. The synthetic contract call should be sufficient. It's debatable if we want to persist HookCall info to all transfer entries.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,912 @@
...
+```
+
+#### Option 2: Transaction Record Sidecars (Alternative Method)
+
+Hook storage updates can also be delivered via transaction record sidecars, similar to how contract state changes are
+handled:
+
+```java
+// Enhanced base transaction handler to process hook storage sidecars
+public abstract class AbstractTransactionHandler implements TransactionHandler {
```
steven-sheehy: We should not add logic to all handlers that only applies to one type.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,912 @@
...
+
+### 3. Response Models
+
+#### HooksResponse.java
+
+```java
+// rest-java/src/main/java/org/hiero/mirror/restjava/dto/HooksResponse.java
+@Data
+@Builder
+public class HooksResponse {
```
steven-sheehy: We shouldn't need DTOs if we use mapstruct mapper and generated OpenAPI models.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,912 @@
...
+    public HooksResponse getAccountHooks(Long accountId, String hookIdParam, int limit, String order) {
+        // Validate account exists
+        entityService.lookup(accountId);
+
+        // Parse hook ID filter if provided
+        var hooks = hookRepository.findActiveHooksByOwner(accountId);
+
+        // Apply filtering and pagination
+        // Build response with proper linking
+        return buildHooksResponse(hooks, accountId, hookIdParam, limit, order);
```
steven-sheehy: Too much code in design. Distill to simplified pseudo code similar to your java comments.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,912 @@
...
+### 1. Hooks Table
+
+Create a new table to store hook information:
+
+```sql
+-- V1.111.0__add_hooks_support.sql
+create type hook_type as enum ('LAMBDA', 'PURE');
+create type extension_point as enum ('ACCOUNT_ALLOWANCE_HOOK');
+
+create table if not exists hook
```
steven-sheehy, line 179: We're missing how we would distribute or shard the data for citus.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,912 @@
...
+        private byte[] slot;
+    }
+}
+```
+
+### 2. Transaction Handler Enhancements
+
+Modify existing transaction handlers to extract `HookCreationDetails`:
+
+#### CryptoCreateTransactionHandler.java
```
steven-sheehy: Missing section for BlockItemTransformer

steven-sheehy (changes requested):

docs/design/hiero-hooks.md
ashumahajan: yes

docs/design/hiero-hooks.md
ashumahajan: Should I then keep the hook_history table?

docs/design/hiero-hooks.md
ashumahajan: We will still need one for owner_id for fetching all hooks for owner?

docs/design/hiero-hooks.md
ashumahajan: This could support fetching transations like Crypto/Contract create update transactions for hook_id. I will remove it

docs/design/hiero-hooks.md
steven-sheehy: YAGNI

docs/design/hiero-hooks.md
steven-sheehy: The primary key will be owner_id+hook_id composite key which is also suitable for looking up just by owner_id.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,912 @@
...
+    @Builder
+    public static class StorageEntryDto {
+        private String key;
+        private String timestamp;
+        private String value;
+    }
+}
+```
+
+## Testing Strategy
```
steven-sheehy, line 564: We should specify the expected acceptance tests we want to write with pseudocode like in our feature files.

docs/design/hiero-hooks.md
ashumahajan: Not sure about this comment

steven-sheehy: These changes are associated with the monitor module, not importer. So should move to the end in a `## Monitor` section

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,584 @@
...
+[HIP-1195](https://hips.hedera.com/hip/hip-1195) introduces programmable hooks to the Hiero network, allowing users to
+customize behavior at specific
+extension points. Hooks can be attached to accounts and contracts via:
+
+- `CryptoCreateTransactionBody`
+- `CryptoUpdateTransactionBody`
+- `ContractCreateTransactionBody`
+- `ContractUpdateTransactionBody`
+
+The `LambdaSStoreTransactionBody` and `ContractCallTransactionBody` will include a `hook_id` field to indicate which
```
xin-hedera: I didn't see in the HIP that `ContractCallTransactionBody` will include a `hook_id` field.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,584 @@
...
+
+**Note**: Hooks in the response are always sorted based on their `hook_id` in descending order. The
+`hook.id`
+parameter in the `links.next` URL uses the format `hook.id=lt:5` for pagination.
+
+#### Query Parameters
+
+| Parameter | Type    | Description                                                                                 | Default | Validation                                                       |
+| --------- | ------- | ------------------------------------------------------------------------------------------- | ------- | ---------------------------------------------------------------- |
+| `hook.id` | string  | Filter hooks by hook ID. Supports comparison operators: `eq:`, `gt:`, `gte:`, `lt:`, `lte:` | none    | Must be a valid positive integer when using comparison operators |
```
xin-hedera, line 95: nit: hook id data type is uint64, so technically `0` is a valid hook id? not a big deal and can always update later if that's the case.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,584 @@
...
+
+```sql
+-- add_hook_storage_table.sql
+create table hook_storage
+(
+    hook_id   bigint not null,
+    owner_id  bigint not null,
+    key       bytea  not null,
+    value     bytea  not null,
+    timestamp bigint not null,
```
xin-hedera:
```suggestion
    consensus_timestamp bigint not null,
```

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,584 @@
...
+
+    primary key (hook_id, owner_id, key)
+);
+
+create index hook_storage__hook_timestamp on hook_storage (hook_id, owner_id, timestamp);
+create index hook_storage__timestamp on hook_storage (timestamp);
```
xin-hedera: not sure where `hook_storage__timestamp` is used so perhaps we don't need it.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,584 @@
...
+
+-- Distribute table for Citus using owner_id as distribution column (co-locate with hook table)
+select create_distributed_table('hook_storage', 'owner_id');
```
xin-hedera: shouldn't we have two tables?

- `hook_storage_change`: rename the current table `hook_storage`
- `hook_storage`: this is the latest state, so as to make the hook storage rest API fast

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,584 @@
...
+// common/src/main/java/org/hiero/mirror/common/domain/entity/Hook.java
+public class Hook {
+    private HookId id;
+
+    private byte[] adminKey;
+    private Long contractId;
+    private Long createdTimestamp;
+    private Boolean deleted;
+    private String extensionPoint;
+    private String type;
```
xin-hedera: should both fields' type be enum instead?

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,584 @@
...
+**Processing Logic:**
+
+```
+for each hookDetails in transactionBody.hookCreationDetailsList:
+    create Hook entity with:
+        - composite ID (owner_id, hook_id) from transaction
+        - extension point from protobuf enum
+        - hook type (PURE or LAMBDA)
+        - admin key if present
+        - contract_id = transaction entity ID if ContractCreate/Update, null if CryptoCreate/Update
```
xin-hedera: isn't the `contractId` inside `HookCreationDetails`, either `pure_evm_hook.spec.contract_id` or `lambda_evm_hook .spec.contract_id`?

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,584 @@
...
+        set composite ID (owner_id, hook_id)
+        map extension point enum
+        extract admin key if present
+        pass to EntityListener
+
+if (transactionBody is LambdaSStoreTransactionBody):
+    for each storageEntry in storageUpdatesList:
+        transform to HookStorage entity
+        set composite ID (hook_id, owner_id, key)
+        pass to EntityListener
```
xin-hedera: we don't need to transform anything in transaction body. From block streams to record streams, in general, transformation happens in block stream transactionOutput/transactionResult/traceData/stateChanges to record stream transactionRecord / sidecarTransactionRecord.

xin-hedera: follow-up, I don't see any block stream specific changes is needed to support hooks atm. If `EvmTraceData` and storage slot statechanges together can't fully support hooks storage changes, the execution team should update the HIP to include any blockstream specific protobuf changes and we can update the design then.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,584 @@
...
+
+```json
+{
+  "hooks": [
+    {
+      "admin_key": {
+        "_type": "ED25519",
+        "key": "0x302a300506032b6570032100e5b2…"
+      },
+      "contract_id": "0.0.456",
```
bilyana-gospodinova, line 72: In the HIP this parameter is named "hook_contract_id".  Also, there is "num_storage_slots", the "admin_key" is "storage_key". There is also: 
"links": {
        "self": "/api/v1/accounts/0.0.123/hooks/1",
        "contract": "/api/v1/contracts/0.0.456",
        "storage": "/api/v1/accounts/0.0.123/hooks/1/storage"
      }
      
  Is there a particular reason why the design doc is different?

docs/design/hiero-hooks.md
ashumahajan: This is based on our discussion with Michael

docs/design/hiero-hooks.md
ashumahajan: I don't think the requirement is about displaying just the current storage. I think it will be a overkill to have a separate table.

docs/design/hiero-hooks.md
ashumahajan: HookCreationDetails  is internal to evm and will not be available to us.

docs/design/hiero-hooks.md
ashumahajan: 1. contract_id is part of hook object.
2. We distributed the apis to now have 2 apis instead of one. storage_key is part of storage api based on protobuf definition and links are based on current api structure in mirror node.
3. @steven-sheehy Do you remember what is the reason for num_storage_slots not being included. Based on what i remember, this is part of EvmHookState which will not be available to mirror node.

docs/design/hiero-hooks.md
xin-hedera: In the doc the Hooks storage API doesn't support a timestamp query parameter, I take that as it always returns the current storage.

Given the current table design, there will be multiple rows for a `(hook_id, owner_id, key)` storage slot, and then the API needs to do a unique by the storage slot and get the latest by timestamp query, and it'll be really slow.

docs/design/hiero-hooks.md
xin-hedera: that contradicts with what's in the HIP. e.g, the protobuf messages changes highlighted in the HIP:

```protobuf
message CryptoCreateTransactionBody {
  // ...

  /**
   * Details of hooks to add immediately after creating this account.
   */
  repeated HookCreationDetails hook_creation_details = 19;
}

```

docs/design/hiero-hooks.md
ashumahajan: The storage api returns the list of storage updates. May be those should be sorted based on timestamp

docs/design/hiero-hooks.md
ashumahajan: you are right. It will be available

docs/design/hiero-hooks.md
xin-hedera: umm, I think it should return the storage state instead of changes. similar to `/api/v1/contracts/{contractIdOrAddress}/state`

ashumahajan: hmm. That will be a different api in that case. There is no requirement for next-link in that case

docs/design/hiero-hooks.md
ashumahajan: Checked with Neeha and this is not a correct assumption. Updated the document based on updated understanding

docs/design/hiero-hooks.md
ashumahajan: Updated as per your suggestions

docs/design/hiero-hooks.md
steven-sheehy: The API in the HIP was just an initial draft by Michael. The design in this doc is the canonical source which once approved we'll push to the HIP.

We chose not to include num_storage_slots since it would require upserts to hook entity for every cryptotransfer hook call. That would be slow at scale.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,721 @@
...
+the
+new REST APIs for hooks and hook storage.
+
+## Goals
+
+- Enhance the database schema to store hook information and hook storage data
+- Ingest hook creation, update, and deletion transactions from the record stream
+- Ingest hook storage updates via `LambdaSStoreTransactionBody` transactions and sidecars
+- Expose hook information via new REST APIs for account hooks and hook storage
+- Support efficient querying and pagination of hooks by owner, contract, and timestamp
```
steven-sheehy: We don't have any filters for contract ID.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,721 @@
...
+  "owner_id": "0.0.123",
+  "storage": [
+    {
+      "key": "0x0000000000000000000000000000000000000000000000000000000000000000",
+      "value": "0x00000000000000000000000000000000000000000000000000000000000003e8",
+      "timestamp": "1726874345.123456789"
+    }
+  ],
+  "links": {
+    "next": "/api/v1/accounts/0.0.123/hooks/1/storage?key=gt:0x0000000000000000000000000000000000000000000000000000000000000001&limit=1"
```
steven-sheehy: Off by one error

```suggestion
    "next": "/api/v1/accounts/0.0.123/hooks/1/storage?key=gt:0x0000000000000000000000000000000000000000000000000000000000000000&limit=1"
```

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,721 @@
...
+create table if not exists hook
+(
+    owner_id          bigint                not null,
+    hook_id           bigint                not null,
+    admin_key         bytea,
+    contract_id       bigint,
+    created_timestamp bigint                not null,
+    deleted           boolean               not null default false,
+    extension_point   hook_extension_point  not null,
+    type              hook_type             not null,
```
steven-sheehy: This order wastes 10 bytes per row on disk due to PostgreSQL bit packing. Please read https://www.enterprisedb.com/blog/rocks-and-sand for more info.

```sql
select
pg_column_size(row(0::bigint, 0::BIGINT, ''::bytea, 0::bigint, 0::bigint, false::boolean, 'LAMBDA'::hook_type, 'ACCOUNT_ALLOWANCE_HOOK'::hook_extension_point)) AS original,
pg_column_size(row(0::bigint, 0::BIGINT, 0::bigint, 0::bigint, 'LAMBDA'::hook_type, 'ACCOUNT_ALLOWANCE_HOOK'::hook_extension_point, false::boolean, ''::bytea)) AS reordered;
```
original | reordered
-|-
76 | 66

```suggestion
    contract_id       bigint,
    created_timestamp bigint                not null,
    hook_id           bigint                not null,
    owner_id          bigint                not null,
    extension_point   hook_extension_point  not null,
    type              hook_type             not null,
    deleted           boolean               not null default false,
    admin_key         bytea,
```

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,721 @@
...
+
+```sql
+-- add_hook_storage_change_table.sql
+create table hook_storage_change
+(
+    hook_id             bigint not null,
+    owner_id            bigint not null,
+    key                 bytea  not null,
+    value               bytea  not null,
+    consensus_timestamp bigint not null,
```
steven-sheehy: Sort and bit packing.

```suggestion
    consensus_timestamp bigint not null,
    hook_id             bigint not null,
    owner_id            bigint not null,
    key                 bytea  not null,
    value               bytea  not null,
```

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,721 @@
...
+```sql
+-- add_hook_storage_table.sql
+create table hook_storage
+(
+    hook_id            bigint not null,
+    owner_id           bigint not null,
+    key                bytea  not null,
+    value              bytea  not null,
+    created_timestamp  bigint not null,
+    modified_timestamp bigint not null,
```
steven-sheehy: The rows are immutable so we just need consensus_timestamp. Also reorder for bit packing.

```suggestion
    consensus_timestamp bigint not null,
    hook_id             bigint not null,
    owner_id            bigint not null,
    key                 bytea  not null,
    value               bytea  not null,
```

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,721 @@
...
+    value               bytea  not null,
+    consensus_timestamp bigint not null,
+
+    primary key (hook_id, owner_id, key, consensus_timestamp)
+);
+
+create index hook_storage_change__hook_timestamp on hook_storage_change (hook_id, owner_id, consensus_timestamp);
+
+-- Distribute table for Citus using owner_id as distribution column (co-locate with hook table)
+select create_distributed_table('hook_storage_change', 'owner_id');
```
steven-sheehy: We should `colocate_with => 'entity'` all 3 tables.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,721 @@
...
+#### 2.1. Hook Storage Change Table (Historical)
+
+```sql
+-- add_hook_storage_change_table.sql
+create table hook_storage_change
+(
+    hook_id             bigint not null,
+    owner_id            bigint not null,
+    key                 bytea  not null,
+    value               bytea  not null,
```
steven-sheehy: Will we get value_read vs value_written? Probably so since it comes from same sidecar data.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,721 @@
...
+
+```java
+// common/src/main/java/org/hiero/mirror/common/domain/entity/Hook.java
+public class Hook {
+    private HookId id;
+
+    private byte[] adminKey;
+    private Long contractId;
+    private Long createdTimestamp;
+    private Boolean deleted;
```
steven-sheehy: We should use primitives if they're not nullable via schema

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,721 @@
...
+    private ExtensionPoint extensionPoint;
+    private HookType type;
+
+    public static class HookId implements Serializable {
+        private Long ownerId;
+        private Long hookId;
```
steven-sheehy: Technically you have to duplicate these values in both outer class and ID class. Same for other domains.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,721 @@
...
+```
+for each hookDetails in transactionBody.hookCreationDetailsList:
+    create Hook entity with:
+        - composite ID (owner_id, hook_id) from transaction
+        - extension point from protobuf enum
+        - hook type (PURE or LAMBDA)
+        - admin key if present
+        - contract_id
+        - deleted = false
+    save Hook via entityListener
```
steven-sheehy, line 341: We will also have hook storage changes if lambda storage updates that needs to be mentioned.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,721 @@
...
+
+    void onHookStorage(HookStorage hookStorage);
+
+    void onHookStorageChange(HookStorageChange hookStorageChange);
+}
+```
+
+### 3.1. Hook Execution Processing
+
+Hook execution appears in the record stream as `ContractCall` transactions to system contract `0.0.365` (EVM address `0x16d`) as children of the triggering transaction. These can be identified by checking `ContractCallTransactionBody.contractID == 0.0.365` and the presence of `parentConsensusTimestamp`. The execution order follows a specific sequence based on the `CryptoTransferTransactionBody` structure:
```
steven-sheehy: Not just presence of parentConsensusTimestamp but that its value is equal to the consensusTimestamp of the crypto transfer. There can be other contract calls that are children of the hook calls.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,721 @@
...
+- API availability must match existing REST API SLAs
+- Data consistency must be maintained across hook, hook_storage, and transaction tables
+
+## Open Questions
+
+1. **Indexing Strategy**: Which additional indexes might be needed for complex hook queries that we haven't identified
+   yet?
+
+2. **Protobuf Changes and Record Stream Timeline**: Check with Michael Tinker on the availability timeline for protobuf
+   changes and record stream implementation for hook support.
```
steven-sheehy: I don't think needs to be mentioned in design.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,721 @@
...
+### 3. Reliability Requirements
+
+- Hook data ingestion must maintain the same reliability guarantees as existing transaction types
+- API availability must match existing REST API SLAs
+- Data consistency must be maintained across hook, hook_storage, and transaction tables
+
+## Open Questions
+
+1. **Indexing Strategy**: Which additional indexes might be needed for complex hook queries that we haven't identified
+   yet?
```
steven-sheehy: YAGNI. We don't need to prematurely consider or add indexes for features that haven't been requested yet.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,721 @@
...
+
+#### Hook Storage Management
+
+**Note**: Will not be part of first phase
+
+```gherkin
+Feature: Hook Storage
+
+  Scenario: Hook storage updates via LambdaSStoreTransaction
+    Given a hook exists with owner_id 123 and hook_id 1
```
steven-sheehy: We can't hardcode an account ID.

```suggestion
    Given a hook exists with owner 'ALICE' and hook ID 1
```

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,721 @@
...
+    When the mirror node processes the transactions
+    And I query "/api/v1/accounts/{account_id}/hooks"
+    Then the response contains 2 hooks
+    And both hooks have null contract_id since they belong to an account
+    And one hook has type "LAMBDA"
+    And one hook has type "PURE"
+
+
+  Scenario: Hook pagination works correctly
+    Given consensus nodes create an account with 50 hooks
```
steven-sheehy: We can't waste hbars creating 50 hooks. We should consolidate all these scenarios to create one lambda hook and one pure hook and perform various CRUD operations on it before we delete them.

steven-sheehy (changes requested):

docs/design/hiero-hooks.md
ashumahajan: I will change as per contract_state_change table. But still don't understand the logic. What is the reason for this?

docs/design/hiero-hooks.md
ashumahajan: This section reflects for purely when hook is created for 
- CryptoCreateTransactionHandler.java
- CryptoUpdateTransactionHandler.java
- ContractCreateTransactionHandler.java
- ContractUpdateTransactionHandler.java

steven-sheehy: And where will the `storage_updates` be persisted during create and update?

```
message LambdaEvmHook {
  EvmHookSpec spec = 1;
  repeated LambdaStorageUpdate storage_updates = 2;
}
```

docs/design/hiero-hooks.md
steven-sheehy: It comes that way in the sidecar and shows contract storage values that were both read and written during a contract call.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,764 @@
...
+create table hook_storage_change
+(
+    consensus_timestamp bigint not null,
+    hook_id             bigint not null,
+    owner_id            bigint not null,
+    key                 bytea  not null,
+    value_read          bytea  not null,
+    value_written       bytea,
+
+    primary key (hook_id, owner_id, key, consensus_timestamp)
```
steven-sheehy: First column is very important in PostgreSQL

```suggestion
    primary key (owner_id, hook_id, key, consensus_timestamp)
```

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,764 @@
...
+-- add_hook_storage_table.sql
+create table hook_storage
+(
+    consensus_timestamp bigint not null,
+    hook_id             bigint not null,
+    owner_id            bigint not null,
+    key                 bytea  not null,
+    value               bytea  not null,
+
+    primary key (hook_id, owner_id, key)
```
steven-sheehy:
```suggestion
    primary key (owner_id, hook_id, key)
```

docs/design/hiero-hooks.md
steven-sheehy: Not addressed

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,764 @@
...
+                - value_written = storageUpdate.value
+            save HookStorageChange via entityListener
+
+            // Create current state record for initial storage
+            create HookStorage entity with:
+                - composite ID (hook_id, owner_id, storageUpdate.slot)
+                - value = storageUpdate.value
+                - created_timestamp = consensus_timestamp
+                - modified_timestamp = consensus_timestamp
+            save HookStorage via entityListener
```
steven-sheehy: Like ContractStateChange, we only need a onHookStorageChange() then generate the HookStorage if valueWritten not null in SqlEntityListener. Also, the processing of sidecar info occurs in ContractResultServiceImpl not the listeners.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,764 @@
...
+
+```gherkin
+Feature: Hook Management
+
+  Scenario: Complete hook lifecycle testing
+    Given I create an account
+    When I create hooks on the account:
+      | type   | extension_point        | admin_key |
+      | LAMBDA | ACCOUNT_ALLOWANCE_HOOK | key123    |
+      | PURE   | ACCOUNT_ALLOWANCE_HOOK | key456    |
```
steven-sheehy: key123 isn't a valid format key. Probably best to just omit it from spec since we can't update hooks anyway

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,764 @@
...
+Create a new table to store hook information:
+
+```sql
+-- add_hooks_support.sql
+create type hook_type as enum ('LAMBDA', 'PURE');
+create type hook_extension_point as enum ('ACCOUNT_ALLOWANCE_HOOK');
+
+create table if not exists hook
+(
+    contract_id       bigint,
```
steven-sheehy: Currently the only way to create a hook and populate its bytecode is via a contract ID

```suggestion
    contract_id       bigint not null,
```

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,764 @@
...
+    created_timestamp bigint                not null,
+    hook_id           bigint                not null,
+    owner_id          bigint                not null,
+    extension_point   hook_extension_point  not null,
+    type              hook_type             not null,
```
steven-sheehy: Add defaults

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,764 @@
...
+### 1. Domain Models
+
+Create new domain classes:
+
+#### Hook.java
+
+```java
+// common/src/main/java/org/hiero/mirror/common/domain/entity/Hook.java
+public class Hook {
+    private HookId id;
```
steven-sheehy: The HookId id field is not present in the class. See AbstractNft for an example. Same comment for other classes.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,764 @@
...
+
+    // Duplicate ID fields in outer class for JPA
+    private Long ownerId;
+    private Long hookId;
```
steven-sheehy: Remove grouping and sort all fields in domain class. Same comment for other classes.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,764 @@
...
+Extend `EntityListener` interface to handle hooks:
+
+```java
+// importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/EntityListener.java
+public interface EntityListener {
+    // ... existing methods ...
+
+    void onHook(Hook hook);
+
+    void onHookStorage(HookStorage hookStorage);
```
steven-sheehy: Remove

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,764 @@
...
+    if (transactionRecord.hasParentConsensusTimestamp()):
+        parentTransaction = context.get(Transaction.class, parentConsensusTimestamp)
+        if (parentTransaction != null && parentTransaction.consensusTimestamp == transaction.parentConsensusTimestamp):
+            // Verify this is a direct hook execution, not a nested contract call
+            // Determine hook execution context using parent transaction body and child index
+            hookExecutionIndex = getChildTransactionIndex(transaction, parentTransaction)
+            hookContext = deriveHookFromCryptoTransferOrder(parentTransaction.body, hookExecutionIndex)
+
+            // Process sidecar ContractStateChanges for contract 0.0.365
+            for each stateChange in transaction.sidecars where contractId == 365:
```
steven-sheehy: We already link them up based upon consensus timestamp so the check for 365 again is redundant.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,764 @@
...
+                    - owner_id from hookContext
+                    - storage key from state change
+                    - value_read from stateChange.valueRead
+                    - value_written from stateChange.valueWritten
+                    - consensus_timestamp from current transaction
+
+                // Also update current state table with effective value
+                effectiveValue = valueWritten != null ? valueWritten : valueRead
+                create/update HookStorage entity with:
+                    - value = effectiveValue
```
steven-sheehy: This goes in listener

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,764 @@
...
+```
+if (transactionBody has HookCreationDetails):
+    for each hookDetails in hookCreationDetailsList:
+        transform hookDetails to Hook entity
+        set composite ID (owner_id, hook_id)
+        map extension point enum
+        extract admin key if present
+        pass to EntityListener
+
+if (transactionBody is LambdaSStoreTransactionBody):
```
steven-sheehy: As Xin mentioned we probably don't need custom logic for processing hooks in block streams. Anything in body would continue to be processed by the handlers. And we already process evm trace data to sidecar info

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,764 @@
...
+
+A `HookService` will implement the business logic for hook queries:
+
+**Account Hooks Service Logic:**
+
+- Validate that the account exists
+- Query hooks by owner_id with deletion filtering
+- Apply hook.id comparison filters (eq, gt, lt, etc.)
+- Implement pagination using hook_id-based cursors
+- Build response with proper next page links
```
steven-sheehy: The links will be generated in controller layer using linkfactory

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,764 @@
...
+    # Test hook pagination with limit
+    When I query "/api/v1/accounts/{account_id}/hooks?limit=1&order=asc"
+    Then I receive 1 hook
+    And the response includes a next page link
+
+    # Test hook updates
+    When I update the LAMBDA hook with new admin key "newkey789"
+    And the mirror node processes the transactions
+    And I query "/api/v1/accounts/{account_id}/hooks?hook.id=eq:{lambda_hook_id}"
+    Then the hook has admin key "newkey789"
```
steven-sheehy: There is no hook update functionality outside its storage. Hooks are immutable.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,764 @@
...
+    # Test hook deletion
+    When I delete both hooks
+    And the mirror node processes the transactions
+    And I query "/api/v1/accounts/{account_id}/hooks"
+    Then the response contains 2 hooks with deleted=true
+```
+
+#### Hook Storage Management
+
+**Note**: Will not be part of first phase
```
steven-sheehy: Why? There are no hook phases. All work will be done in a single consensus node release.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,764 @@
...
+  Scenario: Complete hook lifecycle testing
+    Given I create an account
+    When I create hooks on the account:
+      | type   | extension_point        | admin_key |
+      | LAMBDA | ACCOUNT_ALLOWANCE_HOOK | key123    |
+      | PURE   | ACCOUNT_ALLOWANCE_HOOK | key456    |
+    And the mirror node processes the transactions
+    And I query "/api/v1/accounts/{account_id}/hooks"
+    Then the response contains 2 hooks
+    And both hooks have null contract_id since they belong to an account
```
steven-sheehy: The hook.contract_id will always be populated with the template contract ID. The owner_id will refer to either the owning account or owning contract.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,764 @@
...
+
+### 3. Acceptance Tests
+
+#### Hook Creation and Retrieval
+
+```gherkin
+Feature: Hook Management
+
+  Scenario: Complete hook lifecycle testing
+    Given I create an account
```
steven-sheehy: A hook will also require a contract that implements the hook interface.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,764 @@
...
+    Given I create a contract with for PURE hook
+    Given I update 'ALICE' to add LAMBDA hook
+    Given I update 'ALICE' to add PURE hook
+    # Test storage updates via hook execution during CryptoTransfer
+    When I perform a CryptoTransfer from 'ALICE' that triggers the hooks
+    And I verify mirror node receives 2 ContractCall transaction to address '0.0.135'
+    And I query "/api/v1/accounts/0.0.123/hooks/1/storage"
+    Then I receive storage entries
+    When I successfully update 'ALICE' to delete hooks
+    And I successfully delete the Hook contracts
```
steven-sheehy: We should merge the hook CRUD with hook call tests so we minimize entities

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,764 @@
...
+- Transaction processing tests
+
+### 3. Acceptance Tests
+
+#### Hook Creation and Retrieval
+
+```gherkin
+Feature: Hook Management
+
+  Scenario: Complete hook lifecycle testing
```
steven-sheehy: Probably we'll want to matrix the entire thing for ACCOUNT and CONTRACT as owners. Also, we can probably skip testing both create and update of each and only test update.

docs/design/hiero-hooks.md
ashumahajan: This Was based on my previous understanding which Neeha clarified. Will update.

docs/design/hiero-hooks.md
ashumahajan: Since this was just representation and we removed the annotations. I thought it would be good to add the id field for reference. I will remove it

docs/design/hiero-hooks.md
ashumahajan: One of the things I think we are missing in apis in contract/{contractIdOrAddress}/hooks, when the owner is contract. DO you think there should be another api for that?

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,673 @@
...
+    hook_id             bigint not null,
+    owner_id            bigint not null,
+    key                 bytea  not null,
+    value_read          bytea  not null,
+    value_written       bytea,
+
+    primary key (owner_id, hook_id, key, consensus_timestamp)
+);
+
+create index hook_storage_change__hook_timestamp on hook_storage_change (hook_id, owner_id, consensus_timestamp);
```
steven-sheehy: Why do we need this index that duplicates the primary key but with a different column order?

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,673 @@
...
+        - deleted = false
+    save Hook via entityListener
+
+    // Process initial storage updates from LambdaEvmHook.storage_updates
+    if (hookDetails.type == LAMBDA && hookDetails.lambdaEvmHook.storage_updates is not empty):
+        for each storageUpdate in hookDetails.lambdaEvmHook.storage_updates:
+            // Create historical change record for initial storage
+            create HookStorageChange entity with:
+                - composite ID (hook_id, owner_id, storageUpdate.slot, consensus_timestamp)
+                - value_read = null (new storage entry)
```
steven-sheehy: value_read is not nullable. Should put same value for both read and written.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,673 @@
...
+@IdClass(Hook.Id.class)
+public class Hook {
+
+    private byte[] adminKey;
+    private long contractId;
+    private long createdTimestamp;
+    private boolean deleted;
+    private ExtensionPoint extensionPoint;
+    private Long hookId;
+    private Long ownerId;
```
steven-sheehy:
```suggestion
    private long hookId;
    private long ownerId;
```

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,673 @@
...
+```java
+// common/src/main/java/org/hiero/mirror/common/domain/entity/Hook.java
+@IdClass(Hook.Id.class)
+public class Hook {
+
+    private byte[] adminKey;
+    private long contractId;
+    private long createdTimestamp;
+    private boolean deleted;
+    private ExtensionPoint extensionPoint;
```
steven-sheehy:
```suggestion
    private HookExtensionPoint extensionPoint;
```

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,673 @@
...
+### 3. Acceptance Tests
+
+#### Hook Creation and Retrieval
+
+```gherkin
+Feature: Hook Management
+
+  Scenario: Complete hook lifecycle testing for account
+    Given I create an account
+    Given I creat a contract for hook
```
steven-sheehy:
```suggestion
    Given I create a contract that implements the hook interface
```

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,673 @@
...
+    When I update account 'ALICE' to add hooks on the account:
+      | type   | extension_point        |
+      | LAMBDA | ACCOUNT_ALLOWANCE_HOOK |
+      | PURE   | ACCOUNT_ALLOWANCE_HOOK |
+    And the mirror node processes the transactions
+    And I query mirror node REST API for account hooks
+    Then the response contains 2 hooks
+    And one hook has type "LAMBDA"
+    And one hook has type "PURE"
```
steven-sheehy: Think we can combine these three since they're all verifying the same response.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,673 @@
...
+  Scenario: Complete hook lifecycle testing for contract
+    Given I create a contract
+    When I update contract to add hooks on the contract:
+      | type   | extension_point        |
+      | LAMBDA | ACCOUNT_ALLOWANCE_HOOK |
+      | PURE   | ACCOUNT_ALLOWANCE_HOOK |
+    And the mirror node processes the transactions
+    And I query mirror node REST API for contract hooks
+    Then I call function with transfer that returns the balance
+    And I verify mirror node receives 2 ContractCall transaction to address '0.0.135'
```
steven-sheehy: What is 0.0.135? We should send back to our payer account

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,673 @@
...
+  Scenario: Complete hook lifecycle testing for account
+    Given I create an account
+    Given I creat a contract for hook
+    When I update account 'ALICE' to add hooks on the account:
+      | type   | extension_point        |
+      | LAMBDA | ACCOUNT_ALLOWANCE_HOOK |
+      | PURE   | ACCOUNT_ALLOWANCE_HOOK |
+    And the mirror node processes the transactions
+    And I query mirror node REST API for account hooks
+    Then the response contains 2 hooks
```
steven-sheehy: Why do we not invoke hook, verify storage, and delete hooks for account scenario, but we do for contract? The scenarios should be the same hence why I didn't know if we wanted to matrix the entire scenario with ACCOUNT, CONTRACT.

docs/design/hiero-hooks.md
```diff
@@ -0,0 +1,673 @@
...
+    And the mirror node processes the transactions
+    And I query mirror node REST API for contract hooks
+    Then I call function with transfer that returns the balance
+    And I verify mirror node receives 2 ContractCall transaction to address '0.0.135'
+    And I query mirror node REST API tp get storage for contract hook 'LAMBDA'"
+    Then I receive storage entries
+    When I successfully update contract to delete hooks
+    And I successfully delete the Hook contracts
+
+
```
steven-sheehy: Extra whitespace

```suggestion
```

docs/design/hiero-hooks.md
steven-sheehy: Contracts are also accounts. We can currently query a contract via /api/v1/accounts/{contractId} so guess we can assume the same for hooks.

docs/design/hiero-hooks.md
ashumahajan: It was a mistake. My IDE always undo the ddl format, so I have to make changes and for that I go back to old document and copy from there. Will figure out a way to resolve

docs/design/hiero-hooks.md
ashumahajan: 0.0.135 was suppose to be 0.0.365

PR #11812, opened 2025-08-21
Update web3 HPA

### LINKED ISSUE #11692: Web3 gas used based HPA target causes created pods spike when rolling out update
### Description

When upgrading the web3 deployment, the new gas based HPA target causes a spike of new pods.

```
Events:
  Type     Reason               Age                 From                       Message
  ----     ------               ----                ----                       -------
  Normal   SuccessfulRescale    88s (x2 over 123m)  horizontal-pod-autoscaler  New size: 3; reason: pods metric evm_gas_used_per_second above target
  Normal   SuccessfulRescale    72s                 horizontal-pod-autoscaler  New size: 4; reason: pods metric evm_gas_used_per_second above target
  Normal   SuccessfulRescale    57s                 horizontal-pod-autoscaler  New size: 5; reason: pods metric evm_gas_used_per_second above target
  Warning  FailedGetPodsMetric  12s (x3 over 27s)   horizontal-pod-autoscaler  unable to get metric evm_gas_used_per_second: no metrics returned from custom metrics API
```

### Steps to reproduce

run `kubectl rollout restart deployment/mirror-web3` and observe `hpa/mirror-web3`'s events.

### Additional context

_No response_

### Hedera network

other

### Version

v0.135.0

### Operating system

None

jnels124:
**Description**:
* add recording rule to aggregate gas used for deployment
* scale hpa on gas used average for deployment

**Related issue(s)**:

Fixes #11692 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/charts/hedera-mirror-web3/templates/hpa.yaml
+++ b/charts/hedera-mirror-web3/templates/hpa.yaml
@@ -10,7 +10,7 @@ metadata:
 spec:
   behavior: {{ toYaml .Values.hpa.behavior | nindent 4 }}
   maxReplicas: {{ .Values.hpa.maxReplicas }}
-  metrics: {{ tpl (toYaml .Values.hpa.metrics) . | nindent 4 }}
+  metrics: {{ tpl .Values.hpa.metrics . | nindent 4 }}
   minReplicas: {{ .Values.hpa.minReplicas }}
   scaleTargetRef:
     apiVersion: apps/v1
--- a/charts/hedera-mirror-web3/templates/prometheusrule.yaml
+++ b/charts/hedera-mirror-web3/templates/prometheusrule.yaml
@@ -1,6 +1,6 @@
 # SPDX-License-Identifier: Apache-2.0
 
-{{ if .Values.prometheusRules.enabled -}}
+{{ if or .Values.prometheusRules.enabled .Values.prometheusRecordRules.enabled -}}
 apiVersion: monitoring.coreos.com/v1
 kind: PrometheusRule
 metadata:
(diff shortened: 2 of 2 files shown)


charts/hedera-mirror-web3/values.yaml
```diff
@@ -97,20 +97,26 @@ hpa:
   behavior: {}
   enabled: false
   maxReplicas: 12
-  metrics:
+  metrics: |
```
jnels124, line 100: the hpa already passes this through `tpl`

charts/hedera-mirror-web3/values.yaml
```diff
@@ -287,6 +293,24 @@ prometheusRules:
     labels:
       application: web3
       severity: warning
+prometheusRecordRules:
+  enabled: false
+  evm_gas_used_per_second_by_deployment:
+    enabled: true
+    expr: |
+        sum by (namespace, deployment) (
+          rate(hiero_mirror_web3_evm_gas_used_total{application="web3",namespace!="",pod!=""}[3m])
```
jnels124, line 302: joins are necessary here as the owner of a pod is the replica set which has a changing name and the owner of the replica set is the deployment.

charts/hedera-mirror-common/values.yaml
```diff
@@ -234,16 +234,17 @@ prometheus-adapter:
...
-            pod:
-              resource: "pod"
-        name:
-          as: "evm_gas_used_per_second"
-        metricsQuery: 'sum(rate(<<.Series>>{<<.LabelMatchers>>}[3m])) by (<<.GroupBy>>)'
+              resource: namespace
+            deployment:
+              group: apps
+              resource: deployment
+        metricsQuery: 'sum(<<.Series>>{<<.LabelMatchers>>}) by (<<.GroupBy>>)'
```
steven-sheehy, line 247: It all seems pretty complicated with the extra layer of prometheusRecordRules. Have you tried to revert to how it was but try either:

1) Just dropping pod resource override. Since `hiero_mirror_web3_evm_gas_used_total` is already specific to web3 deployment and we don't have multiple web3 deploys per namespace we just need namespace grouping.
2) If that doesn't work, can we keep it as before but manually specify group by clause to just include namespace?

jnels124: if we do it by pod, we can still have the behavior it had during rolling restarts. We can resolve it by setting `stabilizationWindowSeconds` but if we have it to high then we may miss scaling up as needed and if it is too low (lower than the time it takes to rollout all pods during a rolling restart) then we will see the behavior we saw before. Doing things as a deployment metric makes the behavior more deterministic.

jnels124: I tried doing this without the prometheus rule and just adding the same logic to the adapter config instead but it doesn't work correctly. There aren't errors but it doesn't expose the metric as a deployment object.

jnels124: It is easy for us to try the `stabilizationWindowSeconds ` approach and we can come back to this later if we find we are having issues not scaling up when needed or scaling up during rollout.

jnels124:
closed in favor of pr #11989

PR #11846, opened 2025-08-25
Add missing network api tests

### LINKED ISSUE #11845: Add network/fees and network/supply acceptance tests
### Problem

Based on parent issue currently network/fees and network/supply are not covered by acceptance test.

### Solution

Add tests to cover the endpoints.

### Alternatives

_No response_

kselveliev:
**Description**:
Add acceptance tests for network/fees and network/supply.

Fixes #11845

**Checklist**

### DIFF
--- a/test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/NetworkFeature.java
+++ b/test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/NetworkFeature.java
@@ -56,9 +56,9 @@ public class NetworkFeature {
     public void verifyNetworkFee() {
         final var networkFees = mirrorClient.getNetworkFees();
         assertThat(networkFees).isNotNull();
-        assertThat(networkFees.getFees()).isNotEmpty();
 
         final var fees = networkFees.getFees();
+        assertThat(fees).isNotEmpty();
         assertThat(fees.size()).isEqualTo(3);
 
         final var expectedTypes = Set.of("ContractCall", "ContractCreate", "EthereumTransaction");
--- a/test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/NetworkFeature.java
+++ b/test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/NetworkFeature.java
@@ -54,15 +54,16 @@ public class NetworkFeature {
 
     @When("I verify the network fees")
     public void verifyNetworkFee() {
-        var networkFees = mirrorClient.getNetworkFees();
+        final var networkFees = mirrorClient.getNetworkFees();
         assertThat(networkFees).isNotNull();
         assertThat(networkFees.getFees()).isNotEmpty();
 
         final var fees = networkFees.getFees();
         assertThat(fees.size()).isEqualTo(3);
 
-        var expectedTypes = Set.of("ContractCall", "ContractCreate", "EthereumTransaction");
-        var actualTypes = fees.stream().map(NetworkFee::getTransactionType).collect(Collectors.toSet());
+        final var expectedTypes = Set.of("ContractCall", "ContractCreate", "EthereumTransaction");
+        final var actualTypes =
+                fees.stream().map(NetworkFee::getTransactionType).collect(Collectors.toSet());
         assertThat(actualTypes).containsAll(expectedTypes);
 
         fees.forEach(fee -> {
@@ -74,11 +75,11 @@ public class NetworkFeature {
 
     @When("I verify the network supply")
     public void verifyNetworkSupply() {
-        var networkSupply = mirrorClient.getNetworkSupply();
+        final var networkSupply = mirrorClient.getNetworkSupply();
         assertThat(networkSupply).isNotNull();
 
-        var totalSupply = mustParseToBigDecimal(networkSupply.getTotalSupply());
-        var releasedSupply = mustParseToBigDecimal(networkSupply.getReleasedSupply());
+        final var totalSupply = mustParseToBigDecimal(networkSupply.getTotalSupply());
+        final var releasedSupply = mustParseToBigDecimal(networkSupply.getReleasedSupply());
 
         assertThat(totalSupply).isGreaterThan(BigDecimal.ZERO);
         assertThat(releasedSupply).isGreaterThanOrEqualTo(BigDecimal.ZERO);
--- a/test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/NetworkFeature.java
+++ b/test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/NetworkFeature.java
@@ -3,15 +3,20 @@
 package org.hiero.mirror.test.e2e.acceptance.steps;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.AssertionsForClassTypes.assertThatCode;
 import static org.hiero.mirror.test.e2e.acceptance.util.TestUtil.convertRange;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 
 import io.cucumber.java.en.When;
+import java.math.BigDecimal;
 import java.time.LocalDate;
 import java.time.ZoneOffset;
+import java.util.Set;
+import java.util.stream.Collectors;
 import lombok.CustomLog;
 import lombok.Data;
 import lombok.RequiredArgsConstructor;
+import org.hiero.mirror.rest.model.NetworkFee;
 import org.hiero.mirror.test.e2e.acceptance.client.MirrorNodeClient;
 import org.hiero.mirror.test.e2e.acceptance.props.Order;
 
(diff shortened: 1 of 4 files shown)


test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/NetworkFeature.java
```diff
@@ -47,6 +52,41 @@ public void verifyNetworkStake() {
         assertThat(networkStake.getUnreservedStakingRewardBalance()).isNotNegative();
     }
 
+    @When("I verify the network fees")
+    public void verifyNetworkFee() {
+        final var networkFees = mirrorClient.getNetworkFees();
```
kselveliev, line 61: Not sure if before all verifications we need to call similar method as in the network/stake test.

test/src/test/resources/features/network/networkstake.feature
```diff
@@ -4,3 +4,5 @@ Feature: Network Stake Coverage Feature
   @networkstake @acceptance
   Scenario Outline: Get network stake
```
kselveliev: We should probably rename from **networkstake** to **network**

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/NetworkFeature.java
```diff
@@ -63,4 +103,10 @@ private boolean shouldHaveStake() {
                 LocalDate.now(ZoneOffset.UTC).atStartOfDay(ZoneOffset.UTC).toInstant();
         return timestamp.lowerEndpoint().isBefore(midnight);
     }
+
+    private BigDecimal mustParseToBigDecimal(final String number) {
```
IvanKavaldzhiev: Since this is a utility method, can't it be just named `parseToBigDecimal`?

test/src/test/resources/features/network/network.feature
```diff
@@ -0,0 +1,14 @@
+@network @fullsuite
+Feature: Network Coverage Feature
+
+  @networkstake @acceptance
```
IvanKavaldzhiev: I believe @ acceptance tag can be removed from the individual Scenarios and added on top of the Feature

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/NetworkFeature.java
kselveliev: Addressed

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/NetworkFeature.java
kselveliev: Probably we should skip when response is not found

sdimitrov9: We should probably add an assert that the response is 200 and fail otherwise or assert that RestClientResponseException is not thrown since with the current implementation we can't get the status.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/NetworkFeature.java
```diff
@@ -47,6 +52,41 @@ public void verifyNetworkStake() {
...
+        fees.forEach(fee -> {
+            assertThat(fee.getGas()).isGreaterThan(0);
+        });
+
+        assertThat(networkFees.getTimestamp()).isNotNull();
+    }
+
+    @When("I verify the network supply")
+    public void verifyNetworkSupply() {
+        final var networkSupply = mirrorClient.getNetworkSupply();
```
sdimitrov9, line 79: I would assert RestClientResponseException is not thrown to verify that the response status is not 4xx or 5xx.

test/src/test/resources/features/network/network.feature
```diff
@@ -0,0 +1,14 @@
...
+  Scenario: Get network stake
+    When I verify the network stake
```
sdimitrov9, line 6: I would say scenarios consisting of just one step defeat the purpose of using cucumber and just adds overhead compared to just having the method behind the step as a separate test, but that is just a general observation.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/NetworkFeature.java
kselveliev: I did something similar, added a logic to see if fees or supplies are not found ->  skip verification. Otherwise do verification.

test/src/test/resources/features/network/network.feature
kselveliev: Addressed

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/NetworkFeature.java
kselveliev: I did something similar, added a logic to see if fees or supplies are not found ->  skip verification. Otherwise do verification. I guess we don't want to fail the whole suite when those are not found.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/NetworkFeature.java
```diff
@@ -47,6 +53,49 @@ public void verifyNetworkStake() {
...
+            assertThat(networkFees.getTimestamp()).isNotNull();
+        } catch (HttpClientErrorException.NotFound e) {
+            log.warn("Skipping network fees verification since they are missing");
+        }
+    }
+
+    @When("I verify the network supply")
+    public void verifyNetworkSupply() {
+        try {
+            final var networkSupply = mirrorClient.getNetworkSupply();
```
steven-sheehy: Both network fees and supply may not be populated for partial mirror nodes. This means this PR will cause tests to fail in those environments. You've mitigated it with the not found catch, but we can do better.

We should add a `MirrorNodeClient.hasPartialState()` that queries `/api/v1/blocks?order=asc&limit=1` and returns `block.number > 0`. This method should be cached so we don't need to keep querying the API. Then in both these tests to skip the test if hasPartialState() is true.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/NetworkFeature.java
```diff
@@ -47,6 +53,49 @@ public void verifyNetworkStake() {
...
+            final var actualTypes =
+                    fees.stream().map(NetworkFee::getTransactionType).collect(Collectors.toSet());
+            assertThat(actualTypes).containsAll(expectedTypes);
+
+            fees.forEach(fee -> {
+                assertThat(fee.getGas()).isGreaterThan(0);
+            });
+
+            assertThat(networkFees.getTimestamp()).isNotNull();
+        } catch (HttpClientErrorException.NotFound e) {
```
steven-sheehy: Remove the catch statements in favor of the other approach.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/NetworkFeature.java
```diff
@@ -47,6 +53,49 @@ public void verifyNetworkStake() {
...
+            final var expectedTypes = Set.of("ContractCall", "ContractCreate", "EthereumTransaction");
+            final var actualTypes =
+                    fees.stream().map(NetworkFee::getTransactionType).collect(Collectors.toSet());
+            assertThat(actualTypes).containsAll(expectedTypes);
+
+            fees.forEach(fee -> {
+                assertThat(fee.getGas()).isGreaterThan(0);
+            });
+
+            assertThat(networkFees.getTimestamp()).isNotNull();
```
steven-sheehy:
```suggestion
            assertThat(networkFees)
                .isNotNull()
                .satisfies(f -> assertThat(f.getTimestamp()).isNotNull())
                .extracting(NetworkFeesResponse::getFees, InstanceOfAssertFactories.list(NetworkFee.class))
                .hasSize(3)
                .allSatisfy(fee -> {
                    assertThat(fee.getTransactionType()).isIn(expectedTypes);
                    assertThat(fee.getGas()).isGreaterThan(0);
                });
```

steven-sheehy (changes requested):

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/NetworkFeature.java
kselveliev: I agree with the suggestion. One note, staging env has many blocks but when we query /fees or /supply they both return not found. So hasPartialState() will return true but then still fail.

kselveliev: I think I misunderstood initially. 
We should only run the test if it has blocks and is starting from the block 0. This will work.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/NetworkFeature.java
kselveliev: Addressed

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/NetworkFeature.java
kselveliev: Handled with Steven's suggestion

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/NetworkFeature.java
kselveliev: Addressed

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/NetworkFeature.java
kselveliev: Addressed

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/NetworkFeature.java
kselveliev: Addressed

xin-hedera (approved):
LGTM

PR #11852, opened 2025-08-26
Contract result not returned for eth transaction hash

### LINKED ISSUE #11770: Contract/results endpoint doesn't return a result
### Description

When querying the Mirror Node for contract call results by transaction **hash**, the API returns "Not found", even though the same transaction queried by **ID** returns the expected result.

### Steps to reproduce

1. Verify the transaction on Hashscan:  
   https://hashscan.io/mainnet/transaction/1754443096.485525000  
2. Query by **transaction ID** (works):  
  https://mainnet-public.mirrornode.hedera.com/api/v1/contracts/results/0.0.7321277-1754443067-964256552  
3. Query by **transaction hash** (fails):  
  https://mainnet-public.mirrornode.hedera.com/api/v1/contracts/results/0x8b51976d96c88c23edf8530c6ca7bfb08503d5efb969cb008bf9c071d4609e79  

### Additional context

_No response_

### Hedera network

mainnet

### Version

v0.135.0

### Operating system

macOS

bilyana-gospodinova:
**Description**:
Improve the contract results endpoint to return the successful results with priority when the eth transaction hash is passed and there are multiple results with the same eth transaction hash. If there is no successful result, the latest failed result is returned.

Fixes #11770

### DIFF
--- a/rest/__tests__/specs/contracts/results/{id}/transaction-hash-match-multiple-tx.json
+++ b/rest/__tests__/specs/contracts/results/{id}/transaction-hash-match-multiple-tx.json
@@ -174,7 +174,7 @@
     "s": "0x24e9c602ac800b983b035700a14b23f78a253ab762deab5dc27e3555a750b354",
     "state_changes": [],
     "status": "0x1",
-    "timestamp": "167654.000123456",
+    "timestamp": "167654.000123457",
     "to": "0x0000000000000000000000000000000000001389",
     "transaction_index": 1,
     "type": 2,
--- a/rest/controllers/contractController.js
+++ b/rest/controllers/contractController.js
@@ -39,6 +39,7 @@ import {
   ContractStateViewModel,
   ContractViewModel,
 } from '../viewmodel';
+import ContractTransactionHash from '../model/contractTransactionHash.js';
 
 const contractSelectFields = [
   Entity.AUTO_RENEW_ACCOUNT_ID,
@@ -70,6 +71,7 @@ const contractCreateType = Number(TransactionType.getProtoId('CONTRACTCREATEINST
 const ethereumTransactionType = Number(TransactionType.getProtoId('ETHEREUMTRANSACTION'));
 const duplicateTransactionResult = TransactionResult.getProtoId('DUPLICATE_TRANSACTION');
 const wrongNonceTransactionResult = TransactionResult.getProtoId('WRONG_NONCE');
+const successTransactionResult = TransactionResult.getProtoId('SUCCESS');
 
 /**
  * Extracts the sql where clause, params, order and limit values to be used from the provided contract query
@@ -1114,12 +1116,7 @@ class ContractController extends BaseController {
     const excludeTransactionResults = [duplicateTransactionResult, wrongNonceTransactionResult];
     const {transactionIdOrHash} = req.params;
     if (utils.isValidEthHash(transactionIdOrHash)) {
-      const detailsByHash = await ContractService.getContractTransactionDetailsByHash(
-        utils.parseHexStr(transactionIdOrHash),
-        excludeTransactionResults,
-        1
-      );
-      transactionDetails = detailsByHash[0];
+      transactionDetails = await this.getEthTransactionDetails(transactionIdOrHash, excludeTransactionResults);
     } else {
       const transactionId = TransactionId.fromString(transactionIdOrHash);
       const nonce = getLastNonceParamValue(req.query);
@@ -1217,6 +1214,41 @@ class ContractController extends BaseController {
         contractDetails.contractIds
       ),
     ]);
+  };
+
+  getEthTransactionDetails = async (transactionIdOrHash, excludeTransactionResults) => {
+    const detailsByHash = await ContractService.getContractTransactionDetailsByHash(
+      utils.parseHexStr(transactionIdOrHash),
+      excludeTransactionResults
+    );
+
+    const detailsByHashSuccessful = detailsByHash.filter((t) => t.transactionResult === successTransactionResult);
+    // If there is more than 1 successful transaction -> return the first successful.
+    if (detailsByHashSuccessful.length > 1) {
+      this.sortTransactionsByConsensusTimestampAsc(detailsByHashSuccessful);
+    }
+    let transactionDetails = detailsByHashSuccessful[0];
+
+    if (!transactionDetails) {
+      // If there are no successful transactions -> return the earliest unsuccessful.
+      const detailsByHashUnsuccessful = detailsByHash.filter((t) => t.transactionResult !== successTransactionResult);
+      if (detailsByHashUnsuccessful.length > 1) {
+        this.sortTransactionsByConsensusTimestampAsc(detailsByHashUnsuccessful);
+      }
+      transactionDetails = detailsByHashUnsuccessful[0];
+    }
+    return transactionDetails;
+  };
+
+  sortTransactionsByConsensusTimestampAsc = (transactions) => {
+    if (transactions.length > 1) {
+      transactions.sort((a, b) => {
+        if (a.consensusTimestamp < b.consensusTimestamp) return -1;
+        if (a.consensusTimestamp > b.consensusTimestamp) return 1;
+        return 0;
+      });
+    }
+    return transactions;
   };
 
   getContractActions = async (req, res) => {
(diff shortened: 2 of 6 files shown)


rest/controllers/contractController.js
```diff
@@ -1219,6 +1216,41 @@ class ContractController extends BaseController {
...
+
+    const detailsByHashSuccessful = detailsByHash.filter((t) => t.transactionResult === successTransactionResult);
+    // If there is more than 1 successful transaction -> return the first successful.
+    if (detailsByHashSuccessful.length > 1) {
+      this.sortTransactionsByConsensusTimestampAsc(detailsByHashSuccessful);
+    }
+    let transactionDetails = detailsByHashSuccessful[0];
+
+    if (!transactionDetails) {
+      // If there are no successful transactions -> return the earliest unsuccessful.
```
IvanKavaldzhiev: Probably it would be more clear like this:

// If there is more than 1 unsuccessful transaction -> return the first unsuccessful.

rest/controllers/contractController.js
```diff
@@ -1219,6 +1216,41 @@ class ContractController extends BaseController {
...
+      const detailsByHashUnsuccessful = detailsByHash.filter((t) => t.transactionResult !== successTransactionResult);
+      if (detailsByHashUnsuccessful.length > 1) {
+        this.sortTransactionsByConsensusTimestampAsc(detailsByHashUnsuccessful);
+      }
+      transactionDetails = detailsByHashUnsuccessful[0];
+    }
+    return transactionDetails;
+  };
+
+  sortTransactionsByConsensusTimestampAsc = (transactions) => {
```
bilyana-gospodinova: Not sure if we should return the earliest or the latest entry. I return the earliest in order to extend the current implementation and to be in accordance with the existing tests. I think if we return the latest successful/unsuccessful entry makes more sense. Any thoughts?

kselveliev (approved):

rest/controllers/contractController.js
```diff
@@ -1219,6 +1216,41 @@ class ContractController extends BaseController {
     ]);
   };
 
+  getEthTransactionDetails = async (transactionIdOrHash, excludeTransactionResults) => {
```
xin-hedera: we should change the query in `ContractService.getContractTransactionDetailsByHash` instead,  e.g.,

```sql
select *
from contract_transaction_hash
where hash = $1 and consensus_timestamp = (
  select coalesce((
      select consensus_timestamp
      from contract_transaction_hash
      where hash = $1 and transaction_result = 22
      order by consensus_timestamp
      limit 1
    ), (
      select consensus_timestamp
      from contract_transaction_hash
      where hash = $1 and transaction_result not in (11, 312)
      order by consensus_timestamp
      limit 1
    )
);
```

Further more, `ContractService.getContractTransactionDetailsByHash` is called in two places. The other is by `getContractActions` and it does not pass a `limit` although it only uses the first returned row.  I think we should apply the same logic to `getContractActions` since it doesn't make sense to not exclude WRONG_NONCE / DUPLICATION_TRANSACTION given a transaction hash, those transactions do not have contract actions at all.

So we'd also want to change the function signature to just

```javascript
getContractTransactionDetailsByHash(hash)
```

rest/controllers/contractController.js
xin-hedera: there can be at most one successful transaction given a hash, so the order only applies to unsuccessful transactions. I think returning the latest is better

xin-hedera (changes requested):

rest/controllers/contractController.js
bilyana-gospodinova: Thanks, addressed.

rest/controllers/contractController.js
bilyana-gospodinova: Addressed

rest/controllers/contractController.js
bilyana-gospodinova: Comment addressed.

rest/controllers/contractController.js
```diff
@@ -39,6 +39,7 @@ import {
   ContractStateViewModel,
   ContractViewModel,
 } from '../viewmodel';
+import ContractTransactionHash from '../model/contractTransactionHash.js';
```
xin-hedera: unused?
```suggestion
```

rest/service/contractService.js
```diff
@@ -297,33 +303,29 @@ class ContractService extends BaseService {
...
-      ContractService.transactionHashDetailsQuery,
-      transactionsFilter,
-      `order by ${ContractTransactionHash.CONSENSUS_TIMESTAMP} asc`,
-      limit ? `limit ${limit}` : '',
-    ];
-    const transactionHashRows = await super.getRows(query.join('\n'), [hash]);
-    return transactionHashRows.map((row) => {
-      return new ContractTransactionHash(row);
-    });
+      `select * from ${ContractTransactionHash.tableName}`,
```
xin-hedera: nit, prefer multi-line string with backticks instead of joining an array of strings, e.g.,

```javascript
const sql = `
  select *
  from ${ContractTransactionHash.tableName}`;

```

rest/__tests__/specs/contracts/results/{id}/transaction-hash-match-multiple-tx.json
```diff
@@ -174,7 +174,7 @@
     "s": "0x24e9c602ac800b983b035700a14b23f78a253ab762deab5dc27e3555a750b354",
     "state_changes": [],
     "status": "0x1",
-    "timestamp": "167654.000123456",
+    "timestamp": "167654.000123457",
```
steven-sheehy, line 177: This test case seems setup wrong since it inserts 3 successful ethereum transactions with the same hash. Instead, we should have 1 successful and 2 unsuccessful. We should also probably add one more spec with only unsuccessful.

rest/service/contractService.js
steven-sheehy: Move this field to be a class constant so it's not re-created on each call.

rest/service/contractService.js
```diff
@@ -297,33 +303,29 @@ class ContractService extends BaseService {
    * Retrieves contract transaction details based on the eth hash
    *
    * @param {Buffer} hash eth transaction hash or 32-byte hedera transaction hash prefix
-   * @param {[]}excludeTransactionResults transaction result codes to exclude in result
-   * @param {number} limit number of results to return
    * @return {Promise<{ContractResult}[]>}
```
steven-sheehy: This return type is inaccurate

rest/service/contractService.js
```diff
@@ -297,33 +303,29 @@ class ContractService extends BaseService {
...
-      transactionsFilter,
-      `order by ${ContractTransactionHash.CONSENSUS_TIMESTAMP} asc`,
-      limit ? `limit ${limit}` : '',
-    ];
-    const transactionHashRows = await super.getRows(query.join('\n'), [hash]);
-    return transactionHashRows.map((row) => {
-      return new ContractTransactionHash(row);
-    });
+      `select * from ${ContractTransactionHash.tableName}`,
+      `where ${ContractTransactionHash.HASH} = $1 and ${ContractTransactionHash.CONSENSUS_TIMESTAMP} = (`,
```
steven-sheehy: We don't need to filter again by hash since it's already filtered in the sub-query to get the unique timestamp.

steven-sheehy (changes requested):

rest/service/contractService.js
xin-hedera: Need the hash since it’s the only index on that table.

rest/controllers/contractController.js
bilyana-gospodinova: Correct. Removed.

rest/service/contractService.js
bilyana-gospodinova: Comment addressed

rest/__tests__/specs/contracts/results/{id}/transaction-hash-match-multiple-tx.json
bilyana-gospodinova: Comment addressed

rest/service/contractService.js
bilyana-gospodinova: Moved

rest/service/contractService.js
bilyana-gospodinova: Fixed

steven-sheehy (approved):
LGTM

xin-hedera (approved):
LGTM

PR #11864, opened 2025-08-27
Improve release checklist

xin-hedera:
**Description**:

- Add a step to recreate clean deployments in `mainnet-citus` namespace before deploying to staging

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/docs/checklist/release.md
+++ b/docs/checklist/release.md
@@ -37,6 +37,8 @@ Deployed automatically on every tag.
 
 ### Mainnet Staging
 
+- [ ] Delete all API deployments and let flux reconcile and recreate them to drop manual changes
+  - `kubectl delete deployments mirror-grpc mirror-rest mirror-restjava mirror-web3`
 - [ ] Deployed
 - [ ] Helm Controller logs show successful reconciliation check
 - [ ] Helm release status is healthy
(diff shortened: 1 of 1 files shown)


docs/checklist/release.md
```diff
@@ -37,6 +37,8 @@ Deployed automatically on every tag.
 
 ### Mainnet Staging
 
+- [ ] Delete all API deployments and let flux reconcile and recreate them to drop manual changes
+  - `kubectl delete deployments mirror-grpc mirror-rest mirror-restjava mirror-web3`
```
steven-sheehy: Why not delete all?
```suggestion
  - `kubectl delete deployments --all`
```

xin-hedera: can change to that.

docs/checklist/release.md
```diff
@@ -37,6 +37,8 @@ Deployed automatically on every tag.
 
 ### Mainnet Staging
 
+- [ ] Delete all deployments in `mainnet-citus` and let flux reconcile and recreate them to drop manual changes
+  - `kubectl -n mainnet-citus delete deployments --all`
```
steven-sheehy, line 41: Hopefully they have the right context set 😬

xin-hedera: so we may want to change the namespace to `mainnet-staging-citus` in case it doesn't break too many other things.

steven-sheehy: If we're going to change the namespace then we should do `mainnet-staging` and drop the -citus

jnels124 (approved):

PR #11877, opened 2025-08-29
Add network stake k6 test case to rest-java

xin-hedera:
**Description**:

This PR duplicates the network stake k6 test case for rest-java

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

Wanted to a take a cleaner approach to move it to a common test case folder. It's easy for network stake mainly due to no test case parameter discovery involved. However it'll be more work when later we move other test cases over and it's not worth the effort for non-production code.

**Checklist**

### DIFF
--- a/tools/k6/src/rest-java/libex/common.js
+++ b/tools/k6/src/rest-java/libex/common.js
@@ -26,7 +26,7 @@ class RestJavaTestScenarioBuilder extends TestScenarioBuilder {
     super();
     this.fallbackRequest((testParameters) => {
       // Fallback to a known account ID. Replace this with another API endpoint when one exists not requiring an entity ID.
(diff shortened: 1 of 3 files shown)


tools/k6/src/rest-java/libex/common.js
```diff
@@ -26,7 +26,7 @@ class RestJavaTestScenarioBuilder extends TestScenarioBuilder {
     super();
     this.fallbackRequest((testParameters) => {
       // Fallback to a known account ID. Replace this with another API endpoint when one exists not requiring an entity ID.
-      const url = `${testParameters['BASE_URL_PREFIX']}/accounts/2/allowances/nfts`;
+      const url = `${testParameters['BASE_URL_PREFIX']}/accounts/0.0.2/allowances/nfts`;
```
steven-sheehy: Think we should keep it as before so it works with non-zero realm.

PR #11902, opened 2025-09-01
Enhance acceptance workflow to also support running tests using block node in solo

### LINKED ISSUE #11806: Block node E2E test
### Problem

After #8513 is complete, we'd like to have a similar test that enables block streams and block nodes

### Solution

* Enhance workflow from #8513 to use matrix testing to test record stream and block stream
* Configure importer to use block node and block stream. Consider creating a spring profile if multiple properties need to be set.
* Work with solo team to configure it to disable minio and enable block nodes

### Alternatives

_No response_

IvanKavaldzhiev:
**Description**:
Enhance the acceptance workflow with matrix configuration that runs acceptance tests also using block stream

**Related issue(s)**:

Fixes #11806 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/.github/workflows/acceptance.yaml
+++ b/.github/workflows/acceptance.yaml
@@ -32,6 +32,12 @@ defaults:
 jobs:
   acceptance:
     runs-on: hiero-mirror-node-linux-large
+    strategy:
+      fail-fast: false
+      matrix:
+        stream-type:
+          - RECORD
+          - BLOCK
     timeout-minutes: 40
     env:
       SOLO_CLUSTER_NAME: solo-mirror-acceptance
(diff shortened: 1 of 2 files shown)


.github/workflows/acceptance.yaml
```diff
@@ -88,25 +94,65 @@ jobs:
...
+              tag: "${{ github.base_ref }}"
+          importer:
+            env:
+              HIERO_MIRROR_IMPORTER_BLOCK_NODES_0_HOST: 'block-node-0.${SOLO_NAMESPACE}.svc.cluster.local'
+              HIERO_MIRROR_IMPORTER_BLOCK_NODES_0_PORT: '40840'
+              SPRING_PROFILES_ACTIVE: 'blocknode'
+            image:
+              registry: docker.io
+              repository: xinatswirlds/importer
+              tag: e741ea2b2
```
steven-sheehy, line 133: I assume this image just contains the new `blocknode` spring profile? Can't we create a separate PR to just get that change in first and avoid the hacks to override images in this PR?

.github/workflows/acceptance.yaml
```diff
@@ -88,25 +94,65 @@ jobs:
           solo deployment create --deployment "${SOLO_DEPLOYMENT}" --namespace "${SOLO_NAMESPACE}"
           solo deployment add-cluster --deployment "${SOLO_DEPLOYMENT}" --cluster-ref kind-"${SOLO_CLUSTER_NAME}" --num-consensus-nodes 1
           solo cluster-ref setup --cluster-ref kind-"${SOLO_CLUSTER_NAME}"
+
+          if [ "${{ matrix.stream-type }}" = "BLOCK" ]; then
+            solo block node add --deployment "${SOLO_DEPLOYMENT}" --cluster-ref kind-"${SOLO_CLUSTER_NAME}" --chart-version v0.17.0-rc2 --release-tag v0.65.0
```
steven-sheehy: What about disabling minio and uploaders and setting consensus node's stream type to blocks? Does that all happen automatically with this command?

.github/workflows/acceptance.yaml
```diff
@@ -88,25 +94,65 @@ jobs:
...
+          restjava:
             image:
               tag: "${{ github.base_ref }}"
+          web3:
+            image:
+              tag: "${{ github.base_ref }}"
+          importer:
+            env:
+              HIERO_MIRROR_IMPORTER_BLOCK_NODES_0_HOST: 'block-node-0.${SOLO_NAMESPACE}.svc.cluster.local'
+              HIERO_MIRROR_IMPORTER_BLOCK_NODES_0_PORT: '40840'
```
steven-sheehy, line 128: Why this port when the default port in code is 8080? Is this a static port configured by solo or can it be dynamic?

.github/workflows/acceptance.yaml
```diff
@@ -88,25 +94,65 @@ jobs:
...
+              SPRING_PROFILES_ACTIVE: 'blocknode'
+            image:
+              registry: docker.io
+              repository: xinatswirlds/importer
+              tag: e741ea2b2
           test:
+            cucumberTags: "@acceptance and not @accounts and not @contract and not @equivalence and not @estimateprecompile and not @ethereum and not @historical and not @precompile and not @filebase and not @schedulebase and not @token and not @topic and not @nftallowance and not @erc and not @cryptoallowance and not @tokenallowance and not @ethcall"
```
steven-sheehy: What tests still pass? Maybe it's easier to only list the few that pass?

test/src/test/resources/features/contract/contract.feature
```diff
@@ -1,4 +1,4 @@
-@contractbase @fullsuite
+@contractbase @fullsuite @contract
```
steven-sheehy: Eventually we want to rename `@contractbase` to `@contract` so that will conflict with this. Maybe we can just add `@crud`?

steven-sheehy (changes requested):

.github/workflows/acceptance.yaml
xin-hedera: this is the new default block server port, I'll make it the default in the properties class with my upcoming PR.

.github/workflows/acceptance.yaml
xin-hedera: it's built from the hip-1056 feature branch plus this [PR](https://github.com/hiero-ledger/hiero-mirror-node/pull/11863)

until we merge the feature branch into main, we'll have to run a custom image for consensus node 0.65+

.github/workflows/acceptance.yaml
IvanKavaldzhiev: I think it does not happen automatically. There is a `--minio false` command which can disable minio, however when using it locally, the node fails to start. I believe this can be polished in a future PR, once locally it is successfully tested.

.github/workflows/acceptance.yaml
IvanKavaldzhiev: I tried it but it seems having more than one `and` in the `cucumber.filters` like `@acceptance and @accounts and @contract` is not working.

test/src/test/resources/features/contract/contract.feature
IvanKavaldzhiev: Renamed

.github/workflows/acceptance.yaml
```diff
@@ -76,36 +83,78 @@ jobs:
         run: ./gradlew test:build -x test
 
       - name: Build Mirror Node Test Docker image
-        run: docker build -t gcr.io/mirrornode/hedera-mirror-test:${{ github.base_ref }} ./test
+        run: |
+          TAG=${GITHUB_BASE_REF:-${GITHUB_REF_NAME}}
```
steven-sheehy, line 87: If you do this then you can remove the other definition of TAG. Can also put in a separate step so it's more clear.

```suggestion
          echo "TAG=${GITHUB_BASE_REF:-${GITHUB_REF_NAME}}" >> $GITHUB_ENV
```

steven-sheehy (approved):
LGTM

xin-hedera (approved):
LGTM

Issue #11866, opened 2025-08-27
Add missing contracts api to acceptance tests - part 1


ashumahajan:
### Problem

We're missing acceptance test coverage for some contracts mirror node REST APIs.

### Solution

Based on https://github.com/hiero-ledger/hiero-mirror-node/issues/4051 , add below apis coverage to acceptance tests

1. /api/v1/contracts - List all contracts
2. /api/v1/contracts/{id}/state - Contract state
3. /api/v1/contracts/{id}/results/{timestamp} - Contract results by timestamp

ashumahajan:
Parent: #11807

PR #11903, opened 2025-09-01
Add some contract REST endpoints

bilyana-gospodinova:
**Description**:
This PR adds coverage for some of the missing REST endpoints:
- /api/v1/contracts - List all contracts
- /api/v1/contracts/{id}/state - Contract state
- /api/v1/contracts/{id}/results/{timestamp} - Contract results by timestamp

Related  issues: #11866


kselveliev (approved):

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/ContractFeature.java
```diff
@@ -290,6 +291,18 @@ public void verifyContractNonce() {
...
 
+    @Then("I verify parent and child contracts are deleted")
+    public void verifyParentAndChildContractsAreDeleted() {
+        var contracts = mirrorClient.getContracts(5);
+        if (contracts.getContracts().isEmpty()) {
+            return;
+        }
+        for (var contract : contracts.getContracts()) {
+            assertThat(contract.getContractId()).isNotEqualTo(deployedParentContract.contractId());
+            assertThat(contract.getContractId()).isNotEqualTo(create2ChildContractContractId);
```
steven-sheehy: I'm not sure what this is verifying since the contract list endpoint will return deleted contracts as well. So this could potentially fail unless there are 5 more recent contracts. Probably instead we could use multiple `contract.id` query parameters to search for them.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/ContractFeature.java
```diff
@@ -290,6 +291,18 @@ public void verifyContractNonce() {
         verifyNonceForChildContracts();
     }
 
+    @Then("I verify parent and child contracts are deleted")
```
steven-sheehy: Please merge from main to ensure the acceptance workflow works properly and tests your changes. I don't see `I verify parent and child contracts are deleted` in the logs of the most recent [run](https://github.com/hiero-ledger/hiero-mirror-node/actions/runs/17379640215/job/49333844576).

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/ContractFeature.java
```diff
@@ -122,9 +122,10 @@ public void verifyMirrorAPIAccountResponses(int status) {
 
     @Then("the mirror node REST API should verify the contract")
     public void verifyContract() {
-        verifyContractFromMirror(false);
+        var contractResponse = verifyContractFromMirror(false);
         verifyContractExecutionResultsById();
         verifyContractExecutionResultsByTransactionId();
+        verifyContractExecutionResultByIdAndTimestamp(contractResponse.getCreatedTimestamp());
```
steven-sheehy: It seems we call `the mirror node REST API should verify the contract` for create and call. But here we're only comparing the create timestamp so for the call scenario it will be verifying the older result. Maybe we can return a list of results from verifyContractExecutionResultsByTransactionId() and feed those timestamps to this?

steven-sheehy (changes requested):

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/ContractFeature.java
bilyana-gospodinova: I didn't notice the `contract.id` query parameter. Thanks, this makes more sense now. Comment addressed.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/ContractFeature.java
bilyana-gospodinova: Pulled latest main. I removed this method as it was no longer needed after addressing the other comments.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/ContractFeature.java
bilyana-gospodinova: Addressed. The passed timestamp is now the one returned from the previous endpoint. It is a single result, so we know which contract result we want to fetch.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/client/MirrorNodeClient.java
```diff
@@ -246,6 +248,11 @@ public TokenAllowancesResponse getAccountTokenAllowanceBySpender(
                 spenderId);
     }
 
+    public ContractsResponse getContracts(String contractId) {
+        log.debug("Verify contract '{}' is returned by Mirror Node", contractId);
+        return callRestEndpoint("/contracts?contract.id={contractId}", ContractsResponse.class, contractId);
+    }
+
```
ashumahajan, line 255: I think what was missing was testing api/v1/contracts without contract Id

steven-sheehy: I don't think we have to test every query parameter combination. Main thing to check is the API as a whole is working.

Issue #11868, opened 2025-08-27
Add missing topics and tokens REST APIs to acceptance tests


ashumahajan:
### Problem

We're missing acceptance test coverage for some tokens and topics mirror node REST APIs.

### Solution

Based on #4051 , add below apis coverage to acceptance tests

1. /api/v1/topics/messages/{timestamp} - Topic messages by timestamp
2. /api/v1/tokens - List all tokens
3. /api/v1/tokens/{id}/nfts - NFTs for a token

ashumahajan:
Parent: #11807

PR #11905, opened 2025-09-02
Add acceptance tests for tokens and topics endpoints

sdimitrov9:
**Description**:
Add acceptance tests for tokens and topics endpoints

**Related issue(s)**: #11868 

Fixes #

**Checklist**

### DIFF
--- a/test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/TokenFeature.java
+++ b/test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/TokenFeature.java
@@ -56,6 +56,7 @@ import org.hiero.mirror.rest.model.TokenRelationship;
 import org.hiero.mirror.rest.model.TokenRelationship.FreezeStatusEnum;
 import org.hiero.mirror.rest.model.TokenRelationship.KycStatusEnum;
 import org.hiero.mirror.rest.model.TokenRelationshipResponse;
+import org.hiero.mirror.rest.model.TokensResponse;
 import org.hiero.mirror.rest.model.TransactionByIdResponse;
 import org.hiero.mirror.rest.model.TransactionDetail;
 import org.hiero.mirror.rest.model.TransactionNftTransfersInner;
@@ -89,6 +90,10 @@ public class TokenFeature extends AbstractFeature {
     private TokenId tokenId;
 
     private TokenResponse tokenResponse;
+
+    private TokensResponse tokensWithTreasuryAccount;
+
+    private TokensResponse tokensAssociatedWithPublicKey;
 
     private TransactionDetail transactionDetail;
 
@@ -322,6 +327,11 @@ public class TokenFeature extends AbstractFeature {
 
     @Given("I successfully create a new unfrozen and granted kyc token")
     public void createNewToken() {
+        this.tokensWithTreasuryAccount = mirrorClient.getTokensWithTreasuryAccount(
+                tokenClient.getAccountId().toString());
+        this.tokensAssociatedWithPublicKey = mirrorClient.getTokensAssociatedWithPublicKey(
+                tokenClient.getAccountId().getPublicKey().toString());
+
         this.tokenResponse = tokenClient.getToken(TokenNameEnum.FUNGIBLE_KYC_UNFROZEN_2);
         this.tokenId = tokenResponse.tokenId();
         this.networkTransactionResponse = tokenResponse.response();
@@ -333,6 +343,37 @@ public class TokenFeature extends AbstractFeature {
         this.tokenResponse = tokenClient.getToken(TokenNameEnum.FUNGIBLE_AIRDROP);
         this.tokenId = tokenResponse.tokenId();
         this.networkTransactionResponse = tokenResponse.response();
+    }
+
+    @When("I verify the number of tokens with given treasury account")
+    public void getAllTokensForAccount() {
+        TokensResponse tokensByAccountId = mirrorClient.getTokensWithTreasuryAccount(
+                tokenClient.getAccountId().toString());
+        assertThat(tokensByAccountId.getTokens()).isNotNull();
+        assertThat(this.tokensWithTreasuryAccount.getTokens().size() + 1)
+                .isEqualTo(tokensByAccountId.getTokens().size());
+        assertThat(this.tokenResponse.tokenId().toString())
+                .isEqualTo(tokensByAccountId.getTokens().getLast().getTokenId());
+
+        TokensResponse tokensByPublicKey = mirrorClient.getTokensAssociatedWithPublicKey(
+                tokenClient.getAccountId().getPublicKey().toString());
+        assertThat(tokensByPublicKey.getTokens()).isNotNull();
+        assertThat(tokensByPublicKey.getTokens().stream()
+                        .filter(t -> t.getTokenId()
+                                .equals(this.tokensAssociatedWithPublicKey
+                                        .getTokens()
+                                        .getLast()
+                                        .getTokenId()))
+                        .findFirst())
+                .isNotNull();
+    }
+
+    @When("I verify the number of tokens associated with account")
+    public void getAllTokensByPublicKey() {
+        TokensResponse tokensAssociatedWithPubKey = mirrorClient.getTokensAssociatedWithPublicKey(
+                tokenClient.getAccountId().getPublicKey().toString());
+        assertThat(this.tokensAssociatedWithPublicKey.getTokens().size() + 1)
+                .isEqualTo(tokensAssociatedWithPubKey.getTokens().size());
     }
 
     @Given("I successfully create a new nft {token} with infinite supplyType")
--- a/test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/TopicFeature.java
+++ b/test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/TopicFeature.java
@@ -404,6 +404,19 @@ public class TopicFeature extends AbstractFeature {
         assertThat(getTopicMessageResponse.getMessage()).isEqualTo(base64EncodedMessage);
     }
 
+    @Then("I verify the published message can be retrieved by consensus timestamp")
+    public void getTopicMessageByTimestamp() {
+        var topicMsgBySeqNumber = mirrorClient.getTopicMessageBySequenceNumber(
+                consensusTopicId.toString(), String.valueOf(topicSequenceNumber));
+        assertThat(topicMsgBySeqNumber).isNotNull();
+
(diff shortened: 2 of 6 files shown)


test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/TopicFeature.java
```diff
@@ -404,6 +404,19 @@ public void verifyTopicMessage() {
         assertThat(getTopicMessageResponse.getMessage()).isEqualTo(base64EncodedMessage);
     }
 
+    @Then("I verify the published message can be retrieved by consensus timestamp")
+    public void getTopicMessageByTimestamp() {
+        var topicMsgBySeqNumber = mirrorClient.getTopicMessageBySequenceNumber(
```
steven-sheehy: We can save this response in the previous step `I verify the published message in mirror node REST API` and reuse it here, avoiding an extra call.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/TopicFeature.java
```diff
@@ -404,6 +404,19 @@ public void verifyTopicMessage() {
...
+                consensusTopicId.toString(), String.valueOf(topicSequenceNumber));
+        assertThat(topicMsgBySeqNumber).isNotNull();
+
+        String consensusTimestamp = topicMsgBySeqNumber.getConsensusTimestamp();
```
steven-sheehy:
```suggestion
        final var consensusTimestamp = topicMsgBySeqNumber.getConsensusTimestamp();
```

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/TokenFeature.java
```diff
@@ -133,6 +138,18 @@ public void ensureTokenProperties() {
...
 
+    @RetryAsserts
+    @Given("I verify the number of nfts for token is expected")
+    public void verifyNumOfNFTsForToken() {
+        var apiResponse = mirrorClient.getTokenNFTs(tokenId.toString());
+        assertThat(apiResponse.getNfts()).isNotNull();
+
+        var nftsInitialSize = apiResponse.getNfts().size();
+
+        tokenClient.mint(tokenId, nextBytes(4));
```
steven-sheehy: We shouldn't need to do an extra mint. We already do `I mint a serial number from the token` later in the test. We should just be able to just verify after that test the expected count. We should also verify some of the content of the NFT.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/TokenFeature.java
```diff
@@ -335,6 +357,37 @@ public void createNewTokenKycNotApplicable() {
...
+    }
+
+    @When("I verify the number of tokens associated with account")
+    public void getAllTokensByPublicKey() {
+        TokensResponse tokensAssociatedWithPubKey = mirrorClient.getTokensAssociatedWithPublicKey(
+                tokenClient.getAccountId().getPublicKey().toString());
+        assertThat(this.tokensAssociatedWithPublicKey.getTokens().size() + 1)
+                .isEqualTo(tokensAssociatedWithPubKey.getTokens().size());
+    }
+
```
steven-sheehy: This assert can be combined with `I verify the number of tokens with given treasury account` since you already call getTokensAssociatedWithPublicKey there.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/TokenFeature.java
```diff
@@ -322,6 +339,11 @@ public void createNewToken(List<CustomFee> customFees) {
 
     @Given("I successfully create a new unfrozen and granted kyc token")
     public void createNewToken() {
+        this.tokensWithTreasuryAccount = mirrorClient.getTokensWithTreasuryAccount(
+                tokenClient.getAccountId().toString());
+        this.tokensAssociatedWithPublicKey = mirrorClient.getTokensAssociatedWithPublicKey(
+                tokenClient.getAccountId().getPublicKey().toString());
```
steven-sheehy: These calls should result in identical responses, right? So can just assert equal then store one.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/TokenFeature.java
```diff
@@ -335,6 +357,37 @@ public void createNewTokenKycNotApplicable() {
...
+    public void getAllTokensForAccount() {
+        TokensResponse tokensByAccountId = mirrorClient.getTokensWithTreasuryAccount(
+                tokenClient.getAccountId().toString());
+        assertThat(tokensByAccountId.getTokens()).isNotNull();
+        assertThat(this.tokensWithTreasuryAccount.getTokens().size() + 1)
+                .isEqualTo(tokensByAccountId.getTokens().size());
+        assertThat(this.tokenResponse.tokenId().toString())
+                .isEqualTo(tokensByAccountId.getTokens().getLast().getTokenId());
+
+        TokensResponse tokensByPublicKey = mirrorClient.getTokensAssociatedWithPublicKey(
```
steven-sheehy: This response and getTokensWithTreasuryAccount should be identical. So we can move all asserts to first then just verify first and second are equal for second.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/TokenFeature.java
```diff
@@ -335,6 +357,37 @@ public void createNewTokenKycNotApplicable() {
         this.networkTransactionResponse = tokenResponse.response();
     }
 
+    @When("I verify the number of tokens with given treasury account")
+    public void getAllTokensForAccount() {
+        TokensResponse tokensByAccountId = mirrorClient.getTokensWithTreasuryAccount(
```
steven-sheehy: Use final var here and below

steven-sheehy:
Acceptance tests not passing

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/TopicFeature.java
sdimitrov9: Done

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/TopicFeature.java
sdimitrov9: Done

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/TokenFeature.java
sdimitrov9: Addressed by adding a new step after the minting one - "I verify the number of minted NFTs for token is expected"

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/TokenFeature.java
sdimitrov9: Not necessarily, but I changed the logic to filter both responses for the expected token and assert them for equality.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/TokenFeature.java
sdimitrov9: Done, logic is now combined in "I verify the number of tokens is as expected".

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/TokenFeature.java
sdimitrov9: Done

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/TokenFeature.java
sdimitrov9: Modified

sdimitrov9:
> Acceptance tests not passing

Passing now.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/TokenFeature.java
```diff
@@ -90,8 +93,14 @@ public class TokenFeature extends AbstractFeature {
 
     private TokenResponse tokenResponse;
 
+    private TokensResponse tokensWithTreasuryAccountBeforeCreation;
```
steven-sheehy: This is unused by multiple steps and can be reduced to a local variable.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/TokenFeature.java
```diff
@@ -90,8 +93,14 @@ public class TokenFeature extends AbstractFeature {
...
+
     private TransactionDetail transactionDetail;
 
+    private List<@Valid Nft> tokenNFTsBeforeMint = List.of();
```
steven-sheehy: Valid annotation doesn't work or apply here.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/TokenFeature.java
```diff
@@ -335,6 +347,39 @@ public void createNewTokenKycNotApplicable() {
         this.networkTransactionResponse = tokenResponse.response();
     }
 
+    @RetryAsserts
+    @When("I verify the number of tokens is expected")
+    public void getAllTokensForAccount() {
```
steven-sheehy: Not a getter.

```suggestion
    public void verifyTokenRelationships() {
```

steven-sheehy (changes requested):

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/TokenFeature.java
sdimitrov9: Removed

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/TokenFeature.java
sdimitrov9: Removed

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/TokenFeature.java
sdimitrov9: Addressed

PR #11912, opened 2025-09-02
Added accounts and balances acceptance tests

### LINKED ISSUE #11865: Add missing accounts and balances api to acceptance tests
### Problem

We're missing acceptance test coverage for some mirror node REST APIs.

### Solution

Based on #4051 , add below apis coverage to acceptance tests 
/api/v1/accounts - List all accounts
/api/v1/accounts/{id}/rewards - Staking rewards
/api/v1/balances - Account balances endpoint

### Alternatives

_No response_

ashumahajan:
**Description**:

Added below apis coverage to acceptance tests
/api/v1/accounts - List all accounts
/api/v1/accounts/{id}/rewards - Staking rewards
/api/v1/balances - Account balances endpoint

Fixes #11865 

**Checklist**

### DIFF
--- a/test/src/test/java/org/hiero/mirror/test/e2e/acceptance/client/AccountClient.java
+++ b/test/src/test/java/org/hiero/mirror/test/e2e/acceptance/client/AccountClient.java
@@ -7,6 +7,7 @@ import com.hedera.hashgraph.sdk.AccountAllowanceApproveTransaction;
 import com.hedera.hashgraph.sdk.AccountCreateTransaction;
 import com.hedera.hashgraph.sdk.AccountDeleteTransaction;
 import com.hedera.hashgraph.sdk.AccountId;
+import com.hedera.hashgraph.sdk.AccountUpdateTransaction;
 import com.hedera.hashgraph.sdk.ContractId;
 import com.hedera.hashgraph.sdk.EvmAddress;
 import com.hedera.hashgraph.sdk.Hbar;
@@ -330,6 +331,32 @@ public class AccountClient extends AbstractNetworkClient {
                 spender,
                 tokenId,
                 response.getTransactionId());
+        return response;
+    }
+
+    public NetworkTransactionResponse stakeAccountToNode(ExpandedAccountId accountId, long nodeId) {
+        var transaction = new AccountUpdateTransaction()
+                .setAccountId(accountId.getAccountId())
+                .setStakedNodeId(nodeId)
+                .setDeclineStakingReward(false) // Ensure account can receive rewards
+                .freezeWith(client)
+                .sign(accountId.getPrivateKey());
+
+        var response = executeTransactionAndRetrieveReceipt(transaction);
+        log.info("Staked account {} to node {} via {}", accountId.getAccountId(), nodeId, response.getTransactionId());
+        return response;
+    }
+
+    public NetworkTransactionResponse unstakeAccount(ExpandedAccountId accountId) {
+        var transaction = new AccountUpdateTransaction()
+                .setAccountId(accountId.getAccountId())
+                .clearStakedAccountId()
+                .clearStakedNodeId()
+                .freezeWith(client)
+                .sign(accountId.getPrivateKey());
+
+        var response = executeTransactionAndRetrieveReceipt(transaction);
+        log.info("Unstaked account {} via {}", accountId.getAccountId(), response.getTransactionId());
         return response;
     }
 
--- a/test/src/test/java/org/hiero/mirror/test/e2e/acceptance/client/MirrorNodeClient.java
+++ b/test/src/test/java/org/hiero/mirror/test/e2e/acceptance/client/MirrorNodeClient.java
@@ -27,6 +27,8 @@ import org.awaitility.Awaitility;
 import org.awaitility.Durations;
 import org.hiero.mirror.rest.model.AccountBalanceTransactions;
 import org.hiero.mirror.rest.model.AccountInfo;
+import org.hiero.mirror.rest.model.AccountsResponse;
+import org.hiero.mirror.rest.model.BalancesResponse;
 import org.hiero.mirror.rest.model.BlocksResponse;
 import org.hiero.mirror.rest.model.ContractActionsResponse;
 import org.hiero.mirror.rest.model.ContractCallRequest;
@@ -46,6 +48,7 @@ import org.hiero.mirror.rest.model.NftAllowancesResponse;
 import org.hiero.mirror.rest.model.NftTransactionHistory;
 import org.hiero.mirror.rest.model.Nfts;
 import org.hiero.mirror.rest.model.Schedule;
+import org.hiero.mirror.rest.model.StakingRewardsResponse;
 import org.hiero.mirror.rest.model.TokenAirdropsResponse;
 import org.hiero.mirror.rest.model.TokenAllowancesResponse;
 import org.hiero.mirror.rest.model.TokenBalancesResponse;
@@ -418,6 +421,82 @@ public class MirrorNodeClient {
 
     public boolean hasPartialState() {
         return partialStateSupplier.get();
+    }
+
+    public AccountsResponse getAccounts() {
+        log.debug("Retrieving all accounts from Mirror Node");
+        return callRestEndpoint("/accounts", AccountsResponse.class);
+    }
+
+    public AccountsResponse getAccounts(int limit) {
+        log.debug("Retrieving accounts with limit {} from Mirror Node", limit);
+        return callRestEndpoint("/accounts?limit={limit}", AccountsResponse.class, limit);
+    }
+
+    public StakingRewardsResponse getAccountRewards(String accountId) {
+        log.debug("Retrieving staking rewards for account '{}' from Mirror Node", accountId);
+        return callRestEndpoint("/accounts/{accountId}/rewards", StakingRewardsResponse.class, accountId);
+    }
+
+    public StakingRewardsResponse getAccountRewards(String accountId, int limit) {
+        log.debug("Retrieving staking rewards for account '{}' with limit {} from Mirror Node", accountId, limit);
+        return callRestEndpoint(
+                "/accounts/{accountId}/rewards?limit={limit}", StakingRewardsResponse.class, accountId, limit);
+    }
+
+    public BalancesResponse getBalances() {
+        log.debug("Retrieving all balances from Mirror Node");
+        return callRestEndpoint("/balances", BalancesResponse.class);
+    }
+
+    public BalancesResponse getBalances(int limit) {
+        log.debug("Retrieving balances with limit {} from Mirror Node", limit);
+        return callRestEndpoint("/balances?limit={limit}", BalancesResponse.class, limit);
+    }
+
+    public BalancesResponse getBalancesByAccountId(String accountId) {
+        log.debug("Retrieving balances for account '{}' from Mirror Node", accountId);
+        return callRestEndpoint("/balances?account.id={accountId}", BalancesResponse.class, accountId);
+    }
+
+    public BalancesResponse getBalancesByAccountId(String accountId, int limit) {
+        log.debug("Retrieving balances for account '{}' with limit {} from Mirror Node", accountId, limit);
+        return callRestEndpoint(
+                "/balances?account.id={accountId}&limit={limit}", BalancesResponse.class, accountId, limit);
+    }
+
+    public BalancesResponse getBalancesByAccountBalance(String balance) {
+        log.debug("Retrieving balances with account.balance '{}' from Mirror Node", balance);
+        return callRestEndpoint("/balances?account.balance={balance}", BalancesResponse.class, balance);
+    }
+
+    public BalancesResponse getBalancesByPublicKey(String publicKey) {
+        log.debug("Retrieving balances for public key '{}' from Mirror Node", publicKey);
+        return callRestEndpoint("/balances?account.publickey={publicKey}", BalancesResponse.class, publicKey);
+    }
+
+    public BalancesResponse getBalancesByTimestamp(String timestamp) {
+        log.debug("Retrieving balances at timestamp '{}' from Mirror Node", timestamp);
+        return callRestEndpoint("/balances?timestamp={timestamp}", BalancesResponse.class, timestamp);
+    }
+
+    public BalancesResponse getBalancesWithOrder(String order) {
+        log.debug("Retrieving balances with order '{}' from Mirror Node", order);
+        return callRestEndpoint("/balances?order={order}", BalancesResponse.class, order);
+    }
+
+    public BalancesResponse getBalancesWithAllParams(
+            String accountId, String accountBalance, String publicKey, int limit, String order, String timestamp) {
+        log.debug("Retrieving balances with all parameters from Mirror Node");
+        return callRestEndpoint(
+                "/balances?account.id={accountId}&account.balance={accountBalance}&account.publickey={publicKey}&limit={limit}&order={order}&timestamp={timestamp}",
+                BalancesResponse.class,
+                accountId,
+                accountBalance,
+                publicKey,
+                limit,
+                order,
+                timestamp);
     }
 
     private <T> T callRestEndpoint(String uri, Class<T> classType, Object... uriVariables) {
--- a/test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/AccountFeature.java
+++ b/test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/AccountFeature.java
@@ -24,7 +24,9 @@ import lombok.CustomLog;
 import lombok.Data;
 import lombok.RequiredArgsConstructor;
 import org.hiero.mirror.common.CommonProperties;
+import org.hiero.mirror.rest.model.AccountsResponse;
 import org.hiero.mirror.rest.model.CryptoAllowance;
+import org.hiero.mirror.rest.model.StakingRewardsResponse;
 import org.hiero.mirror.rest.model.TransactionByIdResponse;
 import org.hiero.mirror.rest.model.TransactionDetail;
 import org.hiero.mirror.rest.model.TransactionTransfersInner;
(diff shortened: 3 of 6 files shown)


test/src/test/java/org/hiero/mirror/test/e2e/acceptance/client/MirrorNodeClient.java
```diff
@@ -438,6 +441,41 @@ public boolean hasPartialState() {
         return partialStateSupplier.get();
     }
 
+    public AccountsResponse getAccounts() {
+        log.debug("Retrieving all accounts from Mirror Node");
+        return callRestEndpoint("/accounts", AccountsResponse.class);
+    }
+
+    public AccountsResponse getAccounts(int limit) {
```
steven-sheehy: Combine these two

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/client/MirrorNodeClient.java
```diff
@@ -438,6 +441,41 @@ public boolean hasPartialState() {
...
+        log.debug("Retrieving accounts with limit {} from Mirror Node", limit);
+        return callRestEndpoint("/accounts?limit={limit}", AccountsResponse.class, limit);
+    }
+
+    public StakingRewardsResponse getAccountRewards(String accountId) {
+        log.debug("Retrieving staking rewards for account '{}' from Mirror Node", accountId);
+        return callRestEndpoint("/accounts/{accountId}/rewards", StakingRewardsResponse.class, accountId);
+    }
+
+    public StakingRewardsResponse getAccountRewards(String accountId, int limit) {
```
steven-sheehy: Combine these two

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/AccountFeature.java
```diff
@@ -258,4 +260,76 @@ private void setCryptoAllowance(String accountName, long amount) {
...
+        assertThat(accountsResponse.getLinks()).isNotNull();
+        log.info(
+                "Retrieved {} accounts from Mirror Node (no limit specified)",
+                accountsResponse.getAccounts().size());
+    }
+
+    @Then("the mirror node REST API should return the accounts list with limit {int}")
+    public void verifyAccountsListWithLimit(int limit) {
+        log.info("Calling getAccounts({}) with limit parameter", limit);
+        AccountsResponse accountsResponse = mirrorClient.getAccounts(limit);
```
steven-sheehy: In general we should use var for all non-primitive local variables. And use final. Same comment elsewhere in diff.

```suggestion
        final var accountsResponse = mirrorClient.getAccounts(limit);
```

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/AccountFeature.java
```diff
@@ -258,4 +260,76 @@ private void setCryptoAllowance(String accountName, long amount) {
         assertNotNull(networkTransactionResponse.getTransactionId());
         assertNotNull(networkTransactionResponse.getReceipt());
     }
+
+    @Then("the mirror node REST API should return the accounts list")
+    public void verifyAccountsList() {
+        log.info("Calling getAccounts() without limit parameter");
```
steven-sheehy: Remove all log statements in feature classes. It will already log the step name which should be descriptive enough.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/AccountFeature.java
```diff
@@ -258,4 +260,76 @@ private void setCryptoAllowance(String accountName, long amount) {
...
+    }
+
+    @Then("the mirror node REST API should return the accounts list with limit {int}")
+    public void verifyAccountsListWithLimit(int limit) {
+        log.info("Calling getAccounts({}) with limit parameter", limit);
+        AccountsResponse accountsResponse = mirrorClient.getAccounts(limit);
+        assertThat(accountsResponse).isNotNull();
+        assertThat(accountsResponse.getAccounts()).isNotNull();
+        assertThat(accountsResponse.getAccounts().size()).isLessThanOrEqualTo(limit);
+        assertThat(accountsResponse.getLinks()).isNotNull();
```
steven-sheehy: We should use chaining to assert. Same comment elsewhere.

```suggestion
        assertThat(accountsResponse)
            .isNotNull()
            .satisfies(r -> assertThat(r.getLinks).isNotNull())
            .extracting(AccountsResponse::getAccounts)
            .isNotNull()
            .hasSizeLessThanOrEqualTo(limit);
```

test/src/test/resources/features/balances/balances.feature
```diff
@@ -0,0 +1,18 @@
+@balances @fullsuite @acceptance
+Feature: Balances API Coverage Feature
+
+  Scenario Outline: Balances API validation
+    Given I have created a new account with initial balance of <initialBalance> tinybars
```
steven-sheehy: We should not create a new account just to test balances as it's costly and an existing account like ALICE should suffice.

steven-sheehy: Our features are generally focused around entity types, not API endpoints. So I don't think we need a separate balances.feature. We should add this scenario into the account.feature.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/BalancesFeature.java
```diff
@@ -0,0 +1,195 @@
...
+                .buildQueryParam("account.id", accountId)
+                .build();
+        balancesResponse = mirrorClient.getBalancesForQuery(params);
+        assertThat(balancesResponse).isNotNull();
+        assertThat(balancesResponse.getBalances()).isNotNull();
+        assertThat(balancesResponse.getLinks()).isNotNull();
+        // Unable to verify this as account balances are
+        // sometimes not available by the time this step is executed!
+        //        assertThat(balancesResponse.getBalances()).isNotEmpty();
+        //        assertThat(balancesResponse.getBalances().get(0).getBalance()).isEqualTo(initialBalance);
```
steven-sheehy: We should not have commented out code on main.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/client/AccountClient.java
```diff
@@ -333,6 +334,32 @@ public NetworkTransactionResponse deleteNftAllowance(TokenId tokenId, AccountId
...
+    public NetworkTransactionResponse unstakeAccount(ExpandedAccountId accountId) {
+        var transaction = new AccountUpdateTransaction()
+                .setAccountId(accountId.getAccountId())
+                .clearStakedAccountId()
+                .clearStakedNodeId()
+                .freezeWith(client)
+                .sign(accountId.getPrivateKey());
+
+        var response = executeTransactionAndRetrieveReceipt(transaction);
+        log.info("Unstaked account {} via {}", accountId.getAccountId(), response.getTransactionId());
```
steven-sheehy: This is unused. I don't think we need to test unstaking.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/AccountFeature.java
```diff
@@ -258,4 +260,76 @@ private void setCryptoAllowance(String accountName, long amount) {
...
+    @When("I stake the account to node {long}")
+    public void stakeAccountToNode(long nodeId) {
+        networkTransactionResponse = accountClient.stakeAccountToNode(senderAccountId, nodeId);
+        assertNotNull(networkTransactionResponse.getTransactionId());
+        assertNotNull(networkTransactionResponse.getReceipt());
+        log.info("Account {} staked to node {}", senderAccountId.getAccountId(), nodeId);
+    }
+
+    @When("I unstake the account")
+    public void unstakeAccount() {
```
steven-sheehy: Unused. Can remove

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/AccountFeature.java
```diff
@@ -258,4 +260,76 @@ private void setCryptoAllowance(String accountName, long amount) {
...
+                rewardsResponse.getRewards().size(),
+                accountId,
+                limit);
+    }
+
+    @When("I stake the account to node {long}")
+    public void stakeAccountToNode(long nodeId) {
+        networkTransactionResponse = accountClient.stakeAccountToNode(senderAccountId, nodeId);
+        assertNotNull(networkTransactionResponse.getTransactionId());
+        assertNotNull(networkTransactionResponse.getReceipt());
```
steven-sheehy: We should always use assertThat(). Same comment elsewhere.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/BalancesFeature.java
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    private static final class QueryParamsBuilder {
```
steven-sheehy: I think we're doing something wrong if we're building HTTP query params in the business oriented feature. The query param names, concatenating them, etc should be handled by the MirrorNodeClient.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/AccountFeature.java
```diff
@@ -258,4 +260,76 @@ private void setCryptoAllowance(String accountName, long amount) {
...
+        assertThat(accountsResponse).isNotNull();
+        assertThat(accountsResponse.getAccounts()).isNotNull();
+        assertThat(accountsResponse.getLinks()).isNotNull();
+        log.info(
+                "Retrieved {} accounts from Mirror Node (no limit specified)",
+                accountsResponse.getAccounts().size());
+    }
+
+    @Then("the mirror node REST API should return the accounts list with limit {int}")
+    public void verifyAccountsListWithLimit(int limit) {
```
steven-sheehy: We don't need to separately verify accounts list with a limit and without. In general, we don't want to use acceptance tests to verify every query parameter combination. REST integration tests should already provide good coverage. We probably only want to test important query parameters that cause entirely different SQL statements to be generated.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/AccountFeature.java
```diff
@@ -258,4 +260,76 @@ private void setCryptoAllowance(String accountName, long amount) {
...
+        log.info("Calling getAccounts({}) with limit parameter", limit);
+        AccountsResponse accountsResponse = mirrorClient.getAccounts(limit);
+        assertThat(accountsResponse).isNotNull();
+        assertThat(accountsResponse.getAccounts()).isNotNull();
+        assertThat(accountsResponse.getAccounts().size()).isLessThanOrEqualTo(limit);
```
steven-sheehy: We should also verify the size is greater than zero since we know we just created an account.

test/src/test/resources/features/account/account.feature
```diff
@@ -5,18 +5,21 @@ Feature: Account Coverage Feature
   Scenario Outline: Create crypto account
     When I create a new account with balance <amount> tℏ
     Then the new balance should reflect cryptotransfer of <amount>
+    And the mirror node REST API should return the accounts list
+    When I stake the account to node <nodeId>
```
steven-sheehy: We should move this to the staking specific test below

test/src/test/resources/features/account/account.feature
```diff
@@ -26,3 +29,15 @@ Feature: Account Coverage Feature
       | amount | keyType   |
       | 1      | "ED25519" |
       | 1      | "ECDSA"   |
+
+  @rewards @acceptance
+  Scenario Outline: Validate account staking rewards API
+    Given I send <amount> tℏ to <accountName>
+    Then the mirror node REST API should return the staking rewards for the account
```
steven-sheehy: You will need to wait for the transfer transaction to propagate to the mirror node.

test/src/test/resources/features/account/account.feature
```diff
@@ -26,3 +29,15 @@ Feature: Account Coverage Feature
...
+    And the mirror node REST API should return the staking rewards for the account with limit <rewardLimit>
```
steven-sheehy: Can remove one of the two. Probably we can drop the examples and move it inline since there's only one example.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/BalancesFeature.java
```diff
@@ -0,0 +1,178 @@
...
+        createdAccountId = accountClient.createNewAccount(initialBalance);
+        log.info(
+                "Created account {} with initial balance {} tinybars", createdAccountId.getAccountId(), initialBalance);
+    }
+
+    @When("I query the mirror node REST API for balances")
+    public void queryBalances() {
+        log.info("Calling getBalances() to retrieve all balances");
+        balancesResponse = mirrorClient.getBalancesForQuery(null);
+    }
```
steven-sheehy: We don't usually separate querying of REST API with its response verification. Should combine.

test/src/test/resources/features/balances/balances.feature
```diff
@@ -0,0 +1,17 @@
...
+    When I query the mirror node REST API for balances
+    Then the mirror node REST API should return balances list
+    And the mirror node REST API should return balances list with limit <limit>
+    And the mirror node REST API should return balances with order <order>
+    And the mirror node REST API should return balances filtered by account balance <balanceFilter>
+    When I query historical balances at timestamp <timestamp>
+    Then the mirror node REST API should return balances at timestamp <timestamp>
+    When I query balances with all parameters for account
```
steven-sheehy: I don't think we can test balance queries with a timestamp since it depends upon the 15 minute balance service first running. We don't want the acceptance tests taking up to 15 minutes to execute.

test/src/test/resources/features/balances/balances.feature
```diff
@@ -0,0 +1,17 @@
+@balances @fullsuite @acceptance
+Feature: Balances API Coverage Feature
+
+  Scenario Outline: Balances API validation
+    Given I have created a new account with initial balance of <initialBalance> tinybars
+    When I query the mirror node REST API for balances
+    Then the mirror node REST API should return balances list
+    And the mirror node REST API should return balances list with limit <limit>
+    And the mirror node REST API should return balances with order <order>
+    And the mirror node REST API should return balances filtered by account balance <balanceFilter>
```
steven-sheehy: Since we can't test timestamp param, we only really need one call to balances to verify it. Probably update existing create account test to do a call to `/api/v1/balances?account.id=<createdId>&limit=1&order=desc`. Then we should directly verify its balance info with our expectation based upon its previous transactions in test.

test/src/test/resources/features/account/account.feature
ashumahajan: what do you think we should do to wait?

test/src/test/resources/features/balances/balances.feature
ashumahajan: In that case you think we should move this verification to AccountFeature instead and remove this newly added BalanceFeature

test/src/test/resources/features/account/account.feature
steven-sheehy: There's an existing method `verifyMirrorTransactionsResponse`

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/client/MirrorNodeClient.java
```diff
@@ -438,6 +441,40 @@ public boolean hasPartialState() {
         return partialStateSupplier.get();
     }
 
+    public AccountsResponse getAccounts() {
+        return getAccounts(null);
```
bilyana-gospodinova: Passing `null` seems a little odd. Perhaps we can set some default limit value (e.g. 10, 15, 25) as the REST API internally would set a limit if we don't pass explicitly.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/AccountFeature.java
steven-sheehy: Not addressed yet.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/client/MirrorNodeClient.java
```diff
@@ -438,6 +441,27 @@ public boolean hasPartialState() {
         return partialStateSupplier.get();
     }
 
+    public AccountsResponse getAccounts(Integer limit) {
+        if (limit == null) {
+            return callRestEndpoint("/accounts", AccountsResponse.class);
```
steven-sheehy: This is dead code. The rest of MirrorNodeClient only accepts a primitive limit and callers always set a default limit. We should be consistent and do the same.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/client/MirrorNodeClient.java
```diff
@@ -438,6 +441,27 @@ public boolean hasPartialState() {
...
+        } else {
+            return callRestEndpoint("/accounts?limit={limit}", AccountsResponse.class, limit);
+        }
+    }
+
+    public StakingRewardsResponse getAccountRewards(String accountId, Integer limit) {
+        if (limit == null) {
+            return callRestEndpoint("/accounts/{accountId}/rewards", StakingRewardsResponse.class, accountId);
```
steven-sheehy: Same. Remove and change to primitive and always pass a default.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/client/AccountClient.java
```diff
@@ -333,6 +334,19 @@ public NetworkTransactionResponse deleteNftAllowance(TokenId tokenId, AccountId
         return response;
     }
 
+    public NetworkTransactionResponse stakeAccountToNode(ExpandedAccountId accountId, long nodeId) {
```
steven-sheehy: A signature like this would be more reusable for future use cases. The callback would customize by calling extra setters on the transaction.

```suggestion
    public NetworkTransactionResponse updateAccount(ExpandedAccountId accountId, Consumer<AccountUpdateTransaction> callback) {
```

test/src/test/resources/features/balances/balances.feature
steven-sheehy: Not addressed

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/AccountFeature.java
```diff
@@ -258,4 +264,60 @@ private void setCryptoAllowance(String accountName, long amount) {
...
+
+        assertThat(balancesResponse)
+                .isNotNull()
+                .satisfies(r -> assertThat(r.getLinks()).isNotNull())
+                .extracting(BalancesResponse::getBalances)
+                .isNotNull()
+                .asInstanceOf(InstanceOfAssertFactories.LIST);
+    }
+
+    @And("the mirror node REST API should return the accounts list with limit {int}")
```
steven-sheehy: We should combine `the mirror node REST API should return the accounts list` and this. Probably we can just hardcode the limit here instead of making it configurable. We should keep the verification of size >= 1

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/AccountFeature.java
steven-sheehy: The create account flow still doesn't verify at least one entry in accounts list.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/AccountFeature.java
```diff
@@ -258,4 +264,60 @@ private void setCryptoAllowance(String accountName, long amount) {
...
+        AccountsResponse accountsResponse = mirrorClient.getAccounts(DEFAULT_LIMIT);
+        assertThat(accountsResponse)
+                .isNotNull()
+                .satisfies(r -> assertThat(r.getLinks()).isNotNull())
+                .extracting(AccountsResponse::getAccounts)
+                .isNotNull();
+    }
+
+    @Then("the mirror node REST API should return correct balance of {long}")
+    public void verifyAccountBalance(long initialBalance) {
```
steven-sheehy: Initial balance not asserted anywhere.

test/src/test/resources/features/balances/balances.feature
ashumahajan: I have already removed balance.feature file. It no longer exists

test/src/test/resources/features/account/account.feature
```diff
@@ -26,3 +29,16 @@ Feature: Account Coverage Feature
       | amount | keyType   |
       | 1      | "ED25519" |
       | 1      | "ECDSA"   |
+
+  @rewards @acceptance
+  Scenario Outline: Validate account staking rewards API
+    When I create a new account with balance <amount> tℏ
```
steven-sheehy: We should not create a new account just to test staking as it's costly and an existing account like ALICE should suffice.

test/src/test/resources/features/balances/balances.feature
steven-sheehy: Ok, I saw we create an account still but that's for staking. Will add a separate comment then.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/AccountFeature.java
ashumahajan: This is because we can't verify since the balances are not updated frequently. I had this earlier, but removed it. The testcase only adds coverage. Updated the method signature and description

steven-sheehy:
Some failing acceptance tests

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/AccountFeature.java
```diff
@@ -258,4 +264,52 @@ private void setCryptoAllowance(String accountName, long amount) {
...
+        assertThat(accountsResponse)
+                .isNotNull()
+                .satisfies(r -> assertThat(r.getLinks()).isNotNull())
+                .extracting(AccountsResponse::getAccounts)
+                .isNotNull()
+                .asInstanceOf(InstanceOfAssertFactories.LIST)
+                .hasSizeBetween(1, DEFAULT_LIMIT);
+    }
+
+    @Then("the mirror node REST API should return the balances")
```
IvanKavaldzhiev, line 280: Probably we can specify - 

the mirror node REST API should return the balance for account {string}

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/AccountFeature.java
```diff
@@ -258,4 +264,52 @@ private void setCryptoAllowance(String accountName, long amount) {
         assertNotNull(networkTransactionResponse.getTransactionId());
         assertNotNull(networkTransactionResponse.getReceipt());
     }
+
+    @Then("the mirror node REST API should return the accounts list")
```
IvanKavaldzhiev: IMO, a better wording might be:

1. return the accounts
2. return the list of accounts

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/AccountFeature.java
ashumahajan: Tried that before. But balances in mirror node are not recorded immediately and I believe happens every 5 mins. This testcase would fail

IvanKavaldzhiev (approved):
LGTM

PR #11919, opened 2025-09-02
Add script to copy citus cluster from live environment

### LINKED ISSUE #11640: Ad-hoc staging environment
### Problem

We need a way to spin up a staging environment to run performance tests every release without incurring the cost of it running all the time.

### Solution

Create a workflow that can:
* Restore latest full snapshot from mainnet to mainnet-staging
* Spin up resources from zero
* Run k6
* Wait for k6 to complete
* Scale down to zero

### Alternatives

_No response_

jnels124:
**Description**:
* add script to facilitate copying live environment to another cluster

**Related issue(s)**:

Fixes #11640

**Notes for reviewer**:

**Checklist**


steven-sheehy (commented):
In a follow up PR so we don't lose our diff, I think we'll want to move all these scripts outside the docs folder to the `tools` folder and also put them into a sub-folder based upon runbook. It's not clear which scripts are invoked directly or are helper scripts and sub-folders will help.

jnels124:
> In a follow up PR so we don't lose our diff, I think we'll want to move all these scripts outside the docs folder to the `tools` folder and also put them into a sub-folder based upon runbook. It's not clear which scripts are invoked directly or are helper scripts and sub-folders will help.

Will do. I will create a follow up to address this and also test the github action

docs/runbook/copy-live-environment.md
```diff
@@ -0,0 +1,35 @@
...
+2. If you want to copy the environment, run k6 tests, and automatically tear down resources, set `WAIT_FOR_K6=true` and
+   `DEFAULT_POOL_NAME`
+3. Run script and follow along with all prompts. To auto confirm destructive operations, set `AUTO_CONFIRM=true`
+4. You can skip any prompts or inputs by setting the following variables. If they are not set you will be prompted for
+   their values
+
+- `GCP_SNAPSHOT_PROJECT`
+- `GCP_TARGET_PROJECT`
+- `GCP_K8S_TARGET_CLUSTER_REGION`
+- `GCP_K8S_TARGET_CLUSTER_NAME`
```
steven-sheehy, line 31: You can't just set them you also have to export them, right?

docs/runbook/scripts/input-utils.sh
```diff
@@ -0,0 +1,96 @@
...
+    log "Region '${region}' does not exist in project '${gcpProject}'. Exiting"
+    exit 1
+  }
+
+  echo "${region}"
+}
+
+function promptGcpClusterName() {
+  local project="$1"
+  local region="$2"
```
steven-sheehy: We shouldn't need to accept them as input if they're already set as global variables. Probably just need to export them then can use them everywhere instead of passing them around as local variables. Same thing elsewhere.

docs/runbook/scripts/snapshot-utils.sh
```diff
@@ -0,0 +1,818 @@
+#!/usr/bin/env bash
+
+# SPDX-License-Identifier: Apache-2.0
+
+set -euo pipefail
+
+source ./utils.sh
+
+function waitUntilOutOfRecovery() {
```
steven-sheehy: This seems to duplicate the one in utils.sh?

docs/runbook/scripts/copy-live-environment.sh
```diff
@@ -0,0 +1,447 @@
...
+elif ! contextExists "${K8S_TARGET_CLUSTER_CONTEXT}"; then
+  log "context ${K8S_TARGET_CLUSTER_CONTEXT} doesn't exist"
+  exit 1
+fi
+
+snapshotSource
+restoreTarget
+deleteSnapshots
+
+if [[ "${WAIT_FOR_K6}" ]]; then
```
steven-sheehy, line 440: So k6 will run automatically just because we're scaling the env up?

docs/runbook/copy-live-environment.md
jnels124: Just need to set them and they will be available to the current shell.

docs/runbook/scripts/input-utils.sh
jnels124: Will double check this. There are some places where we call with values from source and target but will double check

docs/runbook/scripts/snapshot-utils.sh
jnels124: oops meant to remove that. Thanks

docs/runbook/scripts/copy-live-environment.sh
jnels124: correct. The process is we make the changes to the helm release and merge but the hr will be suspended so nothing will happen. Then we run this script (or workflow) which will replace the environment, resume the helm release, run acceptance tests, and then k6 will run

docs/runbook/copy-live-environment.md
steven-sheehy: Yes, but by exporting them they're available to the util sub-shells and don't have to pass them around.

jnels124: They are available anyways because those shells are spawned by this one. I will double check but i dont think export is needed unless we want them available to the non executing process

docs/runbook/scripts/snapshot-utils.sh
```diff
@@ -0,0 +1,791 @@
...
+  if [[ -z "${GCP_K8S_TARGET_CLUSTER_REGION}" ]]; then
+    GCP_K8S_TARGET_CLUSTER_REGION="$(promptGcpClusterRegion)"
+  fi
+
+  if [[ -z "${GCP_K8S_TARGET_CLUSTER_NAME}" ]]; then
+    GCP_K8S_TARGET_CLUSTER_NAME="$(promptGcpClusterName)"
+  fi
+
+  if [[ -z "${SNAPSHOT_ID}" ]]; then
+    SNAPSHOT_ID="$(promptSnapshotId))"
```
nirbosl: Extra closing ) on this line, would likely result in a runtime syntax error.
```suggestion
    SNAPSHOT_ID="$(promptSnapshotId)"
```

docs/runbook/scripts/snapshot-utils.sh
```diff
@@ -0,0 +1,791 @@
...
+
+  log "ZFS datasets renamed"
+}
+
+function replaceDisks() {
+  prepareCitusDisksForReplacement
+  if [[ "${REPLACE_DISKS}" == "true" ]]; then
+    log "Will delete disks ${DISK_PREFIX}-(${UNIQUE_NODE_IDS[*]})-zfs"
+    doContinue
+    kubectl delete sgshardedbackups.stackgres.io -n "${namespace}" --all
```
nirbosl: `namespace` var does not exist in this scope, iterate over `CITUS_NAMESPACES` to get the namespaces to delete `sgshardedbackups` from.

```suggestion
    for namespace in "${CITUS_NAMESPACES[@]}"; do
      kubectl delete sgshardedbackups.stackgres.io -n "${namespace}" --all
    done
```

docs/runbook/scripts/snapshot-utils.sh
```diff
@@ -0,0 +1,791 @@
...
+                  ${nodeInfo}"
+      exit 1
+    else
+      log "Snapshot contains all pvcs for node ${nodeId}"
+    fi
+  done
+
+  ZFS_VOLUMES="$(getZFSVolumes)"
+  NODE_ID_MAP="$(getNodeIdToPvcSnapshotsMap)"
+  mapfile -t UNIQUE_NODE_IDS < <(echo "${NODE_ID_MAP}" | jq -r 'keys[]')
```
nirbosl: `UNIQUE_NODE_IDS` was being called before assignment.

```suggestion
  log "Target disk prefix is ${DISK_PREFIX}"

  ZFS_VOLUMES="$(getZFSVolumes)"
  NODE_ID_MAP="$(getNodeIdToPvcSnapshotsMap)"
  mapfile -t UNIQUE_NODE_IDS < <(echo "${NODE_ID_MAP}" | jq -r 'keys[]')

  for nodeId in "${UNIQUE_NODE_IDS[@]}"; do
    local diskName="${DISK_PREFIX}-${nodeId}-zfs"
    local diskZone nodeInfo
    diskZone="$(echo "${nodeId}" | cut -d '-' -f 2-4)"
    nodeInfo=$(echo "${NODE_ID_MAP}" | jq -r --arg NODE_ID "${nodeId}" '.[$NODE_ID]')
    HAS_VALID_SNAPSHOT=$(echo "${nodeInfo}" | jq -r '.snapshot|length == 1')
    if [[ "${HAS_VALID_SNAPSHOT}" == "false" ]]; then
      log "Unable to find valid snapshot for node id ${nodeId} in snapshot id ${SNAPSHOT_ID}.
                  Please verify snapshots contain same namespace, pvc name and postgres version.
                  ${nodeInfo}"
      exit 1
    else
      log "Snapshot contains all pvcs for node ${nodeId}"
    fi
  done
```

docs/runbook/scripts/copy-live-environment.sh
```diff
@@ -0,0 +1,447 @@
...
+          sleep 3
+          ;;
+      esac
+    done
+  done
+}
+
+function runBackupsForAllNamespaces() {
+  if [[ "${CREATE_NEW_BACKUPS}"  != "true" ]]; then
+    retrun 0
```
nirbosl: Fix return typo

```suggestion
    return 0
```

docs/runbook/scripts/utils.sh
```diff
@@ -211,7 +242,7 @@ function scaleDeployment() {
       log "No pods found for deployment ${deploymentLabel} in namespace ${namespace}"
       return
     else
-      log "Removing pods ${deploymentPods} in ${namespace} for ${CURRENT_CONTEXT}"
+      log "Removing pods ${deploymentPods} in ${namespace} for"
```
nirbosl: Removed the leftover "for" in the log print.

```suggestion
      log "Removing pods ${deploymentPods} in ${namespace}"
```

docs/runbook/scripts/volume-snapshot.sh
```diff
@@ -5,78 +5,4 @@
 set -euo pipefail
 
 source ./utils.sh
```
nirbosl, line 7: Add missing import

```suggestion
source ./utils.sh
source ./snapshot-utils.sh
```

docs/runbook/scripts/snapshot-utils.sh
jnels124: just deleting from all namespaces

PR #11926, opened 2025-09-03
Read trace data block items for block root hash calculation

xin-hedera:
**Description**:

This PR adds support of reading trace data block items for block root hash calculation

- Read trace data block items for block root hash calcualtion 
- Make block source and record file downloader exclusive
- Add `blocknode` spring profile

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

The order of the items per consensus node 0.65 is trace data block items appear before transactional state changes, though in the HIP diagram it's the reverse. Michael explained it's more natural to have transactional state changes at the end.

This is needed since otherwise in all environments with BN, importer will just fail if there's any trace data block items.  

**Checklist**


importer/src/main/java/org/hiero/mirror/importer/downloader/record/RecordDownloaderProperties.java
```diff
@@ -37,4 +40,11 @@ public class RecordDownloaderProperties implements DownloaderProperties {
     public StreamType getStreamType() {
         return StreamType.RECORD;
     }
+
+    @PostConstruct
+    void init() {
+        if (enabled && blockProperties.isEnabled()) {
+            throw new IllegalStateException("Cannot enable both block source and record downloader");
+        }
```
steven-sheehy: I don't think it's good to couple the two property classes together. Instead, would prefer this validation moved to a common area like a post construct in `ImporterConfiguration`.

steven-sheehy (approved):
LGTM

PR #11989, opened 2025-09-15
update web3 hpa

### LINKED ISSUE #11692: Web3 gas used based HPA target causes created pods spike when rolling out update

jnels124:
**Description**:
* add custom hpa metric to capture avg gas used for namespace

**Related issue(s)**:

Fixes #11692 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/charts/hedera-mirror-web3/values.yaml
+++ b/charts/hedera-mirror-web3/values.yaml
@@ -97,13 +97,24 @@ hpa:
   behavior: {}
   enabled: false
   maxReplicas: 12
-  metrics:
+  metrics: |
     - type: Resource
       resource:
         name: cpu
         target:
           type: Utilization
           averageUtilization: 80
+    - type: Object
+      object:
+        describedObject:
+          apiVersion: v1
+          kind: Namespace
+          name: {{ include "hedera-mirror-web3.namespace" . }}
+        metric:
+          name: avg_evm_gas_used_per_second
+        target:
+          type: Value
+          value: 2750k
   minReplicas: 1
 
 image:
--- a/charts/hedera-mirror-common/values.yaml
+++ b/charts/hedera-mirror-common/values.yaml
@@ -242,6 +242,14 @@ prometheus-adapter:
         name:
           as: "requests_per_second"
         metricsQuery: "sum(rate(<<.Series>>{<<.LabelMatchers>>}[3m])) by (<<.GroupBy>>)"
+      - seriesQuery: 'hiero_mirror_web3_evm_gas_used_total{application="web3",namespace!=""}'
+        resources:
+          overrides:
+            namespace:
+              resource: "namespace"
+        name:
+          as: "avg_evm_gas_used_per_second"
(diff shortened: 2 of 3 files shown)


jnels124:
this approach chosen in favor of pr #11812

charts/hedera-mirror-web3/values.yaml
```diff
@@ -97,13 +97,24 @@ hpa:
...
   maxReplicas: 12
-  metrics:
+  metrics: |
     - type: Resource
       resource:
         name: cpu
         target:
           type: Utilization
           averageUtilization: 80
+    - type: Object
```
steven-sheehy, line 111: Probably we should add some feature flag to easily disable this extra scaling in case there's an issue?

steven-sheehy: Also, what will occur for users who don't use the common chart and who don't have a avg_evm_gas_used_per_second metric? Maybe that flag needs to be disabled by default and enabled only in our values-prod.yml

jnels124: we can already do that by overriding the hpa property. However, to make this work better when our common chart is not used, will add a separate property. If the metric is not available it will not be considered in scaling operations

steven-sheehy (dismissed):

charts/hedera-mirror-web3/values.yaml
```diff
@@ -96,14 +96,30 @@ global:
 hpa:
   behavior: {}
   enabled: false
+  gasUsed:
+    enabled: false
```
steven-sheehy, line 100: Shouldn't we enable this by default in values-prod?

jnels124: I thought you wanted to enable it in the deploy branch. I can enable in values-prod.yaml

steven-sheehy (approved):

xin-hedera (approved):
LGTM

PR #12019, opened 2025-09-19
Enhance read only estimate feature tests to make ContractCallLocal fo…

### LINKED ISSUE #11859: Verify estimate via call
### Problem

We should verify our own gas estimations via a contract call in our acceptance tests.

### Solution

For existing gas estimation acceptance tests, ensure call is invoked with the estimation to verify accuracy.

### Alternatives

_No response_

IvanKavaldzhiev:
**Description**:
This PR enhances read only operations in acceptance test features related to gas estimation. After a gas estimate for a given read only function is returned by web3, the same gas is used inside a ContractCallLocal making a request to consensus nodes, to validate that the gas is big enough to cover the fee. This will allow higher fidelity of gas estimate validation.

In additions this PR enables again previously commented estimate feature tests.

**Related issue(s)**:

Fixes #11859

**Notes for reviewer**:

**Checklist**


test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/AbstractEstimateFeature.java
```diff
@@ -204,4 +208,44 @@ private Long getGasConsumedByTransactionId(String transactionId) {
...
+        final long estimatedGas = Bytes.fromHexString(gasEstimateResponse.getResult())
+                .toBigInteger()
+                .longValueExact();
+
+        // Validate that the estimated gas is within acceptable deviation
+        assertWithinDeviation(contractMethods.getActualGas(), (int) estimatedGas, lowerDeviation, upperDeviation);
+
+        // Then validate gas is sufficient to execute ContractCallLocal
+        try {
+            contractClient.executeContractQuery(
```
steven-sheehy: As the issue states, there should be a property to enable HAPI call that should only be enabled in CI.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/EstimateFeature.java
```diff
@@ -162,34 +162,43 @@ public void setDeviations(int lower, int upper) {
 
     @Then("I call estimateGas without arguments that multiplies two numbers")
     public void multiplyEstimateCall() {
-        validateGasEstimation(encodeData(ESTIMATE_GAS, MULTIPLY_NUMBERS), MULTIPLY_NUMBERS, contractSolidityAddress);
+        validateGasEstimationAndExecuteContractCallLocal(
+                encodeData(ESTIMATE_GAS, MULTIPLY_NUMBERS),
```
steven-sheehy: Instead of changing which validation method to use everywhere if it's a view function, it would be better to revert to a single `validateGasEstimation()` and add an enum to the `SelectorInterface` implemented by the various contract enums to indicate it's a view. Something like `enum FunctionType {VIEW,PURE,PAYABLE,MUTABLE}`.

steven-sheehy (changes requested):

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/AbstractEstimateFeature.java
IvanKavaldzhiev: Added

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/EstimateFeature.java
IvanKavaldzhiev: Refactored

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/AbstractEstimateFeature.java
```diff
@@ -64,34 +72,69 @@ protected static void assertWithinDeviation(
...
+     * @param data            The function signature and method data of the contract call.
+     * @param method          The method we are going to call.
+     * @param contract        The contract object
+     * @param estimatedGas          The estimated gas to use for the ContractCallLocal execution
+     * *
+     * @throws RuntimeException If the ContractCallLocal execution has thrown an exception.
+     */
+    private void validateGasViaContractCallLocal(
+            String data, ContractMethodInterface method, DeployedContract contract, Long estimatedGas) {
+        if (featureProperties.isContractCallLocalEstimate() && VIEW.equals(method.getFunctionType())) {
```
steven-sheehy: We can probably also do for pure, right?

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/AbstractEstimateFeature.java
```diff
@@ -64,34 +72,69 @@ protected static void assertWithinDeviation(
...
+
+    protected void validateGasEstimation(String data, ContractMethodInterface method, DeployedContract contract) {
+        final var estimatedGas = validateGasEstimationWithSender(
+                data, method, contract.contractId().toEvmAddress(), Optional.empty());
+        validateGasViaContractCallLocal(data, method, contract, estimatedGas);
     }
 
-    protected void validateGasEstimation(String data, ContractMethodInterface actualGasUsed, String solidityAddress) {
-        validateGasEstimation(data, actualGasUsed, solidityAddress, Optional.empty());
+    protected void validateGasEstimation(String data, ContractMethodInterface method, String contractEvmAddress) {
```
steven-sheehy: Is this method still used? Why can't they be converted to DeployedContract? Maybe they're negative test cases

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/config/FeatureProperties.java
```diff
@@ -17,6 +17,8 @@
 @Validated
 public class FeatureProperties {
 
+    private boolean contractCallLocalEstimate = false;
```
ashumahajan, line 20: it will automatically default to false.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/AbstractEstimateFeature.java
```diff
@@ -64,34 +72,69 @@ protected static void assertWithinDeviation(
...
+     * @param contractEvmAddress        The address of the contract.
      * @param sender          The sender's address (optional).
+     * *
+     * @return the estimated gas
      * @throws AssertionError If the actual gas used is not within the acceptable deviation range.
      */
-    protected void validateGasEstimation(
-            String data, ContractMethodInterface actualGasUsed, String solidityAddress, Optional<String> sender) {
-        var contractCallRequest = ModelBuilder.contractCallRequest(actualGasUsed.getActualGas())
+    protected long validateGasEstimationWithSender(
```
steven-sheehy: There are and will never be calls to this method for views? We should probably just have one single implementation that other methods delegate to so we continue to use contractcalllocal in the future.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/EstimateFeature.java
```diff
@@ -162,34 +167,34 @@ public void setDeviations(int lower, int upper) {
 
     @Then("I call estimateGas without arguments that multiplies two numbers")
     public void multiplyEstimateCall() {
-        validateGasEstimation(encodeData(ESTIMATE_GAS, MULTIPLY_NUMBERS), MULTIPLY_NUMBERS, contractSolidityAddress);
```
steven-sheehy, line 165: Do we still need the `contractSolidityAddress` variable here and in other files?

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/config/FeatureProperties.java
IvanKavaldzhiev: I agree. I wanted to be consistent with the other boolean field in this class which is also initialized as false. I either can keep the change as is or delete the initialization in both fields.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/AbstractEstimateFeature.java
IvanKavaldzhiev: Yes, added.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/AbstractEstimateFeature.java
IvanKavaldzhiev: The purpose was to pass fungibleToken addresses as recipients for the call. However, I managed to convert the fungible TokenID to ContractID. So I removed the method.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/AbstractEstimateFeature.java
IvanKavaldzhiev: I reorganized the methods and we have one main private method that validates the gas and performs the ContractCallLocal queries.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/steps/EstimateFeature.java
IvanKavaldzhiev: Yes, it's encoded in the data passed in the request body on several places. However, there were some other fields which were not needed and I deleted them.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/config/FeatureProperties.java
ashumahajan: ok

PR #12059, opened 2025-09-26
Fix "Unhandled failure for input" error

### LINKED ISSUE #11963: Investigate "Unhandled failure for input" error
Several errors (2 for the last 24 hours from opening the ticket) of type "Unhandled failure for input" are observed in Grafana due to java.lang.UnsupportedOperationException: null.

To see the whole stacktrace filter by "|~ "Unhandled failure for input" in Grafana explorer.

bilyana-gospodinova:
**Description**:
In some rare cases when the sender of a transaction is a hollow account and we have an inner child transaction that makes a transfer from this account to another account, the first one needs to be completed. However, by calling the `TransactionExecutor` directly in the MN, we bypass the account completion (which happens in the `HandleWorkflow` on the consensus node). This PR modifies the sender hollow account in the state by setting a default key for it and this way "completing" it, so that the transaction execution can proceed as expected. After this change the transaction passed against MN and CN will be successful in both cases. 

Fixes #11963

### DIFF
--- a/web3/src/main/java/org/hiero/mirror/web3/service/TransactionExecutionService.java
+++ b/web3/src/main/java/org/hiero/mirror/web3/service/TransactionExecutionService.java
@@ -2,10 +2,10 @@
 
 package org.hiero.mirror.web3.service;
 
+import static com.hedera.node.app.hapi.utils.keys.KeyUtils.IMMUTABILITY_SENTINEL_KEY;
 import static com.hedera.services.utils.EntityIdUtils.accountIdFromEvmAddress;
 import static org.hiero.mirror.web3.convert.BytesDecoder.maybeDecodeSolidityErrorStringToReadableMessage;
 import static org.hiero.mirror.web3.state.Utils.DEFAULT_KEY;
-import static org.hiero.mirror.web3.state.Utils.HOLLOW_ACCOUNT_KEY;
 import static org.hiero.mirror.web3.state.Utils.isMirror;
 import static org.hiero.mirror.web3.validation.HexValidator.HEX_PREFIX;
 
@@ -227,7 +227,7 @@ public class TransactionExecutionService {
             throwPayerAccountNotFoundException(SENDER_NOT_FOUND);
         } else if (account.smartContract()) {
             throwPayerAccountNotFoundException(SENDER_IS_SMART_CONTRACT);
-        } else if (!account.hasKey() || account.key().equals(HOLLOW_ACCOUNT_KEY)) {
+        } else if (!account.hasKey() || account.key().equals(IMMUTABILITY_SENTINEL_KEY)) {
             // If the account is hollow, complete it in the state as a workaround
             // as this happens in HandleWorkflow in hedera-app but calling the
             // transaction executor directly skips this account completion and
--- a/web3/src/main/java/org/hiero/mirror/web3/service/TransactionExecutionService.java
+++ b/web3/src/main/java/org/hiero/mirror/web3/service/TransactionExecutionService.java
@@ -4,6 +4,8 @@ package org.hiero.mirror.web3.service;
 
 import static com.hedera.services.utils.EntityIdUtils.accountIdFromEvmAddress;
 import static org.hiero.mirror.web3.convert.BytesDecoder.maybeDecodeSolidityErrorStringToReadableMessage;
+import static org.hiero.mirror.web3.state.Utils.DEFAULT_KEY;
+import static org.hiero.mirror.web3.state.Utils.HOLLOW_ACCOUNT_KEY;
 import static org.hiero.mirror.web3.state.Utils.isMirror;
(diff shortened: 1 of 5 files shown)


web3/src/main/java/org/hiero/mirror/web3/service/TransactionExecutionService.java
```diff
@@ -225,6 +227,15 @@ private AccountID getSenderAccountID(final CallServiceParameters params) {
             throwPayerAccountNotFoundException(SENDER_NOT_FOUND);
         } else if (account.smartContract()) {
             throwPayerAccountNotFoundException(SENDER_IS_SMART_CONTRACT);
+        } else if (!account.hasKey() || account.key().equals(HOLLOW_ACCOUNT_KEY)) {
```
steven-sheehy: Can you use `KeyUtils.IMMUTABILITY_SENTINEL_KEY`?

web3/src/main/java/org/hiero/mirror/web3/state/Utils.java
```diff
@@ -20,6 +20,18 @@ public class Utils {
     public static final int EVM_ADDRESS_LEN = 20;
     public static final Key EMPTY_KEY_LIST =
             Key.newBuilder().keyList(KeyList.DEFAULT).build();
+    public static final Key DEFAULT_KEY = Key.newBuilder()
+            .keyList(KeyList.newBuilder()
+                    .keys(Key.newBuilder()
+                            .ecdsaSecp256k1(com.hedera.pbj.runtime.io.buffer.Bytes.wrap(new byte[] {
```
steven-sheehy:
```suggestion
                            .ecdsaSecp256k1(Bytes.wrap(new byte[] {
```

web3/src/main/java/org/hiero/mirror/web3/service/TransactionExecutionService.java
steven-sheehy, line 1: > Note that this way we skip one synthetic crypto update for the hollow account completion and this may result in a slightly inaccurate gas estimation.

Why would a synthetic update increase gas? Shouldn't that just cause higher payer fees which our code doesn't care about?

web3/src/main/java/org/hiero/mirror/web3/service/TransactionExecutionService.java
```diff
@@ -225,6 +227,15 @@ private AccountID getSenderAccountID(final CallServiceParameters params) {
...
+            // If the account is hollow, complete it in the state as a workaround
+            // as this happens in HandleWorkflow in hedera-app but calling the
+            // transaction executor directly skips this account completion and
+            // this results in failed transactions that would otherwise succeed
+            // against the consensus node.
+            final var writableAccountCache = ContractCallContext.get().getWriteCacheState(AccountReadableKVState.KEY);
+            final var completedAccount = account.copyBuilder().key(DEFAULT_KEY).build();
+            writableAccountCache.put(account.accountId(), completedAccount);
```
steven-sheehy, line 238: Is it possible to ask upstream to move the completion from handle to somewhere invoked by hedera-app so we can remove this workaround in the future?

steven-sheehy (changes requested):

web3/src/main/java/org/hiero/mirror/web3/service/TransactionExecutionService.java
bilyana-gospodinova: Changed

web3/src/main/java/org/hiero/mirror/web3/state/Utils.java
bilyana-gospodinova: Fixed

web3/src/main/java/org/hiero/mirror/web3/service/TransactionExecutionService.java
bilyana-gospodinova: I forwarded the question. I am waiting for a response now.

web3/src/main/java/org/hiero/mirror/web3/service/TransactionExecutionService.java
bilyana-gospodinova: It's possible. I reached out to the execution team to confirm.

steven-sheehy (approved):

web3/src/main/java/org/hiero/mirror/web3/service/TransactionExecutionService.java
bilyana-gospodinova: They confirmed that the gas consumption will not change, so we are good there.

bilyana-gospodinova: I updated the PR description to reflect this discussion.

sdimitrov9 (approved):

PR #12060, opened 2025-09-26
Fix EntityMetadataRegistry duplicate key error with multiple schemas

itsbrandondev:
## Problem
Mirror Node Importer fails with duplicate key error when database contains tables with identical column names in different schemas (experienced when updating to v0.139.0).

## Root Cause
EntityMetadataRegistry queries `information_schema.columns` without filtering by schema, causing duplicate keys when tables not included in the `public` schema have the same `account_id` column as `public.token_account` for example. This can cause errors more broadly for mirror node providers offering enriched data in other schemas for specific use cases (example, [our ERC token indexer](https://docs.hgraph.com/erc-token-data/schema-reference)).

## Solution
Filter information_schema query by the configured schema from DBProperties (default: "public").

## Changes
- Add DBProperties dependency injection
- Update SQL query to include `table_schema = ?` filter  
- Pass schema parameter from `dbProperties.getSchema()`

## Testing
- Can the Hiero Mirror Node team test this fix for us?

## Temporary Workaround (Tested)

Since EntityMetadataRegistry only scans schemas during initial startup, we can work around this issue by temporarily dropping conflicting schemas:

1. Stop services that use non-public schemas (e.g., our ERC indexer)
2. Drop conflicting schemas
3. Start the importer with v0.139.0
4. Wait for importer to be healthy (check logs)
6. Await official fix

**Attempted Workaround**

**Failed:** overridden by `EntityMetadataRegistry`

```yaml
# importer application.yml
db:
  connectionInitSql: "SET search_path TO public; SET timezone TO 'UTC';"
```

### DIFF
--- a/importer/src/main/java/org/hiero/mirror/importer/repository/upsert/EntityMetadataRegistry.java
+++ b/importer/src/main/java/org/hiero/mirror/importer/repository/upsert/EntityMetadataRegistry.java
@@ -32,6 +32,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.hibernate.metamodel.model.domain.SingularPersistentAttribute;
 import org.hiero.mirror.common.domain.UpsertColumn;
 import org.hiero.mirror.common.domain.Upsertable;
+import org.hiero.mirror.importer.db.DBProperties;
 import org.hiero.mirror.importer.exception.FieldInaccessibleException;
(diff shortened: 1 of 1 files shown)


xin-hedera (approved):
LGTM

steven-sheehy:
Thanks for your contribution! 🥂

PR #12112, opened 2025-10-02
Avoid account lookup failure for system accounts during historical calls

### LINKED ISSUE #11768: Non-zero net hbar change when handling body
### Description

The following exception is noticed in the Grafana logs:

```
2025-08-02 10:51:57.504	
2025-08-02T07:51:57.495Z ERROR tomcat-handler-180 c.h.n.a.s.t.i.h.FinalizeRecordHandler Non-zero net hbar change when handling body
TransactionBody[transactionID=TransactionID[transactionValidStart=Timestamp[seconds=1754121114, nanos=0], accountID=AccountID[shardNum=0, realmNum=0, account=OneOf[kind=ACCOUNT_NUM, value=2]], scheduled=false, nonce=0], nodeAccountID=AccountID[shardNum=0, realmNum=0, account=OneOf[kind=ACCOUNT_NUM, value=2]], transactionFee=0, transactionValidDuration=Duration[seconds=15], generateRecord=false, memo=, batchKey=null, data=OneOf[kind=CONTRACT_CALL, value=ContractCallTransactionBody[contractID=ContractID[shardNum=0, realmNum=0, contract=OneOf[kind=EVM_ADDRESS, value=00000000000000000000000000000000003c4370]], gas=15000000, amount=0, functionParameters=cdca17530000000000000000000000000000000000000000000000000000000000000040000000000000000000000000000000000000000000000000000000e8d4a51000000000000000000000000000000000000000000000000000000000000000002b0000000000000000000000000000000000163b5a0005dc000000000000000000000000000000000006f89a000000000000000000000000000000000000000000]], maxCustomFees=[]]
with fee 0; original/modified accounts claimed to be:
	Original : 
	Modified : 
	Original : 
	Modified : 
```

### Steps to reproduce

n/a

### Additional context

_No response_

### Hedera network

_No response_

### Version

v0.135.0

### Operating system

None

sdimitrov9:
**Description**:
Avoid account lookup failure for system accounts during historical calls when the account is not existing yet.

**Related issue(s)**:

Fixes #11768 

**Checklist**

### DIFF
--- a/web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AccountReadableKVState.java
+++ b/web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AccountReadableKVState.java
@@ -29,6 +29,8 @@ import org.hiero.mirror.web3.state.CommonEntityAccessor;
 public class AccountReadableKVState extends AbstractAliasedAccountReadableKVState<AccountID, Account> {
 
     public static final String KEY = "ACCOUNTS";
+    public static final int SYS_ACCOUNT_NUMBER_LOWER_LIMIT = 1;
+    public static final int SYS_ACCOUNT_NUMBER_UPPER_LIMIT = 1000;
 
     private final CommonEntityAccessor commonEntityAccessor;
     private final AliasedAccountCacheManager aliasedAccountCacheManager;
@@ -61,7 +63,7 @@ public class AccountReadableKVState extends AbstractAliasedAccountReadableKVStat
     @Override
     protected Account readFromDataSource(@Nonnull AccountID key) {
         final var timestamp = ContractCallContext.get().getTimestamp();
-        return commonEntityAccessor
+        final var acc = commonEntityAccessor
                 .get(key, timestamp)
                 .filter(entity -> entity.getType() != TOKEN)
                 .map(entity -> {
@@ -73,5 +75,20 @@ public class AccountReadableKVState extends AbstractAliasedAccountReadableKVStat
                     return account;
                 })
                 .orElse(null);
+        // In case a system account doesn't exist in a historical contract call
+        // return a dummy account to avoid errors like
+        // "Non-zero net hbar change when handling body"
+        if (acc == null && isSystemAccount(key)) {
+            return Account.newBuilder().accountId(key).tinybarBalance(1000000).build();
+        }
+        return acc;
+    }
+
+    private boolean isSystemAccount(AccountID key) {
+        if (key.hasAccountNum()) {
+            return key.accountNum() >= SYS_ACCOUNT_NUMBER_LOWER_LIMIT
+                    && key.accountNum() <= SYS_ACCOUNT_NUMBER_UPPER_LIMIT;
+        }
+        return false;
     }
 }


web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AccountReadableKVState.java
```diff
@@ -29,6 +29,8 @@
 public class AccountReadableKVState extends AbstractAliasedAccountReadableKVState<AccountID, Account> {
 
     public static final String KEY = "ACCOUNTS";
+    public static final int SYS_ACCOUNT_NUMBER_LOWER_LIMIT = 1;
```
bilyana-gospodinova: There is a `systemAccountDetector` predicate bean that can be used instead of these two constants.

web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AccountReadableKVState.java
```diff
@@ -73,5 +75,20 @@ protected Account readFromDataSource(@Nonnull AccountID key) {
                     return account;
                 })
                 .orElse(null);
+        // In case a system account doesn't exist in a historical contract call
+        // return a dummy account to avoid errors like
+        // "Non-zero net hbar change when handling body"
+        if (acc == null && isSystemAccount(key)) {
+            return Account.newBuilder().accountId(key).tinybarBalance(1000000).build();
+        }
```
steven-sheehy: And what happens when someone does `balanceOf(0.0.2)`?

web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AccountReadableKVState.java
```diff
@@ -61,7 +63,7 @@ public AccountReadableKVState(
     @Override
     protected Account readFromDataSource(@Nonnull AccountID key) {
         final var timestamp = ContractCallContext.get().getTimestamp();
-        return commonEntityAccessor
+        final var acc = commonEntityAccessor
```
steven-sheehy:
```suggestion
        final var account = commonEntityAccessor
```

web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AccountReadableKVState.java
```diff
@@ -73,5 +75,20 @@ protected Account readFromDataSource(@Nonnull AccountID key) {
...
+            return Account.newBuilder().accountId(key).tinybarBalance(1000000).build();
+        }
+        return acc;
+    }
+
+    private boolean isSystemAccount(AccountID key) {
+        if (key.hasAccountNum()) {
+            return key.accountNum() >= SYS_ACCOUNT_NUMBER_LOWER_LIMIT
+                    && key.accountNum() <= SYS_ACCOUNT_NUMBER_UPPER_LIMIT;
+        }
```
steven-sheehy: What about when council members change their node account ID and it's no longer a system account? Then all node fees to those would still be loaded.

web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AccountReadableKVState.java
```diff
@@ -73,5 +75,20 @@ protected Account readFromDataSource(@Nonnull AccountID key) {
                     return account;
                 })
                 .orElse(null);
+        // In case a system account doesn't exist in a historical contract call
+        // return a dummy account to avoid errors like
+        // "Non-zero net hbar change when handling body"
+        if (acc == null && isSystemAccount(key)) {
```
steven-sheehy: Tests?

web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AccountReadableKVState.java
sdimitrov9: If the account does not exist the request will fail anyway, so to avoid errors(because of failed account lookups in fee charging/distribution scenarios) and further confusion we should probably set the balance of the dummy account to 0.

web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AccountReadableKVState.java
sdimitrov9: Is there a way to check if an account is categorized as system one based on accountId for example?

Edit: Services are following similar approach (AddressChecks.isSystemAccount)

web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AccountReadableKVState.java
sdimitrov9: Added

web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AccountReadableKVState.java
sdimitrov9: Addressed

web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AccountReadableKVState.java
```diff
@@ -61,17 +63,32 @@ public AccountReadableKVState(
...
+        if (account == null && isSystemAccount(key)) {
+            return Account.newBuilder().accountId(key).tinybarBalance(0).build();
+        }
+        return account;
+    }
+
+    private boolean isSystemAccount(AccountID key) {
+        if (key.hasAccountNum()) {
+            return key.accountNum() >= SYS_ACCOUNT_NUMBER_LOWER_LIMIT
+                    && key.accountNum() <= SYS_ACCOUNT_NUMBER_UPPER_LIMIT;
```
steven-sheehy: This logic is off by one compared to `systemAccountDetector` bean. We should just check upper bound and use same constant for both.

```suggestion
            return key.accountNum() <= ServicesConfiguration.STRICT_SYSTEM_ACCOUNT_BOUNDARY;
```

web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AccountReadableKVState.java
steven-sheehy: Think we can ignore for now.

web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AccountReadableKVState.java
sdimitrov9: Addressed

web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AccountReadableKVState.java
sdimitrov9: Addressed, the systemAccountDetector is now used.

web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AccountReadableKVState.java
```diff
@@ -56,22 +62,38 @@ public AccountReadableKVState(
...
+                        aliasedAccountCacheManager.putAccountAlias(acc.alias(), key);
                     }
-                    return account;
+                    return acc;
                 })
                 .orElse(null);
+        // In case a system account doesn't exist in a historical contract call
+        // return a dummy account to avoid errors like
+        // "Non-zero net hbar change when handling body"
+        if (account == null) {
```
bilyana-gospodinova: The comment suggests that the if-statement should be executed only in historical calls. Currently it will be executed for latest transactions as well. If it needs to be executed in both cases, the comment should be adjested. Can't this part of code be moved and chained as a map-function in the code above? I suppose the `null` value comes from `orElse(null)`.

web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AccountReadableKVState.java
```diff
@@ -56,22 +62,38 @@ public AccountReadableKVState(
...
+            try {
+                final var accountAddress =
+                        EvmTokenUtils.toAddress(EntityId.of(key.shardNum(), key.realmNum(), key.accountNumOrThrow()));
+                if (strictSystemAccountDetector.test(accountAddress)) {
+                    return Account.newBuilder().accountId(key).tinybarBalance(0).build();
```
bilyana-gospodinova: If the balance needs to be 0, perhaps this setter can be omitted.

sdimitrov9: Addressed

web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AccountReadableKVState.java
sdimitrov9: Addressed, logic is now chained and extracted in a separate method, clarifying the intent in the method java doc.

web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AccountReadableKVState.java
```diff
@@ -72,6 +80,25 @@ protected Account readFromDataSource(@Nonnull AccountID key) {
...
+     * call for example, return a dummy account to avoid errors like
+     * "Non-zero net hbar change when handling body"
+     */
+    private Optional<Account> getDummySystemAccountIfApplicable(AccountID key) {
+        try {
+            final var accountAddress =
+                    EvmTokenUtils.toAddress(EntityId.of(key.shardNum(), key.realmNum(), key.accountNumOrThrow()));
+            if (strictSystemAccountDetector.test(accountAddress)) {
+                return Optional.of(Account.newBuilder().accountId(key).build());
+            }
```
steven-sheehy: It's inefficient to convert from AccountID to address just so we can reuse the `strictSystemAccountDetector`. Then the detector converts it from address to EntityId for its comparison. Instead, we should have a util method used by both that takes in long and detects system accounts.

jnels124 (dismissed):

web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AccountReadableKVState.java
sdimitrov9: Addressed

steven-sheehy (approved):

web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AccountReadableKVState.java
```diff
@@ -72,6 +74,22 @@ protected Account readFromDataSource(@Nonnull AccountID key) {
...
+
+    /**
+     * In case a system account doesn't exist, in a historical contract
+     * call for example, return a dummy account to avoid errors like
+     * "Non-zero net hbar change when handling body"
+     */
+    private Optional<Account> getDummySystemAccountIfApplicable(AccountID accountID) {
+        if (accountID != null && accountID.hasAccountNum()) {
+            final var accountNum = accountID.accountNum();
+            return AccountDetector.isStrictSystem(accountNum) && accountNum != 0
```
bilyana-gospodinova, line 89: Are we sure that the `accountNum` cannot be negative here? Perhaps it will be safer to check for: `accountNum > 0`.

sdimitrov9: Updated the AccountDetector utility class to properly handle negative input and return false.

PR #12114, opened 2025-10-02
HIP-1056 Reconstruct ContractBytecode

### LINKED ISSUE #11702: HIP-1056 Reconstruct ContractBytecode
### Problem

HIP-1056 is updated to remove references of `ContractBytecode` sidecar message in BS, and only write contract initcode / runtime bytecode if it's not in the transaction body itself or a file, mirrornode should follow the change to gather the information, and maintain file state for cases when bytecode is in a file, to reconstruct `ContractBytecode` sidecar.

### Solution

- runtime bytecode is in statechanges with state id `STATE_ID_CONTRACT_BYTECODE` and proto message `Bytecode`
- initcode
  - for top level contract creates, initcode is either in the transaction body or in the file referred to in the transaction body
  - for non-top level contract creates, with data `EvmTraceData.executed_initcode`, initcode is
    - `executed_initcode.initcode_bookends.deploy_bytecode | runtime bytecode | executed_initcode.initcode_bookends.metadata_bytecode` if `executed_initcode.initcode_bookends` is set
    - `executed_initcode.explicit_initcode` if set

### Alternatives

_No response_

xin-hedera:
**Description**:

This PR adds reconstruction of contract bytecode sidecar transaction record from blockstream

- Parse contract runtime bytecode from statechanges
- Reconstruct contract bytecode sidecar in `AbstractBlockTransactionTransformer`
- Add `FileDataService` to read file content from stitched state
- Add `ContractBytecodeService` to read and decode contract bytecode or ethereum call data from file 
- Add `ContractInitcodeService` for common logic reuse in `ContractCreateTransactionHandler` and `ContractResultService`
- Set amount, functionParameters, and gasLimit in `ContractResult` from decoded ethereum transaction
- Fix flaky `ContractTransformerTest`

**Related issue(s)**:

Fixes #11702

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/importer/src/main/java/org/hiero/mirror/importer/downloader/block/transformer/AbstractBlockTransactionTransformer.java
+++ b/importer/src/main/java/org/hiero/mirror/importer/downloader/block/transformer/AbstractBlockTransactionTransformer.java
@@ -13,10 +13,7 @@ import com.hedera.hapi.block.stream.output.protoc.TransactionOutput;
 import com.hedera.hapi.block.stream.output.protoc.TransactionOutput.TransactionCase;
 import com.hedera.hapi.block.stream.trace.protoc.ContractSlotUsage;
 import com.hedera.hapi.block.stream.trace.protoc.EvmTraceData;
-import com.hedera.hapi.block.stream.trace.protoc.EvmTransactionLog;
-import com.hedera.hapi.block.stream.trace.protoc.ExecutedInitcode;
 import com.hedera.hapi.block.stream.trace.protoc.SlotRead;
-import com.hedera.services.stream.proto.ContractAction;
 import com.hedera.services.stream.proto.ContractActions;
 import com.hedera.services.stream.proto.ContractBytecode;
 import com.hedera.services.stream.proto.ContractStateChange;
@@ -129,14 +126,8 @@ abstract class AbstractBlockTransactionTransformer implements BlockTransactionTr
             BlockTransaction blockTransaction,
             ContractFunctionResult.Builder contractResultBuilder,
             RecordItem.RecordItemBuilder recordItemBuilder) {
-        var evmTraceData = blockTransaction.getEvmTraceData();
-        if (evmTraceData == null) {
-            return;
-        }
-
-        transformEvmTransactionLogs(contractResultBuilder, evmTraceData.getLogsList());
-        transformSidecarRecords(
-                blockTransaction, contractResultBuilder.getContractID(), evmTraceData, recordItemBuilder);
+        transformEvmTransactionLogs(contractResultBuilder, blockTransaction.getEvmTraceData());
+        transformSidecarRecords(blockTransaction, contractResultBuilder.getContractID(), recordItemBuilder);
     }
 
     private void transformSmartContractResult(BlockTransactionTransformation transformation) {
@@ -193,11 +184,12 @@ abstract class AbstractBlockTransactionTransformer implements BlockTransactionTr
     }
 
     private void transformEvmTransactionLogs(
-            ContractFunctionResult.Builder contractResultBuilder, List<EvmTransactionLog> evmTransactionLogs) {
-        if (evmTransactionLogs.isEmpty()) {
+            ContractFunctionResult.Builder contractResultBuilder, EvmTraceData evmTraceData) {
+        if (evmTraceData == null || evmTraceData.getLogsList().isEmpty()) {
             return;
         }
 
+        var evmTransactionLogs = evmTraceData.getLogsList();
         var bloomFilters = new ArrayList<LogsBloomFilter>(evmTransactionLogs.size());
         for (var evmTransactionLog : evmTransactionLogs) {
             var bloomFilter = bloomFor(evmTransactionLog);
@@ -216,17 +208,15 @@ abstract class AbstractBlockTransactionTransformer implements BlockTransactionTr
     }
 
     private void transformContractActions(
-            Timestamp consensusTimestamp,
-            List<ContractAction> contractActions,
-            List<TransactionSidecarRecord> sidecarRecords) {
-        if (contractActions.isEmpty()) {
+            Timestamp consensusTimestamp, EvmTraceData evmTraceData, List<TransactionSidecarRecord> sidecarRecords) {
+        if (evmTraceData == null || evmTraceData.getContractActionsList().isEmpty()) {
             return;
         }
 
         var contractActionSidecarRecord = TransactionSidecarRecord.newBuilder()
                 .setConsensusTimestamp(consensusTimestamp)
                 .setActions(ContractActions.newBuilder()
-                        .addAllContractActions(contractActions)
+                        .addAllContractActions(evmTraceData.getContractActionsList())
                         .build())
                 .build();
         sidecarRecords.add(contractActionSidecarRecord);
@@ -236,7 +226,7 @@ abstract class AbstractBlockTransactionTransformer implements BlockTransactionTr
             BlockTransaction blockTransaction,
             Timestamp consnsusTimestamp,
             ContractID contractId,
-            ExecutedInitcode executedInitcode,
+            EvmTraceData evmTraceData,
             List<TransactionSidecarRecord> sidecarRecords) {
         if (!blockTransaction.getTransactionBody().hasContractCreateInstance()) {
             return;
@@ -245,7 +235,8 @@ abstract class AbstractBlockTransactionTransformer implements BlockTransactionTr
         blockTransaction.getStateChangeContext().getContractBytecode(contractId).ifPresent(runtimeBytecode -> {
             var contractBytecode =
                     ContractBytecode.newBuilder().setContractId(contractId).setRuntimeBytecode(runtimeBytecode);
-            if (!executedInitcode.equals(ExecutedInitcode.getDefaultInstance())) {
+            if (evmTraceData != null && evmTraceData.hasExecutedInitcode()) {
+                var executedInitcode = evmTraceData.getExecutedInitcode();
                 var initcodeCase = executedInitcode.getInitcodeCase();
                 switch (initcodeCase) {
                     case EXPLICIT_INITCODE -> contractBytecode.setInitcode(executedInitcode.getExplicitInitcode());
@@ -337,9 +328,9 @@ abstract class AbstractBlockTransactionTransformer implements BlockTransactionTr
 
     private void transformContractStateChanges(
             BlockTransaction blockTransaction,
-            List<ContractSlotUsage> contractSlotUsages,
+            EvmTraceData evmTraceData,
             List<TransactionSidecarRecord> sidecarRecords) {
-        if (contractSlotUsages.isEmpty()) {
+        if (evmTraceData == null || evmTraceData.getContractSlotUsagesList().isEmpty()) {
             return;
         }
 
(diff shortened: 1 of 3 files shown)


jnels124 (approved):
LGTM

importer/src/main/java/org/hiero/mirror/importer/domain/ContractInitcodeService.java
```diff
@@ -0,0 +1,18 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package org.hiero.mirror.importer.domain;
```
steven-sheehy: It's weird to have a service in a domain package. Should add a `org.hiero.mirror.importer.service` and move all these new services there.

importer/src/main/java/org/hiero/mirror/importer/service/ContractInitcodeServiceImpl.java
```diff
@@ -0,0 +1,37 @@
...
+
+@Named
+@RequiredArgsConstructor
+public final class ContractInitcodeServiceImpl implements ContractInitcodeService {
+
+    private final FileDataService fileDataService;
+
+    @Override
+    public byte[] get(ContractBytecode contractBytecode, RecordItem recordItem) {
+        if (!recordItem.getTransactionBody().hasContractCreateInstance()) {
```
steven-sheehy, line 21: For ethereum transactions with no to_address and call_data present do they still send us the initcode in the sidecar with the child contract create or is it optimized to omit?

importer/src/main/java/org/hiero/mirror/importer/service/FileDataServiceImpl.java
```diff
@@ -0,0 +1,157 @@
...
+                    } else {
+                        contents.addFirst(dbContent);
+                        return combine(contents);
+                    }
+                })
+                .orElse(null);
+    }
+
+    @Override
+    public void onFileData(@Nonnull FileData fileData) {
```
steven-sheehy, line 72: Do we really need the listeners? Can't we just inject the ParserContext and retrieve the data from there? Then the implementation is simpler and there's no need for cache invalidation on start.

importer/src/test/java/org/hiero/mirror/importer/parser/domain/BlockTransactionBuilder.java
```diff
@@ -97,6 +106,7 @@
 public class BlockTransactionBuilder {
 
     private final RecordItemBuilder recordItemBuilder;
+    private final SidecarContractMigration sidecarContractMigration;
```
steven-sheehy: Unused?

importer/src/main/java/org/hiero/mirror/importer/repository/FileDataRepository.java
```diff
@@ -50,25 +50,26 @@ List<FileData> findAddressBooksBetween(
...
+          and (transaction_type = 17
+                 or (transaction_type = 19 and length(file_data) <> 0))
+        order by consensus_timestamp desc
+        limit 1
+      )
+      select
+        (select consensus_timestamp from first) as consensus_timestamp,
+        ?1 as entity_id,
+        string_agg(file_data, '' order by consensus_timestamp) as file_data,
+        (select transaction_type from first) as transaction_type
```
steven-sheehy: Prefer that we join on first once instead of do multiple sub-queries for each row, even though that data should be in memory.

importer/src/main/java/org/hiero/mirror/importer/repository/FileDataRepository.java
```diff
@@ -50,25 +50,26 @@ List<FileData> findAddressBooksBetween(
...
+      with  first as (
+        select consensus_timestamp, transaction_type
+        from file_data
+        where entity_id = ?1
+          and consensus_timestamp <= ?2
+          and (transaction_type = 17
+                 or (transaction_type = 19 and length(file_data) <> 0))
+        order by consensus_timestamp desc
+        limit 1
+      )
```
steven-sheehy: Is there a functional or non-functional benefit to changing this query? If so, we would need to incorporate it in web3 and rest-java (for new exchangerate API) soon as they have identical queries.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/transformer/AbstractBlockTransactionTransformer.java
```diff
@@ -227,6 +231,41 @@ private void transformContractActions(
...
+                switch (initcodeCase) {
+                    case EXPLICIT_INITCODE -> contractBytecode.setInitcode(executedInitcode.getExplicitInitcode());
+                    case INITCODE_BOOKENDS ->
+                        contractBytecode.setInitcode(
+                                asInitcode(executedInitcode.getInitcodeBookends(), runtimeBytecode));
+                    default ->
+                        log.warn(
+                                "Unknown initcode case {} at {}",
+                                initcodeCase,
+                                blockTransaction.getConsensusTimestamp());
```
steven-sheehy: Should be a recoverable error

importer/src/main/java/org/hiero/mirror/importer/downloader/block/transformer/Utils.java
```diff
@@ -24,6 +25,20 @@ static Address asAddress(ContractID contractId) {
         return Address.wrap(Bytes.wrap(DomainUtils.toEvmAddress(contractId)));
     }
 
+    static ByteString asInitcode(InitcodeBookends initcodeBookends, ByteString runtimeBytecode) {
```
steven-sheehy, line 28: Why does it need to be a util? Think it's only used in one transformer class.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/transformer/AbstractBlockTransactionTransformer.java
```diff
@@ -227,6 +231,41 @@ private void transformContractActions(
...
 
+    private void transformContractBytecode(
+            BlockTransaction blockTransaction,
+            Timestamp consnsusTimestamp,
+            ContractID contractId,
+            ExecutedInitcode executedInitcode,
+            List<TransactionSidecarRecord> sidecarRecords) {
+        if (!blockTransaction.getTransactionBody().hasContractCreateInstance()) {
+            return;
+        }
```
steven-sheehy, line 233: Do we need to handle creates via ethereum transaction that have call_data in body differently?

importer/src/main/java/org/hiero/mirror/importer/service/ContractInitcodeServiceImpl.java
xin-hedera: initcode is empty in the sidecar. for blockstream, consensus node should leave `EvmTraceData.executed_initcode` unset and the initcode service will also return empty initcode.

importer/src/main/java/org/hiero/mirror/importer/service/FileDataServiceImpl.java
xin-hedera: Purpose of the cache and the added complexity enables importer to have all file data chunks in the lifecyle so as to avoid slow repository call.

importer/src/main/java/org/hiero/mirror/importer/repository/FileDataRepository.java
xin-hedera: the subquery method should be more performant, the main query itself is an aggregation. a join would create a wide row and then also requires adding the columns from the CTE to an additional group by clause.

importer/src/main/java/org/hiero/mirror/importer/repository/FileDataRepository.java
xin-hedera: the change is now it returns the first chunk's transaction type instead of null. I can sync the change to other modules.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/transformer/Utils.java
xin-hedera: make it directly testable

importer/src/main/java/org/hiero/mirror/importer/downloader/block/transformer/AbstractBlockTransactionTransformer.java
xin-hedera: nope. explained in a previous comment that in recordstream the sidecar for the child contractcreate has empty initcode.  If we think we should populate the initcode from the parent eth tx's embedded calldata or data offloaded in a file, we can create a ticket and make the change in parser.

importer/src/test/java/org/hiero/mirror/importer/parser/domain/BlockTransactionBuilder.java
xin-hedera: yup, likely some weird auto-complete IDE magic.

importer/src/main/java/org/hiero/mirror/importer/service/ContractInitcodeServiceImpl.java
```diff
@@ -0,0 +1,37 @@
...
+        }
+
+        var contractCreate = recordItem.getTransactionBody().getContractCreateInstance();
+        if (contractCreate.hasInitcode()) {
+            return DomainUtils.toBytes(contractCreate.getInitcode());
+        } else if (contractCreate.hasFileID() && recordItem.isBlockstream()) {
+            return fileDataService.get(recordItem.getConsensusTimestamp(), EntityId.of(contractCreate.getFileID()));
+        }
+
+        if (contractBytecode != null) {
```
xin-hedera, line 39: updated the logic in the service so it behaves the same as the original logic in `ContractCreateTransactionhandler`, specifically it now returns empty bytearray when the initcode is not in the transaction body and `ContractBytecode.initcode` is empty.

importer/src/main/java/org/hiero/mirror/importer/repository/FileDataRepository.java
steven-sheehy: Should be? Does the query plan show a performance improvement?

importer/src/main/java/org/hiero/mirror/importer/downloader/block/transformer/AbstractBlockTransactionTransformer.java
steven-sheehy: Maybe it's not necessary since it seems like we probably handle this in view layer.

importer/src/main/java/org/hiero/mirror/importer/service/FileDataServiceImpl.java
steven-sheehy: Right, not saying to get rid of the cache. Just wondering if we can avoid having the listeners and having to do cache cleanup on error if we instead avoid inserting into the cache until after successful commit. So you could have multiple levels of data retrieval: ParserContext (uncomitted) -> cache (committed) -> repository (committed but not in cache). Can use TransactionAwareCacheManagerProxy to insert into cache with data from ParserContext so it's automatic.

xin-hedera: In `ParserContext` file data is a list, and `ParserContext`' state (as a HashMap) can't store a list of file data updates by the entity id. So without adding custom logic and a state map from key to a list of objects, the service then needs to find the file data updates of a specific entity id from the list in `ParserContext`.

PR #12117, opened 2025-10-03
HIP-1056 Get intermediate topic message from trace data

### LINKED ISSUE #11700: HIP-1056 Get intermediate topic message info from trace data
### Problem

In the HIP-1056 update to support atomic batch transaction, `SubmitMessageTraceData` is added as one of trace data to capture intermediate topic message state changes (sequence number and running hash) for multiple ConsensusSubmitMessage transactions in a transaction unit of which state changes are only written to the top level transaction, mirrornode needs to add support to reconstruct relevant info in transaction receipt

### Solution

Note the solution in CN should be generic and apply to future transaction unit, e.g., when system contract adds consensus submit message support.

Right now, due to lack of intermediate state changes, mirrornode's solution can only fill in correct topic sequence numbers given multiple consensus submit transaction transaction in a single transaction unit. With the trace data support, mirrornode should remove the current logic and use `SubmitMessageTraceData` instead.

### Alternatives

_No response_

xin-hedera:
**Description**:

This PR supports getting intermediate topic message info from trace data

- Get topic message info from `SubmitMessageTraceData` and fall back to statechanges
- Remove topic sequence number inference logic in `StateChangeContext`
- Skip creating `TopicMessage` in `ConsensusSubmitMessageTransactionHandler` if no running hash available

**Related issue(s)**:

Fixes #11700 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/common/src/main/java/org/hiero/mirror/common/domain/transaction/StateChangeContext.java
+++ b/common/src/main/java/org/hiero/mirror/common/domain/transaction/StateChangeContext.java
@@ -150,19 +150,7 @@ public final class StateChangeContext {
     }
 
     public Optional<TopicMessage> getTopicMessage(@Nonnull TopicID topicId) {
-        return Optional.ofNullable(topicState.remove(topicId)).map(topicMessage -> {
-            if (topicMessage.getSequenceNumber() > 1) {
-                // running hash is lost for any earlier message, set it to an empty bytearray
-                // since null is not allowed
-                topicState.put(
-                        topicId,
-                        TopicMessage.builder()
-                                .runningHash(DomainUtils.EMPTY_BYTE_ARRAY)
-                                .sequenceNumber(topicMessage.getSequenceNumber() - 1)
-                                .build());
-            }
-            return topicMessage;
-        });
+        return Optional.ofNullable(topicState.remove(topicId));
     }
 
     /**
--- a/importer/src/main/java/org/hiero/mirror/importer/downloader/block/transformer/ConsensusSubmitMessageTransformer.java
+++ b/importer/src/main/java/org/hiero/mirror/importer/downloader/block/transformer/ConsensusSubmitMessageTransformer.java
@@ -18,18 +18,20 @@ final class ConsensusSubmitMessageTransformer extends AbstractBlockTransactionTr
             return;
         }
 
-        var recordBuilder = blockTransactionTransformation.recordItemBuilder().transactionRecordBuilder();
-        var topicMessage = blockTransaction
-                .getStateChangeContext()
-                .getTopicMessage(blockTransactionTransformation
-                        .getTransactionBody()
-                        .getConsensusSubmitMessage()
-                        .getTopicID())
-                .orElseThrow();
-        recordBuilder
+        var receiptBuilder = blockTransactionTransformation
+                .recordItemBuilder()
+                .transactionRecordBuilder()
                 .getReceiptBuilder()
+                .setTopicRunningHashVersion(DEFAULT_RUNNING_HASH_VERSION);
+
+        var topicMessage = blockTransaction.getTopicMessage();
+        if (topicMessage == null) {
+            log.warn("Missing topic message runningHash and sequence at {}", blockTransaction.getConsensusTimestamp());
+            return;
+        }
+
+        receiptBuilder
                 .setTopicRunningHash(DomainUtils.fromBytes(topicMessage.getRunningHash()))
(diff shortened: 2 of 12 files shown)


importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/ConsensusSubmitMessageTransactionHandler.java
```diff
@@ -56,6 +58,13 @@ protected void doUpdateTransaction(Transaction transaction, RecordItem recordIte
         var transactionBody = recordItem.getTransactionBody().getConsensusSubmitMessage();
         var transactionRecord = recordItem.getTransactionRecord();
         var receipt = transactionRecord.getReceipt();
+        if (recordItem.isBlockstream() && receipt.getTopicRunningHash().isEmpty()) {
+            log.warn(
+                    "Skip topic message from blockstream due to missing runningHash at {}",
```
steven-sheehy, line 60: How often will this scenario occur? We don't want it to fill up logs.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/transformer/ConsensusSubmitMessageTransformer.java
```diff
@@ -18,18 +18,22 @@ protected void doTransform(BlockTransactionTransformation blockTransactionTransf
...
+        var receiptBuilder = blockTransactionTransformation
+                .recordItemBuilder()
+                .transactionRecordBuilder()
                 .getReceiptBuilder()
+                .setTopicRunningHashVersion(DEFAULT_RUNNING_HASH_VERSION);
+
+        var topicMessage = blockTransaction.getTopicMessage();
+        if (topicMessage == null) {
+            log.warn(
+                    "Missing topic message runningHash and sequence number at {}",
```
steven-sheehy, line 31: When and how often will this occur?

common/src/main/java/org/hiero/mirror/common/domain/transaction/StateChangeContext.java
```diff
@@ -150,19 +150,7 @@ public Optional<TopicID> getNewTopicId() {
...
-                topicState.put(
-                        topicId,
-                        TopicMessage.builder()
-                                .runningHash(DomainUtils.EMPTY_BYTE_ARRAY)
-                                .sequenceNumber(topicMessage.getSequenceNumber() - 1)
-                                .build());
-            }
-            return topicMessage;
-        });
+        return Optional.ofNullable(topicState.remove(topicId));
```
steven-sheehy, line 153: Do we have to remove it from state context when processed? If we can keep it, then we can avoid the `AtomicReference<TopicMessage>` in `BlockTransaction` and keep it in a central spot.

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/ConsensusSubmitMessageTransactionHandler.java
xin-hedera: only happens when there is bug on our end or on consensus node end

importer/src/main/java/org/hiero/mirror/importer/downloader/block/transformer/ConsensusSubmitMessageTransformer.java
xin-hedera: same, only happens when there's bug

common/src/main/java/org/hiero/mirror/common/domain/transaction/StateChangeContext.java
xin-hedera: yeah, we should remove it, otherwise it's a possibility to run into such a scenario:

given a atomic batch transaction, with multiple inner consensus submit message transactions to the same topic, when 2 or more of the inner transactions don't have `SubmitMessageTraceData` due to upstream bug, they'll end up with the same running hash and sequence number.

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/ConsensusSubmitMessageTransactionHandler.java
steven-sheehy: We should use the recoverable error mechanism for both then so we can alert on it.

steven-sheehy (approved):

PR #12119, opened 2025-10-03
Add gh action to copy stackgres cluster

### LINKED ISSUE #11988: Test github action for copying live environment
### Problem

We added a bash script to copy a citus cluster to a dormant cluster. We need to use this script in a github action so that we can run this action to standup the dormant environment with an up to date view of mainnet and run tests.

### Solution

- add github action
- move scripts out of doc folder

### Alternatives

_No response_

jnels124:
**Description**:
* add gh action to copy stackgres cluster
* move runbook scripts to tools module
* fix bugs introduced by adding copy live environment script

**Related issue(s)**:

Fixes #11988 

**Notes for reviewer**:

**Checklist**


.github/workflows/copy-stackgres-cluster.yml
```diff
@@ -0,0 +1,161 @@
+# SPDX-License-Identifier: Apache-2.0
+
+name: Copy StackGres Citus Cluster
+
+on:
+  workflow_dispatch:
+    inputs:
+      source_cluster_location:
+        description: "Source: GKE location (zone or region)"
```
steven-sheehy, line 57: Maybe default to `us-central1` to reduce typing? Or make an enum? Same for target.

.github/workflows/copy-stackgres-cluster.yml
```diff
@@ -0,0 +1,161 @@
...
+        default: nonprod
+        required: true
+      target_cluster_location:
+        description: "Target: GKE location (zone or region)"
+        required: true
+      target_cluster_name:
+        description: "Target: GKE cluster name"
+        required: true
+      target_default_pool:
+        description: "Target: default pool name"
```
steven-sheehy, line 79: Can this be derived from the other info if omitted?

.github/workflows/copy-stackgres-cluster.yml
```diff
@@ -0,0 +1,161 @@
...
+      SA_EMAIL: ${{ secrets.GH_ACTIONS_KUBECTL_GCP_SERVICE_ACCOUNT }}
+      USE_GKE_GCLOUD_AUTH_PLUGIN: "True"
+      WIF_PROVIDER: ${{ secrets.GH_ACTIONS_KUBECTL_WORKLOAD_ID_PROVIDER }}
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Ensure jq is available
+        run: jq --version || (sudo apt-get update && sudo apt-get install -y jq)
```
steven-sheehy, line 120: I see we use jq in workflows but don't usually install it. Does it not come installed by default?

.github/workflows/copy-stackgres-cluster.yml
```diff
@@ -0,0 +1,161 @@
...
+      K8S_SOURCE_CLUSTER_CONTEXT: "source_gke_context"
+      K8S_TARGET_CLUSTER_CONTEXT: "target_gke_context"
+      PINNED_KUBECONFIG: ${{ github.workspace }}/.kube/config
+      SA_EMAIL: ${{ secrets.GH_ACTIONS_KUBECTL_GCP_SERVICE_ACCOUNT }}
+      USE_GKE_GCLOUD_AUTH_PLUGIN: "True"
+      WIF_PROVIDER: ${{ secrets.GH_ACTIONS_KUBECTL_WORKLOAD_ID_PROVIDER }}
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
```
steven-sheehy: Need to use SHAs for all versions. Can copy from other workflows.

.github/workflows/copy-stackgres-cluster.yml
```diff
@@ -0,0 +1,161 @@
...
+      - name: Setup Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: "1.22"
+
+      - name: Install ksd (Go)
+        run: |
+          go install github.com/mfuentesg/ksd@latest
+          echo "$(go env GOPATH)/bin" >> "$GITHUB_PATH"
+          ksd -h
```
steven-sheehy: We install go just for ksd? If so, we should remove ksd from scripts. It's simple to do what ksd does manually: `kubectl get secret foo -o yaml | yq '.data."values.yaml"' | base64 -d`

.github/workflows/copy-stackgres-cluster.yml
```diff
@@ -0,0 +1,161 @@
...
+
+      - name: Prepare pinned kubeconfig
+        run: |
+          mkdir -p "$(dirname "$PINNED_KUBECONFIG")"
+          : > "$PINNED_KUBECONFIG"
```
steven-sheehy: Does this need to occur here or can it be done later in `Add source context to pinned kube config` step?

.github/workflows/copy-stackgres-cluster.yml
```diff
@@ -0,0 +1,161 @@
...
+        shell: bash
+        env:
+          AUTO_CONFIRM: "true"
+          DEFAULT_POOL_NAME: ${{ inputs.target_default_pool }}
+          WAIT_FOR_K6: ${{ inputs.teardown_target && 'true' || 'false' }}
+        run: |
+          set -Eeuo pipefail
+          echo "Running: script to copy stackgres cluster"
+          cd ./tools/cluster-management/
+          ./copy-live-environment.sh
```
steven-sheehy: Could be simplified. This script itself should already call `set`.

```suggestion
        run: cd ./tools/cluster-management && ./copy-live-environment.sh
```

.github/workflows/copy-stackgres-cluster.yml
```diff
@@ -0,0 +1,161 @@
...
+
+      - name: Cleanup pinned kubeconfig
```
steven-sheehy: Why do we need to clean it up if this env is not reused by anything else?

.github/workflows/copy-stackgres-cluster.yml
jnels124: it is usually the cluster name except in some test clusters

.github/workflows/copy-stackgres-cluster.yml
jnels124: yes it is installed already

.github/workflows/copy-stackgres-cluster.yml
jnels124: yes just to install ksd. I can make this change

.github/workflows/copy-stackgres-cluster.yml
jnels124: it can be done in the add source context step

.github/workflows/copy-stackgres-cluster.yml
jnels124: just cleaning up because it contains access tokens to production clusters (potentially) I can remove this step

.github/workflows/copy-stackgres-cluster.yml
jnels124: just to be clear we will have to change so each individual value is piped to base64 -d

steven-sheehy: Right. Or create a util method if it's easier

xin-hedera (approved):
LGTM

.github/workflows/copy-stackgres-cluster.yml
```diff
@@ -0,0 +1,129 @@
...
+          mv "${PINNED_KUBECONFIG}.merged" "$PINNED_KUBECONFIG"
+          echo "KUBECONFIG=$PINNED_KUBECONFIG" >> "$GITHUB_ENV"
+
+      - name: Execute copy script
+        shell: bash
+        env:
+          AUTO_CONFIRM: "true"
+          DEFAULT_POOL_NAME: ${{ inputs.target_default_pool }}
+          WAIT_FOR_K6: ${{ inputs.teardown_target && 'true' || 'false' }}
+        run: cd ./tools/cluster-management/ && ./copy-live-environment.sh
```
xin-hedera: It doesn't look like the script will sanity check if the target cluster already has everything installed. If so, we should consider adding checks to avoid accidents.

PR #12146, opened 2025-10-06
HIP 1195 Add DB changes for hooks

### LINKED ISSUE #12105: HIP1195 : Add Database Schema Implementation for importer
### Problem

Add Database Schema Implementation for importer

### Solution

1. Adding v1 and v2 sql scripts
2. Adding domain objects
3. Adding repository
4. Update ERD SQL documentation
5. Regenerate ERD diagram

### Alternatives

_No response_

ashumahajan:
**Description**:
- Adding v1 and v2 sql scripts
- Adding domain objects
- Adding repository
- Update ERD SQL documentation

**Related issue(s)**:

Fixes #12105 

**Checklist**

- [x] updated erd.sql to reflect the change

### DIFF
--- a/common/src/main/java/org/hiero/mirror/common/domain/hook/AbstractHook.java
+++ b/common/src/main/java/org/hiero/mirror/common/domain/hook/AbstractHook.java
@@ -4,7 +4,6 @@ package org.hiero.mirror.common.domain.hook;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.google.common.collect.Range;
-import jakarta.persistence.Column;
 import jakarta.persistence.EnumType;
 import jakarta.persistence.Enumerated;
 import jakarta.persistence.IdClass;
@@ -31,20 +30,16 @@ import org.hiero.mirror.common.domain.entity.EntityId;
 public abstract class AbstractHook implements History {
 
     @ToString.Exclude
-    @Column(updatable = false)
     private byte[] adminKey;
 
-    @Column(updatable = false)
     private EntityId contractId;
 
-    @Column(updatable = false)
     private Long createdTimestamp;
 
     private Boolean deleted;
 
     @Enumerated(EnumType.STRING)
     @JdbcTypeCode(SqlTypes.NAMED_ENUM)
-    @Column(updatable = false)
     private HookExtensionPoint extensionPoint;
 
     @jakarta.persistence.Id
@@ -57,7 +52,6 @@ public abstract class AbstractHook implements History {
 
     @Enumerated(EnumType.STRING)
     @JdbcTypeCode(SqlTypes.NAMED_ENUM)
-    @Column(updatable = false)
     private HookType type;
 
     @JsonIgnore


common/src/main/java/org/hiero/mirror/common/domain/hook/HookStorage.java
```diff
@@ -0,0 +1,71 @@
...
+@Builder(toBuilder = true)
+@IdClass(HookStorage.Id.class)
+@NoArgsConstructor
+@Upsertable
+public class HookStorage {
+
+    private static final int KEY_BYTE_LENGTH = 32;
+
+    @Column(updatable = false)
+    private long consensusTimestamp;
```
steven-sheehy: If it's upsertable we should have createdTimestamp and modifiedTimestamp

docs/database/erd.sql
```diff
@@ -326,6 +326,13 @@ alter table transaction_hash add constraint fk_transaction_hash_consensus_timest
 alter table transaction_hash add constraint fk_transaction_hash_payer_account_id foreign key (payer_account_id) references entity (id);
 alter table transaction_signature add constraint fk_transaction_signature_consensus_timestamp foreign key (consensus_timestamp) references transaction (consensus_timestamp);
 alter table transaction_signature add constraint fk_transaction_signature_entity_id foreign key (entity_id) references entity (id);
+alter table hook add constraint fk_hook_contract_id foreign key (contract_id) references entity (id);
+alter table hook add constraint fk_hook_created_timestamp foreign key (created_timestamp) references transaction (consensus_timestamp);
+alter table hook add constraint fk_hook_owner_id foreign key (owner_id) references entity (id);
+alter table hook_storage add constraint fk_hook_storage_consensus_timestamp foreign key (consensus_timestamp) references transaction (consensus_timestamp);
+alter table hook_storage add constraint fk_hook_storage_hook_id_owner_id foreign key (owner_id, hook_id) references hook (owner_id, hook_id);
```
steven-sheehy: We should have separate owner_id -> entity (id) and hook_id -> hook (id). Same for hook_storage_change.

docs/database/erd.sql
```diff
@@ -326,6 +326,13 @@ alter table transaction_hash add constraint fk_transaction_hash_consensus_timest
 alter table transaction_hash add constraint fk_transaction_hash_payer_account_id foreign key (payer_account_id) references entity (id);
 alter table transaction_signature add constraint fk_transaction_signature_consensus_timestamp foreign key (consensus_timestamp) references transaction (consensus_timestamp);
 alter table transaction_signature add constraint fk_transaction_signature_entity_id foreign key (entity_id) references entity (id);
+alter table hook add constraint fk_hook_contract_id foreign key (contract_id) references entity (id);
+alter table hook add constraint fk_hook_created_timestamp foreign key (created_timestamp) references transaction (consensus_timestamp);
+alter table hook add constraint fk_hook_owner_id foreign key (owner_id) references entity (id);
+alter table hook_storage add constraint fk_hook_storage_consensus_timestamp foreign key (consensus_timestamp) references transaction (consensus_timestamp);
```
steven-sheehy: Sort inline

docs/design/hiero-hooks.md
```diff
@@ -173,7 +173,7 @@ Create a new table to store hook information:
 
 ```sql
 -- add_hooks_support.sql
-create type hook_type as enum ('LAMBDA', 'PURE');
+create type hook_type as enum ('LAMBDA');
 create type hook_extension_point as enum ('ACCOUNT_ALLOWANCE_HOOK');
```
steven-sheehy, line 177: We need to also update this file for the changes to domain

importer/src/main/resources/db/migration/v1/V1.113.0__add_hooks_support.sql
```diff
@@ -0,0 +1,20 @@
...
+    created_timestamp bigint                not null,
+    deleted           boolean               not null default false,
+    extension_point   hook_extension_point  not null default 'ACCOUNT_ALLOWANCE_HOOK',
+    hook_id           bigint                not null,
+    owner_id          bigint                not null,
+    type              hook_type             not null default 'LAMBDA',
+
+    primary key (owner_id, hook_id)
+);
+comment on table hook is 'Hooks attached to accounts and contracts';
```
steven-sheehy: We should only have one migration. It should be V1.112.1.

importer/src/main/resources/db/migration/v1/V1.113.0__add_hooks_support.sql
```diff
@@ -0,0 +1,20 @@
...
+create table if not exists hook
+(
+    admin_key         bytea,
+    contract_id       bigint                not null,
+    created_timestamp bigint                not null,
+    deleted           boolean               not null default false,
+    extension_point   hook_extension_point  not null default 'ACCOUNT_ALLOWANCE_HOOK',
+    hook_id           bigint                not null,
+    owner_id          bigint                not null,
+    type              hook_type             not null default 'LAMBDA',
```
steven-sheehy: Please use the creation order from the design doc to have efficient bit packing. Same for other tables.

importer/src/main/resources/db/migration/v2/V2.18.0__add_hooks_support.sql
```diff
@@ -0,0 +1,22 @@
+-- HIP-1195 Mirror Node Hooks Support
+-- Create hook type enums and hooks table
+
+create type hook_type as enum ('LAMBDA');
+create type hook_extension_point as enum ('ACCOUNT_ALLOWANCE_HOOK');
+
+create table if not exists hook
```
steven-sheehy: Combine three migrations into one. Should be V2.17.1

importer/src/main/resources/db/migration/v2/V2.18.1__add_hook_storage_change_table.sql
```diff
@@ -0,0 +1,17 @@
...
+    key                 bytea  not null,
+    owner_id            bigint not null,
+    value_read          bytea  not null,
+    value_written       bytea,
+
+    primary key (owner_id, hook_id, key, consensus_timestamp)
+);
+comment on table hook_storage_change is 'Historical changes to hook storage state';
+
+select create_distributed_table('hook_storage_change', 'owner_id', colocate_with => 'entity');
```
steven-sheehy: We will also need to partition this table via create_time_partitions

common/src/main/java/org/hiero/mirror/common/domain/hook/HookStorage.java
ashumahajan: won't modified timestamp be same as consensus timestamp?

common/src/main/java/org/hiero/mirror/common/domain/hook/Hook.java
```diff
@@ -0,0 +1,80 @@
...
+@NoArgsConstructor
+public class Hook implements Persistable<Hook.Id> {
+
+    @ToString.Exclude
+    private byte[] adminKey;
+
+    private EntityId contractId;
+
+    @Column(updatable = false)
+    private long createdTimestamp;
```
xin-hedera: In case a mirrornode missed the hook creation transaction, we should leave `createdTimestamp` to be `null`.
```suggestion
    private Long createdTimestamp;
```

common/src/main/java/org/hiero/mirror/common/domain/hook/Hook.java
```diff
@@ -0,0 +1,80 @@
...
+    public void setOwnerId(EntityId ownerId) {
+        this.ownerId = ownerId.getId();
+    }
+
+    @AllArgsConstructor
+    @Data
+    @NoArgsConstructor
+    public static class Id implements Serializable {
+        @Serial
+        private static final long serialVersionUID = 1L;
```
xin-hedera: use IDEA generated random `serialVersionUID`

common/src/main/java/org/hiero/mirror/common/domain/hook/Hook.java
```diff
@@ -0,0 +1,80 @@
...
+    @Enumerated
+    private HookType type;
+
+    @Override
+    @JsonIgnore
+    public Id getId() {
+        Id id = new Id();
+        id.setHookId(hookId);
+        id.setOwnerId(ownerId);
+        return id;
```
xin-hedera: nit, `Id` has an all args constructor, so can be simplified to `return new Id(hookId, ownerId);`

common/src/main/java/org/hiero/mirror/common/domain/hook/HookExtensionPoint.java
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package org.hiero.mirror.common.domain.hook;
+
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+
+@Getter
+@RequiredArgsConstructor
```
xin-hedera: Don't think we need the lombok annotations

```suggestion
```

importer/src/main/java/org/hiero/mirror/importer/repository/HookRepository.java
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package org.hiero.mirror.importer.repository;
+
+import org.hiero.mirror.common.domain.hook.Hook;
+import org.springframework.data.repository.CrudRepository;
+import org.springframework.transaction.annotation.Transactional;
+
+@Transactional
```
xin-hedera: not needed

```suggestion
```

importer/src/main/java/org/hiero/mirror/importer/repository/HookStorageRepository.java
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package org.hiero.mirror.importer.repository;
+
+import org.hiero.mirror.common.domain.hook.HookStorage;
+import org.springframework.data.repository.CrudRepository;
+import org.springframework.transaction.annotation.Transactional;
+
+@Transactional
```
xin-hedera:
```suggestion
```

common/src/main/java/org/hiero/mirror/common/domain/hook/HookStorage.java
steven-sheehy: This is an upsertable table i.e. any particular row can be modified by a new timestamp that we should track.

ashumahajan: Primary key for the table is owner_id, hook_id  and key. In this case it will store the current state. The reason why I marked in upsertable is to ensure that any change in the storage is reflected on the table. consensus_timestamp is tied to the transaction that resulted in storage update for the key. I agree that we track created_timestamp which will be the first instance of the entry for the key. consensus_timestamp will always tracks the last updated timestamp. Do you think we still need modified timestamp?

common/src/main/java/org/hiero/mirror/common/domain/hook/Hook.java
ashumahajan: There is no hook creation transaction. Hook is an existing contract, which will be added to the owner entity, with create or update transactions. Incase we do not receive the hookId, we will not have an entry for this

xin-hedera: correct. What I meant is the transactions which creates the hook. Since a hook can be deleted, and importer needs to update the row in database. The requested change still applies in case a mirronode missed the hook creation transaction.

In addition, we should also have a `timestamp_range` column as well as a history table for hook.

ashumahajan: Hook is immutable and thus Steven was not interested in keeping timestamp_range. 
When deleted, the entry will be removed from db. My understanding is that the way to remove the hook for an owner is sending entity update transaction, without HookCreationDetails. There is no specific transaction that supports hook removal.

xin-hedera: I'm confused by the part `When deleted, the entry will be removed from db`

- why do we then have a `deleted` column?
- we've never deleted rows, and we should not do it for both technical limit and data audit concern

xin-hedera: the following transactions can be used to delete hooks per the HIP

```

message CryptoUpdateTransactionBody {
  // ...

  /**
   * The hooks to create for the account.
   */
  repeated HookCreationDetails hook_creation_details = 19;

  /**
   * The ids the hooks to delete from the account.
   */
  repeated int64 hook_ids_to_delete = 20;
}

message ContractUpdateTransactionBody {
  // ...
  
  /**
   * The hooks to create for the contract.
   */
  repeated HookCreationDetails hook_creation_details = 16;

  /**
   * The ids the hooks to delete from the contract.
   */
  repeated int64 hook_ids_to_delete = 20;
}
```

ashumahajan: Sorry for the confusion. My mistake! deleted from the db means that it will be marked deleted

xin-hedera: fine with the no `timestamp_range` column decision decision.

Still, the original feedback stands, that because mirrirnode (non full history specifically) can miss the transactions which creates a hook, and later when the hook is deleted, we should not insert a row to the table with `consensus_timestamp = 0`.

ashumahajan: Good point about hook_ids_to_delete. I got confused as per one of the reponses in slack channel about delete, it was for storage and not the hook

steven-sheehy: Thinking further, we still need to track when the hook was marked as deleted. So we either need a `modified_timestamp` or a `timestamp_range`. Perhaps we should just go with history table plus timestamp_range to keep it similar to other tables?

common/src/main/java/org/hiero/mirror/common/domain/hook/HookStorage.java
steven-sheehy: Standard is if we only have one timestamp we use consensus_timestamp. But if we have created_timestamp and want to track updates we should have modified_timestamp. Using created_timestamp and consensus_timestamp would make the latter column's purpose ambiguous.

Regardless, this table is basically the same as `ContractState` so we should have similar fields and logic.

importer/src/main/resources/db/migration/v1/V1.113.1__add_hooks_support.sql
```diff
@@ -0,0 +1,52 @@
...
+    owner_id            bigint not null,
+    key                 bytea  not null,
+    value_read          bytea  not null,
+    value_written       bytea,
+
+    primary key (owner_id, hook_id, key, consensus_timestamp)
+);
+comment on table hook_storage_change is 'Historical changes to hook storage state';
+
+select create_time_partitions('hook_storage_change', 'consensus_timestamp', true);
```
steven-sheehy: This is for v2 only

common/src/main/java/org/hiero/mirror/common/domain/hook/HookStorage.java
ashumahajan: Made changes as per ContractState

importer/src/main/resources/db/migration/v1/V1.113.1__add_hooks_support.sql
ashumahajan: Figured out when tried running locally :)

common/src/main/java/org/hiero/mirror/common/domain/hook/Hook.java
ashumahajan: @steven-sheehy do we in that case need deleted column at all, or it should be tracked by timestamp_range. 
Pros to the approach is that we are following already established standards. But I feel we are duplicating data to just track deleted column, I like the idea of adding modified_timestamp (or deleted_timestamp) to track, when the hook was deleted

steven-sheehy: Still need deleted column. We can just go with modified_timestamp and no history table for now.

importer/src/main/java/org/hiero/mirror/importer/repository/HookStorageChangeRepository.java
steven-sheehy, line 1: Each repository should have a test that does a basic save and find and verifies the result is the same. They should use the domain builder so this PR is missing DomainBuilder changes.

docs/design/hiero-hooks.md
```diff
@@ -268,8 +270,10 @@ public class Hook {
 ```java
 // common/src/main/java/org/hiero/mirror/common/domain/entity/HookStorage.java
 @IdClass(HookStorage.Id.class)
+@Upsertable
 public class HookStorage {
 
+    private long consensusTimestamp;
```
steven-sheehy: Should be modifiedTimestamp

importer/src/main/resources/db/migration/v1/V1.113.0__add_hooks_support.sql
steven-sheehy: It's still not V1.112.1

importer/src/main/resources/db/migration/v2/V2.18.0__add_hooks_support.sql
steven-sheehy: It's still not V2.17.1

docs/database/erd.sql
steven-sheehy: There are still combined `(owner_id, hook_id)` foreign keys for multiple tables instead of separate for each.

common/src/main/java/org/hiero/mirror/common/domain/hook/HookStorage.java
```diff
@@ -0,0 +1,73 @@
...
+@NoArgsConstructor
+@Upsertable
+public class HookStorage {
+
+    private static final int KEY_BYTE_LENGTH = 32;
+
+    @Column(updatable = false)
+    private long createdTimestamp;
+
+    private long modifiedTimestamp;
```
steven-sheehy: nit: Not sorted

importer/src/main/resources/db/migration/v1/V1.113.1__add_hooks_support.sql
```diff
@@ -0,0 +1,52 @@
...
+-- Create enums for hook types and extension points
+create type hook_type as enum ('LAMBDA');
+create type hook_extension_point as enum ('ACCOUNT_ALLOWANCE_HOOK');
+
+-- Main hook table
+create table if not exists hook
+(
+    contract_id         bigint                not null,
+    created_timestamp   bigint,
+    modified_timestamp  bigint,
```
steven-sheehy: Sort below hook_id. Same for v2

importer/src/main/resources/db/migration/v1/V1.113.1__add_hooks_support.sql
```diff
@@ -0,0 +1,52 @@
...
+(
+    consensus_timestamp bigint not null,
+    hook_id             bigint not null,
+    owner_id            bigint not null,
+    key                 bytea  not null,
+    value_read          bytea  not null,
+    value_written       bytea,
+
+    primary key (owner_id, hook_id, key, consensus_timestamp)
+) partition by range (consensus_timestamp);
```
steven-sheehy: Remove partitioning in V1

importer/src/main/resources/db/migration/v1/V1.113.1__add_hooks_support.sql
```diff
@@ -0,0 +1,52 @@
...
+comment on table hook_storage_change is 'Historical changes to hook storage state';
+
+
+-- Hook storage table (current state)
+create table if not exists hook_storage
+(
+    created_timestamp   bigint not null,
+    modified_timestamp  bigint not null,
```
steven-sheehy: Sort below hook_id. Same for v2

docs/database/erd.sql
ashumahajan: With just hook_id -> hook (id), we get error running the script ERROR: there is no unique constraint matching given keys for referenced table "hook"

importer/src/test/java/org/hiero/mirror/importer/repository/HookStorageChangeRepositoryTest.java
```diff
@@ -0,0 +1,34 @@
...
+@RequiredArgsConstructor
+class HookStorageChangeRepositoryTest extends ImporterIntegrationTest {
+
+    private final HookStorageChangeRepository hookStorageChangeRepository;
+
+    @Test
+    void prune() {
+        domainBuilder.hookStorageChange().persist();
+        var hookStorageChange2 = domainBuilder.hookStorageChange().persist();
+        var hookStorageChange3 = domainBuilder.hookStorageChange().persist();
```
steven-sheehy, line 20: Use `repository.save(domainBuilder.hook().get())` to test the actual repository. persist uses entitymanager.

importer/src/main/resources/db/migration/v1/V1.113.0__add_hooks_support.sql
ashumahajan: We already have V1.112.0__node_account_id that is why i chose V1.113.1__add_hooks_support
same thing for V2.17.0__node_account_id which is already there and thus i used V2.18.1__add_hooks_support

May be I don't understand the naming convention.

steven-sheehy: We only bump minor version of migration for new mirror node version. V1.112.0__node_account_id is in the current release 0.141 so we have to keep this hook migration in V1.112.x.

docs/design/hiero-hooks.md
```diff
@@ -173,21 +173,54 @@ Create a new table to store hook information:
...
-    type              hook_type             not null default 'LAMBDA',
-    deleted           boolean               not null default false,
-    admin_key         bytea,
-
-    primary key (owner_id, hook_id)
+    deleted
+    boolean
+    not
+    null
+    default
```
steven-sheehy: formatting

docs/design/hiero-hooks.md
```diff
@@ -268,12 +302,14 @@ public class Hook {
...
 @IdClass(HookStorage.Id.class)
+@Upsertable
 public class HookStorage {
 
     private long createdTimestamp;
+    private final long modifiedTimestamp;
     private long hookId;
     private byte[] key;
-    private long modifiedTimestamp;
+    private final long modifiedTimestamp;
```
steven-sheehy: There's two modifiedTimestamps now and they're final

importer/src/main/resources/db/migration/v1/V1.112.1__add_hooks_support.sql
```diff
@@ -0,0 +1,52 @@
...
+create table if not exists hook
+(
+    deleted             boolean               not null default false,
+    extension_point     hook_extension_point  not null default 'ACCOUNT_ALLOWANCE_HOOK',
+    type                hook_type             not null default 'LAMBDA',
+    contract_id         bigint                not null,
+    created_timestamp   bigint,
+    hook_id             bigint                not null,
+    modified_timestamp  bigint,
+    owner_id            bigint                not null,
```
steven-sheehy: The bit packing order is again not right. Same for v2. Not sure if some formatter is doing it.

common/src/main/java/org/hiero/mirror/common/domain/hook/Hook.java
```diff
@@ -0,0 +1,84 @@
...
+@IdClass(Hook.Id.class)
+@NoArgsConstructor
+public class Hook implements Persistable<Hook.Id> {
+
+    @ToString.Exclude
+    private byte[] adminKey;
+
+    private EntityId contractId;
+
+    @Column(updatable = false)
```
steven-sheehy: I guess technically all columns are not updatable except modifiedTimestamp and deleted. So we should add this annotation to others except id columns (implicit), modifiedTimestamp, and deleted.

common/src/main/java/org/hiero/mirror/common/domain/hook/Hook.java
```diff
@@ -0,0 +1,84 @@
...
+    @JsonIgnore
+    public Id getId() {
+        return new Id(hookId, ownerId);
+    }
+
+    @JsonIgnore
+    @Override
+    public boolean isNew() {
+        return true;
+    }
```
steven-sheehy: We should remove Persistable and isNew since it's upsertable.

common/src/test/java/org/hiero/mirror/common/domain/DomainBuilder.java
```diff
@@ -575,6 +580,44 @@ public DomainWrapper<FileData, FileData.FileDataBuilder> fileData() {
...
+        var builder = Hook.builder()
+                .hookId(number())
+                .ownerId(id())
+                .extensionPoint(HookExtensionPoint.ACCOUNT_ALLOWANCE_HOOK)
+                .type(HookType.LAMBDA)
+                .createdTimestamp(createdTimestamp)
+                .modifiedTimestamp(createdTimestamp)
+                .deleted(false)
+                .adminKey(key())
+                .contractId(entityId());
```
steven-sheehy: nit: sort this and others

common/src/test/java/org/hiero/mirror/common/domain/DomainBuilder.java
```diff
@@ -575,6 +580,44 @@ public DomainWrapper<FileData, FileData.FileDataBuilder> fileData() {
...
+
+    public DomainWrapper<HookStorage, HookStorage.HookStorageBuilder> hookStorage() {
+        var createdTimestamp = timestamp();
+        var builder = HookStorage.builder()
+                .hookId(number())
+                .ownerId(id())
+                .key(bytes(32))
+                .value(bytes(32))
+                .createdTimestamp(createdTimestamp)
+                .modifiedTimestamp(createdTimestamp);
```
steven-sheehy: Prefer create and modified timestamp with different values by default for storage.

docs/database/erd.sql
steven-sheehy: What did we figure out here?

ashumahajan: I am not sure since it was working with owner_id. I checked online and it seems, the reason the foreign key constraint works with reference just owner_id is because entity table makes sure that owner_id is a unique primary key or has a unique constraint.

common/src/main/java/org/hiero/mirror/common/domain/hook/Hook.java
```diff
@@ -18,20 +18,22 @@
...
 @Data
 @Entity
 @IdClass(Hook.Id.class)
 @NoArgsConstructor
-public class Hook implements Persistable<Hook.Id> {
+@Upsertable
+public class Hook {
 
     @ToString.Exclude
     private byte[] adminKey;
```
steven-sheehy: Should have updatable=false

importer/src/main/resources/db/migration/v1/V1.112.1__add_hooks_support.sql
```diff
@@ -8,14 +8,14 @@ create type hook_extension_point as enum ('ACCOUNT_ALLOWANCE_HOOK');
 -- Main hook table
 create table if not exists hook
 (
-    deleted             boolean               not null default false,
-    extension_point     hook_extension_point  not null default 'ACCOUNT_ALLOWANCE_HOOK',
-    type                hook_type             not null default 'LAMBDA',
     contract_id         bigint                not null,
     created_timestamp   bigint,
     hook_id             bigint                not null,
     modified_timestamp  bigint,
```
steven-sheehy: modified_timestamp should be not null. It's only created_timestamp that can be nullable. Need to also fix v2 and design.

common/src/main/java/org/hiero/mirror/common/domain/hook/Hook.java
```diff
@@ -0,0 +1,81 @@
...
+
+    @Enumerated(EnumType.STRING)
+    @JdbcTypeCode(SqlTypes.NAMED_ENUM)
+    @Column(updatable = false)
+    private HookExtensionPoint extensionPoint;
+
+    @jakarta.persistence.Id
+    private long hookId;
+
+    private Long modifiedTimestamp;
```
steven-sheehy: This can be primitive

steven-sheehy:
Test failures

importer/src/main/resources/db/migration/v2/V2.17.1__add_hooks_support.sql
```diff
@@ -0,0 +1,58 @@
...
+    key                 bytea  not null,
+    value_read          bytea  not null,
+    value_written       bytea,
+
+    primary key (owner_id, hook_id, key, consensus_timestamp)
+) partition by range (consensus_timestamp);
+comment on table hook_storage_change is 'Historical changes to hook storage state';
+
+select create_distributed_table('hook_storage_change', 'owner_id', colocate_with => 'entity');
+select create_time_partitions('hook_storage_change', 'consensus_timestamp', true);
```
steven-sheehy: This doesn't match at all the signature for `create_time_partions`. See `V2.0.2__static_partitioning.sql`. Please make sure you're running importer tests with v2 profile locally before pushing and requesting a review when you're making schema changes.

ashumahajan: tested against v2. Testcases are now passing

IvanKavaldzhiev (dismissed):
LGTM

common/src/main/java/org/hiero/mirror/common/domain/hook/AbstractHook.java
```diff
@@ -0,0 +1,82 @@
...
+@Data
+@MappedSuperclass
+@NoArgsConstructor
+@SuperBuilder(toBuilder = true)
+@IdClass(AbstractHook.Id.class)
+@Upsertable(history = true)
+public abstract class AbstractHook implements History {
+
+    @ToString.Exclude
+    @Column(updatable = false)
```
xin-hedera:
```suggestion
```

common/src/main/java/org/hiero/mirror/common/domain/hook/AbstractHook.java
```diff
@@ -0,0 +1,82 @@
...
+    private byte[] adminKey;
+
+    @Column(updatable = false)
```
xin-hedera:
```suggestion
```

common/src/main/java/org/hiero/mirror/common/domain/hook/AbstractHook.java
```diff
@@ -0,0 +1,82 @@
...
+    private EntityId contractId;
+
+    @Column(updatable = false)
```
xin-hedera:
```suggestion
```

common/src/main/java/org/hiero/mirror/common/domain/hook/AbstractHook.java
```diff
@@ -0,0 +1,82 @@
...
+    private Long createdTimestamp;
+
+    private Boolean deleted;
+
+    @Enumerated(EnumType.STRING)
+    @JdbcTypeCode(SqlTypes.NAMED_ENUM)
+    @Column(updatable = false)
```
xin-hedera:
```suggestion
```

common/src/main/java/org/hiero/mirror/common/domain/hook/AbstractHook.java
```diff
@@ -0,0 +1,82 @@
...
+    private long hookId;
+
+    @jakarta.persistence.Id
+    private long ownerId;
+
+    private Range<Long> timestampRange;
+
+    @Enumerated(EnumType.STRING)
+    @JdbcTypeCode(SqlTypes.NAMED_ENUM)
+    @Column(updatable = false)
```
xin-hedera:
```suggestion
```

xin-hedera (changes requested):

PR #12147, opened 2025-10-07
HIP-1299 Replace account_id if present in node.account_id in gRPC API

### LINKED ISSUE #12095: HIP-1299 Update getNodes gRPC API
### Problem

As a node operator, I'd like to be able to update my node account ID and have that change propagate to mirror nodes and users of the SDKs.

### Solution

* Update gRPC API `getNetworkNodes` APIs to return `node.accountId` if not null, otherwise use `addressBookEntry.nodeAccountId`

### Alternatives

_No response_

martingeorgiev1:
**Description**:
 Update the getNodes gRPC API implementation by replacing the `account_id `when present in the `node entity`, use cache, and add tests.

* Add config property
* Add unit tests
* Add caching

**Related issue(s)**:

Fixes #12095

**Notes for reviewer**:

**Checklist**
- [X] Documented (Code comments, README)
- [X] Confirm that the getNodes API correctly prioritizes node.account_id when available
- [X] Verify caching layer behavior and cache invalidation logic
- [X] All new tests are passing locally (spotlessApply, gradlew build, etc.)

### DIFF
--- a/grpc/src/test/java/org/hiero/mirror/grpc/service/NetworkServiceTest.java
+++ b/grpc/src/test/java/org/hiero/mirror/grpc/service/NetworkServiceTest.java
@@ -246,8 +246,8 @@ class NetworkServiceTest extends GrpcIntegrationTest {
         var addressBook = addressBook();
         var addressBookEntry1 = addressBookEntry();
         var addressBookEntry2 = addressBookEntry();
-        var node1AccountId = EntityId.of(0L, 0L, 1000L);
-        var node2AccountId = EntityId.of(0L, 0L, 1001L);
+        var node1AccountId = domainBuilder.entityId();
+        var node2AccountId = domainBuilder.entityId();
         createNode(addressBookEntry1.getNodeId(), node1AccountId);
         createNode(addressBookEntry2.getNodeId(), node2AccountId);
         var filter = AddressBookFilter.builder().fileId(addressBook.getFileId()).build();
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -378,7 +378,7 @@ value, it is recommended to only populate overridden properties in the custom `a
 
 | Name                                                       | Default          | Description                                                                                               |
 | ---------------------------------------------------------- | ---------------- | --------------------------------------------------------------------------------------------------------- |
-| `hiero.mirror.grpc.addressbook.cacheExpiry`                | 5s               | The amount of time to cache address book entries                                                          |
+| `hiero.mirror.grpc.addressbook.cacheExpiry`                | 2s               | The amount of time to cache address book entries                                                          |
 | `hiero.mirror.grpc.addressbook.cacheSize`                  | 50               | The maximum number of address book pages to cache                                                         |
 | `hiero.mirror.grpc.addressbook.maxPageDelay`               | 250ms            | The maximum amount of time to sleep between paging for address book entries                               |
 | `hiero.mirror.grpc.addressbook.minPageDelay`               | 100ms            | The minimum amount of time to sleep between paging for address book entries                               |


grpc/src/main/java/org/hiero/mirror/grpc/service/NetworkServiceImpl.java
```diff
@@ -87,7 +89,14 @@ private Flux<AddressBookEntry> page(AddressBookContext context) {
                     addressBookTimestamp, nextNodeId, pageSize);
             var endpoints = new AtomicInteger(0);
 
+            Map<Long, EntityId> accountIdMap = nodeRepository.findNodesWithAccountIdsMap();
+
             nodes.forEach(node -> {
+                EntityId accountId = accountIdMap.get(node.getNodeId());
+                if (accountId != null && !EntityId.isEmpty(accountId)) {
```
IvanKavaldzhiev: I think this duplicates the logic inside NodeRepository. No null or empty accountIds should be present.

martingeorgiev1: True, but we are iterating the nodes from the addressBookEntry, and then checking if they are present in the `accountIdMap`, so they can be null. It is true the `isEmpty ` check is redundant here, so I will remove it.

grpc/src/main/java/org/hiero/mirror/grpc/repository/NodeRepository.java
```diff
@@ -0,0 +1,35 @@
...
+public interface NodeRepository extends CrudRepository<Node, Long> {
+
+    @Query(value = "select * from node where deleted is false and account_id is not null", nativeQuery = true)
+    Iterable<Node> findNodesWithAccountIds();
+
+    @Cacheable(
+            cacheManager = NODE_ACCOUNT_CACHE,
+            cacheNames = CACHE_NAME,
+            unless = "#result == null or #result.isEmpty()")
+    default Map<Long, EntityId> findNodesWithAccountIdsMap() {
```
bilyana-gospodinova: I think the `@Cacheable` annotation can be moved to `findNodesWithAccountIds()` and this method can be removed as `findNodesWithAccountIds()` already returns only non-null account ids.

grpc/src/main/java/org/hiero/mirror/grpc/service/NetworkServiceImpl.java
```diff
@@ -87,7 +89,14 @@ private Flux<AddressBookEntry> page(AddressBookContext context) {
                     addressBookTimestamp, nextNodeId, pageSize);
             var endpoints = new AtomicInteger(0);
 
+            Map<Long, EntityId> accountIdMap = nodeRepository.findNodesWithAccountIdsMap();
```
bilyana-gospodinova: nit: `Map<Long, EntityId>` can be a var to match the rest of the code.

grpc/src/main/java/org/hiero/mirror/grpc/service/AddressBookProperties.java
```diff
@@ -29,6 +29,11 @@ public class AddressBookProperties {
     @Min(0)
     private long nodeStakeCacheSize = 5L;
 
+    private Duration nodeAccountCacheExpiry = Duration.ofSeconds(2);
```
bilyana-gospodinova: I think there should be a `@NotNull` annotation here as the other similar properties.

grpc/src/main/java/org/hiero/mirror/grpc/service/AddressBookProperties.java
```diff
@@ -29,6 +29,11 @@ public class AddressBookProperties {
...
+
+    @Min(0)
+    private long nodeAccountCacheSize = 5L;
```
bilyana-gospodinova: I think the value should be bigger - at least `addressBookProperties.getPageSize()` (which by default is 10) as otherwise to return a list of (page size) the node account cache will be overwritten twice and it will not be effective.

grpc/src/main/java/org/hiero/mirror/grpc/repository/NodeRepository.java
martingeorgiev1: I can remove the checks in `findNodesWithAccountIdsMap`  and also make it in foreach so it can be inline.
I kinda want to cache only a map with the nodeId and accountId, instead of list of  node entities and then to iterate them on each addressBookEntity, to find the corresponding node. I can move the forEach and map populating inside the service but caching the list of entities will take more space and will be slower for lookup.

grpc/src/main/java/org/hiero/mirror/grpc/service/AddressBookProperties.java
IvanKavaldzhiev: Currently we have around 30ish consensus validators, so I suggest to set the property to 30 or 35.

grpc/src/main/java/org/hiero/mirror/grpc/repository/AddressBookEntryRepository.java
```diff
@@ -18,8 +18,27 @@ public interface AddressBookEntryRepository extends CrudRepository<AddressBookEn
...
+                    """
+        select abe.consensus_timestamp,
+               abe.description,
+               abe.memo,
+               abe.node_id,
+               abe.node_cert_hash,
+               abe.public_key,
+               abe.stake,
+               case
+                   when n.account_id is not null and (n.deleted is null or n.deleted = false)
```
steven-sheehy: We don't need to check deleted status. Can remove and just do a simple `coalesce(n.account_id, abe.node_account_id)`

grpc/src/main/java/org/hiero/mirror/grpc/repository/AddressBookEntryRepository.java
```diff
@@ -18,8 +18,27 @@ public interface AddressBookEntryRepository extends CrudRepository<AddressBookEn
             cacheNames = CACHE_NAME,
```
steven-sheehy, line 18: The cache expiry is 60s. This means at worst case it could take up to 119s for a client to be made aware of a node account id change. We should probably adjust this down to 2s for now.

grpc/src/main/java/org/hiero/mirror/grpc/repository/AddressBookEntryRepository.java
steven-sheehy, line 1: This doesn't solve the problem we discussed on standup where it doesn't cache the calls to get latest stake and address book timestamps. Let's keep this ticket as is but open another ticket to rework things in a future sprint to change network service to only use a single paginated query that queries all tables that we cache for 2s.

grpc/src/main/java/org/hiero/mirror/grpc/service/AddressBookProperties.java
```diff
@@ -17,7 +17,7 @@ public class AddressBookProperties {
 
     @DurationMin(millis = 500L)
     @NotNull
-    private Duration cacheExpiry = Duration.ofSeconds(60);
+    private Duration cacheExpiry = Duration.ofSeconds(2);
```
steven-sheehy, line 20: Need to also update docs/configuration.md

grpc/src/test/java/org/hiero/mirror/grpc/service/NetworkServiceTest.java
```diff
@@ -240,4 +240,53 @@ private void nodeStake(long nodeId, long stake) {
...
                 .persist();
     }
+
+    @Test
+    void overrideNodeAccountIdWhenPresent() {
+        var addressBook = addressBook();
+        var addressBookEntry1 = addressBookEntry();
+        var addressBookEntry2 = addressBookEntry();
+        var node1AccountId = EntityId.of(0L, 0L, 1000L);
+        var node2AccountId = EntityId.of(0L, 0L, 1001L);
```
steven-sheehy: This will fail with non-zero realm test in CI. Need to use `domainBuilder.entityId()`

grpc/src/main/java/org/hiero/mirror/grpc/repository/AddressBookEntryRepository.java
steven-sheehy: Don't forget to open the ticket.

steven-sheehy: #12234

PR #12151, opened 2025-10-08
Adapt k6 tests to run in CI

bilyana-gospodinova:
**Description**:
TODO

Fixes #


test/src/test/java/org/hiero/mirror/test/e2e/acceptance/client/AccountClient.java
```diff
@@ -61,6 +61,16 @@ public AccountClient(SDKClient sdkClient, RetryTemplate retryTemplate) {
 
     @Override
     public void clean() {
+        if (Boolean.parseBoolean(System.getenv("HIERO_MIRROR_TEST_ACCEPTANCE_SKIP_CLEANUP_ENTITIES"))) {
```
steven-sheehy: This should be an actual documented property in `AcceptanceTestProperties` so it can be set via any of our configuration formats. Probably invert and simplify it so it defaults to true `hiero.mirror.test.acceptance.cleanup=true`.

.github/workflows/acceptance.yaml
```diff
@@ -123,6 +123,7 @@ jobs:
               HIERO_MIRROR_TEST_ACCEPTANCE_FEATURE_CONTRACTCALLLOCALESTIMATE: "true"
               HIERO_MIRROR_TEST_ACCEPTANCE_NETWORK: OTHER
               HIERO_MIRROR_TEST_ACCEPTANCE_WEB3_OPCODETRACER_ENABLED: "true"
+              HIERO_MIRROR_TEST_ACCEPTANCE_SKIP_CLEANUP_ENTITIES: "true"
```
steven-sheehy: Probably we don't need this if we're not running k6 for block?

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/client/ContractClient.java
```diff
@@ -28,21 +28,35 @@
...
     public ContractClient(SDKClient sdkClient, RetryTemplate retryTemplate) {
         super(sdkClient, retryTemplate);
     }
 
     @Override
     public void clean() {
+        if (Boolean.parseBoolean(System.getenv("HIERO_MIRROR_TEST_ACCEPTANCE_SKIP_CLEANUP_ENTITIES"))) {
+            // In CI we don't want to cleanup as the entities are needed in the k6 test in the next step.
+            log.warn("Acceptance tests running in CI -> skip cleanup.");
+            for (var contractName : contractIds.keySet()) {
```
steven-sheehy: We should add a deleteAll in parent that accepts a map and does this logic centrally. We don't have to convert everything to use the map deleteAll at this time.

.github/workflows/acceptance.yaml
```diff
@@ -152,7 +154,149 @@ jobs:
...
+        if: ${{ matrix.stream-type != 'BLOCK' }}
+        uses: grafana/setup-k6-action@v1
+
+      - name: Get web3 base url
+        if: ${{ matrix.stream-type != 'BLOCK' }}
+        id: web3-base-url
+        run: |
+          export IP=$(kubectl get pods -n "${SOLO_NAMESPACE}" -l app.kubernetes.io/name=web3 -o custom-columns='IP:.status.podIP' --no-headers)
+          BASE_URL="http://${WEB3_IP:-127.0.0.1}:8080"
+          echo "BASE_URL=$BASE_URL" >> $GITHUB_OUTPUT
```
steven-sheehy: Since you're always port-forwarding to localhost:8080 you shouldn't need this step. Just hardcode BASE_URL in `Run k6 web3 tests` step.

.github/workflows/acceptance.yaml
```diff
@@ -152,7 +154,149 @@ jobs:
...
+            ./tools/k6/src/web3/test/contractCallSymbol.js
+            ./tools/k6/src/web3/test/contractCallTokenCustomFees.js
+            ./tools/k6/src/web3/test/contractCallTokenDefaultFreezeStatus.js
+            ./tools/k6/src/web3/test/contractCallTokenDefaultKycStatus.js
+            ./tools/k6/src/web3/test/contractCallTokenExpiryInfo.js
+            ./tools/k6/src/web3/test/contractCallTokenInfo.js
+            ./tools/k6/src/web3/test/contractCallTokenKey.js
+            ./tools/k6/src/web3/test/contractCallTokenType.js
+            ./tools/k6/src/web3/test/contractCallTokenURI.js
+            ./tools/k6/src/web3/test/contractCallTotalSupply.js
```
steven-sheehy: Running separately doesn't test the setup and wrapper code and doesn't produce a report. We should use the whitelist variable to run this subset.

.github/workflows/acceptance.yaml
```diff
@@ -152,7 +154,149 @@ jobs:
...
+              sed -E 's/.*at address ([0-9a-fA-F]{40})/\1/' | \
+              tail -n 1 | \
+              tr -d '[:space:]')
+
+          echo "ACCOUNT_ADDRESS=$ACCOUNT_ADDRESS"
+          echo "ACCOUNT_ADDRESS=$ACCOUNT_ADDRESS" >> $GITHUB_OUTPUT
+
+      - name: Install k6
+        if: ${{ matrix.stream-type != 'BLOCK' }}
+        uses: grafana/setup-k6-action@v1
```
steven-sheehy: Need to use sha hashes instead of versions for all action versions for security reasons.

.github/workflows/acceptance.yaml
```diff
@@ -152,7 +154,149 @@ jobs:
...
+              tail -n 1 | \
+              tr -d '[:space:]')
+
+          echo "DEFAULT_CONTRACT_ADDRESS=$DEFAULT_CONTRACT_ADDRESS"
+          echo "DEFAULT_CONTRACT_ADDRESS=$DEFAULT_CONTRACT_ADDRESS" >> $GITHUB_OUTPUT
+
+          ESTIMATE_PRECOMPILE_CONTRACT=$(grep "Skipping cleanup of contract \[ESTIMATE_PRECOMPILE\]" "$LOG_FILE" | \
+              sed -E 's/.*at address ([0-9a-fA-F]{40})/\1/' | \
+              tail -n 1 | \
+              tr -d '[:space:]')
```
steven-sheehy: It's decoupled now but very verbose and brittle. Should use a bash associative array to maintain the mapping then all the logic is generic. Instead of outputting to GITHUB_OUTPUT output the key=value to GITHUB_ENV so it's already available in the next k6 step.

```bash
declare -A k6Map
k6Map["PRECOMPILE"]="PRECOMPILE_CONTRACT"
...

for key in "${!k6Map[@]}"; do
  k6="${k6Map[$key]}"
  value=$(# find acceptance value in log)
  echo "${k6}=${value}" >> $GITHUB_ENV
done
```

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/client/TokenClient.java
```diff
@@ -71,6 +71,16 @@ public TokenClient(SDKClient sdkClient, RetryTemplate retryTemplate) {
 
     @Override
     public void clean() {
+        if (Boolean.parseBoolean(System.getenv("HIERO_MIRROR_TEST_ACCEPTANCE_SKIP_CLEANUP_ENTITIES"))) {
+            // In CI we don't want to cleanup as the entities are needed in the k6 test in the next step.
+            log.warn("Acceptance tests running in CI -> skip cleanup.");
+            for (var tokenName : tokenMap.keySet()) {
+                log.info("Skipping cleanup of token [" + tokenName.getSymbol() + "] at address "
+                        + tokenMap.get(tokenName).tokenId().toEvmAddress());
```
steven-sheehy: Should log this in a form that's easier to parse. Could probably skip logger and use `System.out.println(key+"="+value);`

.github/workflows/acceptance.yaml
bilyana-gospodinova: Correct, removed.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/client/AccountClient.java
bilyana-gospodinova: I defined `HIERO_MIRROR_TEST_ACCEPTANCE_SKIP_CLEANUP_ENTITIES` specifically that way so that it will be less error prone. If we forget to set it in some environment such as testnet, previewnet, etc., this will result in some entities not cleaned up and the remaining balances won't be restored. We need this property only in CI for now.

.github/workflows/acceptance.yaml
```diff
@@ -152,7 +153,74 @@ jobs:
...
+          WEB3_TEST_INCLUDE: "./tools/k6/src/web3/test/contractCallRampUp.js,./tools/k6/src/web3/test/contractCallApproved.js,./tools/k6/src/web3/test/contractCallBalance.js,./tools/k6/src/web3/test/contractCallDecimals.js,./tools/k6/src/web3/test/contractCallFungibleTokenInfo.js,./tools/k6/src/web3/test/contractCallIdentifier.js,./tools/k6/src/web3/test/contractCallIsToken.js,./tools/k6/src/web3/test/contractCallMultiply.js,./tools/k6/src/web3/test/contractCallName.js,./tools/k6/src/web3/test/contractCallNonFungibleTokenInfo.js,./tools/k6/src/web3/test/contractCallOwnerOf.js,./tools/k6/src/web3/test/contractCallSender.js,./tools/k6/src/web3/test/contractCallSymbol.js,./tools/k6/src/web3/test/contractCallTokenCustomFees.js,./tools/k6/src/web3/test/contractCallTokenDefaultFreezeStatus.js,./tools/k6/src/web3/test/contractCallTokenDefaultKycStatus.js,./tools/k6/src/web3/test/contractCallTokenExpiryInfo.js,./tools/k6/src/web3/test/contractCallTokenInfo.js,./tools/k6/src/web3/test/contractCallTokenKey.js,./tools/k6/src/web3/test/contractCallTokenType.js,./tools/k6/src/web3/test/contractCallTokenURI.js,./tools/k6/src/web3/test/contractCallTotalSupply.js"
```
steven-sheehy: It's a regex. So can simplify to `.*/contractCall(RampUp|Approved|...)`.

.github/workflows/acceptance.yaml
```diff
@@ -152,7 +153,74 @@ jobs:
...
+          kubectl port-forward "service/mirror-web3" -n mirror 8080:80 &
+
+      - name: Run k6 web3 tests
+        if: ${{ matrix.stream-type != 'BLOCK' }}
+        uses: grafana/run-k6-action@a15e2072ede004e8d46141e33d7f7dad8ad08d9d # v1
+        env:
+          ACCOUNT_ADDRESS: ${{ steps.log-variables.outputs.ACCOUNT_ADDRESS }}
+          AMOUNT: "0000000000000000000000000000000000000000000000000000000000000001"
+          BASE_URL: "http://localhost:8080"
+          DEFAULT_CONTRACT_ADDRESS: ${{ steps.log-variables.outputs.DEFAULT_CONTRACT_ADDRESS }}
```
steven-sheehy: As mentioned, output to [GITHUB_ENV](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-variables#passing-values-between-steps-and-jobs-in-a-workflow) in the previous step. Then you don't need these `steps.log-variables.outputs` lines here.

.github/workflows/acceptance.yaml
```diff
@@ -152,7 +153,74 @@ jobs:
...
+          k6Map["BOB"]="PAYER_ACCOUNT"
+          k6Map["ALICE"]="ACCOUNT_ADDRESS"
+          k6Map["fungible"]="TOKEN_ADDRESS"
+          k6Map["non_fungible"]="NON_FUNGIBLE_TOKEN_ADDRESS"
+
+          for key in "${!k6Map[@]}"; do
+            value=$(grep -E "${key}=" "$LOG_FILE" | \
+              awk -F'=' '{print $NF}' | \
+              tail -n 1 | \
+              tr -d '[:space:]' || true)
```
steven-sheehy: Should be able to simplify

```suggestion
            value=$(grep -oP "^${key}=(\K.+)$" "$LOG_FILE")
```

.github/workflows/acceptance.yaml
bilyana-gospodinova: Done

.github/workflows/acceptance.yaml
bilyana-gospodinova: Done

.github/workflows/acceptance.yaml
bilyana-gospodinova: Done

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/client/TokenClient.java
bilyana-gospodinova: Done

.github/workflows/acceptance.yaml
bilyana-gospodinova: Simplified

.github/workflows/acceptance.yaml
bilyana-gospodinova: Thanks, I am using GITHUB_ENV now.

.github/workflows/acceptance.yaml
bilyana-gospodinova: Addressed

.github/workflows/acceptance.yaml
bilyana-gospodinova: Done, although this returned multiple results, so I reused "tail -n 1" from the first approach.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/client/ContractClient.java
bilyana-gospodinova: Could you please expand on this comment? Do you mean that I should create a new method in the parent class, called "deleteAll" that accepts a map, checks the acceptance property (`acceptanceTestProperties.isSkipEntitiesCleanup()`) if the entities should be deleted or not, logs that they are not going to be deleted (if `true`), logs the data needed by the k6 tests and does the cleanup (if `false`). If that is the case my concern is that the method will get too complicated, with multiple ifs that check the entity type as the token, contract, account have their addresses logged in its own different way. This way a generic, parent class would need to know much about the child classes.

steven-sheehy: Mostly right, except that logging isn't important. As long as the method checks property and does the cleanup. So can skip logging if it's complicated.

bilyana-gospodinova: I did that partially, because in order to fully implement this we will need to make a larger refactoring. Currently we have:
- Map<String, ContractId> contractMap
- Map<TokenNameEnum, TokenResponse> tokenMap
- Map<AccountNameEnum, ExpandedAccountId> accountMap
As you can see the types in the map are quite different and we will need to have multiple if-statements in the parent method in order to get the id collection for each corresponding entity type. Perhaps we can do this in a followup PR as this will get quite large.

.github/workflows/acceptance.yaml
```diff
@@ -152,11 +153,75 @@ jobs:
           echo "Solo cluster setup completed successfully!"
 
       - name: Run acceptance tests
-        run: helm test mirror -n "${SOLO_NAMESPACE}" --logs --filter name=mirror-acceptance --timeout 20m
+        run: |
+          touch output_log.txt
+          helm test mirror -n "${SOLO_NAMESPACE}" --logs --filter name=mirror-acceptance --timeout 20m >> output_log.txt
+          cat output_log.txt
```
xin-hedera:
```suggestion
          helm test mirror -n "${SOLO_NAMESPACE}" --logs --filter name=mirror-acceptance --timeout 20m | tee output_log.txt
```

.github/workflows/acceptance.yaml
```diff
@@ -152,11 +153,75 @@ jobs:
...
+        with:
+          path: |
+            ./tools/k6/src/web3/apis.js
+          debug: false
+          flags: --quiet
 
       - name: Show Pod Logs on Failure
         continue-on-error: true
-        if: ${{ failure() }}
+        #        if: ${{ failure() }}
```
xin-hedera: should this get reverted?

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/client/TokenClient.java
```diff
@@ -65,18 +66,33 @@ public class TokenClient extends AbstractNetworkClient {
...
         log.info("Deleting {} tokens and dissociating {} token relationships", tokenIds.size(), associations.size());
-        deleteAll(tokenIds, tokenId -> delete(admin, tokenId));
+        deleteOrLogEntities(tokenIds, tokenId -> delete(admin, tokenId));
         deleteAll(associations.keySet(), association -> dissociate(association.accountId, association.tokenId));
     }
 
+    @Override
+    protected void logEntities() {
+        for (var tokenName : tokenMap.keySet()) {
+            log.info("Skipping cleanup of token [" + tokenName.getSymbol() + "] at address "
```
xin-hedera: nit

```suggestion
            log.info("Skipping cleanup of token [{}] at address {}", tokenName.getSymbol(), tokenMap.get(tokenName).tokenId().toEvmAddress());
```

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/client/TokenClient.java
```diff
@@ -65,18 +66,33 @@ public class TokenClient extends AbstractNetworkClient {
...
+                    + tokenMap.get(tokenName).tokenId().toEvmAddress());
+            // Log the values so that they can be parsed in CI and passed to the k6 tests as input.
+            // The token addresses need to be left-padded with zeroes in order to match the expected format.
+            System.out.println(tokenName.getSymbol() + "="
+                    + String.format(
+                            "%s%s",
+                            "0".repeat(24), tokenMap.get(tokenName).tokenId().toEvmAddress()));
```
xin-hedera, line 94: why does the function first log then do stdout print almost the same info?

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/config/AcceptanceTestProperties.java
```diff
@@ -93,6 +93,8 @@ public final class AcceptanceTestProperties {
     @NotNull
     private NodeNameEnum nodeType = NodeNameEnum.MIRROR;
 
+    private boolean skipEntitiesCleanup;
```
xin-hedera, line 97: document it in `docs/configuration.md`

.github/workflows/acceptance.yaml
```diff
@@ -152,11 +153,75 @@ jobs:
...
 
       - name: Run acceptance tests
-        run: helm test mirror -n "${SOLO_NAMESPACE}" --logs --filter name=mirror-acceptance --timeout 20m
+        run: |
+          touch output_log.txt
+          helm test mirror -n "${SOLO_NAMESPACE}" --logs --filter name=mirror-acceptance --timeout 20m >> output_log.txt
+          cat output_log.txt
+
+      - name: Parse acceptance tests logs and set k6 environment variables
+        if: ${{ matrix.stream-type != 'BLOCK' }}
```
xin-hedera, line 161: not in this PR, we should consider group these steps into a [composite action](https://docs.github.com/en/actions/tutorials/create-actions/create-a-composite-action) so don't have to repeat the same condition `if: ${{ matrix.stream-type != 'BLOCK' }}`

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/client/TokenClient.java
bilyana-gospodinova: The log is informative for us to know that the cleanup was not run because of the configured property `skipEntitiesCleanup`. The stdout is in a simplified format in order to be parsed easily in the next step of the flow where we need to set the env vars for the k6 tests.

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/config/AcceptanceTestProperties.java
bilyana-gospodinova: I have documented it in `test/README.md`. I don't see in `docs/configuration.md` any section related to the `test` module. Should I add a new section for the `test` module in `docs/configuration.md` or you meant `test/README.md`?

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/client/TokenClient.java
bilyana-gospodinova: Thanks, addressed.

.github/workflows/acceptance.yaml
bilyana-gospodinova: Fixed, I commented it out during testing and I forgot to revert it.

.github/workflows/acceptance.yaml
bilyana-gospodinova: Addressed

.github/workflows/acceptance.yaml
```diff
@@ -152,7 +153,70 @@ jobs:
...
+      - name: Run k6 web3 tests
+        if: ${{ matrix.stream-type != 'BLOCK' }}
+        uses: grafana/run-k6-action@a15e2072ede004e8d46141e33d7f7dad8ad08d9d # v1
+        env:
+          ACCOUNT_ADDRESS: ${{ env.ALICE }}
+          AMOUNT: "0000000000000000000000000000000000000000000000000000000000000001"
+          BASE_URL: "http://127.0.0.1:8080"
+          DEFAULT_CONTRACT_ADDRESS: ${{ env.PARENT }}
+          DEFAULT_DURATION: "3s"
+          DEFAULT_VUS: 2
```
xin-hedera, line 202: add
```
DEFAULT_GRACEFUL_STOP: '1s'
```

.github/workflows/acceptance.yaml
```diff
@@ -152,7 +153,70 @@ jobs:
...
+          RUN_MODIFICATION_TESTS: "false"
+          RUN_WITH_VARIABLES: "true"
+          SERIAL_NUMBER: "0000000000000000000000000000000000000000000000000000000000000001"
+          TOKEN_ADDRESS: ${{ env.fungible }}
+          WEB3_TEST_EXCLUDE: "^contractCall(Allowance|ApprovedForAll|BalanceOf|IsFrozen|IsKyc|Receive).*$"
+        with:
+          path: |
+            ./tools/k6/src/web3/apis.js
+          debug: false
+          flags: --quiet
```
xin-hedera, line 220: how do we know if it passes or not? should we upload the `report.md` file using `actions/upload-artifact`?

k6 may return non-zero code when some threshold is breached, that doesn't necessarily mean any functionality issues. will this step fail when k6 returns non-zero code?

Don't think we need to address all, however I believe we want to upload `report.md`

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/config/AcceptanceTestProperties.java
xin-hedera: yup you are right.

.github/workflows/acceptance.yaml
bilyana-gospodinova: I am not sure if we want to upload the `report.md` file. We don't aim to test the performance in CI. In the logs we see currently a summary of all of the tests that are run and how many of them passed.

This step fails if k6 returns a non-zero code. Example of a previous run: https://github.com/hiero-ledger/hiero-mirror-node/actions/runs/18158938367/job/51685190875.

xin-hedera (dismissed):
LGTM

.github/workflows/acceptance.yaml
```diff
@@ -152,7 +153,71 @@ jobs:
...
+          k6Map["ESTIMATE_PRECOMPILE"]="ESTIMATE_PRECOMPILE_CONTRACT"
+          k6Map["PRECOMPILE"]="PRECOMPILE_CONTRACT"
+          k6Map["BOB"]="PAYER_ACCOUNT"
+          k6Map["ALICE"]="ACCOUNT_ADDRESS"
+          k6Map["fungible"]="TOKEN_ADDRESS"
+          k6Map["non_fungible"]="NON_FUNGIBLE_TOKEN_ADDRESS"
+
+          for key in "${!k6Map[@]}"; do
+            value=$(grep -oP "^${key}=(\K.+)$" "$LOG_FILE" | tail -n 1  || true)
+            echo "${key}=${value}" >> $GITHUB_ENV
```
steven-sheehy, line 179: Why do we have a k6Map if we don't even use its values? You're still not doing like I provided explicit code for in the original [comment](https://github.com/hiero-ledger/hiero-mirror-node/pull/12151#discussion_r2417903495).

```bash
  k6="${k6Map[$key]}"
  value=$(# find acceptance value in log)
  echo "${k6}=${value}" >> $GITHUB_ENV
```

.github/workflows/acceptance.yaml
```diff
@@ -152,7 +153,71 @@ jobs:
...
 
       - name: Run acceptance tests
-        run: helm test mirror -n "${SOLO_NAMESPACE}" --logs --filter name=mirror-acceptance --timeout 20m
+        run: |
+          touch output_log.txt
+          helm test mirror -n "${SOLO_NAMESPACE}" --logs --filter name=mirror-acceptance --timeout 20m | tee output_log.txt
+
+      - name: Parse acceptance tests logs and set k6 environment variables
+        if: ${{ matrix.stream-type != 'BLOCK' }}
+        id: log-variables
```
steven-sheehy, line 162: This id is not needed and is misnamed

.github/workflows/acceptance.yaml
```diff
@@ -152,7 +153,71 @@ jobs:
...
+          kubectl port-forward "service/mirror-web3" -n mirror 8080:80 &
+
+      - name: Run k6 web3 tests
+        if: ${{ matrix.stream-type != 'BLOCK' }}
+        uses: grafana/run-k6-action@a15e2072ede004e8d46141e33d7f7dad8ad08d9d # v1
+        env:
+          ACCOUNT_ADDRESS: ${{ env.ALICE }}
+          AMOUNT: "0000000000000000000000000000000000000000000000000000000000000001"
+          BASE_URL: "http://127.0.0.1:8080"
+          DEFAULT_CONTRACT_ADDRESS: ${{ env.PARENT }}
```
steven-sheehy, line 199: As I already stated in a previous [comment](https://github.com/hiero-ledger/hiero-mirror-node/pull/12151#discussion_r2420807265), remove all lines that manually map env variables from the `Parse acceptance tests logs and set k6 environment variables` step. These should already be exported into the github environment in that previous step.

bilyana-gospodinova: Addressed in #12206. I tested it before but it seemed like it did not work, so I left it this way. It seems like I did something wrong back them as it seems to work now.

.github/workflows/acceptance.yaml
bilyana-gospodinova: Addressed in a follow up PR - #12206

PR #12166, opened 2025-10-11
Convert exchange rate API to Java

### LINKED ISSUE #12101: Convert `/network/exchangerate` to Java
### Problem

As a maintainer, I'd like to convert `/network/exchangerate` to Java to ease maintenance and improve performance.

### Solution

# REST API Conversion Checklist

This checklist verifies that the JavaScript-based REST API was converted to Java successfully.

## Phase One

Phase one implements the new API in Java with it disabled by default. Thorough testing is conducted to verify no
regressions are introduced.

### REST Java

- [ ] Controller, services, repositories, etc. implemented in rest-java
- [ ] Use Jooq for any dynamic SQL, repositories for any static queries
- [ ] New `Cache-Control` entry added to `rest-java/src/main/resources/application.yml`
- [ ] Controller integration tests added with equivalent or better coverage to existing rest spec tests.

### REST

- [ ] `REST_JAVA_INCLUDE` and `testFiles` updated in `rest/build.gradle.kts`
- [ ] All existing rest spec tests passing against rewritten Java endpoint

### Acceptance

- [ ] Acceptance tests updated to use `MirrorNodeClient.callConvertedRestEndpoint()`
- [ ] Passing acceptance tests

### Performance

- [ ] K6 tests for endpoint duplicated in rest-java folder
- [ ] K6 tests passing with equal to or greater than rps as older API

### Deployment

- [ ] New rest-java route added to `docker-compose.yml` haproxy config
- [ ] New `routes.<apiName>` property added to `charts/hedera-mirror-rest-java/values.yaml` and defaulted to `false`
- [ ] New route added to ingress conditional on routes property
- [ ] New route added to gateway conditional on routes property
- [ ] Enable route in integration environment before tagging and test it
- [ ] Enable route in previewnet and watch metrics/logs
- [ ] Enable route in both testnet clusters for a release and watch metrics/logs

## Phase Two

Phase two enables the new route by default so that it rolls out to mainnet for the first time. The legacy code is kept
until we're confident there are no regressions.

- [ ] Enable route by default in Ingress and Gateway
- [ ] Update JavaScript monitor to use REST Java URL for route
- [ ] Remove temporary enablement in previewnet and testnet so that default is used

## Phase Three

Phase three cleans up the legacy code and configuration so that only the rest-java route remains.

- [ ] Delete JavaScript code and tests
- [ ] Delete `Cache-Control` entry in rest config
- [ ] Delete duplicate rest k6 test
- [ ] Remove `routes.<apiName>` property in chart and its use in ingress and gateway
- [ ] Acceptance tests updated to use `MirrorNodeClient.callRestJavaEndpoint()`
- [ ] `REST_JAVA_INCLUDE` and `testFiles` updated to remove route and specs in `rest/build.gradle.kts`

### Alternatives

_No response_

steven-sheehy:
**Description**:

* Add `/api/v1/network/exchangerate` to rest-java module
* Add k6 tests for rest-java exchange rate
* Add `routes.exchangeRate` with a default of false to Helm chart
* Change acceptance tests to route to both old and new endpoints
* Fix timestamp parameter regex in openapi.yml
* Resolve cross-site scripting attacks by not printing request values to response on validation failure

**Related issue(s)**:

Fixes #12101

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/importer/src/main/java/org/hiero/mirror/importer/repository/FileDataRepository.java
+++ b/importer/src/main/java/org/hiero/mirror/importer/repository/FileDataRepository.java
@@ -48,10 +48,11 @@ public interface FileDataRepository extends CrudRepository<FileData, Long> {
             long startConsensusTimestamp, long endConsensusTimestamp, Collection<Long> entityIds, long limit);
 
     @Query(
+            nativeQuery = true,
             value =
                     """
             select
-              min(consensus_timestamp) as consensus_timestamp,
+              max(consensus_timestamp) as consensus_timestamp,
               ?1 as entity_id,
               string_agg(file_data, '' order by consensus_timestamp) as file_data,
               null as transaction_type
@@ -62,13 +63,11 @@ public interface FileDataRepository extends CrudRepository<FileData, Long> {
                 from file_data
                 where entity_id = ?1
                   and consensus_timestamp <= ?2
-                  and (transaction_type = 17
-                         or (transaction_type = 19
-                              and
-                             length(file_data) <> 0))
+                  and (transaction_type = 17 or (transaction_type = 19 and length(file_data) <> 0))
               order by consensus_timestamp desc
               limit 1
-            ) and consensus_timestamp <= ?2""",
-            nativeQuery = true)
+            ) and consensus_timestamp <= ?2
+              and (transaction_type <> 19 or length(file_data) <> 0)
+            """)
     Optional<FileData> getFileAtTimestamp(long fileId, long timestamp);
 }
--- a/importer/src/test/java/org/hiero/mirror/importer/repository/FileDataRepositoryTest.java
+++ b/importer/src/test/java/org/hiero/mirror/importer/repository/FileDataRepositoryTest.java
@@ -16,7 +16,7 @@ import org.hiero.mirror.importer.ImporterIntegrationTest;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
-class FileDataRepositoryTest extends ImporterIntegrationTest {
+final class FileDataRepositoryTest extends ImporterIntegrationTest {
 
     @Resource
     private FileDataRepository fileDataRepository;
@@ -133,6 +133,7 @@ class FileDataRepositoryTest extends ImporterIntegrationTest {
                 .customize(
                         f -> f.entityId(create.getEntityId()).transactionType(TransactionType.FILEAPPEND.getProtoId()))
                 .persist();
+        expected.setConsensusTimestamp(append.getConsensusTimestamp());
         expected.setFileData(Bytes.concat(create.getFileData(), append.getFileData()));
         softly.assertThat(fileDataRepository.getFileAtTimestamp(entityId, append.getConsensusTimestamp()))
                 .contains(expected);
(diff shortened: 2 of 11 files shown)


rest-java/src/main/java/org/hiero/mirror/restjava/controller/NetworkController.java
```diff
@@ -2,25 +2,61 @@
...
+
+    private Bound timestampBound(TimestampParameter[] timestamp) {
+        if (timestamp == null || timestamp.length == 0) {
+            return Bound.EMPTY;
+        }
+
+        for (int i = 0; i < timestamp.length; ++i) {
+            var param = timestamp[i];
+            if (param.operator() == RangeOperator.EQ) {
+                timestamp[i] = new TimestampParameter(RangeOperator.LTE, param.value());
```
IvanKavaldzhiev, line 56: Why we substitute the EQ operator with LTE?

steven-sheehy: This API returns a single, most recent exchange rate as of the given timestamp. An exchange rate is active not just for a single nanosecond but until the next exchange rate file update. https://github.com/hiero-ledger/hiero-mirror-node/blob/92e69b97cdce97305b1697d6981d4eb2cc1b1df0/rest/controllers/networkController.js#L126

IvanKavaldzhiev: Got it

rest-java/src/main/java/org/hiero/mirror/restjava/service/FileServiceImpl.java
```diff
@@ -0,0 +1,72 @@
...
+                                    try {
+                                        return new SystemFile<>(fileData, parser.apply(fileData.getFileData()));
+                                    } catch (Exception e) {
+                                        log.warn(
+                                                "Attempt {} failed to load file {} at {}, falling back to previous file.",
+                                                context.getRetryCount() + 1,
+                                                entityId,
+                                                fileData.getConsensusTimestamp(),
+                                                e);
+                                        upperBound.set(fileData.getConsensusTimestamp() - 1);
```
bilyana-gospodinova, line 63: If we only adjust the upper bound on retry isn't it possible to get to a case where lowerBound > upperBound which could not succeed? Shouldn't we adjust the lower bound as well?

xin-hedera: not a problem when `lowerBound > upperBound` since the flow will go to the branch ` .orElseThrow(() -> new EntityNotFoundException("File %s not found".formatted(entityId)));`

bilyana-gospodinova:
In the CI logs I see:
`GET /api/v1/network/exchangerate in 3 ms: 404 No static resource`
Perhaps something's missing.

xin-hedera:
> In the CI logs I see: `GET /api/v1/network/exchangerate in 3 ms: 404 No static resource` Perhaps something's missing.

That's expected since the acceptance workflow uses  v0.139.0-rc3.

I'l check why the exchangerate requests are routed to rest-java, we may want to enable it later when an image becomes available, otherwise the workflow will always fail.

PR #12186, opened 2025-10-15
HIP-1299 Replace account_id if present in node in REST API

martingeorgiev1:
**Description:**
Update the getNodes REST API implementation replacing the account_id when present in the node entity, and add tests
Add unit tests

**Related issue(s):**

Fixes https://github.com/hiero-ledger/hiero-mirror-node/issues/12096

**Notes for reviewer:**

**Checklist**
- [X] Confirm that the getNodes API correctly prioritizes node.account_id when available
- [X] All new tests are passing locally (spotlessApply, gradlew build, etc.)

### DIFF
--- a/rest/__tests__/integrationDomainOps.js
+++ b/rest/__tests__/integrationDomainOps.js
@@ -1639,6 +1639,7 @@ const addNode = async (nodeInput) => {
 
   convertByteaFields(['admin_key'], node);
   node.grpc_proxy_endpoint = node.grpc_proxy_endpoint ? JSONStringify(node.grpc_proxy_endpoint) : null;
+  node.account_id = encodedIdFromSpecValue(node.account_id);
   await insertDomainObject('node', Object.keys(nodeDefaults), node);
 };
 


rest/__tests__/service/networkNodeService.test.js
```diff
@@ -576,6 +576,130 @@ describe('NetworkNodeService.getNetworkNodes tests node filter', () => {
   });
 });
 
+describe('NetworkNodeService node_account_id override logic', () => {
+  const nodeAccount1000 = EntityId.parseString('0.0.1000');
+  const nodeAccount1001 = EntityId.parseString('0.0.1001');
```
xin-hedera: do not include `shard.realm.` so the test will also work for non-zero shard / realm

```suggestion
  const nodeAccount1000 = EntityId.parseString('1000');
  const nodeAccount1001 = EntityId.parseString('1001');
```

rest/__tests__/service/networkNodeService.test.js
```diff
@@ -576,6 +576,130 @@ describe('NetworkNodeService.getNetworkNodes tests node filter', () => {
...
+        file_id: EntityId.systemEntity.addressBookFile102.toString(),
+        node_count: 2,
+      },
+    ];
+
+    const entries = [
+      {
+        consensus_timestamp: 10,
+        node_id: 0,
+        node_account_id: EntityId.parseString('0.0.3').toString(), // original
```
xin-hedera: same here

```suggestion
        node_account_id: EntityId.parseString('3').toString(), // original
```

rest/__tests__/service/networkNodeService.test.js
```diff
@@ -576,6 +576,130 @@ describe('NetworkNodeService.getNetworkNodes tests node filter', () => {
...
+      [EntityId.systemEntity.addressBookFile102.getEncodedId()],
+      'asc',
+      10
+    );
+
+    expect(result).toHaveLength(1);
+    expect(result[0].addressBookEntry.nodeAccountId).toEqual(nodeAccountOriginal.getEncodedId());
+  });
+});
+
```
xin-hedera, line 701: please also update one spec based integration test. e.g., rest/__tests__/specs/network/nodes/no-params.json

only need to change one entry in the `setup.nodes` array to have non-null account_id.

rest/model/node.js
```diff
@@ -6,6 +6,7 @@ class Node {
   static tableAlias = 'n';
   static tableName = 'node';
   static ADMIN_KEY = `admin_key`;
+  static ACCOUNT_ID = `account_id`;
```
xin-hedera, line 9: nit, sort it alphabetically so `ACCOUNT_ID` appears before `ADMIN_KEY`

xin-hedera (changes requested):

PR #12189, opened 2025-10-15
Properly set creator for ScheduleCreate transactions if there is top level EthereumTransaction

### LINKED ISSUE #12089: Extend `ScheduleCreate`'s `creator accountId` record parsing logic if the top-level call is an `EthereumTransaciton`
### Problem

With [HIP-756](https://hips.hedera.com/hip/hip-756) and [HIP-1215](https://hips.hedera.com/hip/hip-1215) we are exposing the Scheduling Service with a System Contract.
While testing the new System Contract API we came around an interesting observation for the Schedules created from `EthereumTransaction`s. The creator account was set to be the relay operator account which in this case is not entirely correct.

Here in the importer the `creatorAccountId` or the schedule creator is always set as the payer of the transaction.
https://github.com/hiero-ledger/hiero-mirror-node/blob/6062fc4961d33dd229202873a1cd36fd3d9967bb/importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/ScheduleCreateTransactionHandler.java#L55

This results in schedules linked only to the relay accounts making it impossible to use the account filter in the schedule API for schedules created by an `EthereumTransaction`.

Image

Image

Image

While the relay operator is the designated payer of the transaction the actual creator should be the sender of the transaction. 
This case is rather new and was not in consideration during the implementation of the Schedule Create record parsing so it seems feasible that it should be updated. 

Here is an example to visualise the current behaviour. As we know the ethereum transactions have a sender and a relayer: 

Image
 With the inner transactions "payed" by the relayer and when successful the Schedule Create provides a Schedule ID:

Image

Opening this ID we see the relayer account set as the creator of the schedule:

Image

If this new case is included in the parsing it will enable the user to query the MirrorNode for schedules by the actual creator/sender of the transaction.

### Solution

- Adjust the handling of ScheduleCreate transaction to set the correct creator account when the top-level call is an Ethereum call
- This change might (most likely will) require a migration in order to include previously created schedules with HIP-756 functions.

### Alternatives

_No response_

IvanKavaldzhiev:
**Description**:
This PR properly sets the creator field for ScheduleCreate transactions to be the sender of the Ethereum transaction in case there is a top level Ethereum transaction.

**Related issue(s)**:

Fixes #12089

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
+++ b/common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
@@ -62,11 +62,7 @@ public class RecordItem implements StreamItem {
     private final long consensusTimestamp;
 
     private final boolean blockstream;
-
-    @Setter
-    @NonFinal
-    private RecordItem parent;
-
+    private final RecordItem parent;
     private final EntityId payerAccountId;
     private final RecordItem previous;
     private final SignatureMap signatureMap;
--- a/importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/ScheduleCreateTransactionHandler.java
+++ b/importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/ScheduleCreateTransactionHandler.java
@@ -2,6 +2,7 @@
 
 package org.hiero.mirror.importer.parser.record.transactionhandler;
 
+import com.hederahashgraph.api.proto.java.AccountID;
 import jakarta.inject.Named;
 import org.hiero.mirror.common.domain.entity.Entity;
 import org.hiero.mirror.common.domain.entity.EntityId;
@@ -53,7 +54,7 @@ class ScheduleCreateTransactionHandler extends AbstractEntityCrudTransactionHand
         var body = recordItem.getTransactionBody().getScheduleCreate();
         long consensusTimestamp = recordItem.getConsensusTimestamp();
 
-        EntityId creatorAccount = recordItem.getPayerAccountId();
+        var creatorAccount = recordItem.getPayerAccountId();
 
         var parentRecordItem = recordItem.getParent();
         if (parentRecordItem != null && parentRecordItem.getEthereumTransaction() != null) {
@@ -62,7 +63,7 @@ class ScheduleCreateTransactionHandler extends AbstractEntityCrudTransactionHand
                     ? transactionRecord.getContractCreateResult()
                     : transactionRecord.getContractCallResult();
 
-            if (functionResult.getSenderId().getAccountNum() > 0) {
+            if (!AccountID.getDefaultInstance().equals(functionResult.getSenderId())) {
                 creatorAccount = EntityId.of(functionResult.getSenderId());
             }
         }


common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
```diff
@@ -62,7 +62,11 @@ public class RecordItem implements StreamItem {
     private final long consensusTimestamp;
 
     private final boolean blockstream;
-    private final RecordItem parent;
+
+    @Setter
+    @NonFinal
+    private RecordItem parent;
```
xin-hedera: I don't see why this change is needed.

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/ScheduleCreateTransactionHandler.java
```diff
@@ -52,7 +52,21 @@ protected void doUpdateTransaction(Transaction transaction, RecordItem recordIte
...
+        EntityId creatorAccount = recordItem.getPayerAccountId();
+
+        var parentRecordItem = recordItem.getParent();
+        if (parentRecordItem != null && parentRecordItem.getEthereumTransaction() != null) {
+            var transactionRecord = parentRecordItem.getTransactionRecord();
+            var functionResult = transactionRecord.hasContractCreateResult()
+                    ? transactionRecord.getContractCreateResult()
+                    : transactionRecord.getContractCallResult();
+
+            if (functionResult.getSenderId().getAccountNum() > 0) {
```
xin-hedera:
```suggestion
            if (!AccountID.getDefaultInstance().equals(functionResult.getSenderId())) {
```

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/ScheduleCreateTransactionHandler.java
```diff
@@ -52,7 +52,21 @@ protected void doUpdateTransaction(Transaction transaction, RecordItem recordIte
 
         var body = recordItem.getTransactionBody().getScheduleCreate();
         long consensusTimestamp = recordItem.getConsensusTimestamp();
-        var creatorAccount = recordItem.getPayerAccountId();
+
+        EntityId creatorAccount = recordItem.getPayerAccountId();
```
xin-hedera: nit

```suggestion
        var creatorAccount = recordItem.getPayerAccountId();
```

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
IvanKavaldzhiev: Indeed, it's not needed. Removed.

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/ScheduleCreateTransactionHandler.java
IvanKavaldzhiev: Fixed

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/ScheduleCreateTransactionHandler.java
IvanKavaldzhiev: Fixed

PR #12195, opened 2025-10-16
Upgrade components in acceptance test workflow

### LINKED ISSUE #12152: Update acceptance workflow
### Problem

Upgrade acceptance workflow components

- consensus node: v0.66.0 or v0.67.0 when released
- block node: v0.20.0
- mirror node: v0.140.0
- solo: v0.46.1 - some commands have changed since v0.43.0, refer to solo cli migration guide: https://github.com/hiero-ledger/solo/blob/main/docs/site/content/en/docs/cli-migrations.md

In addition:

- add the workflow file to `on.pull_request.paths`

### Solution

see the description

### Alternatives

_No response_

xin-hedera:
**Description**:

This PR upgrades components in acceptance test workflow

- block node v0.20.1, consensus node v0.67.0, mirror node v0.141.0-rc1, nodejs v22, solo v0.46.1
- migrate solo cli commands
- use new helm release name `mirror-1` to run acceptance test
- use new block node service name `block-node-1`

**Related issue(s)**:

Fixes #12152 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/.github/workflows/acceptance.yaml
+++ b/.github/workflows/acceptance.yaml
@@ -157,7 +157,7 @@ jobs:
           echo "Solo cluster setup completed successfully!"
 
       - name: Run acceptance tests
-        run: stdbuf -oL helm test "${HELM_RELEASE_NAME}" -n "${SOLO_NAMESPACE}" --logs --filter name="${HELM_RELEASE_NAME}-mirror-acceptance" --timeout 20m | tee output_log.txt
+        run: helm test "${HELM_RELEASE_NAME}" -n "${SOLO_NAMESPACE}" --logs --filter name="${HELM_RELEASE_NAME}-mirror-acceptance" --timeout 20m | tee output_log.txt
 
       - name: Parse acceptance tests logs and set k6 environment variables
         if: ${{ matrix.stream-type != 'BLOCK' }}


xin-hedera:
looks like running with block node is unstable, the job succeeded after 2-3 runs.

.github/workflows/acceptance.yaml
```diff
@@ -40,13 +41,15 @@ jobs:
           - BLOCK
     timeout-minutes: 40
     env:
-      BLOCK_NODE_CHART_VERSION: v0.17.1
-      CONSENSUS_VERSION: v0.65.1
-      MIRROR_NODE_VERSION: v0.139.0-rc3
+      BLOCK_NODE_CHART_VERSION: v0.20.1
+      CONSENSUS_VERSION: v0.67.0
+      HELM_RELEASE_NAME: mirror-1
+      MIRROR_NODE_VERSION: v0.141.0-rc1
```
jnels124, line 47: would it be better to find the latest tag and use that for mirrornode version? I guess that potentially will cause incompatibilities with other components but just wondering if it is worth it to do so.

jnels124 (approved):

.github/workflows/acceptance.yaml
rbarker-dev: We should always use a specific version. It might be a good idea **(future)** to add this as an input to the flow that could be specified for testing. That way you could do `${{ inputs.mirror-node-version }}` and pass `latest` in there.

.github/workflows/acceptance.yaml
rbarker-dev, line 1: If solo version is `< 0.44.0` this workflow will be incompatible. There should be an error condition for that.

rbarker-dev (approved):

.github/workflows/acceptance.yaml
xin-hedera: there is no `latest`  tag for helm charts, however that can be retrieved with `helm search repo --devel hedera-mirror-common`. That'll make it automatic and should work for most scenarios.

We do still have the chicken-egg problem through: when a PR changes application logic, the acceptance tests may either fail due to other modules don't have the change or report false positive when acceptance test changes mismatch.

So ideally, we should do

- always use local helm charts from the triggering branch
-  build custom images for modules with code change

.github/workflows/acceptance.yaml
xin-hedera: agree. however, don't think we'll ever downgrade solo.

rbarker-dev: Definitely not a blocker. Just a nice to have in case someone (like me) runs an XTS flow in consensus node with solo version 0.43.2 or 0.38.0 for example.

steven-sheehy (approved):

PR #12196, opened 2025-10-16
Enhance the contract results REST endpoint to return latest failed result

### LINKED ISSUE #12116: Enhance /contracts/results/{transactionIdOrHash} endpoint
### Problem

If a transaction fails against the consensus node, as a user I want to be able to get the contract result and check the reason for the fail. With the current implementation of the endpoint, the `WRONG_NONCE` and `DUPLICATE_TRANSACTION` transaction results are filtered and no result is returned in some cases although there is such. 

### Solution

The /contracts/results/{transactionIdOrHash} endpoint can be changed to remove the `WRONG_NONCE` and `DUPLICATE_TRANSACTION` filter and return the latest successful result (if there is such) or the latest unsuccessful.

### Alternatives

_No response_

bilyana-gospodinova:
**Description**:
This PR changes the `/contracts/results/{transactionIdOrHash}` endpoint in the following way - in case of `WRONG_NONCE` or `DUPLICATE_TRANSACTION` result they are not filtered anymore and the latest unsuccessful result is returned (if there is no successful). This way the user can know in all cases the reason why a transaction failed (when there is a contract result available).

The transaction queries are modified to reflect the change in the implementation above. The explicit filtering of both statuses is removed. Also, only the needed result is returned now by the queries. Previously, a list of results was returned and then in the `contractController` only the first result was used. Also, in case multiple transactions were returned, an error was thrown.

With the new implementation an error is not thrown anymore as only one result is returned by the query (the latest, matching by nonce, if provided). If a nonce is not provided, a 0 is assumed for its value. This way this matches the previous implementation where a list was returned, sorted ascending and only the first result was used (effectively nonce == 0).

To reflect the changes in this PR, multiple tests needed to be adapted to match the change in the transaction queries.

Fixes #12116

### DIFF
--- a/rest/__tests__/service/transactionService.test.js
+++ b/rest/__tests__/service/transactionService.test.js
@@ -92,7 +92,7 @@ describe('TransactionService.getTransactionDetailsFromTransactionId tests', () =
       0
     );
     expect(pickTransactionFields(actual)).toIncludeSameMembers([
-      {consensusTimestamp: 6, payerAccountId: defaultPayerEncodedId}
+      {consensusTimestamp: 6, payerAccountId: defaultPayerEncodedId},
     ]);
   });
 
@@ -101,7 +101,7 @@ describe('TransactionService.getTransactionDetailsFromTransactionId tests', () =
       TransactionId.fromString(`${defaultPayerId.toString()}-0-${duplicateValidStartNs}`)
     );
     expect(pickTransactionFields(actual)).toIncludeSameMembers([
-      {consensusTimestamp: 11, payerAccountId: defaultPayerEncodedId}
+      {consensusTimestamp: 11, payerAccountId: defaultPayerEncodedId},
     ]);
   });
 
--- a/rest/service/transactionService.js
+++ b/rest/service/transactionService.js
@@ -65,10 +65,12 @@ class TransactionService extends BaseService {
    */
   async getTransactionDetailsFromTransactionId(transactionId, nonce = undefined) {
     const maxConsensusTimestamp = BigInt(transactionId.getValidStartNs()) + maxTransactionConsensusTimestampRangeNs;
-    return this.getTransactionDetails(
-      TransactionService.transactionDetailsFromTransactionIdQuery,
-      [transactionId.getEntityId().getEncodedId(), transactionId.getValidStartNs(), maxConsensusTimestamp, nonce],
-    );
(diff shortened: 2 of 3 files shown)


rest/service/contractService.js
```diff
@@ -217,7 +215,7 @@ class ContractService extends BaseService {
           ), (
```
jnels124: would it be better to just do something like this?
```
SELECT *
FROM contract_transaction_hash
WHERE hash = $1
ORDER BY (transaction_result = 22) DESC,  -- prefer result 22
         consensus_timestamp DESC          -- then newest
LIMIT 1;
```

rest/service/transactionService.js
```diff
@@ -19,9 +20,26 @@ class TransactionService extends BaseService {
            ${Transaction.SCHEDULED}, ${Transaction.TYPE},
            ${Transaction.PAYER_ACCOUNT_ID}
     from ${Transaction.tableName}
-    where ${Transaction.PAYER_ACCOUNT_ID} = $1 
-          and ${Transaction.CONSENSUS_TIMESTAMP} >= $2 and ${Transaction.CONSENSUS_TIMESTAMP} <= $3
-          and ${Transaction.VALID_START_NS} = $2`;
+    where ${Transaction.CONSENSUS_TIMESTAMP} = coalesce((
```
jnels124: think this can be similar to previous comment

rest/service/contractService.js
bilyana-gospodinova: Addressed. Thanks, the query looks a lot more simpler and cleaner now.

rest/service/transactionService.js
bilyana-gospodinova: Addressed

rest/__tests__/service/transactionService.test.js
```diff
@@ -17,7 +17,6 @@ const contractCreateType = TransactionType.getProtoId('CONTRACTCREATEINSTANCE');
 const ethereumTxType = TransactionType.getProtoId('ETHEREUMTRANSACTION');
 const duplicateTransactionResult = TransactionResult.getProtoId('DUPLICATE_TRANSACTION');
```
IvanKavaldzhiev, line 18: Is duplicateTransactionResult used any more? Probably we can delete it too.

bilyana-gospodinova: It is used in 3 places.

jnels124 (approved):
LGTM

PR #12204, opened 2025-10-16
add staging deploy workflow trigger

jnels124:
add staging deploy change workflow trigger

### DIFF
--- a/clusters/mainnet-staging-na/mainnet-citus/helmrelease.yaml
+++ b/clusters/mainnet-staging-na/mainnet-citus/helmrelease.yaml
@@ -3,7 +3,6 @@ kind: HelmRelease
 metadata:
   name: mirror
   namespace: mainnet-citus
-  # Trigger staging environemnt - workflow -- Remove in >= 0.142.0-rc1
 spec:
   chart:
     spec:
@@ -132,7 +131,7 @@ spec:
               acceptance:
                 network: MAINNET
       cucumberTags: "@critical"
-      enabled: true
+      enabled: false
     web3:
       env:
         HIERO_MIRROR_WEB3_EVM_NETWORK: MAINNET


.github/workflows/trigger-staging-deploy.yml
jnels124, line 1: this needs to be on the deploy branch I think?

.github/workflows/trigger-staging-deploy.yml
```diff
@@ -0,0 +1,31 @@
...
+    paths:
+      - clusters/mainnet-staging-na/mainnet-citus/helmrelease.yaml
+
+permissions:
+  id-token: write
+  contents: read
+
+jobs:
+  run-reusable-copy:
+    uses: hiero-ledger/hiero-mirror-node/.github/workflows/copy-stackgres-cluster.yml@7d9e935073507e45738098eb05a7c15f7658b565
```
xin-hedera: should we just use `main` so it always runs the latest?

```suggestion
    uses: hiero-ledger/hiero-mirror-node/.github/workflows/copy-stackgres-cluster.yml@main
```

jnels124: That makes sense to me. Changed it

xin-hedera (approved):
LGTM

clusters/mainnet-staging-na/mainnet-citus/helmrelease.yaml
```diff
@@ -131,7 +131,7 @@ spec:
               acceptance:
                 network: MAINNET
       cucumberTags: "@critical"
-      enabled: true
+      enabled: false
```
steven-sheehy, line 134: Why do we disable acceptance tests?

jnels124: acceptance was already ran. I created the 142 checklist already with a step to re-enable. will trigger k6 tests by creating a dummy pod

PR #12243, opened 2025-10-21
Rest-java k6 tests in CI

### LINKED ISSUE #12161: Test k6 tests in CI - rest java
### Problem

k6 tests can have functional issues that we don't catch until late in the testing cycle.

### Solution

Add k6 tests in CI for the rest java module.

### Alternatives

_No response_

bilyana-gospodinova:
**Description**:
Add rest-java k6 tests execution as part of the workflow in order to verify the end-to-end functionality. This will be helpful to catch some issues earlier in the testing cycle. The goal is not a complete coverage of all rest-java k6 tests, so the failing ones are excluded for now.

Fixes #12161

### DIFF
--- a/.github/workflows/acceptance.yaml
+++ b/.github/workflows/acceptance.yaml
@@ -189,15 +189,13 @@ jobs:
           declare -A k6MapRestJava
           k6MapRestJava["DEFAULT_TOPIC"]="DEFAULT_TOPIC_WITH_FEE_ID"
           k6MapRestJava["PRECOMPILE"]="DEFAULT_ACCOUNT_ID_NFTS_ALLOWANCE_SPENDER"
+          k6MapRestJava["DEFAULT_ACCOUNT_ID_BOB"]="DEFAULT_ACCOUNT_ID_AIRDROP_RECEIVER"
 
           for key in "${!k6MapRestJava[@]}"; do
             k6="${k6MapRestJava[$key]}"
             value=$(grep -oP "^${key}=(\K.+)$" "$LOG_FILE" | tail -n 1  || true)
             echo "${k6}=${value}" >> $GITHUB_ENV
           done
-
-          value=$(grep -oP "^DEFAULT_ACCOUNT_ID_BOB=(\K.+)$" "$LOG_FILE" | tail -n 1  || true)
-          echo "DEFAULT_ACCOUNT_ID_AIRDROP_RECEIVER=${value}" >> $GITHUB_ENV
 
       - name: Install k6
         if: ${{ matrix.stream-type != 'BLOCK' }}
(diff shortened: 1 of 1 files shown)


test/src/test/java/org/hiero/mirror/test/e2e/acceptance/client/TopicClient.java
```diff
@@ -49,7 +49,16 @@ public TopicClient(
...
-        deleteAll(topicIds, this::deleteTopic);
+        deleteOrLogEntities(topicIds, this::deleteTopic);
+    }
+
+    @Override
+    protected void logEntities() {
+        for (var topicId : topicIds) {
+            log.info("Skipping cleanup of topic at address {}.", topicId);
+            // Log the values so that they can be parsed in CI and passed to the k6 tests as input.
+            System.out.println("DEFAULT_TOPIC=" + topicId.num);
```
steven-sheehy: Why do we log both via logger and system.out? Should be one or the other.

.github/workflows/acceptance.yaml
```diff
@@ -186,11 +187,26 @@ jobs:
...
+
+      - name: Run rest-java tests
+        if: ${{ matrix.stream-type != 'BLOCK' }}
+        uses: grafana/run-k6-action@a15e2072ede004e8d46141e33d7f7dad8ad08d9d # v1
+        env:
+          BASE_URL: "http://127.0.0.1:8084"
+          DEFAULT_DURATION: "3s"
+          DEFAULT_GRACEFUL_STOP: "1s"
+          DEFAULT_VUS: 2
+          RESTJAVA_TEST_EXCLUDE: "^(accounts(NftAllowanceOwner|NftAllowanceSpender|OutstandingAirdrop|PendingAirdrop))|rampUp.*$"
```
steven-sheehy: Is there not an acceptance test that adds these?

test/src/test/java/org/hiero/mirror/test/e2e/acceptance/client/TopicClient.java
bilyana-gospodinova: The logger is now removed.

.github/workflows/acceptance.yaml
bilyana-gospodinova: There are, but also there are steps that delete the allowances at the end or claim the airdrops, so after the execution of the acceptance tests they are no longer available. I will see what I can do without adding a bunch of new tests just to do the k6 setup.

.github/workflows/acceptance.yaml
```diff
@@ -182,25 +186,38 @@ jobs:
             echo "${k6}=${value}" >> $GITHUB_ENV
           done
 
+          declare -A k6MapRestJava
+          k6MapRestJava["DEFAULT_TOPIC"]="DEFAULT_TOPIC_WITH_FEE_ID"
+          k6MapRestJava["PRECOMPILE"]="DEFAULT_ACCOUNT_ID_NFTS_ALLOWANCE_SPENDER"
+          k6MapRestJava["DEFAULT_ACCOUNT_ID_BOB"]="DEFAULT_ACCOUNT_ID_AIRDROP_RECEIVER"
+
```
steven-sheehy, line 193: We shouldn't need a separate rest java map. Can just add to regular k6Map since they all get added to the GITHUB_ENV in the end anyway.

.github/workflows/acceptance.yaml
```diff
@@ -182,25 +186,38 @@ jobs:
...
-        run: |
-          echo "Port forwarding 8080:80"
-          kubectl port-forward "service/${HELM_RELEASE_NAME}-web3" -n "${SOLO_NAMESPACE}" 8080:80 &
+        uses: grafana/run-k6-action@a15e2072ede004e8d46141e33d7f7dad8ad08d9d # v1
+        env:
+          DEFAULT_ACCOUNT_ID_NFTS_ALLOWANCE_OWNER: "0.0.1003"
+          DEFAULT_ACCOUNT_ID_AIRDROP_SENDER: "0.0.1003"
+        with:
+          path: |
+            ./tools/k6/src/rest-java/apis.js
```
steven-sheehy, line 212: Can't we just run them all at once in a single step by including multiple paths?

.github/workflows/acceptance.yaml
bilyana-gospodinova: The need for the second map came because there are duplicate keys in web3 and rest-java. Like:
`k6Map["PRECOMPILE"]="HTS_CONTRACT_ADDRESS"` in web3
`k6MapRestJava["PRECOMPILE"]="DEFAULT_ACCOUNT_ID_NFTS_ALLOWANCE_SPENDER"` in rest-java
If we use only 1 map the value will be overriden.
With adding more tests, I suppose these duplicates will get even more so I introduced a new map.

.github/workflows/acceptance.yaml
bilyana-gospodinova: We can. I added 2 separate steps because if one of them fails we will know directly in which module there is an issue. Otherwise, we need to go through the logs to check.

xin-hedera (approved):
LGTM

PR #12259, opened 2025-10-23
HIP-1195 Add hooks REST API

### LINKED ISSUE #12121: HIP-1195 Add hooks REST API
### Problem

As per the design document, will need to add api for hooks for a account and contract 
#11731 

### Solution

 - Update openapi.yml with hook endpoint definitions
 - In rest-java module, add New api interface for /accounts/{}/hooks

### Alternatives

_No response_

### LINKED ISSUE #12292: HIP-1195 Optimize the queries of Hooks REST API
### Problem

Measure and potentially Improve the repository layer queries performance of the Hooks REST API ( https://github.com/hiero-ledger/hiero-mirror-node/issues/12121 ), by applying the solution proposed within one of the PR comments: 
https://github.com/hiero-ledger/hiero-mirror-node/pull/12259

### Solution

> ... We only need two repository methods:
> 
> ```java
>     List<Hook> findByOwnerIdAndHookIdGreaterThanEqualAndHookIdLessThanEqual(long ownerId, long hookIdLower, long hookIdUpper, Pageable pageable);
> 
>     List<Hook> findByOwnerIdAndHookIdIn(long ownerId, Collection<Long> hookIds, Pageable pageable);
> ```
> 
> The pageable should auto generate the limit and order by. If it includes an offset keyword in the SQL it shouldn't matter since the full primary key is specified via the other params. Fundamentally, the REST API query params for this particular API will result in 4 scenarios so we can get away with using just the two static queries above:
> 
> * One or more `eq` only params only
>   
>   * Collect all hook Ids and use `findByOwnerIdAndHookIdIn()`
> * One or more `lt|lte|gt|gte` query params only
>   
>   * Normalize lt and gt to lte and gte by adding/subtracting one nanosecond
>   * Pick min lt(e) and max gt(e)
>   * Default missing lower bound to 0 and missing upper bound to MAX_VALUE
>   * Pass resulting single lower and single upper bound to `findByOwnerIdAndHookIdGreaterThanEqualAndHookIdLessThanEqual`
> * Both eq and range params without overlap
>   
>   * Skip repository and return empty list
> * Both eq and range params with overlap in range
>   
>   * Collect all hook Ids and use `findByOwnerIdAndHookIdIn()`

### Alternatives

_No response_

sdimitrov9:
**Description**:
Introducing `GET /api/v1/accounts/{idOrAliasOrEvmAddress}/hooks` endpoint.
As stated in the hiero-hooks.md it should return hooks data in the following format:
```
{
  "hooks": [
    {
      "admin_key": {
        "_type": "ED25519",
        "key": "0x302a300506032b6570032100e5b2…"
      },
      "contract_id": "0.0.456",
      "created_timestamp": "1726874345.123456789",
      "deleted": false,
      "extension_point": "ACCOUNT_ALLOWANCE_HOOK",
      "hook_id": 5,
      "owner_id": "0.0.123",
      "timestamp_range": {
                "from": "1700000000.000000008",
                "to": "1700000001.000000008"
      },
      "type": "LAMBDA"
    }
  ],
  "links": {
    "next": "/api/v1/accounts/0.0.123/hooks?hook.id=lt:5&limit=1"
  }
}
```

Note: Hooks in the response are always sorted based on their hook_id in descending order. The hook.id parameter in the links.next URL uses the format hook.id=lt:5 for pagination.

#### Query Parameters

| Parameter | Type    | Description                                                                                 | Default | Validation                                                       |
| --------- | ------- | ------------------------------------------------------------------------------------------- | ------- | ---------------------------------------------------------------- |
| `hook.id` | string  | Filter hooks by hook ID. Supports comparison operators: `eq:`, `gt:`, `gte:`, `lt:`, `lte:` | none    | Must be a valid positive integer when using comparison operators |
| `limit`   | integer | Maximum number of hooks to return                                                           | `25`    | Must be between 1 and 100                                        |
| `order`   | string  | Sort order for results                                                                      | `desc`  | Must be either `asc` or `desc`                                   |

**Related issue(s)**:

Fixes #12121
Fixes #12292 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/rest-java/src/main/java/org/hiero/mirror/restjava/dto/HooksRequest.java
+++ b/rest-java/src/main/java/org/hiero/mirror/restjava/dto/HooksRequest.java
@@ -16,17 +16,17 @@ public class HooksRequest {
     EntityIdParameter ownerId;
 
     @Builder.Default
-    int limit = 25;
+    private final int limit = 25;
 
     @Builder.Default
-    Sort.Direction order = Direction.DESC;
+    private final Sort.Direction order = Direction.DESC;
 
     @Builder.Default
-    Collection<Long> hookIdEqualsFilters = Collections.emptyList();
+    private final Collection<Long> hookIdEqualsFilters = Collections.emptyList();
 
     @Builder.Default
-    Long hookIdLowerBoundInclusive = 0L;
+    private final long hookIdLowerBoundInclusive = 0L;
 
     @Builder.Default
-    Long hookIdUpperBoundInclusive = Long.MAX_VALUE;
+    private final long hookIdUpperBoundInclusive = Long.MAX_VALUE;
 }
--- a/rest-java/src/test/java/org/hiero/mirror/restjava/controller/HooksControllerTest.java
+++ b/rest-java/src/test/java/org/hiero/mirror/restjava/controller/HooksControllerTest.java
@@ -18,14 +18,12 @@ import org.hiero.mirror.common.domain.hook.Hook;
 import org.hiero.mirror.rest.model.HooksResponse;
 import org.hiero.mirror.rest.model.Links;
 import org.hiero.mirror.restjava.mapper.HookMapper;
-import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.CsvSource;
 import org.junit.jupiter.params.provider.ValueSource;
-import org.springframework.data.domain.Sort;
 import org.springframework.web.client.HttpClientErrorException;
 import org.springframework.web.client.RestClient.RequestHeadersSpec;
 import org.springframework.web.client.RestClient.RequestHeadersUriSpec;
@@ -39,32 +37,55 @@ final class HooksControllerTest extends ControllerTest {
     @Nested
     final class HooksEndpointTest extends EndpointTest {
 
-        private final long accountId = 1234L;
+        private static final long ACCOUNT_ID = 1234L;
+        private static final String ALIAS = "HIQQEXWKW53RKN4W6XXC4Q232SYNZ3SZANVZZSUME5B5PRGXL663UAQA";
+        private static final String EVM_ADDRESS = "ac384c53f03855fa1b3616052f8ba32c6c2a2fec";
 
         @Override
         protected String getUrl() {
-            return "accounts/" + accountId + "/hooks";
+            return "accounts/{account_id}/hooks";
         }
 
         @Override
         protected RequestHeadersSpec<?> defaultRequest(RequestHeadersUriSpec<?> uriSpec) {
-            // given: at least one persisted hook for default request
-            persistHook(accountId);
-            return uriSpec.uri("");
+            persistHook(ACCOUNT_ID);
+
+            return uriSpec.uri("", ACCOUNT_ID);
         }
 
-        @Test
-        void success() {
+        @ParameterizedTest
+        @CsvSource({
+            // entity ID formats
+            "1234",
+            "0.1234",
+            "0.0.1234",
+            // alias formats
+            "HIQQEXWKW53RKN4W6XXC4Q232SYNZ3SZANVZZSUME5B5PRGXL663UAQA",
+            "0.HIQQEXWKW53RKN4W6XXC4Q232SYNZ3SZANVZZSUME5B5PRGXL663UAQA",
+            "0.0.HIQQEXWKW53RKN4W6XXC4Q232SYNZ3SZANVZZSUME5B5PRGXL663UAQA",
+            // evm formats
+            "ac384c53f03855fa1b3616052f8ba32c6c2a2fec",
+            "0.ac384c53f03855fa1b3616052f8ba32c6c2a2fec",
+            "0.0.ac384c53f03855fa1b3616052f8ba32c6c2a2fec",
+            "0xac384c53f03855fa1b3616052f8ba32c6c2a2fec"
+        })
+        void success(String address) {
             // given
-            final var hook1 = persistHook(accountId);
-            final var hook2 = persistHook(accountId);
+            persistAccount(ACCOUNT_ID, ALIAS, EVM_ADDRESS);
+
+            final var accountAddress = normalizeAddress(address);
+
+            final var hook1 = persistHook(ACCOUNT_ID);
+            final var hook2 = persistHook(ACCOUNT_ID);
+
             final var expectedHooks = hookMapper.map(List.of(hook2, hook1));
             final var expectedResponse = new HooksResponse();
             expectedResponse.setHooks(expectedHooks);
             expectedResponse.setLinks(new Links());
 
             // when
-            final var actual = restClient.get().uri("").retrieve().body(HooksResponse.class);
+            final var actual =
+                    restClient.get().uri("", accountAddress).retrieve().body(HooksResponse.class);
 
             // then
             assertThat(actual).isNotNull().isEqualTo(expectedResponse);
@@ -73,33 +94,17 @@ final class HooksControllerTest extends ControllerTest {
         @Test
         void noHooksFound() {
             // given
-            persistHook(accountId + 1);
+            persistHook(ACCOUNT_ID + 1);
             final var expectedHooks = hookMapper.map(Collections.emptyList());
             final var expectedResponse = new HooksResponse();
             expectedResponse.setHooks(expectedHooks);
             expectedResponse.setLinks(new Links());
 
             // when
-            final var actual = restClient.get().uri("").retrieve().body(HooksResponse.class);
+            final var actual = restClient.get().uri("", ACCOUNT_ID).retrieve().body(HooksResponse.class);
 
             // then
             assertThat(actual).isNotNull().isEqualTo(expectedResponse);
-        }
-
-        @Test
-        void orderAsc() {
-            final var hook1 = persistHook(accountId);
-            final var hook2 = persistHook(accountId);
-
-            assertHooksOrder(Sort.Direction.ASC, List.of(hook1, hook2));
-        }
-
-        @Test
-        void orderDesc() {
-            final var hook1 = persistHook(accountId);
-            final var hook2 = persistHook(accountId);
-
-            assertHooksOrder(Sort.Direction.DESC, List.of(hook2, hook1));
         }
 
         @ParameterizedTest
@@ -108,9 +113,9 @@ final class HooksControllerTest extends ControllerTest {
             // given
             final int limit = 2;
 
-            final var hook1 = persistHook(accountId);
-            final var hook2 = persistHook(accountId);
-            final var hook3 = persistHook(accountId);
+            final var hook1 = persistHook(ACCOUNT_ID);
+            final var hook2 = persistHook(ACCOUNT_ID);
+            final var hook3 = persistHook(ACCOUNT_ID);
 
             final List<Hook> expectedHookEntities = order.equals("asc") ? List.of(hook1, hook2) : List.of(hook3, hook2);
 
@@ -123,7 +128,7 @@ final class HooksControllerTest extends ControllerTest {
 
             final var nextLink = String.format(
                     "/api/v1/accounts/%d/hooks?limit=%d&order=%s&hook.id=%s:%d",
-                    accountId, limit, order, operator, lastHook.getHookId());
+                    ACCOUNT_ID, limit, order, operator, lastHook.getHookId());
 
             final var links = new Links();
             links.setNext(nextLink);
@@ -132,7 +137,10 @@ final class HooksControllerTest extends ControllerTest {
             // when
             final var actual = restClient
                     .get()
-                    .uri(String.format("?limit=%d&order=%s", limit, order))
+                    .uri(uriBuilder -> uriBuilder
+                            .queryParam("limit", limit)
+                            .queryParam("order", order)
+                            .build(Map.of("account_id", ACCOUNT_ID)))
                     .retrieve()
                     .body(HooksResponse.class);
 
@@ -187,7 +195,7 @@ final class HooksControllerTest extends ControllerTest {
             // given
             final var hooks = new ArrayList<Hook>();
             for (int i = 0; i < 4; i++) {
-                hooks.add(persistHook(accountId));
+                hooks.add(persistHook(ACCOUNT_ID));
             }
 
             final List<Hook> expectedHookList;
@@ -214,34 +222,12 @@ final class HooksControllerTest extends ControllerTest {
                     hooks.get(1).getHookId(),
                     hooks.get(2).getHookId(),
                     hooks.get(3).getHookId());
-            final var actual = restClient.get().uri(formattedParams).retrieve().body(HooksResponse.class);
+
+            final var actual =
+                    restClient.get().uri(formattedParams, ACCOUNT_ID).retrieve().body(HooksResponse.class);
 
             // then
             assertThat(actual).isNotNull().isEqualTo(expectedResponse);
-        }
-
-        @Test
-        void hookIdWithMissingOperatorDefaultsToEquals() {
-            // given
-            persistHook(accountId); // Another hook, to be filtered out
-            final var hook1 = persistHook(accountId);
-            persistHook(accountId); // Another hook, to be filtered out
-
-            final var expectedHooks = hookMapper.map(List.of(hook1));
-            final var expectedResponse = new HooksResponse();
-            expectedResponse.setHooks(expectedHooks);
-            expectedResponse.setLinks(new Links());
-
-            // when
-            final var queryString = "?hook.id=" + hook1.getHookId();
-            final var actual = restClient.get().uri(queryString).retrieve().body(HooksResponse.class);
-
-            // then
-            // Verify it's not null and matches the expected structure
-            assertThat(actual).isNotNull().isEqualTo(expectedResponse);
-            // Explicitly verify only the one hook we asked for was returned
-            assertThat(actual.getHooks()).hasSize(1);
-            assertThat(actual.getHooks().get(0).getHookId()).isEqualTo(hook1.getHookId());
         }
 
         @Test
@@ -254,7 +240,11 @@ final class HooksControllerTest extends ControllerTest {
 
             // when/then
             validateError(
-                    () -> restClient.get().uri(queryString).retrieve().toEntity(String.class),
+                    () -> restClient
+                            .get()
+                            .uri(queryString, ACCOUNT_ID)
+                            .retrieve()
+                            .toEntity(String.class),
                     HttpClientErrorException.BadRequest.class,
                     // Default JSR-303 message for @Size validation
                     "hookIdFilters size must be between 0 and " + MAX_REPEATED_QUERY_PARAMETERS);
@@ -263,7 +253,11 @@ final class HooksControllerTest extends ControllerTest {
         @Test
         void invalidLimitTooLow() {
             validateError(
-                    () -> restClient.get().uri("?limit=0").retrieve().toEntity(String.class),
+                    () -> restClient
+                            .get()
+                            .uri("?limit=0", ACCOUNT_ID)
+                            .retrieve()
+                            .toEntity(String.class),
                     HttpClientErrorException.BadRequest.class,
                     "limit must be greater than 0");
         }
@@ -271,7 +265,11 @@ final class HooksControllerTest extends ControllerTest {
         @Test
         void invalidLimitTooHigh() {
             validateError(
-                    () -> restClient.get().uri("?limit=101").retrieve().toEntity(String.class),
+                    () -> restClient
+                            .get()
+                            .uri("?limit=101", ACCOUNT_ID)
+                            .retrieve()
+                            .toEntity(String.class),
                     HttpClientErrorException.BadRequest.class,
                     "limit must be less than or equal to 100");
         }
@@ -279,7 +277,11 @@ final class HooksControllerTest extends ControllerTest {
         @Test
         void invalidLimitFormat() {
             validateError(
-                    () -> restClient.get().uri("?limit=abc").retrieve().toEntity(String.class),
+                    () -> restClient
+                            .get()
+                            .uri("?limit=abc", ACCOUNT_ID)
+                            .retrieve()
+                            .toEntity(String.class),
                     HttpClientErrorException.BadRequest.class,
                     "Failed to convert 'limit'");
         }
@@ -287,7 +289,11 @@ final class HooksControllerTest extends ControllerTest {
         @Test
         void invalidOrder() {
             validateError(
-                    () -> restClient.get().uri("?order=foo").retrieve().toEntity(String.class),
+                    () -> restClient
+                            .get()
+                            .uri("?order=foo", ACCOUNT_ID)
+                            .retrieve()
+                            .toEntity(String.class),
                     HttpClientErrorException.BadRequest.class,
                     "Failed to convert 'order'");
         }
@@ -303,118 +309,13 @@ final class HooksControllerTest extends ControllerTest {
                 })
         void invalidHookIdFormat(String queryString) {
             validateError(
-                    () -> restClient.get().uri(queryString).retrieve().toEntity(String.class),
+                    () -> restClient
+                            .get()
+                            .uri(queryString, ACCOUNT_ID)
+                            .retrieve()
+                            .toEntity(String.class),
                     HttpClientErrorException.BadRequest.class,
                     "Failed to convert 'hook.id'");
-        }
-
-        private void assertHooksOrder(Sort.Direction order, List<Hook> expectedHookOrder) {
-            final var expectedHooks = hookMapper.map(expectedHookOrder);
-            final var expectedResponse = new HooksResponse();
-            expectedResponse.setHooks(expectedHooks);
-            expectedResponse.setLinks(new Links());
-
-            // when
-            final var actual = restClient
-                    .get()
-                    .uri("?order=" + order.name().toLowerCase())
-                    .retrieve()
-                    .body(HooksResponse.class);
-
-            // then
-            assertThat(actual).isNotNull().isEqualTo(expectedResponse);
-        }
-    }
-
-    @DisplayName("/api/v1/accounts/{accountId}/hooks [Account ID Formats]")
-    @Nested
-    final class AccountIdFormatsTest extends EndpointTest {
-
-        private String accountId = "1234";
-
-        @ParameterizedTest
-        @ValueSource(strings = {"1234", "0.1234", "0.0.1234"})
-        void successWithEntityIdAddressFormat(String entityId) {
-            this.accountId = entityId;
-
-            final long ownerId = 1234L;
-            final var hook1 = persistHook(ownerId);
-            final var hook2 = persistHook(ownerId);
-
-            final var expectedHooks = hookMapper.map(List.of(hook2, hook1));
-            final var expectedResponse = new HooksResponse();
-            expectedResponse.setHooks(expectedHooks);
-            expectedResponse.setLinks(new Links());
-
-            final var url = "accounts/" + entityId + "/hooks";
-            final var actual = restClient.get().uri(url).retrieve().body(HooksResponse.class);
-
-            assertThat(actual).isNotNull().isEqualTo(expectedResponse);
-        }
-
-        @ParameterizedTest
-        @ValueSource(
-                strings = {
-                    "HIQQEXWKW53RKN4W6XXC4Q232SYNZ3SZANVZZSUME5B5PRGXL663UAQA",
-                    "0.HIQQEXWKW53RKN4W6XXC4Q232SYNZ3SZANVZZSUME5B5PRGXL663UAQA",
-                    "0.0.HIQQEXWKW53RKN4W6XXC4Q232SYNZ3SZANVZZSUME5B5PRGXL663UAQA"
-                })
-        void successWithAliasAddressFormat(String alias) {
-            this.accountId = alias;
-
-            final var aliasBytes = BaseEncoding.base32()
-                    .omitPadding()
-                    .decode("HIQQEXWKW53RKN4W6XXC4Q232SYNZ3SZANVZZSUME5B5PRGXL663UAQA");
-
-            final var account = domainBuilder
-                    .entity()
-                    .customize(e -> e.type(EntityType.ACCOUNT).alias(aliasBytes))
-                    .persist();
-
-            final var hook1 = persistHook(account.getId());
-            final var hook2 = persistHook(account.getId());
-
-            final var expectedHooks = hookMapper.map(List.of(hook2, hook1));
-            final var expectedResponse = new HooksResponse();
-            expectedResponse.setHooks(expectedHooks);
-            expectedResponse.setLinks(new Links());
-
-            final var url = "accounts/" + alias + "/hooks";
-            final var actual = restClient.get().uri(url).retrieve().body(HooksResponse.class);
-
-            assertThat(actual).isNotNull().isEqualTo(expectedResponse);
-        }
-
-        @ParameterizedTest
-        @ValueSource(
-                strings = {
-                    "ac384c53f03855fa1b3616052f8ba32c6c2a2fec",
-                    "0.ac384c53f03855fa1b3616052f8ba32c6c2a2fec",
-                    "0.0.ac384c53f03855fa1b3616052f8ba32c6c2a2fec",
-                    "0xac384c53f03855fa1b3616052f8ba32c6c2a2fec",
-                })
-        void successWithEvmAddressFormat(String evmAddress) throws DecoderException {
-            this.accountId = evmAddress;
-
-            final var evmBytes = Hex.decodeHex("ac384c53f03855fa1b3616052f8ba32c6c2a2fec");
-
-            final var account = domainBuilder
-                    .entity()
-                    .customize(e -> e.type(EntityType.ACCOUNT).evmAddress(evmBytes))
-                    .persist();
-
-            final var hook1 = persistHook(account.getId());
-            final var hook2 = persistHook(account.getId());
-
-            final var expectedHooks = hookMapper.map(List.of(hook2, hook1));
-            final var expectedResponse = new HooksResponse();
-            expectedResponse.setHooks(expectedHooks);
-            expectedResponse.setLinks(new Links());
-
-            final var url = "accounts/" + evmAddress + "/hooks";
-            final var actual = restClient.get().uri(url).retrieve().body(HooksResponse.class);
-
-            assertThat(actual).isNotNull().isEqualTo(expectedResponse);
         }
 
         @DisplayName("GET /api/v1/accounts/{accountId}/hooks [Invalid Account ID Formats]")
@@ -431,45 +332,51 @@ final class HooksControllerTest extends ControllerTest {
                     "9999999999999999999999999" // overflow
                 })
         void invalidAccountIdFormat(String invalidAccountId) {
-            final var url = "accounts/" + invalidAccountId + "/hooks";
-
             validateError(
-                    () -> restClient.get().uri(url).retrieve().toEntity(String.class),
+                    () -> restClient.get().uri("", invalidAccountId).retrieve().toEntity(String.class),
                     HttpClientErrorException.BadRequest.class,
                     "Failed to convert 'accountId'");
         }
 
-        @Override
-        @Test
-        @Disabled("Not applicable for accountId format tests")
-        void etagHeader() {}
+        private void persistAccount(long accountId, String alias, String evmAddress) {
+            final var aliasBytes = BaseEncoding.base32().omitPadding().decode(alias);
 
-        @Override
-        @Test
-        @Disabled("Not applicable for accountId format tests")
-        void headers() {}
-
-        @Override
-        @Test
-        @Disabled("Not applicable for accountId format tests")
-        void methodNotAllowed() {}
-
-        @Override
-        protected String getUrl() {
-            return "";
-        }
-
-        @Override
-        protected RequestHeadersSpec<?> defaultRequest(RequestHeadersUriSpec<?> uriSpec) {
-            // Persist a hook for the currentAccountId if numeric
-            if (accountId.matches("\\d+(\\.\\d+){0,2}")) {
-                persistHook(Long.parseLong(accountId.replaceAll(".*\\.", "")));
+            final byte[] evmBytes;
+            try {
+                evmBytes = Hex.decodeHex(evmAddress);
+            } catch (DecoderException e) {
+                throw new RuntimeException(e);
             }
-            return uriSpec.uri("");
-        }
-    }
 
-    private Hook persistHook(long ownerId) {
-        return domainBuilder.hook().customize(hook -> hook.ownerId(ownerId)).persist();
+            domainBuilder
+                    .entity()
+                    .customize(e -> e.id(accountId)
+                            .type(EntityType.ACCOUNT)
+                            .alias(aliasBytes)
+                            .evmAddress(evmBytes))
+                    .persist();
+        }
+
+        private static String normalizeAddress(String input) {
+            if (input == null) {
+                return null;
+            }
+
+            if (input.startsWith("0.0.")) {
+                return input.substring(4);
+            } else if (input.startsWith("0.")) {
+                return input.substring(2);
+            }
+
+            if (input.startsWith("0x")) {
+                return input.substring(2);
+            }
+
+            return input;
+        }
+
+        private Hook persistHook(long ownerId) {
+            return domainBuilder.hook().customize(hook -> hook.ownerId(ownerId)).persist();
+        }
     }
 }
--- a/rest-java/src/test/java/org/hiero/mirror/restjava/repository/HookRepositoryTest.java
+++ b/rest-java/src/test/java/org/hiero/mirror/restjava/repository/HookRepositoryTest.java
@@ -15,7 +15,7 @@ import org.springframework.data.domain.Pageable;
 import org.springframework.data.domain.Sort;
 
 @RequiredArgsConstructor
-class HookRepositoryTest extends RestJavaIntegrationTest {
+final class HookRepositoryTest extends RestJavaIntegrationTest {
 
     private final HookRepository hookRepository;
 
--- a/rest-java/src/test/java/org/hiero/mirror/restjava/service/HookServiceTest.java
+++ b/rest-java/src/test/java/org/hiero/mirror/restjava/service/HookServiceTest.java
@@ -18,7 +18,7 @@ import org.junit.jupiter.api.Test;
 import org.springframework.data.domain.Sort;
 
 @RequiredArgsConstructor
-class HookServiceTest extends RestJavaIntegrationTest {
+final class HookServiceTest extends RestJavaIntegrationTest {
 
     private static final long OWNER_ID = 1001L;
 


rest-java/src/main/java/org/hiero/mirror/restjava/controller/HooksController.java
```diff
@@ -0,0 +1,51 @@
...
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.RestController;
+
+@NullMarked
+@RequestMapping("/api/v1/accounts")
+@RequiredArgsConstructor
+@RestController
+public class HooksController {
```
xin-hedera: nit

```suggestion
public final class HooksController {
```

rest-java/src/main/java/org/hiero/mirror/restjava/controller/HooksController.java
```diff
@@ -0,0 +1,51 @@
...
+    private final HookService hookService;
+    private final HookMapper hookMapper;
+
+    @GetMapping("/{accountId}/hooks")
+    public ResponseEntity<HooksResponse> getHooks(
+            @PathVariable long accountId,
```
xin-hedera: The path variable type should be `EntityIdParameter`, since we support all different forms as well as alias / evm address.

rest-java/src/main/java/org/hiero/mirror/restjava/controller/HooksController.java
```diff
@@ -0,0 +1,51 @@
...
+            @RequestParam(name = "hook.id", required = false) String hookId,
```
xin-hedera: at a minimum, `hook.id` should be allowed to appear up to two times in a request, check `AllowancesController.getNftAllowances` for reference

rest-java/src/main/java/org/hiero/mirror/restjava/controller/HooksController.java
```diff
@@ -0,0 +1,51 @@
...
+            @RequestParam(defaultValue = "25") @Min(1) @Max(100) int limit,
```
xin-hedera:
```suggestion
            @RequestParam(defaultValue = DEFAULT_LIMIT) @Positive @Max(MAX_LIMIT) int limit,
```

rest-java/src/main/java/org/hiero/mirror/restjava/controller/HooksController.java
```diff
@@ -0,0 +1,51 @@
...
+            @RequestParam(defaultValue = "desc")
+                    @Pattern(regexp = "^(?i)(asc|desc)$", message = "must be either ''asc'' or ''desc''")
```
xin-hedera:
```suggestion
@RequestParam(defaultValue = "desc") Sort.Direction order,
```

rest-java/src/main/java/org/hiero/mirror/restjava/mapper/CommonMapper.java
```diff
@@ -115,6 +117,19 @@ default TimestampRange mapRange(Range<Long> source) {
         return target;
     }
 
+    @Named(QUALIFIER_TIMESTAMP_RANGE_NULLABLE)
+    default TimestampRangeNullable mapTimestampRangeNullable(Range<Long> source) {
```
xin-hedera, line 119: don't think we need it. the existing method `default TimestampRange mapRange(Range<Long> source)` should handle it.

rest-java/src/main/java/org/hiero/mirror/restjava/service/HookService.java
```diff
@@ -0,0 +1,14 @@
...
+
+package org.hiero.mirror.restjava.service;
+
+import java.util.List;
+// import org.hiero.mirror.rest.model.Hook;
+import org.hiero.mirror.common.domain.hook.Hook;
+
+public interface HookService {
+
+    long getActiveHookCount(long ownerId);
```
xin-hedera: where is this used?

rest-java/src/main/java/org/hiero/mirror/restjava/repository/HookRepository.java
```diff
@@ -0,0 +1,28 @@
...
+
+import java.util.List;
+import org.hiero.mirror.common.domain.hook.AbstractHook.Id;
+import org.hiero.mirror.common.domain.hook.Hook;
+import org.springframework.data.domain.Pageable;
+import org.springframework.data.repository.PagingAndSortingRepository;
+
+public interface HookRepository extends PagingAndSortingRepository<Hook, Id> {
+
+    long countByOwnerId(long ownerId);
```
xin-hedera: is this needed?

rest-java/src/main/java/org/hiero/mirror/restjava/repository/HookRepository.java
```diff
@@ -0,0 +1,28 @@
...
+
+    List<Hook> findByOwnerIdOrderByHookIdAsc(long ownerId, Pageable pageable);
```
xin-hedera: I don't think we should use pageable for any of the queries, it'll use `offset` clause and will cause bad performance.

rest-java/src/main/java/org/hiero/mirror/restjava/mapper/HookMapper.java
```diff
@@ -0,0 +1,59 @@
...
+    @Mapping(source = "createdTimestamp", target = "createdTimestamp", qualifiedByName = "formatTimestamp")
+    @Mapping(source = "deleted", target = "deleted")
+    @Mapping(source = "extensionPoint", target = "extensionPoint")
+    @Mapping(source = "hookId", target = "hookId")
+    @Mapping(source = "ownerId", target = "ownerId", qualifiedByName = "formatEntityId")
+    @Mapping(source = "type", target = "type")
+    @Mapping(source = "timestampRange", target = "timestampRange", qualifiedByName = QUALIFIER_TIMESTAMP_RANGE_NULLABLE)
+    org.hiero.mirror.rest.model.Hook mapHook(Hook source);
+
+    @Named("formatEntityId")
```
xin-hedera: I'm not sure why we need these methods, most of them should be supported by `CommonMapper`

xin-hedera (changes requested):

rest-java/src/main/java/org/hiero/mirror/restjava/controller/HooksController.java
steven-sheehy: Can also remove public

rest-java/src/main/java/org/hiero/mirror/restjava/repository/HookRepository.java
steven-sheehy: I think we'll probably need a dynamic jooq query to support arbitrary hook.id params like multiple eq params.

rest-java/src/main/java/org/hiero/mirror/restjava/repository/HookRepository.java
```diff
@@ -0,0 +1,28 @@
...
+
+    List<Hook> findByOwnerIdAndHookIdLessThanOrderByHookIdAsc(long ownerId, long hookId, Pageable pageable);
+
+    List<Hook> findByOwnerIdAndHookIdLessThanOrderByHookIdDesc(long ownerId, long hookId, Pageable pageable);
+
+    List<Hook> findByOwnerIdAndHookIdGreaterThanOrderByHookIdAsc(long ownerId, long hookId, Pageable pageable);
+
+    List<Hook> findByOwnerIdAndHookIdGreaterThanOrderByHookIdDesc(long ownerId, long hookId, Pageable pageable);
+
+    Hook findByOwnerIdAndHookId(long ownerId, long hookId);
```
steven-sheehy: We should remove all of these methods. We only need two repository methods:

```java
    List<Hook> findByOwnerIdAndHookIdGreaterThanEqualAndHookIdLessThanEqual(long ownerId, long hookIdLower, long hookIdUpper, Pageable pageable);

    List<Hook> findByOwnerIdAndHookIdIn(long ownerId, Collection<Long> hookIds, Pageable pageable);
```

The pageable should auto generate the limit and order by. If it includes an offset keyword in the SQL it shouldn't matter since the full primary key is specified via the other params. Fundamentally, the REST API query params for this particular API will result in 4 scenarios so we can get away with using just the two static queries above:

* One or more `eq` only params only
  * Collect all hook Ids and use `findByOwnerIdAndHookIdIn()`
* One or more `lt|lte|gt|gte` query params only
  * Normalize lt and gt to lte and gte by adding/subtracting one nanosecond
  * Pick min lt(e) and max gt(e)
  * Default missing lower bound to 0 and missing upper bound to MAX_VALUE
  * Pass resulting single lower and single upper bound to `findByOwnerIdAndHookIdGreaterThanEqualAndHookIdLessThanEqual`
* Both eq and range params without overlap
  * Skip repository and return empty list
* Both eq and range params with overlap in range
  * Collect all hook Ids and use `findByOwnerIdAndHookIdIn()`

rest-java/src/main/java/org/hiero/mirror/restjava/controller/HooksController.java
```diff
@@ -0,0 +1,51 @@
...
+@NullMarked
+@RequestMapping("/api/v1/accounts")
+@RequiredArgsConstructor
+@RestController
+public class HooksController {
+    private final HookService hookService;
+    private final HookMapper hookMapper;
+
+    @GetMapping("/{accountId}/hooks")
+    public ResponseEntity<HooksResponse> getHooks(
```
steven-sheehy: Can be package private

rest-java/src/main/java/org/hiero/mirror/restjava/controller/HooksController.java
```diff
@@ -0,0 +1,51 @@
...
+
+        // Build next link only if descending and there are results
+        if ("desc".equalsIgnoreCase(order) && hooks.size() == limit) {
+            final long lastHookId = hooks.getLast().getHookId();
+            final var links = new Links();
+            links.setNext(
+                    String.format("/api/v1/accounts/%d/hooks?hook.id=lt:%d&limit=%d", accountId, lastHookId, limit));
+
+            hooksResponse.setLinks(links);
+        }
```
steven-sheehy: We should be using the LinkFactory. And we'll need links regardless of desc or asc.

rest-java/src/main/java/org/hiero/mirror/restjava/controller/HooksController.java
steven-sheehy: We should support more than 2. See my other comment. There should still be a max but it should match our max query param that we allow in js.

rest-java/src/main/java/org/hiero/mirror/restjava/controller/HooksController.java
```diff
@@ -0,0 +1,51 @@
...
+    @GetMapping("/{accountId}/hooks")
+    public ResponseEntity<HooksResponse> getHooks(
+            @PathVariable long accountId,
+            @RequestParam(name = "hook.id", required = false) String hookId,
+            @RequestParam(defaultValue = "25") @Min(1) @Max(100) int limit,
+            @RequestParam(defaultValue = "desc")
+                    @Pattern(regexp = "^(?i)(asc|desc)$", message = "must be either ''asc'' or ''desc''")
+                    String order) {
+        final var hooks = hookService.getAllHooksByOwner(accountId, hookId, limit, order);
+        final var hooksResponse = hookMapper.mapToHooksResponse(hooks);
```
steven-sheehy: Should be consistent with the other mappers.

```suggestion
        final var hooksResponse = hookMapper.map(hooks);
```

rest-java/src/main/java/org/hiero/mirror/restjava/controller/HooksController.java
steven-sheehy: This should be a `NumberRangeParameter[]`. We should not parse strings in the service layer.

rest-java/src/main/java/org/hiero/mirror/restjava/mapper/HookMapper.java
```diff
@@ -0,0 +1,59 @@
...
+public interface HookMapper {
+    @Mapping(source = "adminKey", target = "adminKey", qualifiedByName = "mapAdminKey")
+    @Mapping(source = "contractId", target = "contractId", qualifiedByName = "formatEntityId")
+    @Mapping(source = "createdTimestamp", target = "createdTimestamp", qualifiedByName = "formatTimestamp")
+    @Mapping(source = "deleted", target = "deleted")
+    @Mapping(source = "extensionPoint", target = "extensionPoint")
+    @Mapping(source = "hookId", target = "hookId")
+    @Mapping(source = "ownerId", target = "ownerId", qualifiedByName = "formatEntityId")
+    @Mapping(source = "type", target = "type")
+    @Mapping(source = "timestampRange", target = "timestampRange", qualifiedByName = QUALIFIER_TIMESTAMP_RANGE_NULLABLE)
```
steven-sheehy: We should not need most of these manual mappings. All the field names line up so as long as there's a converter from source type to target type it should work as is. Remove them and see what code the `HookMapperImpl` generates and only map what's truly necessary.

rest-java/src/main/java/org/hiero/mirror/restjava/mapper/HookMapper.java
```diff
@@ -0,0 +1,59 @@
...
+        StringBuilder sb = new StringBuilder("0x");
+        for (byte b : bytes) sb.append(String.format("%02x", b));
+        return sb.toString();
+    }
+
+    default HooksResponse mapToHooksResponse(List<Hook> hooks) {
+        HooksResponse response = new HooksResponse();
+        response.setHooks(hooks.stream().map(this::mapHook).toList());
+        return response;
+    }
```
steven-sheehy: We can probably leverage `CollectionMapper` like `TokenAirdropMapper` does to avoid this method.

rest-java/src/main/java/org/hiero/mirror/restjava/service/HookService.java
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package org.hiero.mirror.restjava.service;
+
+import java.util.List;
+// import org.hiero.mirror.rest.model.Hook;
```
steven-sheehy: Commented out import

rest-java/src/main/java/org/hiero/mirror/restjava/service/HookService.java
```diff
@@ -0,0 +1,14 @@
...
+import org.hiero.mirror.common.domain.hook.Hook;
+
+public interface HookService {
+
+    long getActiveHookCount(long ownerId);
+
+    List<Hook> getAllHooksByOwner(long ownerId, String hookIdFilter, int limit, String order);
```
steven-sheehy: Prefer a DTO builder object like `NftAllowanceRequest` be created to capture all params. Use higher level types like Collection, EntityId, NumberRangeParameter, Sort.Direction.

```suggestion
    Collection<Hook> getHooks(HookRequest request);
```

rest/api/v1/openapi.yml
```diff
@@ -53,6 +53,28 @@ paths:
           $ref: "#/components/responses/NotFoundError"
       tags:
         - accounts
+  /api/v1/accounts/{idOrAliasOrEvmAddress}/hooks:
+    get:
+      summary: List hooks for an account
+      description: Returns a list of hooks associated with a given account ID, alias, or EVM address.
+      operationId: getAccountHooks
```
steven-sheehy:
```suggestion
      operationId: getHooks
```

rest/api/v1/openapi.yml
```diff
@@ -53,6 +53,28 @@ paths:
...
+      parameters:
+        - $ref: "#/components/parameters/accountIdOrAliasOrEvmAddressPathParam"
+        - $ref: "#/components/parameters/limitQueryParam"
+        - $ref: "#/components/parameters/orderQueryParam"
```
steven-sheehy:
```suggestion
        - $ref: "#/components/parameters/orderQueryParamDesc"
```

rest/api/v1/openapi.yml
```diff
@@ -53,6 +53,28 @@ paths:
...
       tags:
         - accounts
+  /api/v1/accounts/{idOrAliasOrEvmAddress}/hooks:
+    get:
+      summary: List hooks for an account
+      description: Returns a list of hooks associated with a given account ID, alias, or EVM address.
+      operationId: getAccountHooks
+      parameters:
+        - $ref: "#/components/parameters/accountIdOrAliasOrEvmAddressPathParam"
+        - $ref: "#/components/parameters/limitQueryParam"
```
steven-sheehy, line 64: Missing hook.id param

rest/api/v1/openapi.yml
```diff
@@ -53,6 +53,28 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/HooksResponse" # Reference the new response schema
+        "400":
+          $ref: "#/components/responses/InvalidParameterError"
+        "404":
+          $ref: "#/components/responses/NotFoundError"
+      tags:
+        - hooks # Keep it grouped with other hooks endpoints
```
steven-sheehy:
```suggestion
        - accounts
```

rest/api/v1/openapi.yml
```diff
@@ -53,6 +53,28 @@ paths:
...
+        - $ref: "#/components/parameters/accountIdOrAliasOrEvmAddressPathParam"
+        - $ref: "#/components/parameters/limitQueryParam"
+        - $ref: "#/components/parameters/orderQueryParam"
+      responses:
+        "200":
+          description: OK
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/HooksResponse" # Reference the new response schema
```
steven-sheehy: Comment isn't really necessary

rest/api/v1/openapi.yml
```diff
@@ -1550,6 +1572,61 @@ components:
...
+        hooks:
+          type: array
+          items:
+            $ref: "#/components/schemas/Hook"
+        links:
+          $ref: "#/components/schemas/Links"
+    Hook:
+      type: object
+      required:
+        - contract_id
```
steven-sheehy, line 1592: Missing admin_key, created_timestamp. Required is for presence of key, not value.

rest/api/v1/openapi.yml
```diff
@@ -1550,6 +1572,61 @@ components:
...
+          example: 1
+        owner_id:
+          description: The entity that owns the hook
+          allOf:
+            - $ref: "#/components/schemas/EntityId"
+        timestamp_range:
+          $ref: "#/components/schemas/TimestampRangeNullable"
+        type:
+          description: The type of the hook implementation
+          type: string
```
steven-sheehy, line 1635: It's an enum, not a string

rest/api/v1/openapi.yml
```diff
@@ -1550,6 +1572,61 @@ components:
...
+          description: The consensus timestamp when the hook was created
+          allOf:
+            - $ref: "#/components/schemas/TimestampNullable"
+        deleted:
+          description: Whether the hook has been deleted
+          type: boolean
+          example: false
+        extension_point:
+          description: The extension point this hook implements
+          type: string
```
steven-sheehy, line 1619: enum

rest/api/v1/openapi.yml
```diff
@@ -1550,6 +1572,61 @@ components:
...
+          example: "ACCOUNT_ALLOWANCE_HOOK"
+        hook_id:
+          description: The unique identifier for the hook within the owner's scope
+          format: int64
+          type: integer
+          example: 1
```
steven-sheehy: nit: prefer all fields sorted by key here and elsewhere in file.

rest/api/v1/openapi.yml
```diff
@@ -1550,6 +1572,61 @@ components:
...
+        owner_id:
+          description: The entity that owns the hook
+          allOf:
+            - $ref: "#/components/schemas/EntityId"
+        timestamp_range:
+          $ref: "#/components/schemas/TimestampRangeNullable"
```
steven-sheehy, line 1630: This is not in the response in the doc. Though maybe we should have it now since we have reusable hooks? @ashumahajan

rest-java/src/test/java/org/hiero/mirror/restjava/controller/HookControllerTest.java
```diff
@@ -0,0 +1,205 @@
...
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
+import org.springframework.web.client.HttpClientErrorException;
+import org.springframework.web.client.RestClient.RequestHeadersSpec;
+import org.springframework.web.client.RestClient.RequestHeadersUriSpec;
+
+@RequiredArgsConstructor
+public class HookControllerTest extends ControllerTest {
```
steven-sheehy:
```suggestion
final class HookControllerTest extends ControllerTest {
```

rest-java/src/main/java/org/hiero/mirror/restjava/service/HookServiceImpl.java
```diff
@@ -0,0 +1,85 @@
...
+import lombok.RequiredArgsConstructor;
+import org.hiero.mirror.common.domain.hook.Hook;
+import org.hiero.mirror.restjava.repository.HookRepository;
+import org.jspecify.annotations.NullMarked;
+import org.springframework.data.domain.PageRequest;
+
+@Named
+@NullMarked
+@RequiredArgsConstructor
+public class HookServiceImpl implements HookService {
```
steven-sheehy:
```suggestion
final class HookServiceImpl implements HookService {
```

steven-sheehy: Needs a test

rest-java/src/main/java/org/hiero/mirror/restjava/repository/HookRepository.java
```diff
@@ -0,0 +1,28 @@
...
+
+package org.hiero.mirror.restjava.repository;
+
+import java.util.List;
+import org.hiero.mirror.common.domain.hook.AbstractHook.Id;
+import org.hiero.mirror.common.domain.hook.Hook;
+import org.springframework.data.domain.Pageable;
+import org.springframework.data.repository.PagingAndSortingRepository;
+
+public interface HookRepository extends PagingAndSortingRepository<Hook, Id> {
```
steven-sheehy, line 12: Needs a test

rest-java/src/main/java/org/hiero/mirror/restjava/mapper/HookMapper.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.List;
+import org.hiero.mirror.common.domain.hook.Hook;
+import org.hiero.mirror.rest.model.HooksResponse;
+import org.hiero.mirror.rest.model.Key;
+import org.mapstruct.Mapper;
+import org.mapstruct.Mapping;
+import org.mapstruct.Named;
+
+@Mapper(config = MapperConfiguration.class, uses = CommonMapper.class)
+public interface HookMapper {
```
steven-sheehy: Needs a test

rest-java/src/test/java/org/hiero/mirror/restjava/controller/HooksControllerTest.java
```diff
@@ -0,0 +1,205 @@
...
+        @Override
+        protected RequestHeadersSpec<?> defaultRequest(RequestHeadersUriSpec<?> uriSpec) {
+            // Persist at least one hook for the default request
+            storeHook(accountId);
+            return uriSpec.uri("");
+        }
+
+        @Test
+        void success() {
+            final var hook1 = storeHook(accountId);
```
steven-sheehy: Use `// given|when|then` section comments.

rest-java/src/main/java/org/hiero/mirror/restjava/service/HookServiceImpl.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    /**
+     * Retrieves all hooks (active or deleted) of a given owner.
+     *
+     * @param ownerId The ID of the owner account.
+     * @return List of all hooks for the owner.
+     */
+    @Override
+    public List<Hook> getAllHooksByOwner(long ownerId, String hookIdFilter, int limit, String order) {
```
steven-sheehy: We need to convert `EntityIdParameter` to EntityId by using `entityService.lookup` to support alias and EVM address lookups. See `TokenAirdropServiceImpl`

rest-java/src/main/java/org/hiero/mirror/restjava/controller/HooksController.java
sdimitrov9: Done

rest-java/src/main/java/org/hiero/mirror/restjava/controller/HooksController.java
sdimitrov9: Done

rest/api/v1/openapi.yml
sdimitrov9: Added

rest/api/v1/openapi.yml
sdimitrov9: Added

rest/api/v1/openapi.yml
sdimitrov9: Fixed

rest/api/v1/openapi.yml
sdimitrov9: Fixed

rest-java/src/main/java/org/hiero/mirror/restjava/controller/HooksController.java
sdimitrov9: Modified

rest-java/src/main/java/org/hiero/mirror/restjava/mapper/CommonMapper.java
sdimitrov9: Removed

**Edit**: After @ashumahajan updated the specs by including the timestamp_range in the response we will need mapping logic for the field and unfortunately `default TimestampRange mapRange(Range<Long> source)` does not seem to handle it  since we are getting Compilation error: `Unmapped target properties: "from, to". Mapping from property "Range<Long> timestampRange" to "TimestampRangeNullable timestampRange"` so we will have to get the above method back.

rest-java/src/main/java/org/hiero/mirror/restjava/service/HookService.java
sdimitrov9: Removed

rest-java/src/main/java/org/hiero/mirror/restjava/repository/HookRepository.java
sdimitrov9: Not for this ticket, removed.

rest-java/src/main/java/org/hiero/mirror/restjava/mapper/HookMapper.java
sdimitrov9: Removed

rest-java/src/main/java/org/hiero/mirror/restjava/controller/HooksController.java
sdimitrov9: Addressed

rest-java/src/main/java/org/hiero/mirror/restjava/controller/HooksController.java
sdimitrov9: Fixed

rest-java/src/main/java/org/hiero/mirror/restjava/repository/HookRepository.java
sdimitrov9: Fixed

rest-java/src/main/java/org/hiero/mirror/restjava/controller/HooksController.java
sdimitrov9: Done

rest-java/src/main/java/org/hiero/mirror/restjava/controller/HooksController.java
sdimitrov9: Fixed

rest-java/src/main/java/org/hiero/mirror/restjava/controller/HooksController.java
sdimitrov9: Fixed

rest-java/src/main/java/org/hiero/mirror/restjava/mapper/HookMapper.java
sdimitrov9: Addressed

rest-java/src/main/java/org/hiero/mirror/restjava/mapper/HookMapper.java
sdimitrov9: Addressed

rest-java/src/main/java/org/hiero/mirror/restjava/service/HookService.java
sdimitrov9: Removed

rest-java/src/main/java/org/hiero/mirror/restjava/service/HookService.java
sdimitrov9: Addressed

rest/api/v1/openapi.yml
sdimitrov9: Addressed

rest/api/v1/openapi.yml
sdimitrov9: Fixed

rest/api/v1/openapi.yml
sdimitrov9: Fixed

rest/api/v1/openapi.yml
sdimitrov9: Removed

rest/api/v1/openapi.yml
sdimitrov9: Done

rest/api/v1/openapi.yml
sdimitrov9: Removed, until spec is changed.

rest-java/src/test/java/org/hiero/mirror/restjava/controller/HookControllerTest.java
sdimitrov9: Done

rest-java/src/main/java/org/hiero/mirror/restjava/service/HookServiceImpl.java
sdimitrov9: Changed

rest-java/src/main/java/org/hiero/mirror/restjava/service/HookServiceImpl.java
sdimitrov9: Done

rest/api/v1/openapi.yml
ashumahajan: Will update the spec.

rest-java/src/main/java/org/hiero/mirror/restjava/controller/HooksController.java
sdimitrov9: After using @Size(max = MAX_REPEATED_QUERY_PARAMETERS) NumberRangeParameter[] hookIds, we can now pass up to 100 repeated params, which I think is the value used in the js.

rest-java/src/test/java/org/hiero/mirror/restjava/controller/HooksControllerTest.java
sdimitrov9: Done

rest-java/src/main/java/org/hiero/mirror/restjava/service/HookServiceImpl.java
sdimitrov9: Added

rest-java/src/main/java/org/hiero/mirror/restjava/repository/HookRepository.java
sdimitrov9: Added

rest-java/src/main/java/org/hiero/mirror/restjava/mapper/HookMapper.java
sdimitrov9: Added

rest-java/src/main/java/org/hiero/mirror/restjava/repository/HookRepository.java
sdimitrov9: Will be addressed within this ticket: https://github.com/hiero-ledger/hiero-mirror-node/issues/12292

rest-java/src/main/java/org/hiero/mirror/restjava/controller/HooksController.java
```diff
@@ -0,0 +1,79 @@
...
+    private final HookMapper hookMapper;
+    private final LinkFactory linkFactory;
+
+    private static final Function<Hook, Map<String, String>> EXTRACTOR =
+            hook -> ImmutableSortedMap.of(HOOK_ID, hook.getHookId().toString());
+
+    @GetMapping("/{accountId}/hooks")
+    ResponseEntity<HooksResponse> getHooks(
+            @PathVariable EntityIdParameter accountId,
+            @RequestParam(name = "hook.id", required = false) @Size(max = MAX_REPEATED_QUERY_PARAMETERS)
```
steven-sheehy:
```suggestion
            @RequestParam(name = HOOK_ID, required = false) @Size(max = MAX_REPEATED_QUERY_PARAMETERS)
```

rest-java/src/main/java/org/hiero/mirror/restjava/controller/HooksController.java
```diff
@@ -0,0 +1,79 @@
...
+                .limit(limit)
+                .order(order)
+                .build();
+        final var hooksServiceResponse = hookService.getHooks(hooksRequest);
+        final var hooks = hookMapper.map(hooksServiceResponse);
+
+        final var hooksResponse = new HooksResponse();
+        hooksResponse.setHooks(hooks);
+
+        if (!hooks.isEmpty() && hooks.size() == limit) {
```
steven-sheehy: The link factory should already handle this check, so can remove if statement. Just always call linkfactory and set the links. We still want a `links: {}` even if no next link instead of `null`.

rest-java/src/main/java/org/hiero/mirror/restjava/repository/HookRepositoryCustomImpl.java
```diff
@@ -0,0 +1,44 @@
...
+import org.hiero.mirror.common.domain.hook.Hook;
+import org.hiero.mirror.restjava.dto.HooksRequest;
+import org.jooq.Condition;
+import org.jooq.DSLContext;
+import org.jooq.Field;
+import org.jooq.SortOrder;
+
+@Named
+@RequiredArgsConstructor
+public class HookRepositoryCustomImpl implements HookRepositoryCustom {
```
steven-sheehy:
```suggestion
final class HookRepositoryCustomImpl implements HookRepositoryCustom {
```

rest-java/src/main/java/org/hiero/mirror/restjava/repository/HookRepositoryCustomImpl.java
```diff
@@ -0,0 +1,44 @@
...
+import org.hiero.mirror.common.domain.entity.EntityId;
+import org.hiero.mirror.common.domain.hook.Hook;
+import org.hiero.mirror.restjava.dto.HooksRequest;
+import org.jooq.Condition;
+import org.jooq.DSLContext;
+import org.jooq.Field;
+import org.jooq.SortOrder;
+
+@Named
+@RequiredArgsConstructor
```
steven-sheehy: `@NullMarked`

rest-java/src/main/java/org/hiero/mirror/restjava/repository/HookRepositoryCustomImpl.java
```diff
@@ -0,0 +1,44 @@
...
+public class HookRepositoryCustomImpl implements HookRepositoryCustom {
+
+    private final DSLContext dsl;
+
+    @Override
+    public Collection<Hook> findAll(HooksRequest request, EntityId accountId) {
+
+        final var hookTable = HOOK;
+
+        final var ownerIdCondition = hookTable.OWNER_ID.eq(accountId.getNum());
```
steven-sheehy:
```suggestion
        final var ownerIdCondition = hookTable.OWNER_ID.eq(accountId.getId());
```

rest-java/src/main/java/org/hiero/mirror/restjava/repository/HookRepositoryCustomImpl.java
```diff
@@ -0,0 +1,44 @@
...
+        return dsl.selectFrom(hookTable)
+                .where(ownerIdCondition.and(hookIdCondition))
+                .orderBy(orderBy)
+                .limit(request.getLimit())
+                .fetchInto(Hook.class);
+    }
+
+    private Condition getBaseCondition(EntityId accountId, Field<Long> baseField) {
+        return getCondition(baseField, EQ, accountId.getId());
+    }
```
steven-sheehy: Unused method?

rest-java/src/main/java/org/hiero/mirror/restjava/repository/HookRepositoryCustom.java
```diff
@@ -0,0 +1,15 @@
...
+
+import jakarta.validation.constraints.NotNull;
+import java.util.Collection;
+import org.hiero.mirror.common.domain.entity.EntityId;
+import org.hiero.mirror.common.domain.hook.Hook;
+import org.hiero.mirror.restjava.dto.HooksRequest;
+
+public interface HookRepositoryCustom extends JooqRepository {
+
+    @NotNull
```
steven-sheehy: Remove

rest-java/src/main/java/org/hiero/mirror/restjava/mapper/CommonMapper.java
```diff
@@ -115,6 +116,18 @@ default TimestampRange mapRange(Range<Long> source) {
...
 
+    default TimestampRangeNullable mapTimestampRangeNullable(Range<Long> source) {
+        final var timestampRange = mapRange(source);
+        if (timestampRange == null) {
+            return null;
+        }
+
+        final var target = new TimestampRangeNullable();
+        target.setFrom(timestampRange.getFrom());
+        target.setTo(timestampRange.getTo());
```
steven-sheehy: We need to check hasLowerBound and hasUpperBound like the other mapRange. Add a test in CommonMapperTest like the existing.

rest-java/src/main/java/org/hiero/mirror/restjava/dto/HooksRequest.java
```diff
@@ -0,0 +1,30 @@
...
+import lombok.Data;
+import org.hiero.mirror.restjava.common.EntityIdParameter;
+import org.hiero.mirror.restjava.service.Bound;
+import org.springframework.data.domain.Sort;
+import org.springframework.data.domain.Sort.Direction;
+
+@Data
+@Builder
+public class HooksRequest {
+    private EntityIdParameter accountId;
```
steven-sheehy:
```suggestion
    private EntityIdParameter ownerId;
```

rest-java/src/main/java/org/hiero/mirror/restjava/dto/HooksRequest.java
```diff
@@ -0,0 +1,30 @@
...
+
+import java.util.List;
+import lombok.Builder;
+import lombok.Data;
+import org.hiero.mirror.restjava.common.EntityIdParameter;
+import org.hiero.mirror.restjava.service.Bound;
+import org.springframework.data.domain.Sort;
+import org.springframework.data.domain.Sort.Direction;
+
+@Data
```
steven-sheehy: Change to `@Value`

rest-java/src/main/java/org/hiero/mirror/restjava/common/Constants.java
```diff
@@ -8,12 +8,14 @@
 public class Constants {
 
     public static final String ACCOUNT_ID = "account.id";
+    public static final String DEFAULT_LIMIT = "25";
```
steven-sheehy: Should move back to other section. Maybe add comments that first group is parameter names and second is parameter constraints and defaults so it's clear the grouping.

rest-java/src/test/java/org/hiero/mirror/restjava/controller/HooksControllerTest.java
```diff
@@ -0,0 +1,194 @@
...
+
+        @Override
+        protected RequestHeadersSpec<?> defaultRequest(RequestHeadersUriSpec<?> uriSpec) {
+            // given: at least one persisted hook for default request
+            storeHook(accountId);
+            return uriSpec.uri("");
+        }
+
+        @Test
+        void success() {
```
steven-sheehy: We should have a parameterized test for different owner id formats: `shard.realm.num`, `realm.num`, `num`, `alias`, `0.0.alias`, `0.alias`, `0x00...123`, `00..123`, `0x123`, `0.0.00..123`, `0.00..123`

rest-java/src/test/java/org/hiero/mirror/restjava/controller/HooksControllerTest.java
```diff
@@ -0,0 +1,194 @@
...
+
+        @ParameterizedTest
+        @CsvSource(
+                delimiter = '|',
+                value = {
+                    "'?hook.id=lt:{2}' | 1,0", // lt hook2's ID -> (hook1, hook0)
+                    "'?hook.id=lte:{2}'| 2,1,0", // lte hook2's ID -> (hook2, hook1, hook0)
+                    "'?hook.id=gt:{1}' | 3,2", // gt hook1's ID -> (hook3, hook2)
+                    "'?hook.id=gte:{1}'| 3,2,1", // gte hook1's ID -> (hook3, hook2, hook1)
+                    "'?hook.id=eq:{1}' | 1", // eq hook1's ID -> (hook1)
```
steven-sheehy: We should have many more tests for no range operator (`hook.id={1}`), multiple eq, multiple no range operators, multiple lt and ltes, multiple gt and gtes, and combinations of lt(e) + gt(e) + eq/no_operator. See also https://github.com/hiero-ledger/hiero-mirror-node/pull/12259#discussion_r2456801615 for a list of some scenarios.

rest-java/src/test/java/org/hiero/mirror/restjava/controller/HooksControllerTest.java
```diff
@@ -0,0 +1,194 @@
...
+            final var expectedResponse = new HooksResponse();
+            expectedResponse.setHooks(expectedHooks);
+            final var links = new Links();
+            links.setNext(String.format(
+                    "/api/v1/accounts/%d/hooks?limit=%d&hook.id=lt:%d", accountId, limit, hook2.getHookId()));
+            expectedResponse.setLinks(links);
+
+            // when
+            final var actual =
+                    restClient.get().uri("?limit=" + limit).retrieve().body(HooksResponse.class);
```
steven-sheehy: Add `&hook.id>=0` just to ensure it's passed through.

rest-java/src/test/java/org/hiero/mirror/restjava/controller/HooksControllerTest.java
```diff
@@ -0,0 +1,194 @@
...
+
+            // when
+            final var actual = restClient.get().uri("").retrieve().body(HooksResponse.class);
+
+            // then
+            assertThat(actual).isNotNull().isEqualTo(expectedResponse);
+        }
+
+        @Test
+        void orderAsc() {
```
steven-sheehy: Should also test non-empty next link for ascending. Either add a separate or do as part of this test.

rest-java/src/test/java/org/hiero/mirror/restjava/controller/HooksControllerTest.java
```diff
@@ -0,0 +1,194 @@
...
+                    hooks.get(2).getHookId(),
+                    hooks.get(3).getHookId());
+            final var actual = restClient.get().uri(formattedParams).retrieve().body(HooksResponse.class);
+
+            // then
+            assertThat(actual).isNotNull().isEqualTo(expectedResponse);
+        }
+
+        @Test
+        void invalidLimitTooLow() {
```
steven-sheehy, line 254: Add parameterized tests for different types of invalid owner id, invalid hook id, and exceeding max query hook.id params.

steven-sheehy (changes requested):

rest-java/src/main/java/org/hiero/mirror/restjava/repository/HookRepositoryCustomImpl.java
sdimitrov9: Done

rest-java/src/main/java/org/hiero/mirror/restjava/repository/HookRepositoryCustomImpl.java
sdimitrov9: Done

rest-java/src/main/java/org/hiero/mirror/restjava/repository/HookRepositoryCustomImpl.java
sdimitrov9: Done

rest-java/src/main/java/org/hiero/mirror/restjava/controller/HooksController.java
sdimitrov9: Done

rest-java/src/main/java/org/hiero/mirror/restjava/controller/HooksController.java
sdimitrov9: Done

rest-java/src/test/java/org/hiero/mirror/restjava/controller/HooksControllerTest.java
sdimitrov9: Done

rest-java/src/test/java/org/hiero/mirror/restjava/controller/HooksControllerTest.java
sdimitrov9: Done

rest-java/src/test/java/org/hiero/mirror/restjava/controller/HooksControllerTest.java
sdimitrov9: Done

rest-java/src/main/java/org/hiero/mirror/restjava/repository/HookRepositoryCustomImpl.java
sdimitrov9: Removed

rest-java/src/main/java/org/hiero/mirror/restjava/repository/HookRepositoryCustom.java
sdimitrov9: Whole interface is removed now as not needed.

rest-java/src/main/java/org/hiero/mirror/restjava/mapper/CommonMapper.java
sdimitrov9: Done

rest-java/src/main/java/org/hiero/mirror/restjava/dto/HooksRequest.java
sdimitrov9: Modified

rest-java/src/main/java/org/hiero/mirror/restjava/dto/HooksRequest.java
sdimitrov9: Done

rest-java/src/main/java/org/hiero/mirror/restjava/common/Constants.java
sdimitrov9: Done

rest-java/src/test/java/org/hiero/mirror/restjava/controller/HooksControllerTest.java
sdimitrov9: Done

rest-java/src/test/java/org/hiero/mirror/restjava/controller/HooksControllerTest.java
steven-sheehy: I don't see this test. Has it been pushed yet?

rest-java/src/main/java/org/hiero/mirror/restjava/controller/HooksController.java
```diff
@@ -0,0 +1,132 @@
...
+                    .ownerId(accountId)
+                    .hookIdEqualsFilters(Collections.emptyList())
+                    .hookIdLowerBoundInclusive(null)
+                    .hookIdUpperBoundInclusive(null)
+                    .limit(limit)
+                    .order(order)
+                    .build();
+        }
+
+        Collection<Long> equalHookIds = new ArrayList<>();
```
steven-sheehy: Collapse duplicates.

```suggestion
        Collection<Long> equalHookIds = new TreeSet<>();
```

rest-java/src/main/java/org/hiero/mirror/restjava/controller/HooksController.java
```diff
@@ -0,0 +1,132 @@
...
+        Long gte = null; // The most restrictive lower bound (max of all gt/gte)
+        Long lte = null; // The most restrictive upper bound (min of all lt/lte)
```
steven-sheehy: We should avoid null if we have a default that works with the logic.
```suggestion
        long lowerBound = 0L; // The most restrictive lower bound (max of all gt/gte)
        long upperBound = Long.MAX_VALUE; // The most restrictive upper bound (min of all lt/lte)
```

rest-java/src/main/java/org/hiero/mirror/restjava/controller/HooksController.java
```diff
@@ -0,0 +1,132 @@
...
+
+        // Apply defaults if a partial range is given
+        if (gte != null || lte != null) {
+            if (gte == null) {
+                gte = 0L;
+            }
+            if (lte == null) {
+                lte = Long.MAX_VALUE;
+            }
+        }
```
steven-sheehy: Can delete with other suggestion.

rest-java/src/main/java/org/hiero/mirror/restjava/controller/HooksController.java
```diff
@@ -0,0 +1,132 @@
...
+                long inclusiveValue = (operator == RangeOperator.GT) ? param.value() + 1 : param.value();
+                if (gte == null || inclusiveValue > gte) {
+                    gte = inclusiveValue;
+                }
+
+            } else if (param.hasUpperBound()) { // lt, lte
+                long inclusiveValue = (operator == RangeOperator.LT) ? param.value() - 1 : param.value();
+                if (lte == null || inclusiveValue < lte) {
+                    lte = inclusiveValue;
+                }
```
steven-sheehy: We can simplify with our default value and adding a new method to `NumberRangeParameter`:

```java
    long getInclusiveValue() {
        if (operator == RangeOperator.GT) {
            return value + 1;
        } else if (operator == RangeOperator.LT) {
            return value - 1;
        } else {
            return value;
        }
    }
```

```suggestion
            upperBound = Math.min(upperBound, param.getInclusiveValue());
```

Same concept for lower bound.

rest-java/src/main/java/org/hiero/mirror/restjava/controller/HooksController.java
```diff
@@ -0,0 +1,132 @@
...
+        if (hookIdFilters == null || hookIdFilters.isEmpty()) {
+            return HooksRequest.builder()
+                    .ownerId(accountId)
+                    .hookIdEqualsFilters(Collections.emptyList())
+                    .hookIdLowerBoundInclusive(null)
+                    .hookIdUpperBoundInclusive(null)
+                    .limit(limit)
+                    .order(order)
+                    .build();
+        }
```
steven-sheehy: We can omit this if we just default hookIdFilters to empty list if null and let the rest of the code execute.

rest-java/src/main/java/org/hiero/mirror/restjava/service/HookServiceImpl.java
```diff
@@ -0,0 +1,63 @@
...
+    public Collection<Hook> getHooks(HooksRequest request) {
+        final var sort = Sort.by(request.getOrder(), "hookId");
+        final var page = PageRequest.of(0, request.getLimit(), sort);
+
+        final var id = entityService.lookup(request.getOwnerId());
+
+        final boolean hasRangeFilters =
+                (request.getHookIdLowerBoundInclusive() != null || request.getHookIdUpperBoundInclusive() != null);
+        final boolean hasEqFilters = (request.getHookIdEqualsFilters() != null
+                && !request.getHookIdEqualsFilters().isEmpty());
```
steven-sheehy: Useless parentheses on both.

rest-java/src/main/java/org/hiero/mirror/restjava/service/HookServiceImpl.java
```diff
@@ -0,0 +1,63 @@
...
+@NullMarked
+@RequiredArgsConstructor
+final class HookServiceImpl implements HookService {
+
+    private final HookRepository hookRepository;
+    private final EntityService entityService;
+
+    @Override
+    public Collection<Hook> getHooks(HooksRequest request) {
+        final var sort = Sort.by(request.getOrder(), "hookId");
```
steven-sheehy: Should store string as a constant.

rest-java/src/main/java/org/hiero/mirror/restjava/service/HookServiceImpl.java
```diff
@@ -0,0 +1,63 @@
...
+        final var id = entityService.lookup(request.getOwnerId());
+
+        final boolean hasRangeFilters =
+                (request.getHookIdLowerBoundInclusive() != null || request.getHookIdUpperBoundInclusive() != null);
+        final boolean hasEqFilters = (request.getHookIdEqualsFilters() != null
+                && !request.getHookIdEqualsFilters().isEmpty());
+
+        if (!hasRangeFilters && !hasEqFilters) {
+            return hookRepository.findByOwnerId(id.getId(), page);
+
```
steven-sheehy: We don't need this if statement or repository method if we default lower and upper bounds to 0/MAX_VALUE.

rest-java/src/main/java/org/hiero/mirror/restjava/service/HookServiceImpl.java
steven-sheehy: Default hookidEqualsFilters to empty list to avoid null checks

steven-sheehy (changes requested):

rest-java/src/test/java/org/hiero/mirror/restjava/controller/HooksControllerTest.java
sdimitrov9: Pushed now ,as part of @ ValueSource(strings = {"asc", "desc"}) void limitAndNextLink(String order) test.

rest-java/src/main/java/org/hiero/mirror/restjava/service/HookServiceImpl.java
sdimitrov9: Done

rest-java/src/main/java/org/hiero/mirror/restjava/service/HookServiceImpl.java
sdimitrov9: Removed

rest-java/src/main/java/org/hiero/mirror/restjava/service/HookServiceImpl.java
sdimitrov9: Done

rest-java/src/main/java/org/hiero/mirror/restjava/service/HookServiceImpl.java
sdimitrov9: Removed

rest-java/src/main/java/org/hiero/mirror/restjava/controller/HooksController.java
sdimitrov9: Done

rest-java/src/main/java/org/hiero/mirror/restjava/controller/HooksController.java
sdimitrov9: Done

rest-java/src/main/java/org/hiero/mirror/restjava/controller/HooksController.java
sdimitrov9: Done

rest-java/src/main/java/org/hiero/mirror/restjava/controller/HooksController.java
sdimitrov9: Addressed

rest-java/src/main/java/org/hiero/mirror/restjava/controller/HooksController.java
sdimitrov9: Addressed

rest-java/src/test/java/org/hiero/mirror/restjava/controller/HooksControllerTest.java
sdimitrov9: Addressed via successWithEntityIdAddressFormat(),successWithAliasAddressFormat(), successWithEvmAddressFormat() and invalidAccountIdFormat() tests.

rest-java/src/test/java/org/hiero/mirror/restjava/controller/HooksControllerTest.java
```diff
@@ -0,0 +1,475 @@
...
+                    .body(HooksResponse.class);
+
+            // then
+            assertThat(actual).isNotNull().isEqualTo(expectedResponse);
+        }
+    }
+
+    @DisplayName("/api/v1/accounts/{accountId}/hooks [Account ID Formats]")
+    @Nested
+    final class AccountIdFormatsTest extends EndpointTest {
```
steven-sheehy: We should not extend EndpointTest for account id as it re-runs a common set of tests like `etagHeader` multiple times per endpoint. Even if you disable them explicitly it's an abuse of `EndpointTest`. There should only be one `EndpointTest` extension for each unique controller method/endpoint.

I also don't think we need so many separate success tests. We can consolidate all success accountID tests into generic `success()` test. We just need to set up one account that has an EVM address and alias and look it up via all possible input types via a single parameterized test.

```java
@ParameterizedTest
@ValueSource(strings = {"1234", "0.1234", "0.0.1234", "0xac384c53f03855fa1b3616052f8ba32c6c2a2fec", 
"HIQQEXWKW53RKN4W6XXC4Q232SYNZ3SZANVZZSUME5B5PRGXL663UAQA",
                    "0.HIQQEXWKW53RKN4W6XXC4Q232SYNZ3SZANVZZSUME5B5PRGXL663UAQA",
                    "0.0.HIQQEXWKW53RKN4W6XXC4Q232SYNZ3SZANVZZSUME5B5PRGXL663UAQA",...})
void success(String ownerId) {
```

rest-java/src/test/java/org/hiero/mirror/restjava/controller/HooksControllerTest.java
```diff
@@ -0,0 +1,475 @@
...
+                    hooks.get(2).getHookId(),
+                    hooks.get(3).getHookId());
+            final var actual = restClient.get().uri(formattedParams).retrieve().body(HooksResponse.class);
+
+            // then
+            assertThat(actual).isNotNull().isEqualTo(expectedResponse);
+        }
+
+        @Test
+        void hookIdWithMissingOperatorDefaultsToEquals() {
```
steven-sheehy: Isn't this covered already with hookIdBounds?

rest-java/src/main/java/org/hiero/mirror/restjava/dto/HooksRequest.java
```diff
@@ -0,0 +1,32 @@
...
+    Sort.Direction order = Direction.DESC;
+
+    @Builder.Default
+    Collection<Long> hookIdEqualsFilters = Collections.emptyList();
+
+    @Builder.Default
+    Long hookIdLowerBoundInclusive = 0L;
+
+    @Builder.Default
+    Long hookIdUpperBoundInclusive = Long.MAX_VALUE;
```
steven-sheehy: These can be primitive

rest-java/src/main/java/org/hiero/mirror/restjava/dto/HooksRequest.java
```diff
@@ -0,0 +1,32 @@
...
+@Value
+@Builder
+public class HooksRequest {
+    EntityIdParameter ownerId;
+
+    @Builder.Default
+    int limit = 25;
+
+    @Builder.Default
+    Sort.Direction order = Direction.DESC;
```
steven-sheehy: Prefer `private final` added to value classes explicitly even if Lombok adds it implicitly. It makes it visually clear and avoids the codacy code smell.

rest-java/src/test/java/org/hiero/mirror/restjava/service/HookServiceTest.java
```diff
@@ -0,0 +1,132 @@
...
+import org.hiero.mirror.restjava.RestJavaIntegrationTest;
+import org.hiero.mirror.restjava.common.EntityIdParameter;
+import org.hiero.mirror.restjava.dto.HooksRequest;
+import org.hiero.mirror.restjava.repository.HookRepository;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.springframework.data.domain.Sort;
+
+@RequiredArgsConstructor
+class HookServiceTest extends RestJavaIntegrationTest {
```
steven-sheehy: final

rest-java/src/test/java/org/hiero/mirror/restjava/repository/HookRepositoryTest.java
```diff
@@ -0,0 +1,100 @@
...
+import org.hiero.mirror.common.domain.hook.Hook;
+import org.hiero.mirror.restjava.RestJavaIntegrationTest;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+import org.springframework.data.domain.PageRequest;
+import org.springframework.data.domain.Pageable;
+import org.springframework.data.domain.Sort;
+
+@RequiredArgsConstructor
+class HookRepositoryTest extends RestJavaIntegrationTest {
```
steven-sheehy: final

Issue #12118, opened 2025-10-03
HIP-1195 Add `LambdaSStoreTransaction` support


ashumahajan:
### Problem

Need to implement new transaction type LambdaSStore  being introduced as part of Hooks

### Solution

- Add new Transaction Type for LambdaSStoreTransactionBody
- Add LambdaSStoreTransactionHandler
- Add processing of HookStorageChange

PR #12261, opened 2025-10-23
HIP-1195 Support hooks in CryptoCreate and CryptoUpdate transactions

### LINKED ISSUE #11999: HIP-1195 update importer
### Problem

Consensus nodes will soon support HIP-1195 hooks so mirror node needs to ensure it can store the new type of data.

### Solution

Implement the importer changes from the design doc.

### Alternatives

_No response_

ashumahajan:
**Description**:
Consensus nodes will support HIP-1195 hooks so mirror node needs to ensure it can store the new type of data.

Add support for ...

* Creation of Hook using CryptoCreate
* Creation of Hook using CryptoUpdate
* Deletion of Hook using CryptoUpdate
* Creation and Deletion of Hook using CryptoUpdate

Fixes #11999 

**Checklist**

### DIFF
--- a/importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListener.java
+++ b/importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListener.java
@@ -640,7 +640,8 @@ public class SqlEntityListener implements EntityListener, RecordStreamFileListen
                 previous.getDeleted() && previous.getTimestampLower().equals(current.getTimestampLower()) ? 1 : 0;
         previous.setTimestampUpper(current.getTimestampLower() + adjustment);
 
-        if (current.getCreatedTimestamp() != null && current.getCreatedTimestamp() == current.getTimestampLower()) {
+        if (current.getCreatedTimestamp() != null
+                && current.getCreatedTimestamp().equals(current.getTimestampLower())) {
             // hook creation, don't merge
             return current;
         }
--- a/importer/src/test/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListenerTest.java
+++ b/importer/src/test/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListenerTest.java
@@ -3607,14 +3607,19 @@ final class SqlEntityListenerTest extends ImporterIntegrationTest {
         return nft;
     }
 
-    @Test
-    void onHook() {
+    @ParameterizedTest
+    @ValueSource(ints = {1, 2})
+    void onHook(int commitIndex) {
         // given
         var hook1 = domainBuilder.hook().get();
         var hook2 = domainBuilder.hook().get();
 
         // when
         sqlEntityListener.onHook(hook1);
+        if (commitIndex > 1) {
+            completeFileAndCommit();
+            assertThat(hookRepository.findAll()).containsExactlyInAnyOrder(hook1);
+        }
         sqlEntityListener.onHook(hook2);
         completeFileAndCommit();
 
@@ -3770,7 +3775,7 @@ final class SqlEntityListenerTest extends ImporterIntegrationTest {
                 .build();
 
         assertThat(hookRepository.findAll()).containsExactly(hookCreate2);
-        assertThat(findHistory(Hook.class)).contains(expectedHookCreate, expectedHookDelete);
+        assertThat(findHistory(Hook.class)).containsExactlyInAnyOrder(expectedHookCreate, expectedHookDelete);
     }
 
     @ParameterizedTest


importer/src/main/resources/db/migration/v1/V1.112.1__add_hooks_support.sql
```diff
@@ -27,6 +27,8 @@ create table if not exists hook_history
 (
```
jnels124, line 27: We should put these changes in a new migration as these have ran in some environments (in integration)

common/src/main/java/org/hiero/mirror/common/domain/hook/AbstractHook.java
```diff
@@ -63,6 +63,10 @@ public void setOwnerId(EntityId ownerId) {
         this.ownerId = ownerId.getId();
     }
 
+    public void setOwnerId(long ownerId) {
```
jnels124: this function can be removed as `@Data` annotation already adds it

importer/src/main/resources/db/migration/v1/V1.112.1__add_hooks_support.sql
jnels124: Shouldn't we always have the contract id even for history?

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/EVMHookHandler.java
```diff
@@ -0,0 +1,75 @@
...
+     * Processes hook creation details from any transaction type and creates corresponding Hook entities. This method
+     * can be used by CryptoCreate, CryptoUpdate, ContractCreate, and ContractUpdate handlers.
+     *
+     * @param recordItem              the record item containing the transaction
+     * @param entityId                the entity ID that owns the hooks
+     * @param hookCreationDetailsList the list of hook creation details from the transaction
+     */
+    void processHookCreationDetails(
+            RecordItem recordItem, EntityId entityId, List<HookCreationDetails> hookCreationDetailsList) {
+        if (hookCreationDetailsList != null && !hookCreationDetailsList.isEmpty()) {
```
jnels124: nit: use `CollectionUtils.isEmpty`

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/EVMHookHandler.java
```diff
@@ -0,0 +1,75 @@
...
+            hookCreationDetailsList.forEach(hookCreationDetails -> {
+                var spec = hookCreationDetails.getLambdaEvmHook().getSpec();
+                Hook hook = Hook.builder()
+                        .adminKey(hookCreationDetails.getAdminKey().toByteArray())
```
jnels124: For updates, do the details contain all previously populated fields?

importer/src/test/java/org/hiero/mirror/importer/parser/domain/RecordItemBuilder.java
```diff
@@ -532,6 +536,34 @@ public Builder<CryptoCreateTransactionBody.Builder> cryptoCreate() {
         return new Builder<>(TransactionType.CRYPTOCREATEACCOUNT, builder).receipt(r -> r.setAccountID(accountId()));
     }
 
+    @SuppressWarnings("deprecation")
+    public Builder<CryptoCreateTransactionBody.Builder> cryptoCreateWithHooks() {
+        var hookCreationDetails = HookCreationDetails.newBuilder()
+                .setExtensionPoint(HookExtensionPoint.ACCOUNT_ALLOWANCE_HOOK)
```
jnels124, line 542: don't we need to set an owner?

importer/src/main/resources/db/migration/v1/V1.112.1__add_hooks_support.sql
ashumahajan: The reason, why i made this non-null is to ensure that we do record delete even though we missed create hook.

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/EVMHookHandler.java
ashumahajan: This is the reason why I made contract_id nullable. In a normal circumstance, the only update will be to delete the hook. other properties either have default value or is the key

importer/src/test/java/org/hiero/mirror/importer/parser/domain/RecordItemBuilder.java
ashumahajan: HookCreationDetails is HAPI proto object and doesn't have owner. Owner is picked up from TransactionBody for account and contract

docs/design/hiero-hooks.md
```diff
@@ -197,6 +197,8 @@ create table if not exists hook_history
 (
     like hook including defaults
 );
+-- Allow contract_id to be null in hook_history
+alter table hook_history alter column contract_id drop not null;
```
xin-hedera: think we can do it in a clean way in the design doc, just update the existing schema in `create table ...`, the other reason is we need to make it nullable in `hook` table as well.

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListener.java
```diff
@@ -629,6 +635,13 @@ private Nft mergeNft(Nft cachedNft, Nft newNft) {
         return dest;
     }
 
+    private Hook mergeHook(Hook previous, Hook current) {
+        if (previous != null && previous.getDeleted() && current.getTimestampUpper() == null) {
```
xin-hedera: - previous & current won't be null
- should also merge all other columns, except `ownerId` & `hookId`.

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/EVMHookHandler.java
```diff
@@ -0,0 +1,75 @@
...
+import org.hiero.mirror.importer.parser.record.entity.EntityListener;
+import org.jspecify.annotations.NullMarked;
+
+@Named
+@RequiredArgsConstructor
+@NullMarked
+final class EVMHookHandler {
+
+    private final EntityListener entityListener;
+
```
xin-hedera, line 29: can have a single public `process` method which takes both a list of hooks to delete, and a list of new hooks.

For `CryptoCreate` and `ContractCreate`, just pass an empty list as the list of hooks to delete.

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/EVMHookHandler.java
```diff
@@ -0,0 +1,75 @@
...
+                        .adminKey(hookCreationDetails.getAdminKey().toByteArray())
+                        .createdTimestamp(recordItem.getConsensusTimestamp())
+                        .contractId(EntityId.of(spec.getContractId()))
+                        .deleted(false)
+                        .extensionPoint(HookExtensionPoint.ACCOUNT_ALLOWANCE_HOOK)
+                        .hookId(hookCreationDetails.getHookId())
+                        .timestampRange(Range.atLeast(recordItem.getConsensusTimestamp()))
+                        .type(HookType.LAMBDA)
+                        .ownerId(entityId.getId())
+                        .build();
```
xin-hedera: - `lambda_evm_hook` is in a protobuf oneof, in general we need to do get the value only if such a field is set. think we want to skip hooks with the field not set and log a recoverable error
- `adminKey` - should do `hasAdminKey` check first
-  what about the `storage_updates` inside `LambdaEvmHook`? I'm not sure if it's a recent change to the protobuf or not, however this is what's in consensus node `main` branch
```
/**
 * Definition of a lambda EVM hook.
 */
message LambdaEvmHook {
  /**
   * The specification for the hook.
   */
  EvmHookSpec spec = 1;

  /**
   * Initial storage updates for the lambda, if any.
   */
  repeated LambdaStorageUpdate storage_updates = 2;
}
```

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/EVMHookHandler.java
```diff
@@ -0,0 +1,75 @@
...
+     * @param recordItem              the record item containing the transaction
+     * @param entityId                the entity ID that owns the hooks
+     * @param hookCreationDetailsList the list of hook creation details from the transaction
+     */
+    void processHookCreationDetails(
+            RecordItem recordItem, EntityId entityId, List<HookCreationDetails> hookCreationDetailsList) {
+        if (hookCreationDetailsList != null && !hookCreationDetailsList.isEmpty()) {
+            hookCreationDetailsList.forEach(hookCreationDetails -> {
+                var spec = hookCreationDetails.getLambdaEvmHook().getSpec();
+                Hook hook = Hook.builder()
```
xin-hedera: can use `final var`

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/EVMHookHandler.java
```diff
@@ -0,0 +1,75 @@
...
+     * functionality to update the database.
+     *
+     * @param recordItem          the record item containing the transaction
+     * @param entityId            the entity ID that owns the hooks
+     * @param hookIdsToDeleteList the list of hook IDs to delete
+     */
+    public void processHookDeletion(RecordItem recordItem, EntityId entityId, List<Long> hookIdsToDeleteList) {
+        if (hookIdsToDeleteList != null && !hookIdsToDeleteList.isEmpty()) {
+            hookIdsToDeleteList.forEach(hookId -> {
+                Hook hook = Hook.builder()
```
xin-hedera:
```suggestion
                final var hook = Hook.builder()
```

importer/src/test/java/org/hiero/mirror/importer/parser/domain/RecordItemBuilder.java
```diff
@@ -532,6 +536,34 @@ public Builder<CryptoCreateTransactionBody.Builder> cryptoCreate() {
...
+                .setHookId(1L)
+                .setLambdaEvmHook(LambdaEvmHook.newBuilder()
+                        .setSpec(EvmHookSpec.newBuilder().setContractId(contractId()))
+                        .build())
+                .setAdminKey(key())
+                .build();
+
+        var builder = CryptoCreateTransactionBody.newBuilder()
```
xin-hedera: should reuse the `cryptoTransfer()` method in the class and customize the hooks, e.g.

```java
  return cryptoCreate().transactionBody(b -> b.addHookCreationDetails(...));
```

importer/src/test/java/org/hiero/mirror/importer/parser/record/transactionhandler/EVMHookHandlerTest.java
```diff
@@ -0,0 +1,249 @@
...
+import org.hiero.mirror.importer.parser.record.entity.EntityListener;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.ArgumentCaptor;
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith(MockitoExtension.class)
+class EVMHookHandlerTest {
```
xin-hedera:
```suggestion
final class EVMHookHandlerTest {
```

importer/src/test/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListenerTest.java
```diff
@@ -3602,6 +3607,61 @@ private Nft getNft(
...
+        // when
+        sqlEntityListener.onHook(hook1);
+        sqlEntityListener.onHook(hook2);
+        completeFileAndCommit();
+
+        // then
+        assertThat(hookRepository.findAll()).containsExactlyInAnyOrder(hook1, hook2);
+        assertThat(findHistory(Hook.class)).isEmpty();
+    }
+
```
xin-hedera, line 3630: there is no test coverage for hook creation and hook delete (same hook) in different transactions.

what we'd expect is

- `adminKey`, `createdTimestamp`, `contractId`, `extentionPoint`, and `type` merged from hook creation to hook delete
- `timestampRange` for hook creation is closed

We also want to make the test parameterized so test

- the two transactions in the same record file
- the two transactions in different record files

In addition, we want to test the scenario that a hook is created with admin key, deleted, and recreated without admin key

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/CryptoCreateTransactionHandler.java
```diff
@@ -87,6 +91,8 @@ protected void doUpdateEntity(Entity entity, RecordItem recordItem) {
 
         updateStakingInfo(recordItem, entity);
         entityListener.onEntity(entity);
+        evmHookHandler.processHookCreationDetails(
+                recordItem, entity.toEntityId(), transactionBody.getHookCreationDetailsList());
```
xin-hedera: looks like the only thing needed from `entity` is the encoded id, so instead of `entityt.toEntityId()`, just pass `entity.getId()`

for the similar reason, pass `recordItem.getConsensusTimestamp()` instead of `recordItem`

docs/design/hiero-hooks.md
ashumahajan: Yes! That is why one of the testcase is failing

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListener.java
ashumahajan: The merge logic is only applicable for delete and we only get hook_id in the proto. No other info is added and if we add all other columns, that means we if some recreate a hook_id (same transaction delete and create) with a different contract_id, we will have incorrect data for delete.

xin-hedera: no this is not how it works.

the merge function will get invoked on the same owner id and hook id since that's the key of the underlying java `Map`.
 
So in the merge function, we'll figure out wether previous & current belong to the same lifecycle or not.

Same for the generate upsert SQL.

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/EVMHookHandler.java
ashumahajan: storage_updates is yet to be implemented, will be part of #12118 
Added Todo, to ensure that it is not missed

importer/src/test/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListenerTest.java
ashumahajan: How will we handle the adminKey being set to null. As per my discussion with you that is being handled by upsert logic. So it will always merge the adminKey from history table?

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/CryptoCreateTransactionHandler.java
ashumahajan: Kept RecordItem, since we add recordItem.addEntityId(hook.getContractId());

common/src/main/java/org/hiero/mirror/common/domain/hook/AbstractHook.java
```diff
@@ -30,16 +31,46 @@
 public abstract class AbstractHook implements History {
 
     @ToString.Exclude
+    @UpsertColumn(
+            coalesce =
+                    """
+                            case when coalesce(e_created_timestamp, 0) != coalesce(created_timestamp, 0) then {0}
+                                 else coalesce({0}, e_{0})
```
steven-sheehy: Store as two private static final strings and reuse across annotations.

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/CryptoCreateTransactionHandler.java
```diff
@@ -87,6 +91,7 @@ protected void doUpdateEntity(Entity entity, RecordItem recordItem) {
 
         updateStakingInfo(recordItem, entity);
         entityListener.onEntity(entity);
+        evmHookHandler.process(recordItem, entity.getId(), transactionBody.getHookCreationDetailsList(), null);
```
steven-sheehy: Prefer non-null values like List.of()

importer/src/main/resources/db/migration/v2/V2.17.1__add_hooks_support.sql
```diff
@@ -8,7 +8,7 @@ create type hook_extension_point as enum ('ACCOUNT_ALLOWANCE_HOOK');
 -- Main hook table
 create table if not exists hook
 (
-    contract_id         bigint                not null,
+    contract_id         bigint ,
```
steven-sheehy: You can't update migrations. This has already been ran on all environments. Please create a new migration.

importer/src/test/java/org/hiero/mirror/importer/parser/domain/RecordItemBuilder.java
```diff
@@ -532,6 +536,20 @@ public Builder<CryptoCreateTransactionBody.Builder> cryptoCreate() {
         return new Builder<>(TransactionType.CRYPTOCREATEACCOUNT, builder).receipt(r -> r.setAccountID(accountId()));
     }
 
+    @SuppressWarnings("deprecation")
+    public Builder<CryptoCreateTransactionBody.Builder> cryptoCreateWithHooks() {
```
steven-sheehy: We don't usually create a separate builder for a transaction with a custom extra fields. Just add this to `cryptoCreate()`

importer/src/test/java/org/hiero/mirror/importer/parser/domain/RecordItemBuilder.java
```diff
@@ -635,6 +653,49 @@ public Builder<CryptoUpdateTransactionBody.Builder> cryptoUpdate() {
...
+                .build();
+
+        var accountId = accountId();
+        var builder = CryptoUpdateTransactionBody.newBuilder()
+                .setAccountIDToUpdate(accountId)
+                .addHookCreationDetails(hookCreationDetails);
+        return new Builder<>(TransactionType.CRYPTOUPDATEACCOUNT, builder);
+    }
+
+    public Builder<CryptoUpdateTransactionBody.Builder> cryptoUpdateWithHookDeletionAndCreation() {
```
steven-sheehy: We shouldn't have any of these as they're specific to certain test cases. The RecordItemBuilder only has a single method per transaction type except in rare cases like different token types. Instead, update cryptoUpdate to include both a single hook create and delete. Then for tests that need the others just customize the cryptoupdate.

importer/src/test/java/org/hiero/mirror/importer/parser/domain/RecordItemBuilder.java
```diff
@@ -635,6 +653,49 @@ public Builder<CryptoUpdateTransactionBody.Builder> cryptoUpdate() {
...
+        var hookCreationDetails = HookCreationDetails.newBuilder()
+                .setExtensionPoint(HookExtensionPoint.ACCOUNT_ALLOWANCE_HOOK)
+                .setHookId(1L) // Same hook ID to delete and create
+                .setLambdaEvmHook(LambdaEvmHook.newBuilder()
+                        .setSpec(EvmHookSpec.newBuilder().setContractId(contractId()))
+                        .build())
+                .setAdminKey(key())
+                .build();
```
steven-sheehy: Since it's common to both create and update (and later contracts) create a `private HookCreationDetails.Builder hookCreationDetails()` helper method.

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListener.java
```diff
@@ -629,6 +635,30 @@ private Nft mergeNft(Nft cachedNft, Nft newNft) {
...
+            previous.setTimestampUpper(current.getTimestampLower() + 1);
+        }
+        Hook src = current.hasHistory() ? current : previous;
+        Hook dest = current.hasHistory() ? previous : current;
```
steven-sheehy: We should always use final var. Also add some newlines for clarity
```suggestion

        final var src = current.hasHistory() ? current : previous;
        final var dest = current.hasHistory() ? previous : current;

```

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/EVMHookHandler.java
```diff
@@ -0,0 +1,147 @@
...
+import org.hiero.mirror.common.domain.hook.HookType;
+import org.hiero.mirror.common.domain.transaction.RecordItem;
+import org.hiero.mirror.importer.parser.record.entity.EntityListener;
+import org.jspecify.annotations.NullMarked;
+import org.springframework.util.CollectionUtils;
+
+@Named
+@RequiredArgsConstructor
+@NullMarked
+@CustomLog
```
steven-sheehy: nit: sort

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/EVMHookHandler.java
```diff
@@ -0,0 +1,147 @@
...
+                    hookCreationDetails.getHookId(),
+                    recordItem.getConsensusTimestamp());
+            return;
+        }
+
+        final var lambdaEvmHook = hookCreationDetails.getLambdaEvmHook();
+
+        // Check if spec is set
+        if (!lambdaEvmHook.hasSpec()) {
+            log.warn(
```
steven-sheehy: Use `Utility.handleRecoverableError` as this is a scenario we should alert on but not stop processing. Same for other warnings.

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/EVMHookHandler.java
```diff
@@ -0,0 +1,147 @@
...
+                    "Skipping hook creation for hookId {} - contract_id not set in spec in transaction at {}",
+                    hookCreationDetails.getHookId(),
+                    recordItem.getConsensusTimestamp());
+            return;
+        }
+
+        final var hookBuilder = Hook.builder()
+                .createdTimestamp(recordItem.getConsensusTimestamp())
+                .contractId(EntityId.of(spec.getContractId()))
```
steven-sheehy: nit: sorting of createdTimestamp and ownerId

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/EVMHookHandler.java
```diff
@@ -0,0 +1,147 @@
...
+        recordItem.addEntityId(hook.getContractId());
+        entityListener.onHook(hook);
+
+        // Process storage updates if present
+        if (!CollectionUtils.isEmpty(lambdaEvmHook.getStorageUpdatesList())) {
+            log.debug(
+                    "Processing {} storage updates for hookId {} in transaction at {}",
+                    lambdaEvmHook.getStorageUpdatesList().size(),
+                    hookCreationDetails.getHookId(),
+                    recordItem.getConsensusTimestamp());
```
steven-sheehy: Don't think we need this debug

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/EVMHookHandler.java
```diff
@@ -0,0 +1,147 @@
...
+    /**
+     * Processes hook deletion by performing soft delete. Creates Hook entities with deleted=true and uses upsert
+     * functionality to update the database.
+     *
+     * @param recordItem          the record item containing the transaction
+     * @param entityId            the entity ID that owns the hooks
+     * @param hookIdsToDeleteList the list of hook IDs to delete
+     */
+    private void processHookDeletion(RecordItem recordItem, long entityId, List<Long> hookIdsToDeleteList) {
+        if (!CollectionUtils.isEmpty(hookIdsToDeleteList)) {
```
steven-sheehy: Can omit if when we switch to List.of()

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/EVMHookHandler.java
```diff
@@ -0,0 +1,147 @@
...
+            hookIdsToDeleteList.forEach(hookId -> {
+                Hook hook = Hook.builder()
+                        .hookId(hookId)
+                        .ownerId(entityId)
+                        .timestampRange(Range.atLeast(recordItem.getConsensusTimestamp()))
+                        .deleted(true)
```
steven-sheehy: nit: sort

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/EVMHookHandler.java
```diff
@@ -0,0 +1,147 @@
...
+                    hookCreationDetails.getHookId(),
+                    recordItem.getConsensusTimestamp());
+            return;
+        }
+
+        final var hookBuilder = Hook.builder()
+                .createdTimestamp(recordItem.getConsensusTimestamp())
+                .contractId(EntityId.of(spec.getContractId()))
+                .deleted(false)
+                .extensionPoint(HookExtensionPoint.ACCOUNT_ALLOWANCE_HOOK)
```
steven-sheehy: We should translate this from HookExtensionPoint in tx body and not hardcode

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/EVMHookHandler.java
```diff
@@ -0,0 +1,147 @@
...
+                .hookId(hookCreationDetails.getHookId())
+                .timestampRange(Range.atLeast(recordItem.getConsensusTimestamp()))
+                .type(HookType.LAMBDA)
```
steven-sheehy: We should translate this from oneof enum case

importer/src/main/resources/db/migration/v2/V2.17.1__add_hooks_support.sql
ashumahajan: Spoken with Jesse and he mentioned that V2 is not yet migrated. So I added another migration file for V1 and for V2 just updated the original sql file

xin-hedera: not really, 2.17.1 is added in last sprint and released in 0.141

ashumahajan: ok, will update.

importer/src/test/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListenerTest.java
ashumahajan: Updated AbstractHook with Upsertable column and added testcases too

common/src/main/java/org/hiero/mirror/common/domain/hook/AbstractHook.java
```diff
@@ -48,21 +65,29 @@ public abstract class AbstractHook implements History {
...
     private Range<Long> timestampRange;
 
     @Enumerated(EnumType.STRING)
     @JdbcTypeCode(SqlTypes.NAMED_ENUM)
+    @UpsertColumn(
+            coalesce =
+                    """
+                            case when coalesce(e_created_timestamp, 0) != coalesce(created_timestamp, 0) then coalesce({0}, {1})
+                                 else coalesce({0}, e_{0}, {1})
+                            end""")
```
steven-sheehy: Can't we replace these two with the constant as well?

common/src/main/java/org/hiero/mirror/common/domain/hook/HookType.java
```diff
@@ -3,5 +3,6 @@
 package org.hiero.mirror.common.domain.hook;
 
 public enum HookType {
+    PURE,
```
steven-sheehy: We shouldn't add this or the mapping since it's removed in v0.68 protobuf.

importer/src/test/java/org/hiero/mirror/importer/parser/domain/RecordItemBuilder.java
```diff
@@ -193,6 +198,7 @@
 /**
  * Generates typical protobuf request and response objects with all fields populated.
  */
+@Slf4j
```
steven-sheehy: Remove

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/EVMHookHandler.java
```diff
@@ -0,0 +1,174 @@
...
+     * Processes hook creation details from any transaction type and creates corresponding Hook entities. This method
+     * can be used by CryptoCreate, CryptoUpdate, ContractCreate, and ContractUpdate handlers.
+     *
+     * @param recordItem              the record item containing the transaction
+     * @param entityId                the entity ID that owns the hooks
+     * @param hookCreationDetailsList the list of hook creation details from the transaction
+     */
+    private void processHookCreationDetails(
+            RecordItem recordItem, long entityId, List<HookCreationDetails> hookCreationDetailsList) {
+        if (!CollectionUtils.isEmpty(hookCreationDetailsList)) {
```
steven-sheehy: Same thing. We know this will never be null so can remove if

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/EVMHookHandler.java
```diff
@@ -0,0 +1,174 @@
...
+                    "Skipping hook creation for hookId {} - no hook implementation set in transaction at {}",
+                    hookCreationDetails.getHookId(),
+                    recordItem.getConsensusTimestamp());
+            return;
+        }
+
+        // Get the spec from whichever hook type is set
+        final var spec = hookCreationDetails.hasLambdaEvmHook()
+                ? hookCreationDetails.getLambdaEvmHook().getSpec()
+                : hookCreationDetails.getPureEvmHook().getSpec();
```
steven-sheehy: Remove pure evm logic as it's going away in 0.68.

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/EVMHookHandler.java
```diff
@@ -0,0 +1,174 @@
...
+     * Processes hook deletion by performing soft delete. Creates Hook entities with deleted=true and uses upsert
+     * functionality to update the database.
+     *
+     * @param recordItem          the record item containing the transaction
+     * @param entityId            the entity ID that owns the hooks
+     * @param hookIdsToDeleteList the list of hook IDs to delete
+     */
+    private void processHookDeletion(RecordItem recordItem, long entityId, List<Long> hookIdsToDeleteList) {
+        hookIdsToDeleteList.forEach(hookId -> {
+            Hook hook = Hook.builder()
```
steven-sheehy:
```suggestion
            final var hook = Hook.builder()
```

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/EVMHookHandler.java
```diff
@@ -0,0 +1,174 @@
...
+     * @return the corresponding domain HookType
+     */
+    private HookType translateHookType(HookCreationDetails hookCreationDetails) {
+        if (hookCreationDetails.hasLambdaEvmHook()) {
+            return HookType.LAMBDA;
+        } else if (hookCreationDetails.hasPureEvmHook()) {
+            Utility.handleRecoverableError(
+                    "PureEvmHook not yet supported for hookId {}, treating as LAMBDA", hookCreationDetails.getHookId());
+            return HookType.PURE;
+        } else {
```
steven-sheehy: Prefer `switch(hookCreationDetails.getHookCase())` like extension point

importer/src/test/java/org/hiero/mirror/importer/parser/domain/RecordItemBuilder.java
```diff
@@ -624,14 +644,16 @@ public Builder<CryptoTransferTransactionBody.Builder> cryptoTransfer(TransferTyp
...
         var builder = CryptoUpdateTransactionBody.newBuilder()
                 .setAutoRenewPeriod(duration(30))
                 .setAccountIDToUpdate(accountId)
                 .setDeclineReward(BoolValue.of(true))
                 .setKey(key())
                 .setProxyAccountID(accountId())
                 .setReceiverSigRequired(false)
-                .setStakedNodeId(1L);
+                .setStakedNodeId(1L)
+                .addHookCreationDetails(hookCreationDetails); // Create hook ID 1
```
steven-sheehy: Can't we also add a deletion by default? Or does that cause too much issues in existing tests?

importer/src/test/java/org/hiero/mirror/importer/parser/domain/RecordItemBuilder.java
```diff
@@ -516,6 +522,9 @@ public Builder<CryptoApproveAllowanceTransactionBody.Builder> cryptoApproveAllow
 
     @SuppressWarnings("deprecation")
     public Builder<CryptoCreateTransactionBody.Builder> cryptoCreate() {
+
+        var hookCreationDetails = hookCreationDetails().build();
+
```
steven-sheehy: Can move inline to `addHookCreationDetails` since you're not using it in multiple places.

importer/src/test/java/org/hiero/mirror/importer/parser/domain/RecordItemBuilder.java
ashumahajan: There are few testcases that just uses hookCreation without deletion. I can clear the hookDeletionId list in individual tests, if that is how it is suppose to be.

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/EVMHookHandler.java
```diff
@@ -0,0 +1,164 @@
...
+     * Translates the protobuf HookExtensionPoint from the transaction body to the domain HookExtensionPoint.
+     *
+     * @param protoExtensionPoint the HookExtensionPoint from the transaction protobuf
+     * @return the corresponding domain HookExtensionPoint
+     */
+    private HookExtensionPoint translateHookExtensionPoint(
+            com.hedera.hapi.node.hooks.legacy.HookExtensionPoint protoExtensionPoint) {
+        return switch (protoExtensionPoint) {
+            case ACCOUNT_ALLOWANCE_HOOK -> HookExtensionPoint.ACCOUNT_ALLOWANCE_HOOK;
+            case UNRECOGNIZED -> {
```
steven-sheehy:
```suggestion
            default -> {
```

steven-sheehy (approved):

common/src/main/java/org/hiero/mirror/common/domain/hook/AbstractHook.java
```diff
@@ -29,17 +30,33 @@
 @Upsertable(history = true)
 public abstract class AbstractHook implements History {
 
+    public static final String UPSERTABLE_COLUMN_COALESCE =
+            """
+                    case when coalesce(e_created_timestamp, 0) != coalesce(created_timestamp, 0) then {0}
```
xin-hedera: this doesn't work for

- hook creation in block X
- hook deletion in block Y

because e_created_timestamp will be > 0, and created_timestamp = null, `coalesce(e_created_timestamp, 0) != coalesce(created_timestamp, 0)` evals to true. The row in `hook` table ends up with every column = null except, deleted, hookId, ownerId, timestamp_range. 

The correct condition should be

- when created_timestamp = lower(timestamp_range), don't coalsece
- otherwise, `coalesce({0}, e_{0})`

This also means lack of test coverage.

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListener.java
```diff
@@ -629,6 +635,32 @@ private Nft mergeNft(Nft cachedNft, Nft newNft) {
         return dest;
     }
 
+    private Hook mergeHook(Hook previous, Hook current) {
+        if (previous != null && previous.getDeleted() && current.getTimestampUpper() == null) {
+            previous.setTimestampUpper(current.getTimestampLower() + 1);
+        }
+
+        final var src = current.hasHistory() ? current : previous;
+        final var dest = current.hasHistory() ? previous : current;
```
xin-hedera: All hook objects are created with history, i.e., with timestampLower set.

what we really need is

```java
    if (current.getCreatedTimestamp() != null && current.getCreatedTimestamp() == current.getTimestampLower()) {
       // hook creation, don't merge
       return current;
     }

    // conditionally merge all fields from previous to current if null, except ownerId, hookId, timestampRange
```

importer/src/test/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListenerTest.java
```diff
@@ -3602,6 +3612,329 @@ private Nft getNft(
...
+    @Test
+    void onHookMergeWithoutPrevious() {
```
xin-hedera: this is duplicate of `onHook`, only difference is 2 hooks in `onHook` vs 1 hook here.

importer/src/test/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListenerTest.java
```diff
@@ -3602,6 +3612,329 @@ private Nft getNft(
...
+                .hookId(hookId)
+                .ownerId(ownerId)
+                .timestampRange(Range.atLeast(createdTimestamp))
+                .type(HookType.LAMBDA)
+                .build();
+
+        var hookDelete = Hook.builder()
+                .deleted(true)
+                .hookId(hookId)
+                .createdTimestamp(createdTimestamp)
```
xin-hedera: we need to make this a parameterized test so we can also test hook creation and hook deletion in different blocks.

can check the following test as reference


```java
    @ParameterizedTest
    @ValueSource(ints = {1, 2})
    void onCustomFee(int commitIndex)
```

note when `commitIndex = 2`, hookDelete.createdTimestamp should be `null`.

importer/src/test/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListenerTest.java
```diff
@@ -3602,6 +3612,329 @@ private Nft getNft(
...
+                .ownerId(ownerId)
+                .timestampRange(Range.atLeast(deletedTimestamp))
+                .build();
+
+        // when
+        sqlEntityListener.onHook(hookCreate);
+        completeFileAndCommit();
+        assertThat(hookRepository.findAll()).containsExactly(hookCreate);
+
+        sqlEntityListener.onHook(hookDelete);
```
xin-hedera, line 3765: The test assumes hookDelete is in a different block, then there's no way to have a non-null `hookDelete.createdTimestamp`. so we are testing with incorrect setup.

importer/src/test/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListenerTest.java
```diff
@@ -3602,6 +3612,329 @@ private Nft getNft(
...
+        completeFileAndCommit();
+
+        var expectedDeleteHistory = hookDelete.toBuilder()
```
xin-hedera: you can do `hookCreation.toBuilder().timestampRange(Range.atLeast(deletedTimestamp)).build()`, less code

importer/src/test/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListenerTest.java
```diff
@@ -3602,6 +3612,329 @@ private Nft getNft(
...
+                () -> assertEquals(contractId, actualOptional.get().getContractId()),
+                () -> assertEquals(
+                        HookExtensionPoint.ACCOUNT_ALLOWANCE_HOOK,
+                        actualOptional.get().getExtensionPoint()),
+                () -> assertEquals(HookType.LAMBDA, actualOptional.get().getType()),
+                () -> assertEquals(
+                        EntityId.of(ownerId).getId(), actualOptional.get().getOwnerId()),
+                () -> assertEquals(createdTimestamp, actualOptional.get().getCreatedTimestamp()),
+                () -> assertNotNull(actualOptional.get().getTimestampUpper()),
+                () -> assertFalse(actualOptional.get().getDeleted()));
```
xin-hedera:
```suggestion
    var expectedHookHistory = hookCreation.toBuilder().timestampRange(Range.closedOpen(createdTimestamp, deletedTimestamp)).build();
assertThat(findHistory(Hook.class)).containsOnly(expectedHookHistory);
```
```

File: importer/src/test/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListenerTest.java [hunk 48]
```diff
@@ -3602,6 +3612,329 @@ private Nft getNft(
...
+    }
+
+    @Test
+    void onHookCreateDeleteRecreateWithoutAdminKey() {
```
xin-hedera: same comment, we need to test with multiple blocks

importer/src/test/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListenerTest.java
```diff
@@ -3602,6 +3612,329 @@ private Nft getNft(
...
+                .ownerId(ownerId)
+                .timestampRange(Range.atLeast(timestamp))
+                .build();
+
+        // when
+        sqlEntityListener.onHook(hookDelete);
+        completeFileAndCommit();
+
+        // then - hook should be persisted with default values for missing fields
+        var expectedHook = Hook.builder()
```
xin-hedera: can use `hookDelete.toBuilder()...` and only set those with default value from schema.

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/EVMHookHandler.java
```diff
@@ -0,0 +1,164 @@
...
+        if (!spec.hasContractId()) {
+            Utility.handleRecoverableError(
+                    "Skipping hook creation for hookId {} - contract_id not set in hook spec in transaction at {}",
+                    hookCreationDetails.getHookId(),
+                    recordItem.getConsensusTimestamp());
+            return;
+        }
+
+        final var hookBuilder = Hook.builder()
+                .contractId(EntityId.of(spec.getContractId()))
```
xin-hedera: we need to use `EntityIdService` to resolve contract id in the form of EVM address here, unless we are sure EVM address is not allowed.

importer/src/test/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListenerTest.java
xin-hedera: correct some of my previous comment. `hookDelete.createdTimestamp` should be always null, and when it's passed to `sqlEntityListener.onHook`, the merge function will merge it from `previous` if available, or the upsert sql will coalesce and copy the existing row's value.

ashumahajan: The reason why it does not make sense to make it parameterized is when we test same record file, delete happens before create and when testing for separate record file, create happens before delete so the verification for history and deletion becomes different.

importer/src/test/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListenerTest.java
ashumahajan: I believe we can ignore this based on your other comment

importer/src/test/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListenerTest.java
ashumahajan: I had created a separate testcase for the usecase. The validation and setup for both cases are different due to sequence

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/EVMHookHandler.java
ashumahajan: not sure what to do here

steven-sheehy: Inject EntityIdService and call `entityIdService.lookup(spec.getContractId()`

steven-sheehy: We should have an integration test for this for when lookup returns empty. I'm worried for partial mirror nodes this will halt importer due to not null constraint violation.

ashumahajan: Added a testcase EVMHookHandlerTest.processHookCreationDetailsWithEmptyContract to handle the usecase

common/src/main/java/org/hiero/mirror/common/domain/hook/AbstractHook.java
```diff
@@ -29,17 +30,33 @@
 @Upsertable(history = true)
 public abstract class AbstractHook implements History {
 
+    public static final String UPSERTABLE_COLUMN_COALESCE =
+            """
+                    case when created_timestamp = lower(timestamp_range) then {0}
+                         else coalesce({0}, e_{0})
+                    end""";
+    public static final String UPSERTABLE_COLUMN_WITH_DEFAULT_COALESCE =
```
xin-hedera:
```suggestion
    private static final String UPSERTABLE_COLUMN_COALESCE =
            """
                    case when created_timestamp = lower(timestamp_range) then {0}
                         else coalesce({0}, e_{0})
                    end""";
    private static final String UPSERTABLE_COLUMN_WITH_DEFAULT_COALESCE =
```

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListener.java
```diff
@@ -629,6 +635,34 @@ private Nft mergeNft(Nft cachedNft, Nft newNft) {
         return dest;
     }
 
+    private Hook mergeHook(Hook previous, Hook current) {
+        if (previous != null && previous.getDeleted() && current.getTimestampUpper() == null) {
+            previous.setTimestampUpper(current.getTimestampLower() + 1);
+        }
```
xin-hedera:
```suggestion
       long adjustment = previous.getDelete() && previous.getTimestampLower() == current.getTimestampLower() ? 1 : 0;
       previous.setTimestampUpper(current.getTimestampLower() + adjustment);
```

- previous will never be null
- current.getTimestampUpper will always be null
- there is a special case for deletion and creation of the same hook in the same transaction, so must add `previous.getTimestampLower() == current.getTImestampLower()`

when deletion and creation of the same hook happens in different transactions, `hookDeletion.timestampUpper` should be set to `hookCreation.timestampLower`

think with this change, you'll need to adjust some test cases.

The logic here also tells us there's lack of test coverage or incorrect test cases (either incorrect setup or incorrect expectation). Specifically, when `previous` is hook creation and `current` is hook deletion, the hook creation's timestamp range will not get closed.

importer/src/test/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListenerTest.java
```diff
@@ -3602,6 +3611,283 @@ private Nft getNft(
...
+                .ownerId(ownerId)
+                .timestampRange(Range.atLeast(timestamp))
+                .type(HookType.LAMBDA)
+                .build();
+
+        var hookDelete = Hook.builder()
+                .deleted(true)
+                .hookId(hookId)
+                .ownerId(ownerId)
+                .timestampRange(Range.atLeast(timestamp)) // different timestamp
```
xin-hedera:
```suggestion
                .timestampRange(Range.atLeast(timestamp)) // same timestamp
```

importer/src/test/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListenerTest.java
```diff
@@ -3602,6 +3611,283 @@ private Nft getNft(
...
+        var expectedPrevious = hookDelete.toBuilder().build();
+        expectedPrevious.setTimestampUpper(hookCreate.getTimestampLower() + 1);
+        expectedPrevious.setExtensionPoint(HookExtensionPoint.ACCOUNT_ALLOWANCE_HOOK);
+        expectedPrevious.setType(HookType.LAMBDA);
+        assertThat(hookRepository.findAll()).containsExactly(hookCreate);
+        assertThat(findHistory(Hook.class)).containsExactly(expectedPrevious);
+    }
+
+    @Test
+    void onHookCreateAndDeleteLifecycle() {
```
xin-hedera: The reason why I have asked you to make multiple test cases parameterized: if this test case was parameterized to have the two hook objects

- in one record file
- separately in two record files

It would have caught the issue I pointed out in `SqlEntityListener:mergeHook`: when hook creation and hook deletion are in the same record file, hook creation will not have its timestamp range closed.

importer/src/test/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListenerTest.java
xin-hedera: My ask is not to have different order / different types of transactions in a parameterized test.

What I mean is given the simplest scenario that there are two hook objects as a result of processing two transactions A and B, we want to parameterize it so

- transaction A and B are in the same block
- transaction A is in block N and transaction B is in block N+1

importer/src/test/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListenerTest.java
```diff
@@ -3602,6 +3612,329 @@ private Nft getNft(
...
+        var hookDelete = Hook.builder()
+                .deleted(true)
+                .hookId(hookId)
+                .ownerId(ownerId)
+                .timestampRange(Range.atLeast(createdTimestamp))
+                .build();
+
+        // when
+        sqlEntityListener.onHook(hookDelete);
+        sqlEntityListener.onHook(hookCreate);
```
xin-hedera: method name says `onHookCreateDeleteRecreate`

however the sequence of changes is

- hookDelete
- hookCreate1
- hookCreate2

hookDelete and hookCreate1 have the same timestamp, which means hookDelete and hookCreate1 belong to different lifecycles.

then how is it possible in the happy path have hookCreate1 followed by hookCreat2 for the same hook without a hook delete in between?

importer/src/test/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListenerTest.java
```diff
@@ -3602,6 +3612,329 @@ private Nft getNft(
...
+        completeFileAndCommit();
+
+        // then
+        var expectedCreateHistory = hookCreate.toBuilder().build();
+        expectedCreateHistory.setTimestampUpper(createdTimestamp);
+
+        var expectedDeleteHistory = hookDelete.toBuilder()
+                .contractId(contractId) // merged from creation
+                .createdTimestamp(createdTimestamp) // merged from creation
```
xin-hedera: this is indicates bugs in the `mergeHook` function  if the test case passes. Again hookDeletion is passed to sqlEntityListener before hookCreate, how can it merge these field from the hook's new lifecycle?

importer/src/test/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListenerTest.java
```diff
@@ -3602,6 +3612,329 @@ private Nft getNft(
...
+
+        // when
+        var hookRecreate = hookCreate.toBuilder()
+                .createdTimestamp(createdTimestamp + 100)
+                .adminKey(null)
+                .build();
+        sqlEntityListener.onHook(hookRecreate);
+        completeFileAndCommit();
+
+        assertThat(hookRepository.findAll()).containsExactly(hookRecreate);
```
xin-hedera: need assertions of history table rows

importer/src/test/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListenerTest.java
```diff
@@ -3602,6 +3612,329 @@ private Nft getNft(
...
+                () -> assertEquals(contractId, actualOptional.get().getContractId()),
+                () -> assertEquals(
+                        HookExtensionPoint.ACCOUNT_ALLOWANCE_HOOK,
+                        actualOptional.get().getExtensionPoint()),
+                () -> assertEquals(HookType.LAMBDA, actualOptional.get().getType()),
+                () -> assertEquals(
+                        EntityId.of(ownerId).getId(), actualOptional.get().getOwnerId()),
+                () -> assertEquals(createdTimestamp, actualOptional.get().getCreatedTimestamp()),
+                () -> assertNotNull(actualOptional.get().getTimestampUpper()),
+                () -> assertFalse(actualOptional.get().getDeleted()));
```
xin-hedera: use the same one liner to assert the history object.

importer/src/test/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListenerTest.java
ashumahajan: hookDelete and hookCreate1 happens as part of the same transaction. But I see your point on we can't have hookCreate when hook was never deleted.
I will add another hookDelete in middle to ensure that hook is deleted.

importer/src/test/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListenerTest.java
ashumahajan: I added another testcase to test the usecase. onHookDeleteAndCreateLifecycle I see the delete record both in hook and hook_history table does not have merged fields when there is no create record initially. I don't think that is an issue

importer/src/test/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListenerTest.java
ashumahajan: I have made the testcases parametrized as you requested. But please be aware
1. If there 2 hook creates in in transaction (not possible) merge will be called
2. If there is a hook create and delete in same transaction( first hookDelete will be called and then hookCreate
3. If hookDelete and hookCreate are called separately, the order doesn't matter.

I had created different testcases to support the use cases above.

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListener.java
ashumahajan: I added few more testcases.

Test Create, Delete, Create again in same block but different transactions
Test Delete without Create
Test Create, Delete in same transaction and in different block
Checked merged data and especially adminKey

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListener.java
```diff
@@ -629,6 +635,34 @@ private Nft mergeNft(Nft cachedNft, Nft newNft) {
         return dest;
     }
 
+    private Hook mergeHook(Hook previous, Hook current) {
+        long adjustment =
+                previous.getDeleted() && previous.getTimestampLower().equals(current.getTimestampLower()) ? 1 : 0;
+        previous.setTimestampUpper(current.getTimestampLower() + adjustment);
+
+        if (current.getCreatedTimestamp() != null && current.getCreatedTimestamp() == current.getTimestampLower()) {
```
xin-hedera: should also do equals here?

```suggestion
        if (current.getCreatedTimestamp() != null && current.getCreatedTimestamp().equals(current.getTimestampLower())) {
```

importer/src/test/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListenerTest.java
```diff
@@ -3602,6 +3607,356 @@ private Nft getNft(
...
+                    .deleted(true)
+                    .extensionPoint(HookExtensionPoint.ACCOUNT_ALLOWANCE_HOOK)
+                    .type(HookType.LAMBDA)
+                    .timestampRange(Range.atLeast(deletedTimestamp))
+                    .build();
+            // Merge Changes
+            expectedDelete.setTimestampUpper(deletedTimestamp + 1);
+            assertThat(hookRepository.findAll()).containsExactly(expectedHookCreate);
+            assertThat(findHistory(Hook.class)).containsExactly(expectedDelete);
+        }
```
xin-hedera, line 3727: this is not how we usually test with variable number of blocks.

what the test case does

- commitIndex > 1: hookCreate then hookDelete
- commitIndex = 1: hookDelete then hookCreate

the general guideline for such parameterized tests are

- the objects passed to `sqlEntityListener.onXXX()` are the same and in the same order
- the expected objects at the end, regardless of how many commits, are the same
- if there are more than 1 commit, we also assert intermediate data persisted to database after each commit

importer/src/test/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListenerTest.java
```diff
@@ -3602,6 +3607,356 @@ private Nft getNft(
...
+    @ParameterizedTest
+    @ValueSource(ints = {1, 2})
+    void onHookCreateAndDeleteWithBlockBoundaries(int commitIndex) {
+        // given
+        var ownerId = EntityId.of(domainBuilder.id()).getId();
+        var hookId = 1L;
+        var contractId = EntityId.of(domainBuilder.id());
+        var adminKey = domainBuilder.bytes(32);
+        var createdTimestamp = domainBuilder.timestamp();
+        var deletedTimestamp = commitIndex > 1 ? createdTimestamp + 100 : createdTimestamp;
```
xin-hedera, line 3665: not supposed to test it this way - delete timestamp should not depend on commitIndex

importer/src/test/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListenerTest.java
```diff
@@ -3602,6 +3607,356 @@ private Nft getNft(
...
+        var expectedHookDelete = hookCreate.toBuilder()
+                .deleted(true)
+                .timestampRange(Range.closedOpen(deletedTimestamp, createdTimestamp2))
+                .build();
+        var expectedHookCreate = hookCreate.toBuilder()
+                .timestampRange(Range.closedOpen(createdTimestamp, deletedTimestamp))
+                .build();
+
+        assertThat(hookRepository.findAll()).containsExactly(hookCreate2);
+        assertThat(findHistory(Hook.class)).contains(expectedHookCreate, expectedHookDelete);
```
xin-hedera:
```suggestion
        assertThat(findHistory(Hook.class)).containsExactlyInAnyOrder(expectedHookCreate, expectedHookDelete);
```

importer/src/test/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListenerTest.java
ashumahajan: This testcase is testing where hookDelete and commit happens in same transaction. For the usecase you defined, I had other testcase, please check onHookDeleteAndCreateLifecycle, I also made onHook parameterize

importer/src/test/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListenerTest.java
ashumahajan: delete timestamp depends on commitIndex to ensure that we are following the exact way transactions would be handled.
1. when hookcreate and delete is part of same commit, I want to test them being part of same transaction.
2. when they are seperate, I want to test they are sent with different timestamp

xin-hedera: Understand what the test case did. However, as I pointed out, this is against our common practice, and makes review harder.

PR #12290, opened 2025-10-27
HIP-1056 Left trim contract storage value read and value written 

### LINKED ISSUE #12286: HIP-1056 Blockstream transformer should trim contract state change's valueWritten
### Description

For contract storage statechanges, the value is of type `SlotKey` and the value in it is left-padded per the doc.

```protobuf
/**
 * The value of a contract storage slot. For the EVM, this is a single "word".
 *
 * Because we iterate through all the storage slots for an expired contract
 * when purging it from state, our slot values also include the words of the
 * previous and next keys in this contract's storage "virtual linked list".
 */
message SlotValue {
    /**
     * The EVM value in this slot, left-padded with zeros to form a 256-bit word.
     */
    bytes value = 1;

    /**
     * The word of the previous key in this contract's storage list (if any).
     */
    bytes previous_key = 2;

    /**
     * The word of the next key in this contract's storage list (if any).
     */
    bytes next_key = 3;
}
```

Blockstream transformer copies the `value` as `ContractStateChange.value_written.value`. However in record stream contract state change sidecar record, the value is left trimmed.

To match recordstream, blockstream transformer should left trim the `value`.

### Steps to reproduce

Run blockstream and recordstream differential test. 

### Additional context

_No response_

### Hedera network

testnet

### Version

v0.142.0-SNAPSHOT

### Operating system

None

xin-hedera:
**Description**:

This PR fixes the contract storage value written padding issue detected by differential testing

- Left trim contract storage value read and value written
 
**Related issue(s)**:

Fixes #12286 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/common/src/main/java/org/hiero/mirror/common/util/DomainUtils.java
+++ b/common/src/main/java/org/hiero/mirror/common/util/DomainUtils.java
@@ -341,8 +341,8 @@ public class DomainUtils {
     }
 
     public static ByteString trim(final ByteString data) {
-        if (data == null) {
-            return null;
+        if (data == null || data.isEmpty() || data.byteAt(0) != 0) {
+            return data;
         }
 
         final byte[] value = DomainUtils.toBytes(data);
--- a/importer/src/test/java/org/hiero/mirror/importer/parser/domain/BlockTransactionBuilder.java
+++ b/importer/src/test/java/org/hiero/mirror/importer/parser/domain/BlockTransactionBuilder.java
@@ -885,7 +885,7 @@ public class BlockTransactionBuilder {
     }
 
     private ByteString leftPad(ByteString value) {
-        if (random.nextInt() % 2 == 0) {
+        if (random.nextBoolean()) {
             return value;
         }
 
--- a/common/src/test/java/org/hiero/mirror/common/util/DomainUtilsTest.java
+++ b/common/src/test/java/org/hiero/mirror/common/util/DomainUtilsTest.java
@@ -44,7 +44,7 @@ import org.junit.jupiter.params.provider.MethodSource;
 import org.junit.jupiter.params.provider.NullAndEmptySource;
 import org.junit.jupiter.params.provider.ValueSource;
 
-class DomainUtilsTest {
+final class DomainUtilsTest {
 
     private static final String ED25519_KEY = "60beee88a761e079f71b03b5fe041979369e96450a7455b203a2578c8a7d4852";
     private static final String ECDSA_384_KEY = "0000001365636473612d736861322d6e69737470333834000000086e697374703338"


common/src/main/java/org/hiero/mirror/common/util/DomainUtils.java
```diff
@@ -339,6 +340,28 @@ public static byte[] trim(final byte[] data) {
         return ArrayUtils.subarray(data, i, data.length);
     }
 
+    public static ByteString trim(final ByteString data) {
+        if (data == null) {
+            return null;
+        }
+
```
steven-sheehy, line 347: We could optimize to return data if leftmost byte is not zero or length zero since toBytes could have overhead and will re-wrap empty arrays.

importer/src/test/java/org/hiero/mirror/importer/parser/domain/BlockTransactionBuilder.java
```diff
@@ -882,6 +884,14 @@ private ByteString getEvmAddress(ContractID contractId) {
         return ByteString.EMPTY;
     }
 
+    private ByteString leftPad(ByteString value) {
+        if (random.nextInt() % 2 == 0) {
```
jnels124: can you clarify this? Why do we only want to left pad ~50% of the time. Also would it just be better to do nextBoolean?

jnels124: nevermind this is test.

xin-hedera: yup it's for test. changed it to nextBoolean.

it's a compromise, valueRead is supposed to be left trimmed (however, consensus node may not stick to the spec..), valueWritten is not. so I just went with randomly pad / not pad them.

common/src/main/java/org/hiero/mirror/common/util/DomainUtils.java
```diff
@@ -339,6 +340,28 @@ public static byte[] trim(final byte[] data) {
...
+        final byte[] value = DomainUtils.toBytes(data);
+        final byte[] trimmed = trim(value);
+
+        return trimmed == value ? data : DomainUtils.fromBytes(trimmed);
+    }
+
+    public static BytesValue trim(final BytesValue data) {
+        if (data == null) {
```
steven-sheehy, line 355: Could do same optimization here

xin-hedera: not needed, since it calls `trim(ByteString)` and then do a value equal comparison.

PR #12291, opened 2025-10-28
HIP-1215 Adapt importer to handle changes related to scheduled contract calls

### LINKED ISSUE #11650: HIP-1215: Impact on block aggregation
### Problem

[HIP-1215](https://hips.hedera.com/hip/hip-1215) generalized schedule contract call will change the definition of a top-level transaction to allow non-user transactions with no triggering user transaction. This impacts how we calculate block gas used and logs bloom.

### Solution

* Update `RecordItem.isTopLevel()` to check `transactionId.nonce == 0 || (nonce > 0 && scheduled) || parent_consensus_timestamp == null`
* Work with execution team to generate a test transaction on integration for examination
* Add test for HIP-1215 type transaction in EntityRecordItemListener tests.
* Check other places nonce is used during read and write where we assume nonce=0 is a top-level user transaction.
* Note also that a synthetic `FileUpdate` with payer `0.0.50` and non-zero nonce is now also considered top-level. The above logic should account for that but needs testing.

### Alternatives

_No response_

IvanKavaldzhiev:
**Description**:
This PR changes the way we determine when a recordItem is a top level one, taking into account the changes related to schedule transactions introduced in HIP-1215.

**Related issue(s)**:

Fixes #11650

**Checklist**

### DIFF
--- a/common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
+++ b/common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
@@ -35,6 +35,7 @@ import lombok.ToString;
 import lombok.Value;
 import lombok.experimental.NonFinal;
 import org.apache.commons.codec.binary.Hex;
+import org.hiero.mirror.common.CommonProperties;
 import org.hiero.mirror.common.domain.StreamItem;
 import org.hiero.mirror.common.domain.contract.ContractTransaction;
 import org.hiero.mirror.common.domain.entity.EntityId;
@@ -182,7 +183,12 @@ public class RecordItem implements StreamItem {
 
     // Whether we have a FileUpdate transaction that is paid by the system account 0.0.50
     private boolean isSystemFileUpdate() {
-        return transactionBody.hasFileUpdate() && EntityId.of(0L, 0L, 50L).equals(payerAccountId);
+        return transactionBody.hasFileUpdate()
+                && EntityId.of(
+                                CommonProperties.getInstance().getShard(),
+                                CommonProperties.getInstance().getRealm(),
+                                50L)
+                        .equals(payerAccountId);
     }
 
     /**
--- a/importer/src/test/java/org/hiero/mirror/importer/parser/domain/RecordFileBuilder.java
+++ b/importer/src/test/java/org/hiero/mirror/importer/parser/domain/RecordFileBuilder.java
@@ -4,7 +4,6 @@ package org.hiero.mirror.importer.parser.domain;
 
 import static org.hiero.mirror.importer.domain.StreamFilename.FileType.DATA;
 
-import com.hederahashgraph.api.proto.java.AccountID;
 import com.hederahashgraph.api.proto.java.Timestamp;
 import com.hederahashgraph.api.proto.java.TransactionID;
 import jakarta.inject.Named;
@@ -222,9 +221,7 @@ public class RecordFileBuilder {
                     case CONTRACT_CALL -> {
                         var transactionID = TransactionID.newBuilder()
                                 .setNonce(nonce)
-                                .setAccountID(AccountID.newBuilder()
-                                        .setAccountNum(4100)
-                                        .build())
+                                .setAccountID(recordItemBuilder.accountId())
                                 .setScheduled(scheduled);
 
                         var contractCallItemBuilder =


common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordFile.java
```diff
@@ -145,4 +145,12 @@ private Version hapiVersion() {
 
         return new Version(hapiVersionMajor, hapiVersionMinor, hapiVersionPatch);
     }
+
+    public long getGasUsed() {
+        return gasUsed;
+    }
+
+    public byte[] getLogsBloom() {
+        return logsBloom;
```
steven-sheehy: These shouldn't be needed as Lombok already generates getters.

importer/src/test/java/org/hiero/mirror/importer/parser/domain/RecordFileBuilder.java
```diff
@@ -118,9 +119,13 @@ public class ItemBuilder {
 
         private int count = 100;
         private int entities = 10;
+        private int nonce = 0;
         private boolean entityAutoCreation = false;
+        private boolean isScheduled = false;
```
steven-sheehy: We shouldn't name fields starting with is, only methods.

```suggestion
        private boolean scheduled = false;
```

importer/src/test/java/org/hiero/mirror/importer/parser/domain/RecordItemBuilder.java
```diff
@@ -337,6 +337,28 @@ public Builder<ContractCallTransactionBody.Builder> contractCall() {
         return contractCall(contractId());
     }
 
+    public Builder<ContractCallTransactionBody.Builder> contractCall(
```
steven-sheehy: We should not add another builder method just because we need to set a few custom fields. Use `contractCall().record(r -> r.setParentConsensusTimestamp())...` etc

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
```diff
@@ -172,7 +172,12 @@ public boolean isInvalidIdError() {
 
     // Whether this is a top level, user submitted transaction that could possibly trigger other internal transactions.
     public boolean isTopLevel() {
-        return transactionRecord.getTransactionID().getNonce() == 0;
+        var transactionNonce = transactionRecord.getTransactionID().getNonce();
+
+        return transactionNonce == 0
+                || (transactionNonce > 0 && transactionRecord.getTransactionID().getScheduled())
+                || (transactionRecord.getParentConsensusTimestamp().getSeconds() == 0
+                        && transactionRecord.getParentConsensusTimestamp().getNanos() == 0);
```
steven-sheehy:
```suggestion
                || !transactionRecord.hasParentConsensusTimestamp();
```

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
```diff
@@ -172,7 +172,12 @@ public boolean isInvalidIdError() {
 
     // Whether this is a top level, user submitted transaction that could possibly trigger other internal transactions.
     public boolean isTopLevel() {
```
steven-sheehy, line 175: Add or update a test in `RecordItemTest`.

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
```diff
@@ -172,7 +172,12 @@ public boolean isInvalidIdError() {
...
-        return transactionRecord.getTransactionID().getNonce() == 0;
+        var transactionNonce = transactionRecord.getTransactionID().getNonce();
+
+        return transactionNonce == 0
+                || (transactionNonce > 0 && transactionRecord.getTransactionID().getScheduled())
```
steven-sheehy, line 179: Do we know if this logic will work retroactively for historical records or do we need to check the proto version and run it conditionally?

importer/src/test/java/org/hiero/mirror/importer/parser/domain/RecordFileBuilder.java
```diff
@@ -118,9 +119,13 @@ public class ItemBuilder {
...
         private SubType subType = SubType.STANDARD;
         private TransactionType type = TransactionType.UNKNOWN;
+        private Timestamp parentConsensusTimestamp =
+                Timestamp.newBuilder().setSeconds(0L).build();
```
steven-sheehy: I don't think we need add explicit fields on ItemBuilder for every field. We can just use `itembuilder.record(r -> r.setParentConsensusTimestamp())...` etc

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordFile.java
IvanKavaldzhiev: Right, for some reason initially the getters were not working. Removed.

importer/src/test/java/org/hiero/mirror/importer/parser/domain/RecordFileBuilder.java
IvanKavaldzhiev: Renamed

importer/src/test/java/org/hiero/mirror/importer/parser/domain/RecordItemBuilder.java
IvanKavaldzhiev: Agree, removed

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
IvanKavaldzhiev: Refactored

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
IvanKavaldzhiev: Added

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
IvanKavaldzhiev: If you are referring to the `scheduled` field in the protobuf - each newly added proto type should be backward compatible. So if we have an older recordFile with this proto field missing, it should return a default value (in this case false), so the check will still be valid.

importer/src/test/java/org/hiero/mirror/importer/parser/domain/RecordFileBuilder.java
IvanKavaldzhiev: Initialized the field with the default proto value. The idea is initially the parentConsensusTimestamp to be empty.

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
```diff
@@ -172,7 +172,17 @@ public boolean isInvalidIdError() {
...
+                || !transactionRecord.hasParentConsensusTimestamp()
+                || isSystemFileUpdate();
+    }
+
+    // Whether we have a FileUpdate transaction that is paid by the system account 0.0.50
+    private boolean isSystemFileUpdate() {
+        return transactionBody.hasFileUpdate() && EntityId.of(0L, 0L, 50L).equals(payerAccountId);
```
steven-sheehy: Need to use CommonProperties.getInstance() to use correct shard and realm.

importer/src/test/java/org/hiero/mirror/importer/parser/domain/RecordFileBuilder.java
```diff
@@ -197,6 +219,24 @@ private Supplier<RecordItemBuilder.Builder<?>> buildTemplate() {
             if (subType != SubType.STANDARD) {
                 return switch (subType) {
                     case TOKEN_TRANSFER -> () -> recordItemBuilder.cryptoTransfer(TransferType.TOKEN);
+                    case CONTRACT_CALL -> {
+                        var transactionID = TransactionID.newBuilder()
+                                .setNonce(nonce)
+                                .setAccountID(AccountID.newBuilder()
+                                        .setAccountNum(4100)
+                                        .build())
```
steven-sheehy: Shouldn't hardcode shard, realm, and num.

```suggestion
                                .setAccountID(recordItemBuilder.accountId())
```

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
IvanKavaldzhiev: Resolved

importer/src/test/java/org/hiero/mirror/importer/parser/domain/RecordFileBuilder.java
IvanKavaldzhiev: Resolved

PR #12294, opened 2025-10-28
Relax condition for valid start timestamp in transactions API

### LINKED ISSUE #12248: Rest API may return 404 for system synthetic transactions by transaction id
### Description

For a couple of releases, consensus node may externalize system synthetic transactions with valid start timestamp > consensus timestamp. This causes 404 from `/api/v1/transactions/{transactionId}` Rest API, for example

- transactions by transaction id returns 404: https://testnet.mirrornode.hedera.com/api/v1/transactions/0.0.50-1733335386-501910000
- transactions by explicit consensus timestamp range: https://testnet.mirrornode.hedera.com/api/v1/transactions?timestamp=gte:1733335386.501909996&timestamp=lte:1733335386.501909999

While consensus node will fix the issue by `wind back` such transaction's valid start timestamp, mirror node still needs to relax the condition for such historical synthetic transactions.

### Steps to reproduce

check links in the description.

### Additional context

_No response_

### Hedera network

testnet

### Version

v0.141.0

### Operating system

None

martingeorgiev1:
**Description**:

Relax the condition for valid start timestamp

* Relaxed the condition by 8s`(maxRecordFileCloseInterval)`
* Fixed unit test to correspond with that change
* Replaced hardcoded timestamps with variables for improved readability

**Related issue(s)**:

Fixes #12248 

**Notes for reviewer**:

**Checklist**


rest/transactions.js
```diff
@@ -836,7 +836,11 @@ const getTransactionsByTransactionIdsSql = (transactionKeys, filters, timestampF
     `${Transaction.CONSENSUS_TIMESTAMP} <= $${params.length + 2}`,
   ];
   params.lowerConsensusTimestampIndex = params.length;
-  params.push(minTimestamp);
+  params.push(
+    timestampField === Transaction.VALID_START_NS
+      ? minTimestamp - config.query.maxRecordFileCloseIntervalNs
```
xin-hedera: Think we should add a new config property, e.g., `maxValidStartTimestampDrift` and the default should be `1s`. `maxRecordFileCloseInterval` is a different concept, should not use it for valid start timestamp.

rest/__tests__/transactions.test.js
```diff
@@ -481,8 +485,11 @@ const addTimestampIndices = (arr, minIndex, maxIndex) => {
 describe('extractSqlFromTransactionsByIdOrHashRequest', () => {
   describe('success', () => {
     const defaultTransactionIdStr = '0.0.200-123456789-987654321';
-    const defaultParams = addTimestampIndices([200n, 123456789987654321n, 123456789987654321n, 123458889987654321n]);
-    const transactionKeys = [200n, 123456789987654321n];
+    const validStartNs = 123456789987654321n;
```
xin-hedera, line 488: can you also add a spec based test to `__tests__/specs/transactions/{id}/no-params.json`. You can add it to the `tests[]` array with a different transaction id and the transaction has consensus timestamp < valid start timestamp.

PR #12306, opened 2025-10-29
Update acceptance workflow to use 0.67.2-rc.2 consensus tag to fix failing acceptance tests

IvanKavaldzhiev:
**Description**:
Currently the acceptance workflow uses a consensus node tag that utilizes the changes from HIP-1249, having 100% gasRefundPercentage. However the mirror node importer and feature properties changed their minimum hapi minor version for these change to be 0.68. This causes inconsistencies in importer and acceptance test behaviour, since we have record files with 0.67.0 version with HIP-1249 behaviour enabled.

This PR changes the consensus version to be 0.67.2 with HIP-1249 disabled and overrides the importer to use 0.68.0 as a minimum hapi throttling version, so that acceptance tests and importer logic match.

This PR updates consensus node to a version which has HIP-1249 disabled and this should fix the failing tests.

**Related issue(s)**:

**Notes for reviewer**:

Fixes #

**Checklist**

### DIFF
--- a/.github/workflows/acceptance.yaml
+++ b/.github/workflows/acceptance.yaml
@@ -119,7 +119,7 @@ jobs:
             cat <<EOF > mirror.yaml
           importer:
             env:
-              HIERO_MIRROR_IMPORTER_SMARTCONTRACTTHROTTLINGVERSION: "0.67.0"
+              HIERO_MIRROR_IMPORTER_SMARTCONTRACTTHROTTLINGVERSION: "0.68.0"
               HIERO_MIRROR_IMPORTER_BLOCK_NODES_0_HOST: 'block-node-1.${SOLO_NAMESPACE}.svc.cluster.local'
               SPRING_PROFILES_ACTIVE: 'blocknode'
           test:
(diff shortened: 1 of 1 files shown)


xin-hedera:
We'll need next mirrornode RC tag with HIP-1249 version updated to 0.68 in importer. right now there is a consensus node version mismatch between importer (0.67) and acceptance test (0.68).

.github/workflows/acceptance.yaml
```diff
@@ -43,7 +43,7 @@ jobs:
     env:
       BASE_URL: "http://127.0.0.1:8081"
       BLOCK_NODE_CHART_VERSION: v0.20.1
-      CONSENSUS_VERSION: v0.67.0
+      CONSENSUS_VERSION: v0.67.2-rc.2
```
steven-sheehy: We should be able to use v0.67.2
```suggestion
      CONSENSUS_VERSION: v0.67.2
```

IvanKavaldzhiev: FIxed

.github/workflows/acceptance.yaml
```diff
@@ -119,6 +119,7 @@ jobs:
             cat <<EOF > mirror.yaml
           importer:
             env:
+              HIERO_MIRROR_IMPORTER_SMARTCONTRACTTHROTTLINGVERSION: "0.67.0"
```
xin-hedera: think needs to be `"0.68.0"`

the default in the importer image is 0.67.0

xin-hedera: and the default in acceptance test in `main` branch is minor version 68.

IvanKavaldzhiev: You're right, I fixed it.

PR #12307, opened 2025-10-29
HIP-1195 Add LambdaSStoreTransaction support

### LINKED ISSUE #12118: HIP-1195 Add `LambdaSStoreTransaction` support

jnels124:
**Description**:
this pr adds support for the new lambda sstore transaction
* add new transaction handler
* logic to persist `HookStorage` and `HookStorageChange`
* adds new temp table `hook_storage`

**Related issue(s)**:

Fixes #12118 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/common/src/main/java/org/hiero/mirror/common/domain/hook/HookStorageChange.java
+++ b/common/src/main/java/org/hiero/mirror/common/domain/hook/HookStorageChange.java
@@ -13,7 +13,6 @@ import lombok.Builder;
 import lombok.Data;
 import lombok.NoArgsConstructor;
 import lombok.ToString;
-import org.apache.commons.lang3.ArrayUtils;
 import org.hiero.mirror.common.util.DomainUtils;
 import org.springframework.data.domain.Persistable;
 


importer/src/test/java/org/hiero/mirror/importer/parser/domain/RecordItemBuilder.java
```diff
@@ -1235,6 +1238,22 @@ private Builder<?> toCreateTransaction(Map.Entry<GeneratedMessage, EntityState>
...
 
+    public Builder<LambdaSStoreTransactionBody.Builder> lambdaSstore() {
+
+        var contractId = contractId();
+        var body = LambdaSStoreTransactionBody.newBuilder()
+                .setHookId(HookId.newBuilder()
+                        .setHookId(id())
+                        .setEntityId(HookEntityId.newBuilder()
+                                .setAccountId(accountId())
+                                .setContractId(contractId)
```
steven-sheehy: This is a oneof so we should only set one.

importer/src/test/java/org/hiero/mirror/importer/parser/domain/RecordItemBuilder.java
```diff
@@ -1235,6 +1238,22 @@ private Builder<?> toCreateTransaction(Map.Entry<GeneratedMessage, EntityState>
         };
     }
 
+    public Builder<LambdaSStoreTransactionBody.Builder> lambdaSstore() {
+
+        var contractId = contractId();
+        var body = LambdaSStoreTransactionBody.newBuilder()
+                .setHookId(HookId.newBuilder()
```
steven-sheehy, line 1275: We need to populate storage updates in the transaction body. As per neeha, storage updates won't even be in the sidecar.

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/LambdaSStoreTransactionHandler.java
```diff
@@ -0,0 +1,57 @@
...
+    public EntityId getEntity(RecordItem recordItem) {
+        if (!recordItem.getTransactionBody().hasLambdaSstore()) {
+            return EntityId.EMPTY;
+        }
+        return EntityId.of(recordItem
+                .getTransactionBody()
+                .getLambdaSstore()
+                .getHookId()
+                .getEntityId()
+                .getAccountId());
```
steven-sheehy: The entityid is a oneof so we need to support either contractid or accountid.

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/LambdaSStoreTransactionHandler.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    @Override
+    protected void doUpdateTransaction(Transaction txn, RecordItem recordItem) {
+        final var transactionBody = recordItem.getTransactionBody();
+
+        if (!transactionBody.hasLambdaSstore()) {
+            log.warn("no lambda sstore in transaction body consensus_timestamp={}", recordItem.getConsensusTimestamp());
```
steven-sheehy: we can remove this as we guarantee elsewhere that this is only called for this type.

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/LambdaSStoreTransactionHandler.java
```diff
@@ -0,0 +1,57 @@
...
+            return;
+        }
+
+        final var sstore = transactionBody.getLambdaSstore();
+        final var owner = getEntity(recordItem);
+        final var ownerId = owner.getId();
```
steven-sheehy:
```suggestion
        final var ownerId = transaction.getEntityId().getId();
```

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/LambdaSStoreTransactionHandler.java
```diff
@@ -0,0 +1,57 @@
...
+    @Override
+    protected void doUpdateTransaction(Transaction txn, RecordItem recordItem) {
+        final var transactionBody = recordItem.getTransactionBody();
+
+        if (!transactionBody.hasLambdaSstore()) {
+            log.warn("no lambda sstore in transaction body consensus_timestamp={}", recordItem.getConsensusTimestamp());
+            return;
+        }
+
+        final var sstore = transactionBody.getLambdaSstore();
```
steven-sheehy: We usually name this `transactionBody` in the handlers.

```suggestion
        final var transactionBody = transactionBody.getLambdaSstore();
```

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/LambdaSStoreTransactionHandler.java
```diff
@@ -0,0 +1,57 @@
...
+    @Override
+    public TransactionType getType() {
+        return TransactionType.LAMBDA_SSTORE;
+    }
+
+    @Override
+    public EntityId getEntity(RecordItem recordItem) {
+        if (!recordItem.getTransactionBody().hasLambdaSstore()) {
+            return EntityId.EMPTY;
+        }
```
steven-sheehy: Not needed

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/LambdaSStoreTransactionHandler.java
```diff
@@ -0,0 +1,57 @@
...
+            return;
+        }
+
+        final var sstore = transactionBody.getLambdaSstore();
+        final var owner = getEntity(recordItem);
+        final var ownerId = owner.getId();
+        final var hookId = sstore.getHookId().getHookId();
+        final var consensusTimestamp = recordItem.getConsensusTimestamp();
+
+        storageHandler.processStorageUpdates(consensusTimestamp, hookId, ownerId, recordItem.getSidecarRecords());
```
steven-sheehy: Does this need to be separate so it can be reused for inline storage updates in crud tx? If so, we can probably place into same EvmHookHandler in other PR once that is merged. If not reused, then we should put logic inline here.

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListener.java
```diff
@@ -206,6 +210,22 @@ public void onFileData(FileData fileData) {
...
+        final var hookStorage = HookStorage.builder()
+                .createdTimestamp(storageChange.getConsensusTimestamp())
+                .hookId(storageChange.getHookId())
+                .ownerId(storageChange.getOwnerId())
+                .key(storageChange.getKey())
+                .modifiedTimestamp(storageChange.getConsensusTimestamp())
+                .value(storageChange.getValueWritten())
+                .build();
+
+        context.merge(hookStorage.getId(), hookStorage, this::mergeHookStorage);
```
steven-sheehy: Should we only merge if valueWritten is present like we do for contract state?

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/LambdaSStoreTransactionHandler.java
jnels124: yes it is reused. Can wait until other pr gets in and move it there

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListener.java
jnels124: I was doing that initially but I removed it as didn't see any indication anywhere that it couldn't be overridden to null.

importer/src/test/java/org/hiero/mirror/importer/parser/domain/RecordItemBuilder.java
jnels124: yes did that initially but yesterday was told there would be a sidecar contract state change that would be sent for this specific transaction. Will re-write again

importer/src/test/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListenerTest.java
```diff
@@ -3486,6 +3491,94 @@ void onEthereumTransactionWFileId() {
...
+                .ownerId(hookStorageChange.getOwnerId())
+                .value(hookStorageChange.getValueWritten())
+                .build();
+
+        assertThat(hookStorageChangeRepository.findAll()).containsExactly(hookStorageChange);
+        assertThat(hookStorageRepository.findAll()).containsExactly(expectedHookStorage);
+    }
+
+    @Test
+    void onHookStorageChangeMerge() {
```
ashumahajan: Like other parameterized testcases, should we also make this parameterized to test merge completely. within same record file and in a separate record file

importer/src/test/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListenerTest.java
```diff
@@ -3486,6 +3491,94 @@ void onEthereumTransactionWFileId() {
...
+                .modifiedTimestamp(hookStorageChange2.getConsensusTimestamp())
+                .ownerId(hookStorageChange.getOwnerId())
+                .value(hookStorageChange2.getValueWritten())
+                .build();
+
+        assertThat(hookStorageChangeRepository.findAll())
+                .containsExactlyInAnyOrder(hookStorageChange, hookStorageChange2);
+        assertThat(hookStorageRepository.findAll()).containsExactly(expectedHookStorage);
+    }
+
```
ashumahajan, line 3606: We should also test hook storage deletion. What happens when valueWritten is not present

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListener.java
```diff
@@ -212,6 +216,23 @@ public void onHook(Hook hook) {
         context.merge(hook.getId(), hook, this::mergeHook);
     }
 
+    @Override
+    public void onHookStorageChange(HookStorageChange storageChange) throws ImporterException {
+        context.add(storageChange);
+
+        if (storageChange.getValueWritten() != null) {
```
jnels124, line 223: Not sure if some other processing needs to occur if this is null?

ashumahajan: If valueWritten is null, it should be treated as delete. So there is no change just the current hook_storage.value should be empty.

jnels124: so should we update the `hook_storage` value to null (and make this column nullable), set to empty byte array, or leave as is?

importer/src/test/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListenerTest.java
jnels124: will look at this

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListener.java
xin-hedera: I think what you implemented is correct.

also want to note we should not set valueWritten to empty when valueWritten is null in our domain object, because in the happy path such `StorageChange` is from a read only operation. read only operation is not supposed to alter the value.

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListener.java
```diff
@@ -212,6 +216,23 @@ public void onHook(Hook hook) {
...
+            final var hookStorage = HookStorage.builder()
+                    .createdTimestamp(storageChange.getConsensusTimestamp())
+                    .hookId(storageChange.getHookId())
+                    .ownerId(storageChange.getOwnerId())
+                    .key(storageChange.getKey())
+                    .modifiedTimestamp(storageChange.getConsensusTimestamp())
+                    .value(storageChange.getValueWritten())
+                    .build();
+            context.merge(hookStorage.getId(), hookStorage, this::mergeHookStorage);
```
xin-hedera, line 233: think we want to add a new `public <T> void put(T object, Object key)` method to `ParserContext` and it finds the `Map` object for domain class `T` and delegate to `Map.put(key, value)`

The reason is for hook storage, due the fact that the same hook identifier by `(ownerId, hookId)` and storage key can exist in many different lifecycles, and importer needs to treat the first time such storage is created in the current lifecycle as its created timestamp, not the created timestamp from previous lifecycles.

As a result, not like the way we merge contract storage, we want to keep its `createdTimestamp` and throw away any previous value in the map, which means, we don't need to merge at all.

In the domain object, for `createdTimestamp`, we need a custom UpsertColumn annotation with the following SQL

```sql
case when length(e_value) = 0 then created_timestamp
          else e_created_timestamp
end
```

common/src/main/java/org/hiero/mirror/common/domain/hook/HookStorage.java
```diff
@@ -27,6 +28,7 @@ public class HookStorage {
 
     private static final int KEY_BYTE_LENGTH = 32;
 
+    @Column(updatable = false)
```
xin-hedera: unfortunately it is updatable since a hook can be deleted, and later a hook with the same hook id can get created. The implication is the slots will get removed, and created later, and they technically belong to different hooks. It'll be bizarre  for us to show that a hook created at T1 has storage slots created before T1.

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/LambdaSStoreTransactionHandler.java
```diff
@@ -0,0 +1,48 @@
...
+    @Override
+    public EntityId getEntity(RecordItem recordItem) {
+        final var body = recordItem.getTransactionBody().getLambdaSstore();
+        final var hookId = body.getHookId().getEntityId();
+
+        if (hookId.hasAccountId()) {
+            return EntityId.of(hookId.getAccountId());
+        }
+
+        return EntityId.of(hookId.getContractId());
```
steven-sheehy: We need to use EntityIdService going forward

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListener.java
xin-hedera: per offline discussion

- the `put` method won't work so that's a bad suggestion. instead `mergeHookStorage` needs to detect when a hook storage is deleted and sets correct created timestamp accordingly. btw will add the feedback to where it belongs
- the custom SQL needs to consider when e_value is `null`, so it should be

```sql
case when coalesce(length(e_value), 0) = 0 then created_timestamp
     else e_created_timestamp
end
```

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListener.java
```diff
@@ -887,6 +908,12 @@ private Topic mergeTopic(Topic previous, Topic current) {
         return current;
     }
 
+    private HookStorage mergeHookStorage(HookStorage previous, HookStorage current) {
+        previous.setValue(current.getValue());
+        previous.setModifiedTimestamp(current.getModifiedTimestamp());
```
xin-hedera, line 923:
```suggestion
        if (ArrayUtils.isEmpty(previous.getValue())) {
          previous.setCreatedTimestamp(current.getCreatedTimestamp());
        }

        previous.setValue(current.getValue());
        previous.setModifiedTimestamp(current.getModifiedTimestamp());
```

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/EVMHookHandler.java
```diff
@@ -163,4 +188,62 @@ private HookType translateHookType(HookCase hookCase) {
...
         };
     }
+
+    private void processMappingEntries(LambdaMappingEntries entries, long consensusTs, long ownerId, long hookId) {
+        final var mappingSlot = entries.getMappingSlot().toByteArray();
+
+        for (final var entry : entries.getEntriesList()) {
+            final var mappingKey = entry.hasKey()
+                    ? leftPad32(entry.getKey().toByteArray())
+                    : keccak256(entry.getPreimage().toByteArray());
```
xin-hedera:
```suggestion
                    : keccak256(DomainUtils.toBytes(entry.getPreimage()));
```

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/EVMHookHandler.java
```diff
@@ -163,4 +188,62 @@ private HookType translateHookType(HookCase hookCase) {
...
+        final var derivedSlot = new byte[mappingSlot.length + key.length];
+        System.arraycopy(mappingSlot, 0, derivedSlot, 0, mappingSlot.length);
+        System.arraycopy(key, 0, derivedSlot, mappingSlot.length, key.length);
+        return keccak256(derivedSlot);
+    }
+
+    static byte[] keccak256(byte[] input) {
+        final var d = new Keccak.Digest256();
+        d.update(input, 0, input.length);
+        return d.digest();
```
xin-hedera: nit,

```suggestion
        return d.digest(input);
```

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/EVMHookHandler.java
```diff
@@ -163,4 +188,62 @@ private HookType translateHookType(HookCase hookCase) {
...
+        change.setValueWritten(valueWritten);
+        entityListener.onHookStorageChange(change);
+    }
+
+    private static byte[] deriveMappingSlot(byte[] mappingSlot, byte[] key) {
+        final var derivedSlot = new byte[mappingSlot.length + key.length];
+        System.arraycopy(mappingSlot, 0, derivedSlot, 0, mappingSlot.length);
+        System.arraycopy(key, 0, derivedSlot, mappingSlot.length, key.length);
+        return keccak256(derivedSlot);
+    }
```
xin-hedera: it's cleaner to just do

```suggestion
    private static byte[] keccak256(byte[] mappingSlot, byte[] key) {
        final var d = new Keccak.Digest256();
        d.update(mappingSlot);
        return d.digest(key);
    }
```

It's brittle to allocate a byte array, copy the two parts, then hash it.

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/EVMHookHandler.java
```diff
@@ -163,4 +188,62 @@ private HookType translateHookType(HookCase hookCase) {
...
+
+    static byte[] keccak256(byte[] input) {
+        final var d = new Keccak.Digest256();
+        d.update(input, 0, input.length);
+        return d.digest();
+    }
+
+    static byte[] leftPad32(byte[] in) {
```
xin-hedera: we have a `DomainUtils.leftPadBytes` utility method.

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/LambdaSStoreTransactionHandler.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    @Override
+    protected void doUpdateTransaction(Transaction transaction, RecordItem recordItem) {
+        final var transactionBody = recordItem.getTransactionBody().getLambdaSstore();
```
xin-hedera, line 47: should only process successful transactions.

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/LambdaSStoreTransactionHandler.java
```diff
@@ -0,0 +1,48 @@
...
+
+    @Override
+    public TransactionType getType() {
+        return TransactionType.LAMBDA_SSTORE;
+    }
+
+    @Override
+    public EntityId getEntity(RecordItem recordItem) {
+        final var body = recordItem.getTransactionBody().getLambdaSstore();
+        final var hookId = body.getHookId().getEntityId();
```
xin-hedera: nit

```suggestion
        final var hookEntityId = body.getHookId().getEntityId();
```

can also fold the 2 lines into 1, since `body` isn't used anywhere else.

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/EVMHookHandler.java
```diff
@@ -163,4 +188,62 @@ private HookType translateHookType(HookCase hookCase) {
             }
         };
     }
+
+    private void processMappingEntries(LambdaMappingEntries entries, long consensusTs, long ownerId, long hookId) {
+        final var mappingSlot = entries.getMappingSlot().toByteArray();
```
xin-hedera:
```suggestion
        final var mappingSlot = DomainUtils.toBytes(entries.getMappingSlot());
```

In addition, `mappingSlot` needs to be left-padded to 32-bytes as well per consensus node implementation: https://github.com/hiero-ledger/hiero-consensus-node/blob/15d522deee039bd63d1a76144e3467910c186165/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/state/WritableEvmHookStore.java#L86

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListener.java
```diff
@@ -212,6 +216,23 @@ public void onHook(Hook hook) {
...
+    @Override
+    public void onHookStorageChange(HookStorageChange storageChange) throws ImporterException {
+        context.add(storageChange);
+
+        if (storageChange.getValueWritten() != null) {
+            final var hookStorage = HookStorage.builder()
+                    .createdTimestamp(storageChange.getConsensusTimestamp())
+                    .hookId(storageChange.getHookId())
+                    .ownerId(storageChange.getOwnerId())
+                    .key(storageChange.getKey())
```
xin-hedera, line 228: we need to override `HookStorage.setKey(byte[] key)` the same way as `ContractState.setSlot(byte[] key)` to left pad to 32 bytes for

- key from contract state change sidecar
- key from `LambdaSStoreTransactionBody.storage_updates.storage_slot.key`

Per protobuf doc, both sources have key left trimmed.

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListener.java
jnels124: per the protobuff implementation it appears valueWritten should always be set. It is initialized in the constructor, the clear method, and setter throws NPE if trying to set to null

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListener.java
jnels124: this method is already overridden 
```
public void setKey(byte[] key) {
        this.key = DomainUtils.leftPadBytes(key, KEY_BYTE_LENGTH);
    }
```

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/EVMHookHandler.java
jnels124: Just to clarify:
```
private static byte[] keccak256(byte[] mappingSlot, byte[] key) {
        final var d = new Keccak.Digest256();
        d.update(mappingSlot);
        return d.digest(key);
    }
```    
should be 
```
    private static byte[] keccak256(byte[] mappingSlot, byte[] key) {
        final var d = new Keccak.Digest256();
        d.update(key);
        return d.digest(mappingSlot);
    }
``` 
right?

xin-hedera: yup you are right, `key` first, then `mappingSlot`

common/src/main/java/org/hiero/mirror/common/domain/hook/HookStorage.java
```diff
@@ -24,14 +25,24 @@
 @NoArgsConstructor
 @Upsertable
 public class HookStorage {
+    private static final String CREATED_TS_COALESCE =
+            """
+                    case when coalesce(e_deleted, true) and NOT deleted then abs(created_timestamp)
+                         when created_timestamp < 0 and NOT deleted then abs(created_timestamp)
```
jnels124: I will remove the `and NOT deleted` here on L31

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListener.java
xin-hedera: do you mean the `bytes value` from the `LambdaSStoreTransactionBody` transaction?

We still need the `if (storageChange.getValueWritten() != null)` check, since this will be called for `HookStorageChange` constructed from contract state change sidecar, in that path `valueWritten` in the protobuf can be left unset, i.e., `storageChange.valueWritten` is `null`

jnels124: yes was only considering sstore transaction. I have corrected.

ashumahajan (approved):
LGTM

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/EVMHookHandler.java
```diff
@@ -163,4 +192,49 @@ private HookType translateHookType(HookCase hookCase) {
...
+
+    private void persistChange(long ownerId, long hookId, byte[] key, byte[] valueWritten, long consensusTs) {
+        final var change = new HookStorageChange();
+        change.setConsensusTimestamp(consensusTs);
+        change.setOwnerId(ownerId);
+        change.setHookId(hookId);
+        change.setKey(key);
+        change.setValueRead(valueWritten);
+        change.setValueWritten(valueWritten);
+        change.setDeleted(ArrayUtils.isEmpty(valueWritten));
```
xin-hedera: it's better to set `deleted` to false as default in the domain class, and set it accordingly inside `HookStorage.setValueWritten`

jnels124: The domain class uses `boolean` instead of `Boolean` so that piece is already being done. Typically i prefer to not do additional work on other fields inside a setter but I can make that change if preferred. I would prefer to add a method such as `updateStorageState` or similar.

jnels124: ~~I guess the alternate naming there doesn't work well with the builder pattern though~~ so will change as suggested

common/src/main/java/org/hiero/mirror/common/domain/hook/HookStorage.java
```diff
@@ -44,6 +53,19 @@ public class HookStorage {
     @ToString.Exclude
     private byte[] value;
 
+    @Builder(toBuilder = true)
+    private HookStorage(
+            long createdTimestamp, long hookId, byte[] key, long modifiedTimestamp, long ownerId, byte[] value) {
+        this.createdTimestamp = createdTimestamp;
+        this.hookId = hookId;
+        this.key = DomainUtils.leftPadBytes(key, KEY_BYTE_LENGTH);
+        ;
```
steven-sheehy:
```suggestion
```

importer/src/test/java/org/hiero/mirror/importer/parser/record/transactionhandler/LambdaSStoreTransactionHandlerTest.java
```diff
@@ -0,0 +1,142 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package org.hiero.mirror.importer.parser.record.transactionhandler;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.*;
```
steven-sheehy: Shouldn't use wildcard imports

importer/src/test/java/org/hiero/mirror/importer/parser/record/transactionhandler/LambdaSStoreTransactionHandlerTest.java
```diff
@@ -0,0 +1,142 @@
...
+import org.hiero.mirror.common.domain.entity.EntityType;
+import org.hiero.mirror.common.domain.transaction.RecordItem;
+import org.hiero.mirror.common.domain.transaction.Transaction;
+import org.hiero.mirror.common.domain.transaction.TransactionType;
+import org.hiero.mirror.importer.parser.domain.RecordItemBuilder;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.mockito.ArgumentCaptor;
+
+class LambdaSStoreTransactionHandlerTest extends AbstractTransactionHandlerTest {
```
steven-sheehy: Can be final

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/EVMHookHandler.java
```diff
@@ -46,6 +56,24 @@ void process(
...
+    public void processStorageUpdates(
+            long consensusTimestamp, long hookId, long ownerId, List<LambdaStorageUpdate> storageUpdates) {
+        for (final var update : storageUpdates) {
+            switch (update.getUpdateCase()) {
+                case STORAGE_SLOT ->
+                    processStorageSlotUpdate(update.getStorageSlot(), consensusTimestamp, ownerId, hookId);
+                case MAPPING_ENTRIES ->
+                    processMappingEntries(update.getMappingEntries(), consensusTimestamp, ownerId, hookId);
+                default ->
+                    log.warn(
```
steven-sheehy: Use Utility.recoverableError

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/EVMHookHandler.java
```diff
@@ -163,4 +191,48 @@ private HookType translateHookType(HookCase hookCase) {
...
+                    ? leftPadBytes(toBytes(entry.getKey()), 32)
+                    : keccak256(toBytes(entry.getPreimage()));
+
+            final var valueWritten = toBytes(entry.getValue());
+            final var derivedSlot = keccak256(mappingKey, mappingSlot);
+            persistChange(ownerId, hookId, derivedSlot, valueWritten, consensusTs);
+        }
+    }
+
+    private void processStorageSlotUpdate(LambdaStorageSlot storageSlot, long consensusTs, long ownerId, long hookId) {
```
steven-sheehy: Prefer not to shorten.

```suggestion
    private void processStorageSlotUpdate(LambdaStorageSlot storageSlot, long consensusTimestamp, long ownerId, long hookId) {
```

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/LambdaSStoreTransactionHandler.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    @Override
+    protected void doUpdateTransaction(Transaction transaction, RecordItem recordItem) {
+        if (!recordItem.isSuccessful()) {
+            return;
+        }
+
+        final var transactionBody = recordItem.getTransactionBody().getLambdaSstore();
+        final var ownerId = transaction.getEntityId().getId();
```
steven-sheehy: I'd prefer we use EntityId to pass between components like `EvmHookHandler`, if possible.

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/EVMHookHandler.java
```diff
@@ -163,4 +191,48 @@ private HookType translateHookType(HookCase hookCase) {
...
+        final var slotKey = toBytes(storageSlot.getKey());
+
+        // Protobuff API ensures that value will never be null
```
steven-sheehy: 💪🏻 
```suggestion
        // Protobuf API ensures that value will never be null
```

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/EVMHookHandler.java
```diff
@@ -163,4 +191,48 @@ private HookType translateHookType(HookCase hookCase) {
...
+        final var valueWritten = toBytes(storageSlot.getValue());
```
steven-sheehy, line 213: I believe historically storage values coming from sidecar will be optimized by CN to remove left padded zeros to save size in stream (and indirectly our database size). Here, these storage values come from the users so they may or may not have left padded zeros (not sure what HAPI requires). Regardless, we should remove left-padded zeros to match sidecar values and to save storage. Of course the key must remain padded to be sortable.

common/src/main/java/org/hiero/mirror/common/domain/hook/HookStorageChange.java
```diff
@@ -42,6 +45,18 @@ public class HookStorageChange implements Persistable<HookStorageChange.Id> {
...
+    @Builder(toBuilder = true)
+    private HookStorageChange(
+            long consensusTimestamp, long hookId, byte[] key, long ownerId, byte[] valueRead, byte[] valueWritten) {
+        this.consensusTimestamp = consensusTimestamp;
+        this.hookId = hookId;
+        this.key = key;
+        this.ownerId = ownerId;
+        this.valueRead = DomainUtils.trim(valueRead);
+        this.valueWritten = DomainUtils.trim(valueWritten);
+        this.deleted = ArrayUtils.isEmpty(this.valueWritten);
```
xin-hedera: `valueWritten` can be null, and it means it's a read-only operation, so `deleted` should be false.

common/src/main/java/org/hiero/mirror/common/domain/hook/HookStorageChange.java
```diff
@@ -53,6 +68,15 @@ public Id getId() {
         return id;
     }
 
+    public void setValueRead(byte[] valueRead) {
+        this.valueRead = DomainUtils.trim(valueRead);
+    }
+
+    public void setValueWritten(byte[] valueWritten) {
+        this.valueWritten = DomainUtils.trim(valueWritten);
+        this.deleted = ArrayUtils.isEmpty(this.valueWritten);
```
xin-hedera: should apply the same change here: a null valueWritten should set deleted to `false`

jnels124: done

xin-hedera (approved):
LGTM

PR #12319, opened 2025-10-31
k6 performance tests for `/opcodes` web3 endpoint

### LINKED ISSUE #12157: Performance test opcodes API
### Problem

We'd like to explore enabling the opcodes endpoint by default but need to quantify its performance characteristics.

### Solution

* Add multiple k6 tests that stress different typical scenarios of the opcodes endpoint. This should include one with all options enabled and all disabled. Also one for a large response returned.

### Alternatives

_No response_

bilyana-gospodinova:
**Description**:
This PR adds 2 k6 performance tests for the `/api/v1/contracts/results/{transactionIdOrHash}/opcodes` endpoint - one with all parameters enabled (storage, stack, memory) and one with all parameters disabled.

Fixes #12157

### DIFF
--- a/.github/workflows/acceptance.yaml
+++ b/.github/workflows/acceptance.yaml
@@ -229,7 +229,7 @@ jobs:
           RUN_OPCODE_TESTS: "false"
           RUN_WITH_VARIABLES: "true"
           SERIAL_NUMBER: "0000000000000000000000000000000000000000000000000000000000000001"
-          WEB3_TEST_EXCLUDE: "^contractCall(Allowance|ApprovedForAll|BalanceOf|IsFrozen|IsKyc|Receive).*$|^.*opcode.*js$"
+          WEB3_TEST_EXCLUDE: "^contractCall(Allowance|ApprovedForAll|BalanceOf|IsFrozen|IsKyc|Receive).*$"
         with:
           path: |
             ./tools/k6/src/web3/apis.js


steven-sheehy (dismissed):

jnels124 (dismissed):

tools/k6/src/web3/test/index.js
```diff
@@ -61,6 +61,8 @@ import * as contractCallPrecompileTransferFungibleToken from './modificationTest
 import * as contractCallRedirectApprove from './modificationTests/contractCallRedirectApprove.js';
 import * as contractCallComplexFunctionTokenLifecycle from './complex-functions/contractCallComplexFunctionsTokenLifecycle.js';
 import * as contractCallComplexFunctionNFTLifecycle from './complex-functions/contractCallComplexFunctionsNFTLifecycle.js';
+import * as contractResultsOpcodesAllDisabled from './opcodes/contractResultsOpcodesAllDisabled.js';
```
IvanKavaldzhiev: The names might be a bit misleading - e.g. we enable or disable the opcodes themselves, while we enable or disable specific properties related to them. Probably a better naming will be:

contractResultsOpcodesAllDisabled -> contractResultsOpcodesAllPropertiesDisabled
contractResultsOpcodesAllEnabled -> contractResultsOpcodesAllPropertiesEnabled

tools/k6/src/web3/test/common.js
```diff
@@ -33,6 +33,22 @@ function isNonErrorResponse(response) {
...
+    return false;
+  }
+
+  const body = JSON.parse(response.body);
+  const list = body[listName];
+  if (!Array.isArray(list)) {
+    return false;
+  }
+
+  return list.length > 0;
```
IvanKavaldzhiev: While it will be very difficult to make a thorough validation about the correctness of the returned response, I think this validation will return true even if we have `opcodes: [ ]`. When we make `contractCall` or `contractCreate` execution we have at least 3-4 opcodes that should be executed, even if the `messageFrame` directly reverts. I think we can check whether the list has a length > 2 at least, so that we exclude empty opcodes list as a valid result.

tools/k6/src/web3/test/index.js
bilyana-gospodinova: Renamed

tools/k6/src/web3/test/common.js
bilyana-gospodinova: Agree, I changed the validation to include min number of elements in the result.

steven-sheehy (dismissed):

.github/workflows/acceptance.yaml
```diff
@@ -226,9 +226,10 @@ jobs:
           RUN_COMPLEX_TESTS: "false"
           RUN_ESTIMATE_TESTS: "false"
           RUN_MODIFICATION_TESTS: "false"
+          RUN_OPCODE_TESTS: "false"
           RUN_WITH_VARIABLES: "true"
           SERIAL_NUMBER: "0000000000000000000000000000000000000000000000000000000000000001"
-          WEB3_TEST_EXCLUDE: "^contractCall(Allowance|ApprovedForAll|BalanceOf|IsFrozen|IsKyc|Receive).*$"
+          WEB3_TEST_EXCLUDE: "^contractCall(Allowance|ApprovedForAll|BalanceOf|IsFrozen|IsKyc|Receive).*$|^.*opcode.*js$"
```
xin-hedera: the regex is tested against the test case name, not the js file name, so for `opcode` tests, it should be `^.*opcodes.*$`.

In addition, I don't think you need to change `WEB3_TEST_EXCLUDE` since `RUN_OPCODE_TESTS` is set to `false`.

bilyana-gospodinova: Right, removed.

IvanKavaldzhiev (approved):
LGTM

PR #12349, opened 2025-11-06
Remove JavaScript based code for /network/exchangerate

### LINKED ISSUE #12305: Remove JavaScript based `/network/exchangerate`
### Problem

After #12101 is successfully deployed to mainnet, the legacy JavaScript implementation should be deleted.

### Solution

Follow the [checklist](https://github.com/hiero-ledger/hiero-mirror-node/issues/12101) to remove the code.

### Alternatives

_No response_

IvanKavaldzhiev:
**Description**:
This PR removes all JavaScript related code to /network/exchangerate rest endpoint.

**Related issue(s)**:

Fixes #12305

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/charts/hedera-mirror-rest-java/values.yaml
+++ b/charts/hedera-mirror-rest-java/values.yaml
@@ -311,6 +311,7 @@ resources:
 
 revisionHistoryLimit: 3
 
+routes: {}
 
 securityContext:
   allowPrivilegeEscalation: false


charts/hedera-mirror-rest-java/values.yaml
```diff
@@ -314,8 +311,6 @@ resources:
 
 revisionHistoryLimit: 3
 
-routes:
-  exchangeRate: true
```
steven-sheehy, line 318: Just do `routes: {}` instead of removing it entirely.

steven-sheehy:
Think you need to merge from main to get acceptance fix.

charts/hedera-mirror-rest-java/values.yaml
IvanKavaldzhiev: Fixed

steven-sheehy:
See https://github.com/hiero-ledger/hiero-mirror-node/pull/12349#issuecomment-3499140892

IvanKavaldzhiev:
> * End-to-end Tests / acceptance (RECORD) (pull_request)

I merged main, but the RECORD test workflow won't fix, until the mirror-node version in acceptance.yaml is changed to 0.143.0. The fix from the merged PR related to the web3 module is not yet published in an official tag.

xin-hedera:
we should keep the spec based integration tests and change them so they will not run against the js rest api. can create a follow-up ticket to address that.

steven-sheehy:
> we should keep the spec based integration tests and change them so they will not run against the js rest api. can create a follow-up ticket to address that.

We've been deleting them. The checklist has an item `Controller integration tests added with equivalent or better coverage to existing rest spec tests`. So we should already have Java controller tests that match all the spec tests and we've verified that the spec tests pass in case there was some bug or missing coverage in our tests. Otherwise when would we know when we could delete old tests and pay down the technical debt?

PR #12351, opened 2025-11-07
HIP-1195 Added Sidecars for hook and updated EVMHookHandler to add initial storage

### LINKED ISSUE #12120: HIP-1195 Enhance sidecar processing for hook storage updates
### Problem

Sidecar will now have storage update for hooks #11731 

### Solution

- Modify ContractResultServiceImpl.processContractStateChange()
- Detect hook storage updates (contract ID 365 with parent transaction)
- Implement hook context derivation from parent transaction order
- Route hook storage changes to HookStorageChange entities instead of ContractStateChange
- Process hook execution order (HBAR → Fungible → NFT transfers)
- Add functionality to store initial storage for hook

### Alternatives

_No response_

ashumahajan:
**Description**:

1. Added Sidecar storage update functionality for hooks
2. Updated EVMHookHandler to support initial hook storage at time of creation

Fixes #12120 

**Checklist**

### DIFF
--- a/common/src/test/java/org/hiero/mirror/common/domain/transaction/HookExecutionCollectorTest.java
+++ b/common/src/test/java/org/hiero/mirror/common/domain/transaction/HookExecutionCollectorTest.java
@@ -10,7 +10,7 @@ import com.hederahashgraph.api.proto.java.HookCall;
 import org.hiero.mirror.common.domain.hook.AbstractHook;
 import org.junit.jupiter.api.Test;
 
-class HookExecutionCollectorTest {
+final class HookExecutionCollectorTest {
 
     @Test
     void createEmptyCollector() {
@@ -87,20 +87,5 @@ class HookExecutionCollectorTest {
         assertEquals(new AbstractHook.Id(202L, 1002L), queue.poll()); // allowPre[1]
         assertEquals(new AbstractHook.Id(102L, 1002L), queue.poll()); // allowPost[0] (same as allowPre[0])
         assertEquals(new AbstractHook.Id(202L, 1002L), queue.poll()); // allowPost[1] (same as allowPre[1])
-    }
-
-    @Test
-    void methodChaining() {
-        var queue = HookExecutionCollector.create()
-                .addAllowExecHook(HookCall.newBuilder().setHookId(1L).build(), 100L)
-                .addPrePostExecHook(HookCall.newBuilder().setHookId(2L).build(), 200L)
-                .addAllowExecHook(HookCall.newBuilder().setHookId(3L).build(), 300L)
-                .buildExecutionQueue();
-
-        assertEquals(4, queue.size());
-        assertEquals(new AbstractHook.Id(1L, 100L), queue.poll());
-        assertEquals(new AbstractHook.Id(3L, 300L), queue.poll());
-        assertEquals(new AbstractHook.Id(2L, 200L), queue.poll());
-        assertEquals(new AbstractHook.Id(2L, 200L), queue.poll());
     }
 }
--- a/importer/src/main/java/org/hiero/mirror/importer/domain/ContractResultServiceImpl.java
+++ b/importer/src/main/java/org/hiero/mirror/importer/domain/ContractResultServiceImpl.java
@@ -506,7 +506,7 @@ final class ContractResultServiceImpl implements ContractResultService {
 
     /**
      * Determines if a RecordItem represents a hook execution. Hook executions are ContractCall transactions to the
-     * system hook contract (0.0.365) that have a parent consensus timestamp.
+     * system hook contract (0x16d) that have a parent consensus timestamp.
      *
      * @param recordItem the record item to check
      * @return true if this is a hook execution, false otherwise
--- a/importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/CryptoTransferTransactionHandler.java
+++ b/importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/CryptoTransferTransactionHandler.java
@@ -42,21 +42,27 @@ class CryptoTransferTransactionHandler extends AbstractTransactionHandler {
 
     private void addNftHookCalls(List<NftTransfer> nftTransfersList, HookExecutionCollector hookExecutionCollector) {
         for (NftTransfer nftTransfer : nftTransfersList) {
-            final var receiverId =
-                    entityIdService.lookup(nftTransfer.getReceiverAccountID()).orElse(EntityId.EMPTY);
-            final var senderId =
-                    entityIdService.lookup(nftTransfer.getSenderAccountID()).orElse(EntityId.EMPTY);
             if (nftTransfer.hasPreTxSenderAllowanceHook()) {
+                final var senderId =
+                        entityIdService.lookup(nftTransfer.getSenderAccountID()).orElse(EntityId.EMPTY);
                 hookExecutionCollector.addAllowExecHook(nftTransfer.getPreTxSenderAllowanceHook(), senderId.getId());
             } else if (nftTransfer.hasPrePostTxSenderAllowanceHook()) {
+                final var senderId =
+                        entityIdService.lookup(nftTransfer.getSenderAccountID()).orElse(EntityId.EMPTY);
                 hookExecutionCollector.addPrePostExecHook(
                         nftTransfer.getPrePostTxSenderAllowanceHook(), senderId.getId());
             }
 
             if (nftTransfer.hasPreTxReceiverAllowanceHook()) {
+                final var receiverId = entityIdService
+                        .lookup(nftTransfer.getReceiverAccountID())
+                        .orElse(EntityId.EMPTY);
                 hookExecutionCollector.addAllowExecHook(
                         nftTransfer.getPreTxReceiverAllowanceHook(), receiverId.getId());
             } else if (nftTransfer.hasPrePostTxReceiverAllowanceHook()) {
+                final var receiverId = entityIdService
+                        .lookup(nftTransfer.getReceiverAccountID())
+                        .orElse(EntityId.EMPTY);
                 hookExecutionCollector.addPrePostExecHook(
                         nftTransfer.getPrePostTxReceiverAllowanceHook(), receiverId.getId());
             }
@@ -65,11 +71,13 @@ class CryptoTransferTransactionHandler extends AbstractTransactionHandler {
 
     private void addHookCalls(List<AccountAmount> accountAmountsList, HookExecutionCollector hookExecutionCollector) {
         for (AccountAmount accountAmount : accountAmountsList) {
-            final var accountId =
-                    entityIdService.lookup(accountAmount.getAccountID()).orElse(EntityId.EMPTY);
             if (accountAmount.hasPreTxAllowanceHook()) {
+                final var accountId =
+                        entityIdService.lookup(accountAmount.getAccountID()).orElse(EntityId.EMPTY);
                 hookExecutionCollector.addAllowExecHook(accountAmount.getPreTxAllowanceHook(), accountId.getId());
             } else if (accountAmount.hasPrePostTxAllowanceHook()) {
+                final var accountId =
+                        entityIdService.lookup(accountAmount.getAccountID()).orElse(EntityId.EMPTY);
                 hookExecutionCollector.addPrePostExecHook(accountAmount.getPrePostTxAllowanceHook(), accountId.getId());
             }
         }
--- a/importer/src/test/java/org/hiero/mirror/importer/parser/record/transactionhandler/CryptoTransferTransactionHandlerTest.java
+++ b/importer/src/test/java/org/hiero/mirror/importer/parser/record/transactionhandler/CryptoTransferTransactionHandlerTest.java
@@ -72,7 +72,6 @@ class CryptoTransferTransactionHandlerTest extends AbstractTransactionHandlerTes
 
         // then
         var queue = recordItem.getHookExecutionQueue();
-        assertNotNull(queue, "Hook execution queue should exist even when empty");
         assertThat(queue).isEmpty();
     }
 
@@ -208,6 +207,110 @@ class CryptoTransferTransactionHandlerTest extends AbstractTransactionHandlerTes
         assertEquals(new AbstractHook.Id(203L, 2003L), recordItem.nextHookContext()); // Token PrePostTx Post
         assertEquals(new AbstractHook.Id(302L, 3002L), recordItem.nextHookContext()); // NFT2 sender PrePostTx Post
         assertEquals(new AbstractHook.Id(312L, 3102L), recordItem.nextHookContext()); // NFT2 receiver PrePostTx Post
+
+        // Verify queue is exhausted
+        assertNull(recordItem.nextHookContext());
+    }
+
+    @Test
+    @DisplayName("Hook execution queue handles unresolvable alias/EVM address")
+    void testHookExecutionQueueWithUnresolvableAlias() {
+        // Mock EntityIdService to fail to resolve alias
+        when(entityIdService.lookup(any(AccountID.class))).thenReturn(Optional.empty());
+
+        var recordItem = recordItemBuilder
+                .cryptoTransfer()
+                .transactionBody(body -> body.setTransfers(TransferList.newBuilder()
+                        .addAccountAmounts(AccountAmount.newBuilder()
+                                .setAccountID(AccountID.newBuilder()
+                                        .setAlias(com.google.protobuf.ByteString.copyFromUtf8("unresolvable_alias")))
+                                .setAmount(100L)
+                                .setPreTxAllowanceHook(HookCall.newBuilder().setHookId(502L)))))
+                .build();
+
+        var transaction = domainBuilder.transaction().get();
+
+        // when
+        transactionHandler.updateTransaction(transaction, recordItem);
+
+        // then
+        var queue = recordItem.getHookExecutionQueue();
+        assertNotNull(queue, "Hook execution queue should be created");
+        assertEquals(1, queue.size(), "Queue should contain 1 hook with EntityId.EMPTY");
+        assertEquals(new AbstractHook.Id(502L, 0L), recordItem.nextHookContext()); // EntityId.EMPTY has ID 0
+        assertNull(recordItem.nextHookContext());
+    }
+
+    @Test
+    @DisplayName("Hook execution queue handles mixed resolvable and unresolvable aliases in NFT transfers")
+    void testHookExecutionQueueWithMixedNftAliases() {
+        // Mock EntityIdService to resolve some aliases and fail others
+        when(entityIdService.lookup(any(AccountID.class))).thenAnswer(invocation -> {
+            AccountID accountId = invocation.getArgument(0);
+            if (accountId.hasAlias()) {
+                var alias = accountId.getAlias().toStringUtf8();
+                if ("resolvable_sender".equals(alias)) {
+                    return Optional.of(EntityId.of(6001L));
+                } else if ("resolvable_receiver".equals(alias)) {
+                    return Optional.of(EntityId.of(6002L));
+                }
+                // Unresolvable aliases return empty
+                return Optional.empty();
+            }
+            return Optional.of(EntityId.of(accountId));
+        });
+
+        var recordItem = recordItemBuilder
+                .cryptoTransfer()
+                .transactionBody(body -> body.addTokenTransfers(TokenTransferList.newBuilder()
+                        .setToken(TokenID.newBuilder().setTokenNum(7000L))
+                        .addNftTransfers(NftTransfer.newBuilder()
+                                .setSenderAccountID(AccountID.newBuilder()
+                                        .setAlias(com.google.protobuf.ByteString.copyFromUtf8("resolvable_sender")))
+                                .setReceiverAccountID(AccountID.newBuilder()
+                                        .setAlias(com.google.protobuf.ByteString.copyFromUtf8("unresolvable_receiver")))
+                                .setSerialNumber(1L)
+                                .setPreTxSenderAllowanceHook(
+                                        HookCall.newBuilder().setHookId(701L))
+                                .setPreTxReceiverAllowanceHook(
+                                        HookCall.newBuilder().setHookId(702L)))
+                        .addNftTransfers(NftTransfer.newBuilder()
+                                .setSenderAccountID(AccountID.newBuilder()
+                                        .setAlias(com.google.protobuf.ByteString.copyFromUtf8("unresolvable_sender")))
+                                .setReceiverAccountID(AccountID.newBuilder()
+                                        .setAlias(com.google.protobuf.ByteString.copyFromUtf8("resolvable_receiver")))
+                                .setSerialNumber(2L)
+                                .setPrePostTxSenderAllowanceHook(
+                                        HookCall.newBuilder().setHookId(703L))
+                                .setPrePostTxReceiverAllowanceHook(
+                                        HookCall.newBuilder().setHookId(704L)))))
+                .build();
+
+        var transaction = domainBuilder.transaction().get();
+
+        // when
+        transactionHandler.updateTransaction(transaction, recordItem);
+
+        // then
+        var queue = recordItem.getHookExecutionQueue();
+        assertNotNull(queue, "Hook execution queue should be created");
+        assertEquals(
+                6, queue.size(), "Queue should contain 6 hooks total"); // 2 PreTx + 2 PrePostTx Pre + 2 PrePostTx Post
+
+        // Phase 1: allowExec (PreTx hooks)
+        assertEquals(new AbstractHook.Id(701L, 6001L), recordItem.nextHookContext()); // Resolvable sender
+        assertEquals(
+                new AbstractHook.Id(702L, 0L), recordItem.nextHookContext()); // Unresolvable receiver -> EntityId.EMPTY
+
+        // Phase 2: allowPre (Pre hooks from PrePostTx)
+        assertEquals(
+                new AbstractHook.Id(703L, 0L), recordItem.nextHookContext()); // Unresolvable sender -> EntityId.EMPTY
+        assertEquals(new AbstractHook.Id(704L, 6002L), recordItem.nextHookContext()); // Resolvable receiver
+
+        // Phase 3: allowPost (Post hooks from PrePostTx - same order as Pre)
+        assertEquals(
+                new AbstractHook.Id(703L, 0L), recordItem.nextHookContext()); // Unresolvable sender -> EntityId.EMPTY
+        assertEquals(new AbstractHook.Id(704L, 6002L), recordItem.nextHookContext()); // Resolvable receiver
 
         // Verify queue is exhausted
         assertNull(recordItem.nextHookContext());
--- a/importer/src/test/java/org/hiero/mirror/importer/parser/record/transactionhandler/EVMHookHandlerTest.java
+++ b/importer/src/test/java/org/hiero/mirror/importer/parser/record/transactionhandler/EVMHookHandlerTest.java
@@ -14,6 +14,7 @@ import static org.mockito.Mockito.verifyNoInteractions;
 import static org.mockito.Mockito.when;
 
 import com.google.common.collect.Range;
+import com.google.common.primitives.Bytes;
 import com.google.protobuf.ByteString;
 import com.hedera.hapi.node.hooks.legacy.EvmHookSpec;
 import com.hedera.hapi.node.hooks.legacy.HookCreationDetails;
@@ -78,13 +79,8 @@ final class EVMHookHandlerTest {
                 .build();
 
         final var key = domainBuilder.bytes(32);
-        final var leadingZeros = new byte[16];
-        final var trimmedValue = ByteBuffer.allocate(16)
-                .put((byte) 0x01)
-                .put(domainBuilder.bytes(15))
-                .array();
-        final var value =
-                ByteBuffer.allocate(32).put(leadingZeros).put(trimmedValue).array();
+        final var trimmedValue = Bytes.concat(new byte[] {0x1}, domainBuilder.bytes(15));
+        final var value = leftPadBytes(trimmedValue, 32);
 
         var lambdaEvmHook = LambdaEvmHook.newBuilder()
                 .addStorageUpdates(LambdaStorageUpdate.newBuilder()
@@ -433,10 +429,8 @@ final class EVMHookHandlerTest {
 
         for (int i = 0; i < numEntries; i++) {
             final var preimage = domainBuilder.bytes(8);
-            final var leadingZeros = new byte[15];
-            final var trimmedValue = domainBuilder.bytes(17);
-            final var value =
-                    ByteBuffer.allocate(32).put(leadingZeros).put(trimmedValue).array();
+            final var trimmedValue = Bytes.concat(new byte[] {0x1}, domainBuilder.bytes(16));
+            final var value = leftPadBytes(trimmedValue, 32);
 
             preimages.add(preimage);
             expectedValues.add(trimmedValue);


common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
```diff
@@ -106,6 +107,29 @@ public class RecordItem implements StreamItem {
     @Setter
     private List<TransactionSidecarRecord> sidecarRecords = Collections.emptyList();
 
+    // Transient hook execution queue for CryptoTransfer transactions that may trigger hooks
+    @NonFinal
+    @Setter
+    Queue<HookId> hookExecutionQueue;
```
jnels124: probably better to keep private and add relevant functions to queue dequeue items

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
```diff
@@ -221,6 +245,11 @@ public Collection<ContractTransaction> populateContractTransactions() {
         return contractTransactions.values();
     }
 
+    /**
+     * Record representing a hook execution context within a parent transaction.
+     */
+    public record HookId(long hookId, long ownerId) {}
```
steven-sheehy: Can't we just reuse `AbstractHook.Id`?

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
steven-sheehy:
```suggestion
    private Queue<HookId> hookExecutionQueue;
```

importer/src/main/java/org/hiero/mirror/importer/domain/ContractResultServiceImpl.java
```diff
@@ -494,6 +502,93 @@ private boolean shouldPersistCreatedContractIDs(RecordItem recordItem) {
...
+        var hookContext = parentRecordItem.nextHookContext();
+        if (hookContext == null) {
+            Utility.handleRecoverableError(
+                    "No hook context available in parent transaction for hook execution at consensus timestamp {}",
+                    recordItem.getConsensusTimestamp());
+            return;
+        }
+
+        // Process each storage change with the same hook context
+        for (var storageChange : stateChange.getStorageChangesList()) {
```
jnels124: possible to add StateChangeList to the process storage updates method on evm hook handler? That way we are processing all storage updates in one place

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/EntityRecordItemListener.java
```diff
@@ -596,6 +663,35 @@ private void insertNonFungibleTokenTransfers(
             recordItem.addEntityId(entityTokenId);
 
             transferNftOwnership(consensusTimestamp, serialNumber, entityTokenId, receiverId);
+
```
jnels124: nit: would make this method simpler to move the hook processing to a different method and call from here

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/EntityRecordItemListener.java
```diff
@@ -115,8 +117,18 @@ public void onItem(RecordItem recordItem) throws ImporterException {
         Transaction transaction = buildTransaction(entityId, recordItem);
         transactionHandler.updateTransaction(transaction, recordItem);
 
+        // hook list for pre-allowance, pre and post hooks. We have to maintain the seq
+        // 1. HBAR transfers
+        // 2. Fungible Token Transfers
+        // 3. NFT Transfers
+        // Since in existing logic the seq is maintained, we will use the function calls as is
+        final var hookExecutionQueue = new LinkedList<HookId>();
+        final var allowExecHookIds = new ArrayList<HookId>();
```
jnels124: Maybe wrap these in a java Record so we don't have to pass the three lists everywhere and can centralize a place where things are going into the lists

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
jnels124: Also, i think `ArrayDeque` is a better interface here since we are not parallelizing the work here

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
xin-hedera: we already have a `AbstractHook.Id` class, should reuse it here.

importer/src/main/java/org/hiero/mirror/importer/domain/ContractResultServiceImpl.java
```diff
@@ -494,6 +502,93 @@ private boolean shouldPersistCreatedContractIDs(RecordItem recordItem) {
...
+    private boolean isHookExecution(RecordItem recordItem) {
+        var transactionRecord = recordItem.getTransactionRecord();
+        var transactionBody = recordItem.getTransactionBody();
+
+        // Must be a ContractCall transaction to 0.0.365 with a parent consensus timestamp
+        return transactionBody.hasContractCall()
+                && transactionRecord.hasParentConsensusTimestamp()
+                && entityIdService
+                        .lookup(transactionBody.getContractCall().getContractID())
+                        .map(entityId -> entityId.getId() == 365)
```
xin-hedera:
```suggestion
                        .map(entityId -> entityId.getNum() == 365)
```

in non-zero shard / realm network, the encoded id will not be 365, so we should only check the entity num. Also should make 365 a constant.

importer/src/main/java/org/hiero/mirror/importer/domain/ContractResultServiceImpl.java
```diff
@@ -494,6 +502,93 @@ private boolean shouldPersistCreatedContractIDs(RecordItem recordItem) {
...
+     * @param stateChange the contract state change containing hook storage updates
+     */
+    private void processHookStorageChanges(RecordItem recordItem, ContractStateChange stateChange) {
+        var transactionRecord = recordItem.getTransactionRecord();
+        if (!transactionRecord.hasParentConsensusTimestamp()) {
+            Utility.handleRecoverableError(
+                    "Hook storage change detected but no parent consensus timestamp found at {}",
+                    recordItem.getConsensusTimestamp());
+            return;
+        }
```
xin-hedera: this is dead code since it's checked in `isHookExecution(RecordItem recordItem)`

It's also debatable whether we want to relax it and skip parent consensus timestamp check. 

```suggestion
```

importer/src/main/java/org/hiero/mirror/importer/domain/ContractResultServiceImpl.java
```diff
@@ -494,6 +502,93 @@ private boolean shouldPersistCreatedContractIDs(RecordItem recordItem) {
...
+
+        var parentRecordItem = recordItem.getParent();
+        if (parentRecordItem == null) {
+            Utility.handleRecoverableError(
+                    "Unable to locate parent RecordItem for hook execution at consensus timestamp {}",
+                    recordItem.getConsensusTimestamp());
+            return;
+        }
```
xin-hedera:
```suggestion
```

we should move the logic into `RecordItem.nextHookContext()`, make it call `parent.nextHookContext()` if parent is not null

importer/src/main/java/org/hiero/mirror/importer/domain/ContractResultServiceImpl.java
```diff
@@ -494,6 +502,93 @@ private boolean shouldPersistCreatedContractIDs(RecordItem recordItem) {
...
+                    .consensusTimestamp(recordItem.getConsensusTimestamp())
+                    .hookId(hookContext.hookId())
+                    .ownerId(hookContext.ownerId())
+                    .key(DomainUtils.toBytes(storageChange.getSlot()))
+                    .valueRead(DomainUtils.toBytes(storageChange.getValueRead()))
+                    .valueWritten(
+                            storageChange.hasValueWritten()
+                                    ? DomainUtils.toBytes(
+                                            storageChange.getValueWritten().getValue())
+                                    : null)
```
xin-hedera: nit, move the ternary expression out and assign it to a local variable for better readability.

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/EntityRecordItemListener.java
```diff
@@ -115,8 +117,18 @@ public void onItem(RecordItem recordItem) throws ImporterException {
         Transaction transaction = buildTransaction(entityId, recordItem);
         transactionHandler.updateTransaction(transaction, recordItem);
 
+        // hook list for pre-allowance, pre and post hooks. We have to maintain the seq
+        // 1. HBAR transfers
+        // 2. Fungible Token Transfers
+        // 3. NFT Transfers
+        // Since in existing logic the seq is maintained, we will use the function calls as is
+        final var hookExecutionQueue = new LinkedList<HookId>();
```
xin-hedera: we should move all hook execution logic to `CryptoTransferTransactionHandler.doUpdateTransaction`

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/EntityRecordItemListener.java
```diff
@@ -115,8 +117,18 @@ public void onItem(RecordItem recordItem) throws ImporterException {
...
+        final var allowExecHookIds = new ArrayList<HookId>();
+        final var allowPreExecHookIds = new ArrayList<HookId>();
+        final var allowPostExecHookIds = new ArrayList<HookId>();
+
         // Insert transfers even on failure
-        insertTransferList(recordItem);
+        insertTransferList(recordItem, allowExecHookIds, allowPreExecHookIds, allowPostExecHookIds);
```
xin-hedera: All these have a common problem: we should read and store hook_call info from crypto transfer transaction body, not from any transaction's transaction record.

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/EVMHookHandler.java
```diff
@@ -57,6 +68,11 @@ void process(
     @Override
     public void processStorageUpdates(
```
xin-hedera, line 70: why can't we unify all callers of `process` and `processStorageUpdates` to either pass in `EntityId ownerEntityId` or `long ownerEntityId`?

importer/src/test/java/org/hiero/mirror/importer/domain/ContractResultServiceImplTest.java
```diff
@@ -260,4 +264,123 @@ void gasConsumedWithOlderHapiVersionThanSmartContractThrottling() {
         var capturedContractResult = contractResultCaptor.getValue();
         assertThat(capturedContractResult.getGasConsumed()).isLessThan(capturedContractResult.getGasUsed());
     }
+
+    @ParameterizedTest
+    @CsvSource({"1", "2", "3", "4"})
```
xin-hedera:
```suggestion
    @ValueSource(ints = {1, 2, 3, 4})
```

importer/src/test/java/org/hiero/mirror/importer/parser/domain/RecordItemBuilder.java
```diff
@@ -1383,7 +1382,7 @@ private ContractLoginfo contractLoginfo(ContractID contractId) {
                 .build();
     }
```
xin-hedera, line 1385: nit, move it up to the public method section

xin-hedera (changes requested):

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/EntityRecordItemListener.java
```diff
@@ -521,8 +576,15 @@ private void insertTokenTransfers(RecordItem recordItem, Transaction transaction
...
-            insertNonFungibleTokenTransfers(recordItem, transaction, tokenTransferList);
+            insertFungibleTokenTransfers(
+                    recordItem, tokenTransferList, allowExecHookIds, allowPreExecHookIds, allowPostExecHookIds);
+            insertNonFungibleTokenTransfers(
+                    recordItem,
+                    transaction,
+                    tokenTransferList,
+                    allowExecHookIds,
+                    allowPreExecHookIds,
+                    allowPostExecHookIds);
```
xin-hedera: this doesn't follow the order outlined in the HIP: fungible token hooks are processed before non-fungible nftTransfers.

For instance, for `pre_tx*allowance` hooks, the logic here definitely can have a fungible token A's `pre_tx_allowance_hook` followed by a nft's `pre_tx_sender_allowance_hook`, then followed by  a fungible token B's `pre_tx_allowance_hook`.

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
```diff
@@ -106,6 +107,29 @@ public class RecordItem implements StreamItem {
...
+    }
+
+    /**
+     * Checks if there are more hook contexts in the execution queue.
+     *
+     * @return true if there are more hook contexts, false otherwise
+     */
+    public boolean hasMoreHookContexts() {
+        return hookExecutionQueue != null && !hookExecutionQueue.isEmpty();
+    }
```
xin-hedera: only used in test? can just assert `nextHookContext() == null` instead.

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/EntityRecordItemListener.java
```diff
@@ -115,8 +117,18 @@ public void onItem(RecordItem recordItem) throws ImporterException {
         Transaction transaction = buildTransaction(entityId, recordItem);
         transactionHandler.updateTransaction(transaction, recordItem);
 
+        // hook list for pre-allowance, pre and post hooks. We have to maintain the seq
+        // 1. HBAR transfers
+        // 2. Fungible Token Transfers
+        // 3. NFT Transfers
+        // Since in existing logic the seq is maintained, we will use the function calls as is
```
xin-hedera: for context, this is what's in the latest HIP

```
1. All pre_tx_allowance_hook calls in the HBAR TransferList, in the order they appear in the CryptoTransferTransactionBody.
2. All pre_tx_allowance_hook calls in the fungible transfers of the tokenTransfers list, in the order they appear in the CryptoTransferTransactionBody.
3. All pre_tx_sender_allowance_hook, pre_tx_receiver_allowance_hook, calls in the non-fungible nftTransfers of the tokenTransfers list, in the order they appear in the CryptoTransferTransactionBody. (Note: if both sender and receiver hooks are present, then the sender hook is executed first.)
4. All pre_post_tx_allowance_hook calls in the HBAR TransferList, in the order they appear in the CryptoTransferTransactionBody.
5. All pre_post_tx_allowance_hook calls in the fungible transfers of the tokenTransfers list, in the order they appear in the CryptoTransferTransactionBody.
6. All pre_post_sender_tx_allowance_hook, pre_post_tx_receiver_allowance_hook, calls in the non-fungible nftTransfers of the tokenTransfers list, in the order they appear in the CryptoTransferTransactionBody.
7. All pre_post hooks previously executed in steps (4) to (6), in the same order they were previously executed; while the first call will be to the allowPre(HookContext, ProposedTransfers) method signature, the second call will be to the allowPost(HookContext, ProposedTransfers) signature.
```

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/EntityRecordItemListener.java
ashumahajan: The logic is based on the logic provided by Michael and Neeha. 
https://github.com/hiero-ledger/hiero-consensus-node/blob/main/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/translators/inputs/BlockTransactionalUnit.java#L140

xin-hedera: should ask them to clarify since there is obvious discrepency.

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/EntityRecordItemListener.java
ashumahajan: The logic is following the sequence suggested here.
we create 3 lists.
1. allowExecution - pre_tx_allowance_hook
2. allowPreExecHookIds - pre_post_tx_allowance_hook for pre
3. allowPostExecHookIds - pre_post_tx_allowance_hook for post

Step 1: allowExecution is populated for HBAR
Step 2: allowPreExecHookIds is populated for HBAR
Step 3: allowPostExecHookIds is populated for HBAR
Step 4: allowExecution is populated for Fungible Token
Step 5: allowPreExecHookIds is populated for Fungible Token
Step 6: allowPostExecHookIds is populated for Fungible Token
Step 7: allowExecution is populated for sender NFT
Step 8: allowPreExecHookIds is populated for sender NFT
Step 9: allowPostExecHookIds is populated for sender NFT
Step 10: allowExecution is populated for receiver NFT
Step 11: allowPreExecHookIds is populated for receiver NFT
Step 12: allowPostExecHookIds is populated for receiver NFT

Based on logic in consensus node implementation, either Step 7/Step 10 is executed or Step 8, 9, 11 and 12 is executed.
Then all the 3 lists are added to the queue(to maintain the sequence).

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/EntityRecordItemListener.java
ashumahajan: The reason I did it this way is to ensure that I don't loop through again and use existing looping logic.

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/EntityRecordItemListener.java
ashumahajan: Transfers are managed in EntityRecordItemListener. If I move all the logic to CryptoTransferTransactionHandler, I will have to re-loop through all the transfers again

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/EntityRecordItemListener.java
ashumahajan: All the AccountAmount and NFTTransfers are currently read from recordItem and since we are already doing it I just added hook_call logic here, instead of looping through the transfers again. Do you want me to change it?

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/EntityRecordItemListener.java
xin-hedera: You probably didn't read all of my comments. doing all these here in the TransacitonRecord is just wrong, since the hooks are all in CryptoTransferTransactionBody.

So this is not about making it efficient, it's done in the wrong way.

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/EntityRecordItemListener.java
xin-hedera: see my other comment. I think there's still misunderstanding.

we should only read hooks from CryptoTransferTransactionBody as stated in the HIP, not any of the transfers in `TransactionRecord`.

ashumahajan: It doesn't make sense, the hook_calls are executed from consensus_node and is part of signed TransactionBody but are later on added. It should be part of TransactionRecord. I have already asked Michael on the channel for the clarification.

xin-hedera: I'm sure Michael will explain in detail why it should and will only exist in the body thus that's the only source MN should read the info from.

In addition, it's also explicitly called out in our design doc: https://github.com/hiero-ledger/hiero-mirror-node/blob/main/docs/design/hiero-hooks.md, section `Hook Context Derivation Logic`

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/EVMHookHandler.java
ashumahajan: Done

importer/src/test/java/org/hiero/mirror/importer/parser/domain/RecordItemBuilder.java
ashumahajan: Done

common/src/main/java/org/hiero/mirror/common/domain/transaction/HookExecutionCollector.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    /**
+     * Returns the total number of hooks across all execution phases.
+     *
+     * @return total number of hooks
+     */
+    public int totalHookCount() {
+        return allowExecHookIds.size() + allowPreExecHookIds.size() + allowPostExecHookIds.size();
+    }
```
xin-hedera: only for tests?

In general we don't want methods whose sole purpose is open internal state of a class for test.

Besides, these are not really required for what the tests what to check.

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/CryptoTransferTransactionHandler.java
```diff
@@ -2,12 +2,89 @@
...
 class CryptoTransferTransactionHandler extends AbstractTransactionHandler {
 
+    @Override
+    protected void doUpdateTransaction(Transaction transaction, RecordItem recordItem) {
+        final var transactionBody = recordItem.getTransactionBody().getCryptoTransfer();
+        HookExecutionCollector hookExecutionCollector = HookExecutionCollector.create();
+        addHookCalls(transactionBody.getTransfers().getAccountAmountsList(), hookExecutionCollector);
+        addTokenHookCalls(transactionBody.getTokenTransfersList(), hookExecutionCollector);
+        recordItem.setHookExecutionQueue(hookExecutionCollector.buildExecutionQueue());
+        super.doUpdateTransaction(transaction, recordItem);
```
xin-hedera: the method in the parent class doesn't do anything

```suggestion
```

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/CryptoTransferTransactionHandler.java
```diff
@@ -2,12 +2,89 @@
...
+import org.hiero.mirror.common.domain.transaction.Transaction;
 import org.hiero.mirror.common.domain.transaction.TransactionType;
 
 @Named
 class CryptoTransferTransactionHandler extends AbstractTransactionHandler {
 
+    @Override
+    protected void doUpdateTransaction(Transaction transaction, RecordItem recordItem) {
+        final var transactionBody = recordItem.getTransactionBody().getCryptoTransfer();
+        HookExecutionCollector hookExecutionCollector = HookExecutionCollector.create();
```
xin-hedera:
```suggestion
        final var hookExecutionCollector = HookExecutionCollector.create();
```

and a lot other places can also use `final var`

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/CryptoTransferTransactionHandler.java
```diff
@@ -2,12 +2,89 @@
...
+        recordItem.setHookExecutionQueue(hookExecutionCollector.buildExecutionQueue());
+        super.doUpdateTransaction(transaction, recordItem);
+    }
+
+    private void addTokenHookCalls(
+            List<TokenTransferList> tokenTransfersList, HookExecutionCollector hookExecutionCollector) {
+        for (TokenTransferList transferList : tokenTransfersList) {
+            for (AccountAmount accountAmount : transferList.getTransfersList()) {
+                addHookCalls(hookExecutionCollector, accountAmount);
+            }
```
xin-hedera:
```suggestion
            addHookCalls(transferList.getTransfersList(), hookExecutionCollector);
```

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/CryptoTransferTransactionHandler.java
```diff
@@ -2,12 +2,89 @@
...
+                if (hookCall.hasHookId()) {
+                    hookExecutionCollector.addPrePostExecHook(hookCall.getHookId(), receiverId.getId());
+                }
+            }
+        }
+    }
+
+    private void addHookCalls(List<AccountAmount> accountAmountsList, HookExecutionCollector hookExecutionCollector) {
+        for (AccountAmount accountAmount : accountAmountsList) {
+            addHookCalls(hookExecutionCollector, accountAmount);
```
xin-hedera: nit, I don't think there's enough complexity to have a separate `void addHookCalls(HookExecutionCollector hookExecutionCollector, AccountAmount accountAmount)` method, instead, we should just do everything inside the for loop.

It's pretty much the opposite that this makes the code slightly more difficult to read.

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/CryptoTransferTransactionHandler.java
```diff
@@ -2,12 +2,89 @@
...
+    private void addHookCalls(HookExecutionCollector hookExecutionCollector, AccountAmount accountAmount) {
+        final var accountId = EntityId.of(accountAmount.getAccountID());
+        if (accountAmount.hasPreTxAllowanceHook()) {
+            final var hookCall = accountAmount.getPreTxAllowanceHook();
+            if (hookCall.hasHookId()) {
+                hookExecutionCollector.addAllowExecHook(hookCall.getHookId(), accountId.getId());
+            }
+        } else if (accountAmount.hasPrePostTxAllowanceHook()) {
+            final var hookCall = accountAmount.getPrePostTxAllowanceHook();
+            hookExecutionCollector.addPrePostExecHook(hookCall.getHookId(), accountId.getId());
```
xin-hedera: hasHookId?

common/src/main/java/org/hiero/mirror/common/domain/transaction/HookExecutionCollector.java
```diff
@@ -0,0 +1,89 @@
...
+
+    /**
+     * Adds a hook ID to the allowExecHookIds list (PreTx hooks).
+     *
+     * @param hookId the hook ID to add
+     * @param ownerId the owner ID for the hook
+     * @return this collector for method chaining
+     */
+    public HookExecutionCollector addAllowExecHook(long hookId, long ownerId) {
```
xin-hedera: can do
```suggestion
    public HookExecutionCollector addAllowExecHook(final HookCall hookCall, long ownerId) {
        if (!hookCall.hasHookId()) {
            return;
        }
        
        // add. hook context
```

and the same to `addPrePostExecHook` to remove all `hookCall.hasHookId()` calls in `CryptoTransferTransactionHandler`.

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/CryptoTransferTransactionHandler.java
```diff
@@ -2,12 +2,89 @@
...
+        for (TokenTransferList transferList : tokenTransfersList) {
+            for (AccountAmount accountAmount : transferList.getTransfersList()) {
+                addHookCalls(hookExecutionCollector, accountAmount);
+            }
+            addNftHookCalls(transferList.getNftTransfersList(), hookExecutionCollector);
+        }
+    }
+
+    private void addHookCalls(HookExecutionCollector hookExecutionCollector, AccountAmount accountAmount) {
+        final var accountId = EntityId.of(accountAmount.getAccountID());
```
xin-hedera: have to use `entityIdService` to resolve all `AccountID` in transaction body,

also we should consider adding an `EMPTY` default `AbstractHook.Id()` and use it as a the placeholder to put in to hook execution queue when `entityIdService` fails to resolve alias / evm address.

In `ContractResultServiceImpl`, we should completely skip persisting contract state changes for a hook execution when `hookExecutionContext == EMPTY`

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/EVMHookHandler.java
```diff
@@ -72,6 +87,26 @@ public void processStorageUpdates(
         }
     }
 
+    @Override
+    public void processStorageUpdatesForSidecar(
+            long consensusTimestamp, long hookId, long ownerId, List<StorageChange> storageUpdates) {
+        for (var storageChange : storageUpdates) {
+            byte[] valueWritten = storageChange.hasValueWritten()
+                    ? toBytes(storageChange.getValueWritten().getValue())
```
xin-hedera, line 94: there are both static import of `toBytes` and qualified reference `DomainUtils.toBytes`

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/CryptoTransferTransactionHandler.java
ashumahajan: Instead of using default AbstractHook.Id(), I am adding EntityId.EMPTY.Id(), which is skipped in ContractResultServiceImpl by checking hookContext == null || hookContext.getOwnerId() == EntityId.EMPTY.getId()

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/EVMHookHandler.java
ashumahajan: Done

common/src/test/java/org/hiero/mirror/common/domain/transaction/HookExecutionCollectorTest.java
```diff
@@ -0,0 +1,106 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import com.hederahashgraph.api.proto.java.HookCall;
+import org.hiero.mirror.common.domain.hook.AbstractHook;
+import org.junit.jupiter.api.Test;
+
+class HookExecutionCollectorTest {
```
xin-hedera: nit

```suggestion
final class HookExecutionCollectorTest {
```

importer/src/main/java/org/hiero/mirror/importer/domain/ContractResultServiceImpl.java
```diff
@@ -494,6 +504,66 @@ private boolean shouldPersistCreatedContractIDs(RecordItem recordItem) {
                 && recordItem.getHapiVersion().isLessThan(RecordFile.HAPI_VERSION_0_23_0);
     }
 
+    /**
+     * Determines if a RecordItem represents a hook execution. Hook executions are ContractCall transactions to the
+     * system hook contract (0.0.365) that have a parent consensus timestamp.
```
xin-hedera: nit, we should not assume it has shard=0 and realm=0

```suggestion
     * system hook contract (0x16d) that have a parent consensus timestamp.
```

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/CryptoTransferTransactionHandler.java
```diff
@@ -2,12 +2,79 @@
...
+            addNftHookCalls(transferList.getNftTransfersList(), hookExecutionCollector);
+        }
+    }
+
+    private void addNftHookCalls(List<NftTransfer> nftTransfersList, HookExecutionCollector hookExecutionCollector) {
+        for (NftTransfer nftTransfer : nftTransfersList) {
+            final var receiverId =
+                    entityIdService.lookup(nftTransfer.getReceiverAccountID()).orElse(EntityId.EMPTY);
+            final var senderId =
+                    entityIdService.lookup(nftTransfer.getSenderAccountID()).orElse(EntityId.EMPTY);
```
xin-hedera: only look up sender / receiver if needed, since for alias / evm address `entityIdService` may end up making a db query, which is expensive, and may not be needed at all.

importer/src/main/java/org/hiero/mirror/importer/parser/record/transactionhandler/CryptoTransferTransactionHandler.java
```diff
@@ -2,12 +2,79 @@
...
+            } else if (nftTransfer.hasPrePostTxReceiverAllowanceHook()) {
+                hookExecutionCollector.addPrePostExecHook(
+                        nftTransfer.getPrePostTxReceiverAllowanceHook(), receiverId.getId());
+            }
+        }
+    }
+
+    private void addHookCalls(List<AccountAmount> accountAmountsList, HookExecutionCollector hookExecutionCollector) {
+        for (AccountAmount accountAmount : accountAmountsList) {
+            final var accountId =
```
xin-hedera: same here, only try to resolve the account when there is either pre_tx_allowance_hook or pre_post_tx_allowance_hook.

importer/src/test/java/org/hiero/mirror/importer/parser/record/transactionhandler/CryptoTransferTransactionHandlerTest.java
```diff
@@ -44,4 +60,156 @@ void updateTransactionSuccessful() {
...
+        var recordItem = recordItemBuilder.cryptoTransfer().build();
+        var transaction = domainBuilder.transaction().get();
+
+        // when
+        transactionHandler.updateTransaction(transaction, recordItem);
+
+        // then
+        var queue = recordItem.getHookExecutionQueue();
+        assertNotNull(queue, "Hook execution queue should exist even when empty");
+        assertThat(queue).isEmpty();
```
xin-hedera:
```suggestion
        assertThat(queue).isEmpty();
```

the single assert will fail if queue is `null`.

importer/src/test/java/org/hiero/mirror/importer/parser/record/transactionhandler/CryptoTransferTransactionHandlerTest.java
```diff
@@ -44,4 +60,156 @@ void updateTransactionSuccessful() {
...
+    }
+
+    @Test
+    @DisplayName("Hook execution queue covers all use cases from HIP-1195 specification")
+    void testComprehensiveHookExecutionQueueSequence() {
+        // Mock EntityIdService to return proper EntityId objects for each account ID
+        when(entityIdService.lookup(any(AccountID.class))).thenAnswer(invocation -> {
```
xin-hedera, line 82: should also test
- one resolvable alias / evm address
- one alias / evm address `entityIdService` fails to resolve

importer/src/test/java/org/hiero/mirror/importer/parser/record/transactionhandler/EVMHookHandlerTest.java
```diff
@@ -75,7 +77,23 @@ void processHookCreationDetailsSuccess() {
...
 
-        var lambdaEvmHook = LambdaEvmHook.newBuilder().setSpec(evmHookSpec).build();
+        final var key = domainBuilder.bytes(32);
+        final var leadingZeros = new byte[16];
+        final var trimmedValue = ByteBuffer.allocate(16)
+                .put((byte) 0x01)
+                .put(domainBuilder.bytes(15))
+                .array();
+        final var value =
+                ByteBuffer.allocate(32).put(leadingZeros).put(trimmedValue).array();
```
xin-hedera: nit 
```suggestion
        final var value = DomainUtils.leftPadBytes(Bytes.concat(new byte[]{0x1}, domainBuilder.bytes(15)), 32);
```

common/src/test/java/org/hiero/mirror/common/domain/transaction/HookExecutionCollectorTest.java
```diff
@@ -0,0 +1,106 @@
...
+                .addPrePostExecHook(HookCall.newBuilder().setHookId(2L).build(), 200L)
+                .addAllowExecHook(HookCall.newBuilder().setHookId(3L).build(), 300L)
+                .buildExecutionQueue();
+
+        assertEquals(4, queue.size());
+        assertEquals(new AbstractHook.Id(1L, 100L), queue.poll());
+        assertEquals(new AbstractHook.Id(3L, 300L), queue.poll());
+        assertEquals(new AbstractHook.Id(2L, 200L), queue.poll());
+        assertEquals(new AbstractHook.Id(2L, 200L), queue.poll());
+    }
```
xin-hedera: the previous test case already uses chaining.

```suggestion
```

nit, chaining isn't really used in main code, so those methods should just return void.

importer/src/test/java/org/hiero/mirror/importer/parser/record/transactionhandler/CryptoTransferTransactionHandlerTest.java
ashumahajan: Added

jnels124 (approved):

PR #12353, opened 2025-11-07
Updates openapi.yml

### LINKED ISSUE #12352: /transactions endpoint defaults to cryptoTransfer for transaction type even though it's not selected in the drop down list
### Description

When trying out the `/transactions` endpoint using swagger (https://testnet.mirrornode.hedera.com/api/v1/docs/), swagger sets the transaction type to `cryptotransfer` even though it's not selected in the list.

Swagger should faithfully reflect the user's selection

### Steps to reproduce

1. navigate to https://testnet.mirrornode.hedera.com/api/v1/docs/
2. scroll to the /transactions endpoint and click "try it out"
3. leave all defaults blank and click on "Execute"

The request url generated by swagger is https://testnet.mirrornode.hedera.com/api/v1/transactions?limit=2&order=asc&transactiontype=cryptotransfer which includes a `transactiontype=cryptotransfer` even though no transaction type was selected from the drop down list

### Additional context

_No response_

### Hedera network

_No response_

### Version

0.142.0

### Operating system

None

gregscullard:
**Description**:
-All transaction types in drop down list shown against the /transactions endpoint in swagger in camel case for readability 
-Removes invalid transaction types (node and unknown) 
-Removes default cryptoTransfer value for /transactions endpoint so users don't have to unset, saving a few clicks and scrolling through a long list of transactions

**Related issue(s)**:

Fixes #12352 

**Notes for reviewer**:
This improves usability by renaming transaction types so that they are more readable, removes invalid transaction types from the list and removes a default which otherwise requires users to unselect the transaction type from a long list of transactions.

**Checklist**

- [X] Tested using local swagger test environment, all renamed transaction types resolve to a working api call, ensuring no typos were introduced


rest/api/v1/openapi.yml
```diff
@@ -3733,62 +3733,60 @@ components:
     TransactionTypes:
       type: string
       enum:
-        - ATOMICBATCH
```
steven-sheehy, line 3736: This change [breaks](https://github.com/hiero-ledger/hiero-mirror-node/actions/runs/19167221806/job/54790666197?pr=12353) the compilation of our test module. It also generates Java code with non-standard naming for enums. Similarly, it would be a breaking change to users that generate client code based upon our OpenAPI schema. We should revert the renaming.

steven-sheehy (approved):

xin-hedera (approved):
LGTM

PR #12355, opened 2025-11-07
enhance copy live env script

### LINKED ISSUE #12318: Copy stackgres script enhancements
### Problem

We have additional disk resources sticking around in the dormant cluster

### Solution

- we should delete all pvcs used by the common chart and let them be recreated on reconcile (ensure disks are deleted as well)
- wait for helm tests to be successful before suspending release, delete acceptance pod if present

### Alternatives

_No response_

jnels124:
**Description**:
* verify helm tests pass and wait for manual rerun if not
* delete common chart pvcs + disks

**Related issue(s)**:

Fixes #12318 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/tools/cluster-management/copy-live-environment.sh
+++ b/tools/cluster-management/copy-live-environment.sh
@@ -619,11 +619,14 @@ function waitForHelmReleaseReady() {
 
     sleep 30
   done
+}
 
+function cleanupAcceptancePod() {
+  local namespace="$1"
   kubectl -n "${namespace}" wait pod -l"app.kubernetes.io/component=hedera-mirror" \
-    --for=jsonpath='{.status.phase}'=Succeeded \
-    --timeout=30m > /dev/null 2>&1 || true
-  kubectl -n "${namespace}" delete pod -l"app.kubernetes.io/component=hedera-mirror" --ignore-not-found >/dev/null 2>&1
+      --for=jsonpath='{.status.phase}'=Succeeded \
+      --timeout=30m || true
+  kubectl -n "${namespace}" delete pod -l"app.kubernetes.io/component=hedera-mirror" --ignore-not-found
 }
 
 if [[ -z "${K8S_SOURCE_CLUSTER_CONTEXT}" || -z "${K8S_TARGET_CLUSTER_CONTEXT}" ]]; then
@@ -646,6 +649,7 @@ if [[ "${WAIT_FOR_K6}" == "true" ]]; then
   for namespace in "${TEST_KUBE_TARGET_NAMESPACE[@]}"; do
     if kubectl get helmrelease -n "${namespace}" "${HELM_RELEASE_NAME}" >/dev/null 2>&1; then
       waitForHelmReleaseReady "${namespace}"
+      cleanupAcceptancePods "${namespace}"
       log "Suspending HelmRelease ${HELM_RELEASE_NAME} in namespace ${namespace}"
       flux suspend helmrelease -n "${namespace}" "${HELM_RELEASE_NAME}"
     fi


tools/cluster-management/copy-live-environment.sh
```diff
@@ -577,6 +581,51 @@ function waitForK6PodExecution() {
...
+  local namespace="$1"
+  local hrJson
+  local testsEnabled
+  local testStatus
+  local testMsg
+  local readyStatus
+  local readyReason
+
+  log "Waiting for helm tests to finish in namespace ${namespace}"
+  while true; do
```
steven-sheehy, line 594: Shouldn't we have some max amount of time to loop waiting for it to complete? Or you want that to be handled by job timeout?

tools/cluster-management/copy-live-environment.sh
```diff
@@ -577,6 +581,51 @@ function waitForK6PodExecution() {
...
+    readyReason="$(jq -r '.status.conditions[]? | select(.type=="Ready") | .reason // empty' <<<"${hrJson}")"
+
+    if [[ "${testStatus}" == "True" ]] || { [[ "${readyStatus}" == "True" && "${readyReason}" == "TestSucceeded" ]]; }; then
+      log "Helm tests SUCCEEDED for ${HELM_RELEASE_NAME} in ${namespace}"
+      break
+    fi
+
+    if [[ "${testStatus}" == "False" ]]; then
+      log "Helm tests FAILED for ${HELM_RELEASE_NAME} in ${namespace} ${testMsg}"
+      log "Waiting for manual intervention to re-run tests and succeed..."
```
steven-sheehy, line 615: So we keep the env up if the tests fail instead of shutting down and retrying via a new workflow?

tools/cluster-management/copy-live-environment.sh
```diff
@@ -577,6 +581,51 @@ function waitForK6PodExecution() {
...
+    else
+      log "Helm tests running or not yet reported for ${HELM_RELEASE_NAME} in ${namespace}"
+    fi
+
+    sleep 30
+  done
+
+  kubectl -n "${namespace}" wait pod -l"app.kubernetes.io/component=hedera-mirror" \
+    --for=jsonpath='{.status.phase}'=Succeeded \
```
steven-sheehy: Is it possible the pod is in a failure state here?

tools/cluster-management/copy-live-environment.sh
```diff
@@ -577,6 +581,51 @@ function waitForK6PodExecution() {
...
+    --timeout=30m > /dev/null 2>&1 || true
+  kubectl -n "${namespace}" delete pod -l"app.kubernetes.io/component=hedera-mirror" --ignore-not-found >/dev/null 2>&1
```
steven-sheehy: Seems out of place to delete pods in a `waitForHelmReleaseReady` function

tools/cluster-management/copy-live-environment.sh
jnels124: I do a lot of infinite loops like this in these scripts so manual intervention can occur if needed

tools/cluster-management/copy-live-environment.sh
jnels124: yes... We keep the env up so we can just do a flux reconcile and have the script continue. At this point, snapshots etc were already taken and restored so it saves on costs and time but not having to create a new snapshot/restored disks. If we need to make actual modifications to the helm release, this job will get canceled by the next push because we only allow one of these workflows to be running at any time and we always prefer the new one starting vs the old completing

tools/cluster-management/copy-live-environment.sh
jnels124: loop above will only exit once tests are successful so if it ever gets to this point, we have completed tests with a success status. This line isn't technically needed because of that.

tools/cluster-management/copy-live-environment.sh
jnels124: this is the acceptance pod. We delete because it is what triggers the k6 run. Usually the acceptance pod deletes once test passes but sometimes it sticks around for a bit

steven-sheehy: I just meant you could move it outside waitForHelmReleaseReady to the calling function

jnels124: I can move this though since it is a little weird to be in this function

steven-sheehy (approved):
LGTM

PR #12409, opened 2025-11-14
Bump acceptance workflow dependencies

bilyana-gospodinova:
**Description**:
Bump the mirror node and consensus node version in the acceptance workflow so that it will run successfully in CI.

### DIFF
--- a/.github/workflows/acceptance.yaml
+++ b/.github/workflows/acceptance.yaml
@@ -43,12 +43,12 @@ jobs:
     env:
       BASE_URL: "http://127.0.0.1:8081"
       BLOCK_NODE_CHART_VERSION: v0.20.1
-      CONSENSUS_VERSION: v0.67.2
+      CONSENSUS_VERSION: v0.68.0
       DEFAULT_DURATION: "3s"
       DEFAULT_GRACEFUL_STOP: "1s"
       DEFAULT_VUS: 2
       HELM_RELEASE_NAME: mirror-1
-      MIRROR_NODE_VERSION: v0.142.0
+      MIRROR_NODE_VERSION: v0.143.0-rc2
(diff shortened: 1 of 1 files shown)


test/src/test/java/org/hiero/mirror/test/e2e/acceptance/client/MirrorNodeClient.java
```diff
@@ -375,7 +375,7 @@ public NetworkStakeResponse getNetworkStake() {
 
     public NetworkFeesResponse getNetworkFees() {
         String feesEndpoint = "/network/fees";
-        return callConvertedRestEndpoint(feesEndpoint, NetworkFeesResponse.class);
```
xin-hedera, line 378: the correct fix is bump up mirror node version in acceptance workflow to v0.143.0-rc2

bilyana-gospodinova: The rest-java endpoint is still disabled there.

xin-hedera: when we run acceptance test inside the k8s cluster, for rest-java, the acceptance test rest api client directly uses the rest-java k8s service, not through a front-end proxy.

The issue we hit in the acceptance tests is because rest-java 0.142.x doesn't have `/network/fees` support. That's why I say the fix is bump up to v0.143.0-rc2.

bilyana-gospodinova: I see, thanks. The versions are updated as suggested.

PR #12430, opened 2025-11-17
HIP-1056 Support consensus node 0.68 blockstream spec changes

### LINKED ISSUE #12428: HIP-1056 Consensus node 0.68.x blockstream spec changes
### Problem

Mirrornode needs to support blockstream spec changes in consensus node release 0.68.x

- new `BlockFooter` block item, immediately before the first `BlockProof`. Note we need to treat it as optional though it's a required block item, because it doesn't exist in blockstream created by consensus node 0.67.x
- multiple `BlockProof` block items; note to be compatible w/ 0.67, only read data from the first `BlockProof` block item
- system transactions which appear as signedTransaction block item w/out transaction result

When all environments are upgraded to post 0.68 release, we should remove the workaround for earlier consensus node releases.

### Solution

Add support of updated blockstream spec with workarounds for prior releases.

### Alternatives

_No response_

xin-hedera:
**Description**:

This PR addresses some blockstream spec change in consensus node 0.68

- Read `BlockFooter` block item
- Handle multiple `BlockProof` block items
- Handle system transaction signedTransaction block item without transactionResult
- Workaround blockchain verification for >= 0.68.1

**Related issue(s)**:

Fixes #12428 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockStreamVerifier.java
+++ b/importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockStreamVerifier.java
@@ -145,7 +145,8 @@ final class BlockStreamVerifier {
         final var version = new Version(
                 consensusNodeVersion.getMajor(), consensusNodeVersion.getMinor(), consensusNodeVersion.getPatch());
         if (version.isGreaterThanOrEqualTo(blockProperties.getNewRootHashAlgorithmVersion())) {
-            // Set both hash and previousHash to all 0s to pass parser validation
+            // Set both hash and previousHash to all 0s to pass parser validation, will remove when the complete support
+            // of redesigned block and state merkle tree is added
             blockFile.setHash(EMPTY_HASH);
             blockFile.setPreviousHash(EMPTY_HASH);
             return;
--- a/importer/src/main/java/org/hiero/mirror/importer/reader/block/BlockStreamReaderImpl.java
+++ b/importer/src/main/java/org/hiero/mirror/importer/reader/block/BlockStreamReaderImpl.java
@@ -35,6 +35,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import lombok.CustomLog;
 import lombok.Setter;
 import lombok.Value;
 import lombok.experimental.NonFinal;
@@ -45,6 +46,7 @@ import org.hiero.mirror.common.util.DomainUtils;
 import org.hiero.mirror.importer.exception.InvalidStreamFileException;
 import org.jspecify.annotations.NonNull;
 
+@CustomLog
 @Named
 public final class BlockStreamReaderImpl implements BlockStreamReader {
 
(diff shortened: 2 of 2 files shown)


importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockStreamVerifier.java
```diff
@@ -142,6 +145,9 @@ private void verifyHashChain(BlockFile blockFile) {
         final var version = new Version(
                 consensusNodeVersion.getMajor(), consensusNodeVersion.getMinor(), consensusNodeVersion.getPatch());
         if (version.isGreaterThanOrEqualTo(blockProperties.getNewRootHashAlgorithmVersion())) {
+            // Set both hash and previousHash to all 0s to pass parser validation
+            blockFile.setHash(EMPTY_HASH);
```
xin-hedera, line 150: we don't calculate the block hash for >= 0.68.1 because there will be breaking upstream changes. As a result, we have to set both hash and previous hash to all 0s here to parser can continue.

importer/src/main/java/org/hiero/mirror/importer/reader/block/BlockRootHashDigest.java
```diff
@@ -72,16 +75,6 @@ public String digest() {
...
 
-    public void setPreviousHash(byte[] previousHash) {
-        validateHash(previousHash, "previousHash");
-        this.previousHash = previousHash;
-    }
-
-    public void setStartOfBlockStateHash(byte[] startOfBlockStateHash) {
-        validateHash(startOfBlockStateHash, "startOfBlockStateHash");
-        this.startOfBlockStateHash = startOfBlockStateHash;
-    }
```
xin-hedera, line 83: as of 0.68.1, these fields are no longer set in `BlockProof`

importer/src/main/java/org/hiero/mirror/importer/reader/block/BlockStreamReaderImpl.java
```diff
@@ -121,6 +128,9 @@ private void readBlockProof(ReaderContext context) {
         blockFile.blockProof(blockProof).previousHash(DomainUtils.bytesToHex(previousHash));
         blockRootHashDigest.setPreviousHash(previousHash);
         blockRootHashDigest.setStartOfBlockStateHash(DomainUtils.toBytes(blockProof.getStartOfBlockStateRootHash()));
+
+        // Read remaining blockProof block items. In a later release, implement new block & state merkle tree support
+        while (context.readBlockItemFor(BLOCK_PROOF) != null) {}
```
steven-sheehy: Add a debug log for now to avoid code smells due to empty block

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockStreamVerifier.java
steven-sheehy: Does this need to be removed later? If so, add a comment about when to remove.

xin-hedera: sure

steven-sheehy (approved):

PR #12436, opened 2025-11-18
Add support for historical blocks in K6 web3 test suite

### LINKED ISSUE #12399: Add K6 web3 test coverage for historical blocks
### Problem

We covered a lot of the web3 functionality with K6 performance tests. However, all of the tests are set up to use the default LATEST block. Performance test coverage needs to be increased using historical blocks as well.

### Solution

Find a way to reuse existing K6 tests and infrastructure to cover historical blocks and inspect DB queries including timestamp as a filter.

### Alternatives

_No response_

IvanKavaldzhiev:
**Description**:
This PR enhances the default behaviour of the K6 web3 tests to accept list of blocks which to use for test execution. For each separate block a new test run will be created. There is a new env variable property that can be set with a historical block to be used for the tests run.

**Related issue(s)**:

Fixes #12399 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/tools/k6/src/web3/test/contractCallEstimateReadStorage.js
+++ b/tools/k6/src/web3/test/contractCallEstimateReadStorage.js
@@ -8,7 +8,7 @@ const runMode = __ENV.RUN_WITH_VARIABLES;
 const data = __ENV.STORAGE_SLOTS_CALLDATA;
 const testName = 'estimateReadStorageSlots';
 
-// If RUN_WITH_VARIABLES=false, use the generic estimate template; otherwise build from provided ENV variables
+//If RUN_WITH_VARIABLES=true will run tests with __ENV variables
 const {options, run} =
   runMode === 'false'
     ? new ContractCallEstimateTestTemplate(testName, false)


tools/k6/src/web3/test/common.js
```diff
@@ -62,6 +62,37 @@ const loadVuDataOrDefault = (filepath, key) =>
     return key in data ? data[key] : [];
   });
 
+function sanitizeScenarioName(name) {
```
bilyana-gospodinova: This function already exists in "src/lib/common.js" and can be reused.

tools/k6/README.md
```diff
@@ -132,6 +132,7 @@ The following parameters can be used to configure a web3 test:
 | ESTIMATE_PRECOMPILE_CONTRACT                              |         | 40 character hex encoded contract address without `0x` prefix for `EstimatePrecompileContract.sol`                                                                                |
 | FUNGIBLE_TOKEN_WITH_FREEZE_KEY_ADDRESS                    |         | 64 character hex encoded fungible token address without `0x` prefix - with a freeze key set                                                                                       |
 | FUNGIBLE_TOKEN_WITH_FREEZE_KEY_ASSOCIATED_ACCOUNT_ADDRESS |         | 64 character hex encoded account evm address without `0x` prefix - an account associated to FUNGIBLE_TOKEN_WITH_FREEZE_KEY_ADDRESS                                                |
+| HISTORICAL_BLOCK_NUMBER                                   |         | 19 decimal digit numeric value representing a block number used to run a test scenario with this block number passed in the request body                                          |
```
bilyana-gospodinova: Maybe we should add that if the value is not provided, it will default to "earliest" block.

tools/k6/src/web3/test/common.js
```diff
@@ -62,6 +62,37 @@ const loadVuDataOrDefault = (filepath, key) =>
...
+  // k6 requires [0-9A-Za-z_-]
+  return name.replace(/[^0-9A-Za-z_-]/g, '_');
+}
+
+const HISTORICAL_BLOCK_NUMBER = __ENV.HISTORICAL_BLOCK_NUMBER || 'earliest';
+
+function toHexBlockNumber(value) {
+  const v = (value ?? '').toString().trim();
+  if (v === '' || v === 'earliest' || v === 'latest' || v === 'pending' || /^0x[0-9a-fA-F]+$/.test(v)) {
```
bilyana-gospodinova: I think "safe" and "finalized" can be passed as well as values and they should match "latest".

IvanKavaldzhiev: That's correct, I've added them.

tools/k6/README.md
IvanKavaldzhiev: I changed it to default to 'latest' as it makes more sense. We won't perform historical tests for each test case.

tools/k6/src/web3/test/common.js
IvanKavaldzhiev: Added an import.

tools/k6/src/web3/test/modificationTests/contractCallPrecompileDeleteToken.js
```diff
@@ -18,6 +18,7 @@ const {options, run} =
         .selector(selector)
         .args([token])
         .to(contract)
+        .blocks(getMixedBlocks())
```
xin-hedera, line 21: isn't this the default so we can revert all such changes?

IvanKavaldzhiev: The default is `blocks: ['latest'],` so for the rest of the test cases only `latest` block will be used. We need to explicitly define which test cases to use for both `latest` and `historical` blocks.

xin-hedera: it's a little hard to follow. this is what I read from the code

1. `defaultVuData.block` is replaced by `defaultVuData.blocks` and the value is `['latest']`
2. `defaultVuData` is only used when there is no `_selector` or `args` AND there is valid `_vuData`
3. in `build`, `getMixedBlocks` is called when `_blocks` is not set or empty, and this happens at the very beginning

From the above, my conclusions are:

- calling `.blocks(getMixedBlocks())` is redundant due to 3
- we should not replace `block` with `blocks` in `defaultVuData` since that's the payload sent as the post request body

IvanKavaldzhiev: I changed the default behaviour of 'blocks' to be initialization with `['latest']` array.

I removed the defaultVuData.block field.

Only the test cases need to be ran with a mix of latest and historical blocks are using `getMixedBlocks`. All of the rest will have a single run with `latest` block only.

PR #12451, opened 2025-11-21
Historical balance updates

jnels124:
**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/importer/src/main/java/org/hiero/mirror/importer/parser/record/historicalbalance/HistoricalBalanceService.java
+++ b/importer/src/main/java/org/hiero/mirror/importer/parser/record/historicalbalance/HistoricalBalanceService.java
@@ -138,20 +138,38 @@ public class HistoricalBalanceService {
                 boolean full = maxConsensusTimestamp.isEmpty();
                 int accountBalancesCount;
                 int tokenBalancesCount;
+
                 if (full) {
                     // get a full snapshot
                     accountBalancesCount = accountBalanceRepository.balanceSnapshot(timestamp, treasuryAccountId);
                     tokenBalancesCount = properties.isTokenBalances()
                             ? tokenBalanceRepository.balanceSnapshot(timestamp, treasuryAccountId)
                             : 0;
+                    tokenBalanceRepository.deleteSnapshotRows();
+                    accountBalanceRepository.deleteSnapshotRows();
                 } else {
                     // get a snapshot that has no duplicates
                     accountBalancesCount = accountBalanceRepository.balanceSnapshotDeduplicate(
                             maxConsensusTimestamp.get(), timestamp, treasuryAccountId);
+                    final var deletedAccountBalanceChangeRows = accountBalanceRepository.deleteSnapshotRows();
                     tokenBalancesCount = properties.isTokenBalances()
                             ? tokenBalanceRepository.balanceSnapshotDeduplicate(
                                     maxConsensusTimestamp.get(), timestamp, treasuryAccountId)
                             : 0;
+                    final var deletedTokenBalanceChangeRows = tokenBalanceRepository.deleteSnapshotRows();
+
+                    if (deletedAccountBalanceChangeRows != accountBalancesCount) {
+                        log.error(
+                                "Deleted account_balance_change rows {} does not match inserted account_balance rows {}",
+                                deletedAccountBalanceChangeRows,
+                                accountBalancesCount);
+                    }
+                    if (deletedTokenBalanceChangeRows != tokenBalancesCount) {
+                        log.error(
+                                "Deleted token_balance_change rows {} does not match inserted token_balance rows {}",
+                                deletedTokenBalanceChangeRows,
+                                tokenBalancesCount);
+                    }
                 }
 
                 long loadEnd = System.currentTimeMillis();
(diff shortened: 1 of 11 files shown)


importer/src/main/java/org/hiero/mirror/importer/parser/record/historicalbalance/HistoricalBalanceService.java
```diff
@@ -138,20 +138,38 @@ public void onRecordFileParsed(RecordFileParsedEvent event) {
...
                     accountBalancesCount = accountBalanceRepository.balanceSnapshotDeduplicate(
                             maxConsensusTimestamp.get(), timestamp, treasuryAccountId);
+                    final var deletedAccountBalanceChangeRows = accountBalanceRepository.deleteSnapshotRows();
                     tokenBalancesCount = properties.isTokenBalances()
                             ? tokenBalanceRepository.balanceSnapshotDeduplicate(
                                     maxConsensusTimestamp.get(), timestamp, treasuryAccountId)
                             : 0;
+                    final var deletedTokenBalanceChangeRows = tokenBalanceRepository.deleteSnapshotRows();
+
+                    if (deletedAccountBalanceChangeRows != accountBalancesCount) {
```
xin-hedera, line 161: think this will be `true` when `treasuryAccountId` is not in account_balance_change table.

importer/src/main/java/org/hiero/mirror/importer/parser/record/historicalbalance/HistoricalBalanceService.java
```diff
@@ -138,20 +138,38 @@ public void onRecordFileParsed(RecordFileParsedEvent event) {
...
                 int tokenBalancesCount;
+
                 if (full) {
                     // get a full snapshot
                     accountBalancesCount = accountBalanceRepository.balanceSnapshot(timestamp, treasuryAccountId);
                     tokenBalancesCount = properties.isTokenBalances()
                             ? tokenBalanceRepository.balanceSnapshot(timestamp, treasuryAccountId)
                             : 0;
+                    tokenBalanceRepository.deleteSnapshotRows();
+                    accountBalanceRepository.deleteSnapshotRows();
```
xin-hedera, line 149: there will be issues when we have concurrent database transactions changing data in the same table

- delete all rows
- the upsert SQL

The two database transactions will see the same data. However I believe the upsert SQL is changed to do nothing if a row already exists, and we'll end up losing track of such entities.

I don't know how to solve it since it depends on different types of lockings.

On the other hand, it looks like at least for account balance change, we could use roaring bitmap

- For each record file, a 64-bit roaring bitmap of the entities with balance change is built
- the roaring bitmap becomes a field in `RecordFileParsedEvent` and the whole event is sent to `HistoricalBalanceService`
- `HistoricalBalanceService` needs a deterministic algorithm to xor roaring bitmaps up to the last record file's consensus end in the `transactionTemplate.executeWithoutResult` block

We can't easily do the same for token account though since there are two 64-bit entity ids.

We also need to figure out a way to build the initial roaring bitmap on application start.

importer/src/main/java/org/hiero/mirror/importer/parser/record/historicalbalance/HistoricalBalanceService.java
jnels124: You are right, i forgot to update this part when realized i needed a union all on the snapshot generation. Will remove or adjust this check

importer/src/main/java/org/hiero/mirror/importer/parser/record/historicalbalance/HistoricalBalanceService.java
jnels124: This should not cause issues with different transactions writing because the transaction is in repeatable read mode. I have some tests i will do to esnure that assumption but since the delete is a data statement and not a ddl statement like drop, It should work. Otherwise we can have the delete be more timestamp aware

jnels124: Oh i see. There is the on conflict issue that would prevent the new snapshot from recording its entry. Yes that is a problem.

jnels124: The only way i can see to handle both entity and token account balances is one of the following.
1. We add a BRIN index to balance_timestamp for both entity and token_account

- Pros:
* makes balance snapshots trivial 
* BRIN index stay relatively small and work well for our use case
* requires no code changes as original queries will be fast enough with a very large entity count
* No additional tables or duplicating data
* No data inconsistencies
- Cons:
* lack of familiarity with maintenance performance. Researching this topic it seems the cost wouldn't be extremely high because how this particular index functions but is still worth mentioning

2. We keep the additional change tables and partition each snapshot window.
- Pros:
* no additional index to maintain
* partition drops are cheap
- Cons:
* will have to have logic to create/prune partitions 
* additional application state needs to be maintained to insert snapshot from correct partition. 
* lots of ddl churn
* provides minimal (if any) savings(db size/performance) vs option 1 and complexity is higher
* Harder to ensure correctness

3. Have a ddl swap that occurs to where we are inserting the snapshot using `balance_change_ready` while writing to `balance_change` table
- Pros:
* no additional index to maintain
- Cons:
* will need to have "complicated" logic to ensure snapshots and new updates are handled appropriately
* provides minimal (if any) savings(db size/performance) vs option 1
* Testing complexity

In summary, I really think option 1 is the correct way to go here. It is simple to reason about, has no extra tables or complex logic.

PR #12474, opened 2025-11-24
Convert /network/supply to Java

### LINKED ISSUE #12302: Convert `/network/supply` API to Java
### Problem

As a maintainer, I'd like to convert `/api/v1/network/supply` to Java to ease maintenance and improve performance.

### Solution

See below comment for checklist.

### Alternatives

_No response_

martingeorgiev1:
**Description**:
Add /api/v1/network/supply to rest-java module
Add k6 tests for rest-java network supply
Add routes.supply* with a default of false to Helm chart

**Related issue(s)**:

Fixes #12302 

**Notes for reviewer**:
Tested on local machine and the results:

Rest Endpoint:
```
    ✓ 'p(95)<500' p(95)=13.23ms
    checks_total.......: 34110   1136.740482/s
    checks_succeeded...: 100.00% 34110 out of 34110
    checks_failed......: 0.00%   0 out of 34110
```

After switching to java-rest endpoint:
```
    ✓ 'p(95)<500' p(95)=2.71ms
    checks_total.......: 165149  5504.656571/s
    checks_succeeded...: 100.00% 165149 out of 165149
    checks_failed......: 0.00%   0 out of 165149
```
Result:
~5× lower latency (p95)
~5× higher throughput

**Checklist**

### DIFF
--- a/rest-java/src/test/java/org/hiero/mirror/restjava/controller/NetworkControllerTest.java
+++ b/rest-java/src/test/java/org/hiero/mirror/restjava/controller/NetworkControllerTest.java
@@ -25,6 +25,7 @@ import org.hiero.mirror.rest.model.NetworkExchangeRateSetResponse;
 import org.hiero.mirror.rest.model.NetworkFeesResponse;
 import org.hiero.mirror.rest.model.NetworkStakeResponse;
 import org.hiero.mirror.rest.model.NetworkSupplyResponse;
+import org.hiero.mirror.restjava.common.EntityIdParameter;
 import org.hiero.mirror.restjava.config.NetworkProperties;
 import org.hiero.mirror.restjava.dto.SystemFile;
 import org.hiero.mirror.restjava.mapper.CommonMapper;
@@ -548,7 +549,7 @@ final class NetworkControllerTest extends ControllerTest {
         }
 
         @Test
-        void success() {
+        void getNetworkSupplyNoQueryParams() {
             // given
             createUnreleasedSupplyAccounts(null);
 
@@ -798,7 +799,8 @@ final class NetworkControllerTest extends ControllerTest {
         }
 
         private void createCustomBalance(long accountNum, long balance, long timestamp) {
-            final var accountId = EntityId.of(0, 0, accountNum);
+            final var entityIdParameter = EntityIdParameter.valueOf(String.valueOf(accountNum));
+            final var accountId = EntityId.of(entityIdParameter.shard(), entityIdParameter.realm(), accountNum);
             domainBuilder
                     .accountBalance()
                     .customize(ab -> ab.balance(balance).id(new AccountBalance.Id(timestamp, accountId)))


rest-java/src/main/java/org/hiero/mirror/restjava/config/NetworkProperties.java
```diff
@@ -0,0 +1,34 @@
...
+
+import jakarta.validation.constraints.NotNull;
+import java.util.List;
+import lombok.Data;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.validation.annotation.Validated;
+
+@Data
+@Validated
+@ConfigurationProperties("hiero.mirror.rest.network")
```
steven-sheehy:
```suggestion
@ConfigurationProperties("hiero.mirror.rest-java.network")
```

rest-java/src/main/java/org/hiero/mirror/restjava/controller/NetworkController.java
```diff
@@ -59,7 +71,43 @@ NetworkStakeResponse getNetworkStake() {
         return networkStakeMapper.map(networkStake);
     }
 
+    @GetMapping("/supply")
+    ResponseEntity<?> getSupply(
+            @RequestParam(required = false) @Size(max = 2) TimestampParameter[] timestamp,
+            @RequestParam(required = false) String q) {
+        // Validate q parameter
+        if (q != null && !(PARAM_Q_TOTALCOINS.equalsIgnoreCase(q) || PARAM_Q_CIRCULATING.equalsIgnoreCase(q))) {
```
steven-sheehy: Should make it an enum

rest/__tests__/specs/network/supply/q-totalcoins-HBARS.json
```diff
@@ -30,8 +30,13 @@
...
   "url": "/api/v1/network/supply?q=totalcoins",
-  "responseHeaders": {
-    "Content-Type": "text/plain; charset=utf-8"
+  "responseHeadersMatrix": {
+    "js": {
+      "Content-Type": "text/plain; charset=utf-8"
+    },
+    "java": {
+      "Content-Type": "text/plain;charset=UTF-8"
+    }
```
steven-sheehy: It would be easier to just adjust the `NetworkController.contentTypeTextPlain` to remove the space.

tools/k6/src/rest-java/test/networkSupply.js
steven-sheehy, line 1: Need to add this to index.js so it runs

rest-java/src/main/java/org/hiero/mirror/restjava/repository/NetworkSupplyRepositoryCustomImpl.java
```diff
@@ -0,0 +1,137 @@
...
+import org.hiero.mirror.restjava.service.Bound;
+import org.jooq.Condition;
+import org.jooq.DSLContext;
+import org.jooq.Field;
+import org.jooq.Record;
+import org.jooq.impl.DSL;
+
+@Named
+@RequiredArgsConstructor
+final class NetworkSupplyRepositoryCustomImpl implements NetworkSupplyRepositoryCustom {
```
steven-sheehy: There's no such thing as a network_supply table so we should not have a NetworkSupplyRepository. This would go in an EntityRepository.

rest-java/src/main/java/org/hiero/mirror/restjava/repository/NetworkSupplyRepositoryCustomImpl.java
```diff
@@ -0,0 +1,137 @@
...
+    private final DSLContext dslContext;
+    private final NetworkProperties networkProperties;
+
+    @Override
+    public NetworkSupply getSupply(Bound timestamp) {
+        if (timestamp.isEmpty()) {
+            return getSupplyFromEntity();
+        }
+        return getSupplyFromAccountBalance(timestamp);
+    }
```
steven-sheehy: We don't need dynamic queries for the supply logic. Expand the from/to accounts into a list of account IDs and use in an IN clause. Create two static queries: 

* `EntityRepository.getSupply(List<Long> unreleasedSupplyAccounts)`
* `AccountBalanceRepository.getSupply(List<Long> unreleasedSupplyAccounts, long lowerTimestamp, long upperTimestamp)`

rest-java/src/main/java/org/hiero/mirror/restjava/config/NetworkProperties.java
```diff
@@ -0,0 +1,34 @@
...
+public class NetworkProperties {
+
+    @NotNull
+    private String currencyFormat = "BOTH";
```
steven-sheehy: We can drop this property and logic and just always return hbars.tinybars. We added it originally since we weren't sure which format coinmarketcap actually wanted in the initial rollout. We should drop it in both js and java code so we don't have differences in tests.

rest-java/src/main/java/org/hiero/mirror/restjava/config/NetworkProperties.java
```diff
@@ -0,0 +1,34 @@
...
+
+    @NotNull
+    private List<AccountRange> unreleasedSupplyAccounts = List.of(
```
steven-sheehy, line 23: Need to document in configuration.md

rest-java/src/main/java/org/hiero/mirror/restjava/dto/NetworkSupply.java
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package org.hiero.mirror.restjava.dto;
+
+import java.math.BigInteger;
+
+public record NetworkSupply(String releasedSupply, Long consensusTimestamp, String totalSupply) {
```
steven-sheehy: We should use Spring Data projection to return this from the repository calls. Can convert to interface if needed

rest-java/src/main/java/org/hiero/mirror/restjava/mapper/NetworkSupplyMapper.java
```diff
@@ -0,0 +1,56 @@
...
+            return null;
+        }
+
+        return new NetworkSupplyResponse()
+                .releasedSupply(networkSupply.releasedSupply())
+                .timestamp(DomainUtils.toTimestamp(networkSupply.consensusTimestamp()))
+                .totalSupply(networkSupply.totalSupply());
+    }
+
+    @Named("convertToCurrencyFormat")
```
steven-sheehy: I don't think any of the `@Named` are necessary in this class.

rest-java/src/main/java/org/hiero/mirror/restjava/dto/NetworkSupply.java
steven-sheehy: We should keep DTO fields as long and convert to string in controller.
```suggestion
public record NetworkSupply(long releasedSupply, long consensusTimestamp, long totalSupply) {
```

rest-java/src/main/java/org/hiero/mirror/restjava/repository/NetworkSupplyRepositoryCustom.java
```diff
@@ -0,0 +1,19 @@
...
+
+public interface NetworkSupplyRepositoryCustom extends JooqRepository {
+
+    /**
+     * Get the network supply at the latest timestamp or at a specific timestamp if provided.
+     *
+     * @param timestamp The timestamp bound for the query
+     * @return The network supply
+     */
+    @NotNull
```
steven-sheehy: We should only use jspecify annotations. Here, we already have repository package-info.java with @NullMarked so don't need any annotations

rest-java/src/main/resources/application.yml
```diff
@@ -2,6 +2,24 @@
 
 hiero:
   mirror:
+    rest:
+      network:
+        unreleasedSupplyAccounts:
```
steven-sheehy: Remove from here since we already specify in NetworkProperties.java

rest-java/src/test/java/org/hiero/mirror/restjava/mapper/NetworkSupplyMapperTest.java
```diff
@@ -0,0 +1,193 @@
...
+final class NetworkSupplyMapperTest {
+
+    private final DomainBuilder domainBuilder = new DomainBuilder();
+    private CommonMapper commonMapper;
+    private NetworkSupplyMapper mapper;
+
+    @BeforeEach
+    void setup() {
+        commonMapper = new CommonMapperImpl();
+        mapper = new NetworkSupplyMapperImpl();
```
steven-sheehy: Can just move to variable declaration and make final

rest-java/src/test/java/org/hiero/mirror/restjava/controller/NetworkControllerTest.java
```diff
@@ -524,4 +524,214 @@ void notFound() {
...
+        @Override
+        protected RequestHeadersSpec<?> defaultRequest(RequestHeadersUriSpec<?> uriSpec) {
+            // Create test entities with balances across unreleased supply account ranges
+            // These ranges are: 2-2, 42-42, 44-71, 73-87, 99-100, 200-349, 400-750
+            // Total unreleased: 3,500,000,000,000,000,000 (3.5 quadrillion tinybars)
+            // Released: 1,500,000,000,000,000,000 (1.5 quadrillion tinybars)
+
+            long timestamp = 1000000000L;
+            long shard = 0L;
+            long realm = 0L;
```
steven-sheehy: This does not work with non-zero realm.

rest-java/src/test/java/org/hiero/mirror/restjava/controller/NetworkControllerTest.java
```diff
@@ -524,4 +524,214 @@ void notFound() {
...
+
+            // Account 0.0.2: 2 quadrillion tinybars
+            domainBuilder
+                    .entity()
+                    .customize(e -> e.id(EntityId.of(shard, realm, 2L).getId())
+                            .balance(2000000000000000000L)
+                            .balanceTimestamp(timestamp))
+                    .persist();
```
steven-sheehy: There's too much duplicated data setup in each test. Create a helper method to create all accounts. For balance, let's just set a static amount like 1M hbars that we can use to detect how many accounts are in the response. We don't need to set a static balanceTimestamp for entities.

```java
private void createUnreleasedSupplyAccounts(Long timestamp) {
    networkProperties.getUnreleasedSupplyAccountIds().forEach(id -> {
        final long balance = 1_000_000 * DomainUtils.TINYBARS_IN_ONE_HBAR;

        if (timestamp != null) {
            domainBuilder.accountBalance().customize(ab -> ab.balance(balance).id(new Id(timestamp, id)).persist();
        } else {
            domainBuilder.entity().customize(e -> e.id(id).balance(balance).persist();
        }
    });
}
```

rest-java/src/test/java/org/hiero/mirror/restjava/controller/NetworkControllerTest.java
```diff
@@ -524,4 +524,214 @@ void notFound() {
                     "No network stake data found");
         }
     }
+
+    @DisplayName("/api/v1/network/supply")
+    @Nested
+    final class SupplyEndpointTest extends EndpointTest {
```
steven-sheehy, line 536: This is missing a lot of test cases like timestamp parameter especially negative test cases like not found, invalid params, etc. One of the checklist items is `Controller integration tests added with equivalent or better coverage to existing rest spec tests.` Every json spec test should be represented here but with dynamic input.

rest-java/src/test/java/org/hiero/mirror/restjava/mapper/NetworkSupplyMapperTest.java
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    @Test
+    void map() {
+        // given
+        final var consensusTimestamp = domainBuilder.timestamp();
+        final var networkSupply = new NetworkSupply("1593057374165809903", consensusTimestamp, "5000000000000000000");
```
steven-sheehy: Use constant from NetworkSupply class

rest-java/src/test/java/org/hiero/mirror/restjava/mapper/NetworkSupplyMapperTest.java
```diff
@@ -0,0 +1,193 @@
...
+    @Test
+    void convertToCurrencyFormatBothLarge() {
+        // given
+        final var valueInTinyCoins = "5000000000000000000"; // 50 billion HBAR
+
+        // when
+        final var result = mapper.convertToCurrencyFormat(valueInTinyCoins, "BOTH");
+
+        // then
+        assertThat(result).isEqualTo("50000000000.00000000");
```
steven-sheehy: These all should've been written as parameterized tests.

rest-java/src/main/java/org/hiero/mirror/restjava/repository/NetworkSupplyRepositoryCustomImpl.java
```diff
@@ -0,0 +1,137 @@
...
+                .from(accountBalances)
+                .fetchOne();
+
+        return toNetworkSupply(Objects.requireNonNull(result), unreleasedSupplyField, consensusTimestampField);
+    }
+
+    private Condition buildUnreleasedSupplyAccountCondition(Field<Long> accountField, long shard, long realm) {
+        Condition condition = DSL.noCondition();
+
+        for (var range : networkProperties.getUnreleasedSupplyAccounts()) {
```
steven-sheehy: Add a helper method to `NetworkProperties` named `List<Long> getUnreleasedSupplyAccountIds()` that flattens the accounts to a single list and uses CommonProperties to converted to EntityId. Use lombok lazy getter to cache it.

rest-java/src/main/java/org/hiero/mirror/restjava/repository/NetworkSupplyRepositoryCustomImpl.java
```diff
@@ -0,0 +1,137 @@
...
+                condition = condition.or(accountField.eq(from));
+            } else {
+                condition = condition.or(accountField.between(from, to));
+            }
+        }
+
+        return condition;
+    }
+
+    private NetworkSupply toNetworkSupply(
```
steven-sheehy: Use repository projection instead.

rest-java/src/main/java/org/hiero/mirror/restjava/repository/NetworkSupplyRepositoryCustomImpl.java
```diff
@@ -0,0 +1,137 @@
...
+        return toNetworkSupply(Objects.requireNonNull(result), unreleasedSupplyField, consensusTimestampField);
+    }
+
+    private NetworkSupply getSupplyFromAccountBalance(Bound timestamp) {
+        var minTimestamp = timestamp.getAdjustedLowerRangeValue();
+        var maxTimestamp = timestamp.adjustUpperBound();
+
+        // Optimize the query by limiting to at most two most recent partitions
+        var optimalLowerBound = getFirstDayOfMonth(maxTimestamp, -1);
+        minTimestamp = Math.max(minTimestamp, optimalLowerBound);
```
steven-sheehy: Move non-SQL logic to service layer.

rest-java/src/test/java/org/hiero/mirror/restjava/repository/NetworkSupplyRepositoryTest.java
```diff
@@ -0,0 +1,209 @@
...
+final class NetworkSupplyRepositoryTest extends RestJavaIntegrationTest {
+
+    private final NetworkSupplyRepositoryCustom networkSupplyRepository;
+
+    @Test
+    void getSupplyFromEntity() {
+        // given
+        var entity1Balance = 1_000_000_000L;
+        var entity2Balance = 2_000_000_000L;
+        var timestamp = domainBuilder.timestamp();
```
sdimitrov9: Consider making local variables final wherever applicable in the class.

rest-java/src/test/java/org/hiero/mirror/restjava/repository/NetworkSupplyRepositoryTest.java
```diff
@@ -0,0 +1,209 @@
...
+        var timestamp2 = timestamp1 + 1000;
+        var timestamp3 = timestamp1 + 2000;
+
+        // Create account balances at different timestamps
+        domainBuilder
+                .accountBalance()
+                .customize(ab -> ab.id(new org.hiero.mirror.common.domain.balance.AccountBalance.Id(
+                                timestamp1, EntityId.of(2L)))
+                        .balance(account2Balance1))
+                .persist();
```
sdimitrov9: Consider creating a helper method like peristAccountBalance() wrapping the logic for account balance persistance. This would reduce code repetition and improve readability.

rest-java/src/test/java/org/hiero/mirror/restjava/repository/NetworkSupplyRepositoryTest.java
```diff
@@ -0,0 +1,209 @@
...
+    @Test
+    void getSupplyFromAccountBalanceNoData() {
+        // given
+        var futureTimestamp = domainBuilder.timestamp() + 1_000_000_000L;
+        var bound = new Bound(
+                new TimestampParameter[] {
+                    new TimestampParameter(org.hiero.mirror.restjava.common.RangeOperator.LTE, futureTimestamp)
+                },
+                true,
+                org.hiero.mirror.restjava.common.Constants.TIMESTAMP,
```
sdimitrov9: Use static import 
`import static org.hiero.mirror.restjava.common.Constants.TIMESTAMP;` 
and omit the whole package here and from the other identical usages in the class.

rest-java/src/test/java/org/hiero/mirror/restjava/repository/NetworkSupplyRepositoryTest.java
```diff
@@ -0,0 +1,209 @@
...
+        var account2Balance2 = 600_000_000L;
+        var account100Balance = 1_500_000_000L;
+        var timestamp1 = domainBuilder.timestamp();
+        var timestamp2 = timestamp1 + 1000;
+        var timestamp3 = timestamp1 + 2000;
+
+        // Create account balances at different timestamps
+        domainBuilder
+                .accountBalance()
+                .customize(ab -> ab.id(new org.hiero.mirror.common.domain.balance.AccountBalance.Id(
```
sdimitrov9: Omit package name an apply import
 `import org.hiero.mirror.common.domain.balance.AccountBalance;`

rest-java/src/test/java/org/hiero/mirror/restjava/controller/NetworkControllerTest.java
```diff
@@ -524,4 +524,214 @@ void notFound() {
...
+            // Account 0.0.250 (range 200-349): 100 trillion tinybars
+            domainBuilder
+                    .entity()
+                    .customize(e -> e.id(EntityId.of(shard, realm, 250L).getId())
+                            .balance(100000000000000000L)
+                            .balanceTimestamp(timestamp + 1))
+                    .persist();
+
+            // when
+            var response =
```
sdimitrov9: Make local variables final wherever applicable.

rest-java/src/test/java/org/hiero/mirror/restjava/controller/NetworkControllerTest.java
```diff
@@ -524,4 +524,214 @@ void notFound() {
...
+                    restClient.get().uri("").retrieve().body(org.hiero.mirror.rest.model.NetworkSupplyResponse.class);
```
sdimitrov9: Fully qualified name for NetworkSupplyResponse shouldn't be necessary here since there isn't another class NetworkSupplyResponse used in the class.

rest-java/src/main/java/org/hiero/mirror/restjava/repository/EntityRepository.java
```diff
@@ -14,4 +16,33 @@ public interface EntityRepository extends CrudRepository<Entity, Long> {
...
+            and account_id in (:unreleasedSupplyAccounts)
+          order by account_id asc, consensus_timestamp desc
+        )
+        select coalesce(sum(balance), 0) as unreleasedSupply,
+               coalesce(max(consensus_timestamp), 0) as consensusTimestamp
+        from account_balances
+        """,
+            nativeQuery = true)
+    NetworkSupplyProjection getSupplyHistory(
+            List<Long> unreleasedSupplyAccounts, long lowerTimestamp, long upperTimestamp);
```
steven-sheehy: This doesn't query the Entity table so should go into an `AccountBalanceRepository`.

rest-java/src/main/java/org/hiero/mirror/restjava/config/NetworkProperties.java
```diff
@@ -0,0 +1,50 @@
...
+            new AccountRange(99, 100),
+            new AccountRange(200, 349),
+            new AccountRange(400, 750));
+
+    /**
+     * Get the unreleased supply account IDs as a list of entity IDs.
+     *
+     * @return List of EntityId objects representing unreleased supply accounts
+     */
+    public List<Long> getUnreleasedSupplyAccountIds() {
```
steven-sheehy: Optimize this to cache it.
```suggestion
    @Getter(lazy = true)
    private final List<Long> unreleasedSupplyAccountIds = createUnreleasedSupplyAccountIds();
    
    public List<Long> createUnreleasedSupplyAccountIds() {
```

rest-java/src/main/java/org/hiero/mirror/restjava/config/NetworkProperties.java
```diff
@@ -0,0 +1,50 @@
...
+        final var commonProperties = CommonProperties.getInstance();
+        final var shard = commonProperties.getShard();
+        final var realm = commonProperties.getRealm();
+        final var accountIds = new ArrayList<Long>();
```
steven-sheehy: Just in case there's a mis-config, change to Collection and use TreeSet.

```suggestion
        final var accountIds = new TreeSet<Long>();
```

rest-java/src/main/java/org/hiero/mirror/restjava/config/NetworkProperties.java
```diff
@@ -0,0 +1,50 @@
...
+        for (final var range : unreleasedSupplyAccounts) {
+            for (long num = range.from(); num <= range.to(); num++) {
+                accountIds.add(EntityId.of(shard, realm, num).getId());
+            }
+        }
+
+        return accountIds;
+    }
+
+    public record AccountRange(long from, long to) {}
```
steven-sheehy: We need to verify `@Min(1)` and from >= to in case these are customized.

rest-java/src/main/java/org/hiero/mirror/restjava/config/NetworkProperties.java
```diff
@@ -0,0 +1,50 @@
...
+import org.hiero.mirror.common.domain.entity.EntityId;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.validation.annotation.Validated;
+
+@Data
+@Validated
+@ConfigurationProperties("hiero.mirror.rest-java.network")
+public class NetworkProperties {
+
+    @NotNull
```
steven-sheehy: IN clause will fail if empty list.

```suggestion
    @NotEmpty
```

rest-java/src/main/java/org/hiero/mirror/restjava/controller/NetworkController.java
```diff
@@ -59,18 +67,47 @@ NetworkStakeResponse getNetworkStake() {
...
+
     private Bound timestampBound(TimestampParameter[] timestamp) {
+        return createBound(timestamp, false, FileData.FILE_DATA.CONSENSUS_TIMESTAMP);
+    }
+
+    private Bound supplyTimestampBound(TimestampParameter[] timestamp) {
+        return createBound(timestamp, true, AccountBalance.ACCOUNT_BALANCE.CONSENSUS_TIMESTAMP);
+    }
+
+    private Bound createBound(
```
steven-sheehy: Need to merge from main and use the new `Bound.of()` method. Can drop these 3 methods and just call Bound.of() inline.

rest-java/src/main/java/org/hiero/mirror/restjava/config/NetworkProperties.java
steven-sheehy: Missing docs still

rest-java/src/test/java/org/hiero/mirror/restjava/controller/NetworkControllerTest.java
```diff
@@ -524,4 +527,328 @@ void notFound() {
...
+
+        private void createUnreleasedSupplyAccounts(Long timestamp) {
+            final var accountRanges = List.of(
+                    new long[] {2, 2},
+                    new long[] {42, 42},
+                    new long[] {44, 71},
+                    new long[] {73, 87},
+                    new long[] {99, 100},
+                    new long[] {200, 349},
+                    new long[] {400, 750});
```
steven-sheehy: Why are we repeating the properties? Should use the `NetworkProperties.getUnreleasedSupplyAccountIds` method to get the flattened ids.

rest-java/src/test/java/org/hiero/mirror/restjava/controller/NetworkControllerTest.java
```diff
@@ -524,4 +527,328 @@ void notFound() {
...
+            createUnreleasedSupplyAccounts(null);
+
+            // when
+            final var response = restClient.get().uri("").retrieve().body(NetworkSupplyResponse.class);
+
+            // then - 548 accounts * 1M hbars = 548M hbars unreleased
+            // released = 50B - 548M = 49,452,000,000 hbars
+            assertThat(response).isNotNull();
+            assertThat(response.getTotalSupply()).isEqualTo("5000000000000000000");
+            assertThat(response.getReleasedSupply()).isEqualTo("4945200000000000000");
```
steven-sheehy: We should verify that `totalsupply - releasedsupply == networkProperties.getUnreleasedSupplyAccountIds().size() * BALANCE_PER_ACCOUNT`.

rest-java/src/test/java/org/hiero/mirror/restjava/controller/NetworkControllerTest.java
```diff
@@ -524,4 +527,328 @@ void notFound() {
...
+            final var response = restClient
+                    .get()
+                    .uri("?timestamp=" + commonMapper.mapTimestamp(timestamp))
+                    .retrieve()
+                    .body(NetworkSupplyResponse.class);
+
+            // then
+            assertThat(response).isNotNull();
+            assertThat(response.getTotalSupply()).isEqualTo("5000000000000000000");
+            assertThat(response.getReleasedSupply()).isEqualTo("4945200000000000000");
```
steven-sheehy, line 616: Probably we should adjust BALANCE_PER_ACCOUNT for historical to be slightly different so we can verify it used historical path.

rest-java/src/test/java/org/hiero/mirror/restjava/controller/NetworkControllerTest.java
```diff
@@ -524,4 +527,328 @@ void notFound() {
...
+
+            final var formattedParams = MessageFormat.format(
+                    parameters,
+                    commonMapper.mapTimestamp(timestamp0),
+                    commonMapper.mapTimestamp(timestamp1),
+                    commonMapper.mapTimestamp(timestamp2));
+
+            final var expected1Timestamp = commonMapper.mapTimestamp(timestamp1);
+            final var expected2Timestamp = commonMapper.mapTimestamp(timestamp2);
+            final var expectedTimestamps = List.of(expected1Timestamp, expected1Timestamp, expected2Timestamp);
```
steven-sheehy: Why is expected1Timestamp here twice?

rest-java/src/test/java/org/hiero/mirror/restjava/controller/NetworkControllerTest.java
```diff
@@ -524,4 +527,328 @@ void notFound() {
...
+            final var response =
+                    restClient.get().uri("?q=totalcoins").retrieve().toEntity(String.class);
+
+            // then - verify Content-Type is text/plain
+            assertThat(response.getHeaders().getContentType()).isNotNull();
+            assertThat(response.getHeaders().getContentType().toString()).startsWith("text/plain");
+        }
+
+        @Test
+        void contentTypeForQueryParameterCirculating() {
```
steven-sheehy: This test seems redundant with `withQueryParameterCirculating` and `withQueryParameterTotalCoins`. Why not use `toEntity()` on the others and check the content type there?

rest-java/src/main/java/org/hiero/mirror/restjava/dto/NetworkSupplyProjection.java
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package org.hiero.mirror.restjava.dto;
+
+public interface NetworkSupplyProjection {
```
steven-sheehy: Is it not possible to return NetworkSupply from the repository? Maybe if you convert to a POJO?

rest-java/src/main/java/org/hiero/mirror/restjava/service/NetworkServiceImpl.java
```diff
@@ -4,20 +4,61 @@
...
+        if (projection.getConsensusTimestamp() == 0L) {
+            throw new EntityNotFoundException("Network supply not found");
+        }
+
+        final var releasedSupply = NetworkSupply.TOTAL_SUPPLY - projection.getUnreleasedSupply();
+        return new NetworkSupply(releasedSupply, projection.getConsensusTimestamp(), NetworkSupply.TOTAL_SUPPLY);
+    }
+
+    private long getFirstDayOfMonth(long timestamp, int monthOffset) {
+        final var instant = Instant.ofEpochSecond(timestamp / 1_000_000_000);
```
steven-sheehy:
```suggestion
        final var instant = Instant.ofEpochSecond(0, timestamp);
```

rest-java/src/main/java/org/hiero/mirror/restjava/service/NetworkServiceImpl.java
```diff
@@ -4,20 +4,61 @@
...
+        final var dateTime = instant.atZone(ZoneOffset.UTC);
+        final var firstDay = dateTime.plusMonths(monthOffset).withDayOfMonth(1);
+
+        return firstDay.toLocalDate().atStartOfDay(ZoneOffset.UTC).toEpochSecond() * 1_000_000_000L;
```
steven-sheehy:
```suggestion
        return firstDay.toLocalDate().atStartOfDay(ZoneOffset.UTC).toEpochSecond() * DomainUtils.NANOS_PER_SECOND;
```

rest-java/src/main/java/org/hiero/mirror/restjava/converter/SupplyQueryConverter.java
```diff
@@ -0,0 +1,18 @@
...
+import org.springframework.boot.context.properties.ConfigurationPropertiesBinding;
+import org.springframework.core.convert.converter.Converter;
+
+@Named
+@ConfigurationPropertiesBinding
+public class SupplyQueryConverter implements Converter<String, SupplyQuery> {
+
+    @Override
+    public SupplyQuery convert(String source) {
+        return SupplyQuery.valueOf(source.toUpperCase());
```
steven-sheehy: Remove this in favor of adding a `valueOf` method to SupplyQuery itself

rest-java/src/main/java/org/hiero/mirror/restjava/config/NetworkProperties.java
```diff
@@ -0,0 +1,56 @@
...
+            new AccountRange(2, 2),
+            new AccountRange(42, 42),
+            new AccountRange(44, 71),
+            new AccountRange(73, 87),
+            new AccountRange(99, 100),
+            new AccountRange(200, 349),
+            new AccountRange(400, 750));
+
+    @Getter(lazy = true)
+    private final TreeSet<Long> unreleasedSupplyAccountIds = createUnreleasedSupplyAccountIds();
```
steven-sheehy: Code smell here. Use `Set` or `Collection`. Whichever is highest level interface that works. Same for `createUnreleasedSupplyAccountIds()`.

```suggestion
    private final Set<Long> unreleasedSupplyAccountIds = createUnreleasedSupplyAccountIds();
```

steven-sheehy (dismissed):
LGTM. Just one nitpick.

steven-sheehy:
Please merge from main to verify chart and acceptance workflows pass.

rest-java/src/main/java/org/hiero/mirror/restjava/common/SupplyQuery.java
```diff
@@ -0,0 +1,25 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package org.hiero.mirror.restjava.common;
+
+public enum SupplyQuery {
```
sdimitrov9: The name SupplyQuery looks semantically incorrect to me, since this enum doesn't represent a query, which usually contains a set of instructions (ordering, filtering, etc.). An enum should be named to what its values represent so I would rename it to SupplyType.

rest-java/src/main/java/org/hiero/mirror/restjava/controller/NetworkController.java
```diff
@@ -57,4 +64,25 @@ NetworkStakeResponse getNetworkStake() {
         final var networkStake = networkService.getLatestNetworkStake();
         return networkStakeMapper.map(networkStake);
     }
+
+    @GetMapping("/supply")
+    ResponseEntity<?> getSupply(
+            @RequestParam(required = false) @Size(max = 2) TimestampParameter[] timestamp,
+            @RequestParam(required = false) String q) {
```
sdimitrov9: Since 'q' stands for query and this parameter is more of a supply type I would rename it to supplyType. To preserve compatibility with the js api parameters we can add name = "q" within the RequestParam: 
`@RequestParam(name = "q", required = false)`

sdimitrov9: We can change the query param type from String to the enum type:
`@RequestParam(name = "q", required = false) SupplyType supplyType)` and avoid this conversion 
`final var supplyQuery = SupplyQuery.of(q);`

To accept input in both upper and lower case we should add SupplyTypeConverter 
```
@Named
@ConfigurationPropertiesBinding
public class SupplyTypeConverter implements Converter<String, SupplyType> {
    @Override
    public SupplyType convert(String order) {
        return SupplyType.of(order);
    }
}
```

rest-java/src/test/java/org/hiero/mirror/restjava/controller/NetworkControllerTest.java
```diff
@@ -524,4 +529,299 @@ void notFound() {
...
+                    .retrieve()
+                    .body(NetworkSupplyResponse.class);
+
+            // then - should use latest balances within range (timestamp2)
+            assertThat(response).isNotNull();
+            assertThat(response.getTimestamp()).isEqualTo("1700000000.000000000");
+        }
+
+        private void createCustomBalance(long accountNum, long balance, long timestamp) {
+            final var accountId = EntityId.of(0, 0, accountNum);
```
sdimitrov9: This is going to fail tests in the Non-Zero Realm workflow. This suggestion should do the work:
```suggestion
            final var entityIdParameter = EntityIdParameter.valueOf(String.valueOf(accountNum));
            final var accountId = EntityId.of(entityIdParameter.shard(), entityIdParameter.realm(), accountNum);
```

rest-java/src/test/java/org/hiero/mirror/restjava/controller/NetworkControllerTest.java
```diff
@@ -524,4 +529,299 @@ void notFound() {
...
+        }
+
+        @Override
+        protected RequestHeadersSpec<?> defaultRequest(RequestHeadersUriSpec<?> uriSpec) {
+            createUnreleasedSupplyAccounts(null);
+            return uriSpec.uri("");
+        }
+
+        @Test
+        void success() {
```
sdimitrov9: Rename to something more meaningful like getNetworkSupplyNoQueryParams()

sdimitrov9:
LGTM

PR #12489, opened 2025-11-28
Avoid cache stampede on latest record file cache expiry

### LINKED ISSUE #12477: Optimize record file query count in web3
### Problem

As a result of the performed testing in https://github.com/hiero-ledger/hiero-mirror-node/issues/12164 the following observations were made:
1. Among all record files the latest one is read the most frequently. It is always the same for all requests in the 2-second window that is valid. Currently it is cached for 500ms and this can be increased up to 2s or at least 1s as we don't need it refreshed that frequently. 
2. There are way too many DB requests that are made concurrently to read the latest record file before it gets cached. In numbers: with only 20% of the real world traffic over a 20-second period there are over 850 actual DB queries (not repository calls) for the latest record file. Even in 3-4 ms period about 20 actual DB calls are made (the average time for the repository call completion is about 80ms). This takes 20 DB connections from the pool that can be used for other queries that are really needed.

### Solution

Find a way to optimize the record file DB queries. Ideally, only 1 thread should make the call for the latest record file and all others should just wait for it and read it. Then after 2 seconds when the file expires, it needs to be read again.
Another suggestion is to make the repository call lazily and skip the record file reading wherever possible.

### Alternatives

_No response_

sdimitrov9:
**Description**:
This PR should reduce the impact from the cache stampede that happens on latest record file cache expiry resulting in hundreds of thousands of unnecessary db calls on daily basis. The latest record file now is queried when is actually needed instead of early in the ContractCallService.callContract() .
 
**Related issue(s)**:

Fixes #12477 

**Checklist**

### DIFF
--- a/web3/src/main/java/org/hiero/mirror/web3/common/ContractCallContext.java
+++ b/web3/src/main/java/org/hiero/mirror/web3/common/ContractCallContext.java
@@ -48,13 +48,6 @@ public class ContractCallContext {
     @Setter
     private CallServiceParameters callServiceParameters;
 
-    /**
-     * Record file which stores the block timestamp and other historical block details used for filtering of historical
-     * data.
-     */
-    @Setter
-    private RecordFile recordFile;
-
     @Setter
     private EntityNumber entityNumber;
 
@@ -69,6 +62,9 @@ public class ContractCallContext {
 
     @Setter
     private long gasRequirement;
+
+    @Setter
+    private Supplier<RecordFile> blockSupplier = () -> null;
 
     private ContractCallContext() {}
 
@@ -117,10 +113,7 @@ public class ContractCallContext {
     }
 
     public boolean useHistorical() {
-        if (callServiceParameters != null) {
-            return callServiceParameters.getBlock() != BlockType.LATEST;
-        }
-        return recordFile != null; // Remove recordFile comparison after mono code deletion
+        return callServiceParameters != null ? callServiceParameters.getBlock() != BlockType.LATEST : false;
     }
 
     /**
@@ -135,7 +128,7 @@ public class ContractCallContext {
     }
 
     private Optional<Long> getTimestampOrDefaultFromRecordFile() {
-        return timestamp.or(() -> Optional.ofNullable(recordFile).map(RecordFile::getConsensusEnd));
+        return timestamp.or(() -> Optional.ofNullable(getRecordFile()).map(RecordFile::getConsensusEnd));
     }
 
     public Map<Object, Object> getReadCacheState(final int stateId) {
@@ -146,15 +139,7 @@ public class ContractCallContext {
         return writeCache.computeIfAbsent(stateId, k -> new HashMap<>());
     }
 
-    /**
-     *  Get record file from context, if not present set it through a supplier.
-     */
-    public RecordFile resolveRecordFile(Supplier<RecordFile> recordFileSupplier) {
-        var recordFile = getRecordFile();
-        if (recordFile == null) {
-            recordFile = recordFileSupplier.get();
-            setRecordFile(recordFile);
-        }
-        return recordFile;
+    public RecordFile getRecordFile() {
+        return blockSupplier.get();
     }
 }
--- a/web3/src/main/java/org/hiero/mirror/web3/service/ContractCallService.java
+++ b/web3/src/main/java/org/hiero/mirror/web3/service/ContractCallService.java
@@ -23,6 +23,7 @@ import org.hiero.mirror.web3.exception.MirrorEvmTransactionException;
 import org.hiero.mirror.web3.service.model.CallServiceParameters;
 import org.hiero.mirror.web3.throttle.ThrottleManager;
 import org.hiero.mirror.web3.throttle.ThrottleProperties;
+import org.hiero.mirror.web3.utils.Suppliers;
 import org.hiero.mirror.web3.viewmodel.BlockType;
 
 @Named
@@ -94,10 +95,10 @@ public abstract class ContractCallService {
             CallServiceParameters params, ContractCallContext ctx) throws MirrorEvmTransactionException {
         ctx.setCallServiceParameters(params);
 
-        if (params.getBlock() != BlockType.LATEST) {
-            ctx.setRecordFile(recordFileService
+        if (params.isModularized() || params.getBlock() != BlockType.LATEST) {
+            ctx.setBlockSupplier(Suppliers.memoize(() -> recordFileService
                     .findByBlockType(params.getBlock())
-                    .orElseThrow(BlockNumberNotFoundException::new));
+                    .orElseThrow(BlockNumberNotFoundException::new)));
         }
 
         return doProcessCall(params, params.getGas(), false);
--- a/web3/src/main/java/org/hiero/mirror/web3/state/singleton/BlockInfoSingleton.java
+++ b/web3/src/main/java/org/hiero/mirror/web3/state/singleton/BlockInfoSingleton.java
@@ -9,10 +9,7 @@ import com.hedera.pbj.runtime.io.buffer.Bytes;
 import jakarta.inject.Named;
 import lombok.RequiredArgsConstructor;
 import org.hiero.mirror.web3.common.ContractCallContext;
-import org.hiero.mirror.web3.exception.BlockNumberNotFoundException;
-import org.hiero.mirror.web3.service.RecordFileService;
 import org.hiero.mirror.web3.state.Utils;
-import org.hiero.mirror.web3.viewmodel.BlockType;
 
 @Named
 @RequiredArgsConstructor
@@ -23,14 +20,9 @@ public class BlockInfoSingleton implements SingletonState<BlockInfo> {
         return BLOCKS_STATE_ID;
     }
 
-    private final RecordFileService recordFileService;
-
     @Override
     public BlockInfo get() {
-        final var recordFile = ContractCallContext.get().resolveRecordFile(() -> recordFileService
-                .findByBlockType(BlockType.LATEST)
-                .orElseThrow(BlockNumberNotFoundException::new));
-
+        final var recordFile = ContractCallContext.get().getRecordFile();
         final var startTimestamp = Utils.convertToTimestamp(recordFile.getConsensusStart());
         final var endTimestamp = Utils.convertToTimestamp(recordFile.getConsensusEnd());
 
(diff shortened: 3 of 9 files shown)


web3/src/main/java/org/hiero/mirror/web3/service/RecordFileServiceImpl.java
```diff
@@ -5,32 +5,66 @@
...
 
 @Service
+@CustomLog
 @RequiredArgsConstructor
 public class RecordFileServiceImpl implements RecordFileService {
 
     private final RecordFileRepository recordFileRepository;
+    private final Web3Properties web3Properties;
+
+    private final AtomicReference<RecordFile> latestRecord = new AtomicReference<>();
```
bilyana-gospodinova: We don't really use the atomic benefits so the two fields can be just regular `RecordFile` and `long`.

web3/src/main/java/org/hiero/mirror/web3/Web3Properties.java
```diff
@@ -20,4 +20,9 @@ public class Web3Properties {
 
     @DurationMin(seconds = 1L)
     private Duration requestTimeout = Duration.ofSeconds(10L);
+
+    @DurationMin(millis = 100L)
+    private Duration frequency = Duration.ofMillis(500L);
```
bilyana-gospodinova: Need to be documented in `configuration.md`.

sdimitrov9: Done

docs/configuration.md
```diff
@@ -742,9 +742,11 @@ value, it is recommended to only populate overridden properties in the custom `a
...
 | `hiero.mirror.web3.evm.modularizedTrafficPercent`            | 1.0                                                | Indicates the probability of processing a transaction through the transaction execution service, instead of the mirrorEvmTxProcessor.                                                            |
 | `hiero.mirror.web3.evm.entityNumBuffer`                      | 1000                                               | A buffer added to the max known entity number to prevent contract call simulations from generating IDs that may conflict with entities persisted by the importer module.                         |
+| `hiero.mirror.web3.frequency`                                | 500                                                | The frequency of execution of the scheduler method saving the latest record file in memory                                                                                                       |
```
bilyana-gospodinova: I think we need a more specific name of the property. It is not clear which frequency exactly it tracks. The description should mention the metric (e.g. milliseconds).

web3/src/main/java/org/hiero/mirror/web3/service/RecordFileServiceImpl.java
sdimitrov9: Obsolete, latest solution does not make changes to the class.

docs/configuration.md
sdimitrov9: Obsolete, latest solution does not make changes to the file.

web3/src/main/java/org/hiero/mirror/web3/service/ContractCallService.java
```diff
@@ -172,4 +178,26 @@ protected final void updateGasLimitMetric(final CallServiceParameters parameters
                 .and("type", parameters.getCallType().toString());
         gasLimitCounter.withTags(tags).increment(parameters.getGas());
     }
+
+    private RecordFile createSyntheticRecordFile() {
```
bilyana-gospodinova: This method is not called anywhere. How does it work?

web3/src/main/java/org/hiero/mirror/web3/state/singleton/RunningHashesSingleton.java
```diff
@@ -7,19 +7,32 @@
...
     }
 
     @Override
     public RunningHashes get() {
-        var recordFile = ContractCallContext.get().getRecordFile();
+        final var context = ContractCallContext.get();
+        var recordFile = context.getRecordFile();
+        if (recordFile == null) {
+            recordFile =
+                    recordFileService.findByBlockType(BlockType.LATEST).orElseThrow(BlockNumberNotFoundException::new);
```
bilyana-gospodinova: What if we are in a historical call?

sdimitrov9: In that case the record file is already set earlier in the context in the callContract of the ContractCallService:

        if (params.getBlock() != BlockType.LATEST) {
                    .findByBlockType(params.getBlock())
                    .orElseThrow(BlockNumberNotFoundException::new));
        }

web3/src/main/java/org/hiero/mirror/web3/service/ContractCallService.java
sdimitrov9: It is not needed anymore, we just call the query for the latest record file when it is actually needed. Will remove it from the code.

web3/src/main/java/org/hiero/mirror/web3/state/singleton/RunningHashesSingleton.java
```diff
@@ -7,19 +7,32 @@
...
     @Override
     public Integer getId() {
         return RUNNING_HASHES_STATE_ID;
     }
 
     @Override
     public RunningHashes get() {
-        var recordFile = ContractCallContext.get().getRecordFile();
+        final var context = ContractCallContext.get();
+        var recordFile = context.getRecordFile();
```
bilyana-gospodinova: These new lines are common for 3 files - `MirrorBlockHashOperation`, `BlockInfoSingleton` and `RunningHashesSingleton`. They can be extracted into a common class with inheritance or composition.

web3/src/main/java/org/hiero/mirror/web3/service/ContractCallService.java
```diff
@@ -94,7 +94,7 @@ protected final HederaEvmTransactionProcessingResult callContract(
             CallServiceParameters params, ContractCallContext ctx) throws MirrorEvmTransactionException {
         ctx.setCallServiceParameters(params);
 
-        if (params.isModularized() || params.getBlock() != BlockType.LATEST) {
+        if (params.getBlock() != BlockType.LATEST) {
```
steven-sheehy: This change isn't really necessary for this PR and causes conflicts in the other PR. Probably better to remove and leave it for the remove mono PR to change.

web3/src/main/java/org/hiero/mirror/web3/state/singleton/BlockInfoSingleton.java
```diff
@@ -20,11 +23,16 @@ public Integer getId() {
...
     }
 
+    private final RecordFileService recordFileService;
+
     @Override
     public BlockInfo get() {
-        var recordFile = ContractCallContext.get().getRecordFile();
-        var startTimestamp = Utils.convertToTimestamp(recordFile.getConsensusStart());
-        var endTimestamp = Utils.convertToTimestamp(recordFile.getConsensusEnd());
+        final var recordFile = ContractCallContext.get().resolveRecordFile(() -> recordFileService
```
steven-sheehy: It's redundant to have to supply the supplier every time you want to get the record file. We should set the supplier once per request in `ContractCallService.callContract()` via `context.setBlockSupplier(() -> ...)`. Then revert this change and others like it.

web3/src/main/java/org/hiero/mirror/web3/state/singleton/BlockInfoSingleton.java
```diff
@@ -20,11 +23,16 @@ public Integer getId() {
...
+                .findByBlockType(BlockType.LATEST)
```
steven-sheehy: Why are we always getting LATEST instead of the requested one in the request here and in other calls to resolveRecordFile()?

web3/src/main/java/org/hiero/mirror/web3/common/ContractCallContext.java
```diff
@@ -144,4 +145,16 @@ public Map<Object, Object> getReadCacheState(final int stateId) {
...
+     *  Get record file from context, if not present set it through a supplier.
+     */
+    public RecordFile resolveRecordFile(Supplier<RecordFile> recordFileSupplier) {
+        var recordFile = getRecordFile();
+        if (recordFile == null) {
+            recordFile = recordFileSupplier.get();
+            setRecordFile(recordFile);
+        }
+        return recordFile;
+    }
```
steven-sheehy: This doesn't do atomic sets. We also don't need to store the recordFile field anymore, just the supplier.

```suggestion
    private Supplier<RecordFile> blockSupplier;
    
    public RecordFile getRecordFile() {
        return blockSupplier.get();
    }
```

Then

```java
    context.setBlockSupplier(Suppliers.memoize(() -> recordFileService.findByBlockType(params.getBlock()).orElseThrow(BlockNumberNotFoundException::new)))
```

web3/src/main/java/org/hiero/mirror/web3/state/singleton/BlockInfoSingleton.java
sdimitrov9: The idea was that if the requested file was not the LATEST, the record file in the context would have already been set in the callContract method, because we would have met the condition in the if statement:
```
protected final HederaEvmTransactionProcessingResult callContract(
            CallServiceParameters params, ContractCallContext ctx) throws MirrorEvmTransactionException {
        ctx.setCallServiceParameters(params);

        if (params.getBlock() != BlockType.LATEST) {
                    .findByBlockType(params.getBlock())
                    .orElseThrow(BlockNumberNotFoundException::new));
        }

        return doProcessCall(params, params.getGas(), false);
    }
```

sdimitrov9:
One possible optimization,  @steven-sheehy what do you think about removing the findLatestIndex call along with the if statement that follows it and instead return the latest record file when recordFileRepository.findByIndex(block.number()) fails to find result? Currently we always call findLatestIndex() during historical call and probably in almost all cases the block.number is valid one.

```
@Override
    public Optional<RecordFile> findByBlockType(BlockType block) {
            return recordFileRepository.findEarliest();
        } else if (block == BlockType.LATEST) {
            return recordFileRepository.findLatest();
        }

        long latestBlock = recordFileRepository
                .findLatestIndex()
                .orElseThrow(() -> new BlockNumberOutOfRangeException(UNKNOWN_BLOCK_NUMBER));

        if (block.number() > latestBlock) {
            throw new BlockNumberOutOfRangeException(UNKNOWN_BLOCK_NUMBER);
        }
        return recordFileRepository.findByIndex(block.number());
    }
```

steven-sheehy:
> One possible optimization, @steven-sheehy what do you think about removing the findLatestIndex call along with the if statement that follows it and instead return the latest record file when recordFileRepository.findByIndex(block.number()) fails to find result? Currently we always call findLatestIndex() during historical call and probably in almost all cases the block.number is valid one.

Yes, I think we should remove call to findLatestIndex(). If we can find the block number then that already means it's less than or equal to latest. Maybe we can do as a follow up.

IvanKavaldzhiev (approved):
LGTM

sdimitrov9:
> > One possible optimization, @steven-sheehy what do you think about removing the findLatestIndex call along with the if statement that follows it and instead return the latest record file when recordFileRepository.findByIndex(block.number()) fails to find result? Currently we always call findLatestIndex() during historical call and probably in almost all cases the block.number is valid one.
> 
> Yes, I think we should remove call to findLatestIndex(). If we can find the block number then that already means it's less than or equal to latest. Maybe we can do as a follow up.

Follow up ticket is created: https://github.com/hiero-ledger/hiero-mirror-node/issues/12627

PR #12509, opened 2025-12-01
Add traffic replay k6 web3 tests

### LINKED ISSUE #12164: Performance test web3 API with real world traffic
### Problem

Our web3 performance tests may not be indicative of real world traffic.

### Solution

* Use the traffic replay tool to send mainnet traffic to staging while simultaneously scaling the traffic for an extended period of time
* Identify any slowdowns and create tickets
* Consider storing simulated requests somewhere outside of this repo in a compressed file and creating a k6 test that loads and executes them

### Alternatives

_No response_

bilyana-gospodinova:
**Description**:
This PR adds k6 tests with real world traffic. The input file is expected to have multiple rows with a single uncompressed request body per row.
The tests were run with an input file containing 1000 requests. The size of that file is 16 KB in its compressed form.

Fixes #12164 

**Checklist**


tools/k6/src/web3/test/common.js
```diff
@@ -206,4 +206,49 @@ function ContractCallTestScenarioBuilder() {
...
+  build() {
+    const that = this;
+
+    let combinedOptions;
+    for (let i = 0; i < that._parsedRequests.length; i++) {
+      const data = that._parsedRequests[i];
+      // Add the index in the scenario name as there might be requests with the same timestamp
+      const scenarioName = `${that._name}_${i}_ts_${data.scenarioKey}`;
+      const options = utils.getOptionsWithScenario(scenarioName, null, {url: that._url, payload: data.payload});
+      options.scenarios[scenarioName].duration = __ENV.TRAFFIC_REPLAY_DURATION ? __ENV.TRAFFIC_REPLAY_DURATION : '5s';
```
bilyana-gospodinova: The `TRAFFIC_REPLAY_DURATION` env variable was introduced because the requests are a lot and it would make sense to run them for a shorter amount of time compared to the other existing tests to reduce the overall execution time.

tools/k6/src/web3/test/traffic-replay/trafficReplay.js
```diff
@@ -0,0 +1,49 @@
+// SPDX-License-Identifier: Apache-2.0
+
+// --- CONFIGURATION ---
+const FILE_PATH = './input.txt';
```
bilyana-gospodinova: Currently this file is not uploaded. It is 16 KB in size (uncompressed). I think we can add it to this repo. It is not that big to complicate the code even more by downloading it from another location or uncompressing it.

tools/k6/src/web3/test/traffic-replay/trafficReplay.js
```diff
@@ -0,0 +1,48 @@
...
+
+const parsedRequests = new SharedArray('parsed payloads', function () {
+  const fileContent = open(FILE_PATH);
+  const requests = [];
+
+  // We use a mutable index to ensure the regex finds the next match correctly
+  let match;
+  while ((match = DATA_REGEX.exec(fileContent)) !== null) {
+    // match[1] is the nanosecond timestamp
+    const scenarioKey = match[1];
```
xin-hedera: `match[1]` is content length not the timestamp, given the following replay log format:

```
POST /api/v1/contracts/call HTTP/1.1
Host: 127.0.0.1:80
User-Agent: curl/8.8.0
Accept: */*
Content-Type: application/json
Content-Length: 406

{"block":"latest","data":"0x1f00ca74000000000000000000000000000000000000000000000000000000e8d4a5100000000000000000000000000000000000000000000000000000000000000000400000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000006f89a0000000000000000000000000000000000000000000000000000000000163b5a","to":"00000000000000000000000000000000002e7a5d"}
🐵🙈🙉
```

tools/k6/src/web3/test/traffic-replay/trafficReplay.js
```diff
@@ -0,0 +1,48 @@
...
+    // match[2] is the full JSON string
+    const rawJsonBody = match[2];
+
+    try {
+      requests.push({
+        scenarioKey: scenarioKey,
```
xin-hedera: I don't see the benefit of creating one scenario per request in the log. Instead, we can have just one scenario, and in each VU, we can maintain a local counter to iterate through the requests in the shared array.

If we make the change, we wouldn't need a separate replay duration.

tools/k6/src/web3/test/common.js
```diff
@@ -206,4 +206,49 @@ function ContractCallTestScenarioBuilder() {
   return this;
 }
 
-export {isNonErrorResponse, isValidListResponse, jsonPost, loadVuDataOrDefault, ContractCallTestScenarioBuilder};
+class TrafficReplayScenarioBuilder extends utils.BaseMultiScenarioBuilder {
```
xin-hedera: the scenario builder should to live in `trafficReplay.js` as it's not shared by any other test case.

tools/k6/src/web3/test/traffic-replay/trafficReplay.js
steven-sheehy: How big is it compressed?

bilyana-gospodinova: My bad - it is 16 KB compressed, 455 KB uncompressed.

tools/k6/src/web3/test/traffic-replay/trafficReplay.js
bilyana-gospodinova: No, the file is simplified and the format is:
`<timestamp> <request_body>`
where:
- `<timestamp>` is the timestamp at which the original request was made
- `<request_body>` is a json body of the request, uncompressed.

tools/k6/src/web3/test/traffic-replay/trafficReplay.js
bilyana-gospodinova: The benefit of having one scenario per request is that we can identify the slow requests and then test them and investigate the reason why they are slow.

tools/k6/src/web3/test/common.js
bilyana-gospodinova: Addressed

tools/k6/src/web3/test/traffic-replay/trafficReplay.js
xin-hedera: I see, it's converted.

Do we need the timestamp? There's also logic in the test case to handle duplicate timestamps.

What if the file is just an large json array of POST request body? For scenario name, it could be just the index of the request in the array.

steven-sheehy:
The attached report.md is for the REST API, not web3.

bilyana-gospodinova:
> The attached report.md is for the REST API, not web3.

I fixed the attachment.

tools/k6/src/web3/test/traffic-replay/trafficReplay.js
bilyana-gospodinova: Agreed. Comment addressed.

tools/k6/src/web3/test/traffic-replay/trafficReplay.js
steven-sheehy: That would be too slow to run as each would run for 5 min. We should just have one scenario added for this since we want to incorporate it into the regular release testing.

bilyana-gospodinova: I thought that too, so I added the `TRAFFIC_REPLAY_DURATION` variable to allow running for a shorter period of time just for this test case.

bilyana-gospodinova: Refactored the PR to have a single scenario that replays the traffic requests iteratively.

tools/k6/src/web3/test/traffic-replay/trafficReplay.js
bilyana-gospodinova: There is no way to unzip a file in a k6 test, so the unzipping needs to be done before test execution. The tools module does not have a build.gradle.kts file - otherwise I could unzip the file in a task. Is there some workflow that executes on a release that I can plug in a script to unzip the compressed file?

bilyana-gospodinova: I changed the test to read the traffic replay input file from another github repo. Tested with my personal repo (https://github.com/bilyana-gospodinova/traffic-replay-input). The file is downloaded only once for all VUs.

bilyana-gospodinova: Perhaps we can create another branch in this repo with an empty history and just the file that we need (similarly to the deploy branch). Then the url that we need to set will look like: 'https://raw.githubusercontent.com/hiero-ledger/hiero-mirror-node/traffic-replay/input.txt' if the branch is named 'traffic-replay'.

tools/k6/src/web3/test/index.js
```diff
@@ -150,6 +152,27 @@ if (__ENV.RUN_OPCODE_TESTS !== 'false' && __ENV.TRANSACTION_IDS) {
...
 }
 
+if (__ENV.RUN_TRAFFIC_REPLAY !== 'false' && __ENV.GITHUB_URL_TRAFFIC_REPLAY_FILE) {
+  Object.assign(tests, {
+    trafficReplay,
+  });
+}
+
+// This method is executed only once regarding of the VUs count.
+export function setup() {
```
xin-hedera, line 162: should move this to `trafficReplay.js` and use the sharedArray approach. Main reason is our test cases are structured in a way that we can

- run it with the whole suite when enabled, e.g., `k6 run web3/apis.js`
- run the test case itself, e.g., `k6 run web3/test/traffic-replay/trafficReplay.js`

tools/k6/src/web3/test/traffic-replay/trafficReplay.js
```diff
@@ -0,0 +1,73 @@
...
+    }
+  }
+  return requests;
+}
+
+function run(testParameters) {
+  if (!testParameters) {
+    // This test case must be run via apis.js in order to download the traffic input file from github only once
+    // instead of once for each VU. With a large amount of VUs, github rejects some of the request which
+    // causes the tests to fail and contaminates the results.
```
xin-hedera, line 38: can't we use `SharedArray` and add the downloading logic into its loading function so it only runs once?

tools/k6/src/web3/test/traffic-replay/trafficReplay.js
```diff
@@ -0,0 +1,73 @@
...
+  const dynamicParams = Object.assign({}, params, {
+    tags: {
+      name: `traffic_replay_index_${requestIndex}`,
+      scenario_key: scenarioKey,
+    },
+  });
+
+  const res = http.post(url, rawPayload, dynamicParams);
+  check(res, {
+    [`Traffic replay index ${requestIndex} processed in ${res.timings.duration}ms`]: (r) =>
```
xin-hedera: won't this clog the output because there are so many different checks?

tools/k6/src/web3/test/traffic-replay/trafficReplay.js
```diff
@@ -0,0 +1,73 @@
...
+  const requestIndex = __ITER % totalRequests;
+  const requestData = parsedRequests[requestIndex];
+
+  const rawPayload = requestData.payload;
+  const scenarioKey = requestData.scenarioKey;
+
+  const url = `${BASE_URL}/api/v1/contracts/call`;
+  const dynamicParams = Object.assign({}, params, {
+    tags: {
+      name: `traffic_replay_index_${requestIndex}`,
```
xin-hedera: don't think there's value of the request index tag

tools/k6/src/web3/test/traffic-replay/trafficReplay.js
```diff
@@ -0,0 +1,73 @@
...
+  let requestIndex = 0;
+  let match;
+
+  while ((match = DATA_REGEX.exec(fileContent)) !== null) {
+    const rawJsonBody = match[1];
+    try {
+      requests.push({
+        scenarioKey: String(requestIndex),
+        payload: rawJsonBody,
+      });
```
xin-hedera: The scenario key is just the index of the item in the array, redundant info.

```suggestion
      requests.push(rawJsonBody);
```

tools/k6/src/web3/test/traffic-replay/trafficReplay.js
```diff
@@ -0,0 +1,73 @@
...
+    return;
+  }
+  const parsedRequests = testParameters.trafficReplayRequests || [];
+  if (parsedRequests.length === 0) {
+    console.log(`No traffic replay requests found.`);
+    return;
+  }
+
+  const totalRequests = parsedRequests.length;
+  const requestIndex = __ITER % totalRequests;
```
xin-hedera, line 49: use `requestIndex` as the scenarioKey because it's equivalent.

tools/k6/src/web3/test/traffic-replay/trafficReplay.js
bilyana-gospodinova: Unfortunately, no. We can't put any network-related logic inside `SharedArray`.

tools/k6/src/web3/test/traffic-replay/trafficReplay.js
bilyana-gospodinova: You are right. I changed it to log only 1 message for all replayed requests.

tools/k6/src/web3/test/traffic-replay/trafficReplay.js
bilyana-gospodinova: Removed

tools/k6/src/web3/test/traffic-replay/trafficReplay.js
bilyana-gospodinova: Addressed

tools/k6/src/web3/test/traffic-replay/trafficReplay.js
bilyana-gospodinova: Not applicable anymore after addressing https://github.com/hiero-ledger/hiero-mirror-node/pull/12509#discussion_r2582098781.

tools/k6/src/web3/test/index.js
bilyana-gospodinova: I am not sure this is possible due to the limitation of the `SharedArray` to not make network requests. (https://github.com/hiero-ledger/hiero-mirror-node/pull/12509#discussion_r2582672270).

tools/k6/README.md
```diff
@@ -143,6 +144,7 @@ The following parameters can be used to configure a web3 test:
 | RUN_ESTIMATE_TESTS                                        | true    | If set to false, estimate gas tests will not run.                                                                                                                                 |
 | RUN_MODIFICATION_TESTS                                    | true    | If set to false, modification tests will not run.                                                                                                                                 |
 | RUN_OPCODE_TESTS                                          | true    | If set to false, opcode tests will not run.                                                                                                                                       |
+| RUN_TRAFFIC_REPLAY                                        | true    | If set to false, traffic replay tests will not run.                                                                                                                               |
```
steven-sheehy: Is the default true? I don't see it as set to true in the code.

bilyana-gospodinova: Actually, all of the "RUN_*" variables are not set in the code and effectively they have `false` value by default. I adjusted the documentation.

tools/k6/src/web3/test/traffic-replay/trafficReplay.js
IvanKavaldzhiev: I like the idea of having a dedicated branch used for such heavy performance tests.

IvanKavaldzhiev (approved):
LGTM

steven-sheehy (approved):

PR #12515, opened 2025-12-01
Removed the temporary enablement of fees api in integration

ashumahajan:
**Description**:
This PR removes temporary enablement restjava /network/fees in integration

**Related issue(s)**:

Fixes #

**Checklist**

### DIFF
--- a/clusters/preprod/integration/helmrelease.yaml
+++ b/clusters/preprod/integration/helmrelease.yaml
@@ -59,9 +59,6 @@ spec:
           enabled: false
     postgresql:
       enabled: false
-    restjava:
-      routes:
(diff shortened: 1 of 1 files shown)


clusters/preprod/integration/helmrelease.yaml
```diff
@@ -61,7 +61,7 @@ spec:
       enabled: false
     restjava:
       routes:
-        fees: true
```
jnels124, line 64: We dont need to set to false here. We just need to remove the entry from values.yaml as true is now the default.

steven-sheehy:
Best to merge this change after the rc deployment

PR #12531, opened 2025-12-03
Balance generation updates

### LINKED ISSUE #11858: Account balance service fails to generate balances with large number of entities
### Problem

After adding a billion accounts, the account balance service is unable to generate balances within the 10 minute timeframe that is configured.

### Solution

Optimize approach to generating historical balances

### Alternatives

_No response_

jnels124:
**Description**:
* increase balance partitions to 6 month ranges
* add async migration to backfill balance data
* remove migration v1.1.89 and`BackfillAndDeduplicateBalanceMigration`
* update rest to use 6 month window for balance queries
* add configurable partition ranges via partition_config table
* add BRIN index to `balance_timestamp` on `account_balance` and `token_balance`

**Related issue(s)**:

Fixes #11858 

**Notes for reviewer**:

**Checklist**


importer/src/main/java/org/hiero/mirror/importer/parser/record/historicalbalance/HistoricalBalanceProperties.java
```diff
@@ -47,7 +47,7 @@ public class HistoricalBalanceProperties {
     @DurationMin(seconds = 30)
     @DurationUnit(ChronoUnit.SECONDS)
     @NotNull
-    private Duration transactionTimeout = Duration.ofMinutes(10);
+    private Duration transactionTimeout = Duration.ofHours(4);
```
steven-sheehy, line 50: Need to update configuration.md

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/sql/SqlEntityListener.java
```diff
@@ -186,7 +186,6 @@ public void onEntity(Entity entity) throws ImporterException {
                 && !entityProperties.getPersist().isEntityHistory()) {
             return;
         }
-
```
steven-sheehy, line 189: Could revert unnecessary whitespace change.

rest/__tests__/integrationContainerOps.js
```diff
@@ -124,6 +124,7 @@ const flywayMigrate = async (connectionParams) => {
       password: password,
       'placeholders.api-password': defaultDbConfig.password,
       'placeholders.api-user': defaultDbConfig.username,
+      'placeholders.balancePartitionTimeInterval': "'10years'",
```
steven-sheehy:
```suggestion
      'placeholders.balancePartitionTimeInterval': "'10 years'",
```

importer/src/main/resources/db/migration/v1/V1.115.0__increase_balance_partition_size.sql
jnels124, line 1: Forgot to put the original indexes on new tables and to rename partitions if exists. Will adjust

importer/src/main/resources/db/migration/v1/V1.115.0__increase_balance_partition_size.sql
```diff
@@ -0,0 +1,187 @@
...
+end;
+$$;
+
+-- Functions added by removed migration v1.89.0
+drop function if exists create_full_account_balance_snapshot(bigint, bigint);
+drop function if exists create_deduped_account_balance_snapshot(bigint, bigint);
+drop function if exists create_full_token_balance_snapshot(bigint, bigint);
+drop function if exists create_deduped_token_balance_snapshot(bigint, bigint);
+
+-- Undo partitioning changes from removed migration v1.89.0
```
steven-sheehy: Believe you mean v1.89.1 for these?

importer/src/main/resources/db/migration/v1/V1.115.0__increase_balance_partition_size.sql
```diff
@@ -0,0 +1,187 @@
...
+language plpgsql
+as
+$$
+declare
+  old_table_name text := base_table_name || '_old';
+begin
+  if to_regclass(old_table_name) is not null then
+    execute format(
+      'drop table if exists %I cascade',
+      base_table_name
```
steven-sheehy, line 108: I'm confused. Did v1 migration have a bug that kept `account_balance_old` and `account_balance` while v2 did not? Won't this drop `account_balance` and `token_balance` which should have the most up to date data?

importer/src/main/resources/db/migration/v2/V2.20.0__improve_balance_snapshot_performance.sql
```diff
@@ -0,0 +1,104 @@
...
+(
+  account_id          bigint not null,
+  balance             bigint not null,
+  consensus_timestamp bigint not null
+) partition by range (consensus_timestamp);
+
+comment on table account_balance is
+  'Account balances (historical) in tinybars at different consensus timestamps';
+
+create table if not exists token_balance
```
steven-sheehy, line 62: So there's no means to re-partition in place?

importer/src/main/java/org/hiero/mirror/importer/migration/ExpandBalancePartitionsMigration.java
```diff
@@ -0,0 +1,271 @@
...
+    }
+
+    private void finalizeMigration() {
+        log.info("Finalizing balance migration and dropping old tables");
+        getJdbcOperations().execute(DROP_OLD_TABLES_SQL);
+    }
+
+    private void processPartition(TimePartition partition) {
+        var stopwatch = Stopwatch.createStarted();
+        getJdbcOperations().execute("set local work_mem = '2048MB'"); // Use higher work_mem to avoid temp files
```
steven-sheehy, line 225: Since there's a transaction around entire loop this can be moved out and done once before loop.

importer/src/main/resources/db/migration/v2/V2.20.0__improve_balance_snapshot_performance.sql
steven-sheehy, line 1: How long will all migrations take to complete?

jnels124: takes ~1 hr (including the async migration)

importer/src/main/resources/db/migration/v2/V2.20.0__improve_balance_snapshot_performance.sql
jnels124: No there is no way to change the bounds and avoid rewriting the data

importer/src/main/resources/db/migration/v1/V1.115.0__increase_balance_partition_size.sql
jnels124: There was an async migration that dropped the old tables but that migration was removed in this PR. We would have had two async migrations trying to populate the balance tables. The changes here are to handle cases where the migrations have ran and also cases where they haven't been ran yet. V2 doesn't have this issue because these tables were already partitioned in the original schema.

buildSrc/src/main/kotlin/jooq-conventions.gradle.kts
```diff
@@ -94,6 +94,7 @@ val flywayMigrate =
             arrayOf("filesystem:${migrationDir}/v1", "filesystem:${migrationDir}/common")
         val placeholders =
             mapOf(
+                "balancePartitionTimeInterval" to "'100 years'",
```
xin-hedera: nit, make it 10 years so it's consistent across the board for tests / non-main code path?

importer/src/main/java/org/hiero/mirror/importer/migration/ExpandBalancePartitionsMigration.java
```diff
@@ -0,0 +1,271 @@
...
+import org.springframework.jdbc.core.JdbcOperations;
+import org.springframework.jdbc.core.JdbcTemplate;
+import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
+import org.springframework.jdbc.datasource.DataSourceTransactionManager;
+import org.springframework.transaction.support.TransactionOperations;
+import org.springframework.transaction.support.TransactionTemplate;
+
+@Named
+@NullMarked
+public class ExpandBalancePartitionsMigration extends AsyncJavaMigration<TimePartition> {
```
xin-hedera: nit

```suggestion
final class ExpandBalancePartitionsMigration extends AsyncJavaMigration<TimePartition> {
```

importer/src/main/java/org/hiero/mirror/importer/migration/ExpandBalancePartitionsMigration.java
```diff
@@ -0,0 +1,271 @@
...
+        Boolean performAsync = getTransactionOperations().execute(status -> {
+            final var maxBalanceFileTimestamp =
+                    Objects.requireNonNull(getJdbcOperations().queryForObject(LAST_BALANCE_FILE, Long.class));
+
+            if (maxBalanceFileTimestamp == -1L) {
+                log.info("No balance files found, skipping migration");
+                return Boolean.FALSE;
+            }
+
+            final var sixHoursNs = 6L * 60 * 60 * 1_000_000_000L;
```
xin-hedera: nit,

```suggestion
            final long sixHoursNs = Duration.ofHours(6).toNanos();
```

importer/src/main/java/org/hiero/mirror/importer/migration/ExpandBalancePartitionsMigration.java
```diff
@@ -0,0 +1,271 @@
...
+
+                if (intersects) {
+                    log.info("Synchronously processing partition {}", range);
+                    processPartition(partition);
+                }
+
+                partitionIndex.incrementAndGet();
+            }
+
+            return Boolean.TRUE;
```
xin-hedera, line 184: why do we need to synchronously process partitions overlaps the 6-hour window from the last account balance snapshot's timestamp?

importer/src/main/java/org/hiero/mirror/importer/migration/ExpandBalancePartitionsMigration.java
```diff
@@ -0,0 +1,271 @@
...
+            return false;
+        }
+
+        return true;
+    }
+
+    @Override
+    protected TimePartition getInitial() {
+        return accountBalancePartitions.get(partitionIndex.getAndIncrement());
+    }
```
xin-hedera: can simplify it to just return the index to the time partitions list

```suggestion
    protected Integer getInitial() {
        return 0;
    }
```

importer/src/main/java/org/hiero/mirror/importer/migration/ExpandBalancePartitionsMigration.java
```diff
@@ -0,0 +1,271 @@
...
+
+    @Override
+    protected Optional<TimePartition> migratePartial(TimePartition partition) {
+        processPartition(partition);
+
+        if (partitionIndex.get() >= accountBalancePartitions.size()) {
```
xin-hedera, line 205: believe this should be


```suggestion
        if (partitionIndex.incrementAndGet() >= accountBalancePartitions.size()) {
```

importer/src/main/java/org/hiero/mirror/importer/migration/ExpandBalancePartitionsMigration.java
```diff
@@ -0,0 +1,271 @@
...
+            finalizeMigration();
+            return Optional.empty();
+        }
+
+        return Optional.of(accountBalancePartitions.get(partitionIndex.getAndIncrement()));
```
xin-hedera: also this should be


```suggestion
        return Optional.of(accountBalancePartitions.get(partitionIndex.get()));
```

importer/src/main/java/org/hiero/mirror/importer/migration/ExpandBalancePartitionsMigration.java
```diff
@@ -0,0 +1,271 @@
...
+        getJdbcOperations().execute("set local work_mem = '2048MB'"); // Use higher work_mem to avoid temp files
+
+        final var params = new MapSqlParameterSource()
+                .addValue("from", partition.getTimestampRange().lowerEndpoint())
+                .addValue("to", partition.getTimestampRange().upperEndpoint());
+
+        final var accountV2HasData = getNamedParameterJdbcOperations()
+                .queryForObject(EXISTS_ACCOUNT_BALANCE_V2_RANGE_SQL, params, Boolean.class);
+        final var tokenV2HasData = getNamedParameterJdbcOperations()
+                .queryForObject(EXISTS_TOKEN_BALANCE_V2_RANGE_SQL, params, Boolean.class);
```
xin-hedera: nit: since when either is true, the processing is skipped for the partition, we can combine the two SQL into one.

importer/src/main/java/org/hiero/mirror/importer/migration/ExpandBalancePartitionsMigration.java
jnels124: The previous backfill migrations made sure to process the last six hours in the versioned migration so applied that same logic here. This ensures "newer" balance updates are available for requests.

buildSrc/src/main/kotlin/jooq-conventions.gradle.kts
jnels124: the config here for the other partition interval was 100 years here so did the same for the balance interval to keep it the same. Maybe the other one was intended to be 10 years as well?

importer/src/main/java/org/hiero/mirror/importer/migration/ExpandBalancePartitionsMigration.java
jnels124: There is also a call coming from the parent class that starts a transaction for each migrate partial. This is placed here to handle that case as well as the async portion

importer/src/main/java/org/hiero/mirror/importer/migration/ExpandBalancePartitionsMigration.java
xin-hedera: Checked this again, I believe the logic is correct so no change is needed. We can make it easier to read though.

importer/src/main/java/org/hiero/mirror/importer/migration/ExpandBalancePartitionsMigration.java
jnels124: I can change this but functionally it is correct as is. If it is clearer as suggested, i will make the change

PR #12539, opened 2025-12-04
Add dynamic logic to validate payer balance

### LINKED ISSUE #12437: Enhance web3 to not check for payer's balance unless explicitly wanted
### Problem

Currently the transaction simulations related to `eth_call` and `eth_estimateGas` have a workaround fix to return big amount of default balance, so that the balance checks inside hedera-app pass. Some users might want to have passing reads without passing `from` account or with passing an account with a low balance. And some might want to pass a `from` account that performs a transfer of `value` or set a `gasPrice`. In these cases payer balance check should be validated. The web3 logic should be dynamic and cover all cases.

### Solution

Enhance web3 logic, to enable or disable payer balance check by having a flag set per request based on the fields populated in the request body.

The existing general overridePayerBalanceValidation flag will be kept to have backwards compatibility.

- `from` + `gasPrice` populated with non-zero value - enable payer balance check
- `from` + `value` populated with non-zero value - enable payer balance check
- `from` + `gasPrice` and `value` populated with non-zero value - enable payer balance check
- `from` not populated - disable payer balance check

### Alternatives

_No response_

martingeorgiev1:
**Description**:
  Enhance web3 to dynamically validate payer balance based on transaction parameters

  Implements dynamic per-request balance validation for the `web3` module. When the global `overridePayerBalanceValidation` flag is enabled, balance validation now depends
  on transaction parameters (`from`, `gasPrice`, `value`) rather than being globally disabled.

**Related issue(s)**:

Fixes #12437 

**Notes for reviewer**:

**Checklist**


web3/src/main/java/org/hiero/mirror/web3/service/model/ContractDebugParameters.java
```diff
@@ -33,6 +33,15 @@ public class ContractDebugParameters implements CallServiceParameters {
...
     long gas;
 
+    @PositiveOrZero
+    long gasPrice;
+
+    @PositiveOrZero
+    long maxFeePerGas;
+
+    @PositiveOrZero
+    long maxPriorityFeePerGas;
```
IvanKavaldzhiev: The `maxFeePerGas` and `maxPriorityFeePerGas` are applicable for eth_sendRawTransaction. eth_call and eth_estimateGas don't support them, so they should be deleted and not taken into account.

web3/src/main/java/org/hiero/mirror/web3/common/ContractCallContext.java
```diff
@@ -77,8 +77,38 @@ public class ContractCallContext {
...
+    public boolean shouldOverridePayerBalanceValidation() {
+        if (overridePayerBalanceValidation != null) {
+            return overridePayerBalanceValidation;
+        }
+
+        if (callServiceParameters == null) {
+            return true;
+        }
+
+        if (callServiceParameters.getSender() == null
```
IvanKavaldzhiev: I would merge the 2 checks like:

if (callServiceParameters == null || callServiceParameters.getSender() == null || callServiceParameters.getSender().isZero()) {
            return true;
        }

web3/src/main/java/org/hiero/mirror/web3/common/ContractCallContext.java
```diff
@@ -77,8 +77,38 @@ public class ContractCallContext {
...
+                || callServiceParameters.getSender().isZero()) {
+            return true;
+        }
+
+        // If sender is provided, check if gasPrice or value fields are non-zero
+        boolean hasGasPrice = callServiceParameters.getGasPrice() > 0;
+        boolean hasValue = callServiceParameters.getValue() > 0;
```
IvanKavaldzhiev: The 2 fields can be declared as `final var`

web3/src/test/java/org/hiero/mirror/web3/service/ContractCallNestedCallsTest.java
```diff
@@ -342,7 +347,8 @@ void createFungibleTokenAndGetIsTokenAndGetDefaultFreezeStatusAndGetDefaultKycSt
                 contract.send_createFungibleTokenAndGetIsTokenAndGetDefaultFreezeStatusAndGetDefaultKycStatus(
                         tokenInfo, BigInteger.ONE, BigInteger.ONE, BigInteger.valueOf(CREATE_TOKEN_VALUE));
 
-        verifyEthCallAndEstimateGasWithValue(functionCall, contract, toAddress(treasury.getId()), CREATE_TOKEN_VALUE);
+        verifyEthCallAndEstimateGasWithValue(
```
IvanKavaldzhiev: Why is treasury changed to sender?

web3/src/test/java/org/hiero/mirror/web3/service/ContractCallServiceTest.java
```diff
@@ -1326,4 +1334,91 @@ void shouldNotCallTransactionExecutionService(boolean isModularizedServices, dou
...
+            // Given
+            mirrorNodeEvmProperties.setOverridePayerBalanceValidation(true);
+            final var receiver = accountEntityWithEvmAddressPersist();
+            final var sender = accountEntityPersistCustomizable(
+                    e -> e.type(EntityType.ACCOUNT).balance(100L));
+            final var params = getContractExecutionParametersWithGasAndValue(
+                    getAliasAddressFromEntity(sender), getAliasAddressFromEntity(receiver), 0L, 1000L);
+
+            // Then
+            if (mirrorNodeEvmProperties.isModularizedServices()) {
```
IvanKavaldzhiev: We shouldn't maintain or take into account mono code any more, so this if is now redundant.

web3/src/test/java/org/hiero/mirror/web3/service/ContractCallServiceTest.java
```diff
@@ -1326,4 +1334,91 @@ void shouldNotCallTransactionExecutionService(boolean isModularizedServices, dou
...
+            // Given
+            mirrorNodeEvmProperties.setOverridePayerBalanceValidation(true);
+            final var receiver = accountEntityWithEvmAddressPersist();
+            final var sender = accountEntityPersistCustomizable(
+                    e -> e.type(EntityType.ACCOUNT).balance(100L));
+            final var params = getContractExecutionParametersWithGasAndValue(
+                    getAliasAddressFromEntity(sender), getAliasAddressFromEntity(receiver), 1_000_000L, 0L);
+
+            // Then
+            if (mirrorNodeEvmProperties.isModularizedServices()) {
```
IvanKavaldzhiev: Remove the if statement

IvanKavaldzhiev (changes requested):

web3/src/main/java/org/hiero/mirror/web3/common/ContractCallContext.java
```diff
@@ -77,8 +77,38 @@ public class ContractCallContext {
     @Setter
     private long gasRequirement;
 
+    @Setter
+    private Boolean overridePayerBalanceValidation;
```
steven-sheehy: Where does this get set?

web3/src/test/java/org/hiero/mirror/web3/service/ContractCallServicePrecompileModificationTest.java
```diff
@@ -792,7 +792,10 @@ void unpauseToken() throws Exception {
     void createFungibleToken() throws Exception {
         // Given
         final var value = 10000L * 100_000_000_000L;
-        final var sender = accountEntityPersist();
+        // Create sender with sufficient balance for value transfer
+        // With dynamic balance validation, value > 0 triggers validation
+        final var sender = accountEntityPersistCustomizable(
+                e -> e.type(EntityType.ACCOUNT).evmAddress(null).alias(null).balance(DEFAULT_ACCOUNT_BALANCE));
```
bilyana-gospodinova: This line is common in multiple test cases and can be extracted into a helper method.

web3/src/test/java/org/hiero/mirror/web3/service/ContractCallServiceTest.java
```diff
@@ -1326,4 +1334,91 @@ void shouldNotCallTransactionExecutionService(boolean isModularizedServices, dou
...
+
+            // Then
+            assertDoesNotThrow(() -> contractExecutionService.processCall(params));
+            mirrorNodeEvmProperties.setOverridePayerBalanceValidation(false);
+        }
+
+        @Test
+        void balanceValidationDisabledWhenAllFieldsAreZero() {
+            // Given
+            mirrorNodeEvmProperties.setOverridePayerBalanceValidation(true);
```
bilyana-gospodinova: `mirrorNodeEvmProperties.setOverridePayerBalanceValidation(true);` - can be extracted into a setup method with `@BeforeEach`

web3/src/test/java/org/hiero/mirror/web3/service/ContractCallServiceTest.java
```diff
@@ -1326,4 +1334,91 @@ void shouldNotCallTransactionExecutionService(boolean isModularizedServices, dou
...
+            final var receiver = accountEntityWithEvmAddressPersist();
+            final var sender = accountEntityPersistCustomizable(
+                    e -> e.type(EntityType.ACCOUNT).balance(100L));
+            final var params = getContractExecutionParametersWithGasAndValue(
+                    getAliasAddressFromEntity(sender), getAliasAddressFromEntity(receiver), 0L, 0L);
+
+            // Then
+            assertDoesNotThrow(() -> contractExecutionService.processCall(params));
+            mirrorNodeEvmProperties.setOverridePayerBalanceValidation(false);
```
bilyana-gospodinova: `mirrorNodeEvmProperties.setOverridePayerBalanceValidation(false);` - can be extracted into a cleanup method with `@AfterEach`

web3/src/main/java/org/hiero/mirror/web3/common/ContractCallContext.java
```diff
@@ -77,8 +77,35 @@ public class ContractCallContext {
...
+        final var hasGasPrice = callServiceParameters.getGasPrice() > 0;
+        final var hasValue = callServiceParameters.getValue() > 0;
+
+        // If any of these are non-zero, DON'T override (enable validation)
+        if (hasGasPrice || hasValue) {
+            return false;
+        }
+
+        // Otherwise, override validation
+        return true;
```
steven-sheehy: This is pretty verbose. Can be written more succinctly.

```suggestion
        return callServiceParameters.getGasPrice() <= 0 && callServiceParameters.getValue() <= 0;
```

web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AbstractAliasedAccountReadableKVState.java
```diff
@@ -165,12 +165,20 @@ private Supplier<Long> getAccountBalance(final Entity entity, final Optional<Lon
...
+        final ContractCallContext context = ContractCallContext.get();
+
+        boolean shouldOverride;
         if (!mirrorNodeEvmProperties.isOverridePayerBalanceValidation()) {
+            shouldOverride = false;
+        } else {
+            shouldOverride = context.shouldOverridePayerBalanceValidation();
+        }
+
+        if (!shouldOverride) {
```
steven-sheehy: I think we should invert the logic of shouldOverridePayerBalanceValidation and rename it so that this if is a positive.

```suggestion
        if (!mirrorNodeEvmProperties.isOverridePayerBalanceValidation() && context.validatePayerBalance()) {
```

web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AbstractAliasedAccountReadableKVState.java
```diff
@@ -165,12 +165,20 @@ private Supplier<Long> getAccountBalance(final Entity entity, final Optional<Lon
...
             return Objects.requireNonNullElse(balance, 0L);
         }
 
-        final ContractCallContext context = ContractCallContext.get();
-        final var isBalanceCall = context.isBalanceCallSafe();
+        final var isBalanceCall = ContractCallContext.isBalanceCallSafe();
```
steven-sheehy: Getting the context is relatively expensive.

```suggestion
        final var isBalanceCall = context.isBalanceCallSafe();
```

steven-sheehy (changes requested):

web3/src/test/java/org/hiero/mirror/web3/service/ContractCallServiceTest.java
```diff
@@ -680,15 +681,15 @@ void transferExceedsBalance(boolean overridePayerBalance) {
...
         if (mirrorNodeEvmProperties.isModularizedServices()) {
-            if (mirrorNodeEvmProperties.isOverridePayerBalanceValidation()) {
-                assertThat(contractExecutionService.processCall(serviceParameters))
-                        .isEqualTo(HEX_PREFIX);
-            } else {
-                assertThatThrownBy(() -> contractExecutionService.processCall(serviceParameters))
-                        .isInstanceOf(MirrorEvmTransactionException.class)
-                        .hasMessage(INSUFFICIENT_PAYER_BALANCE.name());
-            }
+            // Both cases should now fail with INSUFFICIENT_PAYER_BALANCE
```
bilyana-gospodinova: After the if-else statement was removed, the comment is now unclear which both cases it refers to. I think we can remove it.

web3/src/test/java/org/hiero/mirror/web3/service/ContractCallServiceTest.java
```diff
@@ -742,7 +743,7 @@ void transferThruContract() throws Exception {
         final var receiver = accountEntityWithEvmAddressPersist();
         final var receiverAddress = getAliasAddressFromEntity(receiver);
         final var contract = testWeb3jService.deploy(EthCall::deploy);
-        final var payer = accountEntityWithEvmAddressPersist();
+        final var payer = accountEntityWithSufficientBalancePersist();
```
bilyana-gospodinova: The entity before had an evm address. Was this change intentional?

web3/src/test/java/org/hiero/mirror/web3/service/ContractCallServiceTest.java
```diff
@@ -1326,4 +1329,89 @@ void shouldNotCallTransactionExecutionService(boolean isModularizedServices, dou
...
+            // Then
+            assertDoesNotThrow(() -> contractExecutionService.processCall(params));
+        }
+
+        @Test
+        void balanceValidationDisabledWhenAllFieldsAreZero() {
+            // Given
+            final var receiver = accountEntityWithEvmAddressPersist();
+            final var sender = accountEntityPersistCustomizable(
+                    e -> e.type(EntityType.ACCOUNT).balance(100L));
```
bilyana-gospodinova: The entity type is already set inside `accountEntityPersistCustomizable()` - no need to override it.

steven-sheehy:
Merge conflict after deleting mono code

web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AbstractAliasedAccountReadableKVState.java
```diff
@@ -165,12 +165,12 @@ private Supplier<Long> getAccountBalance(final Entity entity, final Optional<Lon
     }
 
     private Long getBalanceOrDefaultToMinimum(final Entity entity, final Long balance) {
-        if (!mirrorNodeEvmProperties.isOverridePayerBalanceValidation()) {
-            return Objects.requireNonNullElse(balance, 0L);
-        }
-
         final ContractCallContext context = ContractCallContext.get();
         final var isBalanceCall = context.isBalanceCallSafe();
```
steven-sheehy: nit: Move this line below if statement to optimize it.

web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AbstractAliasedAccountReadableKVState.java
```diff
@@ -165,12 +165,12 @@ private Supplier<Long> getAccountBalance(final Entity entity, final Optional<Lon
...
+
+        if (!mirrorNodeEvmProperties.isOverridePayerBalanceValidation() || context.validatePayerBalance()) {
```
steven-sheehy: @IvanKavaldzhiev Should we be flipping the `overridePayerBalanceValidation` property to `false` with this change?

steven-sheehy: Shouldn't this be a `&&` between the two? The intent of the property after this change should be to feature flag this new capability. It shouldn't necessarily be equivalent to its old meaning. That is, it should toggle between no payer validation regardless of input (the current mainnet default) and some request-specific payer validation.

Probably we should remove the old property and add a new, documented property `validatePayerBalance` with a default of `true`. Then this line becomes:

```java
        if (mirrorNodeEvmProperties.validatePayerBalance() && context.validatePayerBalance()) {
```

Please discuss with Ivan.

steven-sheehy (changes requested):

steven-sheehy (approved):

web3/src/test/java/org/hiero/mirror/web3/service/ContractCallServiceTest.java
```diff
@@ -710,24 +709,23 @@ void transferExceedsBalanceHistorical(boolean overridePayerBalance) {
...
-            assertThatThrownBy(() -> contractExecutionService.processCall(serviceParameters))
-                    .isInstanceOf(MirrorEvmTransactionException.class)
-                    .hasMessage(
-                            "Cannot remove %s wei from account, balance is only %s",
-                            toHexWith64LeadingZeros(value), toHexWith64LeadingZeros(senderEntity.getBalance()));
+            assertDoesNotThrow(() -> contractExecutionService.processCall(serviceParameters));
         }
         assertGasLimit(serviceParameters);
-        mirrorNodeEvmProperties.setOverridePayerBalanceValidation(false);
+        mirrorNodeEvmProperties.setValidatePayerBalance(true);
```
bilyana-gospodinova, line 728: I think it would be better if we keep a backup of the original value of the property before overriding it in the test and restore the original value in here. In the future, if we decide to switch the default value to false, we will need to refactor all of these tests.

IvanKavaldzhiev: With the change of the property name, its logic also changes, so now the correct default value is `true`. I believe the change in tests is only to align for the reverse meaning of the property.

Basically not overriding payer's balance means we should validate the balance (thus returning correct value from DB).

bilyana-gospodinova: The suggestion was just in case the property's default value is changed in the future.

PR #12554, opened 2025-12-04
Move query parameter parsers to their own package

steven-sheehy:
**Description**:

* Enable case-insensitive enum conversion in Spring MVC
* Disallow users to pass internal `unknown` as a range operator
* Move query parameter parsers to their own package
* Some minor refactoring of recent APIs

**Related issue(s)**:

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/rest-java/src/main/java/org/hiero/mirror/restjava/common/RangeOperator.java
+++ b/rest-java/src/main/java/org/hiero/mirror/restjava/common/RangeOperator.java
@@ -59,7 +59,7 @@ public enum RangeOperator {
 
     private static IllegalArgumentException invalidOperator(String rangeOperator) {
         final var name = rangeOperator != null ? rangeOperator.toLowerCase() : null;
-        throw new IllegalArgumentException(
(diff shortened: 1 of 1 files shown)


bilyana-gospodinova (dismissed):

rest-java/src/main/java/org/hiero/mirror/restjava/common/RangeOperator.java
```diff
@@ -48,4 +56,10 @@ public RangeOperator toInclusive() {
             default -> this;
         };
     }
+
+    private static IllegalArgumentException invalidOperator(String rangeOperator) {
+        final var name = rangeOperator != null ? rangeOperator.toLowerCase() : null;
+        throw new IllegalArgumentException(
```
xin-hedera:
```suggestion
        return new IllegalArgumentException(
```

steven-sheehy: Fixed

PR #12557, opened 2025-12-05
HIP-1340 Pectra EOA Code Delegation Design

### LINKED ISSUE #12380: Research the needed changes for Pectra support
### Problem

We need to figure out the needed changes in the mirror node for Pectra support.

### Solution

Research the available notion pages and EIPs and HIPs.

### Alternatives

_No response_

bilyana-gospodinova:
**Description**:
This PR adds the design for the needed changes in the mirror node in order to support HIP-1340.

Fixes #12380

### DIFF
--- a/docs/design/eoa-code-delegation.md
+++ b/docs/design/eoa-code-delegation.md
@@ -102,13 +102,11 @@ add column if not exists authorization_list jsonb null default null;
 #### EthereumTransaction.java
 
 ```java
-// common/src/main/java/org/hiero/mirror/common/domain/transaction/EthereumTransaction.java
 public class EthereumTransaction implements Persistable<Long> {
 
     // This field needs to be added to the existing class.
     @JsonSerialize(using = ObjectToStringSerializer.class)
     @JdbcTypeCode(SqlTypes.JSON)
-    @UpsertColumn(shouldCoalesce = false)
     private List<Authorization> authorizationList;
 }
 ```
@@ -116,25 +114,23 @@ public class EthereumTransaction implements Persistable<Long> {
 where `Authorization` needs to be created as follows:
 
 ```java
-// common/src/main/java/org/hiero/mirror/common/domain/transaction/Authorization.java
 @Data
 @NoArgsConstructor
 @SuperBuilder(toBuilder = true)
 public class Authorization {
 
-    private String chainId;  // Example: "0x01"
     private String address;  // Example: "0x1111111111111111111111111111111111111111"
+    private String chainId;  // Example: "0x01"
     private Long nonce;      // Example: 5
-    private Integer yParity; // Example: 1
     private String r;        // Example: "0x2222222222222222222222222222222222222222222222222222222222222222"
     private String s;        // Example: "0x3333333333333333333333333333333333333333333333333333333333333333"
+    private Integer yParity; // Example: 1
 }
 ```
 
 #### AbstractEntity.java
 
 ```java
-// common/src/main/java/org/hiero/mirror/common/domain/entity/AbstractEntity.java
 public class AbstractEntity implements History {
 
     // This field needs to be added to the existing class.
@@ -186,8 +182,9 @@ will be that the new transaction type will need to decode the authorizationList 
 - `GET /api/v1/accounts/{idOrAliasOrEvmAddress}`
 
 In both API endpoints the returned account model will contain an additional parameter, named `delegation_indicator`. Its value will be
-the persisted code delegation, if any. The format is: `0xef0100 || 20-byte address`. If a code delegation is not set or
-if it was deleted (set to address 0x0000000000000000000000000000000000000000), the `delegation_indicator` field will be empty.
+the persisted code delegation, if any. The format is: `0xef0100 || 20-byte address`. If a code delegation is not set (equal to `null`
+in the DB) or if it was deleted (persisted as address 0x0000000000000000000000000000000000000000 in the DB), the
+`delegation_indicator` field will be returned as `0x` from the REST endpoints above.
 
 - `GET /api/v1/contracts`
 - `GET /api/v1/contracts/{contractIdOrAddress}`


docs/design/eoa-code-delegation.md
```diff
@@ -0,0 +1,319 @@
+# HIP-1340 Mirror Node EOA Code Delegation - Design Document
+
+## Overview
+
+This design document outlines the implementation plan for supporting HIP-1340 EOA code delegation in the Hiero mirror node.
+The implementation will enable extraction, storage, and querying of code delegation data from consensus node transactions,
+performing `eth_call`, `eth_estimateGas` transactions with code delegations in the mirror node and querying the persisted
```
IvanKavaldzhiev: `in the mirror node` can be omitted

docs/design/eoa-code-delegation.md
```diff
@@ -0,0 +1,319 @@
...
+code delegations from the existing REST APIs.
+
+## Goals
+
+- Enhance the database schema to store code delegations
+- Ingest code delegation creation, update, and deletion transactions from the record stream
```
IvanKavaldzhiev: This line can be removed since these scenarios are covered in the next lines.

docs/design/eoa-code-delegation.md
```diff
@@ -0,0 +1,319 @@
...
+- Simulating code delegation creation, deletion or update in web3 (only execution will be supported)
+
+## Architecture
+
+The HIP-1340 implementation follows the established mirror node architecture pattern:
+
+1. **Transaction Processing**: Code delegation-related transactions are processed by dedicated transaction handlers
+2. **Database Storage**: Code delegation data is stored in normalized tables with proper indexing for efficient queries
+3. **REST APIs**: The existing accounts endpoints expose the new code delegation field. The existing contract endpoints will
+   be enhanced to work for EOAs with code delegations as if they are contracts.
```
IvanKavaldzhiev: Move the whole line to the previous one

docs/design/eoa-code-delegation.md
```diff
@@ -0,0 +1,319 @@
...
+- Validating code delegation logic (code delegations are executed by consensus nodes)
+- Simulating code delegation creation, deletion or update in web3 (only execution will be supported)
+
+## Architecture
+
+The HIP-1340 implementation follows the established mirror node architecture pattern:
+
+1. **Transaction Processing**: Code delegation-related transactions are processed by dedicated transaction handlers
+2. **Database Storage**: Code delegation data is stored in normalized tables with proper indexing for efficient queries
+3. **REST APIs**: The existing accounts endpoints expose the new code delegation field. The existing contract endpoints will
```
IvanKavaldzhiev: The new field can be explained more clearly like:

`expose the new code delegation field called code`

docs/design/eoa-code-delegation.md
```diff
@@ -0,0 +1,319 @@
...
+`protected Bytecode readFromDataSource(@NonNull ContractID contractID);`
+This method needs to be updated to try to find a contract with the `contractRepository`. If it is not found,
+search by contract id in the `codeDelegationRepository`.
+
+## Testing Strategy
+
+### 1. Unit Tests
+
+- Transaction handler tests for code delegation persisting
+- Repository tests for data access
```
IvanKavaldzhiev: Also tests for the new Eip7702EthereumTransactionParser

docs/design/eoa-code-delegation.md
```diff
@@ -0,0 +1,319 @@
...
+```sql
+-- add_code_delegations_support.sql
+
+-- Main code delegation table (current state)
+create table if not exists code_delegation
+(
+    entity_id              bigint,       not null,
+    created_timestamp      bigint,       not null,
+    delegation_identifier  bytea,        not null,
+    timestamp_range        int8range,
```
jnels124: This field should be before the bytea as smaller fields should be defined before larger fields

docs/design/eoa-code-delegation.md
```diff
@@ -0,0 +1,319 @@
...
+Create new domain classes following the EntityHistory pattern:
+
+#### AbstractCodeDelegation.java
+
+```java
+// common/src/main/java/org/hiero/mirror/common/domain/transaction/AbstractCodeDelegation.java
+public abstract class AbstractCodeDelegation implements History {
+
+    private EntityId entityId;
+    private Long createdTimestamp;
```
jnels124: May be good to express upsertable attributes here (i.e `@Id`, `@Column(updatable=false)`)

docs/design/eoa-code-delegation.md
jnels124: `timestamp_range` should be `non null` too right?

docs/design/eoa-code-delegation.md
```diff
@@ -0,0 +1,319 @@
...
+- `CryptoUpdateTransactionBody`
+- `EthereumTransactionBody` - with the new type 4 Ethereum transactions
+
+Code delegation execution will be identified through `ContractCallTransactionBody` transactions to the address of the EOA
+and the transaction will be executed in the context of the EOA. For more detailed explanation, please check HIP-1340.
+
+## API Changes
+
+```
+GET /api/v1/accounts
```
jnels124: Should we explicitly call out that the deleted (ie zero address) entries need to be filtered out?

docs/design/eoa-code-delegation.md
```diff
@@ -0,0 +1,319 @@
...
+
+### 2. Integration Tests
+
+- End-to-end API tests
+  - Web3 integration tests for executing a contract call with code delegation:
+    - with a delegation to an existing contract
+    - with a delegation to another EOA
+    - with a delegation to a contract that needs funds to execute the transaction but the EOA does not have enough funds - to verify that the EOA's context is used as expected
+    - with a delegation to a non-existing address - should result in a hollow account creation. The result is a no-op
+    - with a delegation to a system contract - should result in an error
```
jnels124: From the hip, shouldn't a delegation with system contract be a no-op?

docs/design/eoa-code-delegation.md
```diff
@@ -0,0 +1,319 @@
...
+```
+GET /api/v1/contracts/results/{transactionIdOrHash}
+```
+
+In both API endpoints the returned response needs to contain the new `authorization_list` bytes field.
+
+In all `GET /api/v1/contracts/*` endpoints the DB queries need to be modified so that they work with an EOA account the
+same way as the existing queries with contract id:
+
+1. Create a subquery that gets the contract id from the `contract` table. If it is found then return it. Otherwise, try to
```
jnels124: Can you clarify the logic to look things up in the `code_delegation` table?

docs/design/eoa-code-delegation.md
bilyana-gospodinova: Correct, it seems that there was a change in the requirements at some point after I wrote this. Thanks for pointing that out.

docs/design/eoa-code-delegation.md
bilyana-gospodinova: I tried to clarify but I am not sure which filtering you refer to. Could you please check?

docs/design/eoa-code-delegation.md
bilyana-gospodinova: Addressed

docs/design/eoa-code-delegation.md
bilyana-gospodinova: Addressed

docs/design/eoa-code-delegation.md
bilyana-gospodinova: Addressed

docs/design/eoa-code-delegation.md
bilyana-gospodinova: Addressed

docs/design/eoa-code-delegation.md
bilyana-gospodinova: Addressed

docs/design/eoa-code-delegation.md
bilyana-gospodinova: Addressed

docs/design/eoa-code-delegation.md
bilyana-gospodinova: That's right. Addressed.

docs/design/eoa-code-delegation.md
jnels124: I missed the ` If a code delegation is not set or if it was deleted (set to address 0x0000000000000000000000000000000000000000), the code field will be empty.` in the doc. This is the filtering I was referring too

docs/design/eoa-code-delegation.md
jnels124: I guess it is a little but unclear to me what is considered the contract id. Is the entity (EOA) id considered the contract id for these or is it the delegated contract's id that is needed here?

bilyana-gospodinova: It is the EOA id. In the state/hedera-app the EOA id is converted to contract id and all other operations remain the same as if it is a regular contract. I will add the queries for the REST APIs soon.

bilyana-gospodinova:
The design for `debug_traceTransaction` calls will be added additionally. The rest of the design can be reviewed.

docs/design/eoa-code-delegation.md
bilyana-gospodinova: I added the queries.

docs/design/eoa-code-delegation.md
```diff
@@ -0,0 +1,396 @@
+# HIP-1340 Mirror Node EOA Code Delegation - Design Document
+
+## Overview
+
+This design document outlines the implementation plan for supporting HIP-1340 EOA code delegation in the Hiero mirror node.
+The implementation will enable extraction, storage, and querying of code delegation data from consensus node transactions,
+performing `eth_call`, `eth_estimateGas` transactions with code delegations and querying the persisted code delegations
```
steven-sheehy:
```suggestion
performing `eth_call` and `eth_estimateGas` transactions with code delegations and querying the persisted code delegations
```

docs/design/eoa-code-delegation.md
```diff
@@ -0,0 +1,396 @@
...
+- Simulating code delegation creation, deletion or update in web3 (only execution will be supported)
+
+## Architecture
+
+The HIP-1340 implementation follows the established mirror node architecture pattern:
+
+1. **Transaction Processing**: Code delegation-related transactions are processed by dedicated transaction handlers
+2. **Database Storage**: Code delegation data is stored in normalized tables with proper indexing for efficient queries
+3. **REST APIs**: The existing accounts endpoints expose the new code delegation field called `code`. The existing contract endpoints will be enhanced to work for EOAs with code delegations as if they are contracts.
+4. **WEB3 API**: Code delegation executions can be simulated with `eth_call`, `eth_estimateGas`.
```
steven-sheehy:
```suggestion
4. **Web3 API**: Code delegation executions can be simulated with `eth_call` and`eth_estimateGas`.
```

docs/design/eoa-code-delegation.md
```diff
@@ -0,0 +1,396 @@
...
+
+### 1. Code Delegation Table
+
+Create new tables to store code delegation information with history support:
+
+```sql
+-- add_code_delegations_support.sql
+
+-- Main code delegation table (current state)
+create table if not exists code_delegation
```
steven-sheehy: I don't think we need a separate table. We should add an `entity.code_delegation` and just duplicate the entity whenever the delegation changes. We already create a history row for `CryptoUpdate` which this appears in so it would be zero performance impact on write. And read performance will be great as well since we can avoid reading an extra table. Besides, changing an EOA delegate will be very rare.

docs/design/eoa-code-delegation.md
```diff
@@ -0,0 +1,396 @@
...
+Each EOA can have only one `delegation_identifier` set. To delete it, the address part of the identifier is set to the empty
+address - 0x0000000000000000000000000000000000000000.
+
+### 2. Ethereum Table
+
+The existing `ethereum_transaction` table needs to be altered and a new column needs to be added - `authorization_list`
+of type bytea.
+
+```sql
+alter table if exists ethereum_transaction add column if not exists authorization_list bytea;
```
steven-sheehy: Shouldn't we expose this on the contract result APIs? Especially if relay needs it.

docs/design/eoa-code-delegation.md
```diff
@@ -0,0 +1,396 @@
...
+### 2. Transaction Handler Enhancements
+
+Existing transaction handlers will be enhanced to extract code delegation information from transaction bodies:
+
+The following transaction handlers will be modified to process `CodeDelegation`:
+
+- `EthereumTransactionHandler.java` - the `authorization_list` field from the `EthereumTransactionBody` needs to be saved
+  similarly to the other byte array fields.
+
+A child frame of the parent Ethereum transaction will be a `CryptoCreateTransaction` or a `CryptoUpdateTransaction`
```
steven-sheehy: Frame is an EVM term.

```suggestion
A child transaction of the parent Ethereum transaction will be a `CryptoCreateTransaction` or a `CryptoUpdateTransaction`
```

docs/design/eoa-code-delegation.md
```diff
@@ -0,0 +1,396 @@
...
+(depending on whether the affected EOA already exists or not). Their protobufs will be changed to have an additional
+`code` field that will keep the code delegation.
```
steven-sheehy: Don't think this is true. There's no code field noted in the HIP. The code should be in a sidecar or in the block stream state changes. Please verify.

docs/design/eoa-code-delegation.md
```diff
@@ -0,0 +1,396 @@
...
+
+- `GET /api/v1/accounts`
+- `GET /api/v1/accounts/{idOrAliasOrEvmAddress}`
+
+In both API endpoints the returned account model will contain an additional parameter, named `code`. Its value will be
+the persisted code delegation, if any. The format is: `0xef0100 || 20-byte address`. If a code delegation is not set or
+if it was deleted (set to address 0x0000000000000000000000000000000000000000), the `code` field will be empty.
+
+> All `GET /api/v1/contracts/*` endpoints will be modified so that they work with an EOA account the same way as the
+> existing queries with contract id.
```
steven-sheehy: This means that we will be processing ContractFunctionResults from the TransactionRecord that have a `ContractID` populated that is actually an account, right? Is there any work to handle this scenario or is it automatic?

docs/design/eoa-code-delegation.md
```diff
@@ -0,0 +1,396 @@
...
+## API Changes
+
+- `GET /api/v1/accounts`
+- `GET /api/v1/accounts/{idOrAliasOrEvmAddress}`
+
+In both API endpoints the returned account model will contain an additional parameter, named `code`. Its value will be
+the persisted code delegation, if any. The format is: `0xef0100 || 20-byte address`. If a code delegation is not set or
+if it was deleted (set to address 0x0000000000000000000000000000000000000000), the `code` field will be empty.
+
+> All `GET /api/v1/contracts/*` endpoints will be modified so that they work with an EOA account the same way as the
```
steven-sheehy: In what way will they be modified? Is it automatic by nature of getting delegate executions that point to eoa? Do the queries used break if the entity is not of type CONTRACT?

docs/design/eoa-code-delegation.md
```diff
@@ -0,0 +1,396 @@
...
+```sql
+-- add_code_delegations_support.sql
+
+-- Main code delegation table (current state)
+create table if not exists code_delegation
+(
+    entity_id              bigint,       not null,
+    created_timestamp      bigint,       not null,
+    timestamp_range        int8range,    not null,
+    delegation_identifier  bytea,        not null,
```
steven-sheehy: Instead of code_delegation should it be just `code` to match the output in the API?

docs/design/eoa-code-delegation.md
bilyana-gospodinova: "code" was the working name that we discussed in the meetings. It seems that they named it "delegation_address" in the HIP. I will update the name in the design. The rest is valid.

docs/design/eoa-code-delegation.md
bilyana-gospodinova: It should be automatic.

docs/design/eoa-code-delegation.md
bilyana-gospodinova: After this sentence each individual REST endpoint have the needed changes explained. The queries will break and will need to be modified. I wrote an example queries.

docs/design/eoa-code-delegation.md
bilyana-gospodinova: I chose a separate table because the `code_delegation` field is only relevant for accounts (and not for the other entity types like token, schedule, etc) with no plans for support of other entity types in the future so far. Also, adding a new column in the entity table will add it at the end of the table and the compression might not be optimal (unless we recreate the table with reordered columns with a migration). Recently, there was a suggestion from Joseph for a change in the design. Let's wait for the meeting today first to see if it will be applied and revisit the table design again.

docs/design/eoa-code-delegation.md
bilyana-gospodinova: It will be exposed as mentioned in the design:

> - `GET /api/v1/contracts/{contractIdOrAddress}/results`
> - `GET /api/v1/contracts/{contractIdOrAddress}/results/{timestamp}`
> - `GET /api/v1/contracts/results`
> - `GET /api/v1/contracts/results/{transactionIdOrHash}`
> 
> The returned response will contain the new `authorization_list` bytes field that comes from the ethereum table.

docs/design/eoa-code-delegation.md
bilyana-gospodinova: Do you mean the name of the table?

docs/design/eoa-code-delegation.md
steven-sheehy: Since it's a bytea adding it at the end is already optimal. But regardless we shouldn't let such problems influence a good design. Column ordering is more relevant for new tables.

docs/design/eoa-code-delegation.md
steven-sheehy: No, the field. It would be best if the field in the API response matched the field in the db.

docs/design/eoa-code-delegation.md
```diff
@@ -0,0 +1,327 @@
...
+
+- `EthereumTransactionHandler.java` - the `authorization_list` field from the `EthereumTransactionBody` needs to be saved
+  similarly to the other byte array fields.
+
+A child transaction of the parent Ethereum transaction will be a `CryptoCreateTransaction` or a `CryptoUpdateTransaction`
+(depending on whether the affected EOA already exists or not). Their protobufs will be changed to have an additional
+`delegation_address` field that will keep the code delegation.
+
+- `CryptoCreateTransactionHandler.java` - read the `delegation_address` field and save it in the entity
+- `CryptoUpdateTransactionHandler.java` - read the `delegation_address` field and save it in the entity. If the `delegation_address` is equal to 0x0000000000000000000000000000000000000000, then `null` is going to be saved. This equals to a deletion of the code delegation.
```
xin-hedera: think we want to save an empty byte array `delegation_identifier` when it's the empty address, for two benefits:

- smaller storage footprint
- don't have to deal with custom coalesce for the column since the default `coalesce(delegation_identifier, e_delegation_identifier)` will not work when trying to set it to `null`

docs/design/eoa-code-delegation.md
bilyana-gospodinova: Changed to "`eth_call`, `eth_estimateGas` and `debug_traceTransaction`".

docs/design/eoa-code-delegation.md
bilyana-gospodinova: Addressed

docs/design/eoa-code-delegation.md
bilyana-gospodinova: Added the new column into the `entity` table.

docs/design/eoa-code-delegation.md
bilyana-gospodinova: Changed

docs/design/eoa-code-delegation.md
bilyana-gospodinova: Renamed to "delegation_indicator" both in DB and REST API in order to match. This is the latest naming in the HIP as well.

docs/design/eoa-code-delegation.md
bilyana-gospodinova: Addressed. I am trying to clarify the requirements for the REST APIs now with the rest of the team - if we need to treat accounts with 0x0000000000000000000000000000000000000000 code delegation as a regular account. I will update the REST endpoint queries if needed.

docs/design/eoa-code-delegation.md
```diff
@@ -0,0 +1,327 @@
...
+is non-null. This will include EOAs as well.
+
+```sql
+`select ${contractSelectFields}`,
+`from ${Entity.tableName} ${Entity.tableAlias}`,
+`left join ${Contract.tableName} ${Contract.tableAlias}`,
+`on ${Entity.getFullName(Entity.ID)} = ${Contract.getFullName(Contract.ID)}`,
+`where ${Entity.getFullName(Entity.TYPE)} = 'CONTRACT'
+        or (${Entity.getFullName(Entity.TYPE)} = 'ACCOUNT'
+                and ${Entity.getFullName(Entity.ALIAS) is not null)`,
```
xin-hedera:
```suggestion
                and ${Entity.getFullName(Entity.DELEGATION_IDENTIFIER) is not null)`,
```

docs/design/eoa-code-delegation.md
```diff
@@ -0,0 +1,327 @@
...
+if it was deleted (set to address 0x0000000000000000000000000000000000000000), the `delegation_indicator` field will be empty.
+
+> All `GET /api/v1/contracts/*` endpoints will be modified so that they work with EOAs the same way as the
+> existing queries with contracts.
+
+In order to do this the existing DB queries need to be enhanced.
+
+- `GET api/v1/contracts`
+
+The existing query needs to be changed so that the filter query matches the entity type of 'ACCOUNT' as well if its alias
```
xin-hedera:
```suggestion
The existing query needs to be changed so that the filter query matches the entity type of 'ACCOUNT' as well if its delegation identifier
```

docs/design/eoa-code-delegation.md
```diff
@@ -0,0 +1,327 @@
...
+        or (${Entity.getFullName(Entity.TYPE)} = 'ACCOUNT'
+                and ${Entity.getFullName(Entity.ALIAS) is not null)`,
+<the_rest_of_filter_clauses>
+`order by ${Entity.getFullName(Entity.ID)} ${order}`,
+limitQuery,
+```
+
+- `GET api/v1/contracts/{contractIdOrAddress}`
+
+The existing query needs to be changed so that the filter query matches the entity type of 'ACCOUNT' as well if its alias
```
xin-hedera:
```suggestion
The existing query needs to be changed so that the filter query matches the entity type of 'ACCOUNT' as well if its delegation identifier
```

docs/design/eoa-code-delegation.md
bilyana-gospodinova: I see your point. However, the latest requirement is that the `/contracts/*` endpoints should work with EOAs regardless if they have code delegation set or not. I will try to make this clear in the design.

docs/design/eoa-code-delegation.md
bilyana-gospodinova: Same as https://github.com/hiero-ledger/hiero-mirror-node/pull/12557#discussion_r2624204944

docs/design/eoa-code-delegation.md
bilyana-gospodinova: Same as https://github.com/hiero-ledger/hiero-mirror-node/pull/12557#discussion_r2624204944

docs/design/eoa-code-delegation.md
xin-hedera: I'd disagree, that'll make the contracts endpoint with too much useless information, especially when there are so many more accounts without delegations than accounts with delegation & contracts.

bilyana-gospodinova: I will raise that topic to the smart contracts team.

docs/design/eoa-code-delegation.md
```diff
@@ -0,0 +1,327 @@
...
+
+```sql
+`select <needed_entity_and_contract_fields>,
+    coalesce(${Contract.getFullName(Contract.RUNTIME_BYTECODE)},
+        ${Entity.getFullName(Entity.CODE_DELEGATION)}) as runtime_bytecode`,
+`from ${table} ${Entity.tableAlias}`,
+`left join ${Contract.tableName} ${Contract.tableAlias}`,
+`on ${Entity.getFullName(Entity.ID)} = ${Contract.getFullName(Contract.ID)}`,
+where e.type = 'CONTRACT'
+    or (e.type = 'ACCOUNT' and ${Entity.getFullName(Entity.ALIAS) is not null))
```
jnels124: should this be any account with an alias set, or just when there is a non-empty code delegation?

jnels124: nevermind i see this was discussed in previous comments

docs/design/eoa-code-delegation.md
```diff
@@ -0,0 +1,327 @@
...
+    - with `debug_traceTransaction` call to verify that the Ethereum calls are made successfully and that we can provide historical support
+  - REST spec tests that the `/accounts` endpoint returns the new `delegation_indicator` field as expected
+  - REST spec tests that the contract results endpoints return the new `authorization_list` field as expected
+- Database migration tests
+
+### 3. Acceptance Tests
+
+- Executing a contract call to EOA with code delegation to another contract
+- Call the REST `/accounts/{accountId}` endpoint to verify the delegation identifier is returned as expected
+- Call the REST contract results endpoint to verify the authorization list is returned as expected
```
steven-sheehy, line 336: Should also verify can call all contract endpoints with a EOA that has a delegation.

bilyana-gospodinova: Added

docs/design/eoa-code-delegation.md
bilyana-gospodinova: The query was removed so this is no longer applicable.

docs/design/eoa-code-delegation.md
bilyana-gospodinova: The query was removed so this is no longer applicable.

docs/design/eoa-code-delegation.md
bilyana-gospodinova: The query was removed so this is no longer applicable.

xin-hedera (approved):
LGTM

docs/design/eoa-code-delegation.md
```diff
@@ -0,0 +1,343 @@
...
+#### EthereumTransaction.java
+
+```java
+// common/src/main/java/org/hiero/mirror/common/domain/transaction/EthereumTransaction.java
+public class EthereumTransaction implements Persistable<Long> {
+
+    // This field needs to be added to the existing class.
+    @JsonSerialize(using = ObjectToStringSerializer.class)
+    @JdbcTypeCode(SqlTypes.JSON)
+    @UpsertColumn(shouldCoalesce = false)
```
steven-sheehy: Can remove since the table itself is not upsertable

docs/design/eoa-code-delegation.md
```diff
@@ -0,0 +1,343 @@
...
+where `Authorization` needs to be created as follows:
+
+```java
+// common/src/main/java/org/hiero/mirror/common/domain/transaction/Authorization.java
+@Data
+@NoArgsConstructor
+@SuperBuilder(toBuilder = true)
+public class Authorization {
+
+    private String chainId;  // Example: "0x01"
```
steven-sheehy: nit: Sort fields

docs/design/eoa-code-delegation.md
```diff
@@ -0,0 +1,343 @@
...
+    @JdbcTypeCode(SqlTypes.JSON)
+    @UpsertColumn(shouldCoalesce = false)
+    private List<Authorization> authorizationList;
+}
+```
+
+where `Authorization` needs to be created as follows:
+
+```java
+// common/src/main/java/org/hiero/mirror/common/domain/transaction/Authorization.java
```
steven-sheehy: Can omit these path comments here and elsewhere as they will change over time and we'd have to update the docs.

docs/design/eoa-code-delegation.md
```diff
@@ -0,0 +1,343 @@
...
+will be that the new transaction type will need to decode the authorizationList field as well.
+
+## API Changes
+
+- `GET /api/v1/accounts`
+- `GET /api/v1/accounts/{idOrAliasOrEvmAddress}`
+
+In both API endpoints the returned account model will contain an additional parameter, named `delegation_indicator`. Its value will be
+the persisted code delegation, if any. The format is: `0xef0100 || 20-byte address`. If a code delegation is not set or
+if it was deleted (set to address 0x0000000000000000000000000000000000000000), the `delegation_indicator` field will be empty.
```
steven-sheehy: Can update to be more specific as to what will the db field have for this scenario: null or empty zeroes? What should be returned by the REST API: null, 0x, or 0x000..0?

docs/design/eoa-code-delegation.md
bilyana-gospodinova: Removed

docs/design/eoa-code-delegation.md
bilyana-gospodinova: Sorted

docs/design/eoa-code-delegation.md
bilyana-gospodinova: Removed

docs/design/eoa-code-delegation.md
bilyana-gospodinova: Clarified

jnels124 (approved):

xin-hedera (approved):
LGTM

PR #12582, opened 2025-12-10
HIP-1081 Support block streaming from earliest available block 

### LINKED ISSUE #12264: HIP-1081 Support Block Streaming from earliest available block
### Problem

MNs will be update to solely stream from a block node, in some cases the BNs may not have blocks starting from genesis. This can happen in testing scenario or when pointing to a tier 2 or when you want a partial MN.
This is somewhat supported with the `hiero.mirror.importer.startBlockNumber` config, however, this requires a MN operator to manually inquire the appropriate block to set.
It's therefore valuable to have an autonomous flow when the MN can figure out for itself what block number to start from when accepting a starting gap.

### Solution

- Add logic to scan all the BNs available and capture their `firstAvailableBlock` from the `BlockNodeService.serverStatus` endpoint 
- Take the minimum of the set to decide the BN with the greatest historical blocks
- Designate this BN as the one to start streaming from
- Add considerations for priority differences.
  - At what point would a MN switch back to a higher priority BN => maybe when that BN has the next set of historical blocks
  - What happens if said BN goes down and the other available BNs don't have the next set of blocks => we can come back to this later but worth considering how to handle gaps
- Add config to set this flow of picking up the earliest available block

### Alternatives

_No response_

xin-hedera:
**Description**:

This PR adds support of streaming from earliest available block from a block node.

**Related issue(s)**:

Fixes #12264 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/importer/src/main/java/org/hiero/mirror/importer/ImporterProperties.java
+++ b/importer/src/main/java/org/hiero/mirror/importer/ImporterProperties.java
@@ -3,6 +3,7 @@
 package org.hiero.mirror.importer;
 
 import jakarta.validation.Valid;
+import jakarta.validation.constraints.Min;
 import jakarta.validation.constraints.NotBlank;
 import jakarta.validation.constraints.NotNull;
 import jakarta.validation.constraints.PositiveOrZero;
@@ -64,7 +65,7 @@ public class ImporterProperties {
 
     private Instant startDate;
 
-    @PositiveOrZero
+    @Min(-1)
     private Long startBlockNumber;
 
     private Long topicRunningHashV2AddedTimestamp;


importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockFileSource.java
```diff
@@ -62,24 +64,23 @@ final class BlockFileSource extends AbstractBlockSource {
...
-    public void get() {
-        long blockNumber = getNextBlockNumber();
-        var endBlockNumber = commonDownloaderProperties.getImporterProperties().getEndBlockNumber();
-
-        if (endBlockNumber != null && blockNumber > endBlockNumber) {
-            return;
+    protected void doGet(final long blockNumber) {
+        if (blockNumber == EARLIEST_AVAILABLE_BLOCK_NUMBER) {
+            throw new IllegalStateException(
+                    this.getClass().getSimpleName() + " doesn't support earliest available block number");
```
steven-sheehy, line 70: If we leave blockNumber empty for s3 it gets the earliest available block by default, right? So instead of throwing an error we can just set -1 as default and inherit that logic?

importer/src/main/java/org/hiero/mirror/importer/downloader/block/AbstractBlockSource.java
```diff
@@ -41,4 +45,20 @@ protected final BlockFile onBlockStream(BlockStream blockStream) {
...
+        return endBlockNumber == null || blockNumber <= endBlockNumber;
+    }
+
+    private long getNextBlockNumber() {
+        return blockStreamVerifier
+                .getLastBlockFile()
+                .map(BlockFile::getIndex)
+                .map(v -> v + 1)
+                .or(() -> Optional.ofNullable(
+                        commonDownloaderProperties.getImporterProperties().getStartBlockNumber()))
```
steven-sheehy, line 61: Need to adjust validation on property to `@Min(-1)`

importer/src/main/java/org/hiero/mirror/importer/downloader/block/AbstractBlockSource.java
```diff
@@ -41,4 +45,20 @@ protected final BlockFile onBlockStream(BlockStream blockStream) {
...
+                .orElse(GENESIS_BLOCK_NUMBER);
```
steven-sheehy, line 62: Should we reconsider setting a default block num to either -1 or 0 now so that it's never null?

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockFileSource.java
xin-hedera: No, when start fresh with no `startBlockNumber` set, both block source implementations will start from block 0.

To support the same earliest available block number logic when set to -1, we'd have to add block stream file `list` support in the `StreamFileProvider` implementations and extra logic in `BlockFileSource` to list with a marker file with name lexicographically before block 0.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/AbstractBlockSource.java
xin-hedera: umm, I don't plan to make it non-null with a default.

This is how it works (slightly different than what's in the ticket)

- not set, start from block 0
- -1, start from the earliest available block of a block node, or just fail for blockstream files in cloud bucket
- any other allowed number (non-negative), start from the specified block number

PR #12593, opened 2025-12-11
Fix cluster name for prometheus label in mainnet-staging-na

jnels124:
* fix cluster name for prometheus label in mainnet-staging-na

### DIFF
--- a/clusters/mainnet-staging-na/common/helmrelease.yaml
+++ b/clusters/mainnet-staging-na/common/helmrelease.yaml
@@ -48,7 +48,7 @@ spec:
       prometheus:
         prometheusSpec:
           externalLabels:
-            cluster: mainnet-staging-na
+            cluster: mainnet-staging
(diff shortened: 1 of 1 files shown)


clusters/mainnet-staging-na/common/sealed-secret.yaml
jnels124, line 1: cluster name didn't include -na

steven-sheehy: I'm fine with calling this mainnet-staging because I doubt we'll ever resurrect mainnet-staging-eu

jnels124: perfect... That was my plan initially but didnt realize there was existing config for the metrics to use -na... I will change that existing config to match this config.

steven-sheehy (approved):

PR #12594, opened 2025-12-12
Fix NullPointerException when dummy system accounts are accessed in partial mirror node environments.

### LINKED ISSUE #12534: `NullPointerException`: Field key is null
### Description

Only on preprod the following exception persists for a while:
```
2025-12-03 14:20:07.8342025-12-03T12:20:07.834Z ERROR tomcat-handler-20168 o.h.m.w.c.GenericControllerAdvice Generic error:  java.lang.NullPointerException: Field key is null
	at java.base/java.util.Objects.requireNonNull(Unknown Source)
	at com.hedera.hapi.node.state.token.Account.keyOrThrow(Account.java:642)
	at com.hedera.node.app.workflows.prehandle.PreHandleContextImpl.<init>(PreHandleContextImpl.java:152)
	at com.hedera.node.app.workflows.prehandle.PreHandleContextImpl.<init>(PreHandleContextImpl.java:104)
	at com.hedera.node.app.workflows.prehandle.PreHandleWorkflowImpl.expandAndVerifySignatures(PreHandleWorkflowImpl.java:295)
	at com.hedera.node.app.workflows.prehandle.PreHandleWorkflowImpl.preHandleTransaction(PreHandleWorkflowImpl.java:264)
	at com.hedera.node.app.workflows.prehandle.PreHandleWorkflow.preHandleAllTransactions(PreHandleWorkflow.java:90)
	at com.hedera.node.app.workflows.prehandle.PreHandleWorkflow.getCurrentPreHandleResult(PreHandleWorkflow.java:168)
	at com.hedera.node.app.workflows.standalone.impl.StandaloneDispatchFactory.newDispatch(StandaloneDispatchFactory.java:161)
```
On other environments there are no such occurrences.

### Steps to reproduce

```
2025-12-03 14:20:07.8352025-12-03T12:20:07.834Z  WARN tomcat-handler-20168 o.h.m.w.c.LoggingFilter 10.64.24.29 POST /api/v1/contracts/call in 7 ms (mod=true): 500 Field key is null - {"block":"latest","data":"0x2e3cff6a0000000000000000000000000000000000000000000000000000000000000064","estimate":false,"gas":15000000,"gasPrice":100000000,"to":"0x0000000000000000000000000000000000000168"} 
```
Strangely the same requests passes after that:
```
2025-12-03 14:20:12.421INFO2025-12-03T12:20:12.420Z  INFO tomcat-handler-278636 o.h.m.w.c.LoggingFilter 10.64.20.23 POST /api/v1/contracts/call in 9 ms (mod=true): 200 Success - {"block":"latest","data":"0x2e3cff6a0000000000000000000000000000000000000000000000000000000000000064","estimate":false,"gas":15000000,"gasPrice":100000000,"to":"0x0000000000000000000000000000000000000168"} 
```

### Additional context

_No response_

### Hedera network

other

### Version

v0.144.0

### Operating system

None

martingeorgiev1:
**Description**:
Dummy system accounts were created without keys, causing NPE when Hedera's PreHandle workflow called `account.keyOrThrow()` on preprod with partial data.

 * Add `EMPTY_KEY_LIST` to dummy system account creation in `getDummySystemAccountIfApplicable()`
 * Add test to verify `keyOrThrow()` does not throw `NullPointerException` on dummy accounts
 * Enhance existing test to verify dummy accounts have proper key field set

**Related issue(s)**:

Fixes #12534 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AccountReadableKVState.java
+++ b/web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AccountReadableKVState.java
@@ -5,7 +5,6 @@ package org.hiero.mirror.web3.state.keyvalue;
 import static com.hedera.node.app.service.token.impl.schemas.V0490TokenSchema.ACCOUNTS_STATE_ID;
 import static org.hiero.mirror.common.domain.entity.EntityType.ACCOUNT;
 import static org.hiero.mirror.common.domain.entity.EntityType.CONTRACT;
-import static org.hiero.mirror.web3.state.Utils.EMPTY_KEY_LIST;
 
 import com.hedera.hapi.node.base.AccountID;
 import com.hedera.hapi.node.state.token.Account;
@@ -104,7 +103,7 @@ public class AccountReadableKVState extends AbstractAliasedAccountReadableKVStat
             return AccountDetector.isStrictSystem(accountNum) && accountNum != 0
                     ? Optional.of(Account.newBuilder()
                             .accountId(accountID)
-                            .key(EMPTY_KEY_LIST)
+                            .key(getDefaultKey())
                             .build())
                     : Optional.empty();
         }
--- a/web3/src/test/java/org/hiero/mirror/web3/state/keyvalue/AccountReadableKVStateTest.java
+++ b/web3/src/test/java/org/hiero/mirror/web3/state/keyvalue/AccountReadableKVStateTest.java
@@ -652,22 +652,22 @@ class AccountReadableKVStateTest {
 
     @Test
     void returnsDummyAccountForSystemAccountWhenNotFound() {
-        final var systemKey = new AccountID(0, 0, new OneOf<>(AccountOneOfType.ACCOUNT_NUM, 50L));
-        when(commonEntityAccessor.get(systemKey, Optional.empty())).thenReturn(Optional.empty());
+        final var systemAccount = new AccountID(0, 0, new OneOf<>(AccountOneOfType.ACCOUNT_NUM, 50L));
+        when(commonEntityAccessor.get(systemAccount, Optional.empty())).thenReturn(Optional.empty());
 
-        assertThat(accountReadableKVState.readFromDataSource(systemKey)).satisfies(account -> assertThat(account)
+        assertThat(accountReadableKVState.readFromDataSource(systemAccount)).satisfies(account -> assertThat(account)
                 .isNotNull()
-                .returns(systemKey, Account::accountId)
+                .returns(systemAccount, Account::accountId)
                 .returns(0L, Account::tinybarBalance)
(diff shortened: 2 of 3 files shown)


web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AccountReadableKVState.java
```diff
@@ -101,7 +102,10 @@ private Optional<Account> getDummySystemAccountIfApplicable(AccountID accountID)
         if (accountID != null && accountID.hasAccountNum()) {
             final var accountNum = accountID.accountNum();
             return AccountDetector.isStrictSystem(accountNum) && accountNum != 0
-                    ? Optional.of(Account.newBuilder().accountId(accountID).build())
+                    ? Optional.of(Account.newBuilder()
+                            .accountId(accountID)
+                            .key(EMPTY_KEY_LIST)
```
bilyana-gospodinova: I think you are on the right track with this. However, there is a `getKey(...)` method in the parent `AbstractAliasedAccountReadableKVState` which has a similar idea behind its implementation. I think it will be better if we define a `protected getDefaultKey(...)` method in the parent which gets called in here and in the existing `getKey(...)` method. In the new method we can put the existing comment from `getKey(...)` about the use of `EMPTY_KEY_LIST`.

web3/src/test/java/org/hiero/mirror/web3/state/keyvalue/AccountReadableKVStateTest.java
```diff
@@ -658,7 +658,18 @@ void returnsDummyAccountForSystemAccountWhenNotFound() {
...
                 .isNotNull()
                 .returns(systemKey, Account::accountId)
-                .returns(0L, Account::tinybarBalance));
+                .returns(0L, Account::tinybarBalance)
+                .returns(EMPTY_KEY_LIST, Account::key));
+    }
+
+    @Test
+    void dummySystemAccountDoesNotThrowNPEOnKeyOrThrow() {
+        final var systemKey = new AccountID(0, 0, new OneOf<>(AccountOneOfType.ACCOUNT_NUM, 50L));
```
IvanKavaldzhiev: While the account serves as a key in the state, I think a better naming will be something like "systemAccount".

web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AccountReadableKVState.java
IvanKavaldzhiev: Agree with Bilyana. In the parent class there is an existing `getKey` method that already returns `EMPTY_KEY_LIST`, if the key from DB is null. I believe we can just make the method protected and call it on line 107, without defining any new method.

PR #12624, opened 2025-12-16
Update web3j plugins / Update to Gradle 9 

jjohannes:
**Description**:

* Bump Gradle from 8.14.3 to 9.2.1
* Enable Gradle configuration cache
* Fix missing TokenClaimAirdropTransaction in synthetic event documentation
* Remove unused `com.hedera.cryptography` snapshot maven repository
* Update web3j plugins to custom jitpack version with Gradle 9 fixes

**Notes for Reviewer**

To check that everything works, this uses the plugins _from source_. We still requires these fixes to be released:

- https://github.com/LFDT-web3j/web3j-solidity-gradle-plugin/pull/86
- https://github.com/LFDT-web3j/web3j-solidity-gradle-plugin/pull/88
- https://github.com/LFDT-web3j/web3j-gradle-plugin/pull/100

### DIFF
--- a/buildSrc/src/main/kotlin/java-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/java-conventions.gradle.kts
@@ -46,12 +46,7 @@ dependencies {
 tasks.withType<JavaCompile>().configureEach {
     // Disable serial and this-escape warnings due to errors in generated code
     options.compilerArgs.addAll(
-        listOf(
-            "-parameters",
-            "-Werror",
-            "-Xlint:all",
-            "-Xlint:-this-escape,-preview,-dangling-doc-comments",
-        )
+        listOf("-parameters", "-Werror", "-Xlint:all", "-Xlint:-this-escape,-preview")
     )
     options.encoding = "UTF-8"
     options.errorprone {


buildSrc/build.gradle.kts
```diff
@@ -5,6 +5,7 @@ plugins { `kotlin-dsl` }
 repositories {
     gradlePluginPortal()
     mavenCentral()
+    maven { url = uri("https://www.jitpack.io") } // Temporary until next web3j
```
jjohannes:
```suggestion
repositories {
    // Temporary until next web3j
    maven("https://www.jitpack.io") { content { includeGroupByRegex(".*web3j.*") } }
    gradlePluginPortal()
}
```

This way it won't search for the non-existing versions in `gradlePluginPortal()`. It will go to jitpack first, but only for "org.web3j". `mavenCentral()` can be removed as it is never used. `gradlePluginPortal()` nowadays acts as mirror of Maven Central for everything it does not host directly.

buildSrc/src/main/kotlin/common-conventions.gradle.kts
```diff
@@ -23,6 +23,7 @@ dependencyCheck {
         datafeedUrl = "https://dependency-check.github.io/DependencyCheck_Builder/nvd_cache/"
     }
     suppressionFile = resources.resolve("suppressions.xml").toString()
+    skipConfigurations = listOf("jacocoAgent", "jacocoAnt")
```
jjohannes, line 26: The way `dependencyCheckAggregate` works will become unsupported at some point. It will not be possible from one project to look into the dependencies of another project directly. It's already forbidden if you turn on [Isolated Projects](https://docs.gradle.org/current/userguide/isolated_projects.html). But it's also stricter in General in Gradle 9. That's what causes the following error if we do not add these exclusions.

```
 > Resolution of the configuration ':rest:check-state-proof:jacocoAgent' was attempted without an exclusive lock. This is unsafe and not allowed.
```

On the long run, the _dependency check_ plugin should be used differently. Maybe only use the `dependencyCheckAnalyze` task that analyzes each project directly. It may also be sufficient to only analyze the `runtimeClasspath` configuration.

jnels124 (commented):
Seems I am no longer able to execute dockerBuild tasks with these changes. I get the below error. I get the same error for other modules
```
FAILURE: Build failed with an exception.

* What went wrong:
Could not evaluate onlyIf predicate for task ':importer:dockerBuild'.
> Could not evaluate spec for 'Task satisfies onlyIf spec'.
```

When running `./gradlew :importer:dockerBuild` from cli, I see additional context
```
2 problems were found storing the configuration cache.
- Task `:importer:dockerBuild` of type `com.bmuschko.gradle.docker.tasks.image.DockerBuildImage`: cannot deserialize Gradle script object references as these are not supported with the configuration cache.
  See https://docs.gradle.org/9.2.1/userguide/configuration_cache_requirements.html#config_cache:requirements:disallowed_types
- Task `:importer:dockerBuild` of type `com.bmuschko.gradle.docker.tasks.image.DockerBuildImage`: cannot serialize Gradle script object references as these are not supported with the configuration cache.
  See https://docs.gradle.org/9.2.1/userguide/configuration_cache_requirements.html#config_cache:requirements:disallowed_types
```

jnels124 (approved):

PR #12631, opened 2025-12-17
Remove unneeded findLatestIndex() db call

### LINKED ISSUE #12627: Investigate the option of removing findLatestIndex from the findByBlockType in RecordFileServiceImpl
Currently we always call findLatestIndex() during historical calls and in almost all cases the block.number is a valid one. Optimize it so we can reduce the calls to the database.

```
@Override
    public Optional<RecordFile> findByBlockType(BlockType block) {
            return recordFileRepository.findEarliest();
        } else if (block == BlockType.LATEST) {
            return recordFileRepository.findLatest();
        }

        long latestBlock = recordFileRepository
                .findLatestIndex()
                .orElseThrow(() -> new BlockNumberOutOfRangeException(UNKNOWN_BLOCK_NUMBER));

        if (block.number() > latestBlock) {
            throw new BlockNumberOutOfRangeException(UNKNOWN_BLOCK_NUMBER);
        }
        return recordFileRepository.findByIndex(block.number());
    }
```

Ticket resulted from the following [discussion](https://github.com/hiero-ledger/hiero-mirror-node/pull/12489#issuecomment-3661566589).

sdimitrov9:
**Description**:
Remove findLatestIndex() query always executed in findByBlockType() during historical calls as not needed.

**Related issue(s)**:

Fixes #12627 

**Notes for reviewer**:

**Checklist**


web3/src/main/java/org/hiero/mirror/web3/service/RecordFileServiceImpl.java
```diff
@@ -26,14 +26,9 @@ public Optional<RecordFile> findByBlockType(BlockType block) {
             return recordFileRepository.findLatest();
         }
 
-        long latestBlock = recordFileRepository
-                .findLatestIndex()
```
steven-sheehy, line 30: Please remove now unused `findLatestIndex()` in repository and test

web3/src/main/java/org/hiero/mirror/web3/service/RecordFileServiceImpl.java
```diff
@@ -26,14 +26,9 @@ public Optional<RecordFile> findByBlockType(BlockType block) {
...
-                .orElseThrow(() -> new BlockNumberOutOfRangeException(UNKNOWN_BLOCK_NUMBER));
-
-        if (block.number() > latestBlock) {
+        return recordFileRepository.findByIndex(block.number()).or(() -> {
             throw new BlockNumberOutOfRangeException(UNKNOWN_BLOCK_NUMBER);
```
steven-sheehy: We should not throw an error if it returns an optional. We should either:

a) Change interface to not return an Optional and always throw
b) Remove throw as we already throw in caller class

sdimitrov9: The throw is removed now.

web3/src/main/java/org/hiero/mirror/web3/service/RecordFileServiceImpl.java
```diff
@@ -26,13 +23,6 @@ public Optional<RecordFile> findByBlockType(BlockType block) {
             return recordFileRepository.findLatest();
         }
 
-        long latestBlock = recordFileRepository
-                .findLatestIndex()
-                .orElseThrow(() -> new BlockNumberOutOfRangeException(UNKNOWN_BLOCK_NUMBER));
-
-        if (block.number() > latestBlock) {
-            throw new BlockNumberOutOfRangeException(UNKNOWN_BLOCK_NUMBER);
```
bilyana-gospodinova, line 34: Does `BlockNumberOutOfRangeException` have any other usages? I suppose not. If so, it can be removed.
Also, `ContractControllerTest#callWithBlockNumberOutOfRangeExceptionTest()` test is not relevant anymore and can be removed as well.

sdimitrov9: Done

PR #12643, opened 2025-12-18
Enable monitor testing in CI

### LINKED ISSUE #11990: Test monitor in CI
### Problem

The Java monitor connects directly to consensus nodes but we don't have any tests to verify those calls to catch regressions earlier.

### Solution

* Depends on #12348
* Figure out how to verify monitor is publishing/subscribing successfully. Possibly call `/actuator/health/cluster` or examining metrics endpoint and verifying metrics are increasing
* Add a postman.json and helm test to `hedera-mirror-monitor` sub-chart
* monitor chart should have similar properties as other charts with `test.enabled` defaulted to false except in acceptance workflow

### Alternatives

_No response_

IvanKavaldzhiev:
**Description**:
This PR enhances the CI/CD to also run tests related to monitor module when `helm test` is invoked.

**Related issue(s)**:

Fixes #11990 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/.github/workflows/acceptance.yaml
+++ b/.github/workflows/acceptance.yaml
@@ -199,6 +199,9 @@ jobs:
           web3:
             env:
               HIERO_MIRROR_WEB3_OPCODE_TRACER_ENABLED: "true"
+            test:
+              annotations:
+                helm.sh/hook-delete-policy: before-hook-creation
           EOF
 
           else
@@ -230,6 +233,9 @@ jobs:
           web3:
             env:
               HIERO_MIRROR_WEB3_OPCODE_TRACER_ENABLED: "true"
+            test:
+              annotations:
+                helm.sh/hook-delete-policy: before-hook-creation
           EOF
           fi
 


charts/hedera-mirror-monitor/values.yaml
```diff
@@ -301,6 +301,19 @@ serviceMonitor:
 
 terminationGracePeriodSeconds: 60
 
+test:
+  annotations:
+    helm.sh/hook: test-success
+    helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded
+  enabled: true
```
steven-sheehy: The default should be false for now so it doesn't break our production deploys.

charts/hedera-mirror-monitor/postman.json
```diff
@@ -0,0 +1,107 @@
...
+  "item": [
+    {
+      "name": "Check Prometheus Publish/Subscribe Metrics Increasing",
+      "event": [
+        {
+          "listen": "test",
+          "script": {
+            "exec": [
+              "const baseUrl = pm.environment.get(\"baseUrl\");",
+              "const url = baseUrl + \"/actuator/prometheus\";",
```
steven-sheehy: It's pretty brittle to use the metrics. We already have the `/actuator/health/cluster` which checks the publishing and subscribing rate. The problem is that it just returns UNKNOWN/200 if the rates drop to zero. We should add property that's off by default to return DOWN when these rates drop to zero. Then enable it only in CI. The reason is consensus nodes can be down/upgrading so we can't mark it as down for production but we can be more strict in CI. Then this postman just becomes a simple 200 check.

docs/configuration.md
```diff
@@ -432,6 +432,7 @@ monitor.
 | ----------------------------------------------------------------- | ----------- | ------------------------------------------------------------------------------------------------------------------------------------------------ |
 | `hiero.mirror.monitor.health.release.cacheExpiry`                 | 30s         | The amount of time to cache cluster release health status                                                                                        |
 | `hiero.mirror.monitor.health.release.enabled`                     | false       | Whether to enable cluster release health check                                                                                                   |
+| `hiero.mirror.monitor.health.release.enableDownStatus`            | false       | Whether to enable Status.DOWN when publisher or subscriber metri rates go to zero. Used only for tests in CI.                                    |
```
steven-sheehy: Maybe name `failWhenInactive`?

.github/workflows/acceptance.yaml
```diff
@@ -173,6 +173,11 @@ jobs:
             env:
               HIERO_MIRROR_IMPORTER_BLOCK_NODES_0_HOST: 'block-node-1.${SOLO_NAMESPACE}.svc.cluster.local'
               SPRING_PROFILES_ACTIVE: 'blocknode'
+          monitor:
+            env:
+              HIERO_MIRROR_MONITOR_HEALTH_RELEASE_ENABLE_DOWN_STATUS: "true"
```
steven-sheehy: Believe you need to remove the underscores for it to work.

```suggestion
              HIERO_MIRROR_MONITOR_HEALTH_RELEASE_ENABLEDOWNSTATUS: "true"
```

monitor/src/main/java/org/hiero/mirror/monitor/health/ClusterHealthIndicator.java
```diff
@@ -65,7 +67,7 @@ private Mono<Health> subscribing() {
                 .reduce(0.0, (cur, next) -> cur + next)
                 .filter(sum -> sum > 0)
                 .flatMap(n -> UP)
-                .switchIfEmpty(UNKNOWN);
+                .switchIfEmpty(releaseHealthProperties.isEnableDownStatus() ? DOWN : UNKNOWN);
```
steven-sheehy: Probably move to a helper function so you can reuse

docs/configuration.md
steven-sheehy: Spelling and omit CI sentence as it might change.

```suggestion
| `hiero.mirror.monitor.health.release.enableDownStatus`            | false       | Whether to enable Status.DOWN when publisher or subscriber metric rates go to zero.                                   |
```

charts/hedera-mirror-monitor/postman.json
```diff
@@ -0,0 +1,75 @@
...
+              "",
+              "function checkResponse(res, count) {",
+              "    pm.test(`Call ${count} status code should be 200`, () => {",
+              "        pm.expect(res.code).to.equal(200);",
+              "    });",
+              "    return res.code === 200;",
+              "}",
+              "",
+              "function nextCall() {",
+              "    if (callCount < numCalls) {",
```
steven-sheehy: Does this frequently fail in CI until nodes are ready? Probably instead should adjust it to run after acceptance tests since that has a startup probe. Can remove this retry logic and add `helm.sh/hook-weight:` annotation with an appropriate value.

docs/configuration.md
IvanKavaldzhiev: Fixed

.github/workflows/acceptance.yaml
IvanKavaldzhiev: Fixed

monitor/src/main/java/org/hiero/mirror/monitor/health/ClusterHealthIndicator.java
IvanKavaldzhiev: Fixed

docs/configuration.md
IvanKavaldzhiev: Fixed

charts/hedera-mirror-monitor/postman.json
IvanKavaldzhiev: Adapted

.github/workflows/acceptance.yaml
```diff
@@ -172,6 +172,11 @@ jobs:
             env:
               HIERO_MIRROR_IMPORTER_BLOCK_NODES_0_HOST: 'block-node-1.${SOLO_NAMESPACE}.svc.cluster.local'
               SPRING_PROFILES_ACTIVE: 'blocknode'
+          monitor:
+            env:
+              HIERO_MIRROR_MONITOR_HEALTH_RELEASE_ENABLEDOWNSTATUS: "true"
```
steven-sheehy: Need to update property names in this file.

steven-sheehy (commented):
Any idea why acceptance shows the error: `unable to get pod logs for mirror-1-rest-test: pods "mirror-1-rest-test" not found`?

.github/workflows/acceptance.yaml
IvanKavaldzhiev: Fixed

IvanKavaldzhiev:
> Any idea why acceptance shows the error: `unable to get pod logs for mirror-1-rest-test: pods "mirror-1-rest-test" not found`?

The rest test pod was enabled. I disabled it. However, the same error appear now for monitor-test. I believe the test passes so quick, that the pod shuts down and the logs cannot be fetched. I tried to set a sleep duration, however there is an issue with building solo and the fix currently can't be validated.

steven-sheehy:
The `unable to get pod logs for` error seems to be due to https://github.com/helm/helm/issues/9098. We should revert the sleeps. We can workaround by removing the `hook-succeeded` from the `helm.sh/hook-delete-policy` in CI.

.github/workflows/acceptance.yaml
```diff
@@ -172,6 +172,19 @@ jobs:
...
+            test:
+              annotations:
+                helm.sh/hook-delete-policy: before-hook-creation
+              enabled: true
+          rest:
+            test:
+              enabled: false
+          restjava:
+            test:
+              enabled: false
```
steven-sheehy: It's fine to disable other modules as a test to see if it works, but now let's revert these so they're re-enabled. If we have the log failure for these other modules then do same hook-delete-policy workaround for each.

IvanKavaldzhiev: Fixed

steven-sheehy (approved):

PR #12650, opened 2025-12-19
HIP-1259 fee collection account

### LINKED ISSUE #12580: HIP-1259 fee collection account
### Problem

As a mirror node operator, I want smaller block streams to lower my data ingestion and storage costs.

### Solution

* Add 802 to system entity classes in each module
* Find any place that has logic to detect fee accounts like 98/800/801 and add 802
* Ensure `RecordItemBuilder` generates default transfers with just a single transfer to 802 and any node transfer
* Update logic in mirror-report

### Alternatives

_No response_

martingeorgiev1:
**Description**:

Implement HIP-1259 support in mirror node, by adding fee collection account.

- Added 802 to system entity classes in each module
- Updated logic in mirror-report
- Update the corresponding unit tests
- Added 802 to `entityTransactionExclusion` including it also in `configuration.md`

Fixes #12580 

**Checklist**


common/src/main/java/org/hiero/mirror/common/domain/SystemEntity.java
```diff
@@ -26,6 +26,9 @@ public class SystemEntity {
     @Getter(lazy = true)
     private final EntityId feeCollectorAccount = toEntityId(98L);
```
steven-sheehy: Now these two names are pretty similar. The HIP defines this as the `Network Admin Fee Account` so let's rename this to `networkAdminFeeAccount` so it aligns.

common/src/main/java/org/hiero/mirror/common/domain/SystemEntity.java
```diff
@@ -24,7 +24,10 @@ public class SystemEntity {
     private final EntityId exchangeRateFile = toEntityId(112L);
 
     @Getter(lazy = true)
-    private final EntityId feeCollectorAccount = toEntityId(98L);
+    private final EntityId networkAdminFeeAccount = toEntityId(98L);
```
steven-sheehy: nit: sort by name

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/EntityProperties.java
```diff
@@ -114,7 +114,8 @@ public static class PersistProperties {
 
         public PersistProperties(SystemEntity systemEntity) {
             this.entityTransactionExclusion = Set.of(
-                    systemEntity.feeCollectorAccount(),
+                    systemEntity.networkAdminFeeAccount(),
```
steven-sheehy: nit: sort

rosetta/app/persistence/domain/system_entity.go
```diff
@@ -8,15 +8,17 @@ import (
 )
 
 const (
-	addressBook101FileNum   = 101
-	addressBook102FileNum   = 102
-	stakingRewardAccountNum = 800
-	treasuryAccountNum      = 2
+	addressBook101FileNum       = 101
+	addressBook102FileNum       = 102
+	feeCollectionAccountNum     = 802
```
steven-sheehy: Not sure we need to add it unless it is actually needed somewhere. The task was more to analyze if it's necessary and add it if so. So we can probably revert rosetta changes.

web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AccountReadableKVState.java
```diff
@@ -65,7 +65,8 @@ public AccountReadableKVState(
         this.commonEntityAccessor = commonEntityAccessor;
         this.aliasedAccountCacheManager = aliasedAccountCacheManager;
         this.systemAccounts = Set.of(
-                EntityIdUtils.toAccountId(systemEntity.feeCollectorAccount()),
+                EntityIdUtils.toAccountId(systemEntity.networkAdminFeeAccount()),
+                EntityIdUtils.toAccountId(systemEntity.feeCollectionAccount()),
```
steven-sheehy, line 68: nit: sort

tools/mirror-report/test/report.test.js
```diff
@@ -100,6 +102,7 @@ describe('Generate report', () => {
             {account: '0.0.98', amount: 2},
             {account: '0.0.800', amount: 3},
             {account: '0.0.801', amount: 4},
+            {account: '0.0.802', amount: 5},
```
steven-sheehy, line 105: Did you run the tests? We should see a corresponding increase in the fees in the expected csv output.

martingeorgiev1: Tests passed before because the report tool uses only `charged_tx_fee` for the CSV fees column and ignores all fee account transfers (they're filtered out by the `feeAccount` regex). So adding the 802 transfer had no effect on output until `charged_tx_fee` was also updated.

steven-sheehy (approved):
LGTM

PR #12668, opened 2025-12-23
HIP-1081 Block node metrics

### LINKED ISSUE #12350: HIP-1081 Block node metrics
### Problem

Add block node metrics for observability

### Solution

Add the following block node metrics

- Add `hiero.mirror.importer.stream.latency` to `BlockNodeSubscriber`, defined as the time difference between the last transaction in the block and the time at which the block was verified. Note the metrics should have a block node tag to tell which block node the latency is recorded for
- Add `hiero.mirror.importer.stream.error` to `BlockNode`, defined as the number of errors occurred while streaming from a particular block node
- Update `hedera.mirror.download.stream.verification` in `BlockStreamVerifer` to include block node tag if the block is streamed from a block node

### Alternatives

_No response_

bilyana-gospodinova:
**Description**:
This PR adds the following metrics:
- `hiero.mirror.importer.stream.latency`, defined as the time difference between the last transaction in the block and the time at which the block was verified. The metric includes a block node tag to tell which block node the latency is recorded for.
- `hiero.mirror.importer.stream.error`, defined as the number of errors occurred while streaming from a particular block node. The metric includes a block node tag to tell which block node the latency is recorded for.
- Enhanced `hiero.mirror.importer.stream.verification` to include the block node tag.

Fixes #12350

### DIFF
--- a/importer/src/main/java/org/hiero/mirror/importer/downloader/block/AbstractBlockSource.java
+++ b/importer/src/main/java/org/hiero/mirror/importer/downloader/block/AbstractBlockSource.java
@@ -36,17 +36,12 @@ abstract class AbstractBlockSource implements BlockSource {
 
     protected abstract void doGet(final long blockNumber);
 
-    protected final BlockFile onBlockStream(
-            final BlockStream blockStream, final BlockNodeProperties blockNodeProperties) {
-        var blockFile = readBlockFile(blockStream);
-        blockFile.setNode(blockNodeProperties.getEndpoint());
-        blockStreamVerifier.verify(blockFile);
-        return blockFile;
-    }
-
-    protected final BlockFile onBlockStream(final BlockStream blockStream) {
-        var blockFile = readBlockFile(blockStream);
-        blockFile.setNode("cloud");
+    protected final BlockFile onBlockStream(final BlockStream blockStream, final String blockNodeEndpoint) {
+        var blockFile = blockStreamReader.read(blockStream);
+        if (!properties.isPersistBytes()) {
+            blockFile.setBytes(null);
+        }
+        blockFile.setNode(blockNodeEndpoint);
         blockStreamVerifier.verify(blockFile);
         return blockFile;
     }
@@ -65,13 +60,5 @@ abstract class AbstractBlockSource implements BlockSource {
                 .or(() -> Optional.ofNullable(
                         commonDownloaderProperties.getImporterProperties().getStartBlockNumber()))
                 .orElse(GENESIS_BLOCK_NUMBER);
-    }
-
-    private BlockFile readBlockFile(final BlockStream blockStream) {
-        var blockFile = blockStreamReader.read(blockStream);
-        if (!properties.isPersistBytes()) {
-            blockFile.setBytes(null);
-        }
-        return blockFile;
     }
 }
--- a/importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockNode.java
+++ b/importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockNode.java
@@ -45,6 +45,7 @@ import org.jspecify.annotations.NullMarked;
 @NullMarked
 final class BlockNode implements AutoCloseable, Comparable<BlockNode> {
 
+    static final String ERROR_METRIC_NAME = "hiero.mirror.importer.stream.error";
     private static final Comparator<BlockNode> COMPARATOR = Comparator.comparing(blockNode -> blockNode.properties);
     private static final Range<Long> EMPTY_BLOCK_RANGE = Range.closedOpen(0L, 0L);
     private static final ServerStatusRequest SERVER_STATUS_REQUEST = ServerStatusRequest.getDefaultInstance();
@@ -79,10 +80,10 @@ final class BlockNode implements AutoCloseable, Comparable<BlockNode> {
         this.grpcBufferDisposer = grpcBufferDisposer;
         this.properties = properties;
         this.streamProperties = streamProperties;
-        this.errorsMetric = Counter.builder("hiero.mirror.importer.stream.error")
+        this.errorsMetric = Counter.builder(ERROR_METRIC_NAME)
                 .description("The number of errors that occurred while streaming from a particular block node.")
                 .tag("type", StreamType.BLOCK.toString())
-                .tag("node", properties.getEndpoint())
+                .tag("block_node", properties.getEndpoint())
                 .register(meterRegistry);
     }
 
--- a/importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockStreamVerifier.java
+++ b/importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockStreamVerifier.java
@@ -88,7 +88,7 @@ final class BlockStreamVerifier {
             verifyHashChain(blockFile);
             final var consensusEnd = Instant.ofEpochSecond(0, blockFile.getConsensusEnd());
             streamLatencyMeterProvider
-                    .withTag("node", blockFile.getNode())
+                    .withTag("block_node", blockFile.getNode())
                     .record(Duration.between(consensusEnd, Instant.now()));
             var recordFile = blockFileTransformer.transform(blockFile);
             streamFileNotifier.verified(recordFile);
@@ -106,7 +106,7 @@ final class BlockStreamVerifier {
             throw e;
         } finally {
             streamVerificationMeterProvider
-                    .withTags("success", String.valueOf(success), "node", blockFile.getNode())
+                    .withTags("success", String.valueOf(success), "block_node", blockFile.getNode())
                     .record(Duration.between(startTime, Instant.now()));
         }
     }
--- a/importer/src/test/java/org/hiero/mirror/importer/downloader/block/BlockNodeTest.java
+++ b/importer/src/test/java/org/hiero/mirror/importer/downloader/block/BlockNodeTest.java
@@ -4,6 +4,7 @@ package org.hiero.mirror.importer.downloader.block;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.hiero.mirror.importer.downloader.block.BlockNode.ERROR_METRIC_NAME;
 
 import com.asarkar.grpc.test.GrpcCleanupExtension;
 import com.asarkar.grpc.test.Resources;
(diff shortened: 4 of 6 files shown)


importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockStreamVerifier.java
```diff
@@ -51,6 +53,11 @@ public BlockStreamVerifier(
                 .description("The duration in seconds it took to verify consensus and hash chain of a stream file")
                 .tag("type", StreamType.BLOCK.toString());
 
+        this.streamLatencyMetric = Timer.builder("hiero.mirror.importer.stream.latency")
```
steven-sheehy: Believe BlockFileSource already has this metric so this will double count for block stream in bucket. Should remove it there.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockStreamVerifier.java
```diff
@@ -79,6 +86,11 @@ public void verify(BlockFile blockFile) {
         try {
             verifyBlockNumber(blockFile);
             verifyHashChain(blockFile);
+            final var consensusEnd = Instant.ofEpochSecond(0, blockFile.getConsensusEnd());
+            streamLatencyMetric
+                    .tag("node", blockFile.getNode())
+                    .register(meterRegistry)
```
steven-sheehy: It's slow to use register dynamically this way. Use `MeterProvider<Timer> provider = Timer.builder("...").tag(...).withRegistry(meterRegistry);`

then

`provider.tags("node", blockFile.getNode()).record(...)`

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockStreamVerifier.java
```diff
@@ -96,6 +108,7 @@ public void verify(BlockFile blockFile) {
         } finally {
             streamVerificationMetric
                     .tag("success", String.valueOf(success))
+                    .tag("node", blockFile.getNode())
                     .register(meterRegistry)
                     .record(Duration.between(startTime, Instant.now()));
```
steven-sheehy: Change this to also use MeterProvider.

importer/src/main/java/org/hiero/mirror/importer/reader/block/BlockStreamReaderImpl.java
```diff
@@ -47,6 +48,8 @@
 @NullMarked
 public final class BlockStreamReaderImpl implements BlockStreamReader {
 
+    private final BlockNodeProperties blockNodeProperties = new BlockNodeProperties();
```
steven-sheehy: This will not work since this is not a bean and there can be many BlockNodeProperties in `BlockProperties.nodes`. We should figure out which it is currently connected to in the BlockNodeSubscriber.

importer/src/test/java/org/hiero/mirror/importer/downloader/block/BlockNodeTest.java
```diff
@@ -60,6 +62,8 @@ final class BlockNodeTest extends BlockNodeTestBase {
     private BlockNodeProperties blockNodeProperties;
     private BlockNode node;
     private StreamProperties streamProperties;
+    private final MeterRegistry meterRegistry = new SimpleMeterRegistry();
```
steven-sheehy: Move the registry to parent class and remove from both children.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockStreamVerifier.java
```diff
@@ -79,6 +86,11 @@ public void verify(BlockFile blockFile) {
         try {
             verifyBlockNumber(blockFile);
             verifyHashChain(blockFile);
+            final var consensusEnd = Instant.ofEpochSecond(0, blockFile.getConsensusEnd());
+            streamLatencyMetric
+                    .tag("node", blockFile.getNode())
```
steven-sheehy: For blockfilesource, we won't have a node so this will be null. Instead, we should default to something like "cloud" or "bucket".

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockStreamVerifier.java
bilyana-gospodinova: Removed

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockStreamVerifier.java
bilyana-gospodinova: Addressed

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockStreamVerifier.java
bilyana-gospodinova: Changed

importer/src/main/java/org/hiero/mirror/importer/reader/block/BlockStreamReaderImpl.java
bilyana-gospodinova: Good point. I think I was able to resolve this. Please re-review.

importer/src/test/java/org/hiero/mirror/importer/downloader/block/BlockNodeTest.java
bilyana-gospodinova: Moved

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockStreamVerifier.java
bilyana-gospodinova: Addressed

importer/src/test/java/org/hiero/mirror/importer/downloader/block/BlockNodeTest.java
```diff
@@ -60,6 +60,7 @@ final class BlockNodeTest extends BlockNodeTestBase {
     private BlockNodeProperties blockNodeProperties;
     private BlockNode node;
     private StreamProperties streamProperties;
+    private final String ERROR_METRIC_NAME = "hiero.mirror.importer.stream.error";
```
steven-sheehy: Can store as a package private static var in `BlockNode` to stay DRY.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/AbstractBlockSource.java
```diff
@@ -36,12 +36,17 @@ public final void get() {
...
+            final BlockStream blockStream, final BlockNodeProperties blockNodeProperties) {
+        var blockFile = readBlockFile(blockStream);
+        blockFile.setNode(blockNodeProperties.getEndpoint());
+        blockStreamVerifier.verify(blockFile);
+        return blockFile;
+    }
 
+    protected final BlockFile onBlockStream(final BlockStream blockStream) {
+        var blockFile = readBlockFile(blockStream);
+        blockFile.setNode("cloud");
```
steven-sheehy: Store `"cloud"` as a static var.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/AbstractBlockSource.java
```diff
@@ -36,12 +36,17 @@ public final void get() {
...
-        }
+    protected final BlockFile onBlockStream(
+            final BlockStream blockStream, final BlockNodeProperties blockNodeProperties) {
+        var blockFile = readBlockFile(blockStream);
+        blockFile.setNode(blockNodeProperties.getEndpoint());
+        blockStreamVerifier.verify(blockFile);
+        return blockFile;
+    }
 
+    protected final BlockFile onBlockStream(final BlockStream blockStream) {
```
steven-sheehy: Now there's two `onBlockStream` methods where each is only used by one sub-class. Instead, we should have a single `onBlockStream` where the implementation varies for each sub-class and only `BlockFileSource` has an implementation that sets `cloud`.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockNode.java
```diff
@@ -70,6 +79,11 @@ final class BlockNode implements AutoCloseable, Comparable<BlockNode> {
         this.grpcBufferDisposer = grpcBufferDisposer;
         this.properties = properties;
         this.streamProperties = streamProperties;
+        this.errorsMetric = Counter.builder("hiero.mirror.importer.stream.error")
+                .description("The number of errors that occurred while streaming from a particular block node.")
+                .tag("type", StreamType.BLOCK.toString())
+                .tag("node", properties.getEndpoint())
```
jnels124: nit: maybe `block_node` is a bit more clear as `node` is a highly overloaded term.

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockStreamVerifier.java
```diff
@@ -79,6 +86,10 @@ public void verify(BlockFile blockFile) {
         try {
             verifyBlockNumber(blockFile);
             verifyHashChain(blockFile);
+            final var consensusEnd = Instant.ofEpochSecond(0, blockFile.getConsensusEnd());
+            streamLatencyMeterProvider
+                    .withTag("node", blockFile.getNode())
```
jnels124: nit: same comment as above

importer/src/main/java/org/hiero/mirror/importer/downloader/block/AbstractBlockSource.java
bilyana-gospodinova: I refactored the code to have only one `onBlockStream` method in the parent class and the block node endpoint as a second parameter. This way we pass "cloud" in `BlockFileSource` and pass the configured endpoint from the properties in `BlockNodeSubscriber`. I think it is better this way instead of passing `null` for the block node properties. What do you think?

importer/src/test/java/org/hiero/mirror/importer/downloader/block/BlockNodeTest.java
bilyana-gospodinova: Addressed

importer/src/main/java/org/hiero/mirror/importer/downloader/block/AbstractBlockSource.java
bilyana-gospodinova: Addressed

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockNode.java
bilyana-gospodinova: Sure, renamed

importer/src/main/java/org/hiero/mirror/importer/downloader/block/BlockStreamVerifier.java
bilyana-gospodinova: Renamed

xin-hedera (approved):
LGTM

PR #12670, opened 2025-12-23
Prepare mainnet-na for 0.145

jnels124:
prepare mainnet-na for 0.145

### DIFF
--- a/clusters/mainnet-na/common/helmrelease.yaml
+++ b/clusters/mainnet-na/common/helmrelease.yaml
@@ -13,7 +13,7 @@ spec:
       valuesFiles:
         - values.yaml
         - values-v2.yaml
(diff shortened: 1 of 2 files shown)


xin-hedera (approved):
LGTM

clusters/mainnet-na/common/helmrelease.yaml
```diff
@@ -13,7 +13,7 @@ spec:
       valuesFiles:
         - values.yaml
         - values-v2.yaml
-      version: '0.144.0'
```
xin-hedera, line 16: 0.145.2?

jnels124: oops yes was just double checking things and you are right! Will fix

PR #12675, opened 2025-12-23
Add auto grow for zfs volumes

### LINKED ISSUE #8338: Auto Resize Citus Volumes
### Problem

We need a way to auto size disks for shards/coordinator. Currently, we have to manually modify a config map and restart the zfs init pods. 

### Solution

We should investigate the use of a volume resizer such as  [this one](https://github.com/DevOps-Nirvana/Kubernetes-Volume-Autoscaler/tree/master). 

- [ ] Supports expanding backing zfs disk 
- [ ] Supports auto expanding pvc
  
Issues:
How do we update relevant flux config? 

### Alternatives

Alter zfs daemonset to monitor relevant alerts in prometheus. Use alerts to update disk and pvc.

jnels124:
**Description**:
this pr adds functionality to the common chart to auto grow zfs datasets
* move init container to main container
* add configurable interval to check for resize operations
* automatically grow disk + zpool + pvc

**Related issue(s)**:

Fixes #8338 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/charts/hedera-mirror-common/templates/zfs/configmap-init.yaml
+++ b/charts/hedera-mirror-common/templates/zfs/configmap-init.yaml
@@ -64,13 +64,16 @@ data:
       if ! chroot "${ROOT_MOUNT_DIR}" /bin/bash -c "
         zfs list || true
         partprobe || true
-        zpool online -e \"${POOL}\" /dev/sdb && zfs list || true
+
+        zpool online -e \"${POOL}\" /dev/sdb
+        rc=\$?
+
+        zfs list || true
+        exit \$rc
       "; then
         log "Failed to expand zpool ${POOL} on ${NODE_NAME}"
         return 1
       fi
-
-      return 0
     }
 
     function forceSizeScanWithRetry() {


charts/hedera-mirror-common/values.yaml
```diff
@@ -668,6 +668,11 @@ traefik:
 zfs:
   alloy:
     enabled: false
+  autoResize:
+    growPct: 10
```
steven-sheehy:
```suggestion
    growPercent: 10
```

charts/hedera-mirror-common/templates/zfs/configmap-init.yaml
```diff
@@ -34,233 +33,789 @@ data:
...
+
+    function forceSizeScan() {
+      log "Expanding pool to disk size"
+
+      # Only the zpool expand is "required" for success; the others are best-effort.
+      if ! chroot "${ROOT_MOUNT_DIR}" /bin/bash -c "
+        zfs list || true
+        partprobe || true
+        zpool online -e \"${POOL}\" /dev/sdb
+        zfs list
```
xin-hedera:
```suggestion
        zpool online -e \"${POOL}\" /dev/sdb && zfs list
```

if `zfs list` isn't required,

```suggestion
        zpool online -e \"${POOL}\" /dev/sdb
```

charts/hedera-mirror-common/templates/zfs/configmap-init.yaml
```diff
@@ -34,233 +33,789 @@ data:
...
-      if (kubectl patch configmap $CONFIG_MAP_NAME -n $K8S_NAMESPACE --type merge -p "${MODIFIED_CONFIG_MAP}")
-      then
-        echo "Successfully patched config map"
-        COMMAND_STATUS=0
+    function setGlobals() {
+      log "Getting node metadata"
+      NODE_NAME="$(curl -sS http://metadata.google.internal/computeMetadata/v1/instance/hostname -H 'Metadata-Flavor: Google' | awk -F'.' '{print $1}')"
+      ZONE="$(curl -sS http://metadata.google.internal/computeMetadata/v1/instance/zone -H 'Metadata-Flavor: Google' | awk -F  "/" '{print $NF}')"
+      NODE_LABELS="$(kubectl get node "${NODE_NAME}" -o jsonpath='{.metadata.labels}')"
+      CITUS_ROLE="$(echo "${NODE_LABELS}" | jq --arg ROLE_LABEL "${ROLE_LABEL}" -r '.[$ROLE_LABEL]')"
```
xin-hedera: `setGlobals` is called before `bootstrap`. in bootstrap there's logic to wait for role label. Here it gets the label directly in one try and sets `DISK_SIZE` for coordinator / worker.

why we have to wait for role label in a later step not here?

charts/hedera-mirror-common/templates/zfs/configmap-init.yaml
```diff
@@ -34,233 +33,789 @@ data:
...
         else
-          echo "Successfully removed L2 ARC device ${ATTACHED_CACHE_DEVICE}"
+          # overrides missing: add empty array and then add new entry
+          jq -n --arg rv "${rv}" --arg size "${targetSizeGiStr}" --argjson idx "${idx}" '
+            [
+              {"op":"test","path":"/metadata/resourceVersion","value":$rv},
+              {"op":"add","path":"/spec/shards/overrides","value":[]},
+              {"op":"add","path":"/spec/shards/overrides/-","value":{"index":$idx,"pods":{"persistentVolume":{"size":$size}}}}
+            ]
+          '
```
xin-hedera, line 590: nit, it might be simpler to

1. use jq to create the new `overrides` list given there may be no `overrides` at all / there exists / doesn't exist the override for the particular cluster
2. do `kubectl patch --type=merge` to replaces the whole `overrides`

docs/runbook/increase-zfs-disksize.md
```diff
@@ -44,7 +44,7 @@ The pvc for a shard is running out of space and needs to be increased beyond cur
    ```
 3. Restart the zfs init pods
    ```bash
-   kubectl rollout restart daemonset -n common mirror-zfs-init
+   kubectl rollout restart daemonset -n common mirror-zfs-manager
```
xin-hedera, line 47: there are still mentionings of `zfs init` in the doc.

tools/cluster-management/reduce-citus-disk-size.sh
```diff
@@ -52,7 +52,7 @@ function configureAndValidate() {
 
   DISK_PREFIX="$(kubectl get daemonsets \
   -n "${COMMON_NAMESPACE}" \
-  -l 'app=zfs-init' \
+  -l 'app=zfs-manager' \
   -o json |
   jq -r '.items[0].spec.template.spec.initContainers[0].env[] | select (.name == "DISK_PREFIX") | .value')"
```
jnels124: will fix jq path here

charts/hedera-mirror-common/templates/zfs/configmap-init.yaml
jnels124: The original implementation had this incorrect as well and when things were broken out it carried over. I corrected it to wait for the label in set globals instead.

charts/hedera-mirror-common/templates/zfs/configmap-init.yaml
jnels124: Was running into collisions that way because the resource version wasn't being respected unless we pass the whole sgsharded cluster yaml. There could be different nodes updating thier pvcs at the same time and this approach seemed more clear. Will look at this again

charts/hedera-mirror-common/templates/zfs/configmap-init.yaml
```diff
@@ -34,233 +33,794 @@ data:
...
+    }
+
+    function forceSizeScan() {
+      log "Expanding pool to disk size"
+
+      # Only the zpool expand is "required" for success; the others are best-effort.
+      if ! chroot "${ROOT_MOUNT_DIR}" /bin/bash -c "
+        zfs list || true
+        partprobe || true
+        zpool online -e \"${POOL}\" /dev/sdb && zfs list || true
```
xin-hedera: adding `|| true` will make it always successful? Then the error detection is just dead code.

jnels124: Yes but it will only be successful if the zpool online is successful. And that is the only thing we care about error wise

xin-hedera: umm, I don't think so, because `false && false || true` is `true`

jnels124: You are right. I mis remembered how that worked. I will fix it

PR #12700, opened 2026-01-05
Read trailing non-transaction statechanges in an event

### LINKED ISSUE #12697: HIP-1056 Importer fails to properly read non-transactional trailing statechanges in events
### Description

There appears to be non-transactional trailing statechanges in events as well. An edge case is an event with nothing but such non-transactional statechanges at the end.

This has caused BlockStreamReader to throw an `Missing block footer in block` exception since such statechanges are treated as block-wide statechanges immediately preceding a block proof.

### Steps to reproduce

Run importer to read block 66104759 in testnet

### Additional context

_No response_

### Hedera network

testnet

### Version

v0.146.0-SNAPSHOT

### Operating system

None

xin-hedera:
**Description**:

This PR fixes the bug that trailing non-transaction statechanges in an event causes importer to throw exception

- Skip non-transaction statechanges block items

**Related issue(s)**:

Fixes #12697 

**Notes for reviewer**:

**Checklist**


common/src/main/java/org/hiero/mirror/common/domain/transaction/BlockTransaction.java
```diff
@@ -49,11 +49,13 @@
 @AllArgsConstructor(access = AccessLevel.NONE)
 @CustomLog
 @SuppressWarnings("deprecation")
+@ToString(onlyExplicitlyIncluded = true)
```
xin-hedera, line 52: this addresses a debugging issue that IDEA stuck on rendering `BlockTransaction` in the Variables panel, likely due to some member variables in `BlockTransaction` causing circular references and thus causing infinite string representation.

jnels124 (approved):

PR #12706, opened 2026-01-05
HIP-1195 Transformed blockstream hook storage to ContractStateChanges sidecar record

### LINKED ISSUE #12578: HIP-1195 Transform blockstream hook storage to ContractStateChanges sidecar record
### Problem

Special handling is needed to transform hook storage changes in a transactional unit from blockstream to recordstream `ContractStateChanges` sidecar record especially when there are hook executions from multiple hooks, due to the fact that all hook storage slots are externalized as if belonging to the hook system contract id 0x16d and thus the same (0x16d, key) storage slot under different hook executions pointing to different storage slots.

### Solution

Extend the contract storage transformation logic in `AbstractBlockTransactionTransformer` to handle multiple hook executions in a single transactional unit properly. Details TBD. Note execution team has added a field `proto.HookId executed_hook_id = 7` to `EvmTransactionResult` to avoid the need of alias / evm address account to numerical account id lookup.

### Alternatives

_No response_

ashumahajan:
**Description**:
Extend the contract storage transformation logic in AbstractBlockTransactionTransformer to transform executed_hook_id from EvmTransactionResult. 

- Child Transaction (Contract Call) adds executed_hook_id from EvmTransactionResult to the RecordItem in AbstractBlockTransactionTransformer
- in ContractResultServiceImpl the hook_id is fetched for ContractStateChanges sidecar record from the reordItem

Fixes #12578 

**Checklist**

### DIFF
--- a/common/src/main/java/org/hiero/mirror/common/domain/transaction/StateChangeContext.java
+++ b/common/src/main/java/org/hiero/mirror/common/domain/transaction/StateChangeContext.java
@@ -9,7 +9,6 @@ import com.google.protobuf.BytesValue;
 import com.hedera.hapi.block.stream.output.protoc.MapUpdateChange;
 import com.hedera.hapi.block.stream.output.protoc.StateChanges;
 import com.hedera.hapi.block.stream.output.protoc.StateIdentifier;
-import com.hedera.hapi.node.state.hooks.legacy.LambdaSlotKey;
 import com.hederahashgraph.api.proto.java.Account;
 import com.hederahashgraph.api.proto.java.AccountID;
 import com.hederahashgraph.api.proto.java.ContractID;
@@ -51,7 +50,6 @@ public final class StateChangeContext {
     private final Map<TokenID, Long> tokenTotalSupplies = new HashMap<>();
     private final List<TopicID> topicIds = new ArrayList<>();
     private final Map<TopicID, TopicMessage> topicState = new HashMap<>();
-    private final Map<LambdaSlotKey, BytesValue> lambdaStorageChanges = new HashMap<>();
 
     private StateChangeContext() {}
 


importer/src/main/java/org/hiero/mirror/importer/downloader/block/transformer/AbstractBlockTransactionTransformer.java
```diff
@@ -138,6 +139,25 @@ private void transformSmartContractResult(BlockTransactionTransformation transfo
         }
 
         var evmTransactionResult = evmTransactionInfo.evmTransactionResult();
+
+        if (evmTransactionResult.hasExecutedHookId()) {
```
xin-hedera: record stream parser doesn't need `(owner, hookId)` since it's either explicitly info in the top level crypto transfer transaction or in case the owner is an alias / evm address, parser is capable of resolving it to the numerical entity id.

All the changes should be reverted.

ashumahajan: I made this change for BlockStream and not RecordStream. Are you saying that RecordStream parser will take care of hooks from parent crypto transfer for blockstream too. In that case we don't need to make any change?

common/src/main/java/org/hiero/mirror/common/domain/transaction/BlockTransaction.java
```diff
@@ -79,14 +71,14 @@ public class BlockTransaction implements StreamItem {
...
 
-    @EqualsAndHashCode.Exclude
-    @Getter(lazy = true)
-    private final StateChangeContext stateChangeContext = createStateChangeContext();
-
     private final boolean successful;
     private final List<TraceData> traceData;
     private final TransactionBody transactionBody;
 
+    @EqualsAndHashCode.Exclude
```
xin-hedera, line 82: looks like IDEA reordered these? can you revert them and set IDEA not to do the reordering?

common/src/main/java/org/hiero/mirror/common/domain/transaction/StateChangeContext.java
```diff
@@ -77,6 +78,7 @@ private StateChangeContext() {}
                             processPendingAirdropStateChange(mapUpdate);
                         case StateIdentifier.STATE_ID_TOKENS_VALUE -> processTokenStateChange(mapUpdate);
                         case StateIdentifier.STATE_ID_TOPICS_VALUE -> processTopicStateChange(mapUpdate);
+                        case StateIdentifier.STATE_ID_LAMBDA_STORAGE_VALUE -> processLambdaStorageChange(mapUpdate);
```
xin-hedera, line 78: need to handle lambda storage MapDelete statechanges, check the logic in the else branch for contract storage slot MapDelete statechanges

common/src/main/java/org/hiero/mirror/common/domain/transaction/BlockTransaction.java
```diff
@@ -182,6 +194,18 @@ public BytesValue getValueWritten(final SlotKey slotKey) {
...
+    /**
+     * Get value written to the storage slot at {@link LambdaSlotKey} by the transaction. Returns the first read value
+     * in subsequent transactions, or the value in statechanges.
+     *
+     * @param lambdaSlotKey - The hook storage's slot key
+     * @return The value written
+     */
+    public BytesValue getValueWritten(final LambdaSlotKey lambdaSlotKey) {
+        // fall back to statechanges
+        return getStateChangeContext().getLambdaStorageValueWritten(lambdaSlotKey);
```
xin-hedera: this doesn't handle the scenario with multiple hook executions from a single crypto transfer transaction

- hook execution with hook `X`, updates storage slot `S`
- hook execution with hook `X` again, also updates storage slot `S`

The value written to `S` by the first hook execution is the value read by the second hook execution.

common/src/main/java/org/hiero/mirror/common/domain/transaction/StateChangeContext.java
ashumahajan: Done

common/src/main/java/org/hiero/mirror/common/domain/transaction/BlockTransaction.java
ashumahajan: Done

importer/src/main/java/org/hiero/mirror/importer/downloader/block/transformer/AbstractBlockTransactionTransformer.java
ashumahajan: done

common/src/main/java/org/hiero/mirror/common/domain/transaction/BlockTransaction.java
ashumahajan: Done

common/src/main/java/org/hiero/mirror/common/domain/transaction/StateChangeContext.java
```diff
@@ -51,13 +52,13 @@ public final class StateChangeContext {
     private final Map<TokenID, Long> tokenTotalSupplies = new HashMap<>();
     private final List<TopicID> topicIds = new ArrayList<>();
     private final Map<TopicID, TopicMessage> topicState = new HashMap<>();
+    private final Map<LambdaSlotKey, BytesValue> lambdaStorageChanges = new HashMap<>();
```
xin-hedera: we should unify contract storage and lambda storage here and in `BlockTransaction` instead of duplicating. After all, the logic to parse the related statechanges and evm tracedata is the same.

We can have a `ContractSlotKey` class as the key of the map.

```java
record ContractSlotKey(ContractID contractId, HookId hookId, byte[] key) {}
```

For hook storage, we can set the contractId to `ContractID.newBuilder().setContractNum(365).build()` regardless of the network shard / realm.

For regular contract storage, the slot key would simply have a `null` hookId

common/src/main/java/org/hiero/mirror/common/domain/transaction/BlockTransaction.java
```diff
@@ -182,6 +187,25 @@ public BytesValue getValueWritten(final SlotKey slotKey) {
...
+    /**
+     * Get value written to the storage slot at {@link LambdaSlotKey} by the transaction. Returns the first read value
+     * in subsequent transactions, or the value in statechanges.
+     *
+     * @param lambdaSlotKey - The hook storage's slot key
+     * @return The value written
+     */
+    public BytesValue getValueWritten(final LambdaSlotKey lambdaSlotKey) {
+        final var normalizedSlotKey = normalize(lambdaSlotKey);
+        for (var nextInner = nextInBatch; nextInner != null; nextInner = nextInner.getNextInBatch()) {
```
xin-hedera: the logic in the getter for regular contract storage is specific for inner batch transactions. it wouldn't work for child hook execution contract call transactions triggered by a crypto transfer, i.e., `nextInner` in such a scenario will always be `null`.

xin-hedera (changes requested):

common/src/main/java/org/hiero/mirror/common/domain/transaction/StateChangeContext.java
ashumahajan: Done

common/src/main/java/org/hiero/mirror/common/domain/transaction/ContractSlotKey.java
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Unified key for both contract storage and lambda (hook) storage.
+ * For regular contract storage: contractId is set, hookId is null
+ * For lambda/hook storage: contractId is null, hookId is set
+ */
+@Builder(toBuilder = true)
+@AllArgsConstructor(access = PRIVATE)
+@CustomLog
+@ToString(onlyExplicitlyIncluded = true)
```
steven-sheehy, line 25: Since none are included this will probably print nothing. Think you can remove annotation entirely since `@Value` has it then add `ToString.Exclude` to key.

common/src/main/java/org/hiero/mirror/common/domain/transaction/ContractSlotKey.java
```diff
@@ -0,0 +1,32 @@
...
+@EqualsAndHashCode
```
steven-sheehy, line 26: Can remove since `@Value` has it.

common/src/main/java/org/hiero/mirror/common/domain/transaction/ContractSlotKey.java
```diff
@@ -0,0 +1,32 @@
...
+import lombok.Value;
+
+/**
+ * Unified key for both contract storage and lambda (hook) storage.
+ * For regular contract storage: contractId is set, hookId is null
+ * For lambda/hook storage: contractId is null, hookId is set
+ */
+@Builder(toBuilder = true)
+@AllArgsConstructor(access = PRIVATE)
+@CustomLog
```
steven-sheehy, line 24: Not used. Can remove

common/src/main/java/org/hiero/mirror/common/domain/transaction/ContractSlotKey.java
```diff
@@ -0,0 +1,32 @@
...
+import lombok.ToString;
+import lombok.Value;
+
+/**
+ * Unified key for both contract storage and lambda (hook) storage.
+ * For regular contract storage: contractId is set, hookId is null
+ * For lambda/hook storage: contractId is null, hookId is set
+ */
+@Builder(toBuilder = true)
+@AllArgsConstructor(access = PRIVATE)
```
steven-sheehy, line 23: We might remove these two to avoid extra objects on heap since this only has 3 fields and is created quite frequently.

common/src/main/java/org/hiero/mirror/common/domain/transaction/StateChangeContext.java
```diff
@@ -240,17 +267,36 @@ private void processContractStorageChange(MapUpdateChange mapUpdate) {
         }
 
         var slotKey = mapUpdate.getKey().getSlotKeyKey();
+        final var contractSlotKey = ContractSlotKey.builder()
+                .contractId(slotKey.getContractID())
+                .key(slotKey.getKey())
```
steven-sheehy, line 272: Shouldn't there be a hook id set?

common/src/main/java/org/hiero/mirror/common/domain/transaction/StateChangeContext.java
```diff
@@ -77,15 +75,44 @@ private StateChangeContext() {}
...
+                                        .getKey()
+                                        .getSlotKeyKey()
+                                        .getContractID())
+                                .build();
+                        // use the default BytesValue instance when the storage slot is deleted
+                        processContractStorageChange(contractSlotKey, BytesValue.getDefaultInstance());
+                    }
+                    if (stateChange.getMapDelete().getKey().hasLambdaSlotKey()) {
+                        final var contractSlotKey = ContractSlotKey.builder()
+                                .key(stateChange
```
steven-sheehy, line 102: Don't we need the contract id?

common/src/main/java/org/hiero/mirror/common/domain/transaction/StateChangeContext.java
```diff
@@ -77,15 +75,44 @@ private StateChangeContext() {}
...
                     }
-                } else if (stateChange.hasMapDelete()
-                        && stateChange.getMapDelete().getKey().hasSlotKeyKey()) {
-                    var slotKey = stateChange.getMapDelete().getKey().getSlotKeyKey();
-                    // use the default BytesValue instance when the storage slot is deleted
-                    processContractStorageChange(slotKey, BytesValue.getDefaultInstance());
+                } else if (stateChange.hasMapDelete()) {
+                    if (stateChange.getMapDelete().getKey().hasSlotKeyKey()) {
+                        final var contractSlotKey = ContractSlotKey.builder()
+                                .key(stateChange
```
steven-sheehy, line 86: Need hook id?

common/src/main/java/org/hiero/mirror/common/domain/transaction/StateChangeContext.java
```diff
@@ -240,17 +267,36 @@ private void processContractStorageChange(MapUpdateChange mapUpdate) {
...
+
+    private void processLambdaStorageChange(MapUpdateChange mapUpdate) {
+        if (!mapUpdate.getKey().hasLambdaSlotKey()) {
+            return;
+        }
         var valueWritten = mapUpdate.getValue().getSlotValueValue().getValue();
-        processContractStorageChange(slotKey, BytesValue.of(valueWritten));
+        final var lambdaSlotKey = mapUpdate.getKey().getLambdaSlotKey();
+        final var contractSlotKey = ContractSlotKey.builder()
+                .hookId(lambdaSlotKey.getHookId())
```
steven-sheehy, line 285: contract id?

steven-sheehy:
DCO issue

common/src/main/java/org/hiero/mirror/common/domain/transaction/StateChangeContext.java
ashumahajan: This is for standard contract call, which won't have hookId

common/src/main/java/org/hiero/mirror/common/domain/transaction/StateChangeContext.java
ashumahajan: As per Xin's suggestion, I combined both SlotKey and LambdaSlotKey into one object. Differentiating factor is that for hook execution, contract_id is set to null (since it is always system contract) and for regular contract call, hookId is marked as null

ashumahajan:
Creating another PR as the commit history got corrupted while fixing DCO

PR #12708, opened 2026-01-06
Add opcode tracing tests for native precompiles

IvanKavaldzhiev:
**Description**:
This PR enhances web3 with more tests related to calling besu native precompiles and the tracing opcode functionality used in eth_debugTraceTransaction.

It also fixes a issue related to saving and caching gasRequirement value in the `ContractCallContext` inside `CustomMessageCallProcessor`. This causes issue when we have mixed calls to HTS system contract and besu native precompiles by setting incorrect `gasCost` value when a native precompile is called after a system contract.

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/web3/src/main/java/com/hedera/node/app/service/contract/impl/exec/processors/CustomMessageCallProcessor.java
+++ b/web3/src/main/java/com/hedera/node/app/service/contract/impl/exec/processors/CustomMessageCallProcessor.java
@@ -164,7 +164,6 @@ public class CustomMessageCallProcessor extends MessageCallProcessor {
 
             if (evmPrecompile == null) {
                 handleNonExtantSystemAccount(frame, tracer);
-
                 return;
             }
         }
@@ -236,6 +235,7 @@ public class CustomMessageCallProcessor extends MessageCallProcessor {
             @NonNull final MessageFrame frame, @NonNull final OperationTracer tracer) {
         final PrecompileContractResult result = PrecompileContractResult.success(Bytes.EMPTY);
         frame.clearGasRemaining();
+        ContractCallContext.get().setGasRequirement(0L);
         finishPrecompileExecution(frame, result, PRECOMPILE, (ActionSidecarContentTracer) tracer);
     }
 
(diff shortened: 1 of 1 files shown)


web3/src/test/java/org/hiero/mirror/web3/service/ContractCallServicePrecompileModificationTest.java
```diff
@@ -591,6 +591,36 @@ void burnNFT() throws Exception {
...
+        // When
+        final var functionCall = contract.call_mintTokenNativePrecompileAndBurnExternal(
+                toAddress(tokenId).toHexString(),
+                BigInteger.valueOf(mintAmount),
+                BigInteger.valueOf(burnAmount),
+                new ArrayList<>());
+        final var result = functionCall.send();
+
+        // Then
+        assertThat(result.component1()).isEqualTo(BigInteger.valueOf(22)); // SUCCESS response code
```
bilyana-gospodinova: Maybe use the `ResponseCodeEnum.SUCCESS` constant.

bilyana-gospodinova (dismissed):

web3/src/main/java/com/hedera/node/app/service/contract/impl/exec/processors/CustomMessageCallProcessor.java
```diff
@@ -331,6 +331,7 @@ private void finishPrecompileExecution(
         frame.setState(result.getState());
         frame.setExceptionalHaltReason(result.getHaltReason());
         tracer.tracePrecompileCall(frame, ContractCallContext.get().getGasRequirement(), result.getOutput());
+        ContractCallContext.get().setGasRequirement(0L);
```
steven-sheehy: We're passing a potentially non-zero gasRequirement to tracer then clearing it in the next line. Since this gasRequirement is only used by this class we don't really need to clear it, we just need to ensure we set it in every code path (precompile & system contract). Right now, we're not setting the gasRequirement in doExecutePrecompile(). Why do we not set it there?

IvanKavaldzhiev: You're correct. I now set the gasRequirement in all paths which invoke `finishPrecompileExecution`

PR #12710, opened 2026-01-06
Simplify state initialization

### LINKED ISSUE #11428: Refactor mirror node state initialization
### Problem

Historically, when bumping hedera-app a lot of code changes were needed in the mirror node in order for the code to compile. This was mainly because there are some subcomponents implemented on the mirror node side that mimic the state initialization in the consensus node. These components implement interfaces that change frequently. The following components are:
- org.hiero.mirror.web3.state.components.NetworkInfoImpl
- org.hiero.mirror.web3.state.components.SchemaRegistryImpl
- org.hiero.mirror.web3.state.components.ServiceMigratorImpl
- org.hiero.mirror.web3.state.components.ServicesRegistryImpl
- org.hiero.mirror.web3.state.components.StartupNetworksImpl

### Solution

We need to try to remove these components and write our own simplified implementation that builds the state. This way on hedera-app bump we expect to have less breaking changes.

This task is on hold until mainnet is 100% modularized.

### Alternatives

_No response_

bilyana-gospodinova:
**Description**:
This PR refactors the state initialization by removing the service registration and the schema migrations. Now the readable KV states and singletons are added directly.
The release comparator is enhanced to check for newly added service schemas, which could potentially add new states and flag in the CI.

Fixes #11428

### DIFF
--- a/web3/src/main/java/org/hiero/mirror/web3/state/MirrorNodeState.java
+++ b/web3/src/main/java/org/hiero/mirror/web3/state/MirrorNodeState.java
@@ -22,10 +22,10 @@ import com.swirlds.state.spi.ReadableKVState;
 import com.swirlds.state.spi.ReadableStates;
 import com.swirlds.state.spi.WritableStates;
 import jakarta.inject.Named;
-import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.LinkedList;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Queue;
@@ -53,12 +53,8 @@ public class MirrorNodeState implements State {
     // Key is Service, value is Map of state name to state datasource
     private final Map<String, Map<Integer, Object>> states = new HashMap<>();
 
-    private final Collection<AbstractReadableKVState<?, ?>> readableKVStates;
-
     public MirrorNodeState(
-            final Collection<SingletonState<?>> singletonStates,
-            final Collection<AbstractReadableKVState<?, ?>> readableKVStates) {
-        this.readableKVStates = readableKVStates;
+            final List<SingletonState<?>> singletonStates, final List<AbstractReadableKVState<?, ?>> readableKVStates) {
         initSingletonStates(singletonStates);
         initKVStates(readableKVStates);
         initQueueStates();
@@ -146,26 +142,23 @@ public class MirrorNodeState implements State {
         return Collections.unmodifiableMap(states);
     }
 
-    private void initSingletonStates(final Collection<SingletonState<?>> singletonStates) {
+    private void initSingletonStates(final List<SingletonState<?>> singletonStates) {
+        singletonStates.add(new DefaultSingleton(TokenService.NAME, STAKING_NETWORK_REWARDS_STATE_ID));
+        singletonStates.add(new DefaultSingleton(TokenService.NAME, NODE_REWARDS_STATE_ID));
         singletonStates.forEach(
                 singletonState -> states.computeIfAbsent(singletonState.getServiceName(), k -> new HashMap<>())
                         .put(singletonState.getStateId(), singletonState));
-        Map<Integer, String> defaultSingletonImplementations = new HashMap<>(Map.of(
-                STAKING_NETWORK_REWARDS_STATE_ID, TokenService.NAME,
-                NODE_REWARDS_STATE_ID, TokenService.NAME));
-        defaultSingletonImplementations.forEach((stateId, serviceName) ->
-                states.computeIfAbsent(serviceName, k -> new HashMap<>()).put(stateId, new DefaultSingleton(stateId)));
     }
 
-    private void initKVStates(final Collection<AbstractReadableKVState<?, ?>> readableKVStates) {
+    private void initKVStates(final List<AbstractReadableKVState<?, ?>> readableKVStates) {
         readableKVStates.forEach(kvState -> states.computeIfAbsent(kvState.getServiceName(), k -> new HashMap<>())
                 .put(kvState.getStateId(), kvState));
-        Map<Integer, String> defaultKvImplementations = new HashMap<>(Map.of(
+        final var defaultKvImplementations = Map.of(
                 LAMBDA_STORAGE_STATE_ID, ContractService.NAME,
                 EVM_HOOK_STATES_STATE_ID, ContractService.NAME,
                 SCHEDULE_ID_BY_EQUALITY_STATE_ID, ScheduleService.NAME,
                 SCHEDULED_COUNTS_STATE_ID, ScheduleService.NAME,
-                SCHEDULED_USAGES_STATE_ID, ScheduleService.NAME));
+                SCHEDULED_USAGES_STATE_ID, ScheduleService.NAME);
         defaultKvImplementations.forEach(
                 (stateId, serviceName) -> states.computeIfAbsent(serviceName, k -> new HashMap<>())
                         .put(stateId, createMapReadableStateForId(serviceName, stateId)));
--- a/web3/src/main/java/org/hiero/mirror/web3/state/singleton/DefaultSingleton.java
+++ b/web3/src/main/java/org/hiero/mirror/web3/state/singleton/DefaultSingleton.java
@@ -7,6 +7,8 @@ import lombok.RequiredArgsConstructor;
 
 @RequiredArgsConstructor
 public class DefaultSingleton extends AtomicReference<Object> implements SingletonState<Object> {
+
+    private final String serviceName;
     private final int id;
 
     @Override
@@ -16,6 +18,6 @@ public class DefaultSingleton extends AtomicReference<Object> implements Singlet
 
     @Override
     public String getServiceName() {
-        return "";
+        return serviceName;
     }
 }
--- a/web3/src/test/java/org/hiero/mirror/web3/state/MirrorNodeStateIntegrationTest.java
+++ b/web3/src/test/java/org/hiero/mirror/web3/state/MirrorNodeStateIntegrationTest.java
@@ -208,12 +208,13 @@ final class MirrorNodeStateIntegrationTest extends Web3IntegrationTest {
             serviceStates.removeAll(statesNotNeeded);
             if (!serviceStates.isEmpty()) {
                 missingStates.addAll(serviceStates);
-                System.out.println("Missing states " + serviceStates + " for service " + service);
                 serviceStates.clear();
             }
             implementedStates.clear();
         }
-        assertThat(missingStates).isEmpty();
+        assertThat(missingStates)
+                .withFailMessage("Missing state ids that need to be implemented: " + missingStates)
+                .isEmpty();
     }
 
     private void verifyServiceDataSources(
--- a/web3/src/test/java/org/hiero/mirror/web3/state/MirrorNodeStateTest.java
+++ b/web3/src/test/java/org/hiero/mirror/web3/state/MirrorNodeStateTest.java
@@ -16,26 +16,15 @@ import com.swirlds.state.spi.ReadableSingletonState;
 import com.swirlds.state.spi.WritableKVState;
 import com.swirlds.state.spi.WritableQueueState;
 import com.swirlds.state.spi.WritableSingletonState;
-import java.util.Collection;
 import java.util.LinkedList;
-import org.hiero.mirror.web3.state.singleton.SingletonState;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
-import org.mockito.InjectMocks;
-import org.mockito.Spy;
 import org.mockito.junit.jupiter.MockitoExtension;
 
 @ExtendWith(MockitoExtension.class)
 final class MirrorNodeStateTest {
 
-    @InjectMocks
-    private MirrorNodeState mirrorNodeState;
-
-    @Spy
-    private Collection<ReadableKVState<?, ?>> readableKVStates = new LinkedList<>();
-
-    @Spy
-    private Collection<SingletonState<?>> singletonStates = new LinkedList<>();
+    private final MirrorNodeState mirrorNodeState = new MirrorNodeState(new LinkedList<>(), new LinkedList<>());
 
     @Test
     void getReadableStatesWithSingleton() {


.github/workflows/releases-comparator.yml
```diff
@@ -99,6 +99,20 @@ jobs:
...
+            "com/hedera/node/app/records/schemas"
+            "com/hedera/node/app/blocks/schemas"
+            "com/hedera/node/app/throttle/schemas"
+            "com/hedera/node/app/service/contract/impl/schemas"
+            "com/hedera/node/app/service/entityid/impl/schemas"
+            "com/hedera/node/app/fees/schemas"
+            "com/hedera/node/app/service/file/impl/schemas"
+            "com/hedera/node/app/state/recordcache/schemas"
+            "com/hedera/node/app/service/schedule/impl/schemas"
+            "com/hedera/node/app/service/token/impl/schemas"
```
steven-sheehy: Please sort

web3/src/main/java/org/hiero/mirror/web3/state/MirrorNodeState.java
```diff
@@ -74,49 +98,42 @@ public class MirrorNodeState implements State {
...
-    private final MirrorNodeEvmProperties mirrorNodeEvmProperties;
+    // Singletons
+    private final BlockInfoSingleton blockInfoSingleton;
+    private final BlockStreamInfoSingleton blockStreamInfoSingleton;
+    private final CongestionLevelStartsSingleton congestionLevelStartsSingleton;
+    private final EntityIdSingleton entityIdSingleton;
+    private final EntityCountsSingleton entityCountsSingleton;
+    private final MidnightRatesSingleton midnightRatesSingleton;
+    private final RunningHashesSingleton runningHashesSingleton;
+    private final ThrottleUsageSingleton throttleUsageSingleton;
```
steven-sheehy: Take advantage of Spring's component scanning to find all singletons and inject them as a collection. Then register each singleton generically using `singletonState.getId()`. This reduces coupling between components and allows us to not have to update the `MirrorNodeState` when adding a new singleton.

```suggestion
    private final Collection<SingletonState> singletonStates;
```

web3/src/main/java/org/hiero/mirror/web3/state/MirrorNodeState.java
```diff
@@ -74,49 +98,42 @@ public class MirrorNodeState implements State {
...
+    private final AccountReadableKVState accountReadableKVState;
+    private final AirdropsReadableKVState airdropsReadableKVState;
+    private final AliasesReadableKVState aliasesReadableKVState;
+    private final ContractBytecodeReadableKVState contractBytecodeReadableKVState;
+    private final ContractStorageReadableKVState contractStorageReadableKVState;
+    private final FileReadableKVState fileReadableKVState;
+    private final NftReadableKVState nftReadableKVState;
+    private final TokenReadableKVState tokenReadableKVState;
+    private final ScheduleReadableKVState scheduleReadableKVState;
+    private final TokenRelationshipReadableKVState tokenRelationshipReadableKVState;
```
steven-sheehy:
```suggestion
    private final Collection<ReadableKVState<?, ?>> readableKVStates;
```

web3/src/main/java/org/hiero/mirror/web3/state/MirrorNodeState.java
```diff
@@ -2,66 +2,90 @@
...
+import org.hiero.mirror.web3.state.singleton.EntityCountsSingleton;
+import org.hiero.mirror.web3.state.singleton.EntityIdSingleton;
+import org.hiero.mirror.web3.state.singleton.MidnightRatesSingleton;
+import org.hiero.mirror.web3.state.singleton.RunningHashesSingleton;
 import org.hiero.mirror.web3.state.singleton.SingletonState;
+import org.hiero.mirror.web3.state.singleton.ThrottleUsageSingleton;
 import org.jspecify.annotations.NonNull;
-import org.jspecify.annotations.Nullable;
 
 @SuppressWarnings({"rawtypes", "unchecked"})
```
steven-sheehy: Are these still needed?

web3/src/main/java/org/hiero/mirror/web3/state/MirrorNodeState.java
```diff
@@ -74,49 +98,42 @@ public class MirrorNodeState implements State {
...
-            0L,
-            asAccount(commonProperties.getShard(), commonProperties.getRealm(), 3L),
-            10L,
-            List.of(),
-            Bytes.EMPTY,
-            List.of(),
-            true,
-            Bytes.EMPTY);
-    private static final Metrics NO_OP_METRICS = new NoOpMetrics();
+    private final List<ReadableKVState> readableKVStates;
```
steven-sheehy: Unused. Can be leveraged with my above suggestion though prefer higher level `Collection`.

web3/src/main/java/org/hiero/mirror/web3/state/MirrorNodeState.java
```diff
@@ -74,49 +98,42 @@ public class MirrorNodeState implements State {
...
 
     @PostConstruct
     private void init() {
```
steven-sheehy: We should remove this method and move logic to constructor.

web3/src/main/java/org/hiero/mirror/web3/state/MirrorNodeState.java
```diff
@@ -204,55 +226,88 @@ Map<String, Map<Integer, Object>> getStates() {
...
-                        new ContractServiceImpl(appContext, NO_OP_METRICS),
-                        new BlockRecordService(),
-                        new FeeService(),
-                        new CongestionThrottleService(),
-                        new RecordCacheService(),
-                        new ScheduleServiceImpl(appContext),
-                        new BlockStreamService())
-                .forEach(servicesRegistry::register);
+    private void registerBlockRecordServiceStates() {
+        final var blockRecordServiceStates = new ConcurrentHashMap<Integer, Object>();
```
steven-sheehy: Do all this service maps need to be `ConcurrentHashMap` if they're just created once on startup and read only afterwards? Instead we should use `Map.of()` or some other immutable, non-concurrent map.

web3/src/main/java/org/hiero/mirror/web3/state/MirrorNodeState.java
```diff
@@ -74,49 +98,42 @@ public class MirrorNodeState implements State {
     // Key is Service, value is Map of state name to state datasource
     private final Map<String, Map<Integer, Object>> states = new ConcurrentHashMap<>();
```
steven-sheehy: Probably this doesn't need to be concurrent if we create inside constructor as an immutable map.

web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/StateRegistry.java
```diff
@@ -1,95 +0,0 @@
...
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
-import org.hiero.mirror.web3.state.core.MapReadableKVState;
-import org.hiero.mirror.web3.state.singleton.DefaultSingleton;
-import org.hiero.mirror.web3.state.singleton.SingletonState;
-import org.jspecify.annotations.NonNull;
-
-@Named
-public final class StateRegistry {
-    private static final Set<String> DEFAULT_IMPLEMENTATIONS = Stream.of(
```
steven-sheehy, line 30: This class (or some test using it) had the advantage of detecting new state classes in newer releases that we do not have implementations for. We have to retain that logic somewhere. Preferably in code that introspects library states.

web3/src/main/java/org/hiero/mirror/web3/state/singleton/BlockStreamInfoSingleton.java
```diff
@@ -8,7 +8,7 @@
 import jakarta.inject.Named;
 
 @Named
-final class BlockStreamInfoSingleton implements SingletonState<BlockStreamInfo> {
+public final class BlockStreamInfoSingleton implements SingletonState<BlockStreamInfo> {
```
steven-sheehy: We can undo with my suggestion to scan for all.

web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/ScheduleReadableKVState.java
```diff
@@ -34,7 +34,7 @@
 import org.jspecify.annotations.NonNull;
 
 @Named
-class ScheduleReadableKVState extends AbstractReadableKVState<ScheduleID, Schedule> {
+public class ScheduleReadableKVState extends AbstractReadableKVState<ScheduleID, Schedule> {
```
steven-sheehy: We can undo with my suggestion to scan for all.

web3/src/main/java/org/hiero/mirror/web3/state/MirrorNodeState.java
```diff
@@ -204,55 +226,88 @@ Map<String, Map<Integer, Object>> getStates() {
...
+        tokenServiceStates.put(TOKEN_RELS_STATE_ID, tokenRelationshipReadableKVState);
+        tokenServiceStates.put(AIRDROPS_STATE_ID, airdropsReadableKVState);
+        tokenServiceStates.put(
+                STAKING_NETWORK_REWARDS_STATE_ID, new DefaultSingleton(STAKING_NETWORK_REWARDS_STATE_ID));
+        tokenServiceStates.put(NODE_REWARDS_STATE_ID, new DefaultSingleton(NODE_REWARDS_STATE_ID));
+        states.put(TokenService.NAME, tokenServiceStates);
+    }
+
+    private MapReadableKVState createMapReadableStateForId(final String serviceName, int id) {
+        return new MapReadableKVState<>(serviceName, id, new HashMap<>());
```
steven-sheehy, line 172: This was ConcurrentHashMap in StateRegistry previously

web3/src/main/java/org/hiero/mirror/web3/state/MirrorNodeState.java
```diff
@@ -204,55 +226,88 @@ Map<String, Map<Integer, Object>> getStates() {
...
+        final var tokenServiceStates = new ConcurrentHashMap<Integer, Object>();
+        tokenServiceStates.put(ACCOUNTS_STATE_ID, accountReadableKVState);
+        tokenServiceStates.put(ALIASES_STATE_ID, aliasesReadableKVState);
+        tokenServiceStates.put(TOKENS_STATE_ID, tokenReadableKVState);
+        tokenServiceStates.put(NFTS_STATE_ID, nftReadableKVState);
+        tokenServiceStates.put(TOKEN_RELS_STATE_ID, tokenRelationshipReadableKVState);
+        tokenServiceStates.put(AIRDROPS_STATE_ID, airdropsReadableKVState);
+        tokenServiceStates.put(
+                STAKING_NETWORK_REWARDS_STATE_ID, new DefaultSingleton(STAKING_NETWORK_REWARDS_STATE_ID));
+        tokenServiceStates.put(NODE_REWARDS_STATE_ID, new DefaultSingleton(NODE_REWARDS_STATE_ID));
```
steven-sheehy: Preferably we could have a set of default singletons somewhere that we can auto-register them using like we used to have in StateRegistry.DEFAULT_IMPLEMENTATIONS.

web3/src/main/java/org/hiero/mirror/web3/state/MirrorNodeState.java
```diff
@@ -204,55 +226,88 @@ Map<String, Map<Integer, Object>> getStates() {
...
+        states.put(TokenService.NAME, tokenServiceStates);
```
steven-sheehy: Is there nothing in hedera-app that provides a mapping of NAME to IDs we can leverage to avoid our manual mapping? We can get from Schema, which we can get from service, but neither of these we have easy access to. This might be an enhancement request for upstream.

Meanwhile, we could add a new interface to both kv state and singleton state that returns the name so we can do the grouping generically:

```java
public interface RegisterableState {

    Integer getId();

    String getServiceName();
}
```

.github/workflows/releases-comparator.yml
```diff
@@ -99,6 +99,20 @@ jobs:
             "com/hedera/node/app/service/contract/impl/exec/operations/CustomBalanceOperation.java"
           )
 
+          # Compare the schemas - in case there is a new version we might need to register a new state.
+          DIRS_TO_COMPARE=(
+            "com/hedera/node/app/records/schemas"
```
steven-sheehy: What if there is a new schema in a new folder? Or they rename the package? This is too brittle. We should have a programmatic way in JUnit to detect this scenario.

bilyana-gospodinova: - If there is a new service, this will be a big change and it will probably be defined with a new HIP and we already review the HIPs and this will be covered in the designs.
- If they rename some package, we will get a compile error in the `MirrorNodeStateIntegrationTest` as in there we use all of the state keys when we verify the state is constructed correctly.
- For a programmatic way we will probably need to keep the `ServicesRegistryImpl` and the `SchemaRegistryImpl` or discuss upstream for a direct way to get all states directly.

.github/workflows/releases-comparator.yml
bilyana-gospodinova: Sorted

web3/src/main/java/org/hiero/mirror/web3/state/MirrorNodeState.java
bilyana-gospodinova: Addressed

web3/src/main/java/org/hiero/mirror/web3/state/MirrorNodeState.java
bilyana-gospodinova: Addressed

web3/src/main/java/org/hiero/mirror/web3/state/MirrorNodeState.java
bilyana-gospodinova: Removed

web3/src/main/java/org/hiero/mirror/web3/state/MirrorNodeState.java
bilyana-gospodinova: Addressed

web3/src/main/java/org/hiero/mirror/web3/state/MirrorNodeState.java
bilyana-gospodinova: Moved

web3/src/main/java/org/hiero/mirror/web3/state/singleton/BlockStreamInfoSingleton.java
bilyana-gospodinova: Reverted

web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/ScheduleReadableKVState.java
bilyana-gospodinova: Reverted

web3/src/main/java/org/hiero/mirror/web3/state/MirrorNodeState.java
bilyana-gospodinova: Updated

web3/src/main/java/org/hiero/mirror/web3/state/MirrorNodeState.java
bilyana-gospodinova: Added the interface for now

web3/src/main/java/org/hiero/mirror/web3/state/MirrorNodeState.java
bilyana-gospodinova: Addressed

web3/src/main/java/org/hiero/mirror/web3/state/MirrorNodeState.java
bilyana-gospodinova: Addressed

web3/src/main/java/org/hiero/mirror/web3/state/MirrorNodeState.java
bilyana-gospodinova: They don't need to be. Addressed.

web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/StateRegistry.java
bilyana-gospodinova: I added `MirrorNodeStateIntegrationTest#searchForMissingStates()`. In the release comparator we search for new schemas and it will flag if there is a new one and we need to add it in the `MirrorNodeStateIntegrationTest`. Then in the test we compare the all of the added states and all of the defined states in all of the schemas. The test will fail if there is a new state.

web3/src/test/java/org/hiero/mirror/web3/state/MirrorNodeStateIntegrationTest.java
```diff
@@ -2,76 +2,113 @@
...
 
 @RequiredArgsConstructor
-public class MirrorNodeStateIntegrationTest extends Web3IntegrationTest {
+final class MirrorNodeStateIntegrationTest extends Web3IntegrationTest {
 
     private final MirrorNodeState mirrorNodeState;
-    private final ServicesRegistry servicesRegistry;
-    private final MirrorNodeEvmProperties mirrorNodeEvmProperties;
+
+    // In the releases-comparator.yml there is compare for newly added schemas.
```
IvanKavaldzhiev: nit: there is a comparison

IvanKavaldzhiev (dismissed):
LGTM

web3/src/test/java/org/hiero/mirror/web3/state/MirrorNodeStateIntegrationTest.java
bilyana-gospodinova: Addressed

web3/src/main/java/org/hiero/mirror/web3/state/MirrorNodeState.java
```diff
@@ -131,8 +76,8 @@ public ReadableStates getReadableStates(@NonNull String serviceName) {
             for (final var entry : serviceStates.entrySet()) {
                 final var stateId = entry.getKey();
                 final var state = entry.getValue();
-                if (state instanceof Queue queue) {
-                    data.put(stateId, new ListReadableQueueState(serviceName, stateId, queue));
+                if (state instanceof Queue<?> queue) {
+                    data.put(stateId, new ListReadableQueueState<>(serviceName, stateId, queue));
                 } else if (state instanceof ReadableKVState<?, ?> kvState) {
                     final var readableKVState = readableKVStates.stream()
                             .filter(r -> r.getStateId() == stateId)
```
steven-sheehy: Is this logic here and right below (`if (readableKVState.isPresent())`) still necessary? Shouldn't the `states` already be fully populated with the `readableKVStates`?

web3/src/test/java/org/hiero/mirror/web3/state/MirrorNodeStateIntegrationTest.java
```diff
@@ -80,123 +117,103 @@ void verifyServicesHaveAssignedDataSources() {
...
+                serviceStates.addAll(schema.statesToCreate().stream()
+                        .map(StateDefinition::stateId)
+                        .collect(Collectors.toSet()));
+                serviceStates.removeAll(schema.statesToRemove());
+            }
+            serviceStates.removeAll(implementedStates);
+            serviceStates.removeAll(statesNotNeeded);
+            if (!serviceStates.isEmpty()) {
+                missingStates.addAll(serviceStates);
+                System.out.printf("Missing states %s for service %s.%n", serviceStates, service);
```
steven-sheehy: Should remove printf

web3/src/main/java/org/hiero/mirror/web3/state/singleton/RunningHashesSingleton.java
```diff
@@ -5,18 +5,25 @@
...
 import com.hedera.hapi.node.state.blockrecords.RunningHashes;
+import com.hedera.node.app.records.BlockRecordService;
 import com.hedera.pbj.runtime.io.buffer.Bytes;
 import jakarta.inject.Named;
 import org.hiero.mirror.web3.common.ContractCallContext;
+import org.hiero.mirror.web3.state.RegisterableState;
 
 @Named
-public class RunningHashesSingleton implements SingletonState<RunningHashes> {
+final class RunningHashesSingleton implements SingletonState<RunningHashes>, RegisterableState {
```
steven-sheehy: `SingletonState` already implements `RegisterableState` so each concrete state does not also need to directly implement `RegisterableState`. Same comment for other `SingletonState` classes.

web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AccountReadableKVState.java
```diff
@@ -110,4 +111,14 @@ private Optional<Account> getDummySystemAccountIfApplicable(AccountID accountID)
         }
         return Optional.empty();
     }
+
+    @Override
+    public Integer getId() {
+        return STATE_ID;
+    }
```
steven-sheehy: Rename in `RegisterableState` to `int getStateId()` then you can remove in every KV state sub-class since it already exists.

web3/src/main/java/org/hiero/mirror/web3/state/MirrorNodeState.java
steven-sheehy: Still HashMap? Maybe you've verified it's safe to be not concurrent?

web3/src/main/java/org/hiero/mirror/web3/state/MirrorNodeState.java
steven-sheehy: Let's research and open a ticket upstream for a possible enhancement to get that info.

web3/src/main/java/org/hiero/mirror/web3/state/MirrorNodeState.java
bilyana-gospodinova: Indeed, removed.

web3/src/main/java/org/hiero/mirror/web3/state/singleton/RunningHashesSingleton.java
bilyana-gospodinova: Right, reverted.

web3/src/main/java/org/hiero/mirror/web3/state/keyvalue/AccountReadableKVState.java
bilyana-gospodinova: Addressed

web3/src/main/java/org/hiero/mirror/web3/state/MirrorNodeState.java
bilyana-gospodinova: This map is used only for `LAMBDA_STORAGE_STATE_ID`, `EVM_HOOK_STATES_STATE_ID`, `SCHEDULE_ID_BY_EQUALITY_STATE_ID`, `SCHEDULED_COUNTS_STATE_ID`, `SCHEDULED_USAGES_STATE_ID`. We define a default readable KV state just so the state can be initialized. We don't use the maps in production. Even if we did and I missed that, the map is wrapped in `MapReadableKVState`, which extends `ReadableKVStateBase` and in there we use the scoped per transaction cache so there should be no contamination of the state between the transactions. Same for the writable states.

web3/src/main/java/org/hiero/mirror/web3/state/MirrorNodeState.java
bilyana-gospodinova: I opened a ticket - https://github.com/hiero-ledger/hiero-consensus-node/issues/22873.

web3/src/test/java/org/hiero/mirror/web3/state/MirrorNodeStateIntegrationTest.java
bilyana-gospodinova: Removed

web3/src/main/java/org/hiero/mirror/web3/state/MirrorNodeState.java
```diff
@@ -2,121 +2,66 @@
...
     private final Map<String, WritableStates> writableStates = new ConcurrentHashMap<>();
 
     // Key is Service, value is Map of state name to state datasource
-    private final Map<String, Map<Integer, Object>> states = new ConcurrentHashMap<>();
+    private final Map<String, Map<Integer, Object>> states = new HashMap<>();
 
-    private final List<ReadableKVState> readableKVStates;
-    private final ServicesRegistry servicesRegistry;
-    private final MirrorNodeEvmProperties mirrorNodeEvmProperties;
+    private final Collection<AbstractReadableKVState<?, ?>> readableKVStates;
```
steven-sheehy: Now unused

web3/src/test/java/org/hiero/mirror/web3/state/MirrorNodeStateIntegrationTest.java
steven-sheehy: You didn't remove it, just changed it to println. We should not use System.out anywhere nor should we log warnings in tests. Instead convert to a test assertion if needed so junit prints out any problems.

web3/src/main/java/org/hiero/mirror/web3/state/singleton/DefaultSingleton.java
```diff
@@ -10,7 +10,12 @@ public class DefaultSingleton extends AtomicReference<Object> implements Singlet
...
 
     @Override
-    public Integer getId() {
+    public int getStateId() {
         return id;
     }
+
+    @Override
+    public String getServiceName() {
+        return "";
```
steven-sheehy: It'd be simpler if we changed the constructor to accept the id and name so we can return the name here.

web3/src/main/java/org/hiero/mirror/web3/state/MirrorNodeState.java
```diff
@@ -204,55 +146,36 @@ Map<String, Map<Integer, Object>> getStates() {
...
-                .forEach(servicesRegistry::register);
+    private void initSingletonStates(final Collection<SingletonState<?>> singletonStates) {
+        singletonStates.forEach(
+                singletonState -> states.computeIfAbsent(singletonState.getServiceName(), k -> new HashMap<>())
+                        .put(singletonState.getStateId(), singletonState));
+        Map<Integer, String> defaultSingletonImplementations = new HashMap<>(Map.of(
+                STAKING_NETWORK_REWARDS_STATE_ID, TokenService.NAME,
+                NODE_REWARDS_STATE_ID, TokenService.NAME));
+        defaultSingletonImplementations.forEach((stateId, serviceName) ->
+                states.computeIfAbsent(serviceName, k -> new HashMap<>()).put(stateId, new DefaultSingleton(stateId)));
```
steven-sheehy: Can be simplified:

```suggestion
        singletonStates.add(new DefaultSingleton(TokenService.NAME, STAKING_NETWORK_REWARDS_STATE_ID));
        singletonStates.add(new DefaultSingleton(TokenService.NAME, NODE_REWARDS_STATE_ID));
        singletonStates.forEach(
                s -> states.computeIfAbsent(s.getServiceName(), k -> new HashMap<>()).put(s.getStateId(), s));
```

web3/src/main/java/org/hiero/mirror/web3/state/MirrorNodeState.java
```diff
@@ -204,55 +146,36 @@ Map<String, Map<Integer, Object>> getStates() {
...
-                    @NonNull Bytes bytes,
-                    SignatureVerifier.@NonNull MessageType messageType,
-                    @NonNull SignatureMap signatureMap,
-                    @Nullable Function<Key, SimpleKeyStatus> simpleKeyVerifier) {
-                throw new UnsupportedOperationException("Not implemented");
-            }
+    private void initKVStates(final Collection<AbstractReadableKVState<?, ?>> readableKVStates) {
+        readableKVStates.forEach(kvState -> states.computeIfAbsent(kvState.getServiceName(), k -> new HashMap<>())
+                .put(kvState.getStateId(), kvState));
+        Map<Integer, String> defaultKvImplementations = new HashMap<>(Map.of(
```
steven-sheehy: Unnecessary wrapper map.

```suggestion
        final var defaultKvImplementations = Map.of(
```

steven-sheehy (changes requested):

web3/src/test/java/org/hiero/mirror/web3/state/MirrorNodeStateIntegrationTest.java
bilyana-gospodinova: I see. I though we just prefer to use `System.out.println` instead of `System.out.printf`. I removed this line and moved it into the assertThat's fail message.

web3/src/main/java/org/hiero/mirror/web3/state/MirrorNodeState.java
bilyana-gospodinova: Removed

web3/src/main/java/org/hiero/mirror/web3/state/singleton/DefaultSingleton.java
bilyana-gospodinova: Addressed

web3/src/main/java/org/hiero/mirror/web3/state/MirrorNodeState.java
bilyana-gospodinova: Addressed

web3/src/main/java/org/hiero/mirror/web3/state/MirrorNodeState.java
bilyana-gospodinova: Addressed

steven-sheehy (approved):
LGTM

PR #12713, opened 2026-01-06
Optimize get supply query

### LINKED ISSUE #12632: Enhance performance for REST java /api/v1/network/supply
### Problem

K6 performance test results for this endpoint are lower than the JS API version. 

### Solution

1.  Review implementation and analyze query plan to determine cause for slower performance

### Alternatives

_No response_

sdimitrov9:
**Description**:
Multiple runs of k6 tests were executed and average values were taken. 
Config: 
...
DEFAULT_DURATION=60s \
DEFAULT_LIMIT=100 \
DEFAULT_VUS=150 \
...

Results:
main : 196.1k completed requests
feature : 261k completed requests

Based on the above results the new query seems to improve the network supply endpoint performance with 32.8%

**Related issue(s)**:

Fixes #12632

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/common/src/test/java/org/hiero/mirror/common/config/CommonTestConfiguration.java
+++ b/common/src/test/java/org/hiero/mirror/common/config/CommonTestConfiguration.java
@@ -91,7 +91,7 @@ public class CommonTestConfiguration {
 
     @Bean(POSTGRESQL)
     PostgreSQLContainer<?> postgresql() {
-        var imageName = v2 ? "gcr.io/mirrornode/citus:12.1.1" : "postgres:16-alpine";
+        var imageName = v2 ? "gcr.io/mirrornode/citus:12.1.1" : "gcr.io/mirrornode/postgres:16-alpine";
         var dockerImageName = DockerImageName.parse(imageName).asCompatibleSubstituteFor("postgres");
         var logger = LoggerFactory.getLogger(PostgreSQLContainer.class);
         var excluded = "terminating connection due to unexpected postmaster exit";
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -17,7 +17,7 @@ services:
       POSTGRES_HOST_AUTH_METHOD: scram-sha-256
       POSTGRES_INITDB_ARGS: --auth-host=scram-sha-256
       POSTGRES_PASSWORD: postgres_password
-    image: postgres:16-alpine
+    image: gcr.io/mirrornode/postgres:16-alpine
     ports:
       - 5432:5432
     restart: unless-stopped
--- a/rest/__tests__/globalSetup.js
+++ b/rest/__tests__/globalSetup.js
@@ -15,7 +15,7 @@ const FLYWAY_VERSION = '11.8.2';
 const DB_NAME = 'mirror_node';
 const OWNER_USER = 'mirror_node';
 const DATABASE_IMAGES = {
-  v1: 'postgres:16-alpine',
+  v1: 'gcr.io/mirrornode/postgres:16-alpine',
   v2: 'gcr.io/mirrornode/citus:12.1.1',
 };
 
--- a/rosetta/test/db/db.go
+++ b/rosetta/test/db/db.go
@@ -5,7 +5,6 @@ package db
 import (
 	"database/sql"
 	"fmt"
-	"github.com/ory/dockertest/v3"
 	"os"
 	"path"
 	"path/filepath"
@@ -13,6 +12,8 @@ import (
 	"strconv"
 	"strings"
 	"time"
+
+	"github.com/ory/dockertest/v3"
 
 	"github.com/hiero-ledger/hiero-mirror-node/rosetta/app/config"
 	"github.com/hiero-ledger/hiero-mirror-node/rosetta/app/interfaces"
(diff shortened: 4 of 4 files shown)


rest-java/src/main/java/org/hiero/mirror/restjava/repository/EntityRepository.java
```diff
@@ -18,10 +17,18 @@ public interface EntityRepository extends CrudRepository<Entity, Long> {
...
         from entity
-        where id in (:unreleasedSupplyAccounts)
+        where
+            id = 2
+            or id = 42
+            or id between 44 and 71
+            or id between 73 and 87
+            or id between 99 and 100
+            or id between 200 and 349
+            or id between 400 and 750
```
steven-sheehy: Problem is this does not take into account shard and realm. We could use `num` instead of `id` but then the performance would be horrible.

rest-java/src/main/java/org/hiero/mirror/restjava/repository/AccountBalanceRepository.java
```diff
@@ -11,16 +10,24 @@
...
+        from account_balance
+        where consensus_timestamp between :lowerTimestamp and :upperTimestamp
+          and (
+                account_id = 2
+             or account_id = 42
+             or account_id between 44 and 71
+             or account_id between 73 and 87
+             or account_id between 99 and 100
+             or account_id between 200 and 349
+             or account_id between 400 and 750
```
steven-sheehy: Doesn't work with non-zero shard and realm. Instead, I think we'll have to use a Spel expression like https://www.baeldung.com/spring-data-query-definitions-spel#adding-a-spel-context. Something like `and :#{supply.accounts}`.

steven-sheehy (changes requested):

rest-java/src/main/java/org/hiero/mirror/restjava/repository/EntityRepository.java
sdimitrov9: Support for non-zero shadow and realm is now added, while the performance improvement is kept.

rest-java/src/main/java/org/hiero/mirror/restjava/repository/AccountBalanceRepository.java
sdimitrov9: Support for non-zero shadow and realm is now added, while the performance improvement is kept.

rest-java/src/main/java/org/hiero/mirror/restjava/config/NetworkProperties.java
```diff
@@ -32,6 +33,9 @@ public class NetworkProperties {
     @Getter(lazy = true)
     private final Set<Long> unreleasedSupplyAccountIds = createUnreleasedSupplyAccountIds();
```
steven-sheehy: This can be removed now, right?

rest-java/src/test/java/org/hiero/mirror/restjava/repository/AccountBalanceRepositoryTest.java
```diff
@@ -31,26 +30,24 @@ void getSupplyHistory() {
         createAccountBalance(account2, 2_500_000L, timestamp2);
         createAccountBalance(account1, 3_000_000L, timestamp3);
 
-        final var accountIds = List.of(account1.getId(), account2.getId());
-
         // when / then
-        assertThat(accountBalanceRepository.getSupplyHistory(accountIds, timestamp1, timestamp2))
+        assertThat(accountBalanceRepository.getSupplyHistory("2", "42", timestamp1, timestamp2))
```
steven-sheehy: Don't think the changes in this file will work with non-zero realm. Please set CommonProperties to non-zero shard and realm and re-run and fix.

rest-java/src/test/java/org/hiero/mirror/restjava/repository/EntityRepositoryTest.java
```diff
@@ -57,13 +56,12 @@ void findById() {
...
-        final var account2 = createEntityWithBalance(42L, 2_000_000L, timestamp);
-        final var account3 = createEntityWithBalance(100L, 500_000L, timestamp);
-        final var accountIds = List.of(account1.getId(), account2.getId(), account3.getId());
+        createEntityWithBalance(2L, 1_000_000L, timestamp);
+        createEntityWithBalance(42L, 2_000_000L, timestamp);
+        createEntityWithBalance(100L, 500_000L, timestamp);
 
         // when
-        final var result = entityRepository.getSupply(accountIds);
+        final var result = entityRepository.getSupply("2", "100");
```
steven-sheehy: Test with NZR

steven-sheehy (changes requested):

steven-sheehy:
Can merge from main to get passing snyk workflow

rest-java/src/main/java/org/hiero/mirror/restjava/config/NetworkProperties.java
sdimitrov9: Still used by some tests, so it is now moved to test class.

PR #12750, opened 2026-01-12
Adding support for multi-party fungible token transfer synthetic events

### LINKED ISSUE #12633: Support synthetic multi-party fungible token transfer events
### Problem

As an EVM user, I'd like to consume multi-party (more than 2 parties involved) contract events for all fungible token transfers regardless if they occur via the EVM or regular HAPI transactions.

### Solution

* Feature flag behind `hiero.mirror.importer.parser.record.entity.persist.syntheticContractLogsMulti=true`
* Update `EntityRecordItemListener.insertFungibleTokenTransfers()` to not just store single transfer synthetic events.
* Use transaction body of successful transactions to generate events since they are more likely to be paired and not aggregated.
* Come up with an algorithm that pairs transfers in body based upon matching credits and debits.
* If transfers don't cleanly line up then like for odd numbered parties split out extra transfers using a deterministic algorithm that works regardless of order of transfers in list.
  * `[A=1000, B=-400, C=-600]` => `[[A=400, B=-400], [A=600, C=-600]]`
  * `[A=-1000, B=400, C=600]` => `[[A=-400, B=400], [A=-600, C=600]]`
  * `[A=1000, C=-400, B=-600]` => `[[A=600, B=-600], [A=400, C=-400]]`
* Add back any assessed custom fee synthetic events.
* Update docs/importer/README.md

### Alternatives

* Consensus nodes generate them
* Relay reconstructs them from existing REST APIs
* User generates them from REST APIs

IvanKavaldzhiev:
**Description**:
Adding support for synthetic events related to multi-party fungible token transfers. The following logic is applied:

For one recipient and multiple senders, the events are split into sub-transfers, where the recipients are sorted by AccountID.

`[A=1000, B=-400, C=-600] => [[C=-600, A=600], [C=-400, B=400]]`

For corresponding transfers, the events are created for the directly pairing transfers.

`[A=400, B=-400, C=300, D=-300] => [[B=-400, A=400], [D=-300, C=300]]`

In the cases where we have the same amount for all transfers, we sort both sender and recipients first by their amount in DESC order and then by their AccountID in ASC order, so that we achieve deterministic behaviour.

`[A=400, B=-400, C=400, D=-400] => [[B=-400, A=400], [D=-400, C=400]]`

For more complex cases, with multiple senders, recipients and not matching transfer pairs, we split again the transfers, not only sorting the accounts, but firstly exhausting the transfer amount from the earliest sender possible. After their amount is completely used, we continue using the next sender to create a new transfer pair.

`[A=1000, B=-400, C=-600, D=500, E=400, F=-800, G=-100] => [[F=-800, A=800], [C=-200, A=200], [C=-400, D=400], [B=-100, D=100], [B=-300, E=E00], [G=-100, E=100]]`

A dynamic property for enabling and disabling synthetic event creation for multi-party transfers is added.

After this PR, the importer will back-fill any missing fungible token transfer related events from record/block stream. In case of a mixed case where there is an incoming synthetic event from single transfer but no events for multi-party transfer, the mirror node will populate the events for the multi-party transfer. This logic is valid for any missing fungible transfer related missing events from CN.

In case a SyntheticContractLog has an EVM origin a proper bloom filter will be created and the consensus timestamp of the log will be set to the parent recordItem with existing contractResult.

**Related issue(s)**:

Fixes #12633

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
+++ b/common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
@@ -341,12 +341,12 @@ public class RecordItem implements StreamItem {
 
         private List<ContractLoginfo> parseContractLogs() {
             if (transactionRecord.hasContractCallResult()) {
-                return new ArrayList<>(transactionRecord.getContractCallResult().getLogInfoList().stream()
-                        .toList());
+                return new ArrayList<>(transactionRecord.getContractCallResult().getLogInfoList());
             } else if (transactionRecord.hasContractCreateResult()) {
-                return new ArrayList<>(transactionRecord.getContractCreateResult().getLogInfoList().stream()
-                        .toList());
+                return new ArrayList<>(
+                        transactionRecord.getContractCreateResult().getLogInfoList());
             }
+
             return Collections.emptyList();
         }
 


importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/EntityRecordItemListener.java
```diff
@@ -472,14 +476,113 @@ private void logTokenTransfers(
...
+                    senders.add(transfer);
+                } else if (transfer.getAmount() > 0) {
+                    receivers.add(transfer);
+                }
+            }
+
+            final var transferPairs = createTransferPairs(senders, receivers);
+
+            for (final var pair : transferPairs) {
+                final var first = pair.get(0);
```
bilyana-gospodinova: Doesn't the `createTransferPairs()` method always return the sender at the first place and the receiver on the second? If so, the whole loop can be simplified.

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/EntityProperties.java
```diff
@@ -72,6 +72,8 @@ public static class PersistProperties {
 
         private boolean syntheticContractLogEvmAddressLookup = true;
 
+        private boolean syntheticContractLogsMulti = true;
```
steven-sheehy, line 75: Need to document

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/EntityRecordItemListener.java
```diff
@@ -472,14 +476,113 @@ private void logTokenTransfers(
             EntityId accountId,
             long amount) {
         if (isSingleTransfer && amount > 0) {
-            EntityId senderId = i == 0
+            final var senderId = i == 0
                     ? EntityId.of(tokenTransfers.get(1).getAccountID())
                     : EntityId.of(tokenTransfers.get(0).getAccountID());
             syntheticContractLogService.create(
                     new TransferContractLog(recordItem, tokenId, senderId, accountId, amount));
+        } else if (entityProperties.getPersist().isSyntheticContractLogsMulti() && recordItem.isSuccessful()) {
```
steven-sheehy: Quite a lot of inline logic. Think we should consider moving to a dedicated class to isolate testing and reduce complexity.

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/EntityRecordItemListener.java
```diff
@@ -421,7 +423,9 @@ private void insertFungibleTokenTransfers(RecordItem recordItem, TokenTransferLi
 
             logTokenEvents(recordItem, tokenId, isWipeOrBurn, isMint, accountId, amount);
 
-            logTokenTransfers(recordItem, tokenId, tokenTransfers, isSingleTransfer, i, accountId, amount);
+            if (isSingleTransfer || (isMultipleTransfer && i == 0)) {
```
steven-sheehy: Code smell for unnecessary parentheses

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/EntityRecordItemListener.java
jnels124: This may introduce some performance issues if there are a large amount of transfers. 

Appears this is:
* copy + sort tokenTransfers
* loop to split senders/receivers
* build an intermediate List<List<AccountAmount>> of pairs (lots of tiny objects)
* loop again to turn those pairs into logs

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/EntityRecordItemListener.java
```diff
@@ -472,14 +476,113 @@ private void logTokenTransfers(
...
         if (isSingleTransfer && amount > 0) {
-            EntityId senderId = i == 0
+            final var senderId = i == 0
                     ? EntityId.of(tokenTransfers.get(1).getAccountID())
                     : EntityId.of(tokenTransfers.get(0).getAccountID());
             syntheticContractLogService.create(
                     new TransferContractLog(recordItem, tokenId, senderId, accountId, amount));
+        } else if (entityProperties.getPersist().isSyntheticContractLogsMulti() && recordItem.isSuccessful()) {
+            final var sortedTransfers = new ArrayList<>(tokenTransfers);
+            sortedTransfers.sort(Comparator.comparing(aa -> EntityId.of(aa.getAccountID()), EntityId::compareTo));
```
jnels124: This builds entity id for comparisons. Couldn't we just compare the long value of accountId?

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/EntityRecordItemListener.java
jnels124: Instead of using create pairs method, couldn't we just have 2 pointers (one for the receiver, one for sender) and emit logs as we go? This will reduce extra passes over the data. Overall we would do something like

```
  if NOT (multiLogsEnabled AND recordItem.successful):
    return

  #  Build entries once (and compute account identity once)
  entries = [] # list of lightweight java records
  for each t in tokenTransfers:
    if t.amount != 0:
      entries.add( (accountIdOf(t), t.amount) )

  if entries.size < 2:
    return

  # Sort by accountId
  sort entries by accountId num (instead of comparing EntityId.of()

  # Split into senders and receivers (still sorted by accountId)
  senders = []   # amounts negative
  receivers = [] # amounts positive
  for each e in compact:
    if e.amount < 0: senders.add(e)
    else: receivers.add(e)

  # Two-pointers to pair transfers and emit logs directly
  s = 0
  r = 0
  senderRemaining = abs(senders[s].amount)  
  receiverRemaining = receivers[r].amount 

  while s < senders.size AND r < receivers.size:
    senderId = senders[s].accountId
    receiverId = receivers[r].accountId
    xfer = min(senderRemaining, receiverRemaining)
    emitLog(recordItem, tokenId, senderId, receiverId, xfer)

    senderRemaining -= xfer
    receiverRemaining -= xfer

    if senderRemaining == 0:
      s += 1
      if s < senders.size:
        senderRemaining = abs(senders[s].amount)

    if receiverRemaining == 0:
      r += 1
      if r < receivers.size:
        receiverRemaining = receivers[r].amount

  # sanity check:
   if s < senders.size OR r < receivers.size:
  #   transfers did not net to zero (unexpected); log recoverable error maybe?

```

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/EntityRecordItemListener.java
IvanKavaldzhiev: The whole algorithm was simplified.

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/EntityRecordItemListener.java
IvanKavaldzhiev: I have updated and simplified the logic in a very similar way to what you have suggested.

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/EntityRecordItemListener.java
IvanKavaldzhiev: I'm now comparing directly the Long values from the protobuf itself.

IvanKavaldzhiev: Also decided to use TreeMap structure, so that the loop needed for sorting is avoided.

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/EntityRecordItemListener.java
```diff
@@ -472,14 +477,76 @@ private void logTokenTransfers(
...
             long amount) {
         if (isSingleTransfer && amount > 0) {
-            EntityId senderId = i == 0
+            final var senderId = i == 0
                     ? EntityId.of(tokenTransfers.get(1).getAccountID())
                     : EntityId.of(tokenTransfers.get(0).getAccountID());
             syntheticContractLogService.create(
-                    new TransferContractLog(recordItem, tokenId, senderId, accountId, amount));
+                    new TransferContractLog(recordItem, tokenId, senderId, accountId, amount, false));
+        } else if (entityProperties.getPersist().isSyntheticContractLogsMulti()) {
```
bilyana-gospodinova: In the PR description this case is mentioned:

> For corresponding transfers, the events are created for the directly pairing transfers.
> [A=400, B=-400, C=300, D=-300] => [[A=400, B=-400], [C=300, D=-300]]

However, I don't see this logic here. I see that the account ids are always filtered. Does this happen somewhere else?

bilyana-gospodinova: For example, if the list is: [A=400, B=-300, C=300, D=-400], the current implementation will resolve it to:
[[A=300, B=-300], [A=100, D=-100], [C=300, D=-300]] instead of: [[A=400, D=-400], [C=300, B=-300]]

docs/importer/README.md
```diff
@@ -48,9 +48,15 @@ generate synthetic logs for the following scenarios:
...
-separate transfer pairs for event emission. This could change in the future.
+and this could quickly generate a lot of logs and slow down the system. Multi-party fungible token transfers are supported
+if `hiero.mirror.importer.parser.record.entity.persist.syntheticContractLogsMulti` property is enabled. By default, it
+is set to `true`. The logic for creating transfer pairs is deterministic and uses the following approach:
+
+- senders and receiver accounts are sorted incrementally by `AccountID`'s natural order
+- if sender and receiver amounts are not directly paired, transfers are split by consuming the first available sender
+  balance by priority until it's exhausted. This means a given account can be present in multiple events because of the
+  splitting
+- all remaining senders and receivers are iterated and the same logic applies
```
steven-sheehy: nit: Use proper capitalization and punctuation in docs.

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/EntityRecordItemListener.java
IvanKavaldzhiev: Adapted the algorithm so that this list is resolved to 2 events. The logic now is to sort the senders and receivers first by amount DESC and then by accountID ASC.

importer/src/test/java/org/hiero/mirror/importer/parser/domain/RecordItemBuilder.java
```diff
@@ -645,6 +645,245 @@ public Builder<CryptoTransferTransactionBody.Builder> cryptoTransfer(TransferTyp
...
+        }
+
+        switch (transferType) {
+            case ONE_RECEIVER_TWO_SENDERS:
+                // [A=1000, B=-400, C=-600] =>
+                //
+                // [C=-600, B=-400] senders
+                // [A=1000] receiver
+                //
+                // [[C=-600, A=600], [C=-400, B=400]]
```
bilyana-gospodinova: Shouldn't the result be: [[C=-600, A=600], [B=-400, A=400]]?

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/EntityRecordItemListener.java
```diff
@@ -472,11 +504,66 @@ private void logTokenTransfers(
...
+
+    private void createSyntheticEventsForMultiPartyTransfer(
+            List<AccountTransferChange> senders,
+            List<AccountTransferChange> receivers,
+            RecordItem recordItem,
+            EntityId tokenId) {
+        int s = 0;
+        int r = 0;
+
+        var senderRemainingAmount = Math.abs(senders.get(s).getAmount());
```
bilyana-gospodinova: Do we need `Math.abs()` here? The amount is already positive when `AccountTransferChange` is constructed.

importer/src/test/java/org/hiero/mirror/importer/parser/domain/RecordItemBuilder.java
```diff
@@ -645,6 +645,245 @@ public Builder<CryptoTransferTransactionBody.Builder> cryptoTransfer(TransferTyp
...
+                //
+                // [D=-400, B=-300] senders
+                // [A=400, C=300] receivers
+                //
+                // [[D=-400, A=400], [B=-300, C=300]]
+                tokenTransfers
+                        .addTransfers(accountAmount(accounts.get(0), 400))
+                        .addTransfers(accountAmount(accounts.get(1), -400))
+                        .addTransfers(accountAmount(accounts.get(2), 300))
+                        .addTransfers(accountAmount(accounts.get(3), -300));
```
bilyana-gospodinova: Isn't that the same as the case above? Shouldn't it be: 
```suggestion
                tokenTransfers
                        .addTransfers(accountAmount(accounts.get(0), 400))
                        .addTransfers(accountAmount(accounts.get(1), -300))
                        .addTransfers(accountAmount(accounts.get(2), 300))
                        .addTransfers(accountAmount(accounts.get(3), -400));
```

importer/src/test/java/org/hiero/mirror/importer/parser/domain/RecordItemBuilder.java
```diff
@@ -645,6 +645,245 @@ public Builder<CryptoTransferTransactionBody.Builder> cryptoTransfer(TransferTyp
...
+                        .addTransfers(accountAmount(accounts.get(4), 200))
+                        .addTransfers(accountAmount(accounts.get(5), -200));
+                break;
+            case TWO_RECEIVERS_WITH_DIFFERENT_AMOUNT:
+                // [A=1000, B=-400, C=-600, D=500, E=400, F=-800, G=-100] =>
+                //
+                // [F=-800, C=-600, B=-400, G=-100] senders
+                // [A=1000, D=500, E=400] receivers
+                //
+                // [[F=-800, A=800], [C=-200, A=200], [C=-400, D=400], [B=-100, D=100], [B=-300, E=E00], [G=-100,
```
bilyana-gospodinova: nit: E00 -> 300

importer/src/test/java/org/hiero/mirror/importer/parser/domain/RecordItemBuilder.java
```diff
@@ -645,6 +645,245 @@ public Builder<CryptoTransferTransactionBody.Builder> cryptoTransfer(TransferTyp
...
+                // [[B=-400, A=400], [D=-300, C=300], [F=-200, E=200]]
+                tokenTransfers
+                        .addTransfers(accountAmount(accounts.get(0), 400))
+                        .addTransfers(accountAmount(accounts.get(1), -400))
+                        .addTransfers(accountAmount(accounts.get(2), 300))
+                        .addTransfers(accountAmount(accounts.get(3), -300))
+                        .addTransfers(accountAmount(accounts.get(4), 200))
+                        .addTransfers(accountAmount(accounts.get(5), -200));
+                break;
+            case TWO_RECEIVERS_WITH_DIFFERENT_AMOUNT:
```
bilyana-gospodinova: nit: THREE_RECEIVERS_WITH_DIFFERENT_AMOUNT

importer/src/test/java/org/hiero/mirror/importer/parser/domain/RecordItemBuilder.java
IvanKavaldzhiev: Fixed

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/EntityRecordItemListener.java
IvanKavaldzhiev: You're right. Removed.

importer/src/test/java/org/hiero/mirror/importer/parser/domain/RecordItemBuilder.java
IvanKavaldzhiev: Fixed

importer/src/test/java/org/hiero/mirror/importer/parser/domain/RecordItemBuilder.java
IvanKavaldzhiev: Fixed

importer/src/test/java/org/hiero/mirror/importer/parser/domain/RecordItemBuilder.java
IvanKavaldzhiev: Fixed

bilyana-gospodinova: There is still this case on line 755 `TWO_RECEIVERS_WITH_DIFFERENT_AMOUNT` which has three receivers but the name suggests otherwise.

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/EntityRecordItemListener.java
```diff
@@ -459,7 +466,32 @@ private void logTokenEvents(
             EntityId senderId = amount < 0 ? accountId : EntityId.EMPTY;
             EntityId receiverId = amount > 0 ? accountId : EntityId.EMPTY;
             syntheticContractLogService.create(
-                    new TransferContractLog(recordItem, tokenId, senderId, receiverId, Math.abs(amount)));
+                    new TransferContractLog(recordItem, tokenId, senderId, receiverId, Math.abs(amount), false));
+        }
+    }
+
+    @Getter
+    private static class AccountTransferChange implements Comparable<AccountTransferChange> {
```
steven-sheehy: We should not allocate wrapper objects for every transfer just because we need to have a special comparison. Instead, continue to keep them as `AccountAmount` and pass a static `Comparator` to the collection sort.

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/EntityRecordItemListener.java
```diff
@@ -472,11 +504,66 @@ private void logTokenTransfers(
...
-            EntityId senderId = i == 0
+            final var senderId = i == 0
                     ? EntityId.of(tokenTransfers.get(1).getAccountID())
                     : EntityId.of(tokenTransfers.get(0).getAccountID());
             syntheticContractLogService.create(
-                    new TransferContractLog(recordItem, tokenId, senderId, accountId, amount));
+                    new TransferContractLog(recordItem, tokenId, senderId, accountId, amount, false));
+        } else if (entityProperties.getPersist().isSyntheticContractLogsMulti()) {
+            final var senders = new ArrayList<AccountTransferChange>();
+            final var receivers = new ArrayList<AccountTransferChange>();
```
steven-sheehy: We should specify an estimated size of transfers.length / 2

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/EntityRecordItemListener.java
```diff
@@ -472,11 +504,66 @@ private void logTokenTransfers(
...
+        var senderRemainingAmount = senders.get(s).getAmount();
+        var receiverRemainingAmount = receivers.get(r).getAmount();
+
+        while (s < senders.size() && r < receivers.size()) {
+            final var amountForSyntheticContractLog = Math.min(senderRemainingAmount, receiverRemainingAmount);
+            syntheticContractLogService.create(new TransferContractLog(
+                    recordItem,
+                    tokenId,
+                    EntityId.of(senders.get(s).getAccountId()),
+                    EntityId.of(receivers.get(r).getAccountId()),
```
steven-sheehy: We should only call senders.get(s) or receivers.get(r) once per loop, including the initialization outside the loop.

importer/src/test/java/org/hiero/mirror/importer/parser/domain/RecordItemBuilder.java
```diff
@@ -645,6 +645,245 @@ public Builder<CryptoTransferTransactionBody.Builder> cryptoTransfer(TransferTyp
         return builder;
     }
 
+    /**
+     * Creates a crypto transfer with multi-party token transfers based on transfer type.
+     *
+     * @param transferType the multi-party transfer type
+     * @return the builder
+     */
+    public Builder<CryptoTransferTransactionBody.Builder> cryptoTransferWithMultiPartyTokenTransfers(
```
steven-sheehy: I don't think this test-specific method should be in the generic `RecordItemBuilder`. Should move it to a helper method in the test class it is used in.

importer/src/test/java/org/hiero/mirror/importer/parser/domain/RecordItemBuilder.java
```diff
@@ -645,6 +645,245 @@ public Builder<CryptoTransferTransactionBody.Builder> cryptoTransfer(TransferTyp
...
+                        .addTransfers(accountAmount(accounts.get(2), 300))
+                        .addTransfers(accountAmount(accounts.get(3), -300));
+                break;
+            case PAIRED_SENDERS_AND_RECEIVERS_OF_TWO_PAIRS_WITH_DIFFERENT_AMOUNT_MIXED_ORDER:
+                // [A=400, B=-300, C=300, D=-400] =>
+                //
+                // [D=-400, B=-300] senders
+                // [A=400, C=300] receivers
+                //
+                // [[D=-400, A=400], [B=-300, C=300]]
```
steven-sheehy: I think for all scenarios we can simplify the comment to just show before and after to be less verbose and remove 5 lines. The senders and receivers should be obvious from the resulting transfers and can be omitted.

```suggestion
                // [A=400, B=-300, C=300, D=-400] => [[D=-400, A=400], [B=-300, C=300]]
```

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticContractLogServiceImpl.java
```diff
@@ -19,7 +19,16 @@ public class SyntheticContractLogServiceImpl implements SyntheticContractLogServ
...
     @Override
     public void create(SyntheticContractLog log) {
-        if (isContract(log.getRecordItem()) || !entityProperties.getPersist().isSyntheticContractLogs()) {
+        final boolean isFromMultiTransferAndShouldCreateSyntheticLog =
+                log instanceof TransferContractLog transferContractLog
+                        && transferContractLog.hasMultiPartyOrigin()
+                        && entityProperties
+                                .getPersist()
+                                .getDisableSyntheticEventsForMultiPartyTransfersVersion()
+                                .isGreaterThan(log.getRecordItem().getHapiVersion());
```
steven-sheehy: Why does it matter if multi party transfer or not and specific HAPI version? From what I understand, the new contract events in https://github.com/hiero-ledger/hiero-consensus-node/issues/22615 can also be multi-party. For example: `transferTokens(address token, address[] memory accountId, int64[] memory amount)`. It would be best if we could find a way to detect if events already exist in parent instead of a property.

IvanKavaldzhiev: It does matter if it's a multi-party transfer because in the typical case for a contractCall related transfer - we would directly return and not create an event. In the multi-party transfer case we want to fill missing synthetic events for historically executed precompiles (contract call related transfers), so we should skip the return and actually create an event.

The HAPI version is needed, because synthetic events for precompile transfers will be created directly from consensus nodes and be part of the record/block stream. We would want to create events for contract call related transfers on mirror-node side only before consensus nodes started creating these events.

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/EntityProperties.java
IvanKavaldzhiev: Done

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/EntityRecordItemListener.java
IvanKavaldzhiev: Extracted to a separate class

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/EntityRecordItemListener.java
IvanKavaldzhiev: Fixed

docs/importer/README.md
IvanKavaldzhiev: Reworded the explanations

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/EntityRecordItemListener.java
IvanKavaldzhiev: Fixed

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/EntityRecordItemListener.java
IvanKavaldzhiev: Fixed

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/EntityRecordItemListener.java
IvanKavaldzhiev: Fixed

importer/src/test/java/org/hiero/mirror/importer/parser/domain/RecordItemBuilder.java
IvanKavaldzhiev: Fixed

importer/src/test/java/org/hiero/mirror/importer/parser/domain/RecordItemBuilder.java
IvanKavaldzhiev: Fixed

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/MultiPartyTransferEventsGenerator.java
```diff
@@ -0,0 +1,94 @@
...
+import java.util.List;
+import lombok.RequiredArgsConstructor;
+import org.hiero.mirror.common.domain.entity.EntityId;
+import org.hiero.mirror.common.domain.transaction.RecordItem;
+import org.hiero.mirror.importer.parser.contractlog.SyntheticContractLogService;
+import org.hiero.mirror.importer.parser.contractlog.TransferContractLog;
+
+@Named
+@RequiredArgsConstructor
+public class MultiPartyTransferEventsGenerator {
```
steven-sheehy:
```suggestion
final class MultiPartyTransferEventsGenerator {
```

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/MultiPartyTransferEventsGenerator.java
```diff
@@ -0,0 +1,94 @@
...
+    private static final Comparator<AccountAmount> ACCOUNT_AMOUNT_COMPARATOR = Comparator.<AccountAmount>comparingLong(
+                    e -> Math.abs(e.getAmount()))
+            .reversed()
+            .thenComparing(e -> e.getAccountID().getShardNum())
+            .thenComparing(e -> e.getAccountID().getRealmNum())
+            .thenComparing(
+                    e -> e.getAccountID().hasAccountNum() ? e.getAccountID().getAccountNum() : 0L);
+
+    public void generate(RecordItem recordItem, EntityId tokenId, List<AccountAmount> tokenTransfers) {
+        final var senders = new ArrayList<AccountAmount>(tokenTransfers.size() / 2);
```
steven-sheehy: Could store `tokenTransfers.size() / 2` as a variable and reuse

IvanKavaldzhiev: Fixed

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/MultiPartyTransferEventsGenerator.java
IvanKavaldzhiev: Fixed

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/EntityRecordItemListener.java
```diff
@@ -422,7 +424,9 @@ private void insertFungibleTokenTransfers(RecordItem recordItem, TokenTransferLi
 
             logTokenEvents(recordItem, tokenId, isWipeOrBurn, isMint, accountId, amount);
 
-            logTokenTransfers(recordItem, tokenId, tokenTransfers, isSingleTransfer, i, accountId, amount);
+            if (isSingleTransfer || isMultipleTransfer && i == 0) {
```
xin-hedera: make the code easier to read

```suggestion
            if (isSingleTransfer || (isMultipleTransfer && i == 0)) {
```

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/EntityRecordItemListener.java
```diff
@@ -422,7 +424,9 @@ private void insertFungibleTokenTransfers(RecordItem recordItem, TokenTransferLi
...
+                logTokenTransfers(recordItem, tokenId, tokenTransfers, isSingleTransfer, i, accountId, amount);
```
xin-hedera: we should really have all logic self contained in `logTokenTransfers` including `isSingleTransfer`, `isMultipleTransfer`, and call `logTokenTransfers` outside of the `for (int i = 0; i < tokenTransferCount; i++)` loop.

It's hard to understand how it works in the current form.

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/EntityRecordItemListener.java
```diff
@@ -473,11 +477,13 @@ private void logTokenTransfers(
...
             long amount) {
         if (isSingleTransfer && amount > 0) {
-            EntityId senderId = i == 0
+            final var senderId = i == 0
                     ? EntityId.of(tokenTransfers.get(1).getAccountID())
                     : EntityId.of(tokenTransfers.get(0).getAccountID());
             syntheticContractLogService.create(
                     new TransferContractLog(recordItem, tokenId, senderId, accountId, amount));
+        } else if (entityProperties.getPersist().isSyntheticContractLogsMulti()) {
+            multiPartyTransferEventsGenerator.generate(recordItem, tokenId, tokenTransfers);
```
xin-hedera: nit, we should check `isSyntheticContractLogsMulti()` inside `multiPartyTransferEventsGenerator.generate`

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
```diff
@@ -223,6 +223,18 @@ public byte[] getTransactionHash() {
                         DomainUtils.toBytes(getTransactionRecord().getTransactionHash()), 0, 32));
     }
 
+    /**
+     * Check if the recordItem has a parent and the parent has existing contract logs already present in the incoming
+     * stream
+     *
+     * @return whether the parent record item has contract logs
+     * */
+    public boolean parentHasContractLogs() {
```
xin-hedera: I think instead of `parentHasContractLogs`, we should do

```java
    public boolean hasContractLogs() {
        if (transactionRecord.hasContractCreateResult() || transactionRecord.hasContractCallResult()) {
            final var contractFunctionResult = transactionRecord.hasContractCreateResult() ? transactionRecord.getContractCreateResult() : transactionRecord.getContractCallResult();
            if (contractFunctionResult.getLogInfoCount() > 0) {
                return true;
            }
        }
        
        return parent != null && parent.hasContractLogs();
    }
```

However, I'm not sure if it's a reliable way to determine if consensus node has emitted the previous missing logs or not. The question is what if there are other logs in the result but not the ones consensus node intends to fix and add?

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/EntityRecordItemListener.java
steven-sheehy: This was a code smell that we undid

xin-hedera: Surprise codacy would think this is a code smell…

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/EntityRecordItemListener.java
IvanKavaldzhiev: Refactored.

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/EntityRecordItemListener.java
IvanKavaldzhiev: Refactored.

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
IvanKavaldzhiev: I refactored the whole algorithm.

docs/importer/README.md
```diff
@@ -48,9 +48,18 @@ generate synthetic logs for the following scenarios:
 - Any other current or future transaction that transfers fungible or non-fungible tokens.
 
 For now, the decision was made to omit Hbar transfers because they can be executed at 10,000 transactions per second
-and this could quickly generate a lot of logs and slow down the system. Also, we only support token transfers that
-have a single transfer pair due to the difficulty in splitting aggregated transfers in the `TransactionRecord` into
-separate transfer pairs for event emission. This could change in the future.
+and this could quickly generate a lot of logs and slow down the system. Multi-party fungible token transfers are supported
+if `hiero.mirror.importer.parser.record.entity.persist.syntheticContractLogsMulti` property is enabled. By default, it
```
IvanKavaldzhiev, line 54: Docs need to be updated.

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticContractLogServiceImpl.java
```diff
@@ -3,44 +3,63 @@
...
-        if (isContract(log.getRecordItem()) || !entityProperties.getPersist().isSyntheticContractLogs()) {
+        if (!entityProperties.getPersist().isSyntheticContractLogs()) {
+            return;
+        }
+
+        var recordItem = log.getRecordItem();
+        var parentRecordItem = recordItem.getParent();
+
+        // We will backfill any EVM-related fungible token transfers that don't have synthetic events produced by CN
+        if (isContract(recordItem) && shouldSkipLogCreationForContractTransfer(log, parentRecordItem)) {
```
IvanKavaldzhiev: I think `shouldSkipLogCreationForContractTransfer(log, parentRecordItem)` won't work with a hook transaction. Let's say we have a `CryptoTransfer` transaction that triggers a ContractCall with 3 nested precompile fungible token transfers.

`CryptoTransfer's RecordItem` will have consensusTimestamp X
`ContractCall's RecordItem` will have consensusTimeStamp X + 1 ns and parentConsensusTimestamp X
`1st precompile transfer's RecordItem` will have consensusTimestamp X + 2 ns and parentConsensusTimestamp X
`2nd precompile transfer's RecordItem` will have consensusTimestamp X + 3 ns and parentConsensusTimestamp X
`3rd precompile transfer's RecordItem` will have consensusTimestamp X + 4 ns and parentConsensusTimestamp X

Let's say we are currently importing the 1st precompile transfers' RecordItem and we have a SyntheticContractLog created for the transfer executed by this precompile.

`recordItem.getParent()` will return CryptoTransfer's RecordItem, which don't have ContractLogs and ContractResult and the validations will fail, thus the SyntheticContractLog will persist a ContractLog in DB and cause a duplication. What we are interested in is getting the `ContractCall's RecordItem`. Then the logic will work.

@xin-hedera do you agree with this example? If so, I can add a special check for the hook transaction.

IvanKavaldzhiev: Added tests and a fix for these cases as well.

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/TransferContractLog.java
```diff
@@ -17,4 +21,18 @@ public TransferContractLog(
...
+     *
+     * @param contractLoginfo the ContractLoginfo to compare against
+     * @return true if the logs are equal, false otherwise
+     */
+    public boolean equalsContractLoginfo(ContractLoginfo contractLoginfo) {
+        return Arrays.equals(getTopic0(), Utility.getTopic(contractLoginfo, 0))
+                && Arrays.equals(getTopic1(), Utility.getTopic(contractLoginfo, 1))
+                && Arrays.equals(getTopic2(), Utility.getTopic(contractLoginfo, 2))
+                && Arrays.equals(getTopic3(), Utility.getTopic(contractLoginfo, 3))
+                && Arrays.equals(getData(), Utility.getDataTrimmed(contractLoginfo));
```
steven-sheehy: All 5 of these comparisons result in array allocations. At scale this could hurt performance. We should rewrite to avoid any allocations and compare by iterating over both while skipping leading zeros.

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticContractLogServiceImpl.java
```diff
@@ -3,44 +3,63 @@
...
+        // We will backfill any EVM-related fungible token transfers that don't have synthetic events produced by CN
+        if (isContract(recordItem) && shouldSkipLogCreationForContractTransfer(log)) {
             return;
         }
 
-        long consensusTimestamp = log.getRecordItem().getConsensusTimestamp();
-        int logIndex = log.getRecordItem().getAndIncrementLogIndex();
+        long consensusTimestamp = parentRecordItem != null
+                ? parentRecordItem.getConsensusTimestamp()
+                : recordItem.getConsensusTimestamp();
```
steven-sheehy, line 45: Why are we using the parent timestamp? I assume this is to match how the real logs get attached to the parent contract call? Are we sure that's the behavior we should use for synthetic?

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticContractLogServiceImpl.java
```diff
@@ -50,4 +69,65 @@ private boolean isContract(RecordItem recordItem) {
...
+            // Only TransferContractLog synthetic event creation is supported for an operation with contract origin
+            return true;
+        }
+
+        if (syntheticLog
+                                .getRecordItem()
+                                .getTransactionRecord()
+                                .getTransferList()
+                                .getAccountAmountsCount()
+                        > 2
```
steven-sheehy: Weird formatting here. Maybe pull it out to a separate var.

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticContractLogServiceImpl.java
```diff
@@ -50,4 +69,65 @@ private boolean isContract(RecordItem recordItem) {
...
+                && !entityProperties.getPersist().isSyntheticContractLogsMulti()) {
+            // We have a multi-party fungible transfer scenario and synthetic event creation for
+            // such transfers is disabled
+            return true;
+        } else return logAlreadyImported(transferLog);
```
steven-sheehy:
```suggestion
        }
        
        return logAlreadyImported(transferLog);
```

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
```diff
@@ -125,6 +132,35 @@ public AbstractHook.Id nextHookContext() {
...
+        if (contractLogOccurrences == null || contractLogOccurrences.isEmpty()) {
+            return false;
+        }
+
+        for (var entry : contractLogOccurrences.entrySet()) {
+            var logInfo = entry.getKey();
+            var occurrenceCount = entry.getValue();
+
+            if (occurrenceCount.get() > 0 && matcher.test(logInfo)) {
+                occurrenceCount.decrementAndGet();
```
steven-sheehy: I don't understand why need to track occurrences and decrement them. Can't it just be a set that if we've seen it before regardless of times don't log it?

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
```diff
@@ -294,6 +370,25 @@ private RecordItem parseParent() {
             return this.parent;
         }
 
+        private RecordItem parseHookContractRelatedParent() {
+            if (transactionRecord.hasParentConsensusTimestamp() && previous != null) {
+                if (transactionRecord.getConsensusTimestamp().getNanos()
+                                == previous.transactionRecord
+                                                .getConsensusTimestamp()
+                                                .getNanos()
```
steven-sheehy: Why are we not comparing seconds as well?

steven-sheehy (changes requested):

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
IvanKavaldzhiev: Let's say we have this case and we are pre-HAPI 0.71.0 version, which doesn't emit synthetic events for multi-party transfers:

1. ContractCall with contractResult, tokenTransfers and list of contractLogs containing a single transfer of `A transferring token T to B` 
2. Child RecordItem with a transfer precompile of a single fungible token with a contractResult containing this same `A transferring token T to B` transfer
3. Child RecordItem with a transfer precompile of multi-party fungible token with a contractResult containing a transfer of `A transferring token T to B` and a transfer of `C transferring token T to B`.

We have the same type of event/transfer of A-T-B in 2 different places. For the first Child RecordItem the CN will have created a SyntheticLog. For the second Child RecordItem we won't have any SyntheticLogs.

In this case in the RecordItem for the ContractCall we'll have a contractLogs list with this single A-T-B transfer.

Then we are importing the first child and check whether we should create a ContractLog with A-T-B - since we find exactly the same log already present in the parent - we skip the creation and decrement the occurrence of this transfer inside the parent and now the event will have 0 occurrences in parent.

Then we are importing the second child which will try to create 2 ContractLogs. 1 for A-T-B and 1 for C-T-B. We check whether we have A-T-B in the parent ContractCall - we don't find it and thus create the event by actually backfilling a synthetic event part of a multi-party transfer precompile for record coming from CN without the official support. For C-T-B we also don't find any occurrence - we also create a record for it.

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticContractLogServiceImpl.java
IvanKavaldzhiev: Yes, the idea is to match how real logs are attached in CN. I think this makes sense, especially having a case like the example in one of the comments below:

ContractCall + precompile transfer with single fungible transfer + precompile transfer with multi-party fungible transfer + having a HAPI pre 0.71.0. CN will have a ContractLog for the single transfer with consensusTimestamp as the parent and for the multi-party case - mirror node will backfill the ContractLogs. Then, it makes sense for the ContractLogs from the mirror node to have also the consensus timestamp as the parent (ContractCall).

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
```diff
@@ -240,6 +276,20 @@ public Collection<ContractTransaction> populateContractTransactions() {
...
+        } else {
+            return hasContractResult(parent) ? parent : null;
+        }
+    }
+
+    private boolean hasContractResult(RecordItem recordItem) {
+        return recordItem != null
+                && (recordItem.transactionRecord.hasContractCreateResult()
+                        || recordItem.transactionRecord.hasContractCallResult());
+    }
```
xin-hedera: prefer the following

```suggestion
    private boolean hasContractResult() {
        return  transactionRecord.hasContractCreateResult()
                        || transactionRecord.hasContractCallResult());
    }
```

then in `getContractRelatedParent` change the logic to

```java
      return parent != null && parent.hasContractResult() ? parent : null;
```

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
```diff
@@ -113,6 +116,10 @@ public class RecordItem implements StreamItem {
     @Setter
     private ArrayDeque<AbstractHook.Id> hookExecutionQueue;
 
+    // Map to track remaining occurrences of each contract log for duplicate detection
+    @EqualsAndHashCode.Exclude
+    private Map<ContractLoginfo, AtomicInteger> contractLogOccurrences;
```
xin-hedera: I think it's worth creating a new domain class to hold just the trimmed topics and data. That way, in `consumeMatchingContractLog` we don't have to iterate through the keys of the map.

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
```diff
@@ -294,6 +370,25 @@ private RecordItem parseParent() {
             return this.parent;
         }
 
+        private RecordItem parseHookContractRelatedParent() {
+            if (transactionRecord.hasParentConsensusTimestamp() && previous != null) {
```
xin-hedera: think we can simplify the condition to check that the contract id in the previous's contract call result is 0x365 (the hook system contract id) and the current's contract id is not 0x365


update: however it doesn't work if a hook contract call has multiple child contract calls.

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/TransferContractLog.java
IvanKavaldzhiev: Fixed

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticContractLogServiceImpl.java
IvanKavaldzhiev: Fixed

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticContractLogServiceImpl.java
IvanKavaldzhiev: Fixed

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
IvanKavaldzhiev: Changed the whole check

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
IvanKavaldzhiev: Fixed

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
IvanKavaldzhiev: Created

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
IvanKavaldzhiev: Added a new logic for hook parents.

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/TransferContractLog.java
```diff
@@ -17,4 +20,19 @@ public TransferContractLog(
...
+     * Converts the topics and data fields to a {@link TrimmedTopicsAndData}
+     *
+     * @return TrimmedTopicsAndData the converted object
+     */
+    public TrimmedTopicsAndData getTrimmedTopicsAndData() {
+        var topic1AsByteString = getTopic0() != null ? ByteString.copyFrom(getTopic0()) : ByteString.EMPTY;
+        var topic2AsByteString = getTopic1() != null ? ByteString.copyFrom(getTopic1()) : ByteString.EMPTY;
+        var topic3AsByteString = getTopic2() != null ? ByteString.copyFrom(getTopic2()) : ByteString.EMPTY;
+        var topic4AsByteString = getTopic3() != null ? ByteString.copyFrom(getTopic3()) : ByteString.EMPTY;
+        var dataAsByteString = getData() != null ? ByteString.copyFrom(getData()) : ByteString.EMPTY;
```
steven-sheehy: This still does memory allocation of large byte arrays just for comparison purposes. We can compare without allocating anything on the heap.

common/src/main/java/org/hiero/mirror/common/domain/transaction/TrimmedTopicsAndData.java
```diff
@@ -0,0 +1,29 @@
...
+    ByteString topic4;
+    ByteString data;
+
+    public TrimmedTopicsAndData(ContractLoginfo contractLoginfo) {
+        var topicCount = contractLoginfo.getTopicList().size();
+        topic1 = topicCount > 0 ? DomainUtils.trim(contractLoginfo.getTopic(0)) : ByteString.EMPTY;
+        topic2 = topicCount > 1 ? DomainUtils.trim(contractLoginfo.getTopic(1)) : ByteString.EMPTY;
+        topic3 = topicCount > 2 ? DomainUtils.trim(contractLoginfo.getTopic(2)) : ByteString.EMPTY;
+        topic4 = topicCount > 3 ? DomainUtils.trim(contractLoginfo.getTopic(3)) : ByteString.EMPTY;
+        data = DomainUtils.trim(contractLoginfo.getData());
```
steven-sheehy: Even more memory allocations just for comparing. I know it was just added based upon another comment, but we don't need such this class since it doesn't follow the recommendation from my previous comment to avoid allocations and only compare. This may seem minor but duplicating large byte arrays multiple times when processing tens of thousands of logs per block is a lot of memory to save and could slow down processing.

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/TransferContractLog.java
IvanKavaldzhiev: Agreed, I changed the logic and removed the memory allocation.

common/src/main/java/org/hiero/mirror/common/domain/transaction/TrimmedTopicsAndData.java
IvanKavaldzhiev: Fixed. I removed this new class.

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
```diff
@@ -125,6 +138,42 @@ public AbstractHook.Id nextHookContext() {
...
+        if (consumedContractLogIndices == null) {
+            consumedContractLogIndices = new BitSet(contractLogs.size());
+        }
+
+        for (int i = 0; i < contractLogs.size(); i++) {
+            if (consumedContractLogIndices.get(i)) {
+                continue;
+            }
+            if (contractLogTopicsAndDataMatches(contractLogs.get(i), topic0, topic1, topic2, topic3, data)) {
+                consumedContractLogIndices.set(i);
```
xin-hedera: I think we can just remove the matching contract log from the collection without memorizing it in the bitset.

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
```diff
@@ -294,6 +378,41 @@ private RecordItem parseParent() {
...
+
+                    candidateRecord = candidateRecord.previous;
+                }
+            }
+
+            // we have multiple contract hook-related child items executed via a hook (which are not precompile-related)
+            // and we are already in one of them, so we return the hook triggering transaction instead, which is the
+            // actual
+            // parent of all child items linked to a hook
+            return parent;
```
xin-hedera: can't we just return `null`? 

I'm not seeing setting the crypto transfer transaction as a hook call contract call transaction's `hookParent` being useful.

With `null`, it means there is no hook contract call parent for this transaction.

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
```diff
@@ -294,6 +378,41 @@ private RecordItem parseParent() {
             return this.parent;
         }
 
+        private RecordItem parsePossiblyHookContractRelatedParent() {
+            if (transactionRecord.hasParentConsensusTimestamp() && previous != null) {
```
xin-hedera:
```suggestion
            final var currentTxnRecordContractResult = getContractResult();
            if (transactionRecord.hasParentConsensusTimestamp() && currentTxnRecordContractResult != null && currentTxnRecordContractResult.getContractID().getContractNum() != HOOK_CONTRACT_NUM && previous != null) {
```

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
```diff
@@ -294,6 +378,41 @@ private RecordItem parseParent() {
...
+                var candidateRecord = previous;
+
+                while (candidateRecord != null && candidateRecord != parent) {
+                    var currentTxnRecordContractResult = getContractResult();
```
xin-hedera: there should be null check on `currentTxnRecordContractResult` and this should be done outside of the while loop.

```suggestion
```

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
```diff
@@ -294,6 +378,41 @@ private RecordItem parseParent() {
...
+                    var previousTxnRecordContractResult = candidateRecord.getContractResult();
```
xin-hedera: nit, using previous is confusing.

```suggestion
                    var candidateTxnRecordContractResult = candidateRecord.getContractResult();
```

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
```diff
@@ -294,6 +378,41 @@ private RecordItem parseParent() {
...
+                    if (previousTxnRecordContractResult != null
+                            && previousTxnRecordContractResult.getContractID().getContractNum() == HOOK_CONTRACT_NUM
+                            && currentTxnRecordContractResult.getContractID().getContractNum() != HOOK_CONTRACT_NUM) {
+                        // we found the first hook child item, which has a ContractResult and is executed via a hook
+                        return candidateRecord;
+                    }
```
xin-hedera: - Should shortcut when the candidate isn't an EVM transaction
- The current EVM transaction must not be called with the hook system contract condition is moved up to the top level if condition

```suggestion
                    if (candidateTxnRecordContractResult == null) {
                        break;
                    }

                    if (candidateTxnRecordContractResult.getContractID().getContractNum() == HOOK_CONTRACT_NUM) {
                        // we found the first hook child item, which has a ContractResult and is executed via a hook
                        return candidateRecord;
                    }
```

common/src/main/java/org/hiero/mirror/common/util/DomainUtils.java
```diff
@@ -362,6 +364,48 @@ public static BytesValue trim(final BytesValue data) {
...
+                && trimmedByteStringEquals(topicCount > 2 ? topicList.get(2) : EMPTY_BYTE_STRING, topic2)
+                && trimmedByteStringEquals(topicCount > 3 ? topicList.get(3) : EMPTY_BYTE_STRING, topic3)
+                && trimmedByteStringEquals(log.getData(), data);
+    }
+
+    /**
+     * Compares ByteString to byte[] trim-aware (leading zeros in ByteString skipped), without allocating.
+     */
+    private static boolean trimmedByteStringEquals(ByteString bs, byte[] arr) {
+        byte[] a = arr != null ? arr : EMPTY_BYTE_ARRAY;
```
xin-hedera: It's a bug by assigning `EMPTY_BYTE_ARRAY` to `a` when null.

Because after trimming, an empty byte array indicates there is topic it's just the original content from the contract result is all 0s.

A `null` topic from the domain object means there's no such topic from the contract result.

common/src/main/java/org/hiero/mirror/common/util/DomainUtils.java
```diff
@@ -362,6 +364,48 @@ public static BytesValue trim(final BytesValue data) {
...
+     * Compares a contract log's topics and data to the given byte arrays (trim-aware, no allocation).
+     */
+    public static boolean contractLogTopicsAndDataMatches(
+            ContractLoginfo log, byte[] topic0, byte[] topic1, byte[] topic2, byte[] topic3, byte[] data) {
+        var topicList = log.getTopicList();
+        int topicCount = topicList.size();
+        return trimmedByteStringEquals(topicCount > 0 ? topicList.get(0) : EMPTY_BYTE_STRING, topic0)
+                && trimmedByteStringEquals(topicCount > 1 ? topicList.get(1) : EMPTY_BYTE_STRING, topic1)
+                && trimmedByteStringEquals(topicCount > 2 ? topicList.get(2) : EMPTY_BYTE_STRING, topic2)
+                && trimmedByteStringEquals(topicCount > 3 ? topicList.get(3) : EMPTY_BYTE_STRING, topic3)
```
xin-hedera: Similarly, using `EMPTY_BYTE_STRING` as a fallback is also a bug, for instance

- topicCount = 1
- topic0 matches topicList.get(0)
- topic1 is an empty array
- no topic2 / topic3

The two contract logs should not be a match.

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticContractLogServiceImpl.java
```diff
@@ -50,4 +69,75 @@ private boolean isContract(RecordItem recordItem) {
...
+     * same contract log appears multiple times in the child records as being part of different operations.
+     *
+     * @param transferLog the TransferContractLog to check
+     * @return true if a matching log is found and it is already persisted, false otherwise
+     */
+    private boolean logAlreadyImported(TransferContractLog transferLog) {
+        var parentRecordItem = transferLog.getRecordItem().getContractRelatedParent();
+        return transferLog
+                        .getRecordItem()
+                        .consumeMatchingContractLog(
```
xin-hedera: we should move the logic of looking for parent into `consumeMatchingContractLog`, then remove the parent check here.

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/TransferEventsGenerator.java
```diff
@@ -0,0 +1,99 @@
...
+@Named
+@RequiredArgsConstructor
+final class TransferEventsGenerator {
+
+    private final SyntheticContractLogService syntheticContractLogService;
+    private static final Comparator<AccountAmount> ACCOUNT_AMOUNT_COMPARATOR = Comparator.<AccountAmount>comparingLong(
+                    e -> Math.abs(e.getAmount()))
+            .reversed()
+            .thenComparing(e -> e.getAccountID().getShardNum())
+            .thenComparing(e -> e.getAccountID().getRealmNum())
```
xin-hedera, line 25: do we need to compare shard / realm?

importer/src/main/java/org/hiero/mirror/importer/util/Utility.java
```diff
@@ -173,6 +173,36 @@ public static byte[] getTopic(ContractLoginfo contractLoginfo, int index) {
...
+     * @param contractLoginfo the contract log info
+     * @return a byte array with leading zeros removed, or null if empty
+     */
+    @SuppressWarnings("java:S1168")
+    public static byte[] getDataTrimmed(ContractLoginfo contractLoginfo) {
+        ByteString byteString = contractLoginfo.getData();
+
+        if (byteString == null || byteString.isEmpty()) {
+            return null;
+        }
```
xin-hedera: - `byteString` won't be null
- If `byteString` is ever empty, I believe we should return an empty byte array

xin-hedera (changes requested):

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
IvanKavaldzhiev: Agreed, fixed

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
IvanKavaldzhiev: Makes sense, fixed

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
IvanKavaldzhiev: Fixed

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
IvanKavaldzhiev: Fixed

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
IvanKavaldzhiev: Fixed

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
IvanKavaldzhiev: Fixed

common/src/main/java/org/hiero/mirror/common/util/DomainUtils.java
IvanKavaldzhiev: Good catch! Fixed.

common/src/main/java/org/hiero/mirror/common/util/DomainUtils.java
IvanKavaldzhiev: Fixed

importer/src/main/java/org/hiero/mirror/importer/parser/contractlog/SyntheticContractLogServiceImpl.java
IvanKavaldzhiev: Fixed

importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/TransferEventsGenerator.java
IvanKavaldzhiev: Me and Glib agreed to make such comparison. That's also the logic in CN.

importer/src/main/java/org/hiero/mirror/importer/util/Utility.java
IvanKavaldzhiev: Actually this method was unused, so it's deleted now.

common/src/main/java/org/hiero/mirror/common/util/DomainUtils.java
```diff
@@ -362,6 +363,51 @@ public static BytesValue trim(final BytesValue data) {
...
+                && trimmedByteStringEquals(log.getData(), data);
+    }
+
+    /**
+     * Compares ByteString to byte[] trim-aware (leading zeros in ByteString skipped), without allocating.
+     */
+    private static boolean trimmedByteStringEquals(ByteString bs, byte[] arr) {
+        if (bs == null) {
+            return arr == null;
+        }
```
xin-hedera:
```suggestion
        if (bs == null && arr == null) {
            return true;
        } else if (bs == null || arr == null) {
            return false;
        }
```

Reason for the change request is there is no null check on `arr` and when `bs` is not empty, a null `arr` will cause NPE.

IvanKavaldzhiev: Fixed

common/src/main/java/org/hiero/mirror/common/domain/transaction/RecordItem.java
```diff
@@ -249,15 +327,29 @@ public RecordItem build() {
...
             return buildInternal();
         }
 
+        private List<ContractLoginfo> parseContractLogs() {
+            if (transactionRecord.hasContractCallResult()) {
+                return new ArrayList<>(transactionRecord.getContractCallResult().getLogInfoList().stream()
+                        .toList());
+            } else if (transactionRecord.hasContractCreateResult()) {
+                return new ArrayList<>(transactionRecord.getContractCreateResult().getLogInfoList().stream()
+                        .toList());
```
xin-hedera:
```suggestion
                return new ArrayList<>(transactionRecord.getContractCallResult().getLogInfoList());
            } else if (transactionRecord.hasContractCreateResult()) {
                return new ArrayList<>(transactionRecord.getContractCreateResult().getLogInfoList());
```

IvanKavaldzhiev: Fixed

PR #12758, opened 2026-01-13
Fix GRPC race condition bug

### LINKED ISSUE #11872: Initial topic message sometimes not delivered
### Description

The `ConsensusService.subscribeTopic()` gRPC API can sometimes fail to deliver the initial message upon subscription due to a race condition when transitioning from historical to live streaming.

### Steps to reproduce

Run acceptance test in CI. Here's an example: https://github.com/hiero-ledger/hiero-mirror-node/actions/runs/17270892681/job/49015064782?pr=11761

### Additional context

```
    2025-08-27T15:28:05.565Z  INFO ForkJoinPool-1-worker-1 o.h.m.t.e.a.c.StartupProbe Creating a topic to confirm node connectivity
    2025-08-27T15:28:06.173Z  INFO ForkJoinPool-1-worker-1 o.h.m.t.e.a.c.StartupProbe Created topic 0.0.1002 successfully
    2025-08-27T15:28:06.693Z  INFO ForkJoinPool-1-worker-1 o.h.m.t.e.a.c.MirrorNodeClient GET http://mirror-rest/api/v1/transactions/0.0.2-1756308475-587000483: 200 OK
    2025-08-27T15:28:06.693Z  INFO ForkJoinPool-1-worker-1 o.h.m.t.e.a.c.StartupProbe Subscribing to topic 0.0.1002
    2025-08-27T15:28:06.698Z  INFO ForkJoinPool-1-worker-1 o.h.m.t.e.a.c.StartupProbe Submitting a message to the network
    2025-08-27T15:28:06.949Z  INFO grpc-executor-1 o.h.m.g.s.TopicMessageServiceImpl Subscribing to topic: TopicMessageFilter(endTime=null, limit=0, startTime=0, subscriberId=X1YNDEEq, topicId=0.0.1002)
    2025-08-27T15:28:06.961Z  INFO ForkJoinPool-1-worker-1 o.h.m.t.e.a.c.StartupProbe Waiting for the mirror node to publish the topic message
    2025-08-27T15:28:07.666Z  INFO grpc-executor-1 o.h.m.g.r.PollingTopicMessageRetriever [X1YNDEEq] Finished retrieving 0 messages in 713.5 ms (0/s)
    2025-08-27T15:28:07.678Z  INFO grpc-executor-1 o.h.m.g.l.RedisTopicListener Subscribing: TopicMessageFilter(endTime=null, limit=0, startTime=0, subscriberId=X1YNDEEq, topicId=0.0.1002)
    2025-08-27T15:28:07.679Z  INFO grpc-executor-1 o.h.m.g.l.RedisTopicListener Creating shared subscription to topic.1002
    2025-08-27T15:28:07.680Z  INFO grpc-executor-1 o.h.m.g.l.RedisTopicListener Attempting to connect to Redis
    2025-08-27T15:28:07.681Z  INFO grpc-executor-1 o.h.m.g.l.RedisTopicListener Connected to Redis
```

### Hedera network

other

### Version

v0.137.0

### Operating system

None

bilyana-gospodinova:
**Description**:
When a message is received during the historical data loading and the listener for live messages is not started yet, the current code hangs until there is another message received.
This PR changes the implementation of the GRPC's `TopicMessageServiceImpl` to have a safety check which tries to load messages that were received before the live listener is started.

Fixes #11872

### DIFF
--- a/grpc/src/main/java/org/hiero/mirror/grpc/service/TopicMessageServiceImpl.java
+++ b/grpc/src/main/java/org/hiero/mirror/grpc/service/TopicMessageServiceImpl.java
@@ -65,12 +65,7 @@ public class TopicMessageServiceImpl implements TopicMessageService {
         // Safety Check - Polls missing messages after 1s if we are stuck with no data
         Flux<TopicMessage> safetyCheck = Mono.delay(Duration.ofSeconds(1L))
                 .filter(_ -> !topicContext.isComplete())
-                .flatMapMany(tick -> {
-                    log.info(
-                            "Safety check triggering gap recovery query for subscriber id: {}",
-                            filter.getSubscriberId());
-                    return missingMessages(topicContext, null);
-                })
+                .flatMapMany(_ -> missingMessages(topicContext, null))
                 .subscribeOn(Schedulers.boundedElastic());
 
         Flux<TopicMessage> flux = historical
@@ -149,6 +144,7 @@ public class TopicMessageServiceImpl implements TopicMessageService {
                     : topicContext.getFilter().getStartTime();
             var gapFilter =
                     topicContext.getFilter().toBuilder().startTime(startTime).build();
+            log.info("Safety check triggering gap recovery query with filter {}", gapFilter);
             return topicMessageRetriever.retrieve(gapFilter, false);
         }
 
--- a/grpc/src/test/java/org/hiero/mirror/grpc/service/TopicMessageServiceTest.java
+++ b/grpc/src/test/java/org/hiero/mirror/grpc/service/TopicMessageServiceTest.java
@@ -634,6 +634,39 @@ class TopicMessageServiceTest extends GrpcIntegrationTest {
                 .verify();
     }
 
+    @Test
+    void receivedMessageBeforeSafetyCheckCompletion() {
+        var topicListener = Mockito.mock(TopicListener.class);
+        var entityRepository = Mockito.mock(EntityRepository.class);
+        var topicMessageRetriever = Mockito.mock(TopicMessageRetriever.class);
+        topicMessageService = new TopicMessageServiceImpl(
+                new GrpcProperties(),
+                topicListener,
+                entityRepository,
+                topicMessageRetriever,
+                new SimpleMeterRegistry());
+
+        var filter = TopicMessageFilter.builder().startTime(0).topicId(TOPIC_ID).build();
+        var newFilter = filter.toBuilder().startTime(2L).build();
+
+        var message1 = topicMessage(1);
+
+        Mockito.when(entityRepository.findById(filter.getTopicId().getId())).thenReturn(optionalEntity());
+        // No historical messages
+        Mockito.when(topicMessageRetriever.retrieve(filter, true)).thenReturn(Flux.empty());
+        // Gap message from the safety check
+        Mockito.when(topicMessageRetriever.retrieve(filter, false)).thenReturn(Flux.just(message1));
+        Mockito.when(topicListener.listen(Mockito.any())).thenReturn(Flux.empty());
+
+        StepVerifier.withVirtualTime(
+                        () -> topicMessageService.subscribeTopic(filter).map(TopicMessage::getSequenceNumber))
+                .expectSubscription()
+                .thenAwait(Duration.ofSeconds(2)) // Await the safety check to finish
+                .expectNext(1L)
+                .thenCancel()
+                .verify();
+    }
+
     private void missingMessagesFromListenerTest(TopicMessageFilter filter, Flux<TopicMessage> missingMessages) {
         var topicListener = Mockito.mock(TopicListener.class);
         var entityRepository = Mockito.mock(EntityRepository.class);


grpc/src/test/java/org/hiero/mirror/grpc/service/TopicMessageServiceTest.java
```diff
@@ -343,8 +347,6 @@ void incomingMessagesWithEndTimeBefore() {
                 .subscribeTopic(filter)
                 .map(TopicMessage::getSequenceNumber)
                 .as(StepVerifier::create)
-                .thenAwait(Duration.ofMillis(50))
-                .then(generator::blockLast)
```
steven-sheehy, line 347: The fact that this test case had to change to work implies something regressed. It's no longer generating messages interleaved and always generating them up front. We should revert.

grpc/src/main/java/org/hiero/mirror/grpc/service/TopicMessageServiceImpl.java
```diff
@@ -57,10 +58,18 @@ public Flux<TopicMessage> subscribeTopic(TopicMessageFilter filter) {
         log.info("Subscribing to topic: {}", filter);
         TopicContext topicContext = new TopicContext(filter);
 
+        // Connect to the source as soon as the first subscriber arrives.
+        // Buffer the last BUFFERED_MESSAGES_COUNT messages in case of a DB gap.
+        Flux<TopicMessage> liveStream =
+                incomingMessages(topicContext).replay(BUFFERED_MESSAGES_COUNT).refCount();
```
steven-sheehy: I put some `doOnNext()` and `doOnSubscribe()` logs in the retriever and listener and this doesn't do what you expect. It only subscribes to the listener after the retriever completes all messages just as it did before. This is because of the use of `concatWith` which does not interleave messages or eagerly subscribe.

If you did get the approach that subscribes to listener first working I still have concerns. It would cause the gRPC API to hold 20 messages in memory for every subscriber which at scale could be a lot of memory. It could cause slow performance with topics with high historical volume and/or high current volume. Also, while back-filling historical messages, Redis could receive recent messages that trigger yet another concurrent call to `TopicMessageRetriever` to fill missing messages, which would be very slow.

I think we want an approach that doesn't incur a cost on all streaming messages, just that first message. We should update `missingMessages()` to check for missing messages if last is null. And we should have a delayed mono run once 1 second after redis subscription to search for missing messages if last is null.

```java
Mono.delay(Duration.ofSeconds(1L))
  .thenMany(Flux.defer(() -> last == null ? Flux.empty() : missingMessages(topicContext, null))
  .subscribeOn(Schedulers.boundedElastic()))
  .mergeWith(Flux.defer(() -> incomingMessages(topicContext))
```

steven-sheehy (changes requested):

grpc/src/test/java/org/hiero/mirror/grpc/service/TopicMessageServiceTest.java
bilyana-gospodinova: Reverted

grpc/src/main/java/org/hiero/mirror/grpc/service/TopicMessageServiceImpl.java
bilyana-gospodinova: Thanks for the suggestion! I think I was able to address that. Please re-review.

grpc/src/main/java/org/hiero/mirror/grpc/service/TopicMessageServiceImpl.java
```diff
@@ -119,27 +134,27 @@ private Flux<Object> pastEndTime(TopicContext topicContext) {
...
+        }
+
         if (topicContext.isNext(current)) {
             return Flux.just(current);
         }
 
-        TopicMessage last = topicContext.getLast();
         long numMissingMessages = current.getSequenceNumber() - last.getSequenceNumber() - 1;
 
-        // fail fast on out of order messages
```
bilyana-gospodinova, line 135: Removed this check as with the safety check this one might fail unnecessarily. The order of the messages is still verified in here:
```
.filter(t -> {
                    TopicMessage last = topicContext.getLast();
                    return last == null || t.getSequenceNumber() > last.getSequenceNumber();
                })
```
and also in `onNext`:
```
void onNext(TopicMessage topicMessage) {
            if (!isNext(topicMessage)) {
                throw new IllegalStateException(
                        String.format("Encountered out of order messages, last: %s, current: %s", last, topicMessage));
            }
...
```

grpc/src/main/java/org/hiero/mirror/grpc/service/TopicMessageServiceImpl.java
```diff
@@ -57,10 +59,26 @@ public Flux<TopicMessage> subscribeTopic(TopicMessageFilter filter) {
...
+        Flux<TopicMessage> live = Flux.defer(() -> incomingMessages(topicContext));
+
+        // Safety Check - Polls missing messages after 1s if we are stuck with no data
+        Flux<TopicMessage> safetyCheck = Mono.delay(Duration.ofSeconds(1L))
+                .filter(_ -> !topicContext.isComplete())
+                .flatMapMany(tick -> {
+                    log.info(
+                            "Safety check triggering gap recovery query for subscriber id: {}",
+                            filter.getSubscriberId());
+                    return missingMessages(topicContext, null);
```
steven-sheehy: nit: Maybe move the log into missingMessages() null current path to make this cleaner.

```suggestion
                .flatMapMany(_ -> missingMessages(topicContext, null))
```

grpc/src/test/java/org/hiero/mirror/grpc/service/TopicMessageServiceTest.java
```diff
@@ -582,6 +586,54 @@ void missingMessagesFromRetrieverAndListener() {
                 .verify(WAIT);
     }
 
+    @Test
+    void receivedMessagesBeforeQueryCompletion() {
```
steven-sheehy, line 590: We should also have a scenario where no historical message and no live message occurs. Instead, one topic message just appears in the db before the safety check completes. This will prove that we receive topic messages that appear after historical but before redis.

grpc/src/main/java/org/hiero/mirror/grpc/service/TopicMessageServiceImpl.java
bilyana-gospodinova: Moved

grpc/src/test/java/org/hiero/mirror/grpc/service/TopicMessageServiceTest.java
bilyana-gospodinova: Added `receivedMessageBeforeSafetyCheckCompletion()` test.

PR #12759, opened 2026-01-13
Convert to spring-grpc

### LINKED ISSUE #12452: grpc-spring-boot-starter not compatible with Spring Boot 4
### Problem

We'd like to upgrade to Spring Boot 4 but `net.devh:grpc-spring-boot-starter` is not compatible with Spring Boot 4 and doesn't seem to be maintained anymore.

### Solution

Convert to Spring's new official grpc project:

* Remove `net.devh:grpc-spring-boot-starter` dependency
* Add [spring-grpc](https://docs.spring.io/spring-grpc/reference/index.html) dependency
* Use [Spring Initializer](https://start.spring.io/) and add grpc dependency to use as reference for `build.gradle.kts` changes
* Change grpc controllers to extend synchronous base classes from spring-grpc
* Adapt StreamObserver to Flux/Mono via doFinally, onErrorMap, and subscribe
* Keep service layer reactive
* Enable virtual threads
* Test 10k HCS streaming as in canonical tests
* Also incorporate #9630 into that testing

### Alternatives

_No response_

sdimitrov9:
**Description**:
Replace net.devh:grpc-spring-boot-starter dependency in grpc module with org.springframework.grpc:spring-grpc-spring-boot-starter in order to support Spring Boot 4

**Related issue(s)**:

Fixes #12452 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/grpc/src/main/java/org/hiero/mirror/grpc/config/GrpcHealthIndicator.java
+++ b/grpc/src/main/java/org/hiero/mirror/grpc/config/GrpcHealthIndicator.java
@@ -2,7 +2,6 @@
 
 package org.hiero.mirror.grpc.config;
 
-import jakarta.annotation.PreDestroy;
 import jakarta.inject.Named;
 import java.util.concurrent.atomic.AtomicReference;
 import lombok.CustomLog;
@@ -11,8 +10,9 @@ import org.springframework.boot.actuate.health.Health;
 import org.springframework.boot.actuate.health.HealthIndicator;
 import org.springframework.boot.actuate.health.Status;
 import org.springframework.context.event.EventListener;
-import org.springframework.grpc.autoconfigure.server.GrpcServerProperties;
+import org.springframework.grpc.server.lifecycle.GrpcServerShutdownEvent;
 import org.springframework.grpc.server.lifecycle.GrpcServerStartedEvent;
+import org.springframework.grpc.server.lifecycle.GrpcServerTerminatedEvent;
 
 @CustomLog
 @Named
@@ -20,22 +20,27 @@ import org.springframework.grpc.server.lifecycle.GrpcServerStartedEvent;
 public class GrpcHealthIndicator implements HealthIndicator {
 
     private final AtomicReference<Status> status = new AtomicReference<>(Status.UNKNOWN);
-    private final GrpcServerProperties grpcServerProperties;
 
     @Override
     public Health health() {
         return Health.status(status.get()).build();
     }
 
-    @EventListener(GrpcServerStartedEvent.class)
-    public void onStart() {
-        log.info("Started gRPC server on {}", grpcServerProperties.getAddress());
+    @EventListener
+    public void onStart(GrpcServerStartedEvent event) {
+        log.info("Started gRPC server on {}:{}", event.getAddress(), event.getPort());
         status.set(Status.UP);
     }
 
-    @PreDestroy
-    public void onStop() {
+    @EventListener
+    public void onStop(GrpcServerShutdownEvent event) {
         log.info("Stopping gRPC server");
         status.set(Status.OUT_OF_SERVICE);
+    }
+
+    @EventListener
+    public void onTermination(GrpcServerTerminatedEvent event) {
+        log.info("Stopped gRPC server");
+        status.set(Status.DOWN);
     }
 }


build.gradle.kts
```diff
@@ -72,7 +72,8 @@ dependencies {
         api("io.vertx:vertx-web:4.5.22") // Temporary until next Fabric8 version
         api("jakarta.inject:jakarta.inject-api:2.0.1")
         api("javax.inject:javax.inject:1")
-        api("net.devh:grpc-spring-boot-starter:3.1.0.RELEASE")
+        api("org.springframework.grpc:spring-grpc-spring-boot-starter:0.12.0")
+        api("org.springframework.grpc:spring-grpc-test:0.12.0")
```
steven-sheehy: Store version as a variable and reuse for both.

steven-sheehy: Sort with other packages

grpc/build.gradle.kts
```diff
@@ -17,7 +17,7 @@ dependencies {
     implementation("io.projectreactor.addons:reactor-extra")
     implementation("io.projectreactor:reactor-core-micrometer")
     implementation("jakarta.inject:jakarta.inject-api")
-    implementation("net.devh:grpc-spring-boot-starter")
+    implementation("org.springframework.grpc:spring-grpc-spring-boot-starter")
```
steven-sheehy: Sort with other packages

grpc/src/main/java/org/hiero/mirror/grpc/config/GrpcHealthIndicator.java
```diff
@@ -2,45 +2,40 @@
...
 import org.springframework.boot.actuate.health.HealthIndicator;
 import org.springframework.boot.actuate.health.Status;
+import org.springframework.boot.context.event.ApplicationReadyEvent;
 import org.springframework.context.event.EventListener;
+import org.springframework.grpc.autoconfigure.server.GrpcServerProperties;
 
 @CustomLog
 @Named
 @RequiredArgsConstructor
 public class GrpcHealthIndicator implements HealthIndicator {
```
steven-sheehy, line 20: It seems like there might be a built-in [health](https://github.com/spring-projects/spring-grpc/blob/main/spring-grpc-docs/src/main/antora/modules/ROOT/pages/health.adoc) indicator with spring-grpc. Please double check it covers our needs and remove this and any application.yml config.

grpc/src/main/java/org/hiero/mirror/grpc/controller/ConsensusController.java
```diff
@@ -31,19 +31,21 @@
...
     @Override
-    public Flux<ConsensusTopicResponse> subscribeTopic(Mono<ConsensusTopicQuery> request) {
-        return request.map(this::toFilter)
+    public void subscribeTopic(ConsensusTopicQuery request, StreamObserver<ConsensusTopicResponse> responseObserver) {
+        Mono.fromCallable(() -> toFilter(request))
                 .flatMapMany(topicMessageService::subscribeTopic)
                 .map(this::toResponse)
-                .onErrorMap(ProtoUtil::toStatusRuntimeException);
+                .onErrorMap(ProtoUtil::toStatusRuntimeException)
+                .doFinally(signalType -> log.debug("subscribeTopic stream ended with signal: {}", signalType))
```
steven-sheehy: We can remove this log as it was not present before.

grpc/src/main/java/org/hiero/mirror/grpc/controller/ConsensusController.java
```diff
@@ -31,19 +31,21 @@
...
     static final int DEFAULT_RUNNING_HASH_VERSION = 3;
 
     private final TopicMessageService topicMessageService;
 
     @Override
-    public Flux<ConsensusTopicResponse> subscribeTopic(Mono<ConsensusTopicQuery> request) {
-        return request.map(this::toFilter)
+    public void subscribeTopic(ConsensusTopicQuery request, StreamObserver<ConsensusTopicResponse> responseObserver) {
+        Mono.fromCallable(() -> toFilter(request))
                 .flatMapMany(topicMessageService::subscribeTopic)
```
steven-sheehy, line 44: `toFilter()` doesn't need to be part of the flux chain anymore.

```suggestion
        final var filter = toFilter(request);
        topicMessageService.subscribeTopic(filter)
```

grpc/src/main/java/org/hiero/mirror/grpc/controller/NetworkController.java
```diff
@@ -4,36 +4,38 @@
...
+public class NetworkController extends NetworkServiceGrpc.NetworkServiceImplBase {
 
     private final NetworkService networkService;
 
     @Override
-    public Flux<NodeAddress> getNodes(Mono<AddressBookQuery> request) {
-        return request.map(this::toFilter)
+    public void getNodes(AddressBookQuery request, StreamObserver<NodeAddress> responseObserver) {
+        Mono.fromCallable(() -> toFilter(request))
                 .flatMapMany(networkService::getNodes)
```
steven-sheehy, line 34:
```suggestion
        final var filter = toFilter(request);
        networkService.getNodes(filter)
```

grpc/src/main/java/org/hiero/mirror/grpc/controller/NetworkController.java
```diff
@@ -4,36 +4,38 @@
...
                 .map(this::toNodeAddress)
-                .onErrorMap(ProtoUtil::toStatusRuntimeException);
+                .onErrorMap(ProtoUtil::toStatusRuntimeException)
+                .doFinally(signalType -> log.debug("getNodes stream ended with signal: {}", signalType))
```
steven-sheehy:
```suggestion
```

grpc/src/main/resources/application.yml
steven-sheehy, line 1: Need to delete `grpc.server` config that's specific to old library and move its values to somewhere else that's relevant.

grpc/src/main/resources/application.yml
```diff
@@ -36,7 +36,7 @@ management:
         liveness:
           include: ping
         readiness:
-          include: db, grpc, grpcChannel, ping, redis
+          include: db, grpc, ping, redis
```
steven-sheehy, line 33: Not sure if we have to use `spring.grpc.server.health.actuator.health-indicator-paths`.

grpc/src/test/java/org/hiero/mirror/grpc/GrpcIntegrationTest.java
```diff
@@ -14,7 +14,13 @@
 import org.springframework.transaction.support.TransactionOperations;
 import org.springframework.transaction.support.TransactionTemplate;
 
-@Import({Configuration.class, RedisTestConfiguration.class, GrpcTestConfiguration.class})
+@SpringBootTest(
+        properties = {
+            "spring.grpc.server.enabled=true",
```
steven-sheehy: Isn't this enabled by default and can be omitted?

grpc/src/test/java/org/hiero/mirror/grpc/config/GrpcTestConfiguration.java
```diff
@@ -2,14 +2,31 @@
...
 import org.springframework.context.annotation.Bean;
+import org.springframework.grpc.client.GrpcChannelFactory;
 
 @TestConfiguration
 public class GrpcTestConfiguration {
+
+    private static final String IN_PROCESS_SERVER_NAME = "test-server";
+
+    @Bean
+    ReactorNetworkServiceGrpc.ReactorNetworkServiceStub networkReactiveStub(GrpcChannelFactory channels) {
```
steven-sheehy: We should not have reactor stubs. Need to remove `com.salesforce.servicelibs:reactor-grpc` dependency everywhere.

grpc/src/test/java/org/hiero/mirror/grpc/GrpcIntegrationTest.java
```diff
@@ -14,7 +14,13 @@
...
+            "spring.grpc.server.in-process.name=test-server",
+            "spring.grpc.client.local.address=in-process:test-server"
```
steven-sheehy: Shouldn't need this. Use `@AutoConfigureInProcessTransport` instead.

grpc/src/test/java/org/hiero/mirror/grpc/config/GrpcTestConfiguration.java
```diff
@@ -2,14 +2,31 @@
...
+        return ReactorNetworkServiceGrpc.newReactorStub(channels.createChannel("in-process:" + IN_PROCESS_SERVER_NAME));
+    }
+
+    @Bean
+    ConsensusServiceGrpc.ConsensusServiceBlockingStub consensusBlockingStub(GrpcChannelFactory channels) {
+        return ConsensusServiceGrpc.newBlockingStub(channels.createChannel("in-process:" + IN_PROCESS_SERVER_NAME));
```
steven-sheehy: Shouldn't need this or this entire test config with `@AutoConfigureInProcessTransport`.

grpc/src/test/java/org/hiero/mirror/grpc/controller/NetworkControllerTest.java
```diff
@@ -27,19 +25,22 @@
...
-    private final SystemEntity systemEntity;
+    @Autowired
+    private DomainBuilder domainBuilder;
 
-    @GrpcClient("local")
+    @Autowired
+    private SystemEntity systemEntity;
+
+    @Autowired
     private ReactorNetworkServiceGrpc.ReactorNetworkServiceStub reactiveService;
```
steven-sheehy: Need to remove reactor stubs and any tests associated with it. If regular stubs don't have equal test coverage to reactive then convert reactive tests to regular.

grpc/src/main/java/org/hiero/mirror/grpc/config/GrpcConfiguration.java
```diff
@@ -29,31 +26,26 @@ TransactionOperations transactionOperationsReadOnly(PlatformTransactionManager t
...
-        }
+        return serverBuilder -> {
+            if (serverBuilder instanceof NettyServerBuilder nettyServerBuilder) {
+                nettyServerBuilder
+                        .executor(grpcVirtualThreadExecutor)
+                        .maxConnectionIdle(
+                                nettyProperties.getMaxConnectionIdle().toSeconds(), TimeUnit.SECONDS)
+                        .maxConcurrentCallsPerConnection(nettyProperties.getMaxConcurrentCallsPerConnection())
+                        .maxInboundMessageSize(nettyProperties.getMaxInboundMessageSize())
+                        .maxInboundMetadataSize(nettyProperties.getMaxInboundMetadataSize());
```
steven-sheehy: We can remove some of our netty specific properties as they have spring-grpc generic [properties](https://docs.spring.io/spring-grpc/reference/appendix.html) we can use.

grpc/src/main/java/org/hiero/mirror/grpc/config/GrpcConfiguration.java
```diff
@@ -29,31 +26,26 @@ TransactionOperations transactionOperationsReadOnly(PlatformTransactionManager t
...
-                    .executor(executor)
-                    .maxConnectionIdle(nettyProperties.getMaxConnectionIdle().toSeconds(), TimeUnit.SECONDS)
-                    .maxConcurrentCallsPerConnection(nettyProperties.getMaxConcurrentCallsPerConnection())
-                    .maxInboundMessageSize(nettyProperties.getMaxInboundMessageSize())
-                    .maxInboundMetadataSize(nettyProperties.getMaxInboundMetadataSize());
-        }
+        return serverBuilder -> {
+            if (serverBuilder instanceof NettyServerBuilder nettyServerBuilder) {
+                nettyServerBuilder
+                        .executor(grpcVirtualThreadExecutor)
```
steven-sheehy: I doubt we need to customize netty with our own virtual thread executor. That should be done automatically if you enable virtual threads property. Please debug and check.

grpc/src/main/java/org/hiero/mirror/grpc/config/GrpcHealthIndicator.java
sdimitrov9: Correct, gRPC Health service (grpc.health.v1.Health) is already built-in.

Screenshot 2026-01-14 at 10 10 12

grpc/src/main/java/org/hiero/mirror/grpc/controller/ConsensusController.java
sdimitrov9: Removing toFilter() from the chain means that onErrorMap won't intercept exceptions thrown in it. 
We would have to catch exceptions manually like: 

        final TopicMessageFilter filter;
        try {
            filter = toFilter(request);
        } catch (Exception e) {
            responseObserver.onError(ProtoUtil.toStatusRuntimeException(e));
            return;
        }

otherwise we will experience some test failures like this for the _invalidTopicID_ test:

```
Expecting throwable message:
  "UNKNOWN: Application error processing RPC"
to contain:
  "Invalid entity ID"
but did not.
```

grpc/src/main/resources/application.yml
sdimitrov9: If we use it:

```  grpc:
    server:
      health:
        actuator:
          health-indicator-paths:
            - readiness
 ```
            
 the grpc service health check will return "NOT_SERVING" if the service is not able to connect to Redis for example, but will still respond successfully to client requests:

`WARN loomBoundedElastic-38 o.s.b.a.d.r.RedisReactiveHealthIndicator Redis health check failed org.springframework.data.redis.RedisConnectionFailureException: Unable to connect to Redis` (new attempt is made every 5 seconds)

`INFO virtual-289 o.h.m.g.s.TopicMessageServiceImpl Subscribing to topic: TopicMessageFilter(endTime=1678890000000000000, limit=10, startTime=1678880000000000000, subscriberId=AGgDN4ou, topicId=0.0.1001) `

This doesn't seem to be the current behaviour or at least the logging about Redis being unavailable is not present.

File: grpc/src/main/java/org/hiero/mirror/grpc/config/GrpcHealthIndicator.java (hunk 3 above)
Comment (steven-sheehy) [reply grpc/src/main/java/org/hiero/mirror/grpc/config/GrpcHealthIndicator.java:20]: That's grpc protocol. I meant if there's an actuator grpc health indicator

File: grpc/src/main/resources/application.yml [hunk 16]
```diff
@@ -36,7 +30,7 @@ management:
         liveness:
           include: ping
         readiness:
-          include: db, grpc, grpcChannel, ping, redis
+          include: db, ping, redis
```
jnels124: do we need to include `readinessState` and/or `grpcServer ` or similar to include the spring grpc health check here?

sdimitrov9: We may need to add the below block, but with it the service is spamming db/redis unavailability logs every 5 seconds(if any of them is actually unavailable) and in the meantime grpc continue to respond successfully when getNodes or subscribeTopic calls are made, while the service status is "NOT_SERVING". What is the behaviour we are looking for?

  ```
spring:
    grpc:
      server:
        health:
          actuator:
            health-indicator-paths:
              - db
              - redis
```

File: grpc/src/test/java/org/hiero/mirror/grpc/GrpcIntegrationTest.java [hunk 17]
```diff
@@ -4,17 +4,19 @@
...
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Import;
 import org.springframework.context.annotation.Primary;
+import org.springframework.grpc.test.AutoConfigureInProcessTransport;
 import org.springframework.transaction.PlatformTransactionManager;
 import org.springframework.transaction.support.TransactionOperations;
 import org.springframework.transaction.support.TransactionTemplate;
 
-@Import({Configuration.class, RedisTestConfiguration.class, GrpcTestConfiguration.class})
+@SpringBootTest
```
steven-sheehy: Can omit `@SpringBootTest` since it's on parent.

grpc/src/main/resources/application.yml
jnels124: as long as the health reports down when it is not able to stream results that is the only behavior that i can see that matters. Just not sure if those healthchecks mentioned were evaluated. I will do some testing with it.

grpc/src/test/java/org/hiero/mirror/grpc/controller/NetworkControllerTest.java
```diff
@@ -27,39 +27,50 @@
 import org.hiero.mirror.grpc.GrpcIntegrationTest;
 import org.hiero.mirror.grpc.util.ProtoUtil;
 import org.junit.jupiter.api.Test;
-import reactor.core.publisher.Mono;
-import reactor.test.StepVerifier;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.grpc.client.ImportGrpcClients;
 
-@RequiredArgsConstructor
+@ImportGrpcClients(types = {NetworkServiceGrpc.NetworkServiceBlockingStub.class})
```
steven-sheehy: Code smell
```suggestion
@ImportGrpcClients(types = NetworkServiceGrpc.NetworkServiceBlockingStub.class)
```

grpc/src/main/java/org/hiero/mirror/grpc/config/GrpcConfiguration.java
sdimitrov9: Only maxConcurrentCallsPerConnection is now left as netty specific property.

grpc/src/main/resources/application.yml
sdimitrov9: Updated the application.yml - the readiness check should fail on db,redis unavailability.

grpc/src/main/resources/application.yml
sdimitrov9: `spring.grpc.server.health.actuator.health-indicator-paths` is now added, the readiness check will fail on db/redis unavailability

grpc/src/main/java/org/hiero/mirror/grpc/config/GrpcHealthIndicator.java
sdimitrov9: spring.grpc.server.health.actuator.health-indicator-paths is now added, the readiness check will fail on db/redis unavailability

grpc/src/main/resources/application.yml
```diff
@@ -74,6 +68,19 @@ spring:
       maximum-pool-size: 50
       minimum-idle: 4
       validation-timeout: 3000
+  grpc:
+    server:
+      health:
+        actuator:
```
jnels124: not sure what the default is, but maybe we should add `enabled: true` under actuator to be explicit

jnels124: looking a bit more, I think this is only for the grpc client right? It appears this doesn't automatically include this in the health check used by kubernetes `/actuator/health/readiness`

grpc/src/main/java/org/hiero/mirror/grpc/config/GrpcConfiguration.java
```diff
@@ -30,30 +26,19 @@ TransactionOperations transactionOperationsReadOnly(PlatformTransactionManager t
...
-                            .setNameFormat("grpc-executor-%d")
-                            .build());
 
-            nettyServerBuilder
-                    .executor(executor)
-                    .maxConnectionIdle(nettyProperties.getMaxConnectionIdle().toSeconds(), TimeUnit.SECONDS)
-                    .maxConcurrentCallsPerConnection(nettyProperties.getMaxConcurrentCallsPerConnection())
-                    .maxInboundMessageSize(nettyProperties.getMaxInboundMessageSize())
-                    .maxInboundMetadataSize(nettyProperties.getMaxInboundMetadataSize());
+            nettyServerBuilder.maxConcurrentCallsPerConnection(nettyProperties.getMaxConcurrentCallsPerConnection());
```
xin-hedera: looks like with the changes in `grpcServerConfigurer`, `serverBuilder` is now always a `NettyServerBuilder`, so the `instanceof` should not be needed anymore?

grpc/src/main/resources/application.yml
sdimitrov9: Default value for spring.grpc.server.health.actuator.enabled is indeed true.

steven-sheehy:
build conflict

build.gradle.kts
```diff
@@ -88,9 +86,11 @@ dependencies {
         api("org.hyperledger.besu:evm:$besuVersion")
         api("org.mapstruct:mapstruct:$mapStructVersion")
         api("org.mapstruct:mapstruct-processor:$mapStructVersion")
+        api("org.mockito:mockito-inline:5.2.0")
```
steven-sheehy: This dependency has gone away

grpc/src/test/java/org/hiero/mirror/grpc/controller/NetworkControllerTest.java
```diff
@@ -27,39 +27,50 @@
...
 final class NetworkControllerTest extends GrpcIntegrationTest {
 
-    private static final Duration WAIT = Duration.ofSeconds(10L);
     private static final long CONSENSUS_TIMESTAMP = 1L;
 
-    private final DomainBuilder domainBuilder;
-    private final SystemEntity systemEntity;
+    @Autowired
+    private DomainBuilder domainBuilder;
```
steven-sheehy: Can remove since it's in parent

grpc/src/test/java/org/hiero/mirror/grpc/controller/NetworkControllerTest.java
```diff
@@ -27,39 +27,50 @@
...
 
-    @GrpcClient("local")
-    private ReactorNetworkServiceGrpc.ReactorNetworkServiceStub reactiveService;
+    @Autowired
+    private SystemEntity systemEntity;
```
steven-sheehy: Remove since it's in parent

grpc/src/test/java/org/hiero/mirror/grpc/controller/NetworkControllerTest.java
```diff
@@ -27,39 +27,50 @@
...
+
+    @Autowired
+    private NetworkServiceGrpc.NetworkServiceBlockingStub blockingService;
```
steven-sheehy: Why can't we make final, remove `@Autowired` and use `@RequiredArgsConstructor`?

grpc/src/test/java/org/hiero/mirror/grpc/controller/ConsensusControllerTest.java
```diff
@@ -35,21 +35,16 @@
...
 
-    private static final Duration WAIT = Duration.ofSeconds(10L);
     private final long future = DomainUtils.convertToNanosMax(Instant.now().plusSeconds(10L));
 
-    @GrpcClient("local")
-    private ReactorConsensusServiceGrpc.ReactorConsensusServiceStub grpcConsensusService;
-
-    @GrpcClient("local")
+    @Autowired
     private ConsensusServiceGrpc.ConsensusServiceBlockingStub blockingService;
```
steven-sheehy: Make final, remove `@Autowired`, and use `@RequiredArgsConstructor`

grpc/src/main/resources/application.yml
```diff
@@ -74,6 +68,21 @@ spring:
...
+          health-indicator-paths:
+            - db
+            - ping
+            - redis
```
steven-sheehy: We don't want the `grpc.health.v1.Health/Check` active. We only want the `/actuator/health/*` endpoints. Please set `spring.grpc.server.health.enabled=false` and remove the health-indicator-paths. We need to resurrect the `GrpcHealthIndicator` and make sure it works with actuator health.

grpc/src/main/java/org/hiero/mirror/grpc/config/NettyProperties.java
steven-sheehy: Unused. Remove here and in docs.

steven-sheehy: Unused. Remove here and in docs.

grpc/src/main/java/org/hiero/mirror/grpc/config/NettyProperties.java
```diff
@@ -21,19 +21,9 @@ public class NettyProperties {
     @Max(10000)
     private int executorMaxThreadCount = 1000;
```
steven-sheehy: Unused. Remove here and in docs.

grpc/src/main/resources/application.yml
sdimitrov9: Done

build.gradle.kts
```diff
@@ -70,7 +69,7 @@ dependencies {
         api("io.projectreactor:reactor-core-micrometer:1.2.12")
         api("io.vertx:vertx-web:4.5.22") // Temporary until next Fabric8 version
         api("jakarta.inject:jakarta.inject-api:2.0.1")
-        api("net.devh:grpc-spring-boot-starter:3.1.0.RELEASE")
+        api("javax.inject:javax.inject:1")
```
steven-sheehy: This dependency was removed previously but you resurrected it.

build.gradle.kts
```diff
@@ -86,6 +85,9 @@ dependencies {
         api("org.mapstruct:mapstruct:$mapStructVersion")
         api("org.mapstruct:mapstruct-processor:$mapStructVersion")
         api("org.msgpack:jackson-dataformat-msgpack:0.9.11")
+        api("org.springdoc:springdoc-openapi-webflux-ui:1.8.0")
```
steven-sheehy: Remove this dependency

grpc/src/main/java/org/hiero/mirror/grpc/config/GrpcHealthIndicator.java
```diff
@@ -2,45 +2,40 @@
...
 
     @Override
     public Health health() {
         return Health.status(status.get()).build();
     }
 
-    @EventListener
-    public void onStart(GrpcServerStartedEvent event) {
-        log.info("Started gRPC server on {}:{}", event.getAddress(), event.getPort());
+    @EventListener(ApplicationReadyEvent.class)
```
steven-sheehy: Application ready and grpc server ready events are not quite the same. There should be spring grpc events we can used I think even named the same. Otherwise we'll have to hook into some grpc lifecycle listener.

grpc/src/main/resources/application.yml
```diff
@@ -36,7 +30,7 @@ management:
         liveness:
           include: ping
         readiness:
-          include: db, grpc, grpcChannel, ping, redis
```
steven-sheehy, line 39: We still need grpc here if we have a grpchealthindicator

grpc/src/main/java/org/hiero/mirror/grpc/config/GrpcHealthIndicator.java
```diff
@@ -2,45 +2,40 @@
...
-        log.info("Started gRPC server on {}:{}", event.getAddress(), event.getPort());
+    @EventListener(GrpcServerStartedEvent.class)
+    public void onStart() {
+        log.info("Started gRPC server on {}", grpcServerProperties.getAddress());
         status.set(Status.UP);
     }
 
-    @EventListener
-    public void onStop(GrpcServerShutdownEvent event) {
+    @PreDestroy
```
steven-sheehy: Use GrpcServerShutdownEvent for onStop

sdimitrov9: Original class is fully restored, only imports are changed to reflect the new dependency.

PR #12761, opened 2026-01-13
Add per customer auth limits

### LINKED ISSUE #12716: Per customer response limits
### Problem

Some customers have requested custom limits for API responses.

### Solution

* Add a property
```yaml
hiero:
  mirror:
    rest:
      users:
        - password: bar
          limit: 200
          username: foo
```
* If `Authorization` header is passed verify it's a valid user and password in list inside a middleware. Reject if it's not a valid user and header is present.
* If validated, store the custom user limit in the httpContext
* Retrieve the limit from the httpContext wherever we currently use `response.limit.max`

### Alternatives

_No response_

martingeorgiev1:
**Description**:
Add per-customer authentication limits to the mirror node to enable customized rate limits for different users, improving flexibility and security for customer-specific access.

  - Add parsing and validation for users configuration in config.js
  - Add empty users array to application.yml and document in configuration.md
  - Add authHandler middleware to authenticate users and set custom limits in httpContext
  - Add getEffectiveMaxLimit() helper in utils.js to respect authenticated user limits
  - Update blockController to use getEffectiveMaxLimit() for dynamic limit handling
  - Add basic-auth and tsscmp packages for secure authentication
  - Add unit tests for config parsing, authHandler, and utils
  - Add integration tests for authentication with custom limits
  - Add spec tests for accounts, blocks, and transactions endpoints with authentication
  - Enhance spec test framework to support requestHeaders for authentication testing

**Related issue(s)**:

Fixes #12716 

**Checklist**


rest/middleware/authHandler.js
```diff
@@ -0,0 +1,74 @@
+// SPDX-License-Identifier: Apache-2.0
+
+import httpContext from 'express-http-context';
+
+import config from '../config.js';
+import {httpStatusCodes, userLimitLabel} from '../constants.js';
+
+const extractBasicAuthCredentials = (authHeader) => {
+  if (!authHeader || !authHeader.startsWith('Basic ')) {
```
xin-hedera: can we use https://www.npmjs.com/package/express-basic-auth? Or use the low level https://www.npmjs.com/package/basic-auth to parse the authorization header, with a timing safe string comparison module (e.g., https://www.npmjs.com/package/tsscmp) to validate the user & pass?

rest/utils.js
```diff
@@ -501,17 +502,21 @@ const parseOperatorAndValueFromQueryParam = (paramValue) => {
...
- * the last one is honored.
+ * Gets the limit param value, if not exists, return the default; otherwise cap it at max.
+ * Checks httpContext for authenticated user's custom limit first.
+ * Note if values is an array, the last one is honored.
  * @param {string[]|string} values Values of the limit param
  * @return {number}
  */
 const getLimitParamValue = (values) => {
+  const userLimit = httpContext.get(userLimitLabel);
+  const maxLimit = userLimit !== undefined ? userLimit : responseLimit.max;
```
xin-hedera: - max is only needed when `values !== undefined` so should move the logic inside the if block
- extract the two lines to its own method since it's duplicated in `formatComparator`

xin-hedera (changes requested):

IvanKavaldzhiev (dismissed):
LGTM

rest/config/application.yml
```diff
@@ -148,6 +148,11 @@ hiero:
           tokenBalance:
             multipleAccounts: 50
             singleAccount: 1000
+      users:
+        []
```
jnels124: nit: move array inline

martingeorgiev1: spotless was automatically moving it to new line because of the comment, but that reminded me that the example must be in `configuration.md`, this one is resolved and the comment moved to config

rest/controllers/blockController.js
```diff
@@ -53,7 +53,7 @@ class BlockController extends BaseController {
 
   extractLimitFromFilters = (filters) => {
     const limit = _.findLast(filters, {key: filterKeys.LIMIT});
-    return limit ? (limit.value > maxLimit ? defaultLimit : limit.value) : defaultLimit;
```
steven-sheehy, line 56: Why are we removing this?

rest/controllers/blockController.js
```diff
@@ -11,7 +11,7 @@ import {RecordFileService} from '../service';
 import * as utils from '../utils';
 import {BlockViewModel} from '../viewmodel';
 
-const {default: defaultLimit, max: maxLimit} = getResponseLimit();
+const {default: defaultLimit} = getResponseLimit();
```
steven-sheehy, line 15: We should update `getResponseLimit()` to get custom limit from context.

rest/utils.js
```diff
@@ -501,8 +502,18 @@ const parseOperatorAndValueFromQueryParam = (paramValue) => {
 };
 
 /**
- * Gets the limit param value, if not exists, return the default; otherwise cap it at max. Note if values is an array,
- * the last one is honored.
+ * Gets the effective max limit for the current request, checking httpContext
+ * for authenticated user's custom limit first, falling back to default max.
+ * @return {number}
+ */
+const getEffectiveMaxLimit = () => {
```
steven-sheehy, line 509: We should have spec tests for a number of list endpoints that set user auth header and config and verify custom limit. Can set default and max limit to one in config so can easily go beyond it.

steven-sheehy (changes requested):

rest/utils.js
steven-sheehy: I don't see any spec tests in latest commit.

jnels124 (dismissed):

